PR #3020, opened 2023-06-15
[SM-771] Add new endpoint for bulk enabling users for Secrets Manager

Thomas-Avery:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The purpose of this PR is to provide a new endpoint for bulk enabling users for secrets manager.

## Code changes

* **src/Api/Controllers/OrganizationUsersController.cs:**
Adding in new endpoint for bulk enabling users for secrets manager.

* **src/Core/OrganizationFeatures/OrganizationServiceCollectionExtensions.cs:**
Adding `EnableAccessSecretsManagerCommand` into dependency injection.
Since this is a command for the organization user controller, I thought this was the appropriate place for DI.

* **src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs:**
Command for bulk enabling users for secrets manager.

* **src/Core/SecretsManager/Commands/EnableAccessSecretsManager/Interfaces/IEnableAccessSecretsManagerCommand.cs:**
Interface for the command.

* **test/Core.Test/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommandTests.cs:**
Unit tests for the new command.

## Before you submit


src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+public class EnableAccessSecretsManagerCommand : IEnableAccessSecretsManagerCommand
+{
+    private readonly IOrganizationUserRepository _organizationUserRepository;
+
+    public EnableAccessSecretsManagerCommand(IOrganizationUserRepository organizationUserRepository)
+    {
+        _organizationUserRepository = organizationUserRepository;
+    }
+
+    public async Task<List<(OrganizationUser, string)>> EnableUsersAsync(Guid organizationId,
```
eliykat, line 17: Please use named tuples, especially because the `string` property is not very descriptive (I had to look it up):

```suggestion
    public async Task<List<(OrganizationUser orgUser, string error)>> EnableUsersAsync(Guid organizationId,
```

src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        var filteredUsersByOrg = organizationUsers.Where(ou => ou.OrganizationId == organizationId)
+            .ToList();
+
+        if (!filteredUsersByOrg.Any())
+        {
+            throw new BadRequestException("Users invalid.");
+        }
+
+        var filteredUsers = filteredUsersByOrg.Where(ou => ou.AccessSecretsManager == false).ToList();
```
eliykat, line 28: This is where we would usually set the `string error` property for that orgUser, e.g. "User already has access to Secrets Manager", and include it in the return value.

Then that error message displays in the bulk actions dialog on the front-end, next to that user. (You can see this in other bulk operations today, e.g. bulk confirm users who have already been confirmed.)

src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+    public EnableAccessSecretsManagerCommand(IOrganizationUserRepository organizationUserRepository)
+    {
+        _organizationUserRepository = organizationUserRepository;
+    }
+
+    public async Task<List<(OrganizationUser, string)>> EnableUsersAsync(Guid organizationId,
+        ICollection<OrganizationUser> organizationUsers)
+    {
+        var filteredUsersByOrg = organizationUsers.Where(ou => ou.OrganizationId == organizationId)
+            .ToList();
```
eliykat, line 21: I'm not 100% sure on this, but I wonder if this validation should be done outside the command. Do we tell the command, "enable SM for these orgUsers for this organization"? Or should we just tell it, "enable SM for these orgUsers"? Specifying the organization seems redundant; should the command just trust what's passed into it?

I prefer the latter, but I don't know if there's a right or wrong answer, so it's your call. Just thought I'd mention it for discussion.

eliykat (changes requested):
This is only a high level review on a couple of things that jumped out at me, so please get a second review from SM Team as well. Thanks!

src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs
Thomas-Avery: Thanks for taking a look! 

Resolved in https://github.com/bitwarden/server/pull/3020/commits/015c68aa31ccdeb0b246f7d3a903eb1312174b3d

src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs
Thomas-Avery: Resolved in https://github.com/bitwarden/server/pull/3020/commits/015c68aa31ccdeb0b246f7d3a903eb1312174b3d

src/Core/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommand.cs
Thomas-Avery: I think it makes sense to take that out of the command. It really is a query concern when thinking about the CQRS pattern. 

Resolved in https://github.com/bitwarden/server/pull/3020/commits/015c68aa31ccdeb0b246f7d3a903eb1312174b3d

src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -452,6 +456,29 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRest
...
+        if (!await _currentContext.ManageUsers(orgId))
+        {
+            throw new NotFoundException();
+        }
+
+        var orgUsers = (await _organizationUserRepository.GetManyAsync(model.Ids))
+            .Where(ou => ou.OrganizationId == orgId).ToList();
+        if (orgUsers.Count == 0)
+        {
+            throw new BadRequestException("Users invalid.");
```
coltonhurst, line 473: I would have expected this to be a 404 not found. Is there anything specific we are trying to accomplish by using `BadRequestException`?

coltonhurst (commented):
Looking good, only one comment.

src/Api/Controllers/OrganizationUsersController.cs
Thomas-Avery: I was going with what was pre-established by the other bulk requests. 

Specifically, the bulk delete.

https://github.com/bitwarden/server/blob/ca7ced4e435823edc677efcced06f5689a2e8754/src/Core/Services/Implementations/OrganizationService.cs#L1640-L1647

coltonhurst: Makes sense, thanks for pointing that out. As usual we can leave it for consistency and address the responses in a different PR if we want.

test/Core.Test/SecretsManager/Commands/EnableAccessSecretsManager/EnableAccessSecretsManagerCommandTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Bit.Core.Test.SecretsManager.Commands.EnableAccessSecretsManager;
+
+[SutProviderCustomize]
+public class EnableAccessSecretsManagerCommandTests
+{
+    [Theory]
+    [BitAutoData]
+    public async Task EnableUsers_UsersAlreadyEnabled_DoesNotCallRepository(
+        SutProvider<EnableAccessSecretsManagerCommand> sutProvider, ICollection<OrganizationUser> data)
```
eliykat, line 18: `data` is not very descriptive, I suggest keeping the same name as the parameter in the method under test. i.e. `organizationUsers`. Then it's immediately obvious what arguments you're arranging and asserting.

eliykat (approved):
Looks good to me, just 1 minor comment about the tests. Thanks for handling!

coltonhurst (approved):

PR #2845, opened 2023-04-11
[AC-1387] Resource-based authentication for Groups

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

This is a proposal for us to adopt [Resource-based authentication](https://learn.microsoft.com/en-us/aspnet/core/security/authorization/resourcebased?view=aspnetcore-6.0) to manage our permission checks on the server.

### Outline

The basic outline of this is:
* we have **resources**, which in our case includes ciphers, groups, organizations, etc
* we define **operations**, such as create/read/update/delete
* we define **authentication handlers** which contain the business logic deciding whether a user can perform an operation on a resource

Controllers can then call `_authorizationService.AuthorizeAsync(user, resource, operation)` and get back a success or failure result.

### Implementation notes

I've implemented this on the Groups controller as a proof of concept.

A few further notes on choices I made here (opinionated):
* **operations** should be granular and specific to the resource. For example, I've made "update group", "add members to group", and "remove members from group" all separate operations, even if they share the same code in the handler. This makes it easy to use the correct operation when calling `authorizationService`.
* **authentication handlers** should replace the role and permission logic in `CurrentContext`. In particular, I think the getters on `CurrentContext` can obscure what permissions are really being granted to whom. In this model, `CurrentContext` is mostly used to construct objects from claims and make them available for auth handlers. 
* I've added a proper exception we can throw where resource authorization fails, instead of incorrectly using `NotFoundException` (unless there's a reason we use this?)

### Benefits

I think this is an improvement because:
1. authorization logic is currently spread between controller, services, and `currentContext` getters. `CurrentContext` exposes a mixture of permissions (`ManageGroups`) and roles (`OrganizationAdmin`). This is difficult to keep track of and change. Authorization handlers provide a central location to manage permissions for each resource.
2. our authorization logic is quickly becoming more complex with upcoming initiatives and needs to be better encapsulated
3. `CurrentContext` has too many duties and has too much indirection in it (getters are several layers deep). Auth handlers can be tightly scoped to their resource **and particular operations** (not just "manage"), making them easy to understand and less risky to change
4. almost everything can be thought of as a resource: organizations, orgUsers, providers, ciphers, groups, collections, connections, etc. We can apply it most places
5. this is an ASP.NET-native pattern - we're not reinventing the wheel

### Drawbacks

[...]

PR #3023, opened 2023-06-16
Rewrite Icon fetching

MGibson1:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Rewrites icon fetching to be more object oriented.

## Code changes
- **DomainIcons**: Object controlling all icons retrieved for a given domain. Handles creation of potential uri locations for a given domain
- **IconHttpRequest**: Simplified HttpRequest class for Icon fetching. Handles limited redirect following and ensures use of validated IP addresses only.
- **IconHttpResponse**: HttpResponseMessage wrapper. Parses responses for Icon links, and retrieves them, collating into a list of Icons from a response.
- **IconLink**: Known link to an Icon. Retrives content and parses it into an Icon
- **IconUri**: Handles Icon validation and swap to IP
- **Startup**: Swap over to using `HttpClientFactory` with the same settings as the old static `HttpClient`
- **IPAddressExtension & UriExtension**: Useful validation and manipulation extensions
- **Tests**: Not much testing here, but I did update to match the new DI requirements and added a test that still uses `/favicon.ico` as the resolved icon location.

## Before you submit


justindbaur (changes requested):
Thoughts on passing through the `CancellationToken` from `HttpContext.RequestAborted` to all the async calls, this way if the  connection gets closed we can stop doing all this work. 

With something that has no database side effects I'd normally really want to do that but I can also see an argument against it because we can let the request continue and let it make it into our cache so that we can return the cached version the next time.

src/Icons/Models/IconHttpRequest.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        // Let's add some headers to look like we're coming from a web browser request. Some websites
+        // will block our request without these.
+        message.Headers.Add("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 " +
+            "(KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 Edge/16.16299");
+        message.Headers.Add("Accept-Language", "en-US,en;q=0.8");
+        message.Headers.Add("Cache-Control", "no-cache");
+        message.Headers.Add("Pragma", "no-cache");
+        message.Headers.Add("Accept", "text/html,application/xhtml+xml,application/xml;" +
+            "q=0.9,image/webp,image/apng,*/*;q=0.8");
```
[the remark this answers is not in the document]
MGibson1: I thought about doing that, but was worried about obfuscating. Happy you had the same thought, let's do it

src/Icons/Models/IconHttpResponse.cs
```diff
@@ -0,0 +1,81 @@
...
+    IEnumerator IEnumerable.GetEnumerator() => ((IEnumerable)_icons).GetEnumerator();
+
+    public IconHttpResponse(HttpResponseMessage response, ILogger<IIconFetchingService> logger, IHttpClientFactory httpClientFactory)
+    {
+        _response = response;
+        _logger = logger;
+        _httpClientFactory = httpClientFactory;
+    }
+
+    public async Task<IEnumerable<Icon>> RetrieveIconsAsync(Uri requestUri, string domain)
```
withinfocus, line 41: 💭 Does it make sense to just have this return one? Is there a use case I'm not seeing for providing all of what's found?

withinfocus (commented):
Great suggestions from Justin. I see a couple ⛏️ s beyond that but will skip calling them out until this settles.

src/Icons/Models/IconHttpRequest.cs
```diff
@@ -0,0 +1,121 @@
...
+    private async Task<HttpResponseMessage> GetAsync(IconUri iconUri)
+    {
+        using var message = new HttpRequestMessage();
+        message.RequestUri = iconUri.InnerUri;
+        message.Headers.Host = iconUri.Host;
+        message.Method = HttpMethod.Get;
+
+        // Let's add some headers to look like we're coming from a web browser request. Some websites
+        // will block our request without these.
+        message.Headers.Add("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 " +
```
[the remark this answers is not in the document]
MGibson1: updated to 114

src/Icons/Models/IconHttpResponse.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        var parser = new HtmlParser();
```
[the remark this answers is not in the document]
MGibson1: It [looks like](https://github.com/AngleSharp/AngleSharp/blob/main/src/AngleSharp/Html/Parser/HtmlParser.cs) a singleton should work fine, no state in the parser from what I can see.

src/Icons/Util/UriExtension.cs
```diff
@@ -0,0 +1,43 @@
+﻿
+#nullable enable
+
+using IdentityServer4.Extensions;
```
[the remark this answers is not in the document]
MGibson1: oops, using an extension that I shouldn't.

src/Icons/Models/IconLink.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        using var response = await IconHttpRequest.FetchAsync(uri, logger, httpClientFactory);
+        if (!response.IsSuccessStatusCode)
+        {
+            return null;
+        }
+
+        var format = response.Content.Headers.ContentType?.MediaType;
+        var bytes = await response.Content.ReadAsByteArrayAsync();
+        response.Content.Dispose();
```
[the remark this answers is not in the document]
MGibson1: I was surprised by this in the original code, too.

I read a few blogposts that suggested disposing content when you finish reading it, but aren't going to drop the response out of context for a while. In this example that probably doesn't make sense, since we're immediately returning.

If we run in to connection pool or memory issues we can look here, I guess.

src/Icons/Models/IconHttpResponse.cs
MGibson1: hmm this is pretty interesting. We used to prioritize icon display after pulling, but now that prioritization occurs prior to navigating to a link. That does mean that we could pull only the highest priority and cut down on network requests.

However, if the first link fails, we'd have to go on down the line iteratively. Currently, we just fire off all 10 requests simultaneously and wait for them all to get back to us. Effectively we're discussing changing this from an eager load to a lazy one.

I don't have the data to say which would be better, presumably most links work, but those that don't would suffer in load times.

src/Icons/Models/IconLink.cs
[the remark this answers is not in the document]
MGibson1: After @joseph-flinn 's comment that we have issues with Icon service memory management in prod already, I'd like to keep this in.

I realize it _should_ be cleaned, but calling dispose multiple times by definition does nothing bad.

test/Icons.Test/Services/IconFetchingServiceTests.cs
```diff
@@ -6,6 +6,7 @@ namespace Bit.Icons.Test.Services;
 public class IconFetchingServiceTests : ServiceTestBase<IconFetchingService>
 {
     [Theory]
+    [InlineData("www.twitter.com")] // https site
```
withinfocus, line 9: ❓ Anything special with how this test was handled? Seems the same as the line below.

withinfocus (commented):
Ton of changes here, but I think it's making sense to me and has a lot of test coverage. Few questions.

test/Icons.Test/Services/IconFetchingServiceTests.cs
MGibson1: From when I stepped through the retrieval it's the same as google, we have just been getting reports that twitter in particular is having problems.

Because this work is so pervasive I have no clue if this test passing represents a functionality change or not, but I wanted this as a canary. If it proves to be flaky on twitter's end I want to remove.

test/Common/MockedHttpClient/HttpRequestMatcher.cs
```diff
@@ -86,10 +86,10 @@ public MockedHttpResponse RespondWith(HttpStatusCode statusCode)
     /// Called to produce an HttpResponseMessage for the given request. This is probably something you want to leave alone
     /// </summary>
     /// <param name="request"></param>
-    public HttpResponseMessage RespondTo(HttpRequestMessage request)
+    public Task<HttpResponseMessage> RespondToAsync(HttpRequestMessage request)
     {
         NumberOfMatches++;
-        return _childMatcher == null ? _mockedResponse.RespondTo(request) : _childMatcher.RespondTo(request);
+        return _childMatcher == null ? _mockedResponse.RespondToAsync(request) : _childMatcher.RespondToAsync(request);
```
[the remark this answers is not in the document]
MGibson1: lol thank you, I got [nerd sniped by Justin](https://github.com/bitwarden/server/pull/2845#discussion_r1163388771) a few months ago with this and It almost just happened again!

withinfocus (dismissed):
Should wait for @justindbaur to review as well, but this is looking good to me and CI passes now. Some real-world testing in QA is a must.

src/Icons/Models/IconHttpResponse.cs
```diff
@@ -0,0 +1,83 @@
+﻿#nullable enable
+
+using System.Collections;
+using System.Net;
+using AngleSharp.Html.Parser;
+using Bit.Icons.Services;
+
+namespace Bit.Icons.Models;
+
+public class IconHttpResponse : IEnumerable<Icon>, IDisposable
```
justindbaur, line 10: Should this still implement `IEnumerable<Icon>`. It doesn't look like `_icons` is ever set.

src/Icons/Models/IconLink.cs
```diff
@@ -0,0 +1,212 @@
...
+using Bit.Icons.Extensions;
+using Bit.Icons.Services;
+
+namespace Bit.Icons.Models;
+
+public class IconLink
+{
+    private static readonly HashSet<string> _iconRels = new() { "icon", "apple-touch-icon", "shortcut icon" };
+    private static readonly HashSet<string> _blocklistedRels = new() { "preload", "image_src", "preconnect", "canonical", "alternate", "stylesheet" };
+    private static readonly HashSet<string> _iconExtensions = new() { ".ico", ".png", ".jpg", ".jpeg" };
```
justindbaur, line 14: If we pass `StringComparer.InvariantCultureIgnoreCase` into the constructor we don't have to `.ToLower()` the inputs into the `.Contains` calls and save a few string allocations.

src/Icons/Util/ServiceCollectionExtension.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            AllowAutoRedirect = false,
+            AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
+        });
+    }
+
+    public static void AddHtmlParsing(this IServiceCollection services)
+    {
+        services.AddSingleton<HtmlParser>();
+        services.AddSingleton<IHtmlParser>(s => s.GetRequiredService<HtmlParser>());
```
justindbaur, line 36: Unless we have a direct dependence on `HtmlParser` and if that's the case can we get rid of it?
```suggestion
        services.AddSingleton<IHtmlParser, HtmlParser>();
```

src/Icons/Util/UriExtension.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        if (Uri.TryCreate(uri, paths.First(), out var newUri))
+        {
+            return newUri.ConcatPath(paths.Skip(1));
+        }
+        else
+        {
+            return uri;
+        }
+    }
```
justindbaur, line 40: How badly do we want to make this `params`/`IEnumerable<string>` when we only append a single path? It's a helpful utility but it lives in `Icons` so isn't super likely to be reused.

test/Common/Helpers/HtmlBuilder.cs
justindbaur, line 1: Awesome!

test/Icons.Test/Models/IconHttpRequestTests.cs
```diff
@@ -0,0 +1,38 @@
...
+    [Fact]
+    public async Task FetchAsync_FollowsTwoRedirectsAsync()
+    {
+        var handler = new MockedHttpMessageHandler();
+
+        var request = handler
+            .Fallback
+            .WithStatusCode(HttpStatusCode.Redirect)
+            .WithContent("text/html", "<html><head><title>Redirect 2</title></head><body><a href=\"https://test.local\">Redirect 3</a></body></html>")
+            .WithHeader(HeaderNames.Location, "https://test.local");
```
justindbaur, line 23: I actually learned about special TLD's that are meant for testing/examples that I think we should use for URI's that we don't need to actually be real. https://www.rfc-editor.org/rfc/rfc2606.html#section-2

test/Icons.Test/Models/IconHttpRequestTests.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        var clientFactory = Substitute.For<IHttpClientFactory>();
+        clientFactory.CreateClient("Icons").Returns(handler.ToHttpClient());
+
+        var uriService = Substitute.For<IUriService>();
+        uriService.TryGetUri(Arg.Any<Uri>(), out Arg.Any<IconUri>()).Returns(x =>
+        {
+            x[1] = new IconUri(new Uri("https://test.local"), IPAddress.Parse("3.0.0.0"));
```
justindbaur, line 31: Same thing with special IP blocks https://www.rfc-editor.org/rfc/rfc5735.html#section-3

justindbaur (changes requested):

src/Icons/Models/IconLink.cs
MGibson1: good catch! I didn't think about it at all, just copied 😳

test/Icons.Test/Models/IconHttpRequestTests.cs
MGibson1: ah, I had actually thought `.local` was one...

src/Icons/Models/IconLink.cs
```diff
@@ -99,16 +99,16 @@ public bool IsUsable()
             return false;
         }
 
-        if (Rel != null && _iconRels.Contains(Rel.Value.ToLower()))
```
justindbaur: I did mean you can add that into the constructor of the `HashSet` but this is functionally equivalent so I'm fine with it if you are.

justindbaur (approved):
Great stuff!

src/Icons/Util/UriExtension.cs
justindbaur: I'm fine with it.

src/Icons/Models/IconLink.cs
MGibson1: ah my mistake, I forgot this _was_ a hashset

justindbaur (approved):

src/Icons/Models/IconLink.cs
```diff
@@ -23,12 +23,15 @@ public class IconLink
     private const string _jpegMediaType = "image/jpeg";
     private static readonly byte[] _jpegHeader = new byte[] { 255, 216, 255 };
 
+    private const string _svgXmlMediaType = "image/svg+xml";
```
withinfocus, line 26: 💭 Does there need to be a code expansion below to check the header? I see this done for the other types.

justindbaur: I don't think we need to, those are only for formats that we try to infer if we don't get a `Media-Type` header back or we don't support the media type we got back. I also don't think there are generally "marker" bytes for svg.

justindbaur (approved):

src/Icons/Models/IconLink.cs
withinfocus: Throwing in an additional test theory with a site that uses an SVG icon would be a nice and quick addition.

MGibson1: there doesn't need to be mime type detection on the file level. That's just a fallback if the mimetype doesn't exist.

I don't believe svg+xml has a mime type file header (I couldn't find one) so it's not added.

withinfocus (approved):

PR #3024, opened 2023-06-16
[PM-2594] Added new property "CloudRegion" to GlobalSettings and ConfigResponseModel to be able to override the cloud url value for selfhost instances

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
Output a new value through the `ConfigResponseModel` object using a property named `CloudRegion`, allowing selfhost clients to easily point to the cloud vault on the selfhost web vault instance. The value can be overridden using an environment variable.

## Code changes
This PR is tied to [this clients PR](https://github.com/bitwarden/clients/pull/5629).

- **src/Api/Models/Response/ConfigResponseModel.cs:** Added property `CloudRegion` to `EnvironmentConfigResponseModel` and set the value from `GlobalSettings`
- **src/Core/Settings/GlobalSettings.cs:** Added property `CloudRegion` to read the value from environment variables and if not present then a default value is returned depending if selfhosted
- **src/Core/Settings/IBaseServiceUriSettings.cs:** Added property `CloudRegion`

## Before you submit


src/Core/Settings/GlobalSettings.cs
```diff
@@ -146,6 +147,12 @@ public BaseServiceUriSettings(GlobalSettings globalSettings)
         public string Vault { get; set; }
         public string VaultWithHash => $"{Vault}/#";
 
+        public string CloudVault
+        {
+            get => string.IsNullOrWhiteSpace(_cloudVault) ? _globalSettings.SelfHosted ? "https://vault.bitwarden.com" : Vault : _cloudVault;
```
withinfocus, line 152: ♻️ Want to make sure this cleanup is filed as a task somewhere.

src/Api/Models/Response/ConfigResponseModel.cs
```diff
@@ -47,6 +48,7 @@ public class ServerConfigResponseModel
 public class EnvironmentConfigResponseModel
 {
     public string Vault { get; set; }
+    public string CloudVault { get; set; }
```
withinfocus, line 51: 💭 This name gives me pause and I think your team could chat on alternatives. Other names like "admin" or "management" come to mind and we should certainly ensure this is documented well for general usage.

r-tome: I have made a change to the code by renaming it to `CloudWebVault` to ensure consistency between the server and clients. 
The purpose of this URL is to reference the general cloud vault (vault.bitwarden.com/eu), while we already have the `Vault` environment variable. By using the name `CloudWebVault`, we can distinguish between these two variables more clearly.

src/Core/Settings/GlobalSettings.cs
r-tome: What kind of cleanup? Not relying on the `Vault` variable for a default value if `CloudWebVault` is not present?

withinfocus:
This looks good to me! Would like for your team to review and give the formal approval though.

vincentsalucci (approved):

eliykat (approved):

src/Api/Models/Response/ConfigResponseModel.cs
```diff
@@ -46,6 +47,7 @@ public class ServerConfigResponseModel
 
 public class EnvironmentConfigResponseModel
 {
+    public string CloudVaultRegion { get; set; }
```
eliykat, line 50: Just saw the discussion in Slack re: naming. I agree with @r-tome and @shane-melton that this doesn't need to refer to "Vault", especially because it's just sending a region and not a URL. It's only used to set the cloud vault URL currently, but it might be reused in the future, so a more general name could be a benefit.

I suggest `CloudRegion`.

vincentsalucci (approved):

PR #3028, opened 2023-06-19
[PM-1033] Org invite user creation flow 1

coroiu:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Remove user verification from password reset enrollment because it uses master password hash which does not exist when using TDE

Client PR: https://github.com/bitwarden/clients/pull/5611

## Before you submit


src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -313,17 +313,9 @@ public async Task PutResetPasswordEnrollment(string orgId, string userId, [FromB
...
-        }
-        else
-        {
-            var callingUserId = user.Id;
-            await _organizationService.UpdateUserResetPasswordEnrollmentAsync(
-               new Guid(orgId), new Guid(userId), model.ResetPasswordKey, callingUserId);
-        }
+        var callingUserId = user.Id;
+        await _organizationService.UpdateUserResetPasswordEnrollmentAsync(
+            new Guid(orgId), new Guid(userId), model.ResetPasswordKey, _userService, callingUserId);
```
JaredSnider-Bitwarden, line 318: ❓ : Should this change be feature flagged behind TDE?

coroiu: Hmm, good point, I'd personally don't think so but I'll check!

test/Api.Test/Controllers/OrganizationUsersControllerTests.cs
JaredSnider-Bitwarden, line 1: ❓ do we need any additional tests around the changed `[HttpPost("{organizationUserId}/accept")]` endpoint?

coroiu: I'm haven't actually made any changes to the accept endpoint, only some of the dependecies that the controller calls (which are mocked)

JaredSnider-Bitwarden: Argh, I'm sorry, the code was collapsed. I meant the changes to the `[HttpPut("{userId}/reset-password-enrollment")]` endpoint if there are any tests for those changes.

src/Core/Services/IOrganizationService.cs
```diff
@@ -55,7 +55,7 @@ public interface IOrganizationService
     Task<List<Tuple<OrganizationUser, string>>> DeleteUsersAsync(Guid organizationId,
         IEnumerable<Guid> organizationUserIds, Guid? deletingUserId);
     Task UpdateUserGroupsAsync(OrganizationUser organizationUser, IEnumerable<Guid> groupIds, Guid? loggedInUserId);
-    Task UpdateUserResetPasswordEnrollmentAsync(Guid organizationId, Guid userId, string resetPasswordKey, Guid? callingUserId);
+    Task UpdateUserResetPasswordEnrollmentAsync(Guid organizationId, Guid userId, string resetPasswordKey, IUserService userService, Guid? callingUserId);
```
justindbaur, line 58: Please do not send services through method parameters, if this would cause a circular dependency, then we need to move this method into a command.

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -1269,6 +1269,24 @@ public async Task<OrganizationUser> AcceptUserAsync(string orgIdentifier, User u
         return await AcceptUserAsync(orgUser, user, userService);
     }
 
+    public async Task<OrganizationUser> AcceptUserAsync(Guid organizationId, User user, IUserService userService)
```
justindbaur, line 1272: It doesn't appear this method is added to the interface, if it's not needed there it should be `private` or `internal` if used for testing and it has no need to take the service as a parameter if it exists as a field to this class.

justindbaur (changes requested):

src/Api/Controllers/OrganizationUsersController.cs
JaredSnider-Bitwarden: Resolving as slack conversation has indicated that a feature flag is not necessary in this context. Added complexity for no real win.

src/Core/Services/IOrganizationService.cs
coroiu: @justindbaur 
I was just following what appeared to be a convention in this file:
https://github.com/bitwarden/server/blob/d020c49c0e3957e4bf08a38be32f306e4225fdda/src/Core/Services/IOrganizationService.cs#L43-L44

I assumed that this was because services on the same layer were not allowed to depend on each other directly. Now that you mention it though, there does seem to be a circular dependency. Guess that means I will have to refactor the entire `UpdateUserResetPasswordEnrollmentAsync` into a `Command` 👍

coroiu: I gave it a fair shot spending half a day converting both `UpdateUserResetPasswordEnrollmentAsync` and `AcceptUserAsync` into commands but ran into even more circular dependencies because `UserService.SetPasswordAsync` depends on `AcceptUserAsync` and `AcceptUserAsync` depends on `UserService` meaning that if I convert `SetPasswordAsync` into a command I will end up with a dependency between two commands instead. Which in turn would require that we un-nest that spaghetti by moving the inter-command dependency into controllers and who knows what issues that might cause.

I've removed `IUserService` from `UpdateUserResetPasswordEnrollmentAsync` and instead call `AcceptUserAsync` directly from the controller. I still need to pass in `IUserService` but that method has always needed that service and so I'm at least not making it worse.

I think this whole circular dependency mess needs to be handled in a separate tech debt story

src/Core/Services/Implementations/OrganizationService.cs
coroiu: Fixed by adding to interface

src/Core/Services/IOrganizationService.cs
justindbaur: If it has already happened in this file and it's to much work for now then your solution is fine. But yes, lets make a tech debt story for it. As far as I'm aware the only other service we pass along in parameters is `ICurrentContext` and I also dislike that one 😆.

src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -305,24 +304,22 @@ public async Task PutGroups(string orgId, string id, [FromBody] OrganizationUser
     }
 
     [HttpPut("{userId}/reset-password-enrollment")]
-    public async Task PutResetPasswordEnrollment(string orgId, string userId, [FromBody] OrganizationUserResetPasswordEnrollmentRequestModel model)
```
justindbaur, line 308: ❤️

justindbaur (commented):
Looks good to me, but will leave to @JaredSnider-Bitwarden for final approval.

justindbaur (commented):
Apparently my previous comment didn't clear my request for changes. So 🤞 on this one.

justindbaur (approved):

PR #3037, opened 2023-06-22
[AC-1423] Add AddonProduct and BitwardenProduct properties to BillingSubscriptionItem

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
Add the necessary properties to the billing subscription items so that the client changes in https://github.com/bitwarden/clients/pull/5614 can render the proper prefix and have proper indentation.

## Code changes

* **src/Core/Models/Business/SubscriptionInfo.cs:** Add the new properties to the subscription item info model and response. 

* **src/Core/Utilities/StaticStore.cs:** Add helper methods to determine if a stripe subscription item is an addon product and find the a `Plan` given a Stripe plan id.  

## Before you submit


eliykat (commented):
This looks good, but it doesn't display properly when I tested it against stripe:

Screenshot 2023-06-23 at 2 19 38 pm

When I debugged, `BillingSubscriptionItem.Name` was not being set properly for service accounts. It looks like it's not being sent by Stripe.

@cyprain-okeke can you please investigate the stripe configuration to make sure it's sending this info?

Then it _should_ work, so this PR is probably fine, but I'll hold off approving until we know it's working (unless there's a benefit to merging it into the feature branch asap).

eliykat (approved):
I updated the Stripe config and this now works as expected.

PR #3048, opened 2023-06-27
[PM-1198] Modify `AuthRequest` Purge Job

justindbaur:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

`AuthRequestType.AdminApproval` has different rules for what is considered expired compared to the the other, existing, types. This updates the stored procedure to handle the new type as well as allow it to be configurable from settings with defaults for evolutionary database design. 

## Code changes

* **src/Admin/Auth/Jobs/DeleteAuthRequestsJob.cs:** Change the call to `DeleteExpiredAsync` to include customizable settings.
* **src/Core/Settings/GlobalSettings.cs:** Add new settings for customizing expiration time periods. Will conflict with them also being added #3046 but they are in one commit that I could pick out of here or just deal with a merge request depending on what merges first.
* **src/Infrastructure.Dapper/Auth/Repositories/AuthRequestRepository.cs:** Pass in the seconds to an `int` of the `TotalSeconds` in the `TimeSpan`s. This will lose the precision of any microseconds/nanoseconds that someone could have customized but I think that is more than enough precision that someone might want for these settings.
* **src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs:** Update EF logic for AdminApproval type.
* **util/Migrator/DbScripts/2023-06-27_00_AuthRequestExpirationUpdates.sql:** Migration for updating the logic of `AuthRequest_DeleteIfExpired` to handle AdminApproval rules.

## Before you submit

### DIFF
--- a/src/Core/Auth/Repositories/IAuthRequestRepository.cs
+++ b/src/Core/Auth/Repositories/IAuthRequestRepository.cs
@@ -5,7 +5,7 @@ namespace Bit.Core.Repositories;
 
 public interface IAuthRequestRepository : IRepository<AuthRequest, Guid>
 {
-    Task<int> DeleteExpiredAsync(TimeSpan userExpiration, TimeSpan adminExpiration, TimeSpan adminApprovalExpiration);
+    Task<int> DeleteExpiredAsync(TimeSpan userRequestExpiration, TimeSpan adminRequestExpiration, TimeSpan afterAdminApprovalExpiration);
     Task<ICollection<AuthRequest>> GetManyByUserIdAsync(Guid userId);
     Task<ICollection<OrganizationAdminAuthRequest>> GetManyPendingByOrganizationIdAsync(Guid organizationId);
     Task<ICollection<OrganizationAdminAuthRequest>> GetManyAdminApprovalRequestsByManyIdsAsync(Guid organizationId, IEnumerable<Guid> ids);
--- a/src/Core/Auth/Repositories/IAuthRequestRepository.cs
+++ b/src/Core/Auth/Repositories/IAuthRequestRepository.cs
@@ -5,7 +5,7 @@ namespace Bit.Core.Repositories;
 
 public interface IAuthRequestRepository : IRepository<AuthRequest, Guid>
 {
-    Task<int> DeleteExpiredAsync();
+    Task<int> DeleteExpiredAsync(TimeSpan userExpiration, TimeSpan adminExpiration, TimeSpan adminApprovalExpiration);
     Task<ICollection<AuthRequest>> GetManyByUserIdAsync(Guid userId);
     Task<ICollection<OrganizationAdminAuthRequest>> GetManyPendingByOrganizationIdAsync(Guid organizationId);
     Task<ICollection<OrganizationAdminAuthRequest>> GetManyAdminApprovalRequestsByManyIdsAsync(Guid organizationId, IEnumerable<Guid> ids);
--- a/src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
+++ b/src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
@@ -15,16 +15,17 @@ public class AuthRequestRepository : Repository<Core.Auth.Entities.AuthRequest, 
     public AuthRequestRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper)
         : base(serviceScopeFactory, mapper, (DatabaseContext context) => context.AuthRequests)
     { }
-    public async Task<int> DeleteExpiredAsync(TimeSpan userExpiration, TimeSpan adminExpiration, TimeSpan adminApprovalExpiration)
+    public async Task<int> DeleteExpiredAsync(
+        TimeSpan userRequestExpiration, TimeSpan adminRequestExpiration, TimeSpan afterAdminApprovalExpiration)
     {
 
         using (var scope = ServiceScopeFactory.CreateScope())
         {
             var dbContext = GetDatabaseContext(scope);
             var expiredRequests = await dbContext.AuthRequests
-                .Where(a => (a.Type != AuthRequestType.AdminApproval && a.CreationDate.AddSeconds(userExpiration.TotalSeconds) < DateTime.UtcNow)
-                    || (a.Type == AuthRequestType.AdminApproval && a.Approved != true && a.CreationDate.AddSeconds(adminExpiration.TotalSeconds) < DateTime.UtcNow)
-                    || (a.Type == AuthRequestType.AdminApproval && a.Approved == true && a.ResponseDate.Value.AddSeconds(adminApprovalExpiration.TotalSeconds) < DateTime.UtcNow))
+                .Where(a => (a.Type != AuthRequestType.AdminApproval && a.CreationDate.AddSeconds(userRequestExpiration.TotalSeconds) < DateTime.UtcNow)
+                    || (a.Type == AuthRequestType.AdminApproval && a.Approved != true && a.CreationDate.AddSeconds(adminRequestExpiration.TotalSeconds) < DateTime.UtcNow)
+                    || (a.Type == AuthRequestType.AdminApproval && a.Approved == true && a.ResponseDate.Value.AddSeconds(afterAdminApprovalExpiration.TotalSeconds) < DateTime.UtcNow))
                 .ToListAsync();
             dbContext.AuthRequests.RemoveRange(expiredRequests);
             return await dbContext.SaveChangesAsync();
--- a/src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
+++ b/src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
@@ -15,15 +15,19 @@ public class AuthRequestRepository : Repository<Core.Auth.Entities.AuthRequest, 
     public AuthRequestRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper)
         : base(serviceScopeFactory, mapper, (DatabaseContext context) => context.AuthRequests)
     { }
-    public async Task<int> DeleteExpiredAsync()
+    public async Task<int> DeleteExpiredAsync(TimeSpan userExpiration, TimeSpan adminExpiration, TimeSpan adminApprovalExpiration)
     {
+
         using (var scope = ServiceScopeFactory.CreateScope())
         {
             var dbContext = GetDatabaseContext(scope);
-            var expiredRequests = await dbContext.AuthRequests.Where(a => a.CreationDate < DateTime.Now.AddMinutes(-15)).ToListAsync();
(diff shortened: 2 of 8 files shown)


src/Core/Auth/Repositories/IAuthRequestRepository.cs
```diff
@@ -5,7 +5,7 @@ namespace Bit.Core.Repositories;
 
 public interface IAuthRequestRepository : IRepository<AuthRequest, Guid>
 {
-    Task<int> DeleteExpiredAsync();
+    Task<int> DeleteExpiredAsync(TimeSpan userExpiration, TimeSpan adminExpiration, TimeSpan adminApprovalExpiration);
```
JaredSnider-Bitwarden, line 8: Minor (non-blocking) ❓ :  why the shortened names for the params here vs the full names? 

I found the full names to be very clear, but you lose a decent bit of clarity w/ the shorter names.
Example: `userExpiration` !== `UserRequestExpiration` at a glance

src/Sql/Auth/dbo/Stored Procedures/AuthRequest_DeleteIfExpired.sql
```diff
@@ -1,6 +1,19 @@
...
 BEGIN
     SET NOCOUNT OFF
-    DELETE FROM [dbo].[AuthRequest] WHERE [CreationDate] < DATEADD(minute, -15, GETUTCDATE());
+    DELETE FROM [dbo].[AuthRequest]
+        -- User requests expire after 15 minutes (by default) of their creation
+    WHERE ([Type] != 2 AND DATEADD(second, @UserExpirationSeconds, [CreationDate]) < GETUTCDATE())
+        -- Admin requests expire after 7 days (by default) of their creation if they have not been approved
+        OR ([Type] = 2 AND ([Approved] IS NULL OR [Approved] = 0) AND DATEADD(second, @AdminExpirationSeconds,[CreationDate]) < GETUTCDATE())
+        -- Admin requests expire after 12 hours (by default) of their approval
+        OR ([Type] = 2 AND [Approved] = 1 AND DATEADD(second, @AdminApprovalExpirationSeconds, [ResponseDate]) < GETUTCDATE());
```
JaredSnider-Bitwarden, line 18: ❓ : when is `ResponseDate` set? 

If it is null in any case, then the DATEADD will error, right? Do we need a null check on `ResponseDate` (e.g., AND [ResponseDate] IS NOT NULL)?

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
JaredSnider-Bitwarden, line 1: Awesome work on creating the test suite to cover all these scenarios! 👍

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Bit.Infrastructure.IntegrationTest.Auth.Repositories;
+
+public class AuthRequestRepositoryTests
+{
+    private readonly static TimeSpan _userExpiration = TimeSpan.FromMinutes(15);
+    private readonly static TimeSpan _adminExpiration = TimeSpan.FromDays(6);
+    private readonly static TimeSpan _adminApprovalExpiration = TimeSpan.FromHours(12);
+
+    [DatabaseTheory, DatabaseData]
+    public async Task DeleteExpiredAsync_Works(
```
JaredSnider-Bitwarden, line 16: Minor:  should this test suite be broken up into multiple, smaller tests per best practices?  Also, in the current design, if one assertion fails, then does the entire test method stop executing (which could hide other failing tests)?

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
```diff
@@ -15,15 +15,19 @@ public class AuthRequestRepository : Repository<Core.Auth.Entities.AuthRequest,
...
+    public async Task<int> DeleteExpiredAsync(TimeSpan userExpiration, TimeSpan adminExpiration, TimeSpan adminApprovalExpiration)
     {
+
         using (var scope = ServiceScopeFactory.CreateScope())
         {
             var dbContext = GetDatabaseContext(scope);
-            var expiredRequests = await dbContext.AuthRequests.Where(a => a.CreationDate < DateTime.Now.AddMinutes(-15)).ToListAsync();
+            var expiredRequests = await dbContext.AuthRequests
+                .Where(a => (a.Type != AuthRequestType.AdminApproval && a.CreationDate.AddSeconds(userExpiration.TotalSeconds) < DateTime.UtcNow)
+                    || (a.Type == AuthRequestType.AdminApproval && a.Approved != false && a.CreationDate.AddSeconds(adminExpiration.TotalSeconds) < DateTime.UtcNow)
```
JaredSnider-Bitwarden, line 26: ❓ : SQL checks for `([Approved] IS NULL OR [Approved] = 0)` but this query just checks for false. Should it also remove nulls to match the SQL behavior?

justindbaur: EF is smart enough to recognize that it is nullable and generate the proper SQL for if that language needs a special check for null or not. For example this is the SQL that is generated for Sqlite:
```sql
SELECT "a"."Id", "a"."AccessCode", "a"."Approved", "a"."AuthenticationDate", "a"."CreationDate", "a"."Key", "a"."MasterPasswordHash", "a"."OrganizationId", "a"."PublicKey", "a"."RequestDeviceIdentifier", "a"."RequestDeviceType", "a"."RequestIpAddress", "a"."ResponseDate", "a"."ResponseDeviceId", "a"."Type", "a"."UserId"
FROM "AuthRequest" AS "a"
WHERE ((("a"."Type" <> 2) AND (rtrim(rtrim(strftime('%Y-%m-%d %H:%M:%f', "a"."CreationDate", CAST(@__userExpiration_TotalSeconds_0 AS TEXT) || ' seconds'), '0'), '.') < rtrim(rtrim(strftime('%Y-%m-%d %H:%M:%f', 'now'), '0'), '.'))) OR ((("a"."Type" = 2) AND (("a"."Approved" <> 0) OR ("a"."Approved" IS NULL))) AND (rtrim(rtrim(strftime('%Y-%m-%d %H:%M:%f', "a"."CreationDate", CAST(@__adminExpiration_TotalSeconds_1 AS TEXT) || ' seconds'), '0'), '.') < rtrim(rtrim(strftime('%Y-%m-%d %H:%M:%f', 'now'), '0'), '.')))) OR ((("a"."Type" = 2) AND ("a"."Approved" = 1)) AND (rtrim(rtrim(strftime('%Y-%m-%d %H:%M:%f', "a"."ResponseDate", CAST(@__adminApprovalExpiration_TotalSeconds_2 AS TEXT) || ' seconds'), '0'), '.') < rtrim(rtrim(strftime('%Y-%m-%d %H:%M:%f', 'now'), '0'), '.')))
```

It's kinda insane to read but buried in there is `(("a"."Approved" <> 0) OR ("a"."Approved" IS NULL))` it decided to alternate it but should work all the same. But I don't fully trust every provider to do it properly hence the tests. But this is a good shout, I should also have a test the has an AdminApproval that was denied and make sure that it gets deleted after the 7 days.

src/Sql/Auth/dbo/Stored Procedures/AuthRequest_DeleteIfExpired.sql
justindbaur: `ResponseDate` is set when `Approved` is given a state. https://github.com/bitwarden/server/blob/d020c49c0e3957e4bf08a38be32f306e4225fdda/src/Core/Auth/Services/Implementations/AuthRequestService.cs#L134-L135

So checking that `Approved` is true business logic wise is also me finding out that ResponseDate isn't null. I should add a assertions in #3046 for making sure `ResponseDate` is set on all valid approvals/disapprovals though.

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
JaredSnider-Bitwarden: Ah, thank you for the explanation. Glad it helped id another necessary test case!

src/Core/Auth/Repositories/IAuthRequestRepository.cs
justindbaur: Not sure why I dropped the `Request`. Added it back!

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
justindbaur: This helped me find a bug in the EF implementation, as you can see above what Sqlite was generating for that check it was NOT logically equivalent to the SQL implementation. I fixed the bug and now it will generate `(("a"."Approved" <> 1) OR ("a"."Approved" IS NULL))`, which is different but now logically equivalent. 

Do you think I should change SQL to be ([Approved] IS NULL OR [Approved] != 1) though?

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
justindbaur: I don't think so, the purpose of this test is to both make sure it's deleting items I expect it to test as well as doesn't delete things I expect it to keep around. To do that well I would have to assert the "don't deletes" for each smaller test still but have the "should deletes" all be separate. 

Also since these are integration tests being ran against real databases that are likely to have multiple different `AuthRequest`s that need to be deleted for the various reasons I think this it's worth the larger test to more accurately test the system.

I do have the number of deleted check go last so that if you didn't delete as many as you should have you get a failure on the line for the item you either should have deleted and didn't or an item you shouldn't have deleted and did. The test failure will give you the line number and from that you can get a quick start on figuring out what you did wrong. Plus, if you have only broken one type of database but not the others it will run through all the other types still so you will know the type that you specifically broke.

src/Sql/Auth/dbo/Stored Procedures/AuthRequest_DeleteIfExpired.sql
justindbaur: Added tests for this in https://github.com/bitwarden/server/pull/3046/commits/95240849fa84b8f18018defe6645f926afda007f.

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
JaredSnider-Bitwarden: That's reasonable!

JaredSnider-Bitwarden (approved):
Love the extra clarity! Thank you for the great work on this!

PR #3055, opened 2023-06-29
[AC-1418] Add SecretsManagerPlan property to OrganizationResponseModel

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

Add missing `SecretsManagerPlan` to the `OrganizationResponseModel` so that the SM subscription components have access to the correct SM plan details.

Fixes the issue mentioned in Clients PR: https://github.com/bitwarden/clients/pull/5661 (see [this comment](https://github.com/bitwarden/clients/pull/5661#discussion_r1244624073) for more details)

## Before you submit

### DIFF
--- a/src/Api/Models/Response/Organizations/OrganizationResponseModel.cs
+++ b/src/Api/Models/Response/Organizations/OrganizationResponseModel.cs
@@ -27,6 +27,7 @@ public class OrganizationResponseModel : ResponseModel
         BusinessTaxNumber = organization.BusinessTaxNumber;
         BillingEmail = organization.BillingEmail;
         Plan = new PlanResponseModel(StaticStore.PasswordManagerPlans.FirstOrDefault(plan => plan.Type == organization.PlanType));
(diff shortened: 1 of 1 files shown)


src/Api/Models/Response/Organizations/OrganizationResponseModel.cs
```diff
@@ -65,6 +66,7 @@ public OrganizationResponseModel(Organization organization, string obj = "organi
     public string BusinessTaxNumber { get; set; }
     public string BillingEmail { get; set; }
     public PlanResponseModel Plan { get; set; }
```
vvolkgang, line 68: 🤔 Should we create a tech debt ticket to update this property name to `PasswordManagerPlan`?

eliykat: Kind of. These plan objects need refactoring generally - I think we'll probably end up combining them, imo we should have an "Enterprise Plan" which contains different possible products, not a "PM plan" and an "SM plan". This is tracked in AC-1451 to be done immediately after GA. I'll update the description in Jira.

eliykat (approved):

PR #3066, opened 2023-07-03
[PM-1381] Add Trusted Device Keys to Auth Response

justindbaur:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add the keys from a device when a user is logging in on a device that has previously been trusted.

## Code changes

* **src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs:** Create optional properties for the `EncryptedPublicKey` & `EncryptedUserKey` for when the device being signed into has been trusted. They will be `null` when the device signing in has not been trusted yet.
* **src/Core/Auth/Utilities/DeviceExtensions.cs:** Added an extension method for determining if a device contains the necessary values to consider itself trusted (as far as the server knows). I put it in an extension method so that Auth team can own the logic of what that means without needing to own the full device.
* **src/Identity/IdentityServer/BaseRequestValidator.cs:** Pass `Device` into `CreateUserDecryptionOptionsAsync` so that I can test whether or not it's trusted and if it is pass along the needed details so that the requester can continue trying to decrypt their vault.
* **test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs:** Added a test validating that when a device is trusted it will respond with the needed keys.

## Before you submit

### DIFF
--- a/src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs
+++ b/src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs
@@ -32,13 +32,13 @@ public class UserDecryptionOptions : ResponseModel
 public class TrustedDeviceUserDecryptionOption
 {
     public bool HasAdminApproval { get; }
-    public string? EncryptedPublicKey { get; }
+    public string? EncryptedPrivateKey { get; }
     public string? EncryptedUserKey { get; }
 
-    public TrustedDeviceUserDecryptionOption(bool hasAdminApproval, string? encryptedPublicKey, string? encryptedUserKey)
+    public TrustedDeviceUserDecryptionOption(bool hasAdminApproval, string? encryptedPrivateKey, string? encryptedUserKey)
     {
         HasAdminApproval = hasAdminApproval;
-        EncryptedPublicKey = encryptedPublicKey;
+        EncryptedPrivateKey = encryptedPrivateKey;
         EncryptedUserKey = encryptedUserKey;
     }
 }
--- a/src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs
+++ b/src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs
@@ -32,10 +32,14 @@ public class UserDecryptionOptions : ResponseModel
 public class TrustedDeviceUserDecryptionOption
 {
     public bool HasAdminApproval { get; }
+    public string? EncryptedPublicKey { get; }
+    public string? EncryptedUserKey { get; }
 
-    public TrustedDeviceUserDecryptionOption(bool hasAdminApproval)
+    public TrustedDeviceUserDecryptionOption(bool hasAdminApproval, string? encryptedPublicKey, string? encryptedUserKey)
     {
         HasAdminApproval = hasAdminApproval;
+        EncryptedPublicKey = encryptedPublicKey;
+        EncryptedUserKey = encryptedUserKey;
     }
 }
 
--- a/src/Core/Auth/Utilities/DeviceExtensions.cs
+++ b/src/Core/Auth/Utilities/DeviceExtensions.cs
@@ -9,6 +9,11 @@ public static class DeviceExtensions
     /// <summary>
     /// Gets a boolean representing if the device has enough information on it to determine whether or not it is trusted.
     /// </summary>
+    /// <remarks>
+    /// It is possible for a device to be un-trusted client side and not notify the server side. This should not be
+    /// the source of truth for whether a device is fully trusted and should just be considered that, to the server,
+    /// a device has the necessary information to be "trusted".
+    /// </remarks>
     public static bool IsTrusted(this Device device)
     {
         return !string.IsNullOrEmpty(device.EncryptedUserKey) &&
--- a/src/Core/Auth/Utilities/DeviceExtensions.cs
+++ b/src/Core/Auth/Utilities/DeviceExtensions.cs
@@ -1,0 +1,18 @@
+using Bit.Core.Entities;
+
+#nullable enable
+
+namespace Bit.Core.Auth.Utilities;
+
+public static class DeviceExtensions
+{
+    /// <summary>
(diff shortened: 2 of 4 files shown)


src/Core/Auth/Utilities/DeviceExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Bit.Core.Entities;
+
+#nullable enable
+
+namespace Bit.Core.Auth.Utilities;
+
+public static class DeviceExtensions
+{
+    /// <summary>
+    /// Gets a boolean representing if the device has enough information on it to determine whether or not it is trusted.
```
JaredSnider-Bitwarden, line 10: First off, love the method and the intent behind it!

However, for clarity, should this check be something like `hasTrustedDeviceKeys` vs `isTrusted`?  It's a bit misleading to call it is trusted when the trusted state of a device can only truly be determined client side with a valid device key to decrypt the `EncryptedPrivateKey` here and then have the privateKey decrypt the `encryptedUserKey`.  

If you would prefer to not change the name, it might be good to include that context in the summary at least explaining that the existence of the keys is the only thing that the server has to look at to determine if a device is likely to be trusted. 

In theory, these keys should never be populated unless the device was setup as a trusted device at least once but client side key persistence (e.g., user clears local storage) could make them irrelevant and make their device no longer trusted. 

Happy to discuss!

test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs
```diff
@@ -245,6 +245,88 @@ public async Task SsoLogin_TrustedDeviceEncryptionAndNoMasterPassword_ReturnsOne
...
+        var userDecryptionOptions = AssertHelper.AssertJsonProperty(root, "UserDecryptionOptions", JsonValueKind.Object);
+
+        // Expected to look like:
+        // "UserDecryptionOptions": {
+        //   "Object": "userDecryptionOptions"
+        //   "HasMasterPassword": false,
+        //   "TrustedDeviceOption": {
+        //     "HasAdminApproval": true,
+        //     "EncryptedPublicKey": "2.QmFzZTY0UGFydA==|QmFzZTY0UGFydA==|QmFzZTY0UGFydA==",
+        //     "EncryptedUserKey": "2.QmFzZTY0UGFydA==|QmFzZTY0UGFydA==|QmFzZTY0UGFydA=="
```
JaredSnider-Bitwarden, line 317: Do your other tests already cover the case in which the device is untrusted and these should not come down or be null?

test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs
JaredSnider-Bitwarden, line 1: Once again, awesome work on always adding tests!

JaredSnider-Bitwarden (commented):
Awesome work! Just a few minor questions!

src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs
```diff
@@ -32,10 +32,14 @@ public UserDecryptionOptions() : base("userDecryptionOptions")
...
     public bool HasAdminApproval { get; }
+    public string? EncryptedPublicKey { get; }
+    public string? EncryptedUserKey { get; }
 
-    public TrustedDeviceUserDecryptionOption(bool hasAdminApproval)
+    public TrustedDeviceUserDecryptionOption(bool hasAdminApproval, string? encryptedPublicKey, string? encryptedUserKey)
     {
         HasAdminApproval = hasAdminApproval;
+        EncryptedPublicKey = encryptedPublicKey;
+        EncryptedUserKey = encryptedUserKey;
```
coroiu, line 42: **question/issue:** Doesn't this also need to contain a private key? Trusting a device requires three keys: 
- DevicePublicKey(UserKey)
- UserKey(DevicePublicKey)
- DeviceKey(DevicePrivateKey)

JaredSnider-Bitwarden: *facepalm* 

Thank you for catching that!

justindbaur: Yeah very good catch, I was working off a conversation with Jared and I knew I only needed 2/3 but I someone got it wrong even though our conversation had the correct ones.

test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs
justindbaur: Added a test that asserts it's not there. It's generally a little weird to assert that something does not exist but I think it's worth it for this sensitive stuff.

src/Core/Auth/Utilities/DeviceExtensions.cs
justindbaur: Yes, it's not known to the server it's truest form of trust but I generally like the name, I can add a comment that explains that it's not fully known if it's trusted but to the server it's as known as it can be. 

Each time we use that method (I already use it on some other code) it's for guarding logic that we would only do for a "trusted" device so it feels like it reads better and would evolve better later on for if the logic of what server things a trusted device is. 

If you feel strongly I am more than willing to switch it up though.

src/Core/Auth/Models/Api/Response/UserDecryptionOptions.cs
JaredSnider-Bitwarden: It happens sometimes. I read it and thought, "yep, this is all correct" even though we both had the conversation about the correct keys.

src/Core/Auth/Utilities/DeviceExtensions.cs
justindbaur: Done!

JaredSnider-Bitwarden (approved):
Awesome!

PR #3080, opened 2023-07-09
Vscode devcontainers

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
This adds [VSCode devcontainers](https://code.visualstudio.com/docs/devcontainers/containers), which should make the dev environment setup process easier for those that choose to utilize this workflow.

## Code changes

- **`.devcontainer/bitwarden_common/docker-compose.yml`:** a baseline set of services that are used in both the internal and community dev configurations

- **`.devcontainer/(community_dev|internal_dev)/devcontainer.json`:** the declaration files for the VSCode dev containers
- **`.devcontainer/internal_dev/docker-compose.override.yml`:** adds Azurite container for internal dev config
- **`.devcontainer/(community_dev|internal_dev)/postCreateCommand.sh`:** these optional one-time setup scripts configure `secrets.json` with the DB password, installation ID and key, and certificate thumbprints to enable running `setup_secrets.ps1` automatically after the devcontainer initializes

## Before you submit


.devcontainer/internal_dev/devcontainer.json
```diff
@@ -0,0 +1,16 @@
...
+  "dockerComposeFile": [
+    "../../.devcontainer/bitwarden_common/docker-compose.yml",
+    "../../.devcontainer/internal_dev/docker-compose.override.yml"
+  ],  "service": "bitwarden_server",
+  "workspaceFolder": "/workspace",
+  "customizations": {
+    "vscode": {
+      "settings": {},
+      "features": {},
+      "extensions": ["ms-dotnettools.csharp"]
```
justindbaur, line 12: Same.

```suggestion
      "extensions": ["ms-dotnettools.csdevkit"]
```

.devcontainer/community_dev/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "Bitwarden Community Dev",
+  "dockerComposeFile": "../../.devcontainer/bitwarden_common/docker-compose.yml",
+  "service": "bitwarden_server",
+  "workspaceFolder": "/workspace",
+  "customizations": {
+    "vscode": {
+      "settings": {},
+      "features": {},
+      "extensions": ["ms-dotnettools.csharp"]
```
justindbaur, line 10: Since I sat on this to long the C# Dev Kit has come out and we should probably use that instead. 
```suggestion
      "extensions": ["ms-dotnettools.csdevkit"]
```

justindbaur (commented):
I'm so sorry for sitting on this to long, I still want to get this in and this will make things so much easier.

justindbaur (dismissed):
Thank you so much, @MGibson1 you're up!

MGibson1 (commented):
I _think_ I only have two comments, but this feature is totally new to me
1. I might be old fashioned, but I _like_ manually controlling when migrations are run in dev
2. ⚠️ the correct sdk does not appear to be installed by the containers by default. Is it possible to add this to the post-install script?

tangowithfoxtrot:
@MGibson1 

1. The reason I initially chose to utilize the `Admin` project to run the migrations automatically was because the `migration.ps1` script relies on a docker container to run them, which would require setting up Docker-in-Docker and/or mounting the host's `docker.sock`, which is very hacky, imo. However, @justindbaur informed me about `./util/MsSqlMigratorUtility`, so I've updated the containers to use that instead (in the first-time setup script) and reverted the automatic migrations change in `Admin`. We can now run manual migrations from within the devcontainer by running `dotnet run --project ./util/MsSqlMigratorUtility "$SQL_CONN_STRING"` instead of the `migrate.ps1` script.

2. I've adjusted the devcontainer image to use `mcr.microsoft.com/devcontainers/dotnet:dev-6.0`. Previously, it used `mcr.microsoft.com/devcontainers/dotnet:0-6.0`, which, as you mentioned, has an older SDK version. The dev image uses `6.0.416`. Unfortunately, I can't get it to perfectly match `6.0.415`, as declared in `global.json` because [Microsoft hasn't published that _exact_ version for the devcontainer image](https://mcr.microsoft.com/v2/devcontainers/dotnet/tags/list).

.devcontainer/community_dev/postCreateCommand.sh
```diff
@@ -0,0 +1,63 @@
...
+        sleep 1
+        get_installation_id_and_key
+        configure_other_vars
+        pushd ./dev >/dev/null || exit
+        pwsh ./setup_secrets.ps1 || true
+        popd >/dev/null || exit
+                
+        echo "Running migrations..."
+        sleep 5 # wait for DB container to start
+        dotnet run --project ./util/MsSqlMigratorUtility "$SQL_CONNECTION_STRING"
```
withinfocus, line 57: 🎉 -- this is how our cloud is now performing migrations.

withinfocus:
I like where this is at but will let the other two guys give an approval. Warning: auto-merge is on.

MGibson1 (approved):

PR #3081, opened 2023-07-10
[AC-1495] Extract UpgradePlanAsync into a command

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Move `UpgradePlanAsync` out of `OrganizationService` so that we don't import `ServiceAccountRepository` into `OrganizationService` and therefore all the projects that use it. (It drags in dependencies and the need to manage OSS and non-OSS code, which is best avoided or at least minimised.) This is a simple lift & shift / cut & paste to solve the dependency issues with no additional refactors in order to minimise regression risk.

It also reverts the changes to `Startup.cs` for some projects, these changes are no longer required because they no longer inject that dependency.

## Code changes

## Before you submit

### DIFF
--- a/bitwarden_license/src/Scim/Startup.cs
+++ b/bitwarden_license/src/Scim/Startup.cs
@@ -42,8 +42,6 @@ public class Startup
         // Repositories
         services.AddDatabaseRepositories(globalSettings);
 
-        services.AddOosServices();
-
         // Context
         services.AddScoped<ICurrentContext, CurrentContext>();
         services.AddScoped<IScimContext, ScimContext>();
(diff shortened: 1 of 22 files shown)


r-tome (changes requested):
The tests are failing and its because we're still having the dependency issues with `IServiceAccountRepository` but now its further down.

@cyprain-okeke : the issue is happening because we are injecting AddOrganizationSubscriptionCommandsQueries in AddOrganizationServices which is injected in AddBaseServices. And this AddBaseServices is needed by identity and admin project.

eliykat:
Fixed: I gave the Organization Subscription commands their own service collection extension, then only registered them in the Api project, which is the only place that needs them.

Api also handles the ServiceAccountRepository dependency by registering the noop version for OSS builds (`services.AddOosServices`) or the real implementation otherwise (services.AddSecretsManagerEfRepositories`).

cyprain-okeke (commented):
This looks great

PR #3084, opened 2023-07-11
[AC-1503] Fix Stripe integration on organization upgrade

eliykat:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The upgrade plan command (previously `OrganizationService.UpgradePlanAsync`) was not passing SM parameters through to `StripePaymentService` on upgrade. The `StripePaymentService` method had default parameters which disguised this.

Fixed by:
* removing default values for parameters everywhere, it turns out that all calling locations (except for tests) pass in arguments, so they're not needed
* passing the `OrganizationUpgrade` object into `StripePaymentService`. This is unpacked in the `OrganizationUpgradeSubscriptionOptions` constructor to limit the scope of this change, but it gets us moving in the right direction
* update tests to use this new method signature, also added some assertions on a test (not strictly required but good to have the coverage)

Some minor unrelated fixes:
* I found the cause of flaky tests, we were still overwriting `StaticStore` values which is leaked between tests and was causing unexpected behaviour elsewhere
* removed some extra brackets in an error message

## Code changes

## Before you submit

### DIFF
--- a/bitwarden_license/src/Scim/Startup.cs
+++ b/bitwarden_license/src/Scim/Startup.cs
@@ -42,8 +42,6 @@ public class Startup
         // Repositories
         services.AddDatabaseRepositories(globalSettings);
 
-        services.AddOosServices();
-
         // Context
         services.AddScoped<ICurrentContext, CurrentContext>();
         services.AddScoped<IScimContext, ScimContext>();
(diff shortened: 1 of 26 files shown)


cyprain-okeke (approved):
Great fix. Thanks

r-tome (approved):

r-tome (approved):

PR #3092, opened 2023-07-12
[AC-1508] Stripe changes for the EU datacenter

cturnbull-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
The launch of the EU data center brought up some issues around the Stripe webhook. With the current implementation, The `"webhook"` endpoints in both data centers will each receive all events sent out by Stripe. To combat this in the short term, and to avoid needless errors around looking up entities that exist in another datacenter, we've added a `region` value to the Stripe customer object. Using that value, we're able to determine if a given event is intended for the current instance's webhook or that of another data center.

## Code changes

* **StripePaymentService.cs:** Added logic to supply the configured region value to Stripe when creating or updating a Customer. This will occur when purchasing an organization, purchasing a premium subscription, or when updating a payment method
* **StripeController.cs:** Added logic to check the current instance's region (`"US"` or `"EU"`), and match it with the customer's region. If they match, execution continues as normal. Otherwise, the webhook returns `200 OK` to Stripe.
* **HandledStripeWebhook.cs:** Updated the webhook event type strings to be compile-time constants instead of read-only properties with a single expression. This is to allow them to be used in the switch statement in StripeController.cs.

### DIFF
--- a/src/Billing/Controllers/StripeController.cs
+++ b/src/Billing/Controllers/StripeController.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Options;
 using Stripe;
+using Event = Stripe.Event;
 using TaxRate = Bit.Core.Entities.TaxRate;
 
 namespace Bit.Billing.Controllers;
@@ -41,6 +42,7 @@ public class StripeController : Controller
     private readonly ITaxRateRepository _taxRateRepository;
     private readonly IUserRepository _userRepository;
     private readonly ICurrentContext _currentContext;
+    private readonly GlobalSettings _globalSettings;
 
     public StripeController(
         GlobalSettings globalSettings,
@@ -83,6 +85,7 @@ public class StripeController : Controller
             PrivateKey = globalSettings.Braintree.PrivateKey
         };
         _currentContext = currentContext;
+        _globalSettings = globalSettings;
     }
 
     [HttpPost("webhook")]
@@ -112,6 +115,12 @@ public class StripeController : Controller
         {
             _logger.LogWarning("Getting test events in production.");
             return new BadRequestResult();
+        }
+
+        // If the customer and server cloud regions don't match, early return 200 to avoid unnecessary errors
+        if (!await ValidateCloudRegionAsync(parsedEvent))
+        {
+            return new OkResult();
         }
 
         var subDeleted = parsedEvent.Type.Equals(HandledStripeWebhook.SubscriptionDeleted);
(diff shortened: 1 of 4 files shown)


src/Billing/Controllers/StripeController.cs
```diff
@@ -471,6 +480,71 @@ public async Task<IActionResult> PostWebhook([FromQuery] string key)
...
+
+        switch (eventType)
+        {
+            case HandledStripeWebhook.SubscriptionDeleted:
+            case HandledStripeWebhook.SubscriptionUpdated:
+                {
+                    var subscription = await GetSubscriptionAsync(parsedEvent, true, new List<string>
+                {
+                    "customer"
+                });
```
r-tome, line 506: Nitpick: indentation

src/Billing/Controllers/StripeController.cs
```diff
@@ -471,6 +480,71 @@ public async Task<IActionResult> PostWebhook([FromQuery] string key)
...
+    /// <summary>
+    /// Gets the region from the customer metadata. If no region is present, defaults to "US"
+    /// </summary>
+    /// <param name="customerMetadata"></param>
+    /// <returns></returns>
+    private static string GetCustomerRegionFromMetadata(Dictionary<string, string> customerMetadata)
+    {
+        return !customerMetadata.ContainsKey("region")
+            ? "US"
+            : customerMetadata["region"];
```
r-tome, line 545: Another nitpick: I prefer positive statements
```suggestion
        return customerMetadata.ContainsKey("region")
            ? customerMetadata["region"]
            : "US";
```

r-tome (changes requested):
This looks really great! Other than the minor nitpicks I pointed out the unit tests have to be fixed to look for the region in the metadata.

src/Billing/Controllers/StripeController.cs
```diff
@@ -762,17 +837,19 @@ private async Task<Invoice> GetInvoiceAsync(Stripe.Event parsedEvent, bool fresh
             return eventInvoice;
         }
         var invoiceService = new InvoiceService();
-        var invoice = await invoiceService.GetAsync(eventInvoice.Id);
+        var invoiceGetOptions = new InvoiceGetOptions { Expand = expandOptions };
```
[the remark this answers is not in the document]
cturnbull-bitwarden: It does! It returns the given object without any additional properties expanded

src/Billing/Controllers/StripeController.cs
cturnbull-bitwarden: Fixed!

src/Billing/Controllers/StripeController.cs
cturnbull-bitwarden: Updated in the last push

r-tome (approved):
Great work!

cyprain-okeke (commented):
It looks Great.

src/Billing/Controllers/StripeController.cs
```diff
@@ -114,6 +117,12 @@ public async Task<IActionResult> PostWebhook([FromQuery] string key)
             return new BadRequestResult();
         }
 
+        // If the customer and server cloud regions don't match, early return 200 to avoid unnecessary errors
+        if (!await ValidateCloudRegionAsync(parsedEvent))
```
cyprain-okeke, line 121: for backword comptability, if the region is null from stripe metadata event, it means is US and the existing process take place, but if the region has a value and is not US, the new process takes place.
var cloudRegionResponse = await ValidateCloudRegionAsync(parsedEvent);
if(cloudRegionResponse != "US")
{
    return new OkResult();
}

src/Billing/Controllers/StripeController.cs
```diff
@@ -471,6 +480,68 @@ public async Task<IActionResult> PostWebhook([FromQuery] string key)
...
 
+    /// <summary>
+    /// Ensures that the customer associated with the parsed event's data is in the correct region for this server.
+    /// We use the customer instead of the subscription given that all subscriptions have customers, but not all
+    /// customers have subscriptions
+    /// </summary>
+    /// <param name="parsedEvent"></param>
+    /// <returns>true if the customer's region and the server's region match, otherwise false</returns>
+    /// <exception cref="Exception"></exception>
+    private async Task<bool> ValidateCloudRegionAsync(Event parsedEvent)
```
cyprain-okeke, line 491: What do you think ? If we keep this method light by 
```
var subscriptionEvent = await GetSubscriptionAsync(parsedEvent, true);
return  GetCustomerRegionFromMetadata(subscriptionEvent.Metadata);
```

This way will avoid a lot regression testing on the webhook changes

src/Billing/Controllers/StripeController.cs
cyprain-okeke, line 121: I think we should called this immediately after below code

 ```
if (_hostingEnvironment.IsProduction() && !parsedEvent.Livemode)
        {
            _logger.LogWarning("Getting test events in production.");
            return new BadRequestResult();
        }
```
This way will avoid a lot regression testing on the webhook changes

cturnbull-bitwarden: This should be handled already down in this `private` helper method: [GetCustomerRegionFromMetadata](https://github.com/bitwarden/server/blob/AC-1508-Stripe-changes-for-the-EU-datacenter/src/Billing/Controllers/StripeController.cs#L538C40-L538C40). When the user's region is the US or null, we'll still need to validate that the server's region matches it.

cturnbull-bitwarden: I'm seeing that the new validation logic is already after this bit of code! Let me know if I didn't understand your meaning

cyprain-okeke (approved):
Good Work.

PR #3104, opened 2023-07-14
663-AccessTokens validate expiration date

cd-bitwarden:
## Type of change

```
- [ x ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Validate expiration date is in the future on the server side

## Code changes

src/Api/SecretsManager/Models/Request/AccessTokenCreateRequestModel.cs - implement IValidatableObject so that if the expiration date is in the past it returns an error message that's descriptive
## Before you submit


src/Api/SecretsManager/Models/Request/AccessTokenCreateRequestModel.cs
```diff
@@ -34,4 +34,13 @@ public ApiKey ToApiKey(Guid serviceAccountId)
             EncryptedPayload = EncryptedPayload,
         };
     }
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        if (ExpireAt <= DateTime.Now)
```
Thomas-Avery, line 40:
```suggestion
        if (ExpireAt <= DateTime.UtcNow)
```

We send and handle time as UTC.

I would also add some logic here to handle the null case for when we are doing never expiring tokens.

Thomas-Avery (changes requested):
LGTM

src/Api/SecretsManager/Models/Request/AccessTokenCreateRequestModel.cs
```diff
@@ -34,4 +34,13 @@ public ApiKey ToApiKey(Guid serviceAccountId)
             EncryptedPayload = EncryptedPayload,
         };
     }
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        if (ExpireAt <= DateTime.UtcNow && ExpireAt != null)
```
Thomas-Avery, line 40:
```suggestion
        if (ExpireAt != null && ExpireAt <= DateTime.UtcNow)
```

The conditional will short circuit if the left most check is false. So we would want our null check first.

Here are some [docs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-and-operator-) that will explain it better.

cd-bitwarden: Thank you Thomas!

Thomas-Avery (approved):
LGTM

PR #3111, opened 2023-07-17
[AC-1522] Fix service account check on upgrading

eliykat:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

I noticed that the logic to check service accounts on upgrade is wrong. It was gated behind a `newSecretsManagerPlan.BaseServiceAccount != null` check, which is irrelevant, and it wasn't calculating the new plan total correctly. It basically needs to mirror the seats logic, i.e. if the new total is less than the current total, check current usage and throw if the new total is less than the current usage.

Also added unit tests.

Also added "or increase your subscription" to the end of the error messages, because that's another (and better) way of solving the problem.

## Code changes

## Before you submit


r-tome (approved):
Looks good

eliykat:
Merging to feature branch round 1 because that contains AC-1409 which has the bug.

PR #3117, opened 2023-07-18
[PM-3007] Caching user policies on PolicyService variable

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Enhance performance and reduce SQL operations by modifying the stored procedure `OrganizationUser_ReadByUserIdWithPolicyDetails` to retrieve all policies and filter them in memory for each type, thereby minimizing the number of SQL operations required. The filtered policies will be cached and utilized by the scoped service `IPolicyService`.

## Code changes

* **src/Core/Repositories/IOrganizationUserRepository.cs:** Removed the parameter `PolicyType` so that the method returns all policy types
* **src/Core/Services/Implementations/PolicyService.cs:** Added a private variable to cache the obtained user policies from the repository
* **src/Infrastructure.Dapper/Repositories/OrganizationUserRepository.cs:** Removed the parameter `PolicyType` so that the method returns all policy types
* **src/Infrastructure.EntityFramework/Repositories/OrganizationUserRepository.cs:** Removed the parameter `PolicyType` so that the method returns all policy types
* **src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByUserIdWithPolicyDetails.sql:** Removed the parameter `PolicyType` so that the method returns all policy types and also replaced the nested query with left joins
* **test/Core.Test/Services/PolicyServiceTests.cs:** Updated unit tests
* **test/Infrastructure.EFIntegration.Test/Repositories/OrganizationUserRepositoryTests.cs:** Updated unit tests
* **util/Migrator/DbScripts/2023-07-18_00_OrganizationUserReadByUserIdWithPolicyDetails.sql:** SQL migration script

## Before you submit


r-tome:
I've placed the cached policies inside `PolicyService` but maybe it's more suited inside `CurrentContext`?

eliykat:
> I've placed the cached policies inside PolicyService but maybe it's more suited inside CurrentContext?

CurrentContext is already doing too much imo. Much better to have it within the domain-specific service, which is exactly what you've done.

I'm curious whether you've compared the query execution plans for the old & new sproc? I'm sure we're going to see some boost from the caching either way, so I'd like to know that we've also improved the sproc itself. If you want any help with this let me know.

r-tome:
> CurrentContext is already doing too much in my opinion. I think it's much better to have this cached within the domain-specific service, which is exactly what you've done.

That is what I thought but I needed validation 😅

> I'm curious whether you've compared the query execution plans for the old & new sproc? I'm sure we're going to see some boost from the caching either way, so I'd like to know that we've also improved the sproc itself. If you want any help with this let me know.

I did compare them! Here are both execution plans (old on top):

image

I think its improved by having less nested loops but that my non-expert opinion. Let me know your thoughts.

withinfocus:
How'd you get these query plans by chance? You'll need the actual plans from production.

r-tome:
> How'd you get these query plans by chance? You'll need the actual plans from production.

That is true, who can I ping to get those? These are just from my local dev environment.

withinfocus:
Heh yeah I was asking the same questions to CloudOps and there wasn't an answer. For your sake I think you need to make an ad hoc request for them to jump in and pull some actual execution plans for you.

eliykat:
@withinfocus I was aware of that limitation, but I thought using local query plans would be the next best thing. But comparing them, they're not even close, so that's good to know.

I guess that means we also won't know of any measurable improvements until the new sproc has been running in prod for a while?

withinfocus:
In our situation that's pretty much correct. The usual course of action is to snapshot the table or simulate size and fragmentation so that it could be replicated elsewhere, but we are still adopting better database testing practices.

MGibson1:
> I guess that means we also won't know of any measurable improvements until the new sproc has been running in prod for a while?

It's worth checking qa cloud to see if that matches a little better, but this is probably true.

kspearrin (dismissed):
This seems like an acceptable improvement, but there is certainly much more than can be done to optimize the query within the database alone. For example, I see lots of index scanning that have a large cost to the overall execution plan.

r-tome:
@kspearrin I experimented with adding two new indexes and changing an existing one to include some necessary columns.

```
CREATE NONCLUSTERED INDEX [IX_ProviderUser_UserIdProviderIdStatus]
        ON [dbo].[ProviderUser]([UserId] ASC, [ProviderId] ASC, [Status] ASC)
        WITH (ONLINE = ON);

CREATE NONCLUSTERED INDEX [IX_ProviderOrganization_ProviderIdOrganizationId]
        ON [dbo].[ProviderOrganization]([ProviderId] ASC, [OrganizationId] ASC);
        WITH (ONLINE = ON);

CREATE NONCLUSTERED INDEX [IX_OrganizationUser_UserIdOrganizationIdStatusEmail]
            ON [dbo].[OrganizationUser]([UserId] ASC, [OrganizationId] ASC, [Status] ASC, [Email] ASC)
            INCLUDE ([AccessAll], [Type], [Permissions])
            WITH (ONLINE = ON);
-- Drop existing index that did not include the [Email] column
DROP INDEX [IX_OrganizationUser_UserIdOrganizationIdStatus] ON [dbo].[OrganizationUser];
```

Here is the old plan when I ran the query locally:
image

Here is the new plan with the indexes changes:
image

With these changes I managed to reduce the amount of index scans but of course this is only my local environment so comparing the actual numbers is pointless.

Do you think we should try to add the indexes?

kspearrin:
@r-tome I think those are good results, however, let's hold off on adding lots of indexes right now until we have the right people available to evaluate our indexing strategy. We can always add them on-demand if this query becomes more of a problem in production.

kspearrin (approved):

withinfocus (approved):

PR #3112, opened 2023-07-17
[AC-1468] Exclude Secrets Manager beta users from seat and service account limits

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective
Exclude beta users from seat and service account limits.

## Code changes
- **bitwarden_license/src/Commercial.Core/SecretsManager/Queries/ServiceAccounts/CountNewServiceAccountSlotsRequiredQuery.cs:** If organization participated in Secrets Manager Beta then no new service accounts are needed
- **bitwarden_license/test/Commercial.Core.Test/SecretsManager/Queries/ServiceAccounts/CountNewServiceAccountSlotsRequiredQueryTests.cs:** Updated tests to set value for SecretsManagerBeta and added new test

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3059, opened 2023-06-30
[AC-1462] Add secrets manager service accounts autoscaling commands

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Code changes
- **bitwarden_license/src/Commercial.Core/SecretsManager/Commands/ServiceAccounts/AutoscaleServiceAccountsCommand.cs:** Command to add Service Account slots for an Organization
- **bitwarden_license/src/Commercial.Core/SecretsManager/Queries/ServiceAccounts/CountNewServiceAccountSlotsRequiredQuery.cs:** Query to determine how many slots are required for a number of new service accounts
- **bitwarden_license/test/Commercial.Core.Test/SecretsManager/Commands/ServiceAccounts/AutoscaleServiceAccountsCommandTests.cs:** Unit tests for AutoscaleServiceAccountsCommand
- **bitwarden_license/test/Commercial.Core.Test/SecretsManager/Queries/ServiceAccounts/CountNewServiceAccountSlotsRequiredQueryTests.cs:** Unit tests for CountNewServiceAccountSlotsRequiredQuery
- **src/Core/SecretsManager/Commands/ServiceAccounts/Interfaces/IAutoscaleServiceAccountsCommand.cs:** Command interface
- **src/Core/SecretsManager/Queries/ServiceAccounts/Interfaces/ICountNewServiceAccountSlotsRequiredQuery.cs:** Query interface

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3121, opened 2023-07-19
[AC-1461] Secrets manager seat autoscaling

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Implement seat autoscaling and seat limits for Secrets Manager users.

### Auth team
Hello! I expect the only change to your files is to register some services in Identity. This is because we added these dependencies to `OrganizationService`, which unfortunately affects every project that uses `OrganizationService`. This is temporary until we refactor, as noted in the comments.

### SM team
Hola! When I integrated the new autoscaling code into `OrganizationUsersController.BulkEnableSecretsManagerAsync`, there was no longer any need for the `EnableAccessSecretsManagerCommand`. Most of that command was spent verifying users and collating success/error messages to be returned to the client. However, those error messages are never displayed in the UI, and users who already have SM access are now filtered out in the controller. I have deleted the command and just called the repository directly from the controller.

### AC team
Hey. See my other notes above. Other changes:
* this generally follows the same pattern as #3059 - a new `CountNewSmSeatsRequiredQuery` returns how many seats you need, and that's passed to `UpdateSecretsManagerSubscriptionCommand` to make the change.
* the code is integrated wherever you can give an orgUser access to SM:
  * `OrganizationService` - invite & save user flows
  * `OrganizationUsersController` - bulk grant access to SM endpoint
* I've also gone ahead and incorporated the check for beta users from #3112 
* the biggest change is a refactor to the `SecretsManagerSubscriptionUpdate` object:
  * the main properties that you can set mirror those on the `Organization` object, that is: `SmSeats`, `SmServiceAccounts`, `MaxAutoscaleSmSeats`, and `MaxAutoscaleSmServiceAccounts`
  * all other properties are now getters that calculate the result, instead of being calculated once in the constructor and then possibly becoming out of date later (e.g. previously if you changed `SmSeats`, `SmSeatsAdjustment` would be wrong.)
  * there's a new constructor that takes  the `Organization` and populates the update object using current values. The user can then override values or use the `Adjust` methods to add/subtract. This simplifies instantiation compared to the previous constructor and avoids mistakes
  * **why do this?** It avoids adding many extra methods to the command. We already have `AdjustServiceAccounts`, and this PR would otherwise add `AdjustSmSeats`, `ValidateAdjustSmSeats` and then `RevertSmSeats` - basically because correctly instantiating the update object is difficult at the moment, so we were hiding it in the command. This refactor makes the update object responsible for itself, and keeps the command simple. In general I think it makes it easier to understand.
  * to limit the scope of the change at this point, I've left the old constructor in place which means no update to the existing callers, or the test code. And tests still pass :)

## Before you submit


src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
```diff
@@ -79,6 +83,25 @@ public async Task AdjustServiceAccountsAsync(Organization organization, int smSe
...
+        }
+
+        var update = new SecretsManagerSubscriptionUpdate(
+            organization, seatAdjustment: smSeatAdjustment, maxAutoscaleSeats: organization?.MaxAutoscaleSmSeats,
+            serviceAccountAdjustment: 0, maxAutoscaleServiceAccounts: organization?.MaxAutoscaleSmServiceAccounts);
+
+        if (organization.MaxAutoscaleSmSeats.HasValue && update.SmSeats > organization.MaxAutoscaleSmSeats.Value)
+        {
+            throw new BadRequestException("Secrets Manager seat limit has been reached.");
+        }
```
eliykat, line 100: The main update command will throw an exception in this case anyway, `Cannot set max seat autoscaling below seat count.` But this is a more appropriate error message for the user to see when trying to autoscale.

r-tome (changes requested):
Overall it looks great! 
There's a few things I pointed out that should be sorted:
- missing awaits;
- failing integration test.
One extra thing that I'd like would be unit tests around the catch block reverting changes on `OrganizationService.SaveUsersSendInvitesAsync`

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -835,6 +846,21 @@ public async Task DisableTwoFactorProviderAsync(Organization organization, TwoFa
...
         }
 
+        // Secrets Manager seat autoscaling
+        SecretsManagerSubscriptionUpdate smSubscriptionUpdate = null;
+        var inviteWithSmAccessCount = invites
+            .Where(i => i.invite.AccessSecretsManager)
+            .SelectMany(i => i.invite.Emails)
+            .Count(email => !existingEmails.Contains(email));
+
+        var additionalSmSeatsRequired = await _countNewSmSeatsRequiredQuery.CountNewSmSeatsRequiredAsync(organization.Id, inviteWithSmAccessCount);
```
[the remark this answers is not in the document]
eliykat: Great catch!

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -942,11 +973,25 @@ public async Task DisableTwoFactorProviderAsync(Organization organization, TwoFa
             // Revert any added users.
             var invitedOrgUserIds = orgUsers.Select(u => u.Id).Concat(limitedCollectionOrgUsers.Select(u => u.Item1.Id));
             await _organizationUserRepository.DeleteManyAsync(invitedOrgUserIds);
-            var currentSeatCount = (await _organizationRepository.GetByIdAsync(organization.Id)).Seats;
+            var currentOrganization = await _organizationRepository.GetByIdAsync(organization.Id);
 
-            if (initialSeatCount.HasValue && currentSeatCount.HasValue && currentSeatCount.Value != initialSeatCount.Value)
+            // Revert autoscaling
+            if (initialSeatCount.HasValue && currentOrganization.Seats.HasValue && currentOrganization.Seats.Value != initialSeatCount.Value)
```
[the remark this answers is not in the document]
eliykat: Done, let me know what you think!

src/Core/Models/Business/SecretsManagerSubscriptionUpdate.cs
```diff
@@ -8,87 +9,95 @@ public class SecretsManagerSubscriptionUpdate
...
-    public int SmServiceAccountsAdjustment { get; set; }
+    public DateTime? ProrationDate { get; set; }
 
     /// <summary>
-    /// The total service accounts the organization will have after the update, including the base service accounts
-    /// included in the plan
+    /// Whether the subscription update is a result of autoscaling
     /// </summary>
-    public int SmServiceAccounts { get; set; }
+    public bool Autoscaling { get; init; }
```
eliykat, line 42: @r-tome I'm undecided whether the Autoscaling property is a good idea. It complicates the logic a bit, and it's only used (mostly) to show more accurate error messages. Could we just make the error messages more general and then remove this property? I might experiment on Monday but wanted your thoughts.

r-tome: I doesn't look terrible but if we can simplify it then I think we should.

One possibility on how we could get around this would be have all validation methods moved to `SecretsManagerSubscriptionUpdate` and then create another class that inherits from that (say `SecretsManagerSubscriptionAutoscale`) and override the validation methods with autoscale logic and appropriate messages. I'm not sure if that would work and maybe it's overcomplicating it. We can sync up sometime and discuss it.

src/Core/Services/Implementations/OrganizationService.cs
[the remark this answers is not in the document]
r-tome: Excellent work!

r-tome (approved):
It looks ready for me. We could change the `Autoscale` property but that can be refactored later.

jlf0dev (approved):
Looks good for Auth 🚀

test/Core.Test/Services/OrganizationServiceTests.cs
```diff
@@ -586,6 +589,97 @@ await sutProvider.GetDependency<IMailService>().Received(1)
...
+
+        // Mock SecretsManagerSubscriptionUpdateCommand to actually change the organization's subscription in memory
+        sutProvider.GetDependency<IUpdateSecretsManagerSubscriptionCommand>()
+            .UpdateSubscriptionAsync(Arg.Any<SecretsManagerSubscriptionUpdate>())
+            .ReturnsForAnyArgs(Task.FromResult(0)).AndDoes(x => organization.SmSeats += invitedSmUsers);
+
+        // Throw error at the end of the try block
+        sutProvider.GetDependency<IReferenceEventService>().RaiseEventAsync(default).ThrowsForAnyArgs<BadRequestException>();
+
+        sutProvider.Sut.InviteUsersAsync(organization.Id, savingUser.Id, invites);
```
r-tome, line 643:
```suggestion
        await Assert.ThrowsAsync<AggregateException>(async () => await sutProvider.Sut.InviteUsersAsync(organization.Id, savingUser.Id, invites));
```

src/Identity/Startup.cs
```diff
@@ -138,6 +140,10 @@ public void ConfigureServices(IServiceCollection services)
         services.AddDefaultServices(globalSettings);
         services.AddCoreLocalizationServices();
 
+        // TODO: Remove when OrganizationUser methods are moved out of OrganizationService, this noop dependency should
+        // TODO: no longer be required - see PM-1880
```
r-tome, line 144: PM-1880 won't fix this because `IServiceAccountRepository` will still be a project depedency, it'll just be moved to the command instead

src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -426,7 +430,7 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRest
 
     [HttpPatch("enable-secrets-manager")]
     [HttpPut("enable-secrets-manager")]
-    public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkEnableSecretsManagerAsync(Guid orgId,
+    public async Task BulkEnableSecretsManagerAsync(Guid orgId,
```
Thomas-Avery, line 433: Are we going to have a client PR to clean up the `OrganizationUserServiceImplementation` expecting the list response?

Thomas-Avery (approved):
SM changes look good. One question about cleaning up the client.

test/Core.Test/Services/OrganizationServiceTests.cs
```diff
@@ -586,6 +589,97 @@ await sutProvider.GetDependency<IMailService>().Received(1)
...
+
+        // Set up some invites to grant access to SM
+        invites.First().invite.AccessSecretsManager = true;
+        var invitedSmUsers = invites.First().invite.Emails.Count();
+
+        // Assume we need to add seats for all invited SM users
+        sutProvider.GetDependency<ICountNewSmSeatsRequiredQuery>()
+            .CountNewSmSeatsRequiredAsync(organization.Id, invitedSmUsers).Returns(invitedSmUsers);
+
+        sutProvider.Sut.InviteUsersAsync(organization.Id, savingUser.Id, invites);
```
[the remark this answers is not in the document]
eliykat: Thank you for catching all these missing `await`s, not sure what happened there. Too much SM billing work. (Why does C# even let me do this??)

test/Core.Test/Services/OrganizationServiceTests.cs
eliykat: Nice add!

src/Identity/Startup.cs
eliykat: But the command will be called directly from the Api controller, so it'll only be registered in the Api project, so we won't need it or its dependencies here (in Identity). Or in the other random projects we've added it to here.

src/Api/Controllers/OrganizationUsersController.cs
eliykat: Great point! PR here: https://github.com/bitwarden/clients/pull/5924

src/Core/Models/Business/SecretsManagerSubscriptionUpdate.cs
eliykat: I have considered moving the validation checks inside the object, I'm undecided on it. It might be better - you could have an `OrganizationSubscriptionView` in the db which you query, that returns everything you need including the current usage etc, then `Validate` just compares the internal properties of the object.

I don't think that it helps with this particular question though, one thing I want to ensure is that anyone calling this does not accidentally skip autoscale validation checks. The nice thing about the constructor signature at the moment is that you _have_ to specify whether you're autoscaling or not. Compare to needing to instantiate a subclass, or call a different method - both would be easy to miss.

I reviewed today and it's not as bad as I remember from Friday. I think we can leave as is.

eliykat:
@r-tome I made some changes based on https://bitwarden.atlassian.net/browse/SM-872, and some minor refactors to help me understand the code flow better. That is the last change I'll make to this PR, if Sukhleen comes back with anything else I'll do a separate PR for it.

r-tome (approved):
Great work!

PR #3132, opened 2023-07-25
[PM-3176] Extract IOrganizationService.SaveUserAsync to a command

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Extract `OrganizationService.SaveUserAsync` to a command.

## Code changes

* **src/Api/Controllers/OrganizationUsersController.cs:** Replaced `IOrganizationService.UpdateUserAsync` with `IUpdateOrganizationUserCommand`
* **src/Api/Public/Controllers/MembersController.cs:** Replaced `IOrganizationService.UpdateUserAsync` with `IUpdateOrganizationUserCommand`
* **src/Core/OrganizationFeatures/OrganizationServiceCollectionExtensions.cs:** Registered `UpdateOrganizationUserCommand` for DI
* **src/Core/OrganizationFeatures/OrganizationUsers/Interfaces/IUpdateOrganizationUserCommand.cs:** Command interface
* **src/Core/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommand.cs:** Command with code extracted from `OrganizationService`
* **src/Core/Services/IOrganizationService.cs:** Deleted method `SaveUserAsync` that was extracted
* **src/Core/Services/Implementations/OrganizationService.cs:** Deleted method `SaveUserAsync` that was extracted
* **test/Core.Test/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommandTests.cs:** Added unit tests for `UpdateOrganizationUserCommand`
* **test/Core.Test/Services/OrganizationServiceTests.cs:** Tests moved to `UpdateOrganizationUserCommandTests`

## Before you submit


src/Core/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            var additionalSmSeatsRequired = await _countNewSmSeatsRequiredQuery.CountNewSmSeatsRequiredAsync(user.OrganizationId, 1);
+            if (additionalSmSeatsRequired > 0)
+            {
+                var organization = await _organizationRepository.GetByIdAsync(user.OrganizationId);
+                var update = new SecretsManagerSubscriptionUpdate(organization, true);
+                update.AdjustSeats(additionalSmSeatsRequired);
+                await _updateSecretsManagerSubscriptionCommand.UpdateSubscriptionAsync(update);
+            }
+        }
```
eliykat, line 78: This PR mostly looks like a cut & paste, which generally works fine, except here. Commands shouldn't call queries or other commands. I think this means that it needs to be extracted out to the controller level, which calls the commands/queries in order.

This is how the service account code works, which is similar: https://github.com/bitwarden/server/blob/master/src/Api/SecretsManager/Controllers/ServiceAccountsController.cs#L123-L134

The difference here is that both our commands have validation requirements that should be checked first. We don't want to update the orgUser without knowing we can scale the subscription, and vice versa. This is why we have `UpdateSecretsManagerSubscriptionCommand.ValidateAsync` - maybe call that first, then update the orgUser, then call `UpdateSubscriptionAsync`. Or add your own `ValidateAsync` interface to this command if there's a reason to do it the other way around.

Note that the public API doesn't support updating this property, so it doesn't need this code.

r-tome: Excellent point! I forgot that dependencies between commands/queries should not happen.
I've added a few unit tests for `OrganizationUsersController` to be sure the logic was correct.

src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -501,4 +513,26 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRest
...
+        {
+            var organization = await _organizationRepository.GetByIdAsync(updatedOrganizationUser.OrganizationId);
+            var update = new SecretsManagerSubscriptionUpdate(organization, true);
+            update.AdjustSeats(additionalSmSeatsRequired);
+            await _updateSecretsManagerSubscriptionCommand.ValidateUpdate(update);
+            await _updateOrganizationUserCommand.UpdateUserAsync(updatedOrganizationUser, savingUserId,
+                model.Collections?.Select(c => c.ToSelectionReadOnly()), model.Groups);
+            // Only autoscale (if required) after all validation has passed so that we know it's a valid request before
+            // updating Stripe
+            await _updateSecretsManagerSubscriptionCommand.UpdateSubscriptionAsync(update);
```
eliykat, line 530: This is new, but I think we need a try/catch block around this so that we revert the change to the `OrganizationUser` if the payment fails.

`ValidateUpdate` will pre-validate the update against our data, but it doesn't contact Stripe. I assume the payment could still fail if their card is expired, for example. (or a network outage)

eliykat (changes requested):

src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -498,4 +506,36 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRest
         return new ListResponseModel<OrganizationUserBulkResponseModel>(result.Select(r =>
             new OrganizationUserBulkResponseModel(r.Item1.Id, r.Item2)));
     }
+
+    private async Task HandleOrganizationUserUpdateAsync(OrganizationUser updatedOrganizationUser, Guid savingUserId,
+        OrganizationUserUpdateRequestModel model, int additionalSmSeatsRequired)
+    {
+        if (additionalSmSeatsRequired > 0)
```
eliykat, line 513: Invert conditional to reduce nesting. Maybe even move the simple case back up into the endpoint so you only have the private method for the more complex SM update logic.

src/Api/Controllers/OrganizationUsersController.cs
```diff
@@ -498,4 +506,36 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRest
...
+                // updating Stripe
+                await _updateSecretsManagerSubscriptionCommand.UpdateSubscriptionAsync(update);
+                await _updateOrganizationUserCommand.UpdateUserAsync(updatedOrganizationUser, savingUserId,
+                    model.Collections?.Select(c => c.ToSelectionReadOnly()), model.Groups);
+            }
+            catch
+            {
+                // If we fail to update Stripe, we need to revert the update to Organization
+                await _organizationService.UpdateAsync(organization);
+            }
```
eliykat, line 533: This isn't quite right. The problem is that the sub update and the orgUser update need to pass together or fail together. So you can either do this:

1. validate subscription update
2. do the orgUser update (which includes validation)
4. do the subscription update
5. if (3) fails for some other reason - revert the orgUser update

OR this:

1. validate the orgUser update
2. do the subscription update (which includes validation)
3. do the orgUser update
4. if (3) fails for some other reason - revert the subscription update

In both cases, note that if (2) fails, there's nothing to revert yet.

Here, the `SecretsManagerSubscriptionUpdateCommand` already has a validate interface, and I think internal db changes are easier to revert than external Stripe charges, so I lean towards the first order of operations.

If you agree, compare it to your code - I think you need to swap the order you're doing the updates in (in the `try` block), and revert the orgUser updates, not the organization (in the `catch` block).

src/Core/OrganizationFeatures/OrganizationUserCollectionExtensions.cs
eliykat, line 1: I think this could be moved within the `OrganizationUsers` subfolder for better organization.

eliykat (changes requested):

eliykat:
@r-tome as discussed, when you come back to this I think you can do a simple lift-and-shift of the current code into a command, and we can refine it later. Basically, whatever the minimal change is to get it out of OrganizationService.

PR #3137, opened 2023-07-27
Update Setup project to handle EU region

vgrassia:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This PR adds the ability to select the EU region to the Setup container for self-host installations.  This will fix the issue of being able to use an `installation id` and `installation key` generated for the EU region.

## Code changes

* **util/Setup/Context.cs:** Add the `CloudRegion` enum to Context.
* **util/Setup/Enums/CloudRegion.cs:** Add `CloudRegion` enum to keep track of regions.
* **util/Setup/EnvironmentFileBuilder.cs:** Sets the `globalSettings__baseServiceUri__cloudRegion` in the `global.override.env` file.
* **util/Setup/Helpers.cs:** Update prompt to show the `N` is the default if you hit enter without providing a value. Add `WriteError` function to print out error messages.
* **util/Setup/Program.cs:** Add input validation to questions for `installation id`, `installation key`, and `cloud region`. Also, uses the proper URL when validating the `installation id`.

## Before you submit

### DIFF
--- a/util/Setup/Program.cs
+++ b/util/Setup/Program.cs
@@ -207,7 +207,7 @@ public class Program
         {
             var prompt = "Enter your installation id (get at https://bitwarden.com/host)";
             installationId = Helpers.ReadInput(prompt);
-            while (installationId == string.Empty)
+            while (string.IsNullOrEmpty(installationId))
             {
                 Helpers.WriteError("Invalid input for installation id. Please try again.");
                 installationId = Helpers.ReadInput(prompt);
@@ -228,7 +228,7 @@ public class Program
         {
             var prompt = "Enter your installation key";
             installationKey = Helpers.ReadInput(prompt);
-            while (installationKey == string.Empty)
+            while (string.IsNullOrEmpty(installationKey))
             {
                 Helpers.WriteError("Invalid input for installation key. Please try again.");
                 installationKey = Helpers.ReadInput(prompt);
@@ -243,13 +243,13 @@ public class Program
         {
             var prompt = "Enter your region (US/EU) [US]";
             var region = Helpers.ReadInput(prompt);
-            if (region == string.Empty) region = "US";
+            if (string.IsNullOrEmpty(region)) region = "US";
 
             while (!Enum.TryParse(region, out cloudRegion))
             {
                 Helpers.WriteError("Invalid input for region. Please try again.");
                 region = Helpers.ReadInput(prompt);
-                if (region == string.Empty) region = "US";
+                if (string.IsNullOrEmpty(region)) region = "US";
             }
         }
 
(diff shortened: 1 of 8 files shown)


util/Setup/Program.cs
```diff
@@ -218,26 +226,61 @@ private static bool ValidateInstallation()
...
+            while (installationKey == string.Empty)
+            {
+                Helpers.WriteError("Invalid input for installation key. Please try again.");
+                installationKey = Helpers.ReadInput(prompt);
+            }
+        }
+
+        if (_context.Parameters.ContainsKey("cloud-region"))
+        {
+            Enum.TryParse(_context.Parameters["cloud-region"], out cloudRegion);
```
justindbaur, line 240: This will take something like `blah` and treat it like `CloudRegion.US`, is that what we want?

util/Setup/Program.cs
```diff
@@ -218,26 +226,61 @@ private static bool ValidateInstallation()
...
+            var prompt = "Enter your installation key";
+            installationKey = Helpers.ReadInput(prompt);
+            while (installationKey == string.Empty)
+            {
+                Helpers.WriteError("Invalid input for installation key. Please try again.");
+                installationKey = Helpers.ReadInput(prompt);
+            }
+        }
+
+        if (_context.Parameters.ContainsKey("cloud-region"))
```
justindbaur, line 238: I generally prefer `_context.Parameters.TryGetValue("cloud-region", out var cloudRegion)` because you access the dictionary only once and you save yourself from having the keys differ in both their uses, but I see that isn't the pattern here so I will leave it up to you.

util/Setup/Program.cs
```diff
@@ -196,14 +197,21 @@ private static bool ValidateInstallation()
...
         if (_context.Parameters.ContainsKey("install-id"))
         {
             installationId = _context.Parameters["install-id"].ToLowerInvariant();
         }
         else
         {
-            installationId = Helpers.ReadInput("Enter your installation id (get at https://bitwarden.com/host)");
+            var prompt = "Enter your installation id (get at https://bitwarden.com/host)";
+            installationId = Helpers.ReadInput(prompt);
+            while (installationId == string.Empty)
```
justindbaur, line 210: It's generally safer to do `string.IsNullOrEmpty()` and same for all the other `== string.Empty` checks.

vgrassia: Updated all of them to `string.IsNullOrEmpty()`.

util/Setup/Program.cs
vgrassia: I think I'd like to continue following the pattern here instead of refactoring all of it this time.

util/Setup/Program.cs
vgrassia: Yes.  We currently don't have a use for running the setup container with `cloud-region` specified, but for backwards compatibility I believe the default should be `US`.

justindbaur (approved):
Looks great, thanks for the changes!

joseph-flinn (approved):

PR #3147, opened 2023-08-01
[AC-1571] Handle null reference exception in stripe controller

cturnbull-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Occasionally, Stripe will send an event for an object that is associated with a customer that has `null` metadata. This appears to happen only in the event that the customer itself is deleted on Stripe's side, as even customers without any configured metadata still have their metadata appear as an empty `Dictionary<string, string>` in the Stripe SDK customer entity. This PR aims to address this edge case by first ensuring that the customer metadata isn't `null` before performing any operations on it. If it is `null`, then we just return `200` to prevent it from being shown as an error.

Additionally, the logging around attempting to take a PayPal payment through the StripeController has been added to. The goal is to improve visibility in the logs into errors that happen during this flow.

## Code changes

* **StripeController.cs:** Handling `null` customer metadata objects, and added PayPal error logging

## Before you submit


src/Billing/Controllers/StripeController.cs
```diff
@@ -501,27 +507,29 @@ private async Task<bool> ValidateCloudRegionAsync(Event parsedEvent)
             case HandledStripeWebhook.SubscriptionUpdated:
                 {
                     var subscription = await GetSubscriptionAsync(parsedEvent, true, new List<string> { "customer" });
-                    customerRegion = GetCustomerRegionFromMetadata(subscription.Customer.Metadata);
+                    customerRegion = GetCustomerRegionFromMetadata(subscription.Customer?.Metadata);
```
cyprain-okeke, line 510: The look Great. Is it possible that subscription object will be null? Can we check for null

src/Billing/Controllers/StripeController.cs
```diff
@@ -501,27 +507,29 @@ private async Task<bool> ValidateCloudRegionAsync(Event parsedEvent)
...
                     break;
                 }
             case HandledStripeWebhook.ChargeSucceeded:
             case HandledStripeWebhook.ChargeRefunded:
                 {
                     var charge = await GetChargeAsync(parsedEvent, true, new List<string> { "customer" });
-                    customerRegion = GetCustomerRegionFromMetadata(charge.Customer.Metadata);
+                    customerRegion = GetCustomerRegionFromMetadata(charge.Customer?.Metadata);
```
cyprain-okeke, line 517: Same of charge object. can we check for null before access Customer

cyprain-okeke (commented):
It looks good.  
But Are you are able to resolve the Checkmarx issues? 
And Please try an test  it.

src/Billing/Controllers/StripeController.cs
```diff
@@ -490,60 +496,85 @@ public async Task<IActionResult> PostWebhook([FromQuery] string key)
...
-                    customerRegion = GetCustomerRegionFromMetadata(invoice.Customer.Metadata);
-                    break;
-                }
-            default:
-                {
-                    // For all Stripe events that we're not listening to, just return 200
-                    return false;
-                }
-        }
+            var customerMetadata = eventType switch
```
cyprain-okeke, line 505: I like the previous switch statement, it seems more readable for me. What do you think?

cturnbull-bitwarden: I've been enjoying using pattern matching/expressions, but can definitely update it to use a switch statement if it's not readable. I posted about it in our #code channel [here](https://bitwarden.slack.com/archives/C02LPDKPML6/p1687958869288269). I'll go ahead and update it to be a switch statement, though, as I don't have a strong opinion one way or the other!

cyprain-okeke (approved):
it looks great.

PR #3152, opened 2023-08-03
PM-2128 Enforce one time use of TOTP

ike-kottlowski:
## Type of change

- [ ] Bug 
- [ ] New feature development
- [x] Other

## Objective

The objective is to enforce using TOTP once to be compliant with [RFC 6238 - TOTP: Time-Based One-Time Password Algorithm](https://tools.ietf.org/html/rfc6238)

## Code changes

* **src\Identity\IdentityServer\BaseRequestValidator.cs:** 
  * Added DI for IDistributedCache to allow service to request check if an TOPT has been used already. The service doesn't distinguish between TOPT methods (email/duo/authenticator).
* All other files are updated to match the base class constructor

## Before you submit


jlf0dev (approved):

src/Identity/IdentityServer/BaseRequestValidator.cs
```diff
@@ -133,6 +142,15 @@ public abstract class BaseRequestValidator<T> where T : class
             var verified = await VerifyTwoFactor(user, twoFactorOrganization,
                 twoFactorProviderType, twoFactorToken);
 
+            var isOtpCached = Core.Utilities.DistributedCacheExtensions.TryGetValue(_distributedCache, user.Email, out string cachedToken);
```
withinfocus, line 145: ⛏️ You can eliminate the unused output variable with just `_`.

src/Identity/IdentityServer/BaseRequestValidator.cs
```diff
@@ -87,6 +91,11 @@ public abstract class BaseRequestValidator<T> where T : class
         _tokenDataFactory = tokenDataFactory;
         FeatureService = featureService;
         SsoConfigRepository = ssoConfigRepository;
+        _distributedCache = distributedCache;
+        _cacheEntryOptions = new DistributedCacheEntryOptions
+        {
+            AbsoluteExpirationRelativeToNow = new TimeSpan(0, 15, 0)
```
withinfocus, line 97: ❓ Is this the TOTP expiration window? Centralizing this setting or using whatever is configured elsewhere would help.

src/Identity/IdentityServer/BaseRequestValidator.cs
```diff
@@ -133,6 +142,15 @@ public abstract class BaseRequestValidator<T> where T : class
             var verified = await VerifyTwoFactor(user, twoFactorOrganization,
                 twoFactorProviderType, twoFactorToken);
 
+            var isOtpCached = Core.Utilities.DistributedCacheExtensions.TryGetValue(_distributedCache, user.Email, out string cachedToken);
+            if (isOtpCached)
+            {
+                // Delay for brute force.
+                await Task.Delay(2000);
+                await BuildErrorResultAsync("Two-step token has already been used.", true, context, user);
```
withinfocus, line 150: 💭 Should we reveal this vs. just saying it's invalid?

src/Identity/IdentityServer/BaseRequestValidator.cs
withinfocus, line 145: ❓ The task mentions considering the provider as well. Is that still needed?

src/Identity/IdentityServer/ResourceOwnerPasswordValidator.cs
```diff
@@ -44,11 +45,12 @@ public class ResourceOwnerPasswordValidator : BaseRequestValidator<ResourceOwner
...
         IDataProtectorTokenFactory<SsoEmail2faSessionTokenable> tokenDataFactory,
         IFeatureService featureService,
-        ISsoConfigRepository ssoConfigRepository)
+        ISsoConfigRepository ssoConfigRepository,
+        IDistributedCache distributedCache)
         : base(userManager, deviceRepository, deviceService, userService, eventService,
               organizationDuoWebTokenProvider, organizationRepository, organizationUserRepository,
               applicationCacheService, mailService, logger, currentContext, globalSettings, userRepository, policyService,
-              tokenDataFactory, featureService, ssoConfigRepository)
+              tokenDataFactory, featureService, ssoConfigRepository, distributedCache)
```
withinfocus, line 53: ❓ What scopes this cache to this use case of user-specific TOTP codes? Is this automatic and somehow scoped to the service? I don't have a lot of experience with `IDistributedCache`. My concern is another usage with email addresses could clash.

src/Identity/IdentityServer/BaseRequestValidator.cs
```diff
@@ -133,6 +142,15 @@ public abstract class BaseRequestValidator<T> where T : class
             var verified = await VerifyTwoFactor(user, twoFactorOrganization,
                 twoFactorProviderType, twoFactorToken);
 
+            var isOtpCached = Core.Utilities.DistributedCacheExtensions.TryGetValue(_distributedCache, user.Email, out string cachedToken);
+            if (isOtpCached)
+            {
+                // Delay for brute force.
+                await Task.Delay(2000);
```
withinfocus, line 149: ❓ Do our higher-level protections like rate limiting already cover this?

withinfocus (commented):
Been following the task due to the implied security issues and have a few comments on the change.

src/Identity/IdentityServer/BaseRequestValidator.cs
kspearrin: Yes

src/Identity/IdentityServer/BaseRequestValidator.cs
ike-kottlowski: The issue sloved here is Authenticator TOTPs have a 1 minute TTL, and email TOTPs has a 5 minute TTL. Based on the RFC the recomendation is to accept TOTPs where T = T+1, where T is the vaild time-window step. Meaning that all TOTPs are vaild for the window they are created in and the next (T+1). This means an Authenticator TOTP will be vaild for 2 minutes, and an email TOTP will be vaild for 10 minutes. This is how it is implemented in OTP.NET library we use.

Chaching the TOTPs for 15 minutes ensures an Email and Authenticator TOTP is well past the T+1 mark. It also ensures if the clocks are misaligned between servers the cache will not clear too soon and allow the use of a TOTP twice.

The TTL for the cache being hardcoded gives code smells but I believe in this situation it's permissible.

src/Identity/IdentityServer/BaseRequestValidator.cs
ike-kottlowski: Error message updated.

src/Identity/IdentityServer/BaseRequestValidator.cs
ike-kottlowski: Updated to use `_`.

ike-kottlowski: I don't think so. At its core this solution is to just track a user and the TOTPs they've used. I'm not sure why we care which provider it was. The possibility of a collision between TOTP providers I feel like is slim to none when considering most users probably use a single TOTP provider and the cache is cleared regularly.

If there is an argument for tracking the provider then I'm all for implementing it, I just didn't see the need, but I could also be missing something.

src/Identity/IdentityServer/ResourceOwnerPasswordValidator.cs
ike-kottlowski: It's a [singleton ](https://github.com/bitwarden/server/blob/d94a54516eafe2d86ca6288d19b54adf7b7bf284/src/SharedWeb/Utilities/ServiceCollectionExtensions.cs#L665)so this is an issue. I'll add information to the key to avoid any issues with the email being the Key.

src/Identity/IdentityServer/BaseRequestValidator.cs
withinfocus: Great explanation, thanks. I think a brief comment in code could help here.

src/Identity/IdentityServer/BaseRequestValidator.cs
withinfocus: Was this undone?

src/Identity/IdentityServer/BaseRequestValidator.cs
withinfocus: Also still here -- did a merge undo your changes?

src/Identity/IdentityServer/BaseRequestValidator.cs
withinfocus: Samesies.

withinfocus (commented):
Maybe I jumped the gun on re-review, but a few comments.

src/Identity/IdentityServer/ResourceOwnerPasswordValidator.cs
ike-kottlowski: Added "TOTP_" prefix to the key to address the singleton issue.

src/Identity/IdentityServer/BaseRequestValidator.cs
ike-kottlowski: Comment added.

JaredSnider-Bitwarden (approved):
LGTM! Nice work on this 🚀 !

PR #3153, opened 2023-08-03
[PM-3246] Check the user directly if they have the ResetPasswordKey

justindbaur:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Check the `OrganizationUser` directly for if they have the reset password key which is the better identifier for if the org has the policy and they have enrolled in that policy since that is what is needed for admin approval to work.

## Code changes

* **src/Identity/IdentityServer/BaseRequestValidator.cs:** Grab the user from `IOrganizationUserRepository` so that we can check their `ResetPasswordKey` value.
* **test/Common/Helpers/AssertHelper.cs:** Usability improvement so that when an exception comes from this assertion helper the first line in the stack trace is user code and can be clicked on to find the actual problem in the test code.
* **test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs:** Refactored tests for reuse of code, and fixed the has admin approval test.
## Before you submit

### DIFF
--- a/src/Identity/IdentityServer/BaseRequestValidator.cs
+++ b/src/Identity/IdentityServer/BaseRequestValidator.cs
@@ -3,12 +3,12 @@ using System.Reflection;
 using System.Security.Claims;
 using System.Text.Json;
 using Bit.Core;
+using Bit.Core.Auth.Entities;
 using Bit.Core.Auth.Enums;
 using Bit.Core.Auth.Identity;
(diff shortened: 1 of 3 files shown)


JaredSnider-Bitwarden (approved):
LGTM! I tested locally as well, and it properly resolved the issue!

PR #3155, opened 2023-08-04
[AC-1577] [AC-1575] Prefill Bitwarden portal values etc

eliykat:
## Type of change

```
- [x] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Improve default values and validation around Secrets Manager in the Bitwarden Portal, and fix a related bug. Specifically:

**Default values**
* when Secrets Manager is enabled for an org, set the initial service accounts value based on the amount included in the plan
* when an org migrates from SM Beta to SM GA, set their initial seats and service accounts based on their current usage
* when a CS-assisted trial is started for a Free Org, prefill 10 SmSeats (consistent with PM behaviour) and the base service accounts

**Validation**
* SM Beta requires SM access - enforce this by automatically selecting the SM checkbox if beta is selected, or deselecting the beta checkbox if SM is deselected
* reject the form if the plan does not support SM - to avoid undefined behaviour

**Bug fixes**
* ([AC-1575]) don't overwrite seats, storage, expiration date and license key configuration when the plan is changed
* Fix the Service Account field label by removing the word "Max", to avoid confusion with max autoscale

## Code changes

* `OrganizationsController` - add check for unsupported plans
* `OrganizationEditModel` - update label
* `OrganizationViewModel` - clarify these property names, this matches the naming convention for Collections etc.
* `Edit.cshtml` - per AC-1575, the `togglePlanSettings` function was doing too much and was incorrectly overwriting current values (seats, license key, etc) when the plan changed. This logic is only needed when initiating a trial, so move it here into its own `setTrialDefaults` function.
* `OrganizationFormScripts`
  *  rename `togglePlanSettings` to `togglePlanFeatures` and make it only set the features, not any other values
  * create `setInitialSecretsManagerConfiguration` function to set sane defaults (the higher of current usage or plan amounts), and `clearSecretsManagerConfiguration` to clear them. These are called when SM checkboxes are toggled
  * add `getBaseServiceAccountValue` because this is reused in various places

## Before you submit

[AC-1575]: https://bitwarden.atlassian.net/browse/AC-1575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Admin/Views/Shared/_OrganizationFormScripts.cshtml
```diff
@@ -103,12 +94,52 @@
...
+        document.getElementById('@(nameof(Model.SmSeats))').value = '';
+        document.getElementById('@(nameof(Model.SmServiceAccounts))').value = '';
+        document.getElementById('@(nameof(Model.MaxAutoscaleSmSeats))').value = '';
+        document.getElementById('@(nameof(Model.MaxAutoscaleSmServiceAccounts))').value = '';
+    }
+
+    function getBaseServiceAccountValue(planType) {
+        switch(planType) {
+            case '@((byte)PlanType.Free)':
+                return 3;
```
r-tome, line 131:
```suggestion
                return @Bit.Core.Utilities.StaticStore.GetSecretsManagerPlan(PlanType.Free).BaseServiceAccount;
```
This way we don't have to worry about updating this if the value ever changes. Same for the other plans.

src/Admin/Views/Organizations/Edit.cshtml
```diff
@@ -12,25 +12,45 @@
...
+
     <script>
         (() => {
             document.getElementById('teams-trial').addEventListener('click', () => {
                 if (document.getElementById('@(nameof(Model.PlanType))').value !== '@((byte)Bit.Core.Enums.PlanType.Free)') {
                     alert('Organization is not on a free plan.');
                     return;
                 }
-                togglePlanSettings('@((byte)Bit.Core.Enums.PlanType.TeamsAnnually)');
+                setTrialDefaults('@((byte)Bit.Core.Enums.PlanType.TeamsAnnually)');
```
r-tome, line 23: `Bit.Core.Enums` can be removed

src/Admin/Views/Shared/_OrganizationFormScripts.cshtml
eliykat: This turned into a long switch statement that basically just matched planType to baseServiceAccount for that plan, so I went 1 better and just started bringing the plan objects directly into JS so we can look them up as required. We could use this for the Feature checkboxes as well - moving away from hardcoded values and providing more dynamic behaviour here. Let me know what you think.

eliykat:
>     * The "Secrets Manager Beta" checkbox name could be clearer on what it is for. Maybe move it to the right sort of a sub-option to "Secrets Manager" a rename it to "Beta Access"?

It's a good idea, I've asked design what they think.

>     * Changing the Organization plan does not update the Secrets Manager values for Seats and Service Accounts (this should also happen when creating an organization for a Reseller Provider)

I'm reluctant to do this, as this caused the bug in AC-1575. We don't necessarily want to overwrite current values, maybe sometimes but not all the time, and I'm not sure what the business logic should be here. I think this needs more consultation with CS before implementing. Keep in mind that previously, changing the plan didn't update _anything_ (not even the Feature checkboxes!) so it should be expected (or at least reasonable) behaviour for the subscription to require manual adjustment.

When creating an org for a reseller, you should probably work top-to-bottom on the form. If you choose the plan first, _then_ enable SM, you'll get the right default values per the plan.

The whole Portal feels a little janky but this is probably the limit of the scope for this PR. Would love to invest some time into it in another initiative.

That said - I did notice that `setInitialSecretsManagerConfiguration` was using the saved PlanType, not necessarily the currently selected PlanType, so I have fixed that.

eliykat:
Design said to just leave the SM beta checkbox as-is. It'll be removed in a few months anyway.

r-tome (approved):
Great work!

eliykat:
@r-tome I just spotted a potential `null` and made a small change to cover it. I'll move on to QA in the meantime because I think it's straightforward but let me know if you disagree.

r-tome (approved):

shane-melton (approved):

PR #3168, opened 2023-08-09
[AC-108] Enterprise policies are not removed when downgraded to teams

r-tome:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
If an organization has `UsePolicies = false` then any of its members should not be affected by any previously configured policy.

I tested these changes by using a Service Bus resource on Azure to share `OrganizationAbilities` data between the projects.

## Code changes

- **src/Core/Models/Data/Organizations/OrganizationAbility.cs:** Add a property for the column `UsePolicies`
- **src/Core/Services/Implementations/PolicyService.cs:** Added a dependency for `IApplicationCacheService` and use that to check if an organization has policies enabled
- **src/Infrastructure.EntityFramework/Repositories/OrganizationRepository.cs:** Retrieving `UsePolicies` column value
- **src/Sql/dbo/Stored Procedures/Organization_ReadAbilities.sql:** Retrieving `UsePolicies` column value
- **test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs:** Fixed failing test by setting `UsePolicies = true`
- **test/Identity.IntegrationTest/Endpoints/IdentityServerTests.cs:** Fixed failing test by setting `UsePolicies = true`
- **util/Migrator/DbScripts/2023-08-09_00_OrgAbilitiesUsePolicies.sql:** Migration script to retrieve column `UsePolicies`

## Before you submit


src/Core/Services/Implementations/PolicyService.cs
```diff
@@ -206,7 +209,9 @@ private async Task<IEnumerable<OrganizationUserPolicyDetails>> QueryOrganization
         }
 
         var excludedUserTypes = GetUserTypesExcludedFromPolicy(policyType);
+        var orgAbilities = await _applicationCacheService.GetOrganizationAbilitiesAsync();
         return _cachedOrganizationUserPolicyDetails.Where(o =>
+            (!orgAbilities.ContainsKey(o.OrganizationId) || orgAbilities[o.OrganizationId].Enabled && orgAbilities[o.OrganizationId].UsePolicies) &&
```
eliykat, line 214: Please add brackets to make the order of evaluation clearer:

```suggestion
            (!orgAbilities.ContainsKey(o.OrganizationId) || (orgAbilities[o.OrganizationId].Enabled && orgAbilities[o.OrganizationId].UsePolicies)) &&
```

More importantly - this means that if the org isn't cached in `orgAbilities`, we assume policies do apply? Is there a situation where the org would not be cached?

eliykat: Also, have you confirmed with product that policies do not apply to disabled orgs? It's _probably_ fine, but it is a change in behaviour. For example, if the subscription lapses temporarily and is then reinstated, there's been a window where users could export their personal vault, or login without SSO, or remove 2FA from their account. And these may not be retroactively enforced when the subscription is reinstated. (e.g. users who removed 2FA will remain without 2FA)

src/Core/Models/Data/Organizations/OrganizationAbility.cs
eliykat, line 1: As I mentioned in Slack - how can we force the cache to update after we add this new property?

eliykat (changes requested):
Great work finding out how the cache works!

eliykat (changes requested):
You also made Bitwarden Portal changes in your other PR which need to be replicated here.

r-tome:
> You also made Bitwarden Portal changes in your other PR which need to be replicated here. Or are they covered by my PR?

They are mostly covered by your PR with the exception of unchecking features when downgrading to the 'Free' plan. Do you think that is good addition?

src/Core/Models/Data/Organizations/OrganizationAbility.cs
r-tome: I can think of two ways:
- ask the Devops team if the API application is restarted when deploying a new release. If it's not we'd have to ask them to force a restart.
- have `OrganizationAbility.UsePolicies` be nullable and then when retrieving the cached data check if that property has a value - if not then update the cached data.

What do you think?

src/Core/Services/Implementations/PolicyService.cs
r-tome: > More importantly - this means that if the org isn't cached in orgAbilities, we assume policies do apply? Is there a situation where the org would not be cached?

I added this as a precaution in the case we have a connection issue on `ServiceBus` and then the new `OrgAbility` wouldn't propagate throughout the different projects (for example creating an Organization in the Admin project). Does this make sense (or am I overthinking)? 

> Also, have you confirmed with product that policies do not apply to disabled orgs? It's probably fine, but it is a change in behaviour. 

I did not realise I was changing the behaviour, I have reverted this.

eliykat:
> > You also made Bitwarden Portal changes in your other PR which need to be replicated here. Or are they covered by my PR?
> 
> They are mostly covered by your PR with the exception of unchecking features when downgrading to the 'Free' plan. Do you think that is good addition?

I do think it's a good addition, but I also know that Sukhleen is working on some more comprehensive UI/UX improvements here, so let's leave it for that initiative.

src/Core/Models/Data/Organizations/OrganizationAbility.cs
eliykat: After discussing this further, this will be done automatically when the `InMemoryApplicationCacheService` initialises itself after the update.

src/Core/Services/Implementations/PolicyService.cs
eliykat: lgtm, and I think it's sensible to default to the stricter behaviour (assuming policies apply) if the value is missing for any reason.

eliykat (approved):
LGTM! Thanks for stepping me through it.

PR #3186, opened 2023-08-11
[PM-289] Add taxIdType for more countries in TaxInfo

cturnbull-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When adding tax information to a payment method for certain countries, the VAT/GST Tax ID would fail to save. This PR addresses this by adding the remaining `taxIdType` values for all possible countries supported by Stripe.

## Code changes

* **TaxInfo.cs:** Updated the getter for TaxIdType to include all possible tax ID type for every country that stripe supports

## Before you submit


src/Core/Models/Business/TaxInfo.cs
```diff
@@ -37,12 +37,24 @@ public string TaxIdType
 
             switch (BillingAddressCountry)
```
cyprain-okeke, line 38: It looks Good. 

If this switch grows or changes frequently, you might consider refactoring it. Using a Dictionary<string, string> for mapping country codes to tax ID types can be more efficient and easier to maintain.

```
private static readonly Dictionary<string, string> TaxIdTypeMap = new Dictionary<string, string>
{
    { "AD", "ad_nrt" },
    { "AE", "ae_trn" },
    { "AF", "af_tin" }, 
    { "AR", "ar_cuit" },
    // ... add all other mappings ...
    { "ZA", "za_vat" },
};
```

cyprain-okeke, line 38: It's also important to handle potential issues with input data. If BillingAddressCountry could be in lowercase, ensure it's transformed to uppercase before the lookup.

src/Core/Models/Business/TaxInfo.cs
```diff
@@ -37,12 +37,24 @@ public string TaxIdType
...
             {
+                case "AD":
+                    _taxIdType = "ad_nrt";
+                    break;
                 case "AE":
```
cyprain-okeke, line 43: There's a repeated value for the country code "AF" - it has the tax ID type "ae_trn", which appears to be the same as "AE". Please, ensure this isn't a mistake.

src/Core/Models/Business/TaxInfo.cs
cturnbull-bitwarden: Nice, thanks for these suggestions. I'll push up a change to address them

cyprain-okeke (approved):
Look Good

PR #3188, opened 2023-08-13
[PM-3356][PM-3292] Allow OTP For All

justindbaur:
- On a trusted device isn't a good check because a user might be using a trusted device locally but not trusted it long term
- The logic wasn't working for KC users anyways

## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

My logic wasn't good for KC users in #3184 and also didn't work for users of TDE where they chose not to remember the device on the server. After discussion, we are going to allow the use of OTP for all, we still only show it on the client side as an option when absolutely needed but OTP will work for any user with an email.

## Code changes

* **src/Core/Services/Implementations/UserService.cs:** Remove all limitations from OTP request and verification.
* **test/Core.Test/Services/UserServiceTests.cs:** Add tests for secret verification (with hack).

## Before you submit

### DIFF
--- a/test/Core.Test/Services/UserServiceTests.cs
+++ b/test/Core.Test/Services/UserServiceTests.cs
@@ -186,8 +186,6 @@ public class UserServiceTests
         OTP = 0x2,
     }
 
-    // This test currently doesn't work because of UserService not using my IPasswordHasher that I mocked up, giving up for now
-    // but this code is a good starter so I wanted it here.
     [Theory]
     // A user who has a password, and the password is valid should only check for that password
     [BitAutoData(true, "test_password", true, ShouldCheck.Password)]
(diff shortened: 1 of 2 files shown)


test/Core.Test/Services/UserServiceTests.cs
```diff
@@ -212,79 +188,91 @@ public enum ShouldCheck
 
     // This test currently doesn't work because of UserService not using my IPasswordHasher that I mocked up, giving up for now
```
JaredSnider-Bitwarden, line 189: Is this comment still accurate?

JaredSnider-Bitwarden (approved):
LGTM with one minor question

JaredSnider-Bitwarden (approved):
LGTM!

PR #3189, opened 2023-08-14
[AC-1568] Finish refactor of update subscription interface, minor fixes

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Follow-on from #3210, which will be merged first.

This updates the production code to use the new constructor and removes the old constructor.

It also contains some minor opportunistic refactors and fixes.

## Code changes

**SecretsManagerSubscriptionUpdate object**
* delete old ctor and update the last of the code that used it
* make the "adjust" methods return the object so that you can chain these calls

**UpdateSecretsManagerSubscriptionCommand**
* update error messages to not capitalise "service accounts"
* fix an incorrect error message re: current usage, it didn't correctly describe the problem
* reduce nesting in the internal method calls - 1 deep is fine, but beyond that I think it's hard to understand
* simplify interfaces and logic, we ended up attaching everything to the `update` object so let's just pass that around
* throw if the org is using SM Beta, and add a test for this

**AddSecretsManagerSubscriptionCommand**
* throw if the org is using SM already, or has SM Beta, and add tests for this

## Before you submit


src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
```diff
@@ -51,112 +49,76 @@ public async Task UpdateSubscriptionAsync(SecretsManagerSubscriptionUpdate updat
...
 
-        if (update.MaxAutoscaleSmServiceAccountsChanged)
-        {
-            organization.MaxAutoscaleSmServiceAccounts = update.MaxAutoscaleSmServiceAccounts;
-        }
+        var organization = update.Organization;
+        organization.SmSeats = update.SmSeats;
+        organization.SmServiceAccounts = update.SmServiceAccounts;
+        organization.MaxAutoscaleSmSeats = update.MaxAutoscaleSmSeats;
+        organization.MaxAutoscaleSmServiceAccounts = update.MaxAutoscaleSmServiceAccounts;
```
eliykat, line 86: Note: there's no point checking if these have changed before assigning the new values. If they haven't changed, assigning them won't do anything anyway.

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
```diff
@@ -341,30 +298,33 @@ private async Task ValidateSmServiceAccountsUpdateAsync(Organization organizatio
...
         {
             // autoscale limit has been turned off, no validation required
             return;
         }
 
-        if (organization.SmSeats.HasValue && maxAutoscaleSeats.Value < organization.SmSeats.Value)
+        if (update.SmSeats.HasValue && update.MaxAutoscaleSmSeats.Value < update.SmSeats.Value)
         {
             throw new BadRequestException($"Cannot set max Secrets Manager seat autoscaling below current Secrets Manager seat count.");
         }
```
eliykat, line 321: Note: this check was using the org's current `SmSeats` (out of date) rather than new `update.SmSeats`, which meant it sometimes threw incorrectly if you were updating both SmSeats and its autoscaling at the same time.

Same for the service account equivalent.

src/Core/Models/Business/SecretsManagerSubscriptionUpdate.cs
```diff
@@ -6,16 +6,15 @@ namespace Bit.Core.Models.Business;
 
 public class SecretsManagerSubscriptionUpdate
```
r-tome, line 7: Since we already have the methods `AdjustSeats` and `AdjustServiceAccounts` what do you think of adding methods `SetMaxAutoscaleSmSeats` and `SetMaxAutoscaleSmServiceAccounts` and then make the properties have `private set`?
I just think it looks a bit weird to use methods to set a few properties and then use the object initializer to set other properties.

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
r-tome: Nice catch!

r-tome (commented):
Looks great, I'm happy with this how it is but I think the `SecretsManagerSubscriptionUpdate` initialisation could be better.

src/Core/Models/Business/SecretsManagerSubscriptionUpdate.cs
eliykat: I don't really see any issue with it to be honest. We could move to all method setters, but there are times we also want to set seats and service accounts directly, so we'd end up with `SetSmSeats`, `SetSmServiceAccounts`, `SetMaxAutoscaleSmSeats` and `SetMaxAutoscaleSmServiceAccounts`, as well as the 2 `Adjust` methods. (Plus proration date?)

I thought that would be unnecessary boilerplate when we can just set those in the initializer. Whereas the `Adjust` methods do actually encapsulate some additional logic.

r-tome: Fair enough, that is a good point. Thanks for considering it.

r-tome (dismissed):
LGTM!

Thomas-Avery (dismissed):
SM changes look good.

r-tome (approved):

PR #3194, opened 2023-08-15
[AC-1174] CollectionUser and CollectionGroup authorization handlers

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
Introduce a new bulk authorization handler concept for `CollectionUser` and `CollectionGroup` resources. ~~Each resource has its own authorization handler, however for now, they do share a base `CollectionAccessAuthorizationHandlerBase` class to reduce redundancy in authorization logic.~~ 
We ultimately decided to go with a single `CollectionAuthorizationHandler` for collection access and collection creation/deletion. This PR adds the initial implementation and the `ModifyAccess` operation. 

### Authorization Handlers

* **BulkAuthorizationHandler.cs:** A new abstract authorization handler class to assist in handling bulk resource authorization requests. It's only responsibility is to find either individual or `IEnumerable` resources in the authorization context, and pass them to an abstract `HandleRequirementAsync()` that accepts the strongly typed `ICollection<TResource>` that can be more easily used by implementations.

* **CollectionAuthorizationHandler.cs:** The authorization handler the performs the bulk of the authorization logic for collection access. It ensures the acting user has the appropriate permissions for the requested collections. Should any of these requirements fail, the authorization request/context will fail.

### Testing Helpers

* **CollectionFixture.cs:** An AutoFixture customization attribute to help building up the tests for the new authorization handlers. Its responsible for organizing the relevant resources being tested and ensures they have the same `Ids` when requested in the same unit test. Helps reduce a lot of boilerplate assigning the same organization and collection id's to all the resources that are required to match in the authorization handlers.

* **AutoFixtureExtensions.cs:** Some helpful extension methods I created to help with the `CollectionFixture.cs` customization above. 
  * `WithGuidFromSeed()` - Allows you to customize entities to generate a Guid from a given seed. This ensures that all Guids generated for entities of different types, given the same seed, will always have the same Id values. e.g. If AutoFixture generates 3 `Collections` and 3 `CollectionUsers` and you give the same seed for the `Id` and `CollectionId`, respectively, the Ids will match up.
  * `WithValueFromList()` - Allows you to specify multiple values to be used by AutoFixture when populating properties. It will iterate through the list of resources and pick the next item in the list, eventually looping to the first item and repeating. This is used to specify multiple organization Ids in the `CollectionFixture` customization, as the handler is not restricted to a single organization and needs to be aware/test for resources that may belong to different orgs.

## Before you submit


src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+                                           join cu in resource on c.Id equals cu.CollectionId
+                                           where cu.OrganizationUserId == targetUser.Id
+                                           select c;
+
+            // Target user is not in the same org as a collection they're being assigned to, fail the requirement
+            if (targetCollectionsForUser.Any(tc => tc.OrganizationId != targetUser.OrganizationId))
+            {
+                context.Fail();
+                return;
+            }
```
eliykat, line 107: You could incorporate this into the join, e.g.

```cs
var validCollections = from c in targetCollections
  join cu in resource on c.Id equals cu.CollectionId
  where cu.OrganizationUserId == targetUser.Id
  where c.OrganizationId == targetUser.OrganizationId
  select c;

if (validCollections.Count != resource.Count)
{
  context.Fail();
  return;
}
```

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+        var targetCollections = await _collectionRepository.GetManyByManyIdsAsync(distinctTargetCollectionIds);
+
+        // A target collection does not exist, fail the requirement
+        if (targetCollections.Count != distinctTargetCollectionIds.Count)
+        {
+            context.Fail();
+            return;
+        }
+
+        // TODO: Add check for future organization Collection Management setting here and skip the next check if it's enabled
```
eliykat, line 71: Please link the follow-up ticket in all TODOs

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+    ///   The target collection(s) is manageable by the acting user. 
+    /// </remarks>
+    private async Task CanDeleteAsync(AuthorizationHandlerContext context,
+        CollectionUserOperationRequirement requirement, ICollection<CollectionUser> resource)
+    {
+        if (!_currentContext.UserId.HasValue)
+        {
+            context.Fail();
+            return;
+        }
```
eliykat, line 130: Will this be the case for all AuthorizationHandlers? What does it mean if the currentContext doesn't have a userId set?

In any case, this can go in the `HandleRequirementAsync` method to avoid repeating it in each private method.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+    /// specified collection (that has not be created in the database yet).
+    /// </summary>
+    /// <remarks>
+    /// Checks that the following are all true:
+    /// - The target collection is provided in the requirement
+    /// - All collection users are assigned to the target collection in the requirement
+    /// - The target users exist and belong to the same organization as the target collection
+    /// </remarks>
+    private async Task CanCreateForNewCollectionAsync(AuthorizationHandlerContext context,
+        CollectionUserOperationRequirement requirement, ICollection<CollectionUser> resource)
```
eliykat, line 170: Hmm I see the problem here. Can you push the code where these auth handlers are being used? It'll be helpful to see the complete picture, and also to see the code we're replacing.

I suspect that this Auth Handler includes too much general validation logic. Authorization is just about the user's permissions, not about making sure the general business logic of the request is valid (such as making sure the orgUsers match up to the organization). We might end up moving some of it back out to the controller or command level, which might solve some of our problems here.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+    /// Ensure the acting user can create the requested <see cref="CollectionUser"/> resource(s).
+    /// </summary>
+    /// <remarks>
+    /// Checks that the following are all true:
+    /// - The target collection(s) exists 
+    /// - The acting user is an owner/admin AND the collection management setting is enabled
+    ///   OR
+    ///   The target collection(s) is manageable by the acting user. 
+    /// - The target user(s) exists and belongs to the same organization as the target collection
+    /// </remarks>
```
eliykat, line 49: I think this is too much detail for an xmldoc comment, and it's likely to become out of date. It's not really required given that the method itself is commented as well.

eliykat (commented):
Great work! And thank you for implementing this pattern 👏

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
shane-melton: Fair point about it becoming out of date; it started out as pseudo code for myself when writing the method and I figured it would also work as documentation/summary for the method so I moved it there. I agree that its a bit verbose, so I can remove it.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+        if (targetCollections.Count != distinctTargetCollectionIds.Count)
+        {
+            context.Fail();
+            return;
+        }
+
+        // TODO: Add check for future organization Collection Management setting here and skip the next check if it's enabled
+
+        // List of collections the user is allowed to manage
+        var manageableCollections = (await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value)).Where(c => c.Manage).ToList();
```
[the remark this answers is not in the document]
shane-melton: I had originally thought Owners/Admins can only manage access to all collections if the organization has the appropriate "Owners and admin can manage all collections" setting enabled as [described in the last row of this table.](https://bitwarden.atlassian.net/wiki/spaces/PROD/pages/203292673/Collection+Management+Enhancements#Migration-Strategy)

However, reading that document/table again, I see some other possibly conflicting statements and am not sure what is always meant by "manage" in that document, so I'm working on getting clarification there.

As for the repository call, it does not consider owners/admins and only checks for `CollectionUser` and `CollectionGroup` relationships and the `AccessAll` flag that will eventually be deprecated. 

This is the reason I queried for _all_ of the `targetCollections` prior to this point; to verify that all the requested collections actually exist before checking those organizational permissions in the TODO comment above (which could then skip manageable collections query).

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
shane-melton: I believe it will be the most common case for AuthorizationHandlers, at least initially. All expected use cases for the handlers in this ticket will have a userId as they will be triggered by controller endpoints used by users via the Web client.

I would expect us to adapt the handler to account for any use cases where a `userId` is not expected (e.g. another type of authenticated subject is making a request that doesn't have a userId). But, for now, I think we're safe to assume a `userId` must be present to authorize successfully. 

And yep, we can move the check to the `HandleRequirementAsync` method to reduce repetition (and redundant tests).

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
shane-melton: This was just an initial strategy I wanted to demonstrate and get feedback on after our discussion with @vincentsalucci and his work. The only example for this particular operation requirement is in the unit tests at this time. I'll have a separate PR draft I can put up that demonstrates the `Create` operation shortly.

> I suspect that this Auth Handler includes too much general validation logic.

Totally fair point and I started to get that feeling as well. I felt the "orgUser/organization" check was authorization related because it ensured the acting user was only acting on users it had permission to assign (i.e. belong to the org). 

It sounds like you'd prefer to only check if the target organization users exist and belong to **any** organization the acting user has a role at. If the target user belongs to an unrelated org then the acting user obviously shouldn't have access to them. Then it would be the responsibility of the command/controller to ensure the users and collection match the same organization. 

It think that sounds reasonable and do admit it would likely simplify these authorization handlers quite a bit. I can work on making that change.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        // List of collections the user is performing the operation on
+        var targetCollections = await _collectionRepository.GetManyByManyIdsAsync(distinctTargetCollectionIds);
+
+        // A target collection does not exist, fail the requirement
+        if (targetCollections.Count != distinctTargetCollectionIds.Count)
+        {
+            context.Fail();
+            return;
+        }
```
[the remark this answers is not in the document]
shane-melton: It is somewhat redundant, but the `manageableCollections` query below excludes collections the user has no association with (i.e. via `CollectionUser` or `CollectionGroup`). So, the collection would come back as missing and fail the requirement. 

However, because there are organization settings that can allow users to modify access without a direct association with a collection; I needed to ensure they exist and belong to an organization that has those settings enabled or they're an Owner/Admin. 

And then, if no org settings are enabled and they're not an Owner/Admin, we need to continue and check for the `CanManage` setting for each targeted collection. 

We could possibly combine these query/repository calls into a single query that accepts multiple collection Ids _and_ a userId to also grab the permissions for that particular user. It would be an outer join though, that way if the permission is missing, the collection would still appear as existing for the organizational level checks. I opted to avoid that path for now to avoid introducing those new, more complicated query concepts at this time, but its definitely an area has potential for optimization.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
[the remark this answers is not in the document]
eliykat: The 2 collection management settings are being implemented separately:

* the current epic, AC-1373, adds the "limit collection creation and deleting to owners and admins" settings. It also adds the collection-level "Can Manage" permission. Together, they effectively replace the Manager role. But we're not changing owner/admin access.
* in the _next_ epic, AC-1114, we're adding the "owners and admins can manage all collections" setting which allows you to restrict collection access. But for now, owners/admins retain their current behaviour (equivalent to that setting being ON)

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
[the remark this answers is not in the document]
eliykat: The path here might be a bit clearer after our latest discussion about the collection management settings. But from the discussion above it sounds like the flow needs to be something like:

1. get the `CurrentContentOrganization` (misspelling is intentional, should be `CurrentContextOrganization`) and check if the user is owner/admin for each relevant organization
    * you'll need to bring across this method from my PR: https://github.com/bitwarden/server/pull/2845/files#diff-a0d048d2dd3fd53b7dc8f8349cd348baf79c0ce075af29faff0a7abbc3683468R567-R570
    * this contains information about the user's permissions for that org, based on claims in the JWT - saves us a database call. CurrentContext will still construct that for us and make it available, but the actual business logic will be in handlers
3. if so, no further checks required for those orgs, we know they have permissions
4. do the `manageableCollections` db query + logic for the remaining collections, if any

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
eliykat: It's a bit hard to say where to draw the line. Here's how I'm thinking of it (not to be too prescriptive, just to try to be clear):

**Can create collection-user and collection-group associations**
* validation logic in the command: is this a valid request
  * do all these collections exist?
  * do all orgUsers/groups exist?
  * do all orgUsers/groups match the orgId of the collection they're being associated to?
  * maybe we do this check first so that by the time it gets to the handler we can assume it's a valid object
* authorization handler: does the user have permission
  * is the user an admin or owner for that org?
  * or, does the user have Can Manage permissions for that collection?

**Can delete collection-user and collection-group associations**
* validation logic in the command: is this a valid request
  * much less required here, maybe nothing? We can just fire delete requests off to the db
* authorization handler: does the user have permission
  * is the user an admin or owner for that org?
  * or, does the user have Can Manage permissions for that collection?

(I may not have followed this strictly in my own PR, just thinking it through in more detail now)

An advantage of this is that then the `CanCreateForNewCollectionAsync` disappears altogether, it's entirely validation logic. I think if you're creating a collection then it's assumed you can create the initial associations with that collection as part of that same operation.

eliykat: Some null checks may be required as part of the authorization logic (like if the org doesn't exist, we can't check permissions on it). But it's not the main focus.

shane-melton: That all sounds reasonable to and makes sense to get validation logic out of the authorization handler.

One downside I see to splitting the validation and authorization logic is that we'll have to query the database in each context. First in the validation logic to see if the entities exist and are configured correctly (everything matches up). Then again when we check if the user has the Can Manage permission for each collection.

Not that this extra database call would be enough reason to avoid the clear separation of concerns, it just seems to be an unfortunate side effect.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
shane-melton: Resolving as this is no longer relevant after removing this validation logic from the handler.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
shane-melton: Comments removed!

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
shane-melton: These changes turned out nicely I think, the handler is more straightforward now and only handles authorization concerns.

Another benefit of the simplification is that the authorization logic for both `CollectionUser` and `CollectionGroup` resources became identical, so I was able to create a shared `CollectionAccessAuthorizationHandlerBase` to use for both resource handlers internally. Each still have their own handlers with different operation requirements for `Create` and `Delete` to follow the authorization/operation pattern and allow for modification for each operation should it be required in the future. For now though, they all call the same underlying `CanManageCollectionAccess` method in the base class to keep things less repetitive and easier to test.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBase.cs
```diff
@@ -0,0 +1,99 @@
...
+            (await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value))
+            .Where(c => c.Manage).ToList();
+
+        foreach (var org in restrictedOrganizations)
+        {
+            // User must have explicit "Manage" permission on the target collections for this organization
+            foreach (var targetCollection in targetCollections.Where(tc => tc.OrganizationId == org.Id))
+            {
+                // Target collection is not in the list of manageable collections, fail
+                if (!manageableCollections.Exists(c => c.Id == targetCollection.Id))
```
eliykat, line 89: There are 2 nested loops here (plus the outer loop), can it be rewritten to avoid nesting? e.g. by un-nesting the loops or by using hashsets for faster lookup

src/Core/OrganizationFeatures/OrganizationServiceCollectionExtensions.cs
```diff
@@ -47,6 +49,9 @@ public static void AddOrganizationServices(this IServiceCollection services, IGl
         services.AddOrganizationAuthCommands();
         services.AddOrganizationUserCommandsQueries();
         services.AddBaseOrganizationSubscriptionCommandsQueries();
+
+        services.AddScoped<IAuthorizationHandler, CollectionUserAuthorizationHandler>();
+        services.AddScoped<IAuthorizationHandler, CollectionGroupAuthorizationHandler>();
```
eliykat, line 54: This class is used by many different projects (e.g. scim, billing, identity, etc) but these dependencies are only needed for Api. I suggest registering them in a serviceCollectionExtension class that only Api uses. (maybe a dedicated `AuthorizationHandlerServiceCollectionExtension`?)

That suggests to me that these should also be moved to somewhere in the Api project, rather than in Core.

test/Core.Test/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBaseTests.cs
```diff
@@ -0,0 +1,228 @@
...
+        ICollection<CollectionDetails> collectionDetails,
+        ICollection<CurrentContentOrganization> organizations)
+    {
+        var actingUserId = Guid.NewGuid();
+
+        // Simulate not having manage collection permission
+        foreach (var collectionDetail in collectionDetails)
+        {
+            collectionDetail.Manage = false;
+        }
```
eliykat, line 153: Suggestion: do this for just 1 `collectionDetail`, being the minimum required to get a `Context.Fail()` result.

test/Core.Test/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBaseTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            collectionUsers
+        );
+        var expectedCollectionIds = collectionUsers.Select(cu => cu.CollectionId);
+
+        sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
+
+        await sutProvider.Sut.HandleAsync(context);
+
+        await sutProvider.GetDependency<ICollectionRepository>()
+            .Received(2) // Called twice, once for each operation requirement
```
eliykat, line 199: Hm, it only just clicked that all our db calls etc. are going to be repeated for each `OperationRequirement`. And often if we're modifying associations, we're both adding and deleting them, which means we'll often do this work twice when it's not really required. _And_ we'll do it for users and groups. (I assume - you might have some other insight here if you've been looking at the controller logic.)

Do you have any ideas on how to mitigate this? Maybe we do need to make the handlers and/or the operations a little less granular? (other suggestions welcome)

test/Core.Test/AutoFixture/CollectionFixture.cs
eliykat, line 1: This is some autofixture magic that I wouldn't have thought of. I can see how it helps you create matching organization resources though. Nice work!

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionUserAuthorizationHandler.cs
eliykat: I agree this is looking really good! I'll leave a fresh round of feedback below.

eliykat (changes requested):
This is looking really good with a much tighter focus on just the auth logic. 🎉

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBase.cs
```diff
@@ -0,0 +1,99 @@
...
+        var targetCollections = await _collectionRepository.GetManyByManyIdsAsync(distinctTargetCollectionIds);
+
+        // A target collection does not exist, fail the requirement
+        if (targetCollections.Count != distinctTargetCollectionIds.Count)
+        {
+            context.Fail();
+            return;
+        }
+
+        var userOgs = await _currentContext.OrganizationMembershipAsync(_organizationUserRepository, _currentContext.UserId.Value);
```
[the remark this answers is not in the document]
shane-melton: I used `OrganizationMembershipAsync()` as it had a check to ensure the `Organizations` property (of type `List<CurrentContentOrganization>`) in `CurrentContext` is not null. 

- If its not null, meaning it was likely populated from the JWT, it simply returns the existing list of `CurrentContentOrganization`s without a database query. 
- Only if the list is null, does it attempt to query the database to construct the list directly.

This way we can be sure the handler always has a copy of the user's organizations. Admittedly, it seems unlikely for the list to be null if its coming from the JWT, but I assume the logic to build the list from the database in case its missing from the JWT exists for certain scenarios? 

If you think we should instead fail the auth request if the list of `Organizations` is missing from `CurrentContext` I can switch to using the `Organizations` property directly and fail if its null.

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBase.cs
```diff
@@ -0,0 +1,99 @@
...
+                return;
+            }
+
+            // Owner/Admins or users with EditAnyCollection permission can always manage collection access
+            if (
+                org.Permissions.EditAnyCollection ||
+                org.Type is OrganizationUserType.Admin or OrganizationUserType.Owner)
+            {
+                continue;
+            }
```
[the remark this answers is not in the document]
shane-melton: Ah, good catch, thanks!

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBase.cs
shane-melton: Rewritten to remove the nested loops and instead joins/queries the lists together to find any collections that don't meet the requirements; then if any exist, fail the request.

src/Core/OrganizationFeatures/OrganizationServiceCollectionExtensions.cs
shane-melton: I agree that it shouldn't be added in that shared organization service collection extension method, as its called in places this handler is not needed. 

However, I still feel that they should belong in the `Core` project with its own `AddAuthorizationHandlers()` service collection extension. This is also stays consistent with what SecretsManager did with their authorization handlers and registration. Authorization, as a general concept, is something that spans across many projects in some form or another, so to me, it makes sense to keep these together and at a higher level than the API project. 

If you feel strongly that it should be moved into the API project let me know and we can discuss!

src/Core/OrganizationFeatures/AuthorizationHandlers/CollectionAccessAuthorizationHandlerBase.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// </summary>
+    protected async Task CanManageCollectionAccessAsync(AuthorizationHandlerContext context, IAuthorizationRequirement requirement, IEnumerable<Guid> collectionIds)
+    {
+        if (!_currentContext.UserId.HasValue)
+        {
+            context.Fail();
+            return;
+        }
+
+        var distinctTargetCollectionIds = collectionIds.Distinct().ToList();
```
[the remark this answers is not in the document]
eliykat: Thanks for confirming!

src/Api/Vault/AuthorizationHandlers/CollectionAccessAuthorizationHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+        var manageableCollectionIds =
+            (await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value))
+            .Where(c => c.Manage)
+            .Select(c => c.Id);
+
+        // Target collections that require explicit "Manage" permission, but the user does not have it
+        var collectionWithoutPermission = from tc in targetCollections
+                                          join o in restrictedOrganizations on tc.OrganizationId equals o.Id
+                                          where !manageableCollectionIds.Contains(tc.Id)
+                                          select tc;
```
eliykat, line 101: I don't know if changing to LINQ gets us any performance improvements here, without any change in the underlying logic. Is join more efficient than a loop?

I went down a rabbit hole here, I thought I should work it through myself if I was going to ask for more changes, and here's what I came up with: https://github.com/bitwarden/server/compare/ac/ac-1174/collection-user-auth-handler...ac/ac-1174/collection-user-auth-handler-alt

However I'm not convinced that it's necessarily better, it might just be different. What do you think? Your choice.

If you want to stick with the LINQ approach then my only recommendation is to use a hashset for constant time lookup on `manageableCollectionIds`.

```cs
        var manageableCollectionIds =
            (await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value))
            .Where(c => c.Manage)
            .Select(c => c.Id)
            .ToHashSet();
```

test/Core.Test/Vault/AutoFixture/CollectionFixture.cs
```diff
@@ -0,0 +1,52 @@
...
+
+public class CollectionCustomization : ICustomization
+{
+    private const int _collectionIdSeed = 1;
+    private const int _userIdSeed = 2;
+    private const int _groupIdSeed = 3;
+
+    public void Customize(IFixture fixture)
+    {
+        var orgIds = new[] { Guid.NewGuid(), Guid.NewGuid() };
```
eliykat, line 18: When playing with this, I found that autofixture always generates 3 organizations using `BitAutoData`, which means it'll wrap around to the start of this array and the you'll end up with the first and last org sharing the same Guid. I solved this by adding another `Guid.NewGuid()` element here, but maybe there's a better way.

eliykat (changes requested):

src/Api/Vault/AuthorizationHandlers/CollectionAccessOperation.cs
```diff
@@ -0,0 +1,15 @@
...
+public class CollectionAccessOperationRequirement : OperationAuthorizationRequirement { }
+
+public static class CollectionAccessOperation
+{
+    /// <summary>
+    /// The operation that represents creating or removing collection access.
+    /// Combined together to allow for a single requirement to be used for both operations as they
+    /// currently share the same underlying authorization logic.
+    /// </summary>
+    public static readonly CollectionAccessOperationRequirement CreateDelete = new() { Name = nameof(CreateDelete) };
```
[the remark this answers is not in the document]
shane-melton: Good point, updated!

src/Api/Vault/AuthorizationHandlers/CollectionAccessAuthorizationHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+            var org = userOrgs.FirstOrDefault(o => orgId == o.Id);
+
+            // Acting user is not a member of the target organization, fail
+            if (org == null)
+            {
+                context.Fail();
+                return;
+            }
+
+            // Owner/Admins or users with EditAnyCollection permission can always manage collection access
```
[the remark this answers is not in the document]
shane-melton: Fixed

src/Api/Vault/AuthorizationHandlers/CollectionAccessAuthorizationHandler.cs
shane-melton: Thanks for digging into this; however, I went ahead and simplified this logic by requiring that all of the resources belong to the same organization. I did still add the `ToHashSet()` call to improve the lookup time on `managableCollectionIds`, but the other nested loops are no longer a problem this way.

Supporting multiple orgs in a single request has been something I kept going back and forth on. It adds additional complexity and is for something we currently have no use cases for. Users never modify resources belonging to multiple organizations at once.

Additionally, as I have been working on the the command/SQL to actually create these associations, trying to support multiple organizations blows up the complexity of the SQL and validation logic; so it seemed best to just focus on our actual use case and restrict requests to have the same organization for all collections.

Let me know what you think!

test/Core.Test/Vault/AutoFixture/CollectionFixture.cs
shane-melton: With the change to only support single orgs this became no longer necessary and I've switched it back to a single `orgId`. (This was another complexity that was pulling me away from trying to support multiple orgs in the same request)

src/Core/Entities/ICollectionAccess.cs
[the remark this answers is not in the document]
shane-melton: Despite the name and purpose being very similar they do represent the collection <-> access relationship from different perspectives at the moment. 

`CollectionAccessSelection` uses `Id` to represent either an `OrganizationUser` or a `Group` selection made from the UI / request model. It has never needed a `CollectionId` property as all requests / model that used it had the collection Id available in a route parameter or parent model property. (see the various collection PUT and POST endpoints) 

The `ICollectionAccess` interface requires `CollectionId` as its used to provide a common ancestor for `CollectionUser` and `CollectionGroup` in cases where we don't care about the grantee of the access (i.e. this authorization handler for now).

I would really like to consolidate on a better naming convention / pattern for these concepts as a whole, but is likely out of scope of this ticket and will require its own research to develop a solid solution. Fortunately, for now, this `ICollectionAccess` interface is internal and not adding any new patterns to public APIs, so we should not be creating anymore headache for ourselves in the future.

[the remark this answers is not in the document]
eliykat: I agree, thanks for considering it!

src/Api/Vault/AuthorizationHandlers/CollectionAccessAuthorizationHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        var targetOrganizationId = targetCollections.First().OrganizationId;
+
+        // Ensure all target collections belong to the same organization
+        if (targetCollections.Any(tc => tc.OrganizationId != targetOrganizationId))
+        {
+            context.Fail();
+            return;
+        }
```
eliykat, line 64: I recommend throwing an error here, it's not that the user doesn't have access, we've just decided not to support this operation. So throwing an error with some feedback seems more useful.

src/Api/Vault/AuthorizationHandlers/CollectionAccessAuthorizationHandler.cs
eliykat: I think your reasoning is sound, especially given that we don't actually need that extra complexity. (and we're unlikely to need it in the future)

src/Api/Vault/AuthorizationHandlers/CollectionAccessAuthorizationHandler.cs
shane-melton: Good point, done!

eliykat (approved):
Great work on this! I assume there'll be a separate PR to actually call it.

vincentsalucci (approved):

PR #3199, opened 2023-08-15
[AC-244] Consider a user's email as verified when they accept an organization invitation via the email link

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
If a user is invited to an organization and accepts the email invite, the member record should be considered automatically verified.

## Code changes

This PR is tied to [this clients PR](https://github.com/bitwarden/clients/pull/6094).

- **src/Core/Services/IOrganizationService.cs:** Added a summary for one of the signatures of the method `AcceptUserAsync`
- **src/Core/Services/Implementations/OrganizationService.cs:** Mark the user's email as verified after the process of accepting the user into the organization
- **test/Core.Test/AutoFixture/OrganizationFixtures.cs:** Created `EphemeralDataProtectionAutoDataAttribute` to use an instance of `EphemeralDataProtectionProvider` for testing
- **test/Core.Test/Services/OrganizationServiceTests.cs:** Added a unit test to verify the expected behaviour of `OrganizationService.AcceptUserAsync` for a valid input

## Before you submit


test/Core.Test/Services/OrganizationServiceTests.cs
```diff
@@ -1831,4 +1832,31 @@ public void ValidateSecretsManagerPlan_ThrowsException_WhenMoreSeatsThanPassword
...
+        user.Email = organizationUser.Email;
+
+        var dataProtector = sutProvider.GetDependency<IDataProtectionProvider>()
+            .CreateProtector("OrganizationServiceDataProtector");
+        // Token matching the format used in OrganizationService.InviteUserAsync
+        var token = dataProtector.Protect(
+            $"OrganizationUserInvite {organizationUser.Id} {organizationUser.Email} {CoreHelpers.ToEpocMilliseconds(DateTime.UtcNow)}");
+        var userService = Substitute.For<IUserService>();
+
+        sutProvider.GetDependency<IGlobalSettings>().OrganizationInviteExpirationHours.Returns(24);
```
eliykat, line 1850: This seems useful, is there any test helper class we can put it in for future use?

src/Core/Services/Implementations/OrganizationService.cs
eliykat, line 1: lgtm, but can you please make sure you've checked all the flows that this is called from, and make sure that this is applicable to all of them? If a user is moved into the accepted status in any way other than via the email link, their email should not be confirmed.

One flow that jumped out to me is that one of the public `AcceptUserAsync` methods is called by `OrganizationUsersController.PutResetPasswordEnrolment`. But that sounds like something you do from the web vault, not the email link. (I could be wrong though - maybe you only end up on that flow by clicking the email link to begin with?)

test/Core.Test/Services/OrganizationServiceTests.cs
r-tome: I didn't find an appropriate place. In the future if we reuse this elsewhere I'll add it to `Core.Utilities`.

test/Core.Test/Services/OrganizationServiceTests.cs
```diff
@@ -1831,4 +1832,38 @@ public void ValidateSecretsManagerPlan_ThrowsException_WhenMoreSeatsThanPassword
...
+    {
+        var token = SetupAcceptUserAsyncTest(sutProvider, user, organizationUser);
+        var userService = Substitute.For<IUserService>();
+
+        await sutProvider.Sut.AcceptUserAsync(organizationUser.Id, user, token, userService);
+
+        await sutProvider.GetDependency<IOrganizationUserRepository>().Received(1).ReplaceAsync(
+            Arg.Is<OrganizationUser>(ou => ou.Id == organizationUser.Id && ou.Status == OrganizationUserStatusType.Accepted));
+        await sutProvider.GetDependency<IUserRepository>().Received(1).ReplaceAsync(
+            Arg.Is<User>(u => u.Id == user.Id && u.Email == user.Email && user.EmailVerified == true));
```
eliykat, line 1849: I think we need to set `user.EmailVerified = false` at the start of the test, to make sure it's in the state we expect.

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -1093,7 +1093,12 @@ private async Task SendInviteAsync(OrganizationUser orgUser, Organization organi
             throw new BadRequestException("User email does not match invite.");
         }
 
-        return await AcceptUserAsync(orgUser, user, userService);
+        var organizationUser = await AcceptUserAsync(orgUser, user, userService);
+
+        user.EmailVerified = true;
+        await _userRepository.ReplaceAsync(user);
```
eliykat, line 1099: We should only do this if their email isn't already verified. Potentially save ourselves a db call.

src/Core/Services/IOrganizationService.cs
```diff
@@ -39,6 +39,16 @@ public interface IOrganizationService
...
+    /// <summary>
+    /// Accepts a user into an organization based on the provided invitation token and user information.
+    /// Validates the invitation token, checks for existing memberships, and ensures email correspondence before acceptance.
+    /// Marks the user's email as verified upon successful acceptance.
+    /// </summary>
+    /// <param name="organizationUserId">The unique identifier of the organization user.</param>
+    /// <param name="user">Information about the user accepting the invitation.</param>
+    /// <param name="token">Token associated with the invitation.</param>
+    /// <param name="userService">Service for user-related operations.</param>
+    /// <returns>The accepted OrganizationUser.</returns>
```
eliykat, line 51: This may be a nit, or just a style difference, but I think this is too much info, and unnecessary explanation on the parameters. I suggest shorter & simpler, e.g.

```suggestion
    /// <summary>
    /// Moves an OrganizationUser into the Accepted status and marks their email as verified.
    /// This method is used where the user has clicked the invitation link sent by email.
    /// </summary>
    /// <param name="token">The token embedded in the email invitation link</param>
    /// <returns>The accepted OrganizationUser.</returns>
```

One thing I don't know is if we need to describe all the other params as well. They still show up in the method signature so I'm inclined to only comment on those that are unclear, but feel free to disagree on that.

eliykat (changes requested):
Looking good!

src/Core/Services/IOrganizationService.cs
r-tome: I tested it and (at least on Rider) all the parameters showed up but only `token` had a description.

eliykat (dismissed):
Great work!

eliykat (approved):

PR #3118, opened 2023-07-18
PM-2427-Defect-Cannot-delete-account-if-User-has-auth-request-on-table-FK-constraint

ike-kottlowski:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

When a user uses a device to log in, they were not able to delete their account immediately after. This update to the stored proc and UserRepository.cs allows users to delete their account without having to wait for the AuthRequest table to purge old request.

## Code changes

* **UserRepository.cs:** To keep EF changes up-to-date with Stored Proc
* **User_DeleteById.sql:** Added Delete from AuthRequest table.
* **2023-07-17_00_DeleteUserSproc.sql:** Migration file to account for changes to SQL Store Procedure

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3202, opened 2023-08-15
[PM-2427] Added back alias that was accidentally introduced

trmartin4:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

In https://github.com/bitwarden/server/pull/3118/, we accidentally introduced a change that was meant to be part of PM-1247, which is still in review.

This commit adds back the alias with an additional note referencing the issue that will remove it.  We cannot remove it yet as clients need to be fixed.

## Code changes

* **SsoController.cs:** Added back the `account` alias and added a note referencing the Jira ticket currently in progress to fix it.

## Before you submit

### DIFF
--- a/src/Identity/Controllers/SsoController.cs
+++ b/src/Identity/Controllers/SsoController.cs
@@ -14,6 +14,8 @@ using Microsoft.AspNetCore.Mvc;
 
 namespace Bit.Identity.Controllers;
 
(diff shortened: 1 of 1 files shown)


JaredSnider-Bitwarden (approved):
Ty for fixing this!

PR #3209, opened 2023-08-17
[AC-1601] Added RequireSso policy as a dependency of TDE

trmartin4:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

We are making the RequireSSO policy a prerequisite for enabling TDE.  This is because allowing users to use Login with Device puts users in a bad state that was unanticipated during implementation.

This change in policy application has been discussed and approved by Product as well.

## Code changes
* **SsoConfigService.cs:** Added code to set RequireSso policy when TDE is enabled.
* **PolicyService.cs:** Added code to prevent disabling RequireSso policy when TDE is enabled.
* **PolicyServiceTests.cs:** Added test to ensure that the RequireSso not enabled when TDE is enabled throws an error.

## Before you submit

### DIFF
--- a/src/Core/Auth/Services/Implementations/SsoConfigService.cs
+++ b/src/Core/Auth/Services/Implementations/SsoConfigService.cs
@@ -84,7 +84,7 @@ public class SsoConfigService : ISsoConfigService
                               new Policy { OrganizationId = config.OrganizationId, Type = PolicyType.RequireSso, };
 
             ssoRequiredPolicy.Enabled = true;
-            await _policyService.SaveAsync(resetPolicy, _userService, _organizationService, null);
+            await _policyService.SaveAsync(ssoRequiredPolicy, _userService, _organizationService, null);
         }
 
         await LogEventsAsync(config, oldConfig);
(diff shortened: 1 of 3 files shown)


src/Core/Auth/Services/Implementations/SsoConfigService.cs
```diff
@@ -78,8 +78,13 @@ public async Task SaveAsync(SsoConfig config, Organization organization)
...
             resetPolicy.Enabled = true;
             resetPolicy.SetDataModel(new ResetPasswordDataModel { AutoEnrollEnabled = true });
-
             await _policyService.SaveAsync(resetPolicy, _userService, _organizationService, null);
+
+            var ssoRequiredPolicy = await _policyRepository.GetByOrganizationIdTypeAsync(config.OrganizationId, PolicyType.RequireSso) ??
+                              new Policy { OrganizationId = config.OrganizationId, Type = PolicyType.RequireSso, };
+
+            ssoRequiredPolicy.Enabled = true;
+
```
shane-melton, line 87: issue: We need to save the `ssoRequiredPolicy` with the `_policyService`. Something like:

```c#
await _policyService.SaveAsync(ssoRequiredPolicy, _userService, _organizationService, null);
```

shane-melton (changes requested):
Just need to fix one issue and then I think we're good to go!

src/Core/Auth/Services/Implementations/SsoConfigService.cs
trmartin4: Fixed! I accidentally deleted that with an errant keystroke before committing, after testing locally. Thank you for catching.

shane-melton (approved):
Looks good!

PR #3210, opened 2023-08-18
[AC-1568] Refactor UpdateSecretsManagerSubscriptionCommandTests to use new ctor

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

In https://github.com/bitwarden/server/pull/3121 we added a new, simpler constructor on the SecretsManagerSubscriptionUpdate object, which takes an Organization object and uses it to populate the initial values.

Due to time constraints, that change was backwards-compatible with the old constructor interface. This is part 1 of an effort to go back and update all calling locations to use the new constructor, and then remove the old one.

This PR only updates test code, which was most of the uses of the old ctor. I split it so we can merge the bulk of the changes without needing QA approval. The next PR will make any necessary changes to production code.

While updating the tests I also made some changes to improve their structure and simplify the arrange phase, including:
* autofixture customizations
* ordering tests to match the order of the sut
* using regions

## Code changes

## Before you submit

### DIFF
--- a/test/Core.Test/OrganizationFeatures/OrganizationSubscriptionUpdate/UpdateSecretsManagerSubscriptionCommandTests.cs
+++ b/test/Core.Test/OrganizationFeatures/OrganizationSubscriptionUpdate/UpdateSecretsManagerSubscriptionCommandTests.cs
@@ -280,7 +280,6 @@ public class UpdateSecretsManagerSubscriptionCommandTests
     [Theory]
     [BitAutoData]
     public async Task UpdateSubscriptionAsync_SmSeatAutoscaling_Subtracting_ThrowsBadRequestException(
-        PlanType planType,
         Organization organization,
         SutProvider<UpdateSecretsManagerSubscriptionCommand> sutProvider)
     {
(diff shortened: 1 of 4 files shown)


test/Core.Test/OrganizationFeatures/OrganizationSubscriptionUpdate/UpdateSecretsManagerSubscriptionCommandTests.cs
```diff
@@ -17,711 +18,617 @@
 namespace Bit.Core.Test.OrganizationFeatures.OrganizationSubscriptionUpdate;
 
 [SutProviderCustomize]
+[SecretsManagerOrganizationCustomize]
 public class UpdateSecretsManagerSubscriptionCommandTests
 {
+    #region General
```
r-tome, line 24: We had a vote and `#regions` lost https://contributing.bitwarden.com/contributing/code-style/#code-regions

r-tome (commented):
I love the tests, nice work!
Regarding the regions, should we split this file?

test/Core.Test/OrganizationFeatures/OrganizationSubscriptionUpdate/UpdateSecretsManagerSubscriptionCommandTests.cs
eliykat: I forgot about that. I removed the regions, they're probably not necessary, and I'm not sure I want to split the test code into separate files.

r-tome (approved):
LGTM!

PR #3219, opened 2023-08-21
[PM-3563] Prevent org name from injecting HTML into FD notes

kspearrin:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Prevent org name from injecting HTML into Freshdesk notes.

## Code changes

* **src/Billing/Controllers/FreshdeskController.cs:** Replace `<` and `>` from org names.

## Before you submit

### DIFF
--- a/src/Billing/Controllers/FreshdeskController.cs
+++ b/src/Billing/Controllers/FreshdeskController.cs
@@ -77,7 +77,9 @@ public class FreshdeskController : Controller
 
                 foreach (var org in orgs)
                 {
-                    var orgNote = $"{org.Name} ({org.Seats.GetValueOrDefault()}): " +
+                    // Prevent org names from injecting any additional HTML
+                    var orgName = org.Name.Replace("<", string.Empty).Replace(">", string.Empty);
+                    var orgNote = $"{orgName} ({org.Seats.GetValueOrDefault()}): " +
                         $"{_globalSettings.BaseServiceUri.Admin}/organizations/edit/{org.Id}";
                     note += $"<li>Org, {orgNote}</li>";
                     if (!customFields.Any(kvp => kvp.Key == _billingSettings.FreshDesk.OrgFieldName))
--- a/src/Billing/Controllers/FreshdeskController.cs
+++ b/src/Billing/Controllers/FreshdeskController.cs
@@ -1,6 +1,7 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using System.Reflection;
 using System.Text;
+using System.Web;
 using Bit.Billing.Models;
 using Bit.Core.Repositories;
 using Bit.Core.Settings;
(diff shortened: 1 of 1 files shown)


cyprain-okeke (commented):
Looks Good

src/Billing/Controllers/FreshdeskController.cs
```diff
@@ -77,7 +77,9 @@ public class FreshdeskController : Controller
 
                 foreach (var org in orgs)
                 {
-                    var orgNote = $"{org.Name} ({org.Seats.GetValueOrDefault()}): " +
+                    // Prevent org names from injecting any additional HTML
+                    var orgName = org.Name.Replace("<", string.Empty).Replace(">", string.Empty);
+                    var orgNote = $"{orgName} ({org.Seats.GetValueOrDefault()}): " +
```
cyprain-okeke, line 82: We are considering Password Manager Seat only? because org.Seats.GetValueOrDefault() is for password maanger seat only while org.SmSeats.GetValueOrDefault() is for secrets Manager Seat

src/Billing/Controllers/FreshdeskController.cs
```diff
@@ -77,7 +77,9 @@ public class FreshdeskController : Controller
 
                 foreach (var org in orgs)
                 {
-                    var orgNote = $"{org.Name} ({org.Seats.GetValueOrDefault()}): " +
+                    // Prevent org names from injecting any additional HTML
+                    var orgName = org.Name.Replace("<", string.Empty).Replace(">", string.Empty);
```
cturnbull-bitwarden, line 81: suggestion:

An alternative approach we could consider is utilizing `HtmlUtility.Encode(org.Name)` from the `System.Web` namespace. Using that utility method should have the added benefit of taking care of addressing other HTML-reserved characters like `"`, `'`, and `&`. We could consider using `HtmlUtility.Encode` as more resilient since it would also handle any future additions or changes to reserved characters.
That said, the output of `HtmlUtility.Encode` would be different from this approach since it replaces characters with their encoded equivalents instead of just stripping them out. However, this could ensure that we aren't inadvertently discarding any non-malicious data!
Feel free to resolve this comment if you think the current approach suffices!

kspearrin: I switched to HtmlEncode. This is a little less desirable from a display perspective since the `'` character will get encoded and I suspect it is somewhat common in org names, but overall I think is a safer solution.

src/Billing/Controllers/FreshdeskController.cs
kspearrin: I don't see org name being included anywhere else.

cturnbull-bitwarden: IMO updating the content of the notes field is out of scope here. The changes address the stated issue of preventing an HTML injection attack. I'll resolve this and create a spike to investigate if we should add `SmSeats` to the FD note.

cturnbull-bitwarden (approved):
Looks good!

cyprain-okeke (approved):
LGTM

PR #2471, opened 2022-12-05
[PM-2444] Add Pipeline for Testing All Database Variants in CI

justindbaur:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [x] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Introduce automated testing of various database types, so that we can trust database changes across all supported versions.

## Code changes

* **.github/workflows/infrastructure-tests.yml:** Brand new file that runs an automated test by first composes all databases defined in our docker compose and then migrates them all to the latest migration. Then it runs the tests will all defined databases as it's config. It's built in a way such that many more databases/versions can be added later and validated as being able to run bitwarden. 
* There were several updates to the existing tests to ensure that they behaved the same across all versions and ensure repeatability when ran locally.
* **test/Infrastructure.IntegrationTest/DatabaseDataAttribute.cs:** Change the paradigm of this data to make it possible to have multiple databases per database type if it is configured for it.
* **util/\*\*/Factories.cs:** Made it possible to pass in configuration through command line args for easier use in a CI scenario.

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3229, opened 2023-08-24
[AC-1174] Bulk Collection Management

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
Add support for bulk adding access to multiple collections at once.

Related client changes: https://github.com/bitwarden/clients/pull/6133

## Code changes

- `src/Api/Controllers/CollectionsController.cs`
  -  Add the new endpoint and make use of the new `CollectionAuthorizationHandler` introduced in #3194

- `src/Core/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommand.cs`
  - Add a new `BulkAddCollectionAccessCommand`
  - Validates that all collections, users, and groups exist and belong to the same organization
  - Creates `Collection_Updated` event logs for every collection modified
  - Command is registered in the `OrganizationServiceCollectionExtensions.cs`

- `src/Infrastructure.Dapper/Repositories/CollectionRepository.cs`
  - Add new repository method `CreateOrUpdateAccessForManyAsync()`
  - **Note: This new method only adds or updates collection access, it does not remove any existing access relationships.**

- `src/Infrastructure.EntityFramework/Repositories/DatabaseContextExtensions.cs`
  - Add a new extension method to bump user account revision dates by _multiple_ collection ids at once.
  - A similar sproc was also created `User_BumpAccountRevisionDateByCollectionIds.sql`

## Before you submit


test/Core.Test/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommandTests.cs
```diff
@@ -0,0 +1,246 @@
...
+    private static ICollection<CollectionAccessSelection> ToAccessSelection(IEnumerable<CollectionGroup> collectionGroups)
+    {
+        return collectionGroups.Select(cg => new CollectionAccessSelection
+        {
+            Id = cg.GroupId,
+            Manage = cg.Manage,
+            HidePasswords = cg.HidePasswords,
+            ReadOnly = cg.ReadOnly
+        }).ToList();
+    }
```
eliykat, line 245: It looks like some `CollectionRepository` methods do a transformation like this (e.g. see `GetManyByOrganizationIdWithAccessAsync`), maybe we could move this into production code? e.g. as a method on `CollectionUser`/`Group` or as a constructor overload on `CollectionAccessSelection`.

eliykat (changes requested):
Reviewed everything except for the db changes - I'll come back to those tomorrow with a fresh set of eyes. However they're complex enough that I'd definitely want a second reviewer on them.

Great work overall 👏

test/Core.Test/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommandTests.cs
shane-melton: I like this idea; however, as I started doing so, I re-discovered the `CollectionAccessSelection.Id` property flips what it is identifying based on the perspective of the "owning" entity of the list of `CollectionAccessSelection`.

In the above code, the perspective is from a single collection that owns multiple `CollectionAccessSelection` for various Group/Users, so the `Id` property refers to a `OrganizationUserId` or `GroupId`.

However, this relationship flips when a Group or OrgUser owns multiple `CollectionAccessSelection` and the `Id` property now refers to a `ColllectionId`. See below for an example in the `GroupRepository.cs`.  https://github.com/bitwarden/server/blob/2aaef3cf64119440be897e937c0f68a40f27328a/src/Infrastructure.Dapper/Repositories/GroupRepository.cs#L69-L75

So, because we use the same entity for both perspectives, it makes it tricky to create a method or constructor overload. We would need to take another argument or create additional overloads to specify if the `Id` property should be a `CollectionId` OR a `GroupId`/`OrganizationUserId`. I don't really like either of those options at this point, and think this further highlights that we should look into renaming/refactoring these `CollectionAccessSelection` objects at some point (seems out of scope for this ticket, but is probably worthwhile for Flexible Collections).

Let me know what you think!

test/Core.Test/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommandTests.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        var userAccessSelections = ToAccessSelection(collectionUsers);
+        var groupAccessSelections = ToAccessSelection(collectionGroups);
+        await sutProvider.Sut.AddAccessAsync(org.Id, collections,
+            userAccessSelections,
+            groupAccessSelections
+        );
+
+        await sutProvider.GetDependency<IOrganizationUserRepository>().ReceivedWithAnyArgs().GetManyAsync(default);
+        await sutProvider.GetDependency<IGroupRepository>().ReceivedWithAnyArgs().GetManyByManyIds(default);
```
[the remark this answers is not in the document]
shane-melton: These don't use an `orgId` here, but I did modify them to check for the correct `orgUserIds` and `groupIds`.

test/Api.Test/Controllers/CollectionsControllerTests.cs
```diff
@@ -250,4 +253,123 @@ await sutProvider.GetDependency<IDeleteCollectionCommand>()
...
+    public async Task PostBulkCollectionAccess_Success(Guid orgId, User actingUser, ICollection<Collection> collections, SutProvider<CollectionsController> sutProvider)
+    {
+        // Arrange
+        var userId = Guid.NewGuid();
+        var groupId = Guid.NewGuid();
+        var model = new BulkCollectionAccessRequestModel
+        {
+            CollectionIds = collections.Select(c => c.Id),
+            Users = new[] { new SelectionReadOnlyRequestModel { Id = userId, ReadOnly = true } },
+            Groups = new[] { new SelectionReadOnlyRequestModel { Id = groupId, ReadOnly = true } },
```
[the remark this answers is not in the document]
shane-melton: Good idea, done!

src/Core/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommand.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            throw new BadRequestException("No collections were provided.");
+        }
+
+        if (collections.Any(c => c.OrganizationId != orgId))
+        {
+            throw new BadRequestException("All collections must belong to the same organization.");
+        }
+
+        var collectionUserIds = usersAccess?.Select(u => u.Id).ToList();
```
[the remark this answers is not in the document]
shane-melton: It is unlikely to have duplicates here, but there was no guarantee so I added the `distinct()` to be sure.

src/Core/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommand.cs
```diff
@@ -0,0 +1,88 @@
...
+        IGroupRepository groupRepository,
+        IEventService eventService)
+    {
+        _collectionRepository = collectionRepository;
+        _organizationUserRepository = organizationUserRepository;
+        _groupRepository = groupRepository;
+        _eventService = eventService;
+    }
+
+    public async Task AddAccessAsync(Guid organizationId, ICollection<Collection> collections,
```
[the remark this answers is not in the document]
shane-melton: Fair point, updated and removed the `organizationId` parameter.

test/Api.Test/Controllers/CollectionsControllerTests.cs
```diff
@@ -250,4 +253,123 @@ await sutProvider.GetDependency<IDeleteCollectionCommand>()
...
+    public async Task PostBulkCollectionAccess_CollectionsNotFound_Throws(Guid orgId, User actingUser, ICollection<Collection> collections, SutProvider<CollectionsController> sutProvider)
+    {
+        var userId = Guid.NewGuid();
+        var groupId = Guid.NewGuid();
+        var model = new BulkCollectionAccessRequestModel
+        {
+            CollectionIds = collections.Select(c => c.Id),
+            Users = new[] { new SelectionReadOnlyRequestModel { Id = userId, ReadOnly = true } },
+            Groups = new[] { new SelectionReadOnlyRequestModel { Id = groupId, ReadOnly = true } },
+        };
```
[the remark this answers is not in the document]
shane-melton: I looked into writing a fixture for this, but it was proving to be a bit of a headache to keep the models and Ids in sync with each other so that they could be easily asserted in the tests. While it is somewhat repetitive, I believe its fairly minimal with only 3 tests. The fixture I was attempting to create seemed like it was going to add lot more complexity for little gain and reduced readability. 

I could be overcomplicating your suggestion, so if you have something more specific in mind I'd be happy to take a look again.

gbubemismith:
I think the only change touching the `vault` team is on the `CollectionFixture` : (  , which looks good IMO. I would be approving based on that.

test/Api.Test/Controllers/CollectionsControllerTests.cs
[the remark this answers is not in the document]
eliykat: Nope, I trust you on this one!

test/Core.Test/OrganizationFeatures/OrganizationCollections/BulkAddCollectionAccessCommandTests.cs
eliykat: That's messy that the `Id` can refer to different things. Let's leave the explicit initialization, it seems safest.

src/Sql/dbo/Stored Procedures/Collection_CreateOrUpdateAccessForMany.sql
```diff
@@ -0,0 +1,95 @@
+CREATE PROCEDURE [dbo].[Collection_CreateOrUpdateAccessForMany]
```
gbubemismith, line 1: This looks good. Maybe some more comments would make it easier to maintain in future.

gbubemismith (approved):

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -473,6 +473,97 @@ public async Task DeleteManyAsync(IEnumerable<Guid> collectionIds)
...
+                        existingCollectionGroup.ReadOnly = requestedGroup.ReadOnly;
+                        existingCollectionGroup.Manage = requestedGroup.Manage;
+                        dbContext.CollectionGroups.Update(existingCollectionGroup);
+                    }
+                }
+            }
+
+            await dbContext.SaveChangesAsync();
+            await dbContext.UserBumpAccountRevisionDateByCollectionIdsAsync(collectionIdsList, organizationId);
+            await dbContext.SaveChangesAsync();
```
eliykat, line 563: Do we need to call `SaveChangesAsync` twice?

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -473,6 +473,97 @@ public async Task DeleteManyAsync(IEnumerable<Guid> collectionIds)
...
+                                Manage = requestedUser.Manage
+                            });
+                            continue;
+                        }
+
+                        // It already exists, update it
+                        existingCollectionUser.HidePasswords = requestedUser.HidePasswords;
+                        existingCollectionUser.ReadOnly = requestedUser.ReadOnly;
+                        existingCollectionUser.Manage = requestedUser.Manage;
+                        dbContext.CollectionUsers.Update(existingCollectionUser);
```
eliykat, line 518: Any idea on performance here, calling `Update` for each entity that we're updating, vs. adding them to a list and calling `UpdateRange` at the end? Compare this to other bulk methods in this repository, e.g. `ReplaceCollectionGroupsAsync`, which follow this pattern. I don't know for sure, but I'm inclined to follow the other methods as examples.

Same for calling `Add` vs. `AddRange`.

eliykat (changes requested):

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
shane-melton: Great question! I did a bit digging and it looks like there was a performance difference in EF6. However, in EF Core the `AddRange` and `UpdateRange` are [purely convenience methods and do not have a significant performance difference](https://learn.microsoft.com/en-us/ef/core/change-tracking/miscellaneous#addrange-updaterange-attachrange-and-removerange).

I'm not familiar with exactly when we updated to EF core, but that is likely why the other `ReplaceCollectionGroupsAsync` goes out of its way to use the `*Range` methods (it looks like it written in '21). 

I feel updating the `dbContext` collections directly is more readable, so I'd prefer to keep it as is unless you have any objections. _(also fwiw, we use this same pattern in the `UpdateUsersAsync` method in the repository)_

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
shane-melton: I'm not completely familiar with the problem, but we "recently" [had to make adjustments](https://github.com/bitwarden/server/pull/2471#discussion_r1166072351) in a similar method to call `SaveChangesAync` before bumping the user account revision dates. I followed that pattern here as well, I'll add the same comment in that method to help with any confusion for any changes in the future.

src/Sql/dbo/Stored Procedures/Collection_CreateOrUpdateAccessForMany.sql
```diff
@@ -0,0 +1,95 @@
...
+	 -- Groups
+	;WITH [NewCollectionGroups] AS (
+		SELECT
+			cId.[Id] AS [CollectionId],
+			gu.[Id] AS [GroupId],
+			gu.[ReadOnly],
+			gu.[HidePasswords],
+			gu.[Manage]
+		FROM
+			@Groups AS gu
```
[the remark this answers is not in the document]
shane-melton: Done!

vincentsalucci (approved):

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
eliykat: Thank you for the reference, all good with me.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
eliykat: Makes sense, thank you!

eliykat (approved):

PR #3239, opened 2023-08-30
[PM-3487] prevent account enumeration on auth request endpoint

jlf0dev:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The create auth request endpoint was returning different errors based on if the user existed or not. We can't leak that information on anonymous endpoint due to account enumeration.

The solution here is to return the same error message for if the user or device isn't found.

## Code changes

## Before you submit


JaredSnider-Bitwarden (approved):
LGTM!

PR #3241, opened 2023-08-30
[AC-1125] Enforce org setting for creating/deleting collections

vincentsalucci:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Code changes
- **src/Api/Controllers/CollectionsController.cs**: Updated `Post` `Delete` and `DeleteMany` endpoints to use the new `AuthorizationService` flow. 
- **src/Api/Startup.cs**: Added call to `AddAuthorizationHandlers` from new service collection extension
- **src/Api/Utilities/ServiceCollectionExtensions.cs**: Created new service collection extension to scope the `CollectionAuthorizationHandler`
- **src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs**: Created `CollectionAuthorizationHandler` for (initially handling) the `Create` and `Delete` operations.
- **src/Api/Vault/AuthorizationHandlers/Collections/CollectionOperations.cs**: Created `Create` and `Delete` `CollectionOperations` for use within the `CollectionAuthorizationHandler`
- **src/Core/Context/CurrentContext.cs**: Deprecated calls to `CreateNewCollections` and `DeleteAnyCollections`. Added `GetOrganization` helper method.
- **src/Core/Context/CurrentContextOrganization.cs**: Added `LimitCollectionCdOwnerAdmin`
- **src/Core/Services/Implementations/CollectionService.cs**: Newly created collections will assign the creating user along with the `Manage` permission.
- **src/Core/Services/Implementations/OrganizationService.cs**: Use `CurrentContextOrganization.Permissions` for direct comparison (Owner, Admin, Provider user are handled at the top of the method)
- **Tests**: Updated to not use `CreateNewCollections` / `DeleteAnyCollections`

## Before you submit


src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            case not null when requirement == CollectionOperations.ModifyAccess:
+                await CanManageCollectionAccessAsync(context, requirement, resources);
+                break;
+        }
+    }
+
+    private async Task CanCreateAsync(AuthorizationHandlerContext context, CollectionOperationRequirement requirement, ICollection<Collection> resources)
+    {
+        // Bulk creation of collections is not supported
+        if (resources.Count > 1)
```
shane-melton, line 46: thought: Not sure if this is relevant to the authorization logic and should instead be handled by command/service business logic. I don't see any particular reason we couldn't authorize the creation of multiple collections at once? Just need to check that the all belong to the same organization like the other operations check.

I imagine it could be useful in the future, so we can use this auth handler for creating multiple collections during a data import or something similar.

shane-melton (changes requested):
Just had a few questions/suggestions, but otherwise looks really good!

src/Core/Context/CurrentContext.cs
```diff
@@ -321,6 +321,7 @@ public async Task<bool> AccessReports(Guid orgId)
                     && (o.Permissions?.AccessReports ?? false)) ?? false);
     }
 
+    [Obsolete("CreateNewCollections is deprecated, use the CollectionAuthorizationHandler for controller level permission checks or CurrentContext.GetOrganization if permission checks are needed outside of that scope.")]
     public async Task<bool> CreateNewCollections(Guid orgId)
```
eliykat, line 325: This looks unused. Let's delete it! (even better than marking it as obsolete)

src/Core/Context/CurrentContext.cs
```diff
@@ -333,6 +334,7 @@ public async Task<bool> EditAnyCollection(Guid orgId)
                     && (o.Permissions?.EditAnyCollection ?? false)) ?? false);
     }
 
+    [Obsolete("DeleteAnyCollection is deprecated, use the CollectionAuthorizationHandler for controller level permission checks or CurrentContext.GetOrganization if permission checks are needed outside of that scope.")]
     public async Task<bool> DeleteAnyCollection(Guid orgId)
```
eliykat, line 338: Same here

src/Core/Context/CurrentContext.cs
```diff
@@ -507,6 +510,11 @@ public bool AccessSecretsManager(Guid orgId)
         return Providers;
     }
 
+    public CurrentContextOrganization GetOrganization(Guid orgId)
+    {
+        return Organizations.Find(o => o.Id == orgId);
+    }
+
```
eliykat, line 517: @shane-melton pointed out to me that `CurrentContext` already has an interface for this, `OrganizationMembershipAsync`. Let's use that instead of adding our own. (I'll also update my own PR to match)

src/Core/Context/ICurrentContext.cs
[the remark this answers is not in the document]
eliykat: I get the impression it's mostly an individual effort so far, the idea being that more type safety is better. It's a good question to ask in `#code` though.

test/Api.Test/Controllers/CollectionsControllerTests.cs
```diff
@@ -210,7 +210,7 @@ await sutProvider.GetDependency<IDeleteCollectionCommand>()
 
 
     [Theory, BitAutoData]
-    public async Task DeleteMany_UserCanNotAccessCollections_FiltersOutInvalid(Guid orgId, User user, Collection collection1, Collection collection2, SutProvider<CollectionsController> sutProvider)
+    public async Task DeleteMany_UserCanNotAccessCollections_FiltersOutInvalid(Guid orgId, Collection collection1, Collection collection2, SutProvider<CollectionsController> sutProvider)
```
eliykat, line 213: Is this test still accurate? I think the auth handler will throw if the user doesn't have access to all collections requested, it doesn't filter as described in the test name.

test/Api.Test/Vault/AuthorizationHandlers/CollectionAuthorizationHandlerTests.cs
eliykat, line 1: Can you please write tests for the new operations you've added?

eliykat (changes requested):
Great work!

src/Api/Controllers/CollectionsController.cs
```diff
@@ -141,19 +149,16 @@ public async Task<CollectionResponseModel> Post(Guid orgId, [FromBody] Collectio
...
         }
 
         var groups = model.Groups?.Select(g => g.ToSelectionReadOnly());
         var users = model.Users?.Select(g => g.ToSelectionReadOnly());
 
-        var assignUserToCollection = !(await _currentContext.EditAnyCollection(orgId)) &&
-            await _currentContext.EditAssignedCollections(orgId);
-
-        await _collectionService.SaveAsync(collection, groups, users, assignUserToCollection ? _currentContext.UserId : null);
+        await _collectionService.SaveAsync(collection, groups, users, _currentContext.UserId);
```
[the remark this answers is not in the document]
vincentsalucci: That is correct! That was my interpretation from this statement in the Flexible Collections confluence document:

> When a user creates a collection, they will automatically be granted a “can manage” permission to the newly created collection.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -189,35 +194,35 @@ public async Task PutUsers(Guid orgId, Guid id, [FromBody] IEnumerable<Selection
     [HttpPost("{id}/delete")]
     public async Task Delete(Guid orgId, Guid id)
     {
-        if (!await CanDeleteCollectionAsync(orgId, id))
+        var collection = await GetCollectionAsync(id, orgId);
```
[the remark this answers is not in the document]
vincentsalucci: I'm fine with letting the Authorization Handler deal with a null resource (what Shane linked above for the `BulkAuthorizationHandler`). Thumbs up if ya'll agree?

src/Core/Context/CurrentContext.cs
eliykat: Actually I just realised that the existing method gets _all_ organizations rather than a specific org. I still like adding an interface for this rather than having to `Find` the org yourself every time.

Maybe this should just become an async method which is a wrapper over `OrganizationMembershipAsync`?

src/Api/Controllers/CollectionsController.cs
```diff
@@ -189,35 +194,35 @@ public async Task PutUsers(Guid orgId, Guid id, [FromBody] IEnumerable<Selection
...
     {
-        var orgId = new Guid(model.OrganizationId);
         var collectionIds = model.Ids.Select(i => new Guid(i));
-        if (!await _currentContext.DeleteAssignedCollections(orgId) && !await _currentContext.DeleteAnyCollection(orgId))
+        var collections = await _collectionRepository.GetManyByManyIdsAsync(collectionIds);
+
+        if (!collections.Any())
         {
-            throw new NotFoundException();
+            throw new BadRequestException("No collections found.");
```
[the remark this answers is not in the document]
vincentsalucci: I think that you could make a case for using a `BadRequestException` because the resulting collections are based solely on the `ids` that were passed in, thus (technically) resulting in a bad request. But since we've agreed upon the handler handling null resource checks, a `NotFoundException` is also an arguable exception!

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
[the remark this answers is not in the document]
vincentsalucci: I implemented all of the above **except** for the `_currentContext.UserId` check. The only reason was because Rider was still warning me that the value could be null because we checked it outside of the context of the methods using it.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
vincentsalucci: I agree that it takes minimal effort to make this method handle bulk resources. I've updated the method accordingly.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+        if (org == null)
+        {
+            context.Fail();
+            return;
+        }
+
+        var success = !org.LimitCollectionCdOwnerAdmin || (org.Type == OrganizationUserType.Owner ||
+                                                           org.Type == OrganizationUserType.Admin ||
+                                                           org.Permissions.CreateNewCollections ||
+                                                           org.Permissions.EditAnyCollection ||
```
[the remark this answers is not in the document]
vincentsalucci: Correct - purely maintaining existing logic. I will connect with Product to discuss the permission structure for both the `Create` and `Delete` actions.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+        var org = _currentContext.GetOrganization(targetOrganizationId);
+        if (org == null)
+        {
+            context.Fail();
+            return;
+        }
+
+        // Owners, Admins, Providers, and users with DeleteAnyCollection or EditAnyCollection permission can always delete collections
+        if (
+            org.Permissions is { DeleteAnyCollection: true, EditAnyCollection: true } ||
```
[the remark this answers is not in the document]
vincentsalucci: Ahhh - good catch. Reverted that conditional to explicit ORs and aligned all methods with how that conditional is structured.

See above comment within the `CanCreateAsync` method in regards to `EditAnyCollection` inclusion.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            org.Type is OrganizationUserType.Admin or OrganizationUserType.Owner ||
+            await _currentContext.ProviderUserForOrgAsync(org.Id))
+        {
+            context.Succeed(requirement);
+            return;
+        }
+
+        // Other members types should have the Manage capability for all collections being deleted
+        if (!org.LimitCollectionCdOwnerAdmin)
```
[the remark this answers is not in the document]
vincentsalucci: Thank you - cleaned up the affected methods.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionOperations.cs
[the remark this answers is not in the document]
vincentsalucci: I have **no** idea how that happened? Regardless - fixed!

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -2120,12 +2120,13 @@ private async Task<bool> ValidateCustomPermissionsGrant(Guid organizationId, Per
             return false;
         }
 
-        if (permissions.CreateNewCollections && !await _currentContext.CreateNewCollections(organizationId))
+        var org = _currentContext.GetOrganization(organizationId);
```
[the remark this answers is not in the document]
vincentsalucci: Can do - I moved the dependent conditionals to the end of the method just in case the org returned is null (shouldn't happen).

src/Core/Context/CurrentContext.cs
vincentsalucci: Deleted.

src/Core/Context/CurrentContext.cs
vincentsalucci: Deleted

src/Core/Context/CurrentContext.cs
vincentsalucci: This gets really messy because the `OrganizationMembershipAsync` methods requires a reference to the `OrganizationUserRepository`. This poses problems for any usages within the current context itself.  Shane and I talked a week or two ago about this and both came to the conclusion that `Organizations` shouldn't ever be `null` since its initialized when the context is set via the `claimsDict`. Do you know of an instance where the `Organizations` property won't be set?

test/Api.Test/Controllers/CollectionsControllerTests.cs
vincentsalucci: Good call! Removed since the handler handles that and the test above handles invalid permissions.

shane-melton (changes requested):
Looks really good! Only had a few more comments.

The shared test logic is also looking nice now! Are you still planning on adding specific tests for the new authorization operations? (You may already be in the middle of this; if so, please ignore)

 I believe the two new operations should only require two new tests similar to `CanManageCollectionAccessAsync_Success` with their own `BitAutoData` related to the different combinations of user type/permissions and org settings (e.g. `LimitCollectionCdOwnerAdmin`).

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -24,101 +25,103 @@ public CollectionAuthorizationHandler(ICurrentContext currentContext, ICollectio
...
         {
-            throw new NotSupportedException();
+            context.Succeed(requirement);
+            return;
         }
 
-        // Acting user is not a member of the target organization, fail
-        var org = _currentContext.GetOrganization(resources.First().OrganizationId);
-        if (org == null)
+        // The above conditional should have passed if the following collection management setting is true, fail check
```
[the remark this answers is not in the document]
vincentsalucci: Updated!

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -24,101 +25,103 @@ public CollectionAuthorizationHandler(ICurrentContext currentContext, ICollectio
...
+            org.Permissions.DeleteAnyCollection || org.Permissions.EditAnyCollection ||
             await _currentContext.ProviderUserForOrgAsync(org.Id))
         {
             context.Succeed(requirement);
             return;
         }
 
-        // Other members types should have the Manage capability for all collections being deleted
-        if (!org.LimitCollectionCdOwnerAdmin)
+        // The above conditional should have passed if the following collection management setting is true, fail check
```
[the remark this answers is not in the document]
vincentsalucci: Updated!

src/Api/Models/Request/CollectionRequestModel.cs
```diff
@@ -34,7 +34,8 @@ public virtual Collection ToCollection(Collection existingCollection)
 public class CollectionBulkDeleteRequestModel
 {
     [Required]
-    public IEnumerable<string> Ids { get; set; }
+    public IEnumerable<Guid> Ids { get; set; }
+    [Obsolete("OrganizationId is no longer required and will be removed in a future release")]
```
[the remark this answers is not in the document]
vincentsalucci: I made a task for it. Just cleanliness (imo). Trying to keep this PR within its original scope while also being able to group small, one-off PRs for server/client with the removal of the property.

test/Api.Test/Vault/AuthorizationHandlers/CollectionAuthorizationHandlerTests.cs
vincentsalucci: Refactored existing auth handler tests to match the changes to some pre-req conditional checks. Added a couple of success path tests for the new methods

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+        if (!org.LimitCollectionCdOwnerAdmin)
+        {
+            context.Succeed(requirement);
+            return;
+        }
+
+        // Owners, Admins, Providers, and users with CreateNewCollections or EditAnyCollection permission can always create collections
+        if (
+            org.Type is OrganizationUserType.Owner or OrganizationUserType.Admin ||
+            org.Permissions.CreateNewCollections || org.Permissions.EditAnyCollection ||
```
shane-melton, line 83: I believe that `org.Permissions` can be null if the user type is not `Custom` and this can possible throw a null reference exception. I ended up using pattern matching in the `CanManageCollectionAccessAsync` method to avoid having extra null checks.

However, I also believe @eliykat [mentioned in his PR](https://github.com/bitwarden/server/pull/2845#discussion_r1306745876) that he was going to modify `CurrentContextOrganization` so that `Permissions` is never null; so this may not be a problem in the end.

shane-melton (changes requested):
I think we're pretty much set with this. I just remembered a potential null reference exception I ran into my own work and wanted to bring attention to it before approving.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
vincentsalucci: I was going through the motions here and assumed @eliykat change was already in place. I've pulled it into my PR!

shane-melton (commented):
All my comments have been resolved, thanks! 

Any idea why the tests/checks are failing after the merge with the `feature/flexible-collections` branch?

vincentsalucci:
> All my comments have been resolved, thanks!
> 
> Any idea why the tests/checks are failing after the merge with the `feature/flexible-collections` branch?

@shane-melton Yup - an old reference to the now deleted `CreateNewCollections` method. Fixed it with the necessary changes until the next iteration of the `CollectionAuthorizationHandler` is applied (will handled view/manage collection access)

shane-melton (approved):
Looks good to go!

src/Core/Context/CurrentContext.cs
```diff
@@ -361,9 +347,20 @@ public async Task<bool> DeleteAssignedCollections(Guid orgId)
 
     public async Task<bool> ViewAssignedCollections(Guid orgId)
     {
-        return await CreateNewCollections(orgId) // Required to display the existing collections under which the new collection can be nested
-               || await EditAssignedCollections(orgId)
-               || await DeleteAssignedCollections(orgId);
+        /*
+         * Required to display the existing collections under which the new collection can be nested.
+         * Owner, Admin, Manager, and Provider checks are handled via the EditAssigned/DeleteAssigned context calls.
+         * This entire method will be moved to the ControllerAuthorizationHandler in the future
```
eliykat, line 353: typo 

```suggestion
         * This entire method will be moved to the CollectionAuthorizationHandler in the future
```

src/Core/Context/CurrentContext.cs
```diff
@@ -361,9 +347,20 @@ public async Task<bool> DeleteAssignedCollections(Guid orgId)
...
+         */
+        var canCreateNewCollections = false;
+        var org = GetOrganization(orgId);
+        if (org != null)
+        {
+            canCreateNewCollections = !org.LimitCollectionCdOwnerAdmin || org.Permissions.CreateNewCollections;
```
eliykat, line 359: This doesn't properly replace the `CurrentContext.CreateNewCollections` method (or replicate the new check in the auth handler). The second half of the boolean is only checking the custom permission `CreateNewCollections`, not whether the user is an owner/admin/provider.

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -2083,7 +2083,7 @@ private async Task ValidateOrganizationCustomPermissionsEnabledAsync(Guid organi
 
     private async Task<bool> ValidateCustomPermissionsGrant(Guid organizationId, Permissions permissions)
     {
-        if (permissions == null || await _currentContext.OrganizationOwner(organizationId) || await _currentContext.OrganizationAdmin(organizationId))
+        if (permissions == null || await _currentContext.OrganizationAdmin(organizationId))
```
eliykat, line 2086: This is an optimization to take advantage of the cascading nature of these checks, right? `OrganizationAdmin` includes owners and providerUsers?

eliykat (changes requested):
Just a few things. I'm OOO tomorrow so feel free to finalise it without me.

src/Core/Services/Implementations/OrganizationService.cs
vincentsalucci: That's correct! Checking for an explicit `OrganizationOwner` is unnecessary when paired with an `OrganizationAdmin` check because the latter also checks for owner/providers user level permissions.

src/Core/Context/CurrentContext.cs
vincentsalucci: This is correct - however, because it is paired with two other current context permission checks using the `OR` operator, the remaining owner/admin/manager/provider checks are handled there **first** and if nothing passes above, the remaining collection management and `CreateNewCollections` permission is checked.

src/Core/Context/CurrentContext.cs
vincentsalucci: Updated - thank you!

shane-melton (approved):

src/Core/Context/CurrentContext.cs
eliykat: I accept that!

eliykat (approved):

PR #3244, opened 2023-08-31
[PM-3729]  Add old client / new server backward compatibility

kspearrin:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
https://bitwarden.atlassian.net/browse/PM-3729

Old clients do not know how to deal with Fido2Key ciphers, so filter them from the sync call so that they never get them.

## Code changes

* **SyncController.cs:** Private method for filtering ciphers by type for older clients.
* **Constants.cs:** Define Fido2KeyCipherMinimumVersion constant

## Before you submit


kspearrin:
Note that `collectionCiphersGroupDict` may send cipher ids down to the client that we have since filtered out, but I am unsure if this will cause an actual problem. I don't think so, since I assume clients would gracefully skip over them, but I have not been able to confirm.

differsthecat (approved):

PR #3245, opened 2023-09-03
[AC-1373] Flexible Collections

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+            throw new BadRequestException("Requested collections must belong to the same organization.");
+        }
+
+        // Acting user is not a member of the target organization, fail
+        var org = _currentContext.GetOrganization(targetOrganizationId);
+        if (org == null)
+        {
+            context.Fail();
+            return;
+        }
```
eliykat, line 52: This will give a false negative result for providers.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -248,17 +272,6 @@ private async Task<Collection> GetCollectionAsync(Guid id, Guid orgId)
...
-    {
-        if (collectionId != default)
-        {
-            return false;
-        }
-
-        return await _currentContext.CreateNewCollections(orgId);
-    }
-
     private async Task<bool> CanEditCollectionAsync(Guid orgId, Guid collectionId)
```
eliykat, line 275: This is duplicating logic in the authorization handler and should be moved there. I know that we're doing it incrementally, but this is a bit different because it's identical to the "manage access" check. It should at least be marked deprecated, but preferably just moved there so we don't have identical logic in 2 separate places.

eliykat (commented):
Some things I noticed while reviewing the WIP today.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
eliykat: Logged as AC-1696

src/Api/Controllers/CollectionsController.cs
eliykat: Will be handled in AC-1139

shane-melton (approved):
Looks good to go! 🎉

r-tome (approved):
LGTM!

PR #3267, opened 2023-09-13
no longer cache remember me token

ike-kottlowski:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Users are unable to "remember me" when logging in with 2FA. "Remember Me" flow uses the `twoFactorToken` which was being saved to the cache, this caused the "Remember Me Token" to be saved to the cache and to cause the client to request 2FA even though the token was valid.

## Code changes

* **BaseRequestValidator.cs:** "Remember Me" token is no longer saved to cache.

## Before you submit


JaredSnider-Bitwarden (approved):
Pulled this down and tested against web with email 2FA. Remember me worked so LGTM!

PR #3271, opened 2023-09-14
Migration to remove WebAuthn 2FA method from Free users

trmartin4:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Other (migration script)
```

## Objective

We have users who currently do not have any 2FA method, with the only one being WebAuthn, which is effectively disabled by a server-side permission check for Premium status.

With WebAuthn being made free, we want to avoid these users suddenly being forced to provide 2FA using a key that they haven't used in a long time, by deleting that key from their TwoFactorProviders.

## Before you submit

### DIFF
--- a/util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
+++ b/util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
@@ -33,7 +33,7 @@ from @TwoFactorByUser t1
 where t1.TwoFactorType = 7
 AND NOT EXISTS (SELECT * FROM @TwoFactorByUser t2 WHERE t2.Id = t1.Id and t2.TwoFactorType <> 7)
 
-UPDATE [User]
-SET TwoFactorProviders = NULL
-FROM @TwoFactorByUser tf
-WHERE tf.Id = [User].Id
+-- UPDATE [User]
+-- SET TwoFactorProviders = NULL
+-- FROM @TwoFactorByUser tf
+-- WHERE tf.Id = [User].Id
--- a/util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
+++ b/util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
@@ -2,24 +2,24 @@
 -- When a 2FA method is disabled, it is removed from the TwoFactorProviders array
 
 -- Problem statement:
--- We have users who currently do not have any 2FA method, with the only one being
+-- We have users who currently do not have any available 2FA method, with the only one being
 -- WebAuthn, which is effectively disabled by a server-side permission check for Premium status.
 -- With WebAuthn being made free, we want to avoid these users suddenly being forced
 -- to provide 2FA using a key that they haven't used in a long time, by deleting that key from their TwoFactorProviders.
 
-declare @UsersWithoutPremium TABLE
+DECLARE @UsersWithoutPremium TABLE
 (
     Id UNIQUEIDENTIFIER,
     TwoFactorProviders NVARCHAR(MAX)
 );
 
-declare @TwoFactorMethodsForUsersWithoutPremium TABLE
+DECLARE @TwoFactorMethodsForUsersWithoutPremium TABLE
 (
     Id UNIQUEIDENTIFIER,
     TwoFactorType INT
 )
 
-declare @UsersToAdjust TABLE
+DECLARE @UsersToAdjust TABLE
 (
     Id UNIQUEIDENTIFIER
 );
@@ -46,19 +46,19 @@ CROSS APPLY OPENJSON(u.TwoFactorProviders) tfp1
 CROSS APPLY OPENJSON(tfp1.[value]) WITH (
    [Enabled] BIT '$.Enabled'
 ) tfp2
-WHERE [Enabled] = 'true' -- We only want enabled 2FA methods
+WHERE [Enabled] = 1 -- We only want enabled 2FA methods
 
-insert into @UsersToAdjust
-select t1.Id
-from @TwoFactorMethodsForUsersWithoutPremium t1
-where t1.TwoFactorType = 7
+INSERT INTO @UsersToAdjust
+SELECT t1.Id
+FROM @TwoFactorMethodsForUsersWithoutPremium t1
+WHERE t1.TwoFactorType = 7
 AND NOT EXISTS 
     (SELECT * 
     FROM @TwoFactorMethodsForUsersWithoutPremium t2 
     WHERE t2.Id = t1.Id AND t2.TwoFactorType <> 7 AND t2.TwoFactorType <> 4)
 
-select *
-from @UsersToAdjust
+SELECT *
+FROM @UsersToAdjust
 
 -- UPDATE [User]
 -- SET TwoFactorProviders = NULL
--- a/util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
+++ b/util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
@@ -1,0 +1,38 @@
+-- Assumptions:
+-- When a 2FA method is disabled, it is removed from the TwoFactorProviders array
+
+-- Problem statement:
+-- We have users who currently do not have any 2FA method, with the only one being
+-- WebAuthn, which is effectively disabled by a server-side permission check for Premium status.
+-- With WebAuthn being made free, we want to avoid these users suddenly being forced
+-- to provide 2FA using a key that they haven't used in a long time, by deleting that key from their TwoFactorProviders.
+declare @TwoFactorByUser TABLE
+(
+    Id UNIQUEIDENTIFIER,
+    Email NVARCHAR(256),
+    TwoFactorType INT,
+    TypeEnabled BIT
+)
+
+INSERT INTO @TwoFactorByUser
+SELECT  u.Id, 
+        u.Email,
+        tfp1.[key] as TwoFactorType, 
+        CASE [Enabled] WHEN 'true' THEN 1 ELSE 0 END as TypeEnabled
+FROM [User] u
+CROSS APPLY OPENJSON(u.TwoFactorProviders) tfp1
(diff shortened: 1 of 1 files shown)


trmartin4:
📓 I will remove the `SELECT` prior to opening the PR for review - this is for initial testing.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,39 @@
...
+
+select *
+from @TwoFactorByUser t1
+where t1.TwoFactorType = 7
+AND NOT EXISTS (SELECT * FROM @TwoFactorByUser t2 WHERE t2.Id = t1.Id and t2.TwoFactorType <> 7)
+
+-- UPDATE [User]
+-- SET TwoFactorProviders = NULL
+-- FROM @TwoFactorByUser tf
+-- WHERE tf.Id = [User].Id
```
withinfocus, line 39: ⛏️ Newline.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,39 @@
...
+INSERT INTO @TwoFactorByUser
+SELECT  u.Id, 
+        u.Email,
+        tfp1.[key] as TwoFactorType
+FROM [User] u
+LEFT OUTER JOIN [OrganizationUser] ou on ou.UserId = u.Id
+LEFT OUTER JOIN [Organization] o on o.Id = ou.OrganizationId
+CROSS APPLY OPENJSON(u.TwoFactorProviders) tfp1
+CROSS APPLY OPENJSON(tfp1.[value]) WITH (
+   [Enabled] nvarchar(10) '$.Enabled'
```
withinfocus, line 25: 💭 Maybe cast to a `BOOL` for performance on the later `WHERE`?

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,39 @@
...
+) tfp2
+WHERE [Enabled] = 'true' -- We only want enabled 2FA methods
+AND Premium = 0 -- User isn't Premium
+AND (o IS NULL OR (o IS NOT NULL AND o.Enabled = 1 AND o.UsersGetPremium = 0)) -- User doesn't have Premium from their org
```
withinfocus, line 29: 🎨 Since we're running this as a script we have the benefit of time and an ad hoc run, so why not make this less complicated? I think it'd be cleaner and with more straightforward results if we ran the above twice for users in or out of an org. This looks fine to me regardless.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,39 @@
...
+
+INSERT INTO @TwoFactorByUser
+SELECT  u.Id, 
+        u.Email,
+        tfp1.[key] as TwoFactorType
+FROM [User] u
+LEFT OUTER JOIN [OrganizationUser] ou on ou.UserId = u.Id
+LEFT OUTER JOIN [Organization] o on o.Id = ou.OrganizationId
+CROSS APPLY OPENJSON(u.TwoFactorProviders) tfp1
+CROSS APPLY OPENJSON(tfp1.[value]) WITH (
```
withinfocus, line 24: ⚠️ Could be really, really slow in our cloud but don't see an alternative.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,39 @@
...
+declare @TwoFactorByUser TABLE
+(
+    Id UNIQUEIDENTIFIER,
+    Email NVARCHAR(256),
+    TwoFactorType INT
+)
+
+INSERT INTO @TwoFactorByUser
+SELECT  u.Id, 
+        u.Email,
```
withinfocus, line 18: ⚠️ This is unused.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,39 @@
...
+   [Enabled] nvarchar(10) '$.Enabled'
+) tfp2
+WHERE [Enabled] = 'true' -- We only want enabled 2FA methods
+AND Premium = 0 -- User isn't Premium
+AND (o IS NULL OR (o IS NOT NULL AND o.Enabled = 1 AND o.UsersGetPremium = 0)) -- User doesn't have Premium from their org
+
+select *
+from @TwoFactorByUser t1
+where t1.TwoFactorType = 7
+AND NOT EXISTS (SELECT * FROM @TwoFactorByUser t2 WHERE t2.Id = t1.Id and t2.TwoFactorType <> 7)
```
withinfocus, line 34: 💭 I think there's a way we could filter this out in the above insert and gain in performance, but the subselect while inserting results may not work. Still thinking about this one.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: Fixed, thank you.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: Adjusted to a `BIT` - looks like SQL handled the casting.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: I changed the approach here with a new commit - instead pre-selecting all the users without Premium, and then using that as the base for the `CROSS APPLY`.  This simplifies this logic and perhaps also reduces the number of rows for which the `CROSS APPLY` needs to work.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: Removed.  I had it on there to see what accounts I was getting back in testing, but not necessary any more.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,66 @@
+-- Assumptions:
+-- When a 2FA method is disabled, it is removed from the TwoFactorProviders array
+
+-- Problem statement:
+-- We have users who currently do not have any 2FA method, with the only one being
```
withinfocus, line 5: ⛏️ "any _other_" right?

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,66 @@
...
+-- From users without Premium, get their enabled 2FA methods
+INSERT INTO @TwoFactorMethodsForUsersWithoutPremium
+SELECT  u.Id, 
+        tfp1.[key] as TwoFactorType
+FROM @UsersWithoutPremium u
+CROSS APPLY OPENJSON(u.TwoFactorProviders) tfp1
+CROSS APPLY OPENJSON(tfp1.[value]) WITH (
+   [Enabled] BIT '$.Enabled'
+) tfp2
+WHERE [Enabled] = 'true' -- We only want enabled 2FA methods
```
withinfocus, line 49: ❓ Does this work? Probably some Boolean helper in the engine. I think `= 1` will help a bit.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,66 @@
...
+
+insert into @UsersToAdjust
```
withinfocus, line 51: ⛏️ You start using all lowercase here 😜.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
```diff
@@ -0,0 +1,66 @@
...
+select t1.Id
+from @TwoFactorMethodsForUsersWithoutPremium t1
+where t1.TwoFactorType = 7
+AND NOT EXISTS 
+    (SELECT *
```
withinfocus, line 56: 💭 I don't think it matters, but `SELECT 1` vs. `SELECT *` may slightly relax I/O. Probably not worth pursuing.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: From their perspective on the client, they don't have any, as WebAuthn is filtered out on the server with a permission check.  I clarified with "available" here.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: Fixed!  Thank you for catching that.

util/Migrator/DbScripts/2023-09-14_00_Remove_WebAuthn_For_Free_Users.sql
trmartin4: Ah, yes.  Definitely clearer too.

withinfocus (approved):
Short of the lack of newline 😜

kspearrin (approved):

PR #3272, opened 2023-09-15
[AC-1660] Bump SignalR Redis Package

justindbaur:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Bump `Microsoft.AspNetCore.SignalR.StackExchangeRedis` to the latest version that supports .NET 6. 

I used `dotnet restore` to get packages.lock.json changes.

## Code changes

* Single package update and associated lock file updates.

## Before you submit


src/Notifications/Notifications.csproj
```diff
@@ -8,7 +8,7 @@
   <PropertyGroup Condition=" '$(RunConfiguration)' == 'Notifications-SelfHost' " />
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="6.0.4" />
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.StackExchangeRedis" Version="6.0.4" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.StackExchangeRedis" Version="6.0.22" />
```
Hinton, line 11: Is there a reason redis is part of self-hosted? And do we want to bump the other SignalR dependencies, I'm not familiar with if they expect to stay in sync.

src/Notifications/Notifications.csproj
```diff
@@ -8,7 +8,7 @@
   <PropertyGroup Condition=" '$(RunConfiguration)' == 'Notifications-SelfHost' " />
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="6.0.4" />
```
withinfocus, line 10: ⛏️ I would bump this at the same time.

src/Notifications/packages.lock.json
```diff
@@ -2638,71 +2638,71 @@
...
-          "SendGrid": "9.27.0",
-          "Sentry.Serilog": "3.16.0",
-          "Serilog.AspNetCore": "5.0.0",
-          "Serilog.Extensions.Logging": "3.1.0",
-          "Serilog.Extensions.Logging.File": "2.0.0",
-          "Serilog.Sinks.AzureCosmosDB": "2.0.0",
-          "Serilog.Sinks.SyslogMessages": "2.0.6",
-          "Stripe.net": "40.0.0",
-          "YubicoDotNetClient": "1.2.0"
+          "AWSSDK.SQS": "[3.7.2.47, )",
```
withinfocus, line 2641: ⚠️ I feel like we have to get this figured out. Different tools are doing different things, and after a bit of research I don't think we know how to control it.

src/Notifications/Notifications.csproj
justindbaur: It's not really that it's a part of self-hosted it's just not explicitly dis-allowed. I suspect the number of SH using it is 0. 

No they don't need to be updated at the same time. This package just manages the connection lifetime and the other package is just for how we encode our messages.

src/Notifications/Notifications.csproj
justindbaur: I disagree for something that we want to pick this late into testing, they don't have dependencies on each other and they both interact with separate parts of the SignalR stack. I'm fine with making a separate PR that only goes into `master` though. Would that be okay?

src/Notifications/packages.lock.json
justindbaur: Yeah I don't know... that's why I made a note of the tool I used to do this one so maybe we can track it a little.

Hinton: Rider seems to prefer the old syntax. The dotnet CLI prefers `[)`. I think we should settle on `[)` and ask developers to run `dotnet restore -f --force-evaluate --no-cache` if they change things in Rider.

withinfocus (approved):
Short of the conflicting lockfile output, works for me. I think a PR to replace all the lockfiles that just goes to `master` would be nice to do after this.

PR #3274, opened 2023-09-15
Refresh lockfiles

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Refresh lockfiles using `dotnet restore -f --force-evaluate --no-cache`. Only change should be the internal bitwarden projects which will use the latest version instead of whichever was previously defined.

## Before you submit


bitwarden_license/src/Commercial.Core/packages.lock.json
```diff
@@ -2424,43 +2424,43 @@
...
-          "SendGrid": "9.27.0",
-          "Sentry.Serilog": "3.16.0",
-          "Serilog.AspNetCore": "5.0.0",
-          "Serilog.Extensions.Logging": "3.1.0",
-          "Serilog.Extensions.Logging.File": "2.0.0",
-          "Serilog.Sinks.AzureCosmosDB": "2.0.0",
-          "Serilog.Sinks.SyslogMessages": "2.0.6",
-          "Stripe.net": "40.0.0",
-          "YubicoDotNetClient": "1.2.0"
+          "AWSSDK.SQS": "[3.7.2.47, )",
```
JaredSnider-Bitwarden, line 2427: ❓ is it safe to go from exact version specification to use exclusive bound specification (allowing the use of any newer versions than the currently specified one)?

Hinton: While they look different behavior wise are identical. https://learn.microsoft.com/en-us/nuget/concepts/package-versioning.

To lock to exact versions you need `[version]`.

JaredSnider-Bitwarden: Ah, I'm sorry. Mixed that up. Thank you for the link and clarification.

withinfocus (approved):
Did a quick scan but this is generated content.

From a DevOps and build perspective, the inclusion of inter-project versioning in these e.g. `Core` is maybe something we need to automate with version bumps. I will mention it to them.

JaredSnider-Bitwarden (approved):
👍

PR #3275, opened 2023-09-15
only tracking TOTPs in cache

ike-kottlowski:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Created "white-list" logic for the cache so we do not store things we don't want to, such as `rememberMe` and `yubikey` tokens.

Purpose is to enforce TOTPs so we should only store TOTPs and nothing else.

## Code changes

* **BaseRequestValidator.cs:** added "white-list" logic when we write to the cache for 2FA tokens.

## Before you submit


JaredSnider-Bitwarden (approved):
LGTM!

PR #3276, opened 2023-09-15
Reverting TOTP changes from rc branch

trmartin4:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

Reverting 3 commits that implement PM-2128.  The initial implementation of this work assumed that we needed to cache all types of 2FA tokens, not just TOTP tokens.  Rather than try to patch this for the release, we are reverting it so that we can have a more complete solution in the next release.  The changes will stay in `master` and be iterated upon there in preparation for the November release.

## Code change

- Reverted changes from 3 commits.

## Before you submit

### DIFF
--- a/src/Identity/IdentityServer/BaseRequestValidator.cs
+++ b/src/Identity/IdentityServer/BaseRequestValidator.cs
@@ -163,10 +163,7 @@ public abstract class BaseRequestValidator<T> where T : class
                 }
                 return;
             }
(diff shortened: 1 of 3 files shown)


JaredSnider-Bitwarden (approved):
Reviewed reversions. All look correct and match old state before the work started.

PR #3279, opened 2023-09-18
[PM-3980] Added creationDate property to the Fido2Key object

gbubemismith:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
PR is linked to https://github.com/bitwarden/clients/pull/6334

## Before you submit

### DIFF
--- a/src/Api/Vault/Models/CipherFido2KeyModel.cs
+++ b/src/Api/Vault/Models/CipherFido2KeyModel.cs
@@ -20,6 +20,7 @@ public class CipherFido2KeyModel
         UserDisplayName = data.UserDisplayName;
         Counter = data.Counter;
         Discoverable = data.Discoverable;
+        CreationDate = data.CreationDate;
     }
 
     [EncryptedString]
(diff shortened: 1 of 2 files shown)


src/Api/Vault/Models/CipherFido2KeyModel.cs
```diff
@@ -70,7 +73,8 @@ public CipherLoginFido2KeyData ToCipherLoginFido2KeyData()
             UserHandle = UserHandle,
             UserDisplayName = UserDisplayName,
             Counter = Counter,
-            Discoverable = Discoverable
+            Discoverable = Discoverable,
+            CreationDate = CreationDate.Value
```
coroiu, line 77: issue: it seems we need to use `Value` because `CipherFido2KeyModel.CreationDate` is marked as nullable but `CipherLoginFido2KeyData.CreationDate` is not. I think we need to actually handle the nullable case here, otherwise `.Value` will just initialize with a default value, right?

coroiu (approved):

PR #3281, opened 2023-09-19
Admin: Update Stripe Customer Email on Organization Edit

amorask-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

[PM-186](https://bitwarden.atlassian.net/browse/PM-186)

This PR uses the `StripeSyncService` to update the customer email address in Stripe when the organization is updated from Admin.

## Code changes

* **OrganizationsController.cs:** Inject the `StripeSyncService`. Use it to update the customer email address during an Organization `Edit`. If the Stripe update fails, log an error and re-throw the exception to prevent DB/cache updates and reference event production.

https://github.com/bitwarden/server/assets/144709477/9e3a0fbb-3efe-4337-940c-efafb2cba553

## Before you submit

[PM-186]: https://bitwarden.atlassian.net/browse/PM-186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Admin/Controllers/OrganizationsController.cs
+++ b/src/Admin/Controllers/OrganizationsController.cs
@@ -215,19 +215,20 @@ public class OrganizationsController : Controller
         try
         {
             await _stripeSyncService.UpdateCustomerEmailAddress(organization.GatewayCustomerId, organization.BillingEmail);
-            await _organizationRepository.ReplaceAsync(organization);
-            await _applicationCacheService.UpsertOrganizationAbilityAsync(organization);
-            await _referenceEventService.RaiseEventAsync(new ReferenceEvent(ReferenceEventType.OrganizationEditedByAdmin, organization, _currentContext)
-            {
-                EventRaisedByUser = _userService.GetUserName(User),
-                SalesAssistedTrialStarted = model.SalesAssistedTrialStarted,
-            });
         }
         catch (StripeException stripeException)
         {
             _logger.LogError(stripeException, "Failed to update billing email address in Stripe for Organization with ID '{organizationId}'", organization.Id);
             throw;
         }
+
+        await _organizationRepository.ReplaceAsync(organization);
+        await _applicationCacheService.UpsertOrganizationAbilityAsync(organization);
+        await _referenceEventService.RaiseEventAsync(new ReferenceEvent(ReferenceEventType.OrganizationEditedByAdmin, organization, _currentContext)
+        {
+            EventRaisedByUser = _userService.GetUserName(User),
+            SalesAssistedTrialStarted = model.SalesAssistedTrialStarted,
+        });
 
         return RedirectToAction("Edit", new { id });
     }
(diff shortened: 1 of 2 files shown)


src/Admin/Controllers/OrganizationsController.cs
```diff
@@ -208,13 +212,23 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
+            await _organizationRepository.ReplaceAsync(organization);
+            await _applicationCacheService.UpsertOrganizationAbilityAsync(organization);
+            await _referenceEventService.RaiseEventAsync(new ReferenceEvent(ReferenceEventType.OrganizationEditedByAdmin, organization, _currentContext)
+            {
+                EventRaisedByUser = _userService.GetUserName(User),
+                SalesAssistedTrialStarted = model.SalesAssistedTrialStarted,
+            });
+        }
+        catch (StripeException stripeException)
+        {
```
amorask-bitwarden, line 227: Don't love this by any stretch of the imagination, but the client is only using basic HTML 5 email validation, which something like "test@email" go through. I figured fixing the client-side validation was out of scope for this.

src/Admin/Controllers/OrganizationsController.cs
```diff
@@ -208,13 +212,23 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
 
-        await _organizationRepository.ReplaceAsync(organization);
-        await _applicationCacheService.UpsertOrganizationAbilityAsync(organization);
-        await _referenceEventService.RaiseEventAsync(new ReferenceEvent(ReferenceEventType.OrganizationEditedByAdmin, organization, _currentContext)
+        try
         {
-            EventRaisedByUser = _userService.GetUserName(User),
-            SalesAssistedTrialStarted = model.SalesAssistedTrialStarted,
-        });
+            await _stripeSyncService.UpdateCustomerEmailAddress(organization.GatewayCustomerId, organization.BillingEmail);
```
cturnbull-bitwarden, line 217: 👍 Nice, I wasn't aware that [IStripeSyncService](https://github.com/bitwarden/server/blob/90d600db9d55c8bf91dbfde286301f4097a71fc0/src/Core/Services/IStripeSyncService.cs#L3) already did this. Good find!

src/Admin/Controllers/OrganizationsController.cs
cturnbull-bitwarden: ⛏️ I think you should be fine to add validation for the email field if you're comfortable doing so, but no worries either way.

I'd suggest reducing the scope of the try/catch to just include line `217`. Since you're only catching `StripeException`, there's no danger that the existing code will trigger that exception, and since you're rethrowing the exception, execution will halt in the catch block preventing any further updates.

cturnbull-bitwarden (commented):
👍 Looks good!

cturnbull-bitwarden (approved):
👍 LGTM

PR #3287, opened 2023-09-21
[SM-896] restricting access to disabled orgs

cd-bitwarden:
## Type of change

```
- [ ] Bug fix
- [ x ] New feature development
- [ ] Other
```

## Objective
Disabled access to api endpoints that are not longer enabled

## Code changes
src/Core/Context/CurrentContextOrganization.cs - Added check for orgUser.Enabled
src/Identity/IdentityServer/ClientStore.cs - Added a check for org.Enabled
## Before you submit


Thomas-Avery (changes requested):
Looks good! 

We should update all the Secrets Manager endpoint `SmNotEnabled` integration tests to test for this new condition. 

Here is an example of one https://github.com/bitwarden/server/blob/90d600db9d55c8bf91dbfde286301f4097a71fc0/test/Api.IntegrationTest/SecretsManager/Controllers/SecretsControllerTests.cs#L58-L69

test/Api.IntegrationTest/SecretsManager/Controllers/AccessPoliciesControllerTests.cs
```diff
@@ -167,12 +168,13 @@ public async Task CreateProjectAccessPolicies_Success(PermissionType permissionT
...
 
     [Theory]
-    [InlineData(false, false)]
-    [InlineData(true, false)]
-    [InlineData(false, true)]
-    public async Task UpdateAccessPolicy_SmNotEnabled_NotFound(bool useSecrets, bool accessSecrets)
+    [InlineData(false, false, false)]
+    [InlineData(true, false, false)]
+    [InlineData(false, true, false)]
+    [InlineData(false, false, true)]
```
Thomas-Avery, line 174:
```suggestion
    [InlineData(false, false, false)]
    [InlineData(false, false, true)]
    [InlineData(false, true, false)]
    [InlineData(false, true, true)]
    [InlineData(true, false, false)]
    [InlineData(true, false, true)]
    [InlineData(true, true, false)]
```

A couple of more cases to test for each of these to cover everything.

Thomas-Avery (changes requested):

Thomas-Avery (approved):
Changes look good from SM side. 

We will want auth to review as well.

test/Api.IntegrationTest/SecretsManager/SecretsManagerOrganizationHelper.cs
```diff
@@ -25,13 +25,22 @@ public SecretsManagerOrganizationHelper(ApiApplicationFactory factory, string ow
         _ownerEmail = ownerEmail;
     }
 
-    public async Task<(Organization organization, OrganizationUser owner)> Initialize(bool useSecrets, bool ownerAccessSecrets)
+    public async Task<(Organization organization, OrganizationUser owner)> Initialize(bool useSecrets, bool ownerAccessSecrets, bool organizationEnabled = true)
```
JaredSnider-Bitwarden, line 28: Minor question: would it be safer to default this to false?

JaredSnider-Bitwarden (commented):
This is looking great! Only two minor questions.

test/Api.IntegrationTest/SecretsManager/Controllers/AccessPoliciesControllerTests.cs
```diff
@@ -56,12 +56,16 @@ private async Task LoginAsync(string email)
     }
 
     [Theory]
-    [InlineData(false, false)]
-    [InlineData(true, false)]
-    [InlineData(false, true)]
-    public async Task CreateProjectAccessPolicies_SmNotEnabled_NotFound(bool useSecrets, bool accessSecrets)
+    [InlineData(false, false, false)]
```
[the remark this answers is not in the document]
cd-bitwarden: For these tests we want one of the flags to be false because we are testing for the scenario where it fails because of one being false, so we don't want all true's passed in 😄

test/Api.IntegrationTest/SecretsManager/SecretsManagerOrganizationHelper.cs
cd-bitwarden: Spoke with Thomas Avery and we decided that it's probably best to explicitly pass in the value instead of defaulting to either false/true, thank you for pointing this out 😄

JaredSnider-Bitwarden (approved):
LGTM!

Thomas-Avery (approved):
Looks good, thank you!

PR #3297, opened 2023-09-25
[AC-1638] Disallow Secrets Manager for MSP-managed organizations

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We do not support MSP access to Secrets Manager, so we need to prevent:
* an org with SM being added to an MSP, or
* an MSP adding SM to one of their client organizations.

The internal Bitwarden Portal will not have this restriction, so that Customer Success can make case-by-case exceptions if required.

Clients changes: https://github.com/bitwarden/clients/pull/6392

## Code changes

- **ProviderService** - block a provider from adding an existing org if it has SM
- **AddSecretsManagerSubscriptionCommand** - block SM from being added to an existing org if it has an MSP
- **OrganizationService** - block a provider from creating a new org with SM

## Before you submit


amorask-bitwarden (approved):
LGTM!

cturnbull-bitwarden (approved):
🎉

PR #3300, opened 2023-09-26
[AC-1646] Rename LimitCollectionCdOwnerAdmin column

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Rename this to `LimitCollectionCreationDeletion` to be more human readable and accurate.

Also bumped the date of the SQL migration so that it runs again for local developers to re-create the column with the new name.

See https://github.com/bitwarden/clients/pull/6409 for clients changes.

## Code changes

## Before you submit


vincentsalucci (approved):
Looks good to me. One note: since we've changed the name of the column but didn't remove the old column - will those who've already run the original migration script have an extraneous column in their database? It won't matter because this hasn't been pushed live anywhere outside of the feature branch, of course and isn't referenced anywhere. Just wanted to clarify.

eliykat:
Yes good call, I'll share a SQL query for devs to clean up their own database. However, because it hasn't gone outside the feature branch I didn't think it was worth including in prod migrations for all time.

PR #3302, opened 2023-09-27
[AC-1637] Sanitize Business and Organization Names from html script injection prior to storing in db

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Other
```

## Objective
Sanitize user input values for Javascript script injection prior to storage.

## Code changes
In the API project, the user input values are automatically encoded using `HtmlEncodingStringConverter` and the same applies to output values. This converter needs to be configured on the desired properties using the data annotation `JsonConverter`.

In the Admin project, since its MVC, theres no automatic approach so we need to manually change the views to decode the values for them to be correctly displayed.

PR with a console app to update existing records [here](https://github.com/bitwarden/misc/pull/220).

- **src/Admin/Views/Organizations/Edit.cshtml:** Decoding values to display
- **src/Admin/Views/Organizations/Index.cshtml:** Decoding values to display
- **src/Admin/Views/Organizations/View.cshtml:** Decoding values to display
- **src/Admin/Views/Shared/_OrganizationForm.cshtml:** Decoding values to display
- **src/Api/Models/Request/Organizations/OrganizationCreateRequestModel.cs:** Configured Name and BusinessName properties to be converted using HtmlEncodingStringConverter on API create request
- **src/Api/Models/Request/Organizations/OrganizationUpdateRequestModel.cs:** Configured Name and BusinessName properties to be converted using HtmlEncodingStringConverter on API update request
- **src/Api/Models/Response/Organizations/OrganizationResponseModel.cs:** Configured Name and BusinessName properties to be converted using HtmlEncodingStringConverter on API response
- **src/Api/Models/Response/ProfileOrganizationResponseModel.cs:** Configured Name and BusinessName properties to be converted using HtmlEncodingStringConverter on API response
- **src/Core/Utilities/JsonHelpers.cs:** Custom JsonConverter that uses HttpUtility.HtmlEncode when deserializing and HttpUtility.HtmlDecode when serializing
- **test/Core.Test/Utilities/HtmlEncodingStringConverterTests.cs:** Unit tests

## Screenshots
The error messages related to the character limit for Organization `Name` and `Business Name` had to be modified. This was necessary because, due to the encoding of specific characters, the actual count depended on the inputted string.

Before:
image

After:
image

## Before you submit


vincentsalucci (dismissed):
Thanks for making the `HtmlDecode` changes to use `WebUtility` throughout!

justindbaur (approved):
This looks good to me too, you do have some conflicts though but they look like they are just renames so hopefully it merges cleanly and I'll be happy to re-approve.

vincentsalucci (approved):

cturnbull-bitwarden (approved):
Billing changes look good!

eliykat:
@r-tome last we spoke about this, you were having problems with the encoded value exceeding the max length and then getting truncated in the database. What was the resolution to that?

r-tome:
> @r-tome last we spoke about this, you were having problems with the encoded value exceeding the max length and then getting truncated in the database. What was the resolution to that?

@eliykat that was a mistake from me, that does not happen here. If the user input exceeds the char limit (either by having encoded chars or not) they will get an error message as shown above.

> Also, I see you've updated the error messages - is the intention here to _not_ tell the user the max length, given that it's the max length after encoding and therefore slightly misleading?

Exactly! The encoding is done on the `JsonSerializer` level (which happens before input validation) therefore the input length is validated with already encoded values. Specifying "50 char" limit on the error message could mislead the user because it would entirely depend on the number of chars that would be encoded.

eliykat (approved):
Sounds good to me, latest changes look good and renewing @vincentsalucci's previous approval.

bitwarden_license/src/Sso/Controllers/AccountController.cs
```diff
@@ -481,7 +483,7 @@ public async Task<IActionResult> Logout(string logoutId)
             if (orgUser.Status == OrganizationUserStatusType.Invited)
             {
                 // Org User is invited - they must manually accept the invite via email and authenticate with MP
-                throw new Exception(_i18nService.T("UserAlreadyInvited", email, organization.Name));
+                throw new Exception(_i18nService.T("UserAlreadyInvited", email, WebUtility.HtmlDecode(organization.Name)));
```
jlf0dev, line 486: Can we add this as a getter on the model?

r-tome: Excellent suggestion! I've added methods to output the Decoded values and used them everywhere where the value would be used for display purposes.

src/Core/AdminConsole/Entities/Organization.cs
```diff
@@ -100,19 +107,35 @@ public void SetNewId()
         }
     }
 
+    /// <summary>
+    /// Returns the name of the organization, decoded from HTML ready for display
+    /// </summary>
+    public string OrganizationName()
```
jlf0dev, line 113: nit: could we name this `DisplayName()`? Just so there's a little more differentiation from the `Name` property at a glance.

jlf0dev (dismissed):
Looks good! Had a minor suggestion but nothing blocking

jlf0dev (dismissed):
Awesome! Thanks for the changes, looks great!

src/Admin/Views/Providers/Organizations.cshtml
```diff
@@ -37,7 +36,7 @@
                         {
                             <tr>
                                 <td class="align-middle">
-                                    <a asp-controller="Organizations" asp-action="Edit" asp-route-id="@org.OrganizationId">@WebUtility.HtmlDecode(org.OrganizationName)</a>
+                                    <a asp-controller="Organizations" asp-action="Edit" asp-route-id="@org.OrganizationId">@org.DecodedOrganizationName()</a>
```
eliykat, line 39: 💭  (non-blocking) This could also follow the `DisplayName` naming scheme, rather than `Decoded` which is a bit less clear. `OrganizationDisplayName`.

eliykat (approved):
Those getters really tidied this up, nice work.

src/Admin/Views/Providers/Organizations.cshtml
r-tome: I forgot about that method, thanks for pointing it out. I have updated it.

eliykat (approved):

jlf0dev (approved):

cturnbull-bitwarden (approved):

r-tome:
@cturnbull-bitwarden @eliykat @jlf0dev I have merged main into this branch, hence the dismissed reviews

cturnbull-bitwarden (approved):

jlf0dev (approved):

eliykat (approved):

PR #3312, opened 2023-09-29
[AC-1669] Bug - Remove obsolete assignUserId from CollectionService.SaveAsync(...)

vincentsalucci:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Code changes
* **src/Api/Controllers/CollectionsController.cs**: Removed parameter
* **src/Core/Services/ICollectionService.cs**: Removed parameter
* **src/Core/Services/Implementations/CollectionService.cs**: Removed parameter. Added conditionals around making sure either a group or user have the `manage` access

## Before you submit

### DIFF
--- a/src/Core/Services/Implementations/CollectionService.cs
+++ b/src/Core/Services/Implementations/CollectionService.cs
@@ -61,7 +61,17 @@ public class CollectionService : ICollectionService
                 }
             }
 
-            await _collectionRepository.CreateAsync(collection, org.UseGroups ? groups : null, users);
+            var groupsList = groups?.ToList();
+            var usersList = users?.ToList();
+            var groupHasManageAccess = groupsList?.Any(g => g.Manage) ?? false;
+            var userHasManageAccess = usersList?.Any(u => u.Manage) ?? false;
+            if (!groupHasManageAccess && !userHasManageAccess)
+            {
+                throw new BadRequestException(
+                    "At least one User or Group must have Manage access to the newly created collection.");
+            }
+
+            await _collectionRepository.CreateAsync(collection, org.UseGroups ? groupsList : null, usersList);
             await _eventService.LogCollectionEventAsync(collection, Enums.EventType.Collection_Created);
             await _referenceEventService.RaiseEventAsync(new ReferenceEvent(ReferenceEventType.CollectionCreated, org, _currentContext));
         }
--- a/src/Core/Services/Implementations/CollectionService.cs
+++ b/src/Core/Services/Implementations/CollectionService.cs
@@ -49,6 +49,16 @@ public class CollectionService : ICollectionService
             throw new BadRequestException("Organization not found");
         }
 
+        var groupsList = groups?.ToList();
+        var usersList = users?.ToList();
+        var groupHasManageAccess = groupsList?.Any(g => g.Manage) ?? false;
+        var userHasManageAccess = usersList?.Any(u => u.Manage) ?? false;
+        if (!groupHasManageAccess && !userHasManageAccess)
+        {
+            throw new BadRequestException(
+                "At least one User or Group must have Manage access to the newly created collection.");
+        }
+
         if (collection.Id == default(Guid))
         {
             if (org.MaxCollections.HasValue)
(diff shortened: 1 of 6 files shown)


test/Core.Test/Services/CollectionServiceTests.cs
```diff
@@ -39,10 +44,15 @@ public async Task SaveAsync_DefaultIdWithUsers_CreatesCollectionInTheRepository(
         collection.Id = default;
         sutProvider.GetDependency<IOrganizationRepository>().GetByIdAsync(organization.Id).Returns(organization);
         var utcNow = DateTime.UtcNow;
+        // Grant the user manage access
+        var userList = users.ToList();
+        userList.First().Manage = true;
```
shane-melton, line 49: question/thought: While not a problem, the list of `users` here will sometimes already have `Manage` set to true due to how the Autofixture creates objects with random values for properties. If we want the test to pass with the minimum requirements, we should probably forcefully set all but one of the users to `Manage=false`.

Alternatively, we could add a customization for `CollectionAccessSelection` to default all of the `Manager`, `Readonly`, `HidePasswords` fields to false and remove the level of randomness from the tests completely.

Up to you though, I don't think its super critical, but I think would strengthen the tests overall.

test/Core.Test/Services/CollectionServiceTests.cs
```diff
@@ -57,26 +67,36 @@ await sutProvider.GetDependency<IEventService>().Received()
...
         var utcNow = DateTime.UtcNow;
+        userAccess.Manage = true;
+        List<CollectionAccessSelection> userList = new() { userAccess };
 
-        await sutProvider.Sut.SaveAsync(collection);
+        await sutProvider.Sut.SaveAsync(collection, null, userList);
 
-        await sutProvider.GetDependency<ICollectionRepository>().Received().ReplaceAsync(collection, null, null);
+        await sutProvider.GetDependency<ICollectionRepository>().Received().ReplaceAsync(collection,
+            Arg.Any<List<CollectionAccessSelection>>(),
```
shane-melton, line 98: question: Should this group list arg still be `null`?

src/Core/Services/Implementations/CollectionService.cs
```diff
@@ -41,14 +41,24 @@ public class CollectionService : ICollectionService
...
         }
 
+        var groupsList = groups?.ToList();
+        var usersList = users?.ToList();
+        var groupHasManageAccess = groupsList?.Any(g => g.Manage) ?? false;
+        var userHasManageAccess = usersList?.Any(u => u.Manage) ?? false;
+        if (!groupHasManageAccess && !userHasManageAccess)
+        {
+            throw new BadRequestException(
+                "At least one User or Group must have Manage access to the newly created collection.");
```
shane-melton, line 59: issue: The `SaveAsync()` method isn't always creating a new collection so we should drop that part of the exception message.

Side-note, we ended up using slightly different language for our error messages. I used the following on the client to match more closely with the access dialog language:
`At least one member or group must have can manage permission.`

Might be worthwhile to reach out to product to determine what language they want because I think yours could work just as well.

shane-melton (changes requested):
Just had a few comments on the test changes and the bad request exception message. Otherwise, looks good!

test/Core.Test/Services/CollectionServiceTests.cs
vincentsalucci: Added a `CollectionAccessSelection` fixture to guarantee the state of the `Manage` boolean. This helped to clean up the tests as a nice side effect.

test/Core.Test/Services/CollectionServiceTests.cs
vincentsalucci: I'm honestly not 100% sure: i also expected that passing in `null` would work but NSubstitute was tripping on the params with an error message similar to this:

`NSubstitute.Exceptions.AmbiguousArgumentsException
Cannot determine argument specifications to use. Please use specifications for all arguments of the same type.
Method signature:
    CreateAsync(Collection, IEnumerable<CollectionAccessSelection>, IEnumerable<CollectionAccessSelection>)
Method arguments (possible arg matchers are indicated with '*'):
    CreateAsync(Collection, *<null>*, *<null>*)`

shane-melton: Ah, well that is an annoying error 😦 

I think we still want to have the explicit check for `null` instead of `Any`, so we should be able to swap that with something like the following:

```Arg.Is<IEnumerable<CollectionAccessSelection>>(l => l == null)```

**Review by shane-melton — CHANGES_REQUESTED** [review]
Just had one more small naming nit/suggestion. Otherwise, looks good to go once we get that [null check](https://github.com/bitwarden/server/pull/3312#discussion_r1344857016) working in the test again

File: test/Core.Test/AutoFixture/CollectionAccessSelectionFixtures.cs [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public void Customize(IFixture fixture)
+    {
+        fixture.Customize<CollectionAccessSelection>(composer => composer
+            .With(o => o.Manage, Manage));
+    }
+}
+
+public class CollectionAccessSelectionAttribute : CustomizeAttribute
```
shane-melton, line 24: nit: Most of our customization attributes end with `Customize`, can we update this to follow that same pattern?

test/Core.Test/Services/CollectionServiceTests.cs
vincentsalucci: Updated!

shane-melton (approved):
Looks good to go!

PR #3313, opened 2023-09-29
Tools - Make Entities and Repositories nullable

audrey-jensen:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Turning on the nullable feature introduced in C# 8 improves compile time safety by ensuring code handles null values appropriately.

## Code changes
This PR enables nullability for entity and repository code owned by the tools group by adding `#nullable enable` pragmas to each file. Classes were updated to include nullable values where its dependencies were written in a null-aware fashion.

This PR also documents the types and methods in affected files, noting cases where null values should be anticipated. A brief description of all types and members was added for consistency's sake.

Nullability checks and documentation were added to all files in the commit. The following files contain additional changes: 
* **ReferenceEvents.cs:** - The nullable fields are all conditional values. Details in xmldoc comments.
* **IReferenceable.cs:** - `ReferenceData` is an optional value
* **Send.cs:** - `Password` is optional. `Data` and `Key` are nullable for database compatibility. Created PM-4128 to reconcile this.
* **SendData.cs:** `Notes` is optional. `Name` is required so I initialize it to `String.Empty` to satisfy the compiler.
* **SendFileData.cs:** `Id` is set by the service when a file is uploaded, and is `null` beforehand. `FileName` is read from a non-null field in a content header, and so is initialized to `String.Empty` to satisfy the compiler. This also includes a static import to improve readability of the number handling attribute.
* **SendTextData.cs:** - `Text` is optional. I was surprised by this.

No functional changes were applied in the following files:
* **ISendRepository.cs:** 
* **...Dapper/.../SendRepository.cs:** 
* **...EntityFramework/.../SendRepository.cs:**


src/Core/Tools/Entities/Send.cs
```diff
@@ -1,29 +1,123 @@
...
+    /// unless they're the creator.
+    /// </summary>
     public bool Disabled { get; set; }
+
+    /// <summary>
+    /// When this is false the creator's email address is not shown to the recipient.
+    /// </summary>
+    /// <remarks>
+    /// This should is null for compatibility with clients that don't set the field.
+    /// ::ADD TECH DEBT TICKET TO REMOVE NULLABILITY::
```
justindbaur, line 114: nit: Is this meant to be `This field is null`? Also, maybe missing a ticket ID. 

It might be nice to include in the remarks what true boolean value should be defaulted to in the case where this is null. I'm guessing `false` but I'll leave the exacts to you.

src/Core/Tools/Models/Data/SendFileData.cs
```diff
@@ -1,22 +1,54 @@
-﻿using System.Text.Json.Serialization;
+﻿#nullable enable
+
+using System.Text.Json.Serialization;
+using static System.Text.Json.Serialization.JsonNumberHandling;
```
justindbaur, line 4: I like it, especially for long enum names such as this one!

src/Core/Tools/Models/Data/SendFileData.cs
```diff
@@ -1,22 +1,54 @@
...
+    /// </remarks>
+    [JsonNumberHandling(WriteAsString | AllowReadingFromString)]
     public long Size { get; set; }
 
-    public string Id { get; set; }
-    public string FileName { get; set; }
+    /// <summary>
+    /// Uniquely identifies the file.
+    /// </summary>
+    public string? Id { get; set; }
```
justindbaur, line 41: I like the extra comments you have about this in the PR description, mind adding them here as well? Also, should this maybe be marked with `[DisallowNull]` while it being nullable is the right annotation I'm guessing we should never assign it to a non-null ID.

src/Core/Tools/Repositories/ISendRepository.cs
```diff
@@ -1,10 +1,36 @@
-﻿using Bit.Core.Repositories;
+﻿#nullable enable
+
+using Bit.Core.Repositories;
 using Bit.Core.Tools.Entities;
 
 namespace Bit.Core.Tools.Repositories;
 
+/// <summary>
+/// A repository for loading <see cref="Send"/>s.
```
justindbaur, line 9: nit: `A repository for interacting with <see cref="Send"/>s.` since `IRepository` does bring methods for creating and updating them.

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
```diff
@@ -7,12 +9,30 @@
 
 namespace Bit.Infrastructure.EntityFramework.Tools.Repositories;
 
+/// <summary>
+/// A repository for loading and creating <see cref="Send"/>s.
+/// </summary>
```
justindbaur, line 14: `<inheritdoc />` maybe?

src/Core/Tools/Models/Business/ReferenceEvent.cs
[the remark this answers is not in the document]
audrey-jensen: Tbh, I'd much rather create different event types, but that's even further afield of the scope of this PR than the documentation.

src/Core/Tools/Repositories/ISendRepository.cs
audrey-jensen: "interacting with sends" is what the whole system does, so I scoped the doc specifically to loads and saves.

src/Core/Tools/Entities/Send.cs
audrey-jensen: That comment was notes from a conversation that I forgot to update. Will clean it up presently! Thanks! :)

src/Core/Tools/Repositories/ISendRepository.cs
justindbaur: But since `ISendRepository` inherits from `IRepository` anything that implements it will be able to do things from the whole system, I feel like that's helpful to mention in its comment. I could maybe see an argument for that if we only did an explicit implementation of the interface but we very rarely (never?) do that in this repo.

src/Core/Tools/Models/Data/SendFileData.cs
justindbaur: Knowing why the field is null and what it is useful for indicating seems valuable in a comment without crossing domain boundaries. Maybe something like: `Expected to be <see langword="null" /> when the file has not been uploaded yet, and set to a unique identifier for a file provider once the file has been uploaded.`?

src/Core/Tools/Repositories/ISendRepository.cs
audrey-jensen: There's all kinds of ways to interact with sends once you have an instance, yes. But the only thing the repository does is load and save. There's special ways to do that (e.g. upsert/replace/create/etc), but at the end of the day the repo's scope is handling persistence of the send data. Anything beyond that, and you shouldn't use the repo to do it.

src/Core/Tools/Models/Data/SendFileData.cs
audrey-jensen: I've included guidance about its nullability in the latest commit. I also updated the file name member to distinguish when it should/should not contain an empty string value.

justindbaur (dismissed):
Looks fantastic, thank you for this!

djsmith85 (changes requested):
Love the added docs to this, especially for the ReferenceEvents

Just one small thing and this should be ready.

src/Core/Tools/Models/Business/ReferenceEvent.cs
```diff
@@ -26,48 +41,196 @@ public ReferenceEvent(ReferenceEventType type, IReferenceable source, ICurrentCo
...
     public bool? SendHasNotes { get; set; }
 
+    /// <summary>
+    /// Whether the created send has private notes.
+    /// </summary>
+    /// <value>
+    /// Should contain a value only on <see cref="ReferenceEventType.SendAccessed"/>
+    /// and <see cref="ReferenceEventType.SendCreated"/> events.
+    /// Otherwise the value should be <see langword="null"/>.
+    /// </value>
```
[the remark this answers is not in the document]
audrey-jensen: Good spot! Fixed.

djsmith85 (dismissed):

justindbaur (approved):

djsmith85 (approved):

PR #3328, opened 2023-10-05
Add server version compatibility check for Fido2Credentials on sharing with org

trmartin4:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

We need to ensure that a Login cipher that is updated from an old client doesn't overwrite a `Fido2Credential` that exists in the `Data` property from a newer client.

This should be done:
- When updating a cipher
- When moving a cipher to an org 
- When moving multiple ciphers to an org

## Code changes

**CiphersController.cs:** 
- Refactored `Fido2Credential` version check into a separate method for reuse on multiple methods.
- Added check to additional methods.
- Added missing check for vault item encryption on `PutShareMany()` as well.

## Before you submit


LRNcardozoWDF (approved):
LGTM!

PR #3334, opened 2023-10-09
[AC-1713] [Flexible collections] Add feature flags to server

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add feature flags for the server-side changes in Flexible Collections.

The goal is to allow an early merge of the feature branch into master. This means that additional changes can branch from master and have access to our WIP, however they'll have to be flagged where appropriate.

Database changes are not flagged; they just won't be used if the flags are off. Most flagging is around the `CollectionsController` authorization checks, which also required us to (temporarily) restore some of the old logic for when flags are OFF.

Both sets of authorization logic (old & new) have checks in them to make sure they're only being called when appropriate.

## Code changes

See my self-review below for commentary.

## Before you submit


test/Api.Test/Controllers/LegacyCollectionsControllerTests.cs
eliykat, line 1: This is `CollectionsControllerTests` as it exists on `master`, with a few minor adjustments. It's not new code. This way we keep unit tests covering the existing behaviour when the feature flag is OFF.

test/Core.Test/AutoFixture/FeatureServiceFixtures.cs
eliykat, line 1: This is a customization which intercepts the call to create the `IFeatureService` mock. We use it to arrange and return a mock with the specified feature flag turned on. The customization attribute must be applied to the test class, not individual test methods, but it's useful when an entire test suite assumes a certain flag state.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -232,16 +246,38 @@ public async Task Delete(Guid orgId, Guid id)
...
+            if (!result.Succeeded)
+            {
+                throw new NotFoundException();
+            }
+
+            await _deleteCollectionCommand.DeleteManyAsync(collections);
+            return;
+        }
+
+        // Old pre-flexible collections logic follows
```
eliykat, line 265: Unlike the other controller methods above, this logic diverged quite a bit between old & new. It was easiest to include both in full.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -272,6 +308,33 @@ private async Task<Collection> GetCollectionAsync(Guid id, Guid orgId)
...
+
+    private void DeprecatedPermissionsGuard()
+    {
+        if (FlexibleCollectionsIsEnabled())
+        {
+            throw new FeatureUnavailableException("Flexible Collections is ON when it should be OFF.");
+        }
+    }
+
+    [Obsolete("Pre-Flexible Collections logic. Will be replaced by CollectionsAuthorizationHandler.")]
```
eliykat, line 324: The rest of the code in this file is largely the existing logic restored from `master`, although with the `DeprecatedPermissionsGuard()` to ensure that it's not called accidentally. I also moved `DeleteAnyCollection` here rather than putting it back in `CurrentContext` where others might call it.

eliykat (commented):
Points of interest for the reviewer.

r-tome (commented):
Nice work! It looks great to me, I just left a question and a suggestion, but it'd be great to have someone else review it as well.
Nice work!

src/Api/Controllers/CollectionsController.cs
```diff
@@ -232,16 +246,38 @@ public async Task Delete(Guid orgId, Guid id)
 
     [HttpDelete("")]
     [HttpPost("delete")]
-    public async Task DeleteMany([FromBody] CollectionBulkDeleteRequestModel model)
+    public async Task DeleteMany(Guid orgId, [FromBody] CollectionBulkDeleteRequestModel model)
```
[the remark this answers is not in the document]
eliykat: That property has already been removed in the feature branch (on both clients and server), because it's not required for the auth handler interface. Rather than restore it for the legacy logic, and then have to remove it again later, we can just grab the orgId from the URL and use that instead.

shane-melton (approved):
This looks good to go to me and lines up with the changes made to client's repo!

PR #3339, opened 2023-10-12
[PM-4167] Add PRF attestation flow during passkey registration

coroiu:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Allow users to create and register all necessary keys to allow vault decryption using a passkey

## Before you submit


src/Core/Services/IUserService.cs
```diff
@@ -28,7 +28,7 @@ public interface IUserService
     Task<bool> DeleteWebAuthnKeyAsync(User user, int id);
     Task<bool> CompleteWebAuthRegistrationAsync(User user, int value, string name, AuthenticatorAttestationRawResponse attestationResponse);
     Task<CredentialCreateOptions> StartWebAuthnLoginRegistrationAsync(User user);
-    Task<bool> CompleteWebAuthLoginRegistrationAsync(User user, string name, CredentialCreateOptions options, AuthenticatorAttestationRawResponse attestationResponse);
+    Task<bool> CompleteWebAuthLoginRegistrationAsync(User user, string name, bool supportsPrf, string encryptedUserKey, string encryptedPublicKey, string encryptedPrivateKey, CredentialCreateOptions options, AuthenticatorAttestationRawResponse attestationResponse);
```
withinfocus, line 31: 🎨 Could these new parameters move to the end and be optional? For non-PRF default assumptions.

src/Core/Auth/Entities/WebAuthnCredential.cs
```diff
@@ -29,4 +33,18 @@ public void SetNewId()
...
+        if (SupportsPrf && EncryptedUserKey != null && EncryptedPrivateKey != null && EncryptedPublicKey != null)
+        {
+            return WebAuthnPrfStatus.Enabled;
+        }
+        else if (SupportsPrf)
+        {
+            return WebAuthnPrfStatus.Supported;
+        }
+
+        return WebAuthnPrfStatus.Unsupported;
```
jlf0dev, line 48: nit (non-blocking): just a tiny suggestion for readability
```suggestion
        if (!SupportsPrf)
        {
            return WebAuthnPrfStatus.Unsupported;
        }
        if (EncryptedUserKey != null && EncryptedPrivateKey != null && EncryptedPublicKey != null)
        {
            return WebAuthnPrfStatus.Enabled;
        }
        return WebAuthnPrfStatus.Supported;

```

src/Core/Services/IUserService.cs
jlf0dev: You could also consider making a data model for all these parameters and build it from the response model.

ike-kottlowski (approved):
Overall looks good. No notes.

jlf0dev (approved):

src/Core/Auth/Entities/WebAuthnCredential.cs
coroiu: Fixed!

src/Core/Services/IUserService.cs
coroiu: Fixed! I'll look into the model approach in my refactor task, where I'll convert these functions to commands!

PR #3343, opened 2023-10-13
Change DomainsResponseModel to use byte instead of GlobalEvalentDomainsType

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Generating API bindings from the Swagger Docs drags in all the `GlobalEvalentDomainsType` which may be frequently updated on the server. When the server returns a domain that is unknown to the client they will throw an error which is undesired.

This changes the `DomainsResponseModel` to use the inner `byte` value of the `GlobalEquivalentDomainsType` instead.

## Before you submit


eliykat (approved):
Is this something we're going to have to do for all Response models that contain enums?

Hinton:
@eliykat I hope not. It's something I'm looking into. In this case the enum doesn't exist on the other clients and only exists on server for convenience.

PR #3349, opened 2023-10-17
[AC-1593] Auto-Grant SM access to org owner when they add SM

amorask-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

[AC-1593](https://bitwarden.atlassian.net/browse/AC-1593)

This PR updates the flow of adding secrets manager to an organization by ensuring that the owner who adds secrets manager is updated to have access to it immediately.

## Code changes

* **[AddSecretsManagerSubscriptionCommand.cs](https://github.com/bitwarden/server/pull/3349/files#diff-1287d071ad8e7778d36fb065e743e1f9dccef783cfe850eb9349186948b9df72):** Take the `userId` of the user who requested SM addition. If it's the owner, give them SM access.
* **[OrganizationController.cs](https://github.com/bitwarden/server/pull/3349/files#diff-191777172f466a300e71192d0198f4d1b6be33c610e10b103e3c846d582bd712):** Pass the `userId` of the user who requested SM addition to the command handler.

## Before you submit

[AC-1593]: https://bitwarden.atlassian.net/browse/AC-1593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/OrganizationFeatures/OrganizationSubscriptions/AddSecretsManagerSubscriptionCommand.cs
eliykat, line 1: In my view, this command does exactly 1 thing: it adds SM to an organization. I don't think it should be responsible for granting access to individual users as well - I would just put that logic in the controller.

eliykat:
There is also the flow where an org adds SM as part of upgrading their plan. Do we need to provision user access as part of that flow as well?

amorask-bitwarden:
@eliykat I think I realized you're referring to upgrading from a Free Organization. I've removed the original commit and pushed a new commit that addresses both parts of your feedback: https://github.com/bitwarden/server/pull/3349/commits/4ebed973a613d9fa3959c8d645c724f9098ca200

Thanks for the review!

src/Api/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -766,4 +777,15 @@ public async Task<OrganizationSsoResponseModel> PostSso(Guid id, [FromBody] Orga
 
         return new OrganizationSsoResponseModel(organization, _globalSettings, ssoConfig);
     }
+
+    private async Task TryGrantOwnerAccessToSecretsManagerAsync(Guid organizationId, Guid userId)
+    {
+        var organizationUser = await _organizationUserRepository.GetByOrganizationAsync(organizationId, userId);
+
+        if (organizationUser is { Status: OrganizationUserStatusType.Confirmed, Type: OrganizationUserType.Owner })
```
eliykat, line 785: Each endpoint that uses this method already has its own permission check, so I don't think we need to check for Owner again here. The Jira ticket says that anyone who's managed to add SM should be given access, so I assume if they've reached this point in the code, we should just give it to them.

It also uses slightly different logic (`CurrentContext.EditSubscription` vs. `OrganizationUserType.Owner`) which risks diverging in the future.

It's true that we don't want to give providers access to SM, but they don't use these endpoints (granted that is not obvious) and in any case would return `null` from the `organizationUser` repo call.

eliykat (changes requested):
Yes you understood me exactly, nice work!

Two further requests:
1. Please add some tests in `OrganizationsControllerTests`
2. If you haven't already, please test these flows in the web client and make sure that the user automatically has access to SM without having to refresh the page to force a refresh of sync data. Looking at the code this is _probably_ OK but it would be good to check. (The risk I'm thinking of is that the server/database is updated without telling the client.)

amorask-bitwarden:
@eliykat Alright, all I needed to do in the client was refresh that identity token so we could send the updated claims. 

Implementation of your feedback is in this commit: https://github.com/bitwarden/server/pull/3349/commits/8538fd2a110677aa5db7a9b3f7ae36360fdc7c5e

The associated client PR is [here](https://github.com/bitwarden/clients/pull/6610) in case you want to see the screen recordings of both flows working.

Thanks!

eliykat (approved):
Great work! Thanks for double checking the flow as well.

cturnbull-bitwarden (approved):
🎉

eliykat:
FYI @amorask-bitwarden , please avoid force pushing (see the danger callout in https://contributing.bitwarden.com/contributing/pull-requests/)

We prefer merging `master` into the branch rather than rebasing the branch on top of `master`. Your commits will be squashed when you merge back into `master` anyway.

cturnbull-bitwarden (approved):

vincentsalucci (approved):
Approving on behalf of @eliykat !

PR #3355, opened 2023-10-18
Fix flaky UpdateSecretsManagerSubscriptionCommandTest

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

@Thomas-Avery reported that the `UpdateSubscriptionAsync_ThrowsBadRequestException_WhenOccupiedSeatsExceedNewSeatTotal` test sometimes fails for no reason, because an exception is expected but not thrown: https://github.com/bitwarden/server/actions/runs/6539171183/job/17756678563

My guess is that this occurs when autodata generates `organization.SmSeats = 7`. That would be equal to the new `SmSeat` value in the update model, so it won't be considered a change and [validation will be skipped by the sut](https://github.com/bitwarden/server/blob/45e883cc185093e4b9cf53a57f8a3c2058912948/src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs#L138-L142).

Fix this by expressly setting this value.

## Code changes

## Before you submit


Thomas-Avery (approved):
LGTM

PR #3360, opened 2023-10-19
[AC-1748] Flexible collections: deprecate Manage/Edit/Delete Assigned Collections custom permissions

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


src/Core/Services/Implementations/CollectionService.cs
```diff
@@ -53,21 +53,40 @@ public class CollectionService : ICollectionService
...
+                var user = usersList.FirstOrDefault(u => u.Id == orgUser.Id);
+                if (user != null)
+                {
+                    user.Manage = true;
+                }
+                else
+                {
+                    usersList.Add(new CollectionAccessSelection { Id = orgUser.Id, Manage = true });
+                }
+            }
```
eliykat, line 86: My bad - I see I mixed up `EditAny` and `EditAssigned` in my breakdown.

We're not deprecating `EditAny`, only `EditAssigned`.

So this logic should only set `user.Manage = true` if they're already being assigned access. In other words:
* check the `EditAssignedCollection` permission, not `Any`
* delete the `usersList.Add` block

Then I think it should work? But just double check for me.

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -915,14 +919,22 @@ public async Task DisableTwoFactorProviderAsync(Organization organization, TwoFa
...
+                        if (orgUser.GetPermissions()?.EditAnyCollection ?? false)
+                        {
+                            var orgCollections = await _collectionRepository.GetManyByOrganizationIdAsync(orgUser.OrganizationId);
+                            collections = orgCollections.Select(c => new CollectionAccessSelection { Id = c.Id, Manage = true });
+                        }
+                        else if (orgUser.AccessAll)
+                        {
+                            var orgCollections = await _collectionRepository.GetManyByOrganizationIdAsync(orgUser.OrganizationId);
+                            collections = orgCollections.Select(c => new CollectionAccessSelection { Id = c.Id, ReadOnly = true });
+                        }
```
eliykat, line 935: You'll probably need to revisit this per my comment above.

I think we'll need similar logic in a few places, can we extract it to like a static helper class or something? In any case - I definitely wouldn't want to nest this this far in this foreach block.

eliykat (changes requested):
@r-tome this is great! Glad to have you working on Flexible Collections stuff 😁 

One challenge here is limiting nesting when we have multiple feature-flagged implementations. Particularly on the controllers. It was less of a problem on the methods I touched because it was only the authorization call that changed, e.g.

```cs
var authorized = FlexibleCollectionsIsEnabled
  ? (await _authorizationService.AuthorizeAsync(User, collection, CollectionOperations.Delete)).Succeeded
  : await CanDeleteCollectionAsync(orgId, id);
```

But with your further changes to the controllers, it's most of the implementation that diverges, and we can get several levels deep in nested conditionals.

I suggest trying a pattern like

```cs
[HttpGet("")]
public async Task<ListResponseModel<CollectionResponseModel>> Get(Guid orgId)
{
  if (FlexibleCollectionsIsEnabled)
  {
    return await Get_FC(orgId);
  }
  
  // pre-flexible collections logic follows...
```

This keeps the logic totally separate, should make for a cleaner diff (mostly just new methods), and will also be easy to clean up when we remove the flags post-release.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -95,6 +126,28 @@ public async Task<CollectionAccessDetailsResponseModel> GetDetails(Guid orgId, G
...
     public async Task<ListResponseModel<CollectionAccessDetailsResponseModel>> GetManyWithDetails(Guid orgId)
     {
+        if (FlexibleCollectionsIsEnabled)
+        {
+            var readAllAuthorization = await _authorizationService.AuthorizeAsync(User, null, CollectionOperations.ReadAll(orgId));
+            if (readAllAuthorization.Succeeded)
+            {
+                var collections = await _collectionRepository.GetManyByOrganizationIdWithAccessAsync(orgId);
+                return new ListResponseModel<CollectionAccessDetailsResponseModel>(collections.Select(c =>
+                    new CollectionAccessDetailsResponseModel(c.Item1, c.Item2.Groups, c.Item2.Users)));
```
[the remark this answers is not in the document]
r-tome: Actually I'm not sure if we need it or not, removing it was an oversight. I'll add it back in and we can remove it later if we don't need it.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -69,22 +72,41 @@ public class CollectionAuthorizationHandler : BulkAuthorizationHandler<Collectio
...
     private async Task CanCreateAsync(AuthorizationHandlerContext context, CollectionOperationRequirement requirement,
-        CurrentContextOrganization org)
+        ICollection<Collection> targetCollections, CurrentContextOrganization org)
     {
+        if (targetCollections.Any(c => c.Id != default))
+        {
+            context.Fail();
+            return;
+        }
+
```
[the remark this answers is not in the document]
r-tome: I copied the logic from https://github.com/bitwarden/server/blob/728cd1c0b5eff982351cab7bab32a38caa66ef71/src/Api/Controllers/CollectionsController.cs#L254
But you're right, it has nothing to do with authorization

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -105,9 +127,63 @@ await _currentContext.ProviderUserForOrgAsync(org.Id))
...
+            !org.Permissions.ManageGroups &&
+            !org.Permissions.ManageUsers &&
+            !org.Permissions.EditAnyCollection &&
+            !org.Permissions.DeleteAnyCollection &&
+            !org.Permissions.AccessImportExport)
+        {
+            context.Fail();
+        }
+
+        context.Succeed(requirement);
```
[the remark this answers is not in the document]
r-tome: You're right, that was missing!

src/Core/Services/Implementations/CollectionService.cs
r-tome: I have revisited this, it should now be according to your comment above

src/Core/Services/Implementations/OrganizationService.cs
r-tome: This has been extracted into a private method

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,205 @@
...
+        await CheckCollectionPermissionsAsync(context, requirement, targetCollections, org, requireManagePermission: true);
+    }
+
+    private async Task CheckCollectionPermissionsAsync(
+        AuthorizationHandlerContext context,
+        IAuthorizationRequirement requirement,
+        ICollection<Collection> targetCollections,
+        CurrentContextOrganization org,
+        bool requireManagePermission)
+    {
```
eliykat, line 185: This method would be much simpler if it returned a `bool`. Then you don't have to pass in so many arguments, and the name would be easier to understand, e.g. `CanManageCollections`.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,205 @@
...
+            .ToHashSet();
+
+        // The acting user does not have permissions for all target collections, fail
+        if (targetCollections.Any(tc => !manageableCollectionIds.Contains(tc.Id)))
+        {
+            context.Fail();
+            return;
+        }
+
+        context.Succeed(requirement);
```
eliykat, line 203: I recommend inverting this logic to state the positive/affirmative case (assuming we're returning a bool):

```suggestion
  return targetCollections.All(tc => manageableCollectionIds.Contains(tc.Id));
```

eliykat (changes requested):
Partial review in case you want to action anything in the meantime!

src/Api/AdminConsole/Controllers/GroupsController.cs
```diff
@@ -185,4 +202,18 @@ public async Task Delete(string orgId, string id, string orgUserId)
 
         await _groupService.DeleteUserAsync(group, new Guid(orgUserId));
     }
+
+    private async Task<ListResponseModel<GroupDetailsResponseModel>> Get_FC(Guid orgId)
+    {
+        var authorized =
+            (await _authorizationService.AuthorizeAsync(User, null, GroupOperations.ReadAll(orgId))).Succeeded;
```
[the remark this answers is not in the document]
r-tome: I did not find that overload so I created a new extension method (`AuthorizationServiceExtensions`).

src/Api/Controllers/CollectionsController.cs
```diff
@@ -284,14 +337,32 @@ public async Task DeleteMany(Guid orgId, [FromBody] CollectionBulkDeleteRequestM
...
+    {
+        if (FlexibleCollectionsIsEnabled)
+        {
+            // New flexible collections logic
+            await DeleteUser_FC(id, orgUserId);
+        }
+
+        // Old pre-flexible collections logic follows
+        var collection = await GetCollectionAsync(id, orgId);
+        await _collectionService.DeleteUserAsync(collection, orgUserId);
```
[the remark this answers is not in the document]
r-tome: We do, the method `GetCollectionAsync` already checks the user's permissions to view all collections or that specific one.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -411,8 +480,150 @@ private async Task<bool> CanViewCollectionAsync(Guid orgId, Guid collectionId)
...
+    }
+
+    private async Task<ListResponseModel<CollectionAccessDetailsResponseModel>> GetManyWithDetails_FC(Guid orgId)
+    {
+        // We always need to know which collections the current user is assigned to
+        var assignedOrgCollections = await _collectionRepository
+            .GetManyByUserIdWithAccessAsync(_currentContext.UserId.Value, orgId);
+
+        var readAllAuthorized =
+            (await _authorizationService.AuthorizeAsync(User, null, CollectionOperations.ReadAll(orgId))).Succeeded;
```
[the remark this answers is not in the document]
r-tome: Thanks for pointing that out, I found a difference in their current authorization checks, `Get` allows users with `AccessImportExport` permission whereas `GetManyWithDetails` does not. Please re-check to be sure I didn't miss anything.

When I asked about different view models it was because some repository methods return classes that are different from the database model (usually from SQL views). So if we had an AuthorizationHandler using the database model as the type it wouldn't work, we would have to create another one.

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
```diff
@@ -0,0 +1,119 @@
...
+    [BitAutoData(OrganizationUserType.Custom, false, true, false, false, false, true)]
+    [BitAutoData(OrganizationUserType.Custom, false, false, true, false, false, true)]
+    [BitAutoData(OrganizationUserType.Custom, false, false, false, true, false, true)]
+    [BitAutoData(OrganizationUserType.Custom, false, false, false, false, true, true)]
+    [BitAutoData(OrganizationUserType.Custom, false, false, false, false, false, false)]
+    public async Task CanReadAllAccessAsync_ReturnsExpectedResult(
+        OrganizationUserType userType, bool editAnyCollection, bool deleteAnyCollection,
+        bool manageGroups, bool manageUsers, bool accessImportExport, bool expectedSuccess,
+        Guid userId, SutProvider<GroupAuthorizationHandler> sutProvider,
+        CurrentContextOrganization organization)
```
eliykat, line 34: This is really hard to read and maintain. I'd rather some test code repetition than a big matrix of bools like this.

I also think that an `expectedSuccess` bool is a bit of a smell. Better to have separate test methods for separate test cases.

To be clear, `BitAutoData` is still really good and useful, I just think this leans on it too heavily.

Suggestion for how you might improve this (you might think of a better way):
* `CanReadAllAccessAsync_WhenAdminOrOwner_Success`
* `CanReadAllAccessAsync_WhenUser_Fail`
* `CanReadAllAccessAsync_WhenCustomUserWithRequiredPermissions_Success`

src/Api/Controllers/CollectionsController.cs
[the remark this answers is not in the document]
eliykat: I'll dig into this more in my next review.

eliykat (changes requested):
Today's review. I'll find some time to dig further into the auth logic and verify that it's not being widened or narrowed unintentionally.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,222 @@
...
+            // Flexible collections is OFF, should not be using this handler
+            throw new FeatureUnavailableException("Flexible collections is OFF when it should be ON.");
+        }
+
+        // Establish pattern of authorization handler null checking passed resources
+        if (resources == null)
+        {
+            context.Fail();
+            return;
+        }
```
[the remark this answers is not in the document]
r-tome: Do we want to fail auth if the list has no results? For example if the organization has no collections that doesn't mean the user doesn't have access to them.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        var targetOrganizationId = requirement.OrganizationId != default
+            ? requirement.OrganizationId : resources.FirstOrDefault()?.OrganizationId ?? default;
+
+        if (targetOrganizationId == default)
+        {
+            context.Fail();
+            return;
+        }
```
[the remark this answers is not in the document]
r-tome: Fair enough but the point I made above can also be made here. If the list is empty then there is no `OrganizationId` to be obtained. Perhaps this check should only be made when the list has items?

src/Core/Services/Implementations/CollectionService.cs
```diff
@@ -53,21 +53,36 @@ public class CollectionService : ICollectionService
...
+            var organizationUsers = await _organizationUserRepository
+                .GetManyByOrganizationAsync(collection.OrganizationId, null);
+            foreach (var orgUser in organizationUsers.Where(ou => ou.GetPermissions()?.EditAssignedCollections ?? false))
+            {
+                var user = usersList.FirstOrDefault(u => u.Id == orgUser.Id);
+                if (user != null)
+                {
+                    user.Manage = true;
+                }
+            }
```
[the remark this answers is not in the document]
r-tome: Yes, I have moved those changes to this PR https://github.com/bitwarden/server/pull/3406

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
r-tome: I have split this test, it should be easier to read/maintain

eliykat (changes requested):
Reviewed Group and OrganizationUser changes. Looking really good! Continuing on the Collections changes.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -411,8 +485,150 @@ private async Task<bool> CanViewCollectionAsync(Guid orgId, Guid collectionId)
...
     private async Task<bool> ViewAtLeastOneCollectionAsync(Guid orgId)
     {
+        DeprecatedPermissionsGuard();
+
         return await _currentContext.ViewAllCollections(orgId) || await _currentContext.ViewAssignedCollections(orgId);
     }
+
+    private async Task<CollectionResponseModel> Get_vNext(Guid collectionId)
+    {
+        var collection = await _collectionRepository.GetByIdAsync(collectionId, _currentContext.UserId.Value);
```
eliykat, line 498: By specifying the userId here, it'll get by collectionId _and_ userId, which will only return collections the user is assigned to.

Because we have the permissions check in the auth handler, we don't need this kind of db-level permissions check. So you can just `GetByIdAsync(collectionId)` (which is an overload in the base repository class).

(this is kind of nice! the db call is now super simple.)

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
eliykat: I think these tests look really good now!

eliykat (changes requested):
WIP review on collections changes.

src/Api/Vault/AuthorizationHandlers/Groups/GroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+        switch (requirement)
+        {
+            case not null when requirement.Name == nameof(GroupOperations.ReadAll):
+                await CanReadAllAsync(context, requirement, org);
+                break;
+        }
+    }
+
+    private async Task CanReadAllAsync(AuthorizationHandlerContext context, GroupOperationRequirement requirement,
+        CurrentContextOrganization org)
```
eliykat, line 61: org should be nullable (and enable `#nullable`)

```suggestion
        CurrentContextOrganization? org)
```

src/Api/Vault/AuthorizationHandlers/OrganizationUsers/OrganizationUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        switch (requirement)
+        {
+            case not null when requirement.Name == nameof(OrganizationUserOperations.ReadAll):
+                await CanReadAllAsync(context, requirement, org);
+                break;
+        }
+    }
+
+    private async Task CanReadAllAsync(AuthorizationHandlerContext context, OrganizationUserOperationRequirement requirement,
+        CurrentContextOrganization org)
```
eliykat, line 60: Same here

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        // Allow provider users to create collections if they are a provider for the target organization
+        if (await _currentContext.ProviderUserForOrgAsync(_targetOrganizationId))
+        {
+            context.Succeed(requirement);
+        }
+    }
+
+    private async Task CanReadAsync(AuthorizationHandlerContext context, CollectionOperationRequirement requirement,
+        ICollection<Collection> targetCollections, CurrentContextOrganization org)
```
eliykat, line 114: After #3381, `CurrentContextOrganization` is nullable here. This is because the user might be a provider and have no org returned from `CurrentContext`. You'll need to allow for a potential null org, please see the other methods in this handler as an example. (or what you did in Group/OrgUser handlers)

eliykat (changes requested):

src/Api/Vault/AuthorizationHandlers/Groups/GroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            // Acting user is a member of the target organization, check permissions
+            if (org.Type is OrganizationUserType.Owner or OrganizationUserType.Admin ||
+                org.Permissions.ManageGroups ||
+                org.Permissions.ManageUsers ||
+                org.Permissions.EditAnyCollection ||
+                org.Permissions.DeleteAnyCollection ||
+                org.Permissions.CreateNewCollections ||
+                !org.LimitCollectionCreationDeletion ||
+                org.Permissions.AccessImportExport)
```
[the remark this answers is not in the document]
r-tome: You're right, this was added by mistake.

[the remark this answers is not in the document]
eliykat: resolved

src/Api/Vault/AuthorizationHandlers/Groups/GroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+                context.Succeed(requirement);
+            }
+        }
+        else
+        {
+            // Check if acting user is a provider user for the target organization
+            if (await _currentContext.ProviderUserForOrgAsync(requirement.OrganizationId))
+            {
+                context.Succeed(requirement);
+            }
```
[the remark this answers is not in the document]
eliykat: resolved

src/Api/Vault/AuthorizationHandlers/Groups/GroupAuthorizationHandler.cs
eliykat: resolved

src/Api/Vault/AuthorizationHandlers/OrganizationUsers/OrganizationUserAuthorizationHandler.cs
eliykat: resolved

src/Api/Controllers/CollectionsController.cs
eliykat: resolved

src/Api/Vault/AuthorizationHandlers/OrganizationUsers/OrganizationUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+        else
+        {
+            // Check if acting user is a provider user for the target organization
+            if (await _currentContext.ProviderUserForOrgAsync(requirement.OrganizationId))
+            {
+                context.Succeed(requirement);
+            }
+        }
```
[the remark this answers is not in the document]
eliykat: resolved

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionOperations.cs
eliykat, line 1: Not a big deal, but why did these need to be separated from the `ReadAll` requirement? ([commit](https://github.com/bitwarden/server/pull/3360/commits/0229165849eb67d53b32b8cb7b6b9053bfc4018b)) I thought it would've worked fine how it was.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -411,8 +485,150 @@ private async Task<bool> CanViewCollectionAsync(Guid orgId, Guid collectionId)
...
+        }
+
+        return new CollectionResponseModel(collection);
+    }
+
+    private async Task<CollectionAccessDetailsResponseModel> GetDetails_vNext(Guid id)
+    {
+        // New flexible collections logic
+        var (collection, access) = await _collectionRepository.GetByIdWithAccessAsync(id);
+        var authorized = (await _authorizationService.AuthorizeAsync(User, collection, BulkCollectionOperations.Read)).Succeeded;
```
eliykat, line 512: We could make this a bit more specific, `ReadWithAccess`, similar to how we separated `ReadAllWithAccess`.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+        // The acting user is a member of the target organization,
+        // ensure they have access for the collection being read
+        if (org is not null)
+        {
+            var canManageCollections = await HasCollectionAccessAsync(resources, org, requireManagePermission: false);
+            if (canManageCollections)
+            {
+                context.Succeed(requirement);
+                return;
+            }
```
eliykat, line 141: Variable name `canManageCollections` does not match argument `requireManagePermissions: false`. Which should it be?

Comparing to the current logic in both `Get` and `GetDetails`, I think this should be `requireManagePermission: true` for both. Both seem to require manager permissions (they generally trickle down to `currentContext.ViewAssignedCollections` -> `currentContext.EditAssignedCollections` which is a manager check). But please double check me.

eliykat (changes requested):
Did a bit more today

eliykat (changes requested):
Review completed, thanks again for your patience. 🎉

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionOperations.cs
r-tome: If `BulkCollectionAuthorizationHandler` and `CollectionAuthorizationHandler` shared the same `OperationAuthorizationRequirement` class then the first class registered on DI would handle all operation types. For example, `ReadAll` would be handled by `BulkCollectionAuthorizationHandler` when it should have been `CollectionAuthorizationHandler`.

src/Api/Controllers/CollectionsController.cs
r-tome: Yes, and its permissions are identical with `Read` so I've used the same method,

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
r-tome: You're right, I've also reviewed these and removed the `requireManagePermission` parameter as it was no longer used.

r-tome: I double checked and I think you're right. I got confused (you can understand why 😵‍💫)

test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
```diff
@@ -0,0 +1,843 @@
...
+    [BitAutoData(OrganizationUserType.Owner)]
+    public async Task CanCreateAsync_WhenAdminOrOwner_Success(
+        OrganizationUserType userType,
+        Guid userId, SutProvider<BulkCollectionAuthorizationHandler> sutProvider,
+        ICollection<Collection> collections,
+        CurrentContextOrganization organization)
+    {
+        organization.Type = userType;
+        organization.LimitCollectionCreationDeletion = true;
+        organization.Permissions = new Permissions();
```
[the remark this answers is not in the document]
r-tome: That is neat! Good spot

src/Api/Controllers/CollectionsController.cs
```diff
@@ -411,8 +485,150 @@ private async Task<bool> CanViewCollectionAsync(Guid orgId, Guid collectionId)
...
+                })
+            );
+        }
+
+        return new ListResponseModel<CollectionAccessDetailsResponseModel>(assignedOrgCollections.Select(c =>
+            new CollectionAccessDetailsResponseModel(c.Item1, c.Item2.Groups, c.Item2.Users)
+            {
+                Assigned = true // Mapping from assignedOrgCollections implies they're all assigned
+            })
+        );
```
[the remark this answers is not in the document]
r-tome: I added a new operation for this `CollectionOperations.ReadManyWithDetails`

src/Api/Controllers/CollectionsController.cs
```diff
@@ -411,8 +485,150 @@ private async Task<bool> CanViewCollectionAsync(Guid orgId, Guid collectionId)
...
+
+        var readAllAuthorized = (await _authorizationService.AuthorizeAsync(User, CollectionOperations.ReadAll(orgId))).Succeeded;
+        if (readAllAuthorized)
+        {
+            orgCollections = await _collectionRepository.GetManyByOrganizationIdAsync(orgId);
+        }
+        else
+        {
+            var collections = await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value);
+            orgCollections = collections.Where(c => c.OrganizationId == orgId);
```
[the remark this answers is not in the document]
eliykat: resolved

src/Api/Controllers/CollectionsController.cs
```diff
@@ -520,6 +520,13 @@ private async Task<CollectionAccessDetailsResponseModel> GetDetails_vNext(Guid i
 
     private async Task<ListResponseModel<CollectionAccessDetailsResponseModel>> GetManyWithDetails_vNext(Guid orgId)
     {
+        var readManyWithDetailsAuthorized =
+            (await _authorizationService.AuthorizeAsync(User, CollectionOperations.ReadManyWithDetails(orgId))).Succeeded;
+        if (!readManyWithDetailsAuthorized)
+        {
+            throw new NotFoundException();
+        }
```
eliykat, line 528: Isn't this expressing the same thing as ReadAllWithAccess, which we already check below? ("Details" just means Access.)

And anyway, I was thinking of this - we need to check whether the user has access to these specific collections, which we already have logic for:

```cs
        if (readAllAuthorized)
        {
             // snip
            // early return
        }

        var readAssignedAuthorized = await _authorizationService.AuthorizeAsync(User, assignedOrgCollections, BulkCollectionOperations.ReadWithAccess);
        // throw if not authorized

        return new ListResponseModel<CollectionAccessDetailsResponseModel>(assignedOrgCollections.Select(c =>
            new CollectionAccessDetailsResponseModel(c.Item1, c.Item2.Groups, c.Item2.Users)
            {
                Assigned = true // Mapping from assignedOrgCollections implies they're all assigned
            })
          );
```

src/Api/Controllers/CollectionsController.cs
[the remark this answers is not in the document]
eliykat: Left a comment on this in my latest review

src/Api/Controllers/CollectionsController.cs
r-tome: Yes! I was overcomplicating it, thank you!

test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
```diff
@@ -32,7 +32,6 @@ public class BulkCollectionAuthorizationHandlerTests
     {
         organization.Type = userType;
         organization.LimitCollectionCreationDeletion = true;
-        organization.Permissions = new Permissions();
```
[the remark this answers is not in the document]
eliykat: I think this is the safest way to go, thanks for following up on it!

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionOperations.cs
eliykat: Hmm, this isn't my understanding of how auth handlers should work, but it's not particularly important. It's a bit cleaner this way anyway.

eliykat (commented):
I think we've worked over this one well enough, I think it's looking good other than the unresolved comment [here](https://github.com/bitwarden/server/pull/3360#discussion_r1403913351). I wasn't sure from your latest DM whether you ended up testing that or whether regressions got in the way.

The situation to test would be:
* User role in the organization
* canManage permission in 1 collection
* `LimitCollectionCreationDeletion` set to true
* try to add another user to your collection - at the moment I think you'll hit an error?

While I think the feature flagging here is pretty clean, it's quite a large PR and does still touch some production code, so I'd like QA to regression test before merge.

r-tome:
Thanks for that summary of the situation to test, its really helpful!
I cannot test that yet because there is an issue on clients that prevents a user with Manage permission to assign another user to the collection (`this.manage` undefined issue).

I have revised `BulkOperations.Read` to not require Manage `permission` because a user with Edit permission still needs to access it. Or are we changing clients to not load the users tab on a collection?

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -1,176 +1,106 @@
...
         if (org is
         { Type: OrganizationUserType.Owner or OrganizationUserType.Admin } or
-        { Permissions.DeleteAnyCollection: true })
+        { Permissions.EditAnyCollection: true } or
+        { Permissions.DeleteAnyCollection: true } or
+        { Permissions.AccessImportExport: true })
         {
             context.Succeed(requirement);
             return;
         }
```
eliykat, line 79: We also need to add `ManageGroups` to the ReadAll permissions. See #3514 for context.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -1,176 +1,106 @@
...
+        // permission can always read a collection
         if (org is
         { Type: OrganizationUserType.Owner or OrganizationUserType.Admin } or
-        { Permissions.EditAnyCollection: true })
+        { Permissions.EditAnyCollection: true } or
+        { Permissions.DeleteAnyCollection: true })
         {
             context.Succeed(requirement);
             return;
         }
```
eliykat, line 100: And this is used by `ManageUsers` custom permission to get collections the user is associated with. You may need to add it here, or add it to your filtering logic in `GetManyWithDetails_vNext`.

shane-melton (commented):
Partial review for now, I will finish up more tomorrow. It looks pretty good, just a couple questions surrounding the new Read collection authorization requirement methods.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -0,0 +1,265 @@
...
+        { Permissions.EditAnyCollection: true } or
+        { Permissions.DeleteAnyCollection: true })
+        {
+            context.Succeed(requirement);
+            return;
+        }
+
+        // The acting user is a member of the target organization,
+        // ensure they have access for the collection being read
+        if (org is { LimitCollectionCreationDeletion: false })
```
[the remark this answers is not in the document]
eliykat: Agree. I don't quite follow the changes in https://github.com/bitwarden/server/pull/3360/commits/fd0d5d744f66fcd0e4c63b4bee26c148b0307a02

src/Api/Controllers/CollectionsController.cs
```diff
@@ -438,8 +512,164 @@ private async Task<bool> CanViewCollectionAsync(Guid orgId, Guid collectionId)
...
+            return new ListResponseModel<CollectionAccessDetailsResponseModel>(allOrgCollections.Select(c =>
+                new CollectionAccessDetailsResponseModel(c.Item1, c.Item2.Groups, c.Item2.Users)
+                {
+                    // Manually determine which collections they're assigned to
+                    Assigned = assignedOrgCollections.Any(ac => ac.Item1.Id == c.Item1.Id)
+                })
+            );
+        }
+
+        var readAssignedAuthorized = await _authorizationService.AuthorizeAsync(User, assignedOrgCollections, BulkCollectionOperations.ReadWithAccess);
```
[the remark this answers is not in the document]
r-tome: I did exactly that but to get the `Manage` property I also had to modify `ICollectionRepository.GetManyByUserIdWithAccessAsync` to return `CollectionDetails` which the sproc was returning already.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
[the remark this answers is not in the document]
r-tome: I agree with you both, I got confused with so many rules and tests 😅

r-tome:
@shane-melton and @eliykat thank you again for the reviews, all problems mentioned in the reviews have been addressed.

> The situation to test would be:
> User role in the organization
> canManage permission in 1 collection
> LimitCollectionCreationDeletion set to true
> try to add another user to your collection - at the moment I think you'll hit an error?

I tested this scenario and worked fine.

r-tome:
Btw, I agree this needs to be tested before merge so I've added the needs-qa tag

src/Api/Controllers/CollectionsController.cs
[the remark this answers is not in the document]
shane-melton: Sweet, glad the sproc was already grabbing that info!

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
shane-melton, line 1: thought: After looking at these Collection authorization handlers some more, their names don't seem to clearly distinguish their differences all that well and might be confusing to those unfamiliar.

`BulkCollectionAuthorizationHandler` makes sense to me and implies that it can handle authorizing multiple collections at a once.

However, seeing `CollectionAuthorizationHandler` would then be a bit confusing because just by the name I wouldn't expect it to do anything different that the Bulk handler couldn't already do.

I'm not suggesting we need to necessarily make a change for this PR, but I'm curious to what you both think on this?

My suggestion to potentially clear it up would be to rename the `CollectionAuthorizationHandler` to something like `OrganizationCollectionsAuthorizationHandler`. I would prefix it with Organization because I think it could be argued that the Organization is the resource being examined here and not the Collections; especially because the current handler has no resource and is pulling the organization from the auth requirement anyways. Then you could simply pass the Organization in as the resource and the operation would simply by `OrganizationCollectionOperations.ReadAll`.

This could also be used as a pattern for other similar Organization focused operations like read all Groups or Users.

shane-melton (approved):
Finished up my review and this looks good to go. Only saw two spots were the comments became out of date with the code and then just a general thought on how we name the authorization handlers. But, nothing that should be considered blocking from my perspective for this PR!

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
eliykat: Yes, this was a subject of some discussion when I first did the proof of concept for this pattern for Groups.

I originally used Organization as the resource as you suggest @shane-melton, simply because there is no single "Collection" object to use as the resource instead. I think the objection there was that it split the auth handlers. However, that ended up happening anyway, just because the logic is sufficiently different. Also, embedding the orgId in the requirement still feels like a hack, and a bad pattern to establish.

So I think your suggestion is a good one, it's just a lot more straightforward. I'd like to do a bit of a retro on the handler pattern more generally once FC is done to review how it's going and whether we should make any changes to our approach before extending the pattern.

eliykat (approved):

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
r-tome: > rename the `CollectionAuthorizationHandler` to something like `OrganizationCollectionsAuthorizationHandler`. I would prefix it with Organization because I think it could be argued that the Organization is the resource being examined here and not the Collections

Absolutely! I struggle with naming and during the development of this branch both AuthorizationHandlers went through multiple versions. That's why some comments also require updating. I'm interested in exploring ways to eliminate OrgId from the requirement. Let's talk about this in one of our upcoming FC dev syncs.

PR #3370, opened 2023-10-23
Add Bulk Move to Project Endpoint

justindbaur:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add new endpoint for updating the projects of many secrets to the same project.

Clients PR: bitwarden/clients#6665

## Code changes

* **bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs:** This interacts with the virtual table `ProjectSecret` to avoid making large changes in the code base or to make this update one row at a time. It deletes all current relationships for the given secrets and then creates new relationships for them based on the supplied project ids.
* **src/Api/SecretsManager/Controllers/SecretsController.cs:** Controller method for validating that the current user has the permissions to do the specified action.

## Before you submit

### DIFF
--- a/bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
+++ b/bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
@@ -313,6 +313,34 @@ public class SecretRepository : Repository<Core.SecretsManager.Entities.Secret, 
         await dbContext.SaveChangesAsync();
     }
 
+    public async Task MoveSecretsAsync(IEnumerable<Core.SecretsManager.Entities.Secret> secrets, IEnumerable<Guid> projectIds)
+    {
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+
+        var projectSecretsSet = dbContext.Set<Dictionary<string, object>>("ProjectSecret");
+
+        using var transaction = await dbContext.Database.BeginTransactionAsync();
+
+        var secretIds = secrets.Select(s => s.Id).ToList();
+
+        var projectSecrets = secretIds.SelectMany(s => projectIds.Select(p => new Dictionary<string, object>
+        {
+            ["ProjectsId"] = p,
+            ["SecretsId"] = s
+        }));
+
+        await projectSecretsSet
+            .Where(ps => secretIds.Contains((Guid)ps["SecretsId"]))
+            .ExecuteDeleteAsync();
+
+        projectSecretsSet.AddRange(projectSecrets);
+
+        await dbContext.SaveChangesAsync();
+
+        await transaction.CommitAsync();
+    }
+
     private IQueryable<SecretPermissionDetails> SecretToPermissionDetails(IQueryable<Secret> query, Guid userId, AccessClientType accessType)
     {
         var secrets = accessType switch
--- a/src/Api/SecretsManager/Controllers/SecretsController.cs
+++ b/src/Api/SecretsManager/Controllers/SecretsController.cs
@@ -33,6 +33,7 @@ public class SecretsController : Controller
     private readonly IEventService _eventService;
     private readonly IReferenceEventService _referenceEventService;
     private readonly IAuthorizationService _authorizationService;
+    private readonly IMoveSecretsCommand _moveSecretsCommand;
 
     public SecretsController(
         ICurrentContext currentContext,
@@ -45,7 +46,8 @@ public class SecretsController : Controller
         IUserService userService,
         IEventService eventService,
         IReferenceEventService referenceEventService,
-        IAuthorizationService authorizationService)
+        IAuthorizationService authorizationService,
+        IMoveSecretsCommand moveSecretsCommand)
     {
         _currentContext = currentContext;
         _projectRepository = projectRepository;
@@ -58,7 +60,7 @@ public class SecretsController : Controller
         _eventService = eventService;
         _referenceEventService = referenceEventService;
         _authorizationService = authorizationService;
-
+        _moveSecretsCommand = moveSecretsCommand;
     }
 
     [HttpGet("organizations/{organizationId}/secrets")]
@@ -245,5 +247,37 @@ public class SecretsController : Controller
 
         var responses = secrets.Select(s => new BaseSecretResponseModel(s));
         return new ListResponseModel<BaseSecretResponseModel>(responses);
+    }
+
+    [HttpPost("secrets/{organizationId}/move")]
+    public async Task BulkMoveToProjectAsync(Guid organizationId, [FromBody] BulkMoveToProjectsRequestModel request)
+    {
+        var secrets = (await _secretRepository.GetManyByIds(request.Secrets)).ToList();
+        if (!secrets.Any() || secrets.Count != request.Secrets.Count)
+        {
+            throw new NotFoundException();
+        }
+
+        if (secrets.Any(secret => secret.OrganizationId != organizationId) ||
+            !_currentContext.AccessSecretsManager(organizationId))
+        {
+            throw new NotFoundException();
+        }
+
+        foreach (var secret in secrets)
+        {
+            var authorizationResult = await _authorizationService.AuthorizeAsync(User, secret, SecretOperations.Update);
+            if (!authorizationResult.Succeeded)
+            {
+                throw new NotFoundException();
+            }
+        }
+
+        if (request.Projects.Count != 1)
+        {
+            throw new BadRequestException("You are only allowed to assign a single project.");
+        }
+
+        await _moveSecretsCommand.MoveSecretsAsync(secrets, request.Projects);
     }
 }
--- a/bitwarden_license/src/Commercial.Core/SecretsManager/Commands/Secrets/MoveSecretsCommand.cs
+++ b/bitwarden_license/src/Commercial.Core/SecretsManager/Commands/Secrets/MoveSecretsCommand.cs
@@ -1,0 +1,20 @@
+using Bit.Core.SecretsManager.Commands.Secrets.Interfaces;
+using Bit.Core.SecretsManager.Entities;
+using Bit.Core.SecretsManager.Repositories;
+
+namespace Bit.Commercial.Core.SecretsManager.Commands.Secrets;
+
+public class MoveSecretsCommand : IMoveSecretsCommand
+{
+    private readonly ISecretRepository _secretRepository;
+
+    public MoveSecretsCommand(ISecretRepository secretRepository)
+    {
+        _secretRepository = secretRepository;
+    }
+
+    public async Task MoveSecretsAsync(IEnumerable<Secret> secrets, IEnumerable<Guid> projects)
+    {
+        await _secretRepository.MoveSecretsAsync(secrets, projects);
+    }
+}
--- a/bitwarden_license/src/Commercial.Core/SecretsManager/Commands/Secrets/MoveSecretsCommand.cs
+++ b/bitwarden_license/src/Commercial.Core/SecretsManager/Commands/Secrets/MoveSecretsCommand.cs
@@ -1,4 +1,4 @@
-using Bit.Core.SecretsManager.Commands.Secrets.Interfaces;
+﻿using Bit.Core.SecretsManager.Commands.Secrets.Interfaces;
 using Bit.Core.SecretsManager.Entities;
 using Bit.Core.SecretsManager.Repositories;
 
--- a/bitwarden_license/src/Commercial.Core/SecretsManager/SecretsManagerCollectionExtensions.cs
+++ b/bitwarden_license/src/Commercial.Core/SecretsManager/SecretsManagerCollectionExtensions.cs
@@ -56,5 +56,6 @@ public static class SecretsManagerCollectionExtensions
         services.AddScoped<IImportCommand, ImportCommand>();
         services.AddScoped<IEmptyTrashCommand, EmptyTrashCommand>();
         services.AddScoped<IRestoreTrashCommand, RestoreTrashCommand>();
+        services.AddScoped<IMoveSecretsCommand, MoveSecretsCommand>();
     }
 }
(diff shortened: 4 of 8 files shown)


src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -246,4 +248,36 @@ public async Task<ListResponseModel<BulkDeleteResponseModel>> BulkDeleteAsync([F
...
+            if (!authorizationResult.Succeeded)
+            {
+                throw new NotFoundException();
+            }
+        }
+
+        if (request.Projects.Count != 1)
+        {
+            throw new BadRequestException("You are only allowed to assign a single project.");
+        }
```
Thomas-Avery, line 279: What are your thoughts on just changing the request model to enforce this restriction? 

```
public class BulkMoveToProjectsRequestModel
{
    [Required]
    public IReadOnlyList<Guid> Secrets { get; set; }

    [Required]
    public Guid Project { get; set; }
}
```

I think it makes the intention clearer from an API contract standpoint.

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
```diff
@@ -313,6 +313,34 @@ public async Task EmptyTrash(DateTime currentDate, uint deleteAfterThisNumberOfD
...
+        await projectSecretsSet
+            .Where(ps => secretIds.Contains((Guid)ps["SecretsId"]))
+            .ExecuteDeleteAsync();
+
+        projectSecretsSet.AddRange(projectSecrets);
+
+        await dbContext.SaveChangesAsync();
+
+        await transaction.CommitAsync();
+    }
```
Thomas-Avery, line 342: This is cool, I didn't know you could do this.

Is the warning [here](https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many#basic-many-to-many) applicable to what we're doing? 

```
Currently, EF Core uses Dictionary<string, object> to represent join entity instances for which no .NET class has been configured. 
However, to improve performance, a different type may be used in a future EF Core release. 
Do not depend on the join type being Dictionary<string, object> unless this has been explicitly configured.
```

Another option is, we could create a class for the join entity and add that to our EF config.

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
```diff
@@ -313,6 +313,34 @@ public async Task EmptyTrash(DateTime currentDate, uint deleteAfterThisNumberOfD
...
     }
 
+    public async Task MoveSecretsAsync(IEnumerable<Core.SecretsManager.Entities.Secret> secrets, IEnumerable<Guid> projectIds)
+    {
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+
+        var projectSecretsSet = dbContext.Set<Dictionary<string, object>>("ProjectSecret");
+
+        using var transaction = await dbContext.Database.BeginTransactionAsync();
```
Thomas-Avery, line 323:
```suggestion
        await using var transaction = await dbContext.Database.BeginTransactionAsync();
```
My IDE is suggesting this since it implements `IAsyncDisposable`

Thomas-Avery (changes requested):
Thank you for working on this! 

The team appreciates it. 😃 

This is looking good, just a few questions to address.

src/Api/SecretsManager/Controllers/SecretsController.cs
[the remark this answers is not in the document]
justindbaur: This code was ripped of from https://github.com/bitwarden/server/blob/75cae907e82fa2b2afc0ce7517a18d018ec08daa/src/Api/SecretsManager/Controllers/SecretsManagerPortingController.cs#L70-L73

From an outsiders perspective it was built this way because there is a chance a secret might be in multiple projects down the road. If that is not the case I'd be totally fine with making it singular but if that is indeed the plan then I'd probably want to keep this shape and then remove the exception when that becomes possible.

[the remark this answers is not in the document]
Thomas-Avery: For context, at one point we were implementing having on secret in multiple projects. That is why you see things in the code base as it is now.

A decision was made to only support one project for now. 

In the future, it is possible multiple projects will be implemented but for now it isn't on our roadmap. 

For things like this, I typically follow the YAGNI mantra to avoid the very confusion we're currently experiencing.

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
justindbaur: Good catch!

src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -246,4 +248,36 @@ public async Task<ListResponseModel<BulkDeleteResponseModel>> BulkDeleteAsync([F
...
+
+        foreach (var secret in secrets)
+        {
+            var authorizationResult = await _authorizationService.AuthorizeAsync(User, secret, SecretOperations.Update);
+            if (!authorizationResult.Succeeded)
+            {
+                throw new NotFoundException();
+            }
+        }
```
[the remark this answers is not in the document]
justindbaur: I did something sort of like this, I created a `BulkSecretAuthorizationHandler` that just validates that all the selected secrets are editable. I wrote a repository method a lot like you sketched out above. But I did instead call `AuthorizeAsync` twice. Once for the list of secrets and once for the target project so I could reuse the logic in the Project authorization handler. Let me know what you think.

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
justindbaur: I think it would have been applicable. I did decide to go the route of defining it as an entity. I did initially try that but it messed with the model a lot so I ditched it but I figured out how to do it better and there are only very minor changes to the model. Essentially now it just have a namespace name instead of an auto generated one.

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
```diff
@@ -274,35 +274,58 @@ public async Task<(bool Read, bool Write)> AccessToSecretAsync(Guid id, Guid use
         var secret = dbContext.Secret
             .Where(s => s.Id == id);
 
-        var query = accessType switch
+        var query = BuildSecretAccessQuery(secret, userId, accessType);
```
Thomas-Avery, line 277: Nice! Thanks for DRYing this up.

src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -246,4 +248,38 @@ public async Task<ListResponseModel<BulkDeleteResponseModel>> BulkDeleteAsync([F
...
+    public async Task BulkMoveToProjectAsync(Guid organizationId, [FromBody] BulkMoveToProjectsRequestModel request)
+    {
+        var secrets = (await _secretRepository.GetManyByIds(request.Secrets)).ToList();
+        if (!secrets.Any() || secrets.Count != request.Secrets.Count)
+        {
+            throw new NotFoundException();
+        }
+
+        if (secrets.Any(secret => secret.OrganizationId != organizationId) ||
+            !_currentContext.AccessSecretsManager(organizationId))
```
Thomas-Avery, line 262: It looks like we are already checking `!_currentContext.AccessSecretsManager(organizationId)` in 
` _authorizationService.AuthorizeAsync(
            User, secrets.AsReadOnly(), BulkSecretOperations.Update);`
            
 Should we remove it here since it is an authz concern?

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
```diff
@@ -180,6 +181,8 @@ protected override void OnModelCreating(ModelBuilder builder)
         eOrganizationDomain.ToTable(nameof(OrganizationDomain));
         aWebAuthnCredential.ToTable(nameof(WebAuthnCredential));
 
+        builder.Entity<ProjectSecret>(b => b.ToTable(nameof(ProjectSecret)));
```
Thomas-Avery, line 184: Thoughts on moving this into `src/Infrastructure.EntityFramework/SecretsManager/Configurations` into its own `IEntityTypeConfiguration` file?

For this case, it isn't much configuration, but I find it easier for managing the team's specific configurations.

bitwarden_license/src/Commercial.Core/SecretsManager/AuthorizationHandlers/Secrets/BulkSecretAuthorizationHandler.cs
Thomas-Avery, line 1: Can we write unit tests for this file? 

For examples see `bitwarden_license/test/Commercial.Core.Test/SecretsManager/AuthorizationHandlers`.

test/Api.IntegrationTest/SecretsManager/Controllers/SecretsControllerTests.cs
```diff
@@ -801,6 +801,70 @@ public async Task GetSecretsByIds_Success(PermissionType permissionType)
...
+
+        var response = await _client.PostAsJsonAsync($"/organizations/{org.Id}/secrets/move", new BulkMoveToProjectsRequestModel
+        {
+            Secrets = secretIds,
+            Project = otherProject.Id,
+        });
+
+        Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
+    }
+
```
Thomas-Avery, line 867: Nice, thanks for adding integration tests. 

Can we add one for checking SM access? 

Example here https://github.com/bitwarden/server/blob/061253e4285a9320a34189a851c80b788e4b1e09/test/Api.IntegrationTest/SecretsManager/Controllers/SecretsControllerTests.cs#L58C1-L73

src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -246,4 +248,38 @@ public async Task<ListResponseModel<BulkDeleteResponseModel>> BulkDeleteAsync([F
         var responses = secrets.Select(s => new BaseSecretResponseModel(s));
         return new ListResponseModel<BaseSecretResponseModel>(responses);
     }
+
+    [HttpPost("organizations/{organizationId}/secrets/move")]
+    public async Task BulkMoveToProjectAsync(Guid organizationId, [FromBody] BulkMoveToProjectsRequestModel request)
```
Thomas-Avery, line 253: Can we write some unit tests for this controller logic/validation? 

Some examples in `test/Api.Test/SecretsManager/Controllers/SecretsControllerTests.cs`

src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -246,4 +248,38 @@ public async Task<ListResponseModel<BulkDeleteResponseModel>> BulkDeleteAsync([F
...
+
+        var authorizationResult = await _authorizationService.AuthorizeAsync(
+            User, secrets.AsReadOnly(), BulkSecretOperations.Update);
+
+        if (!authorizationResult.Succeeded)
+        {
+            throw new NotFoundException();
+        }
+
+        var project = await _projectRepository.GetByIdAsync(request.Project);
```
Thomas-Avery, line 275: Can we move this project fetch up with the secrets fetching so we can check that both exist and both are in the same organization first?

Thomas-Avery (changes requested):

src/Api/SecretsManager/Controllers/SecretsController.cs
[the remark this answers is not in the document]
Thomas-Avery: Looked good and sounds reasonable.

src/Api/SecretsManager/Controllers/SecretsController.cs
justindbaur: Added the one to check for SM access, let me know if you think it's missing any still.

src/Api/SecretsManager/Controllers/SecretsController.cs
justindbaur: I moved the project fetch and validation before we do anything with secrets this way we only get a single item from the database if it's wrong and fail the request vs getting `n` number of secrets and validating those before getting to the project. Should make the load on our DB be lower in the worst case, is that what you meant by this or did I misunderstand?

util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
```diff
@@ -533,6 +533,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("varchar(50)");
 
                     b.HasKey("Id")
+                        .HasName("PK_Grant")
```
justindbaur, line 536: @withinfocus Any idea why my migrations are adding this?

withinfocus: I do not. Unsure why my changes didn't already do this but my guess is it has something to do with Pomelo's workaround with key modifications not immediately representing where the named PK is.

Thomas-Avery (dismissed):
Looks good, thank you for the work on this. 

Ideally, we would have some unit tests at `test/Api.Test/SecretsManager/Controllers/SecretsControllerTests.cs` to cover what codecov is warning on. 
I'll leave that up to you, as I think the integration tests cover most of our concerns.

Thomas-Avery (changes requested):
Looks like the EF migrations will need to be refreshed.

Other than that, LGTM

rkac-bw (approved):
Looks good to me

PR #3372, opened 2023-10-24
[AC-1330] [AC-1815] [Server] Deprecate access control indicator - UserCipherDetails

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Deprecating `AccessAll` logic has a large impact on our db sprocs because it's used frequently to check who's allowed to access a cipher. By my count, it requires updates to 26 separate sprocs. Because that's a lot, I'm going to chunk it up into separate PRs.

This PR updates the `UserCipherDetails` function and the sprocs that depend on it.

## Code changes

The changes are broadly as follows:

**1. SQL sprocs**
* copy each sproc to a `_V2` variant
* update the `_V2` variant with the new logic
* update Dapper repository to call the old sproc or the new sproc depending on feature flag value
* existing sprocs will be dropped 1-2 releases after the feature flags are removed, per EDD requirements

**2. EF queries**
* code is a bit more flexible than SQL query files, so this might be a bit more case-by-case. In this PR I duplicated the `Run` method of the `UserCipherDetailsQuery` into a `Run_Legacy` and `Run_V2`, and made the public `Run` method contain the conditional logic to switch between them

**3. Repository interface update**
* the repository implementations need to know the feature flag state. Per the discussion below, each caller evaluates the flag state and passes a boolean in to the repository.

## Before you submit


src/Events/Controllers/CollectController.cs
```diff
@@ -70,7 +76,8 @@ public async Task<IActionResult> Post([FromBody] IEnumerable<EventModel> model)
                     else
                     {
                         cipher = await _cipherRepository.GetByIdAsync(eventModel.CipherId.Value,
-                           _currentContext.UserId.Value);
+                           _currentContext.UserId.Value,
+                           _currentContext);
```
withinfocus, line 80: (grabbed this one at random) ⚠️ Sending the entire context into repositories seems like an overreach to me.

src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs
```diff
@@ -14,20 +17,31 @@ namespace Bit.Infrastructure.Dapper.Vault.Repositories;
...
+    {
+        _featureService = featureService;
+    }
 
     public CipherRepository(string connectionString, string readOnlyConnectionString)
         : base(connectionString, readOnlyConnectionString)
     { }
 
-    public async Task<CipherDetails> GetByIdAsync(Guid id, Guid userId)
+    public async Task<CipherDetails> GetByIdAsync(Guid id, Guid userId, ICurrentContext currentContext)
```
withinfocus, line 35: ⚠️ To further my other comment, we should pass results of flag evaluation into the repository from the service layer. The repository merely owns the decision of what database entity to use.

src/Infrastructure.EntityFramework/Repositories/Queries/UserCipherDetailsQuery.cs
```diff
@@ -78,6 +88,71 @@ public virtual IQueryable<CipherDetails> Run(DatabaseContext dbContext)
         return union;
     }
 
+    private IQueryable<CipherDetails> Run_V2(DatabaseContext dbContext)
```
withinfocus, line 91: 💭 This is a lot of EF code ... will be valuable to debug its resulting SQL. I also support EF calling stored procedures to get known and measured performance.

src/Infrastructure.EntityFramework/Repositories/Queries/UserCipherDetailsQuery.cs
```diff
@@ -8,11 +8,21 @@ namespace Bit.Infrastructure.EntityFramework.Repositories.Queries;
...
     }
+
     public virtual IQueryable<CipherDetails> Run(DatabaseContext dbContext)
+    {
+        return _useFlexibleCollections
+            ? Run_V2(dbContext)
+            : Run_Legacy(dbContext);
+    }
+
+    private IQueryable<CipherDetails> Run_Legacy(DatabaseContext dbContext)
```
withinfocus, line 25: ⛏️ The "legacy" seems unnecessary to me as it's the V1. With thorough cleanup this returns and the "V2" is dropped anyway. Per some past conversations this is why I like tech debt tasks for cleanup and restoring naming, so this could be "vNext" even. Don't need to dwell too much on naming though.

withinfocus:
Left a few comments that might deserve light discussion. My biggest one is keeping the concept of flag providers (our service) out of the data layer.

src/Infrastructure.EntityFramework/Repositories/Queries/UserCipherDetailsQuery.cs
eliykat: I'm warming to the vNext idea, just so we don't have ever-incrementing version numbers.

eliykat:
@withinfocus thanks for the review. So you would have, for example

```cs
    // CiphersController
    private async Task<CipherDetails> GetByIdAsync(Guid cipherId, Guid userId)
    {
        var useFlexibleCollections = await _featureService.IsEnabled(FeatureFlagKeys.FlexibleCollections, _currentContext);
        return await _cipherRepository.GetByIdAsync(cipherId, userId, useFlexibleCollections);
    }
```

and then

```cs
    // CipherRepository
    public async Task<CipherDetails> GetByIdAsync(Guid id, Guid userId, bool useFlexibleCollections)
```

?

That was my first approach, however that requires every calling location to inject `FeatureService` and repeat the `var useFlexibleCollections = await _featureService.IsEnabled(FeatureFlagKeys.FlexibleCollections, _currentContext);` logic. Which is not the end of the world, but I thought it had less repetition and less code impact to encapsulate it in the repository.

On the other hand, I do understand that it's contaminating the data layer to give it knowledge of currentContext and feature flags. I'll go ahead and make that change.

src/Infrastructure.EntityFramework/Repositories/Queries/UserCipherDetailsQuery.cs
eliykat: Changed to `VNext` and `VCurrent` (??) or would you go back the legacy wording in this pattern?

withinfocus: `Run` and `Run_vNext` feel fine to me with the consideration that this is a transitional state.

withinfocus:
I like where this is at now, thanks. While more verbose, I think it's important to keep the layers independent; we should split `.Core` into a few assemblies, with one being for data and another for services so that this is more clear.

src/Infrastructure.EntityFramework/Repositories/Queries/UserCipherDetailsQuery.cs
eliykat: `Run` is already taken as the public interface that just switches between the 2 private methods.

jlf0dev:
@eliykat have you considered creating a new version of the `CipherRepository` class and registering it in DI based on the feature flag instead? That would lock flag changes to application startup of course, but you wouldn't need to touch code all over the place. @withinfocus interested in your thoughts too.

eliykat:
@jlf0dev I did briefly consider it, but it makes the feature flags almost no longer run-time, given that it would require a redeploy (or at least a restart, not sure what the devops process is here) to change its value. That's not consistent with how feature flags otherwise behave so it didn't seem like a good idea to introduce that here.

I guess you could register 2 different `ICipherRepository` implementations and then have a scoped service that chooses which one to call based on the value for the flag? Feels messy though, and it doesn't work well for other repositories where only 1 method is different.

Not sure if it sparks any other ideas @withinfocus ?

jlf0dev:
The more I consider it, losing run-time capabilities seems like a fair sacrifice in this situation. Tagging @bitwarden/dept-devops since I'm sure if thats even possible.

I do think your second idea might still require changes to all the repository references, right? I would prefer the current strategy than that.

withinfocus:
Services have the especially-valuable per-request lifetime to assess feature flag states and I think that's the presiding requirement. While we're not pushing more advanced use cases of flags right now, their ability to change in real-time will be incredibly important to us with regressions, targeting, and experiments. We won't be able to expect recycling to change singleton registrations.

My old-guy default is to keep repositories as dumb as possible with parameters (like a flag state, assuming what will be a very common pattern with a Boolean flag of "old" vs. "new") driving what one returns, but your idea around a scoped service that depends on `IFeatureService` and we give it a constant flag value to make a pivot on is very intriguing to me. Perhaps it's too limiting or relies on uglier language features, but I'd love to explore a fixed use case like I think we'll need with EDD of "flag represents current database call, or new one".

eliykat:
@withinfocus to your point, I kind of like how "dumb" this current implementation turned out to be. It's a bit laborious but very clear and simple to understand (which might seem better than clever DI solutions). I'm working through updating more sprocs, and that's less work now that the `UseFlexibleCollections` getter is established on the main controllers and services that use the repository.

@jlf0dev
> I do think your second idea might still require changes to all the repository references, right? I would prefer the current strategy than that.

My thinking is that a scoped "wrapper" service would implement `ICipherRepository` and would be injected wherever that is currently required with no/minimal changes. It would then calculate the flag value and pass it through to the underlying "real" repository with the rest of the params.

I'm still unsure about whether this extra layer of abstraction is worth it though, my tentative suggestion is that most flags will not have this big an impact on our code base. For now at any rate I'd like to keep it simple, maybe we keep a more advanced solution up our sleeve for next time round.

jlf0dev:
@eliykat Oh nice! I completely misunderstood your idea. I do think thats actually much more preferable from the perspective of a different team. Rather than having to understand the implications of this new feature flag, we can still call the repository the same way. We also have changes sitting in code review to call this repository, and we'll need to make compatible them with these changes.

But I don't want to stop this work from being done, so let me know if you'd still rather go with this route and I'll stop slowing you down 😁

eliykat:
It's a good point - not ideal that other teams need to know about the flag as well.

This implementation is far enough along, let's keep it simple for now and revisit it next time we need to switch db sprocs.

jlf0dev (approved):

shane-melton (dismissed):
DB sprocs and EF query changes look good to go!

eliykat:
Had to fix merge conflicts

jlf0dev (approved):

shane-melton (approved):

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Delete_V2.sql
```diff
@@ -0,0 +1,73 @@
...
+
+    CREATE TABLE #Temp
+    (
+        [Id] UNIQUEIDENTIFIER NOT NULL,
+        [UserId] UNIQUEIDENTIFIER NULL,
+        [OrganizationId] UNIQUEIDENTIFIER NULL,
+        [Attachments] BIT NOT NULL
+    )
+
+    INSERT INTO #Temp
```
rkac-bw, line 16: I find joins to have better plans than in clauses this should be the same:

INSERT INTO #Temp
SELECT
    ucd.[Id],
    ucd.[UserId],
    ucd.[OrganizationId],
    CASE WHEN ucd.[Attachments] IS NULL THEN 0 ELSE 1 END
FROM
    [dbo].[UserCipherDetails_V2](@UserId) ucd
INNER JOIN
    @Ids ids ON ucd.[Id] = ids.Id
WHERE
    ucd.[Edit] = 1

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Delete_V2.sql
```diff
@@ -0,0 +1,73 @@
...
+        [Id],
+        [UserId],
+        [OrganizationId],
+        CASE WHEN [Attachments] IS NULL THEN 0 ELSE 1 END
+    FROM
+        [dbo].[UserCipherDetails_V2](@UserId)
+    WHERE
+        [Edit] = 1
+        AND [Id] IN (SELECT * FROM @Ids)
+
```
rkac-bw, line 27: Possibly join here also:

DELETE c
FROM
    [dbo].[Cipher] c
INNER JOIN
    #Temp t ON c.[Id] = t.[Id]

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Delete_V2.sql
```diff
@@ -0,0 +1,73 @@
...
+    -- Delete ciphers
+    DELETE
+    FROM
+        [dbo].[Cipher]
+    WHERE
+        [Id] IN (SELECT [Id] FROM #Temp)
+
+    -- Cleanup orgs
```
rkac-bw, line 35: Also while loops tend to be less resource intense and faster than cursors:

DECLARE @OrgId UNIQUEIDENTIFIER
DECLARE @OrgIndex INT = 1

-- Create a table variable to hold the distinct OrganizationId values
DECLARE @OrgTable TABLE (IndexId INT IDENTITY(1,1), OrganizationId UNIQUEIDENTIFIER)
INSERT INTO @OrgTable (OrganizationId)
SELECT DISTINCT [OrganizationId]
FROM #Temp
WHERE [OrganizationId] IS NOT NULL

WHILE EXISTS (SELECT 1 FROM @OrgTable WHERE IndexId = @OrgIndex)
BEGIN
    SELECT @OrgId = OrganizationId FROM @OrgTable WHERE IndexId = @OrgIndex
    EXEC [dbo].[Organization_UpdateStorage] @OrgId
    EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationId] @OrgId
    SET @OrgIndex = @OrgIndex + 1
END

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Move_V2.sql
```diff
@@ -0,0 +1,36 @@
...
+    @Ids AS [dbo].[GuidIdArray] READONLY,
+    @FolderId AS UNIQUEIDENTIFIER,
+    @UserId AS UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @UserIdKey VARCHAR(50) = CONCAT('"', @UserId, '"')
+    DECLARE @UserIdPath VARCHAR(50) = CONCAT('$.', @UserIdKey)
+
```
rkac-bw, line 11: Might be more efficient to join here also:

;WITH [IdsToMoveCTE] AS (
    SELECT
        UCDV2.[Id]
    FROM
        [dbo].[UserCipherDetails_V2](@UserId) UCDV2
    INNER JOIN
        @Ids I
    ON
        UCDV2.[Id] = I.Id
)
UPDATE C
SET
    [Folders] =
        CASE
        WHEN @FolderId IS NOT NULL AND [Folders] IS NULL THEN
            CONCAT('{', @UserIdKey, ':"', @FolderId, '"', '}')
        WHEN @FolderId IS NOT NULL THEN
            JSON_MODIFY([Folders], @UserIdPath, CAST(@FolderId AS VARCHAR(50)))
        ELSE
            JSON_MODIFY([Folders], @UserIdPath, NULL)
        END
FROM
    [dbo].[Cipher] C
INNER JOIN
    [IdsToMoveCTE] CTE
ON
    C.[Id] = CTE.[Id]

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Move_V2.sql
```diff
@@ -0,0 +1,36 @@
...
+            WHEN @FolderId IS NOT NULL AND [Folders] IS NULL THEN
+                CONCAT('{', @UserIdKey, ':"', @FolderId, '"', '}')
+            WHEN @FolderId IS NOT NULL THEN
+                JSON_MODIFY([Folders], @UserIdPath, CAST(@FolderId AS VARCHAR(50)))
+            ELSE
+                JSON_MODIFY([Folders], @UserIdPath, NULL)
+            END
+    WHERE
+        [Id] IN (SELECT * FROM [IdsToMoveCTE])
+
```
rkac-bw, line 34: Join here possibly:

;WITH [IdsToMoveCTE] AS (
    SELECT
        UCDV2.[Id]
    FROM
        [dbo].[UserCipherDetails_V2](@UserId) UCDV2
    INNER JOIN
        @Ids I
    ON
        UCDV2.[Id] = I.Id
)
UPDATE C
SET
    [Folders] =
        CASE
        WHEN @FolderId IS NOT NULL AND [Folders] IS NULL THEN
            CONCAT('{', @UserIdKey, ':"', @FolderId, '"', '}')
        WHEN @FolderId IS NOT NULL THEN
            JSON_MODIFY([Folders], @UserIdPath, CAST(@FolderId AS VARCHAR(50)))
        ELSE
            JSON_MODIFY([Folders], @UserIdPath, NULL)
        END
FROM
    [dbo].[Cipher] C
INNER JOIN
    [IdsToMoveCTE] CTE
ON
    C.[Id] = CTE.[Id]

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Restore_V2.sql
```diff
@@ -0,0 +1,62 @@
...
+        [Id],
+        [UserId],
+        [OrganizationId]
+    FROM
+        [dbo].[UserCipherDetails_V2](@UserId)
+    WHERE
+        [Edit] = 1
+        AND [DeletedDate] IS NOT NULL
+        AND [Id] IN (SELECT * FROM @Ids)
+
```
rkac-bw, line 26: Possibly avoid join

INSERT INTO #Temp
SELECT
    UCDV2.[Id],
    UCDV2.[UserId],
    UCDV2.[OrganizationId]
FROM
    [dbo].[UserCipherDetails_V2](@UserId) UCDV2
INNER JOIN
    @Ids I
ON
    UCDV2.[Id] = I.Id
WHERE
    UCDV2.[Edit] = 1
    AND UCDV2.[DeletedDate] IS NOT NULL

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_Restore_V2.sql
```diff
@@ -0,0 +1,62 @@
...
+    DECLARE @UtcNow DATETIME2(7) = GETUTCDATE();
+    UPDATE
+        [dbo].[Cipher]
+    SET
+        [DeletedDate] = NULL,
+        [RevisionDate] = @UtcNow
+    WHERE
+        [Id] IN (SELECT [Id] FROM #Temp)
+
```
rkac-bw, line 35: Avoid in

UPDATE C
SET
    [DeletedDate] = NULL,
    [RevisionDate] = @UtcNow
FROM
    [dbo].[Cipher] C
INNER JOIN
    #Temp T
ON
    C.[Id] = T.[Id]

src/Sql/Vault/dbo/Stored Procedures/Cipher/Cipher_SoftDelete_V2.sql
```diff
@@ -0,0 +1,60 @@
...
+        [Id],
+        [UserId],
+        [OrganizationId]
+    FROM
+        [dbo].[UserCipherDetails_V2](@UserId)
+    WHERE
+        [Edit] = 1
+        AND [DeletedDate] IS NULL
+        AND [Id] IN (SELECT * FROM @Ids)
+
```
rkac-bw, line 26: Possible to avoid in:
INSERT INTO #Temp
SELECT
    UCDV2.[Id],
    UCDV2.[UserId],
    UCDV2.[OrganizationId]
FROM
    [dbo].[UserCipherDetails_V2](@UserId) UCDV2
INNER JOIN
    @Ids I
ON
    UCDV2.[Id] = I.Id
WHERE
    UCDV2.[Edit] = 1
    AND UCDV2.[DeletedDate] IS NULL

util/Migrator/DbScripts/2023-11-03_00_DeprecateAccessAll_UserCollectionDetails.sql
```diff
@@ -0,0 +1,334 @@
...
+            [OrganizationId]
+    OPEN [OrgCursor]
+    FETCH NEXT FROM [OrgCursor] INTO @OrgId
+    WHILE @@FETCH_STATUS = 0 BEGIN
+        EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationId] @OrgId
+        FETCH NEXT FROM [OrgCursor] INTO @OrgId
+    END
+    CLOSE [OrgCursor]
+    DEALLOCATE [OrgCursor]
+
```
rkac-bw, line 329: can avoid cursor:
   DECLARE @OrgIndex INT = 1

    -- Create a table variable to hold the distinct OrganizationId values
    DECLARE @OrgTable TABLE (IndexId INT IDENTITY(1,1), OrganizationId UNIQUEIDENTIFIER)
    INSERT INTO @OrgTable (OrganizationId)
    SELECT DISTINCT [OrganizationId]
    FROM #Temp
    WHERE [OrganizationId] IS NOT NULL

    WHILE @OrgIndex <= (SELECT COUNT(*) FROM @OrgTable)
    BEGIN
        SELECT @OrgId = OrganizationId FROM @OrgTable WHERE IndexId = @OrgIndex
        EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationId] @OrgId
        SET @OrgIndex = @OrgIndex + 1
    END

rkac-bw:
Might want to look at the missing index warning for table cipher from prod and check if it applies to one of the queries:

image

TableName: Cipher
Equality_columns: [UserId]
Include columns: [OrganizationId], [Type], [Data], [Favorites], [Folders], [Attachments], [CreationDate], [RevisionDate], [DeletedDate], [Reprompt], [Key]
Total cost of missing index: 145875

rkac-bw:
Might want to see if the missing indexes on organization might be applicable, besides email status which is already in another PR:

image

eliykat:
@rkac-bw Thanks for the review - all these sprocs are frequently used so optimizations here are definitely worthwhile. However these are all improvements that are unrelated to this changeset - which just focuses on removing the `[OrganizationUser].[AccessAll]` and `[Group].[AccessAll]`. Can I schedule a follow-up ticket for the Vault Team to tackle those changes after this initiative?

rkac-bw (dismissed):
Looks good to me

eliykat:
Thanks! follow up ticket created here: https://bitwarden.atlassian.net/browse/PM-4891

rkac-bw (approved):

shane-melton (approved):

JaredSnider-Bitwarden (approved):
Stamping for auth team as has been approved already. Brief skim looked good!

PR #3378, opened 2023-10-25
Auth/PM-1658 - Dynamic Org Invite Link to accelerate users through org invite accept process 

JaredSnider-Bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

[PM-1658](https://bitwarden.atlassian.net/browse/PM-1658)

Associated clients PR here: https://github.com/bitwarden/clients/pull/6702

In order to support the above clients changes, we needed to send down additional information on the Org Invite Email Join Organization Link URL.  We needed to know the following pieces of information:
- Does the the newly invited org user have an existing account with Bitwarden?
- Does the Org have SSO enabled or SSO login required policy enabled?
    -  We also needed the Org's SSO identifier

### DB and `UserRepository` changes
* **src/Sql/dbo/Stored Procedures/User_ReadByEmails.sql** - Add new `User_ReadByEmails` stored proc as we will need to be able to retrieve N users by N org user emails to determine if they have an existing account
* **util/Migrator/DbScripts/2023-10-21_00_User_ReadByEmails.sql** - Add new `User_ReadByEmails` stored proc to migrations
* **src/Core/Repositories/IUserRepository.cs** - Add new `GetManyByEmailsAsync(...)` method 
* **src/Infrastructure.Dapper/Repositories/UserRepository.cs** - Add Dapper implementation for `GetManyByEmailsAsync(...)`
* **src/Infrastructure.EntityFramework/Repositories/UserRepository.cs** - Add EF implementation for `GetManyByEmailsAsync(...)`

### `OrganizationService` and `HandleBarsMailService` Changes

We had a good bit of duplicated logic in sending org email invites with separate handling for single org invites vs n org invites.  I consolidated the processes to reduce duplication and remove unnecessary methods. We now have a single method for handling one or many org invite emails for simplicity. The number of additional parameters also required the creation of the new `OrganizationInvitesInfo` parameter object to be passed into the mail service for sending org invite emails. 
* **src/Core/Models/Mail/OrganizationInvitesInfo.cs** Create new Org Invites info parameter object to be used in the mail service org invite bulk send email method to consolidate all the parameters into an object instead
* **src/Core/Services/Implementations/OrganizationService.cs** 
    * Refactor and consolidate the building of Org Invite information into a helper method `BuildOrganizationInvitesInfoAsync(...)` for both `SendInvitesAsync(...)` and `SendInviteAsync(...)` which leverages the new `OrganizationInvitesInfo` class to populate all the existing and newly added org invite email information 
    * Replace `_mailService.SendOrganizationInviteEmailAsync(...)` with `_mailService.BulkSendOrganizationInviteEmailAsync(...)` so we can remove it as it is no longer necessary
* **test/Core.Test/Services/OrganizationServiceTests.cs** - Fix tests 
* **src/Core/Models/Mail/OrganizationUserInvitedViewModel.cs** - Update email view model with new required properties and use them in the url
* **src/Core/Services/IMailService.cs** - Remove no longer necessary `SendOrganizationInviteEmailAsync(...)` method and consolidate `BulkSendOrganizationInviteEmailAsync(...)` params using `OrganizationInvitesInfo`
* **src/Core/Services/Implementations/HandlebarsMailService.cs** 
    * Delete `SendOrganizationInviteEmailAsync(...)`
    * Update `BulkSendOrganizationInviteEmailAsync(...)` to use `OrganizationInvitesInfo`
        * Add new params to `OrganizationUserInvitedViewModel` creation
* **src/Core/Services/NoopImplementations/NoopMailService.cs** - Update NoOp service to match standard implementation and remove deleted method 

## Before you submit

[PM-1658]: https://bitwarden.atlassian.net/browse/PM-1658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


util/Migrator/DbScripts/2023-10-21_00_User_ReadByEmails.sql
jlf0dev, line 1: nit: formatting

jlf0dev (commented):
I'm gonna go ahead and request a review from AC on this. Most of this is going to be in their domain.

src/Core/Services/Implementations/HandlebarsMailService.cs
```diff
@@ -206,33 +205,35 @@ public async Task SendOrganizationConfirmedEmailAsync(string organizationName, s
...
+                Token = WebUtility.UrlEncode(invite.Token.Token),
+                ExpirationDate = $"{invite.Token.ExpirationDate.ToLongDateString()} {invite.Token.ExpirationDate.ToShortTimeString()} UTC",
+                OrganizationNameUrlEncoded = WebUtility.UrlEncode(orgInvitesInfo.OrganizationName),
                 WebVaultUrl = _globalSettings.BaseServiceUri.VaultWithHash,
                 SiteName = _globalSettings.SiteName,
-                InitOrganization = initOrganization
+                InitOrganization = orgInvitesInfo.InitOrganization,
+                OrgSsoIdentifier = orgInvitesInfo.OrgSsoIdentifier,
+                OrgSsoEnabled = orgInvitesInfo.OrgSsoEnabled,
+                OrgSsoLoginRequiredPolicyEnabled = orgInvitesInfo.OrgSsoLoginRequiredPolicyEnabled,
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: The [task](https://bitwarden.atlassian.net/browse/PM-1658) flows required me to send down all 3 pieces of information with different routing decisions for combinations of SSO enabled and SSO login policy required. If you still think we can consolidate this, let me know and we can discuss!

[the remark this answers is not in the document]
JaredSnider-Bitwarden: Good call. I've done so without passing too many deps for it to be an issue (imho).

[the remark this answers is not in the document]
JaredSnider-Bitwarden: Per our discussion, I've simplified the query params sent down. The `orgUserHasExistingUser` param is always sent down, but the `orgSsoIdentifier` is only sent down if SSO is enabled and the SSO required policy is on. This allows for the same routing acceleration logic with fewer query params sent to the client. Thanks for the comment!

jlf0dev (dismissed):
LGTM

src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -1053,22 +1054,70 @@ public async Task ResendInviteAsync(Guid organizationId, Guid? invitingUserId, G
...
-        await _mailService.SendOrganizationInviteEmailAsync(organization.Name, orgUser, new ExpiringToken(token, now.AddDays(5)), organization.PlanType == PlanType.Free, initOrganization);
+        // convert single org user into array of 1 org user
+        var orgUsers = new[] { orgUser };
+
+        var orgInvitesInfo = await BuildOrganizationInvitesInfoAsync(orgUsers, organization, initOrganization);
+
+        await _mailService.SendOrganizationInviteEmailsAsync(orgInvitesInfo);
     }
 
+    private async Task<OrganizationInvitesInfo> BuildOrganizationInvitesInfoAsync(
```
vincentsalucci, line 1072: Some may not like these comments as much as I do - but I appreciate the walkthrough!

src/Core/Services/NoopImplementations/NoopMailService.cs
```diff
@@ -53,12 +52,7 @@ public Task SendOrganizationConfirmedEmailAsync(string organizationName, string
         return Task.FromResult(0);
     }
 
-    public Task SendOrganizationInviteEmailAsync(string organizationName, OrganizationUser orgUser, ExpiringToken token, bool isFreeOrg, bool initOrganization = false)
-    {
-        return Task.FromResult(0);
-    }
-
-    public Task BulkSendOrganizationInviteEmailAsync(string organizationName, IEnumerable<(OrganizationUser orgUser, ExpiringToken token)> invites, bool isFreeOrg, bool initOrganization = false)
+    public Task SendOrganizationInviteEmailsAsync(OrganizationInvitesInfo orgInvitesInfo)
```
vincentsalucci, line 55: I like this!

vincentsalucci (approved):

jlf0dev (approved):

jlf0dev (dismissed):

r-tome (approved):

rkac-bw (dismissed):
Looks good to me

src/Sql/dbo/Stored Procedures/User_ReadByEmails.sql
```diff
@@ -0,0 +1,19 @@
+﻿CREATE PROCEDURE [dbo].[User_ReadByEmails]
+    @Emails AS [dbo].[EmailArray] READONLY
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    IF (SELECT COUNT(1) FROM @Emails) < 1
+    BEGIN
+        RETURN(-1)
+    END
```
withinfocus, line 10: ❓ Is this necessary? Seems like it wouldn't happen via the code usage and even if it did it'd return an empty result right?

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -48,6 +48,27 @@ public async Task<User> GetByEmailAsync(string email)
...
+        foreach (var email in emails)
+        {
+            emailTable.Rows.Add(email);
+        }
+
+        using (var connection = new SqlConnection(ConnectionString))
+        {
+            var results = await connection.QueryAsync<User>(
+                $"[{Schema}].[{Table}_ReadByEmails]",
+                new { Emails = emailTable.AsTableValuedParameter("dbo.EmailArray") },
```
withinfocus, line 64: 🌱 I get that this type was already there, but I am wondering about its value as such a structured concept and a type vs. a simple generic string array that has reuse (and to ignore field lengths).

src/Sql/dbo/Stored Procedures/User_ReadByEmails.sql
JaredSnider-Bitwarden: It very well might not be necessary. However, the existing `User_ReadByIds` that I used as inspiration for this stored proc had the logic so I copied it to be safe. Do you want me to remove it?

image

src/Infrastructure.Dapper/Repositories/UserRepository.cs
JaredSnider-Bitwarden: Fair. I'll probably leave that as a future enhancement if that's alright? 

However, I will ask: what types of issues can arise from allowing any length strings vs scoping the strings to a specific domain which matche the length of that domain in our DB (ex: email is NVARCHAR 256)?  If a string longer than a DB column is passed in will it throw an error or truncate the data?

rkac-bw: When you try to update an NVARCHAR(256) field with a value from a variable that is longer the that the actual value in the variable is cut off. SQL Server will truncate the value to fit into the NVARCHAR(300) field. This means that only the first 300 characters of the string will be saved in the NVARCHAR(300) field, and the rest will be lost.

src/Sql/dbo/Stored Procedures/User_ReadByEmails.sql
rkac-bw: if there is not data to process it would be ideal to not make the database call

JaredSnider-Bitwarden: Ok, I'll keep the early return. Thank you for the discussion though!

src/Infrastructure.Dapper/Repositories/UserRepository.cs
JaredSnider-Bitwarden: Ah, good to know! So, it's more safe to use the explicit length types for these operations.

withinfocus: Can certainly be future work. If there's nothing above the database call that's checking this length then this is gonna happen anyway, so it seems more reusable to get rid of this constricted passthrough type and use something open since the expectation is sanitization at a higher level.

rkac-bw: Agreed on an open approach. In this case I had ANSI_warnings off for my session due to some previous testing , it is on by default 

I set it to ON and tried inserting to a tvp and a table with max nvarchar(256)

Both gave an error and aborted:

set ANSI_WARNINGS ON

DECLARE @EmailList AS dbo.EmailArray;
   

INSERT INTO email (Email) VALUES ('email1aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaadddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddaaaaaaaaaaaaaaa@example.com');


INSERT INTO @EmailList (Email) VALUES ('email1aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaadddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddaaaaaaaaaaaaaaa@example.com');


Msg 2628, Level 16, State 1, Line 6
String or binary data would be truncated in table 'vault_dev.dbo.email', column 'email'. Truncated value: 'email1aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaadddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd'.
The statement has been terminated.
Msg 2628, Level 16, State 1, Line 9
String or binary data would be truncated in table 'tempdb.dbo.#B40707EB', column 'Email'. Truncated value: 'email1aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaadddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd'.
The statement has been terminated.

vincentsalucci (approved):

rkac-bw (approved):
previously reviewed, merge conflicts resolved

vincentsalucci (approved):

withinfocus (dismissed):

jlf0dev (dismissed):

jlf0dev (dismissed):

withinfocus (dismissed):

vincentsalucci (approved):

r-tome (approved):

jlf0dev (approved):

withinfocus (approved):

JaredSnider-Bitwarden:
Passed QA! Merging!

PR #3380, opened 2023-10-26
[AC-1761] Delete unused CipherDetails_CreateWithCollections sproc

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This sproc and repository interfaces are dead code. 🧹 

## Code changes

## Before you submit


shane-melton (approved):
LGTM! 🎉 🧹

eliykat:
Nope! I found that it's called from `CipherService.SaveDetailsAsync`. However, because `CipherDetails` is a subclass of `Cipher`, it'll silently fall back to using `CipherService.SaveAsync` if we delete the `SaveDetails` method. At a glance those sprocs are similar but not identical so I think we should hold off on this one.

withinfocus:
Worth looking at that pattern and why a domain object (representing a database table) was subclassed. That could break a lot of things.

PR #3381, opened 2023-10-27
[AC-1696] Fix Provider permissions for Flexible Collections

shane-melton:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
The `CollectionAuthorizationHandler` had an early return if the user was not a member of the org, which prevented providers from passing authorization when they should have. This PR updates the authorization handler to check for provider status of the target organization when the user is not member of the org (`org == null`); and only if they are not a provider and are not a member does the handler return early with a failure.

As a side effect of this change, the `org` variable can be null when passed into the other private methods that do the authorization checks; to account for this, the conditionals related to the `org` variable were switched to pattern matching.

Lastly, the `targetOrganizationId` variable was made a private field to avoid additional parameter passing to the other private methods now that `org` can be null and the target organization ID is still necessary in some cases.

## Code changes

* **src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs:** See above.
* **test/Api.Test/Vault/AuthorizationHandlers/CollectionAuthorizationHandlerTests.cs:** Add tests and a helper enum/method to create multiple tests using the `BitAutoData` attribute. You cannot put `static` fields in an attribute, so I created an enum to be used in the attributes and mapping function to be used in the test body.

## Before you submit


src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -83,20 +85,20 @@ public class CollectionAuthorizationHandler : BulkAuthorizationHandler<Collectio
...
         }
 
         // Owners, Admins, Providers, and users with CreateNewCollections permission can always create collections
         if (
-            org.Type is OrganizationUserType.Owner or OrganizationUserType.Admin ||
-            org.Permissions is { CreateNewCollections: true } ||
-            await _currentContext.ProviderUserForOrgAsync(org.Id))
+            org is { Type: OrganizationUserType.Owner or OrganizationUserType.Admin } ||
+            org is { Permissions.CreateNewCollections: true } ||
+            await _currentContext.ProviderUserForOrgAsync(_targetOrganizationId))
```
eliykat, line 101: If the user is a providerUser, this sproc is now called twice: once in the `HandleRequirementAsync` method and once here.

We could just move it exclusively to the main method, but I like each switch case being self-contained. And it's kind of awkward, e.g.

```cs
if (org is null && isProvider)
{
  context.Succeed(...);
  return;
}
else if (org is null)
{
  return;
}

// switch statement here which exclusively deals with orgUsers
```

I think this shows that the structure of the handler isn't quite as clear as it could be.

What if we removed the providerUser check in the public method, and then each private method was laid out like this:

```cs
private async Task CanCreateAsync(...)
{
  if (org is not null)
  {
  // check organizationUser requirements
  // early return if success
  }

  if (providerUser)
  {
  // success
  }

// if it's not authorised above then it just drops through the if statements and is unauthorized
}
```

At the moment the org logic is too long to nest it like this, but @r-tome extracted the `manageableCollections` logic in https://github.com/bitwarden/server/pull/3360 which makes it more concise. I think with both your refactors it could work quite well.

eliykat (changes requested):

test/Api.Test/Vault/AuthorizationHandlers/CollectionAuthorizationHandlerTests.cs
```diff
@@ -221,4 +247,22 @@ public class CollectionAuthorizationHandlerTests
...
+    private static CollectionOperationRequirement GetRequirementFromEnum(CollectionOperationEnum collectionOperationEnum)
+    {
+        return collectionOperationEnum switch
+        {
+            CollectionOperationEnum.Create => CollectionOperations.Create,
+            CollectionOperationEnum.Delete => CollectionOperations.Delete,
+            CollectionOperationEnum.ModifyAccess => CollectionOperations.ModifyAccess,
+            _ => throw new ArgumentOutOfRangeException(nameof(collectionOperationEnum), collectionOperationEnum, null)
+        };
+    }
```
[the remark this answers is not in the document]
shane-melton: Yeah, its really tricky with the static operations. I thought of the `foreach` as well, but favored the `enum` to avoid re-using the same `sut` and dependencies for what should be independent tests. With that said, I agree and really don't like needing to now maintain the extra `enum`/mapping.

So, I took another look and found that I can do the loop and "recreate" the `sutProvider` in between each operation's test. Still kind of ugly, but does avoid the additional enum and mapping

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
shane-melton: I do like your suggestion and went ahead and refactored the private methods to something similar and removed the provider check in the public method. It should also work nicely with @r-tome's `HasCollectionAccess()` refactor that you mentioned.

eliykat:
You just need to merge in `master` again now that we've merged the feature branch.

shane-melton:
@eliykat Done!

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -29,7 +31,7 @@ public class CollectionAuthorizationHandler : BulkAuthorizationHandler<Collectio
     }
 
     protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context,
-        CollectionOperationRequirement requirement, ICollection<Collection> resources)
+        CollectionOperationRequirement requirement, ICollection<Collection>? resources)
```
r-tome, line 34: Do we need to add the nullable modifier to `ICollection<Collection>`

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
```diff
@@ -83,95 +79,100 @@ public class CollectionAuthorizationHandler : BulkAuthorizationHandler<Collectio
     }
 
     private async Task CanCreateAsync(AuthorizationHandlerContext context, CollectionOperationRequirement requirement,
-        CurrentContextOrganization org)
+        CurrentContextOrganization? org)
```
r-tome, line 82: Same question here, do we need the nullable modifier?

r-tome (commented):
This looks good to me, I just have a question about the nullable modifiers

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
shane-melton: We decided (at some point, can't find a link at the moment) to let the auth handlers be responsible for checking for a null resource. So, once I added `#nullable enable`, I started getting warnings about the `resource == null` check below always being false according to the nullable annotations.

Admittedly, this check is slightly redundant here because this handler extends the `BulkAuthorizationHandler` which already does null checking for us. With that said, I still feel its beneficial to continue being explicit with the null check, even if only to continue setting a precedent for future auth handlers that may not use `BulkAuthorizationHandler`. If you think otherwise let me know!

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
shane-melton: In this case its not strictly required for this method, as we don't have any explicit `org == null` in the `CanCreateAsync()` method. 

However, with the `#nullable enable` directive I found it beneficial to explicitly highlight that the `org` parameter has the potential to be null (in case of Provider users). This is also to be consistent with the other private `Can...Async()` methods that do have some `org is not null` checks.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
r-tome: That makes sense to me! Thanks for taking the time to explain it

r-tome (approved):

Jingo88 (dismissed):

r-tome (approved):

eliykat (approved):
This turned out great 🎉

PR #3385, opened 2023-10-30
[PM-4395] Prevent org scaling when org has reseller provider

amorask-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

[PM-4395](https://bitwarden.atlassian.net/browse/PM-4395)

Associated `clients` change: https://github.com/bitwarden/clients/pull/6698
This PR blocks organization seat scaling when the organization has an enabled resale provider.

## Code changes

* **OrganizationService.cs:** Updates the `CanScale` utility method to return `false` when the organization trying to be scaled has a reseller provider.

## Before you submit

[PM-4395]: https://bitwarden.atlassian.net/browse/PM-4395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eliykat (approved):
What about MSPs? It seems like they'd have the same problem. Or is that OK because they can set a seat autoscale limit? (question for Product)

r-tome (approved):

amorask-bitwarden:
> What about MSPs? It seems like they'd have the same problem. Or is that OK because they can set a seat autoscale limit? (question for Product)

@eliykat Yeah this is only for resellers who cannot allow any auto-scaling because of the manner in which they need pre-approved quotes for invoices: https://bitwarden.atlassian.net/browse/PM-4395?focusedCommentId=45800

eliykat (dismissed):

eliykat (approved):

PR #3388, opened 2023-10-30
[AC-1721] Include limit collection creation/deletion in license file

vincentsalucci:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Code changes
* **src/Core/Entities/Organization.cs** Added `LimitCollectionCreationDeletion` property from license
* **src/Core/Models/Business/OrganizationLicense.cs**: Added `LimitCollectionCreationDeletion` property from org. Bumped license file version. Updated `ValidLicenseVersion` range. Updated `GetDataBytes` to check for new version and property. Updated `VerifyData` to account for new version.
* **src/Core/Services/Implementations/OrganizationService.cs**: Sets property if the feature flag allows or defaults to `true` otherwise

## Before you submit


r-tome (approved):

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -333,6 +337,11 @@ public bool VerifyData(Organization organization, IGlobalSettings globalSettings
                 organization.SmServiceAccounts == SmServiceAccounts;
             }
 
+            if (valid && Version >= 14)
+            {
+                valid = organization.LimitCollectionCreationDeletion == LimitCollectionCreationDeletion;
+            }
```
eliykat, line 343: Can you check with product on this one: should the user be able to change this setting on the self-hosted instance as well?

At the moment they can, however that could put it out of sync with the license file, and this validation would fail.

It also means that the setting value could bounce around unexpectedly if self-hosted and cloud have different values and they update their license file. (i.e. cloud will overwrite the self-hosted value.)

I think these should be editable in 1 location only, or they shouldn't be synced. But largely a product/UX question I think.

vincentsalucci: @eliykat Updating this PR to include endpoint protection for cloud-only access.

eliykat:
Can Billing Team please review organizationLicense changes carefully as the codeowners to be.

eliykat:
I've added unit tests in #3474, I'd like us to merge those first so that this PR can benefit from them.

src/Core/Models/Business/OrganizationLicense.cs
eliykat: Thanks for following this up!

eliykat (commented):
LGTM, this is basically an approval except I'd like us to merge the unit tests first.

src/Core/Entities/Organization.cs
```diff
@@ -258,5 +258,6 @@ public void UpdateFromLicense(OrganizationLicense license)
         UseSecretsManager = license.UseSecretsManager;
         SmSeats = license.SmSeats;
         SmServiceAccounts = license.SmServiceAccounts;
+        LimitCollectionCreationDeletion = license.LimitCollectionCreationDeletion;
```
eliykat, line 261: This just occurred to me - this needs similar feature flag logic to `OrganizationService`. The OrgService method creates an organization, this method updates an existing organization. In both cases we want to ignore the license property and use the default value unless FC has been enabled for self-hosted.

eliykat, line 261: This just occurred to me - this needs similar feature flag logic to `OrganizationService`. The OrgService method creates an organization, this method updates an existing organization. In both cases we want to ignore the license property and use the default value unless FC has been enabled for self-hosted.

I suggest just passing in a `bool` for the flag state.

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -189,6 +191,8 @@ public byte[] GetDataBytes(bool forHash = false)
                     (Version >= 13 || !p.Name.Equals(nameof(UsePasswordManager))) &&
                     (Version >= 13 || !p.Name.Equals(nameof(SmSeats))) &&
                     (Version >= 13 || !p.Name.Equals(nameof(SmServiceAccounts))) &&
+                    // LimitCollectionCreationDeletion was added in Version 14
+                    (Version >= 14 || !p.Name.Equals(nameof(LimitCollectionCreationDeletion))) &&
```
eliykat, line 195: One problem with the current approach is that if the feature flag is off, the license file won't match the org object (because the create/update methods will choose to ignore the value in the license file). So there's a risk that validation will fail. Given that it's not a billing-critical property, maybe we just exclude it from validation for now.

Alternatively, maybe we don't flag any of this logic and we just let the value be updated from the license file. After all, the actual logic will still be flagged on self-hosted. My concern there was that it'll be a latent change that will suddenly take effect after updating the self-hosted instance, and that could be unexpected.

There's a bit to this so let's discuss in the next FC sync.

eliykat (commented):
A couple of things that just occurred to me.

src/Core/Models/Business/OrganizationLicense.cs
eliykat: @vincentsalucci I think the conclusion from our sync was that my feedback here is fine. Amended the comment above.

eliykat (dismissed):
lgtm! It's flagged & has tests so I think we can merge (if you agree!) but please get QA to test as well (e.g. generate a license file with the new code, upload it to a self-hosted instance on old code, etc)

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -135,6 +136,7 @@ public OrganizationLicense()
     public bool UseSecretsManager { get; set; }
     public int? SmSeats { get; set; }
     public int? SmServiceAccounts { get; set; }
+    public bool LimitCollectionCreationDeletion { get; set; }
```
eliykat, line 139: This might need a default value of `true`. i.e.

```suggestion
    public bool LimitCollectionCreationDeletion { get; set; } = true;
```

Think of the situation where a self-hosted server updates its code and has the flag on (so we'll use this value), but it's using an existing (old) license file that doesn't have this value in it. In that case, this class will be instantiated and this will default to `false`, which will then overwrite the default value of `true` in the db.

This is part of the trouble of having `true` as a default value instead of `false`. Won't be doing that again for sure.

eliykat (approved):

cturnbull-bitwarden (approved):

gbubemismith (approved):

PR #3389, opened 2023-10-31
[AC-85] Set Max Seats Autoscale and Current Seats via Public API

cyprain-okeke:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

As an organizational admin user of the public API, I want to be able to update max seat autoscaling, current seats, and storage via the Public API so that I do not have to do this manually and so that I can perform this in bulk automatically on a given date each month to combine invoice/billing changes, which will reduce overhead and load on our accounts payable team in dealing with autoscaling invoices as users provision on-demand.

## Code changes

## Before you submit


src/Api/Billing/Public/Models/SecretsManagerSubscriptionUpdateRequestModel.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Bit.Api.Billing.Public.Models;
+
+public class SecretsManagerSubscriptionUpdateRequestModel
+{
+    [Required]
+    public int SeatAdjustment { get; set; }
+    public int? MaxAutoscaleSeats { get; set; }
+    public int ServiceAccountAdjustment { get; set; }
+    public int? MaxAutoscaleServiceAccounts { get; set; }
```
eliykat, line 13: I don't like that seats and service account values use deltas ("adjustment" amounts) whereas max autoscale values use totals. We do it in our private API and I don't like it there either. I think it's confusing to have similar values treated differently, and "adjustment" values are not idempotent (i.e. if you accidentally send it multiple times you'll have multiple charges).

I suggest that all properties use totals, i.e.
```suggestion
    public int Seats { get; set; }
    public int? MaxAutoscaleSeats { get; set; }
    public int ServiceAccounts { get; set; }
    public int? MaxAutoscaleServiceAccounts { get; set; }
```

However, this is just my personal bugbear, has there been any input from Product or the CS Integration team as to their preferred API design?

Note: I have not done a full review, just flagging this issue.

src/Api/Billing/Public/Controllers/OrganizationSubscriptionsController.cs
```diff
@@ -0,0 +1,91 @@
...
+using Bit.Core.OrganizationFeatures.OrganizationSubscriptions.Interface;
+using Bit.Core.Repositories;
+using Bit.Core.Services;
+using Bit.Core.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Bit.Api.Billing.Public.Controllers;
+
+[Route("public/organizationsubscriptions")]
```
amorask-bitwarden, line 15: This might just be personal preference, but I think we could simplify this endpoint by making it a bit more RESTful. 

Presumably, our primary resource here is an organization's subscription, and seemingly we just want to update it rather than create one from scratch. As such, I think it would almost be easiest to provide just 1 endpoint:
`PUT /public/organizations/{organizationId}/subscription`

And keep it in an `OrganizationsController` in the same folder. Doing so would let us expand this public API to perform other organization-related billing changes that might not have to do with subscriptions exactly.

This endpoint could have a request body like the following so that the client wouldn't need to invoke two different endpoints to update the same resource:
``` json
{
    "passwordManager": {
        "seats": 1,
        "maxAutoScaleSeats": 10
    },
    "secretsManager": {
        "seats": 1,
        "maxAutoScaleSeats": 10,
        "serviceAccounts": 1,
        "maxAutoScaleServiceAccounts": 2
    },
    "storage": 1
}
```

src/Api/Billing/Public/Models/OrganizationSubscriptionUpdateRequestModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Bit.Api.Billing.Public.Models;
+
+public class OrganizationSubscriptionUpdateRequestModel
```
amorask-bitwarden, line 5: Looks like the ticket also calls for the ability to update storage in the description. Might want to check with the reporter on whether that is actually required.

src/Api/Billing/Public/Models/SecretsManagerSubscriptionUpdateRequestModel.cs
amorask-bitwarden: I agree with @eliykat, if have the opportunity to keep these consistent in a new controller, I think it's worth doing so.

src/Api/Billing/Public/Controllers/OrganizationSubscriptionsController.cs
eliykat: Great suggestion, also this is why you should absolutely get feedback from our integration team as to what API design they prefer. They deal with customers on this stuff every day. And public APIs need to be stable, which means it'll be hard to change whatever the initial implementation here is. (e.g. Jordan C)

eliykat: Also just a cautionary note that some of these properties are nullable (e.g. max autoscale), the API will need to be clear about what happens if a value is not provided. (Does it clear the existing saved value or not update it?)

cyprain-okeke: Thanks for the feedback, I have reach out to @notnamed  for prefered API design.

notnamed: Apologies for the delay @cyprain-okeke !

Agreed with the above. Because of the design of the existing Public API endpoint, I would as a customer look under the existing `/public/organization` structure (we currently have `/public/organization/import` which is used in bulk inviting members). However, when it comes to combining the Password Manager and Secrets Manager endpoints, I will defer to someone who's brain works better thinking RESTfully than mine. If I don't have SM enabled in my subscription, is my `secretsManager` JSON object empty, or not present? If I'm adding SM to my PM sub via the API, is that a PUT to this endpoint or a POST?

On the topic of `maxAutoScale`, I would expect a null value to leave my value unchanged, and a -1 to disable the seat limit entirely, so that I don't have to perform a separate call to look up what my existing max is set to. In a PUT request, I think it makes sense to leave the value unchanged - I'm not PUTing a value to that object so keep what is already there.

src/Api/Billing/Public/Controllers/OrganizationSubscriptionsController.cs
```diff
@@ -0,0 +1,95 @@
...
+using Bit.Core.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using OrganizationSubscriptionUpdateRequestModel = Bit.Api.Billing.Public.Models.OrganizationSubscriptionUpdateRequestModel;
+
+namespace Bit.Api.Billing.Public.Controllers;
+
+[Route("public/organizations")]
+[Authorize("Organization")]
+public class OrganizationSubscriptionsController : Controller
```
amorask-bitwarden, line 18: Since this route is going to be `/public/organizations`, I think we'll want to use `OrganizationsController` as the name of the class and the file.

amorask-bitwarden (changes requested):

amorask-bitwarden (approved):

eliykat (changes requested):
This is looking good, some feedback below.

One outstanding issue is that this doesn't look like it addresses Jordan's comment [here](https://github.com/bitwarden/server/pull/3389#discussion_r1385608733):

> On the topic of maxAutoScale, I would expect a null value to leave my value unchanged, and a -1 to disable the seat limit entirely, so that I don't have to perform a separate call to look up what my existing max is set to. In a PUT request, I think it makes sense to leave the value unchanged - I'm not PUTing a value to that object so keep what is already there.

I agree with that suggestion, that is,
* if the value is `null`, don't change it
* for nullable properties like max autoscale - if the value is `-1`, set it to null. The idea here is to distinguish between an "intentional null" (clearing the value) and an unintentional null (not supplying a value, therefore leaving it unchanged).

(This is only an issue for properties that are nullable in our database. For non-nullable properties, the user can never update it to `null`, so we can safely skip it if not provided.)

I'm sure this isn't the only way to deal with this, feel free to see what we're doing elsewhere in our public API at the moment, or what stackoverflow recommends, or what `#code` thinks.

src/Api/Billing/Public/Controllers/OrganizationsController.cs
```diff
@@ -0,0 +1,95 @@
...
+using Bit.Core.Services;
+using Bit.Core.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using OrganizationSubscriptionUpdateRequestModel = Bit.Api.Billing.Public.Models.OrganizationSubscriptionUpdateRequestModel;
+
+namespace Bit.Api.Billing.Public.Controllers;
+
+[Route("public/organizations")]
+[Authorize("Organization")]
```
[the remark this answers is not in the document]
withinfocus: Seconded. Is there any reason why the existing controller can't be used?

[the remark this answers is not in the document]
cyprain-okeke: I thinking since activites in https://github.com/bitwarden/server/blob/master/src/Api/AdminConsole/Public/Controllers/OrganizationController.cs relates to adminConsole, We can change then new route to [Route("public/billing-organization")] or [Route("public/organization")] and still have it in Billing/Public/Controllers/OrganizationController.cs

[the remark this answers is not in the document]
eliykat: @withinfocus the existing controller is owned by AC Team, this is owned by Billing Team. Separate files for separate code ownership. But from a public api perspective, they both belong together on an `organizations/{id}/` endpoint.

eliykat (dismissed):
By the way, I recommend you run this code locally (if you're not already) and test out different requests against your new endpoint to make sure everything's working as you expect. See our current [API documentation](https://bitwarden.com/help/public-api/) and [OpenAPI specification](https://bitwarden.com/help/api/). If you have any trouble getting this set up, you can ask in `#code`.

src/Api/Billing/Public/Models/OrganizationSubscriptionUpdateRequestModel.cs
```diff
@@ -0,0 +1,95 @@
...
+    public virtual SecretsManagerSubscriptionUpdate ToSecretsManagerSubscriptionUpdate(Organization organization)
+    {
+        return new SecretsManagerSubscriptionUpdate(organization, false)
+        {
+            MaxAutoscaleSmSeats = MaxAutoScaleSeats ?? organization.MaxAutoscaleSmSeats,
+            MaxAutoscaleSmServiceAccounts = MaxAutoScaleServiceAccounts ?? organization.MaxAutoscaleSmServiceAccounts
+        }
+            .AdjustSeats(Seats)
+            .AdjustServiceAccounts(ServiceAccounts);
+    }
```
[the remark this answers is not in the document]
cyprain-okeke: I have handled this with set { _seats = value is < 0 or null ? 0 : value; } or set { _maxAutoScaleSeats = value < 0 ? null : value; }

src/Api/Billing/Public/Controllers/OrganizationsController.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private async Task UpdatePasswordManagerSubscriptionAsync(Guid id, OrganizationSubscriptionUpdateRequestModel model)
+    {
+        await _organizationService.UpdateSubscription(id, model.PasswordManagerSubscriptionUpdateModel.Seats,
+            model.PasswordManagerSubscriptionUpdateModel.MaxAutoScaleSeats);
+        if (model.PasswordManagerSubscriptionUpdateModel.Storage.HasValue)
+        {
+            await _organizationService.AdjustStorageAsync(id, (short)model.PasswordManagerSubscriptionUpdateModel.Storage);
+        }
```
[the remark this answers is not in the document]
cyprain-okeke: I will create a ticket for this

src/Api/Billing/Public/Controllers/OrganizationsController.cs
```diff
@@ -0,0 +1,76 @@
...
+    [ProducesResponseType((int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [ProducesResponseType((int)HttpStatusCode.NotFound)]
+    public async Task<IActionResult> PostSubscriptionAsync([FromBody] OrganizationSubscriptionUpdateRequestModel model)
+    {
+        var organization = await _organizationRepository.GetByIdAsync(_currentContext.OrganizationId.Value);
+        if (organization == null)
+        {
+            throw new NotFoundException();
+        }
```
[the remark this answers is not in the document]
cyprain-okeke: I have tested it

withinfocus (commented):
I haven't followed the entire conversation here but would still think a shared controller is possible.

src/Api/Billing/Public/Models/OrganizationSubscriptionUpdateRequestModel.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        get { return _maxAutoScaleSeats; }
+        set { _maxAutoScaleSeats = value < 0 ? null : value; }
+    }
+
+    private int? _serviceAccounts;
+    public int? ServiceAccounts
+    {
+        get { return _serviceAccounts; }
+        set { _serviceAccounts = value is < 0 or null ? 0 : value; }
```
cturnbull-bitwarden, line 67: Also here, `SmServiceAccounts` is allowed to be `null`.

src/Api/Billing/Public/Models/OrganizationSubscriptionUpdateRequestModel.cs
```diff
@@ -0,0 +1,98 @@
...
+        if (ServiceAccounts.HasValue)
+        {
+            update.AdjustServiceAccounts(ServiceAccounts.Value);
+        }
+
+        return update;
+    }
+}
+
+
```
cturnbull-bitwarden, line 97: Remove the extra whitespace; there should just be one newline at the end of the file

src/Api/Billing/Public/Controllers/OrganizationsController.cs
```diff
@@ -0,0 +1,73 @@
...
+using Bit.Core.Repositories;
+using Bit.Core.Services;
+using Bit.Core.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using OrganizationSubscriptionUpdateRequestModel = Bit.Api.Billing.Public.Models.OrganizationSubscriptionUpdateRequestModel;
+
+namespace Bit.Api.Billing.Public.Controllers;
+
+[Route("public/billing-organization")]
```
cturnbull-bitwarden, line 15: The value for this route should be `public/organization` in keeping with the AC public controller of the same name. The API user shouldn't care how we set up our teams and code ownership

cturnbull-bitwarden (changes requested):
Sorry that it took me some time to take this over this PR review from @eliykat! Thank you for your work on this so far. I did leave some comments, however. Feel free to reach out if you'd like to discuss them

cturnbull-bitwarden (approved):
Nice work!

eliykat:
I removed my review given that @cturnbull-bitwarden took over for me. Thanks all!

cturnbull-bitwarden (approved):

PR #3398, opened 2023-11-01
[AC-1772] Check null or whitespace on `GatewayCustomerId` and `BillingEmail`

amorask-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/src/Admin/Controllers/OrganizationsController.cs
+++ b/src/Admin/Controllers/OrganizationsController.cs
@@ -215,7 +215,10 @@ public class OrganizationsController : Controller
 
         try
         {
-            await _stripeSyncService.UpdateCustomerEmailAddress(organization.GatewayCustomerId, organization.BillingEmail);
+            if (!string.IsNullOrWhiteSpace(organization.GatewayCustomerId) && !string.IsNullOrWhiteSpace(organization.BillingEmail))
+            {
+                await _stripeSyncService.UpdateCustomerEmailAddress(organization.GatewayCustomerId, organization.BillingEmail);
+            }
         }
         catch (StripeException stripeException)
         {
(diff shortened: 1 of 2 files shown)


cturnbull-bitwarden (approved):

src/Admin/Controllers/OrganizationsController.cs
```diff
@@ -215,7 +215,10 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
 
         try
         {
-            await _stripeSyncService.UpdateCustomerEmailAddress(organization.GatewayCustomerId, organization.BillingEmail);
+            if (!string.IsNullOrWhiteSpace(organization.GatewayCustomerId) && !string.IsNullOrWhiteSpace(organization.BillingEmail))
```
eliykat, line 218: We should check if the email has changed before hitting the Stripe API.

src/Admin/Controllers/OrganizationsController.cs
```diff
@@ -215,7 +215,10 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
+            {
+                await _stripeSyncService.UpdateCustomerEmailAddress(organization.GatewayCustomerId, organization.BillingEmail);
+            }
         }
         catch (StripeException stripeException)
```
eliykat, line 223: Is it worth rethrowing so that the entire operation is aborted, just because we can't update their email in Stripe?

It might be more appropriate to prompt the user ("unable to update Stripe, do you want to continue?"), or provide an override switch or something. A question for Product.

eliykat (commented):
I know this is a hotfix so these can be follow up items if required.

cturnbull-bitwarden (approved):

PR #3399, opened 2023-11-01
Auth/PM-3659 - Disable Passkey registration if Require SSO Policy Enabled 

JaredSnider-Bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

[PM-3659](https://bitwarden.atlassian.net/browse/PM-3659) -  If the require SSO policy applies to the user (doesn't apply to admins / owners), disable the user's ability to create new credentials.
Associated clients PR: https://github.com/bitwarden/clients/pull/6729

## Code changes

* **src/Api/Auth/Controllers/WebAuthnController.cs:** Add validation step on passkey creation which checks if the require SSO policy applies to the user and throws if it does. 

Recording of server responses (commented out client logic which hides the ability to register a passkey if require SSO is enabled and applicable for this testing):

https://github.com/bitwarden/server/assets/116684653/460ec552-494f-4902-93d8-cbbc279789c6

## Before you submit

[PM-3659]: https://bitwarden.atlassian.net/browse/PM-3659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coroiu:
@JaredSnider-Bitwarden a thought: what about also running the validation as part of the `options` call? That way you wouldn't even be able to create a credential

JaredSnider-Bitwarden:
> @JaredSnider-Bitwarden a thought: what about also running the validation as part of the `options` call? That way you wouldn't even be able to create a credential

Good point. I've done so!

src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -75,6 +82,17 @@ public async Task Post([FromBody] WebAuthnCredentialRequestModel model)
         }
     }
 
+    private async Task ValidateRequireSsoPolicyDisabledOrNotApplicable(Guid userId)
+    {
+        var requireSsoLogin = await _policyService.AnyPoliciesApplicableToUserAsync(userId, PolicyType.RequireSso);
+
+        if (requireSsoLogin)
+        {
+            // User is not an admin or owner of this org and the org requires SSO login so do not allow passkey creation.
```
coroiu, line 91: suggestion(non-blocking): this comment is not relevant anymore, right?

coroiu (commented):
@JaredSnider-Bitwarden some unit tests? 😄

src/Api/Auth/Controllers/WebAuthnController.cs
JaredSnider-Bitwarden: No, it is still relevant. The logic is just applied via the `AnyPoliciesApplicableToUserAsync` method.

coroiu: Ok but since we're not actually checking for admin or owner in this class, someone could change the implementation of `AnyPoliciesApplicableToUserAsync` and not have any idea that this comment needed to be updated? 

My point was mostly that we don't need to worry about those things here, since the service takes care of them for us

JaredSnider-Bitwarden: Ah, fair. Ok, I'll remove the comment. Thank you for the discussion

coroiu (dismissed):

coroiu (approved):

PR #3400, opened 2023-11-01
[AC-1578] Fixed issue where legacy plans couldn't sign up for SM

cturnbull-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

An issue was encountered in production where old 2019 plans couldn't sign up for secrets manager, even after my latest work on this was merged which should have accomplished this. This PR addresses this by removing the validation which ensures legacy plans aren't included when upgrading a plan to SM

## Code changes

* **OrganizationService.cs:** Removed validation that ensured legacy plans couldn't be upgraded to SM
* **OrganizationServiceTests.cs:** Removed unit test that ensured enterprise 2019 couldn't be upgraded to SM

## Before you submit

### DIFF
--- a/src/Core/Services/Implementations/OrganizationService.cs
+++ b/src/Core/Services/Implementations/OrganizationService.cs
@@ -1950,6 +1950,11 @@ public class OrganizationService : IOrganizationService
 
     private static void ValidatePlan(Models.StaticStore.Plan plan, int additionalSeats, string productType)
     {
+        if (plan is null)
+        {
+            throw new BadRequestException($"{productType} Plan was null.");
+        }
+
         if (plan.Disabled)
         {
             throw new BadRequestException($"{productType} Plan not found.");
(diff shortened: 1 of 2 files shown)


src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -1950,11 +1950,6 @@ private async Task<Organization> GetOrgById(Guid id)
 
     private static void ValidatePlan(Models.StaticStore.Plan plan, int additionalSeats, string productType)
     {
-        if (plan is not { LegacyYear: null })
-        {
-            throw new BadRequestException($"Invalid {productType} plan selected.");
-        }
-
```
eliykat, line 1957: This method is used (via `ValidatePasswordManagerPlan`) to prevent a user from signing up or upgrading to a legacy plan - which we still want to enforce.

Given that you can only signup/upgrade to PM, and SM is still an add-on, the quick solution is to move this `if` block into `ValidatePasswordManagerPlan` so that we retain the check on signup/upgrade but not on adding SM.

This validation logic is a mess and should be refactored before we allow separate subscriptions to PM and/or SM.

Separately - this pattern matching hides a `null` check which we should keep. So after moving it, add back a null check here.

eliykat (changes requested):

src/Core/Services/Implementations/OrganizationService.cs
cturnbull-bitwarden: Good catch, I updated `ValidatePasswordManagerPlan` to contain this check and update the validation method to ensure `plan` isn't `null`

eliykat (approved):

PR #3402, opened 2023-11-02
Add message for FeatureUnavailableException

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Add a separate message for `FeatureUnavailableException` to help with debugging, both while developing and if a customer stumbles across it in prod.

As noted in the xml comment, it is not intended that a user ever sees this; it indicates that we've messed up our client-side feature flagging logic. However, that is helpful to know!

## Code changes

## Before you submit


withinfocus (changes requested):
You can still move to `NotFoundObjectResult` per our discussion if you'd like too. I see a few good ideas at https://stackoverflow.com/questions/20139621/how-do-i-return-notfound-ihttpactionresult-with-an-error-message-or-exception.

src/Api/Utilities/ExceptionHandlerFilterAttribute.cs
```diff
@@ -82,6 +82,11 @@ public override void OnException(ExceptionContext context)
             errorMessage = "Resource not found.";
             context.HttpContext.Response.StatusCode = 404;
         }
+        else if (exception is FeatureUnavailableException)
+        {
+            errorMessage = "This feature is not available.";
```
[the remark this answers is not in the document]
eliykat: Not sure this is right - we own [NotFoundException](https://github.com/bitwarden/server/blob/feature-unavailable-exception-improvements/src/Core/Exceptions/NotFoundException.cs#L3-L3), for example, and many others in this `if...else if` block.

The message given to the constructor will still be included in the error response if it's running in dev mode: https://github.com/bitwarden/server/blob/feature-unavailable-exception-improvements/src/Api/Utilities/ExceptionHandlerFilterAttribute.cs#L133-L141. But this provides a guaranteed user-appropriate message for production use.

This does suggest that these properties should just be on the custom Exception object though; this long `if...else if` block is basically polymorphic behaviour.

I am unsure how far to go in this PR in refactoring this class.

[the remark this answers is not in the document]
withinfocus: `NotFoundException` is ours from inheritance yes, but I mean it has a way to set the message on the base. Why not keep `FeatureUnavailableException` inheriting that and inside that class set the static "This feature is not available." message; `FeatureUnavailableException` stays parameterless in its constructor.

The status code will still be set.

src/Api/Utilities/ExceptionHandlerFilterAttribute.cs
```diff
@@ -82,6 +82,11 @@ public override void OnException(ExceptionContext context)
             errorMessage = "Resource not found.";
             context.HttpContext.Response.StatusCode = 404;
         }
+        else if (exception is FeatureUnavailableException)
```
withinfocus, line 85: ❌ The current state of this PR has `FeatureUnavailableException` unused.

src/Core/Utilities/RequireFeatureAttribute.cs
```diff
@@ -30,7 +31,7 @@ public override void OnActionExecuting(ActionExecutingContext context)
 
         if (!featureService.IsEnabled(_featureFlagKey, currentContext))
         {
-            throw new FeatureUnavailableException();
+            context.Result = new NotFoundObjectResult("This feature is unavailable.");
```
withinfocus, line 34: ❓ If this setting of the result is acceptable to our middleware and everything works as expected we'd be complete with this change here, but per previous comments it isn't in our situation right? If we can make it move that way that's great, but I thought we need an exception to throw.

test/Core.Test/Utilities/RequireFeatureAttributeTests.cs
```diff
@@ -23,8 +22,14 @@ public void Throws_When_Feature_Disabled()
...
-        // Act & Assert
-        Assert.Throws<FeatureUnavailableException>(() => rfa.OnActionExecuting(GetContext(enabled: false)));
+        // Act
+        var context = GetContext(enabled: false);
+        rfa.OnActionExecuting(context);
+
+        // Assert
+        Assert.IsType<NotFoundObjectResult>(context.Result);
+        var result = (NotFoundObjectResult)context.Result;
+        Assert.Contains("This feature is unavailable", (string)result.Value);
```
withinfocus, line 32: ❓ Do we need to assert error messages in tests?

withinfocus (changes requested):

src/Core/Utilities/RequireFeatureAttribute.cs
eliykat: I've tested this and confirm that it returns a 404 error with the message. Also note [per the docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.filters.actionexecutingcontext.result?view=aspnetcore-7.0&viewFallbackFrom=net-6.0):

> Setting Result to a non-null value inside an action filter will short-circuit the action and any remaining action filters.

test/Core.Test/Utilities/RequireFeatureAttributeTests.cs
eliykat: It's our current practice. In this case (for example), it ensures that the `NotFoundObjectResult` is the one I expect and not for some other reason (a false negative).

src/Api/Utilities/ExceptionHandlerFilterAttribute.cs
eliykat: It's still used in `CollectionsController` and `CollectionAuthorizationHandler`. We've used those as guards at the top of feature-flagged methods, e.g. 

```cs
if (flagEnabled)
{
  callnewMethod();
}

public void newMethod() {
  // just to make sure noone calls this accidentally...
  if (!flagEnabled)
  {
     throw FeatureUnavailableException();
  }
  // etc etc
}
```

That is maybe overkill and misuse of the exception. What do you think? Maybe we should just remove this use of it and rely on unit testing instead, then we don't need the exception at all.

src/Core/Utilities/RequireFeatureAttribute.cs
withinfocus: If we mix patterns but this is the recommended way and we'd eventually want to move everything to it, I support the mixing.

src/Api/Utilities/ExceptionHandlerFilterAttribute.cs
withinfocus: If your approach to set the result is working well I think it's best to use it everywhere and drop any inconsistent usages.

eliykat:
This is pretty old by this point and has never come up again, so I'm leaving it be.

PR #3403, opened 2023-11-02
[AC-1779] Add comment to clarify ExpirationWithoutGracePeriod in OrganizationLicense

cyprain-okeke:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

OrganizationLicense.VerifyData is missing validation for the ExpirationWithoutGracePeriod property that was added by Billing Team. Please fix this following the pattern in the rest of that method and request a review from AC Team.

## Code changes

## Before you submit


withinfocus:
This needs to be some kind of datetime-oriented check, right? To ensure one date is less than the other.

MGibson1 (changes requested):
⚠️ This is in direct conflict to creation of an org license done [here](https://github.com/bitwarden/server/blob/master/src/Core/Entities/Organization.cs#L254).

Unless expiration date is equal to expiration date without grace period, these aren't going to be equal

eliykat:
I agree with @MGibson1 - I didn't look too closely at this before creating a ticket, it just looked like a miss because it's different to the established pattern.

I'm not sure about adding a db column for it - it's been a while since I looked at this, but why can't we just calculate the grace period date from the expiration date, or vice versa? It would be good to follow up and see if there's a better way of doing this.

For now my only real request is to add a comment here noting that v12 is intentionally excluded, e.g.

```cs
// Version 12 added ExpirationWithoutDatePeriod, but that property is informational only and is not saved
// to the Organization object. It's validated as part of the hash but does not need to be validated here.
```

eliykat (approved):

cturnbull-bitwarden (approved):

r-tome (approved):

MGibson1 (approved):

PR #3406, opened 2023-11-02
[AC-1784] Lining up new Manage collection permissions for users with deprecated EditAssignedCollections permission

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

> An EditAssignedCollection permission (pre-flexible collections) is equivalent to having a Can Manage permission on all collections you’re assigned to (post-flexible collections). We want to start setting up this data now.
> 
> Whenever an orgUser is assigned to a collection, or has their collection access updated, we want to give them Can Manage permissions for that collection if they have the EditAssignedCollection permission and the feature flag is OFF. (Once the feature flag is ON, then the Can Manage permissions are live and we can’t mess with them.)
> 
> Whenever an orgUser has their role/permissions changed, they need to have CanManage set to true /false for their existing collections if they’re gaining/losing this custom permission and the feature flag is OFF.

## Code changes

* **src/Core/Services/Implementations/CollectionService.cs:** Saving a collection now adds the permission `Manage = true` to assigned users that have the `EditAssignedCollections` permission
* **src/Core/Services/Implementations/OrganizationService.cs:** Inviting or saving an organization user with collection access now adds the permission `Manage = true` if they also have the `EditAssignedCollections` permission

## Before you submit


src/Core/Services/Implementations/OrganizationService.cs
```diff
@@ -61,6 +61,8 @@ public class OrganizationService : IOrganizationService
     private readonly IDataProtectorTokenFactory<OrgUserInviteTokenable> _orgUserInviteTokenDataFactory;
     private readonly IFeatureService _featureService;
 
+    private bool UseFlexibleCollections => _featureService.IsEnabled(FeatureFlagKeys.FlexibleCollections, _currentContext);
```
vincentsalucci, line 64: Do we need both this and the var down below?

vincentsalucci (approved):

PR #3422, opened 2023-11-08
[AC-1126] Flexible collections: Deprecate manager role

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
With Flexible Collections enabled the Manager role is deprecated.

## Code changes
This is targeting https://github.com/bitwarden/server/pull/3360 so that should be merged first.

* **src/Core/Context/CurrentContext.cs:** Only add the organizations that the user is the manager of if Flexible Collections feature flag is not enabled
* **src/Core/Context/ICurrentContext.cs:** Marked the method `OrganizationManager` as obsolete
* **src/Core/Services/Implementations/OrganizationService.cs:** Reject any attempt to assign the Manager role to a user if the feature flag is enabled
* **test/Core.Test/Services/OrganizationServiceTests.cs:** Added unit tests to check that any attempt to assign the Manager role is blocked if the feature flag is enabled

## Before you submit


src/Core/Context/CurrentContext.cs
```diff
@@ -224,7 +224,7 @@ private List<CurrentContextOrganization> GetOrganizations(Dictionary<string, IEn
                 }));
         }
 
-        if (claimsDict.ContainsKey(Claims.OrganizationManager))
+        if (!UseFlexibleCollections && claimsDict.ContainsKey(Claims.OrganizationManager))
```
eliykat, line 227: If we skip this and don't add any claim in its place, it's going to look like the user isn't a member of the organization at all.

If FC is enabled, you could add a `CurrentContextOrganization` with `Type = OrganizationUserType.User`, however there are still other sources (e.g. db sprocs) that will be returning the "real" value. So should we try to overwrite it like this, or just update the logic that checks it?

r-tome: > If we skip this and don't add any claim in its place, it's going to look like the user isn't a member of the organization at all.

You're absolutely right! I've reverted this change.

> Where is this consumed?

Only on the method `CurrentContext.OrganizationManager` so I've modified it to throw an exception if the feature flag is enabled.

vincentsalucci (dismissed):
This is looks fine to me! Please make sure @eliykat does one final pass for his requested change(s). 

One small nit: `UseFlexibleCollections` by itself doesn't really tell me what the variable represents (of course, reviewing the value being assigned does offer that context). Maybe changing `Use` to `Using` or even `FlexibleCollectionsEnabled`.

eliykat (approved):
This is a very small changeset, am I right in thinking that basically all the work was already done in #3360? That is, by deprecating those `CurrentContext` checks, we got this ticket for (almost) free?

In that case - lgtm! Although you've been following @vincentsalucci's feedback above in other PRs, but you haven't made that change here. Once that's done, this should be good to go.

It's a small changeset behind flags, so I think it can merged without QA.

r-tome:
I have renamed `UseFlexibleCollections` to `FlexibleCollectionsIsEnabled`. `UseFlexibleCollections` was the old name we were using everywhere which is why it was also here.
@eliykat yes all the work is mostly done in https://github.com/bitwarden/server/pull/3360 hence why this PR is targeting that branch, so that should be merged first.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1968,6 +1970,11 @@ public async Task ValidateOrganizationUserUpdatePermissions(Guid organizationId,
         {
             throw new BadRequestException("Custom users can only grant the same custom permissions that they have.");
         }
+
+        if (FlexibleCollectionsIsEnabled && newType == OrganizationUserType.Manager)
```
eliykat, line 1974: I'm a little concerned this might block other updates that are unrelated to the role, given that we still have the Manager enum type floating around. I suggest we check that either the orgUser is being created, or that their role is being explicitly changed. e.g. 

```suggestion
        if (FlexibleCollectionsIsEnabled && newType == OrganizationUserType.Manager && (oldType == null) || oldType != OrganizationUserType.Manager)
```

This is probably not necessary but seems safer.

eliykat (changes requested):
Unrelated to this PR: is it correct that we call `ValidateOrganizationUserUpdatePermissions` here: https://github.com/bitwarden/server/blob/5152e4bc2c800757df8cbd492cd3b2b2f1de39e6/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserGroupsCommand.cs#L27-L31 ?

That command is updating user-group associations, not the user role.

eliykat:
Created a follow-up ticket here: https://bitwarden.atlassian.net/browse/AC-1961

eliykat (approved):

PR #3425, opened 2023-11-08
[PM-3797 Part 1] Layout new key rotation methods

jlf0dev:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We currently have multiple api calls when rotating the user key. The intention here is to consolidate these api calls and, as a result, using a single transaction for the database call.

The server changes for these improvements ended up being large so I'm splitting them out as much as I can. I am introducing broken code, but it is hidden behind a feature flag.

## Code changes

* **UpdateKeyRequestModel.cs:** Add `EmergencyAccessKeys` and `ResetPasswordKeys`. They will both require new models in the future with Id to identify the data to rotate.
* **AccountsController.cs:** Adds feature flag with template of what the command call will look like. Validation will be done by controller, and implemented by each team. Future PRs will finish validation and building data model to be passed to `RotateUserKeyCommand`.
* **RotateUserKeyCommand.cs:** New command for rotation of user key. Responsible for verifying master password hash. A list of db calls for each domain will be collected and passed to the repo.
* **UserRepository.cs:** new repository method to update keys and encrypted data. Each team will be responsible for owning a method that can update their required data using the given transaction (EF won't pass transaction since it should catch it)
* **src/Core/Constants.cs:** new feature flag for key improvements

## Before you submit


src/Api/Controllers/AccountsController.cs
```diff
@@ -370,6 +375,45 @@ public async Task PostKdf([FromBody] KdfRequestModel model)
...
+        {
+            return;
+        }
+
+        foreach (var error in result.Errors)
+        {
+            ModelState.AddModelError(string.Empty, error.Description);
+        }
+
+        await Task.Delay(2000);
```
ike-kottlowski, line 413: ❓ Why is there a delay here?

src/Api/Auth/Models/Request/Accounts/RotateUserKeyRequestModel.cs
```diff
@@ -0,0 +1,18 @@
...
+using Bit.Api.AdminConsole.Models.Request.Organizations;
+using Bit.Api.Tools.Models.Request;
+using Bit.Api.Vault.Models.Request;
+
+namespace Bit.Api.Auth.Models.Request.Accounts;
+
+public class RotateUserKeyRequestModel
+{
+    [Required][StringLength(300)] public string MasterPasswordHash { get; set; }
+    [Required] public string Key { get; set; }
```
ike-kottlowski, line 11: nit: shouldn't this be `UserKey`? Or is it implied 🤔

src/Api/Controllers/AccountsController.cs
jlf0dev: I believe its to prevent timing attacks for the MP verification. I copied it from previous endpoint, but I've since consolidated the two endpoints.

src/Api/Auth/Models/Request/Accounts/RotateUserKeyRequestModel.cs
jlf0dev: I decided to not create a new endpoint and request model. This was a good suggestion, but making this change to an existing request model requires compatibility changes which might be out of scope right now.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -179,88 +177,57 @@ public async Task UpdateRenewalReminderDateAsync(Guid id, DateTime renewalRemind
...
-                        // emergencyAccessKeys.UpdateEncryptedData(user.Id, connection, transaction);
-                    }
-
-                    if (resetPasswordKeys.Any())
-                    {
-                        // resetPasswordKeys.UpdateEncryptedData(user.Id, connection, transaction);
-                    }
-
-                    transaction.Commit();
+                    cmd.Parameters.Add("@PrivateKey", SqlDbType.VarChar).Value = DBNull.Value;
```
ike-kottlowski, line 200: nit: This format might improve readability.

```suggestion
                cmd.Parameters.Add("@PrivateKey", SqlDbType.VarChar).Value = string.IsNullOrWhiteSpace(user.PrivateKey) ? DBNull.Value : user.PrivateKey;
```

ike-kottlowski (approved):
One suggestion to improve readability. But not a blocker.

src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Bit.Core.Auth.Models.Data;
+using Bit.Core.Repositories;
+using Bit.Core.Services;
+using Microsoft.AspNetCore.Identity;
+
+namespace Bit.Core.Auth.UserFeatures.UserKey.Implementations;
+
+public class RotateUserKeyCommand : IRotateUserKeyCommand
```
coroiu, line 8: question: should this be internal?
```suggestion
internal class RotateUserKeyCommand : IRotateUserKeyCommand
```

src/Infrastructure.Dapper/Repositories/UserRepository.cs
jlf0dev: Good idea, added!

ike-kottlowski (approved):

src/Api/Controllers/AccountsController.cs
MGibson1, line 1: Can you move this controller into auth?

test/Api.Test/Controllers/AccountsControllerTests.cs
MGibson1, line 1: Can you move this controller into auth?

MGibson1 (approved):
I like the new design, nice work.

I think auth should own that controller, though. What do you think about Auth handling user repositories too?

src/Api/Controllers/AccountsController.cs
jlf0dev: Yeah I've been planning to move them at some point. I'll move them in the next couple PRs in this train.

PR #3432, opened 2023-11-09
[PM-4619] Rewrite `UserService` methods as commands

coroiu:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Rewrite methods to commands for better:
  - Maintainability
  - Code ownership

## Before you submit


src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/CreateWebAuthnLoginCredentialCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+using Bit.Core.Auth.Repositories;
+using Bit.Core.Entities;
+using Bit.Core.Utilities;
+using Fido2NetLib;
+
+namespace Bit.Core.Auth.UserFeatures.WebAuthnLogin.Implementations;
+
+internal class CreateWebAuthnLoginCredentialCommand : ICreateWebAuthnLoginCredentialCommand
+{
+    public const int MaxCredentialsPerUser = 5;
```
ike-kottlowski, line 11: ❓ Is there a reason this is hard coded? might be worth plopping in a comment 🤔

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/CreateWebAuthnLoginCredentialCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+    public CreateWebAuthnLoginCredentialCommand(IFido2 fido2, IWebAuthnCredentialRepository webAuthnCredentialRepository)
+    {
+        _fido2 = fido2;
+        _webAuthnCredentialRepository = webAuthnCredentialRepository;
+    }
+
+    public async Task<bool> CreateWebAuthnLoginCredentialAsync(User user, string name, CredentialCreateOptions options, AuthenticatorAttestationRawResponse attestationResponse, bool supportsPrf, string encryptedUserKey = null, string encryptedPublicKey = null, string encryptedPrivateKey = null)
+    {
+        var existingCredentials = await _webAuthnCredentialRepository.GetManyByUserIdAsync(user.Id);
+        if (existingCredentials.Count >= 5)
```
ike-kottlowski, line 25: Shouldn't this use the class variable `MaxCredentialsPerUser`?

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/CreateWebAuthnLoginCredentialCommand.cs
coroiu: Just didn't really know where else to put it 🤷 Suggestions are welcome!

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/CreateWebAuthnLoginCredentialCommand.cs
coroiu: Woooops, yep! Good catch!

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/CreateWebAuthnLoginCredentialCommand.cs
ike-kottlowski: Is the max number of credentials 5 for all security keys? or is this an arbitrary number we've selected? I don't really know enough of the fine details to suggest a comment. 🫠

ike-kottlowski (approved):

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/CreateWebAuthnLoginCredentialCommand.cs
coroiu: It's an arbitrary number picked by product to limit the amount of passkeys that users are allowed to register. It's an artificial limitation, i.e. from a purely technical perspective we could allow users to create hundreds

test/Core.Test/Auth/UserFeatures/WebAuthnLogin/AssertWebAuthnLoginCredentialCommandTests.cs
ike-kottlowski, line 1: I like that we have tests for the command. What I'm concerned about is confidence the logic trees are being executed as we expect. Since all the exceptions return the same error message it is difficult to follow the full execution.

The tests add mock returns for services that _imply_ the execution trees. I would feel better about the tests if we verified the mock returns.

For example we could use the `sutProvider.GetDependency<T>().Received(n).method()` this would verify the execution. 

```
internal async void UserNotFound_ThrowsBadRequestException(...)
{
  // Arrange
  response.Response.UserHandle = user.Id.ToByteArray();
  sutProvider.GetDependency<IUserRepository>().GetByIdAsync(user.Id).ReturnsNull();

  // Act
  try
  {
    await sutProvider.Sut.AssertWebAuthnLoginCredential(options, response);
  }
  catch (BadRequestException ex)
  {
  //Assert
    await sutProvider.GetDependency<IUserRepository>().Received(1).GetByIdAsync(user.Id);
    Assert.Equal("Invalid credential.", ex.Message);
  }
}
```

Not a blocker but Something to consider.

ike-kottlowski (approved):

test/Core.Test/Auth/UserFeatures/WebAuthnLogin/AssertWebAuthnLoginCredentialCommandTests.cs
coroiu: Interesting point! As a counter argument, I'm not really trying to test the execution trees, I'm trying to test behavior. In my mind, I don't really mind if the error were to come from somewhere completely different, as long as I get an error when a user is not found. To put other way:

> I'm not testing `if (user === null)`, I'm testing `User Empty => BadRequestException`. If I could remove `if (user == null)` and still get `BadRequestException` out I would do it

PR #3433, opened 2023-11-09
[AC-1809] JWT claim for LimitCollectionCreationDeletion

vincentsalucci:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Code changes
* **src/Core/Context/CurrentContext.cs**: Added mapping from claims and applied them to the `CurrentContextOrganization` being created
* **src/Core/Identity/Claims.cs**: Added `LimitCollectionCreationDeletion` claim key
* **src/Core/Utilities/CoreHelpers.cs**: Added logic to create the claim with the `orgId` as the value
* **src/Identity/IdentityServer/ApiResources.cs**: Added new claim property to list of claims resources available to the `api`

## Before you submit

### DIFF
--- a/src/Core/Context/CurrentContext.cs
+++ b/src/Core/Context/CurrentContext.cs
@@ -175,6 +175,10 @@ public class CurrentContext : ICurrentContext
             ? claimsDict[Claims.SecretsManagerAccess].ToDictionary(s => s.Value, _ => true)
             : new Dictionary<string, bool>();
 
+        var limitCollectionCreationDeletion = claimsDict.TryGetValue(Claims.LimitCollectionCreationDeletion, out var claims)
+            ? claims.ToDictionary(s => s.Value, _ => true)
+            : new Dictionary<string, bool>();
+
         var organizations = new List<CurrentContextOrganization>();
         if (claimsDict.ContainsKey(Claims.OrganizationOwner))
         {
@@ -184,6 +188,7 @@ public class CurrentContext : ICurrentContext
                     Id = new Guid(c.Value),
                     Type = OrganizationUserType.Owner,
                     AccessSecretsManager = accessSecretsManager.ContainsKey(c.Value),
+                    LimitCollectionCreationDeletion = limitCollectionCreationDeletion.ContainsKey(c.Value)
                 }));
         }
         else if (orgApi && OrganizationId.HasValue)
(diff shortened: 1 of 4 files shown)


shane-melton (approved):

withinfocus:
Was there a discussion I can read on why this would be in our tokens' claims? I am not understanding here vs. in app logic.

src/Core/Context/CurrentContext.cs
```diff
@@ -175,6 +175,10 @@ private List<CurrentContextOrganization> GetOrganizations(Dictionary<string, IEn
             ? claimsDict[Claims.SecretsManagerAccess].ToDictionary(s => s.Value, _ => true)
             : new Dictionary<string, bool>();
 
+        var limitCollectionCreationDeletion = claimsDict.TryGetValue(Claims.LimitCollectionCreationDeletion, out var claims)
```
shane-melton, line 178: question: Can we follow the pattern used above for `accessSecretsManager`? 

I ask because I'll be doing something similar for the 2nd setting and the `out var claims` is "polluting" the scope with a `claims` variable that is only related to the `limitCollectionCreationDeletion` setting and never used again.

withinfocus: This extends upon my earlier comment. A claim to me is "Secrets Manager" -- if you have access it's there. Same would go for Password Manager really as we embrace being multi-product.

Collections management feels like a stretch for a claim -- that's some kind of internal permission.

vincentsalucci: Hey all! 

@shane-melton Surely can. Will just have to set aside my OCD for squiggly lines. 🤣 

@withinfocus - There were a few on-call discussions about why we thought about having the value in the claims made sense:

- It prevented extraneous database calls when needing to check this organization setting within the `CurrentContext` class. This is seemingly necessary due to our permission chains stemming within the `CurrentContext` class. 
- While it does appear to be outside of the scope of a token claim, it is actually tightly coupled with a user's role (which is what the majority of this token's existence is used for in this scope) and their permission to create/delete collections (applied in parallel to existing permission checks).

withinfocus: Hmm, OK ... let's think about the definition of the claim some more then. With the mindset that a claim is an assertion of an ability, can this be structured different about the capabilities with collections? I would think the claim is the converse about the _ability_ to create and delete collections.

Additionally, is it worth separating creation vs. deletion, as two claims?

PR #3437, opened 2023-11-10
[AC-1682] Flexible collections: data migrations for deprecated permissions

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionGroups.sql
+++ b/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionGroups.sql
@@ -1,10 +1,10 @@
--- Step 1: Retrieve relevant data from [dbo].[Group] where [AccessAll] is 1
+-- Step 1: Retrieve Groups with [AccessAll] permission
 SELECT [Id] AS [GroupId], [OrganizationId]
 INTO #TempGroup
 FROM [dbo].[Group]
 WHERE [AccessAll] = 1;
 
--- Step 2: Declare variables for group and organization IDs
+-- Step 2: Declare variables for GroupId and OrganizationId
 DECLARE @GroupId UNIQUEIDENTIFIER;
 DECLARE @OrganizationId UNIQUEIDENTIFIER;
 
@@ -33,7 +33,7 @@ UPDATE SET
 INSERT ([CollectionId], [GroupId], [ReadOnly], [HidePasswords], [Manage])
 VALUES (source.[CollectionId], source.[GroupId], 0, 0, 1);
 
--- Step 6: Fetch the next group and organization IDs
+-- Step 6: Fetch the next GroupId and OrganizationId
 FETCH NEXT FROM GroupCursor INTO @GroupId, @OrganizationId;
 END;
 
--- a/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionGroups.sql
+++ b/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionGroups.sql
@@ -1,0 +1,45 @@
+-- Step 1: Retrieve relevant data from [dbo].[Group] where [AccessAll] is 1
+SELECT [Id] AS [GroupId], [OrganizationId]
+INTO #TempGroup
+FROM [dbo].[Group]
+WHERE [AccessAll] = 1;
+
+-- Step 2: Declare variables for group and organization IDs
+DECLARE @GroupId UNIQUEIDENTIFIER;
+DECLARE @OrganizationId UNIQUEIDENTIFIER;
+
+-- Step 3: Create a cursor to iterate through the results of the temporary table
+DECLARE GroupCursor CURSOR FOR
+SELECT [GroupId], [OrganizationId]
+FROM #TempGroup;
+
+OPEN GroupCursor;
+
+-- Step 4: Loop through the groups
+FETCH NEXT FROM GroupCursor INTO @GroupId, @OrganizationId;
+
+WHILE @@FETCH_STATUS = 0
+BEGIN
+    -- Step 5: Use MERGE to insert or update into [dbo].[CollectionGroup] for each [dbo].[Collection] entry
+MERGE INTO [dbo].[CollectionGroup] AS target
+    USING (SELECT C.[Id] AS [CollectionId], @GroupId AS [GroupId] FROM [dbo].[Collection] C WHERE C.[OrganizationId] = @OrganizationId) AS source
+    ON (target.[CollectionId] = source.[CollectionId] AND target.[GroupId] = source.[GroupId])
+    WHEN MATCHED THEN
+UPDATE SET
+    target.[ReadOnly] = 0,
+    target.[HidePasswords] = 0,
+    target.[Manage] = 1
+    WHEN NOT MATCHED THEN
+INSERT ([CollectionId], [GroupId], [ReadOnly], [HidePasswords], [Manage])
+VALUES (source.[CollectionId], source.[GroupId], 0, 0, 1);
+
+-- Step 6: Fetch the next group and organization IDs
+FETCH NEXT FROM GroupCursor INTO @GroupId, @OrganizationId;
+END;
+
+-- Step 7: Close and deallocate the cursor
+CLOSE GroupCursor;
+DEALLOCATE GroupCursor;
+
+-- Step 8: Drop the temporary table
+DROP TABLE #TempGroup;
--- a/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionUsers.sql
+++ b/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionUsers.sql
@@ -1,10 +1,10 @@
--- Step 1: Retrieve relevant data from [dbo].[OrganizationUser] where [AccessAll] is 1
+-- Step 1: Retrieve OrganizationUsers with [AccessAll] permission
 SELECT [Id] AS [OrganizationUserId], [OrganizationId]
 INTO #TempOrgUser
 FROM [dbo].[OrganizationUser]
 WHERE [AccessAll] = 1;
 
--- Step 2: Declare variables for organization user and organization ID
+-- Step 2: Declare variables for OrganizationUserId and OrganizationId
 DECLARE @OrgUserId UNIQUEIDENTIFIER;
 DECLARE @OrganizationId UNIQUEIDENTIFIER;
 
@@ -22,7 +22,7 @@ WHILE @@FETCH_STATUS = 0
 BEGIN
     -- Step 5: Use MERGE to insert or update into [dbo].[CollectionUser] for each [dbo].[Collection] entry
 MERGE INTO [dbo].[CollectionUser] AS target
-    USING (SELECT C.[Id] AS [CollectionId], @OrgUserId AS [OrganizationUserId] FROM [dbo].[Collection] C WHERE C.[OrganizationId] = @OrganizationId) AS source -- Adjusted to use OrganizationId
+    USING (SELECT C.[Id] AS [CollectionId], @OrgUserId AS [OrganizationUserId] FROM [dbo].[Collection] C WHERE C.[OrganizationId] = @OrganizationId) AS source
     ON (target.[CollectionId] = source.[CollectionId] AND target.[OrganizationUserId] = source.[OrganizationUserId])
     WHEN MATCHED THEN
 UPDATE SET
@@ -33,7 +33,7 @@ UPDATE SET
 INSERT ([CollectionId], [OrganizationUserId], [ReadOnly], [HidePasswords], [Manage])
 VALUES (source.[CollectionId], source.[OrganizationUserId], 0, 0, 1);
 
--- Step 6: Fetch the next organization user and organization ID
+-- Step 6: Fetch the next OrganizationUserId and OrganizationId
 FETCH NEXT FROM OrgUserCursor INTO @OrgUserId, @OrganizationId;
 END;
 
--- a/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionUsers.sql
+++ b/util/Migrator/DbScripts/2023-11-10_00_AccessAllCollectionUsers.sql
@@ -1,0 +1,45 @@
+-- Step 1: Retrieve relevant data from [dbo].[OrganizationUser] where [AccessAll] is 1
+SELECT [Id] AS [OrganizationUserId], [OrganizationId]
+INTO #TempOrgUser
+FROM [dbo].[OrganizationUser]
+WHERE [AccessAll] = 1;
+
+-- Step 2: Declare variables for organization user and organization ID
+DECLARE @OrgUserId UNIQUEIDENTIFIER;
+DECLARE @OrganizationId UNIQUEIDENTIFIER;
+
+-- Step 3: Create a cursor to iterate through the results of the temporary table
+DECLARE OrgUserCursor CURSOR FOR
+SELECT [OrganizationUserId], [OrganizationId]
+FROM #TempOrgUser;
+
+OPEN OrgUserCursor;
+
+-- Step 4: Loop through the organization users
+FETCH NEXT FROM OrgUserCursor INTO @OrgUserId, @OrganizationId;
+
+WHILE @@FETCH_STATUS = 0
+BEGIN
+    -- Step 5: Use MERGE to insert or update into [dbo].[CollectionUser] for each [dbo].[Collection] entry
+MERGE INTO [dbo].[CollectionUser] AS target
+    USING (SELECT C.[Id] AS [CollectionId], @OrgUserId AS [OrganizationUserId] FROM [dbo].[Collection] C WHERE C.[OrganizationId] = @OrganizationId) AS source -- Adjusted to use OrganizationId
+    ON (target.[CollectionId] = source.[CollectionId] AND target.[OrganizationUserId] = source.[OrganizationUserId])
+    WHEN MATCHED THEN
+UPDATE SET
+    target.[ReadOnly] = 0,
+    target.[HidePasswords] = 0,
+    target.[Manage] = 1
+    WHEN NOT MATCHED THEN
+INSERT ([CollectionId], [OrganizationUserId], [ReadOnly], [HidePasswords], [Manage])
+VALUES (source.[CollectionId], source.[OrganizationUserId], 0, 0, 1);
+
+-- Step 6: Fetch the next organization user and organization ID
+FETCH NEXT FROM OrgUserCursor INTO @OrgUserId, @OrganizationId;
+END;
+
+-- Step 7: Close and deallocate the cursor
+CLOSE OrgUserCursor;
+DEALLOCATE OrgUserCursor;
+
+-- Step 8: Drop the temporary table
+DROP TABLE #TempOrgUser;
--- a/util/Migrator/DbScripts/2023-11-10_00_ManagersEditAssignedCollectionUsers.sql
+++ b/util/Migrator/DbScripts/2023-11-10_00_ManagersEditAssignedCollectionUsers.sql
@@ -1,0 +1,12 @@
+-- Update [dbo].[CollectionUser] with [Manage] = 1 for all users with Manager role or 'EditAssignedCollections' permission
+UPDATE cu
+SET cu.[ReadOnly] = 0,
+    cu.[HidePasswords] = 0,
+    cu.[Manage] = 1
+FROM [dbo].[CollectionUser] cu
+JOIN [dbo].[Collection] c
+    ON cu.[CollectionId] = c.[Id]
+JOIN [dbo].[OrganizationUser] ou
+    ON cu.[OrganizationUserId] = ou.[Id]
+WHERE (ou.[Type] = 3 OR (ou.[Permissions] IS NOT NULL AND
+    ISJSON(ou.[Permissions]) > 0 AND JSON_VALUE(ou.[Permissions], '$.editAssignedCollections') = 'true'))


r-tome:
EF migrations are missing at this moment, I will add them soon.

eliykat:
Data migrations should be in the `dbScripts_transition` please

rkac-bw (dismissed):
Looks good to me

eliykat:
For visibility - I'm working through other PRs first so probably won't get to this until next week.

eliykat:
Started looking at this today, but would like a second review from another Flexible Collections expert.

eliykat (changes requested):

util/Migrator/DbScripts_transition/2023-11-10_00_AccessAllCollectionUsers.sql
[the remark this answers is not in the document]
r-tome: Yes, that is the correct order. I have added the order to the file names and renamed them to today's date.

eliykat (approved):

util/Migrator/DbScripts_transition/2023-11-10_00_ManagersEditAssignedCollectionUsers.sql
```diff
@@ -0,0 +1,12 @@
+-- Update [dbo].[CollectionUser] with [Manage] = 1 for all users with Manager role or 'EditAssignedCollections' permission
+UPDATE cu
+SET cu.[ReadOnly] = 0,
+    cu.[HidePasswords] = 0,
+    cu.[Manage] = 1
+FROM [dbo].[CollectionUser] cu
+JOIN [dbo].[Collection] c
+    ON cu.[CollectionId] = c.[Id]
```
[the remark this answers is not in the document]
rkac-bw: Agreed no need for collection join as I can see

util/Migrator/DbScripts_transition/2023-12-06_02_ManagersEditAssignedCollectionUsers.sql
eliykat, line 1: I just realised that this does not handle access via groups. If a Manager has access to collections via a group, then we need to create a new CollectionUser association for them with CanManage. (given that we cannot upgrade the entire Group to CanManage without also giving other group members this access)

eliykat (changes requested):

util/Migrator/DbScripts_transition/2023-12-06_02_ManagersEditAssignedCollectionUsers.sql
r-tome: I've updated the scripts to include a migration for that scenarion. Thanks for looking out!

eliykat (approved):
Nicely done!

rkac-bw (dismissed):
Looks good to me

util/Migrator/DbScripts_transition/2023-12-06_02_ManagersEditAssignedCollectionUsers.sql
```diff
@@ -0,0 +1,25 @@
...
+SET cu.[ReadOnly] = 0,
+    cu.[HidePasswords] = 0,
+    cu.[Manage] = 1
+FROM [dbo].[CollectionUser] cu
+INNER JOIN [dbo].[OrganizationUser] ou
+    ON cu.[OrganizationUserId] = ou.[Id]
+WHERE (ou.[Type] = 3 OR (ou.[Permissions] IS NOT NULL AND
+    ISJSON(ou.[Permissions]) > 0 AND JSON_VALUE(ou.[Permissions], '$.editAssignedCollections') = 'true'))
+
+-- Insert rows to [dbo].[CollectionUser] for Managers and users with 'EditAssignedCollections' permission assigned to groups with collection access
```
eliykat, line 12: What if they also have direct access (i.e. a CollectionUser entry)? This will create duplicate rows.

eliykat (changes requested):
We need to bump the account revision date for all affected users. e.g. see `Collection_UpdateWithGroupsAndUsers.sql` - the last thing it does is call `User_BumpAccountRevisionDateByCollectionId`.

You should be able to use that sproc here (or a similar `BumpAccountRevisionDate` sproc, there are a few).

Also, with the changes discussed in Slack (user-initiated migrations), these migration scripts would become sprocs that are called for a specific OrganizationId only. That probably makes the batching logic unnecessary. However, you can wait for our dev sync to discuss before doing this extra work if you like.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,132 @@
...
+                )
+
+        -- Step 5: Set all Managers to Users
+            UPDATE [dbo].[OrganizationUser]
+            SET [Type] = 2  -- User
+            WHERE [OrganizationId] = @OrganizationId
+              AND [Type] = 3; -- Manager
+
+        -- Step 6: Bump the account revision dates for all users in the organization
+        EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationId] @OrganizationId
```
eliykat, line 123: Not all users will be affected by these changes. I think you'd want to use `ByCollectionIds` or `ByOrganizationUserId` to target only those users we've actually updated.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,132 @@
...
+
+        COMMIT TRANSACTION;
+    END TRY
+    BEGIN CATCH
+        ROLLBACK TRANSACTION;
+        THROW;
+    END CATCH;
```
eliykat, line 130: Do we need to do anything else here, or does SQL just keep track of this for us and knows how to rollback the changes in the event of a db error?

util/Migrator/DbScripts_transition/2024-01-10_00_AllOrgsEnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,44 @@
...
+    WHERE [Batch] = @CurrentBatch;
+
+    OPEN OrgCursor;
+
+    FETCH NEXT FROM OrgCursor INTO @OrganizationId;
+
+    WHILE @@FETCH_STATUS = 0
+    BEGIN
+        -- Execute the stored procedure for the current OrganizationId
+        EXEC [dbo].[Organization_EnableCollectionEnhancements] @OrganizationId;
```
eliykat, line 31: As discussed, not sure if this will perform well enough to run for the whole db, but will defer to @rkac-bw on that one

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
eliykat, line 1: I'm a little nervous about so much happening in a single transaction. My understanding is that changes are not written to the db until you call `COMMIT`, so later queries may not pick up the changes you made earlier in the sproc. This was written to execute in a particular order as separate migration scripts, so this might cause unexpected behavior. Please discuss with @rkac-bw

eliykat (commented):
Some high level feedback

util/Migrator/DbScripts_transition/2024-01-10_00_AllOrgsEnableCollectionEnhancements.sql
r-tome: That is a valid concern but I thought it wouldn't be a problem since we're applying the changes per each `OrganizationId`

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
r-tome: That is a very good point, I don't know what is the correct way. My reasoning for having a single transaction was that in case of any failure then everything should be reverted.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
rkac-bw: It will keep track of changes involved and will recognize updated data within the transaction, how qquickl

r-tome:
Currently I have two approaches for the data migration:
- Multiple data migration scripts
    - `util/Migrator/DbScripts_transition/2024-01-10_00_AccessAllCollectionGroups.sql`
    - `util/Migrator/DbScripts_transition/2024-01-10_01_AccessAllCollectionUsers.sql`
    - `util/Migrator/DbScripts_transition/2024-01-10_02_ManagersEditAssignedCollectionUsers.sql`
    - `util/Migrator/DbScripts_transition/2024-01-10_03_EnableOrgsFlexibleCollections.sql`
- Single migration script executing stored procedure per each `OrganizationId`
    - `util/Migrator/DbScripts_transition/2024-01-10_04_AllOrgsEnableCollectionEnhancements.sql`
        - executes `src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql` which will exist anyway for when an Organization opts it

@rkac-bw which approach do you think is best?

util/Migrator/DbScripts_transition/2024-01-10_00_AllOrgsEnableCollectionEnhancements.sql
rkac-bw: I think this one will do well as it is batched and per org

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
rkac-bw: You could change it to be a transaction per group, org , would be slower but less of a locking risk

r-tome: I have updated the sproc to have a transaction per block.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
r-tome: I have updated the sproc to bump the dates by `OrganizationUserId`

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
r-tome: I've modified the script to have multiple transactions (one per block).

rkac-bw (approved):
looks good to me

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
rkac-bw: @eliykat Even with the transactions scoped to original blocks, do you think they may cause more issues than solve?
I am leaning in that direction

eliykat: I agree, it was much cleaner before with the single TRY/CATCH block. But for example, referring back to the [previous iteration of the sproc](https://github.com/bitwarden/server/blob/54cc6fab8f162448446eeb06822e44e97a2b6534/src/Sql/dbo/Stored%20Procedures/Organization_EnableCollectionEnhancements.sql), what happens if:

1. we insert new rows into CollectionUser table [here](https://github.com/bitwarden/server/blob/54cc6fab8f162448446eeb06822e44e97a2b6534/src/Sql/dbo/Stored%20Procedures/Organization_EnableCollectionEnhancements.sql#L68-L77)
2. and then try to update those rows [here](https://github.com/bitwarden/server/blob/54cc6fab8f162448446eeb06822e44e97a2b6534/src/Sql/dbo/Stored%20Procedures/Organization_EnableCollectionEnhancements.sql#L89-L98) when they haven't been committed to the db yet
3. and then potentially need to check if those rows have already been created [here](https://github.com/bitwarden/server/blob/54cc6fab8f162448446eeb06822e44e97a2b6534/src/Sql/dbo/Stored%20Procedures/Organization_EnableCollectionEnhancements.sql#L100-L114)?

Am I right in thinking that those later queries won't see the rows created in the first step?

If so - what's our other alternative? Restructure our queries so that they can be run independently of each other?

If I am wrong and we can just go back to that previous version, all the better.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,299 @@
...
+                SELECT [OrganizationUserId]
+                FROM #TempOrganizationUsers
+
+                OPEN UniqueOrgUserIdCursor
+                FETCH NEXT FROM UniqueOrgUserIdCursor INTO @Step1OrganizationUserId
+
+                WHILE (@@FETCH_STATUS = 0)
+                BEGIN
+                    -- Execute the stored procedure for the current OrganizationUserId
+                    EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationUserId] @Step1OrganizationUserId
```
eliykat, line 61: There is a bulk version of this sprocs which will accept multiple orgUser Ids. If that's cleaner

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
eliykat: Maybe I'm overcomplicating it - is your suggestion just to ditch the TRY/CATCH altogether?

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,299 @@
...
+        END TRY
+        BEGIN CATCH
+            ROLLBACK TRANSACTION;
+            THROW;
+        END CATCH;
+
+        -- Drop the temporary table
+        DROP TABLE #TempStep4;
+
+    -- Step 5: Set all Managers to Users
```
eliykat, line 253: I think it would make more sense to do this in Steps 4 & 5 for each user affected by those steps. You could have the same UPDATE statement but joining on #TempStep4 / #TempStep5 instead. Then you save this whole step and you don't have to bump their account revision date again.

It's also how you treated clearing AccessAll.

withinfocus (commented):
Please remove the bulk operation here for now as we'll want to ship that as a separate change after we plan for its execution by hand for our cloud.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
eliykat: We've discussed this and I misunderstood how transactions work. Marking this as resolved but we'll continue discussing the approach here.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
r-tome: I have condensed steps 3, 4 and 5 into step 3.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,222 @@
...
+          AND [OrganizationId] = @OrganizationId;
+
+        -- Create a temporary table to store distinct OrganizationUserIds
+        SELECT DISTINCT GU.[OrganizationUserId]
+        INTO #TempOrganizationUsers
+        FROM [dbo].[GroupUser] GU
+        JOIN #TempGroup TG ON GU.[GroupId] = TG.[GroupId];
+
+        BEGIN TRY
+            BEGIN TRANSACTION;
```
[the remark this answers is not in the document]
r-tome: I've updated the script to have the just the one big transaction.

r-tome:
@eliykat / @rkac-bw  I have removed the cursors from the script and also combined all distinct `OrganizationUserId` into one variable and then use that to bump all revision dates in one go. Does it make sense or is it safer to update them once per step like before?

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,160 @@
...
+                    INSERT INTO [dbo].[CollectionUser] ([CollectionId], [OrganizationUserId], [ReadOnly], [HidePasswords], [Manage])
+                    SELECT source.[CollectionId], source.[OrganizationUserId], 0, 0, 0
+                    FROM (
+                             SELECT C.[Id] AS [CollectionId], T.[OrganizationUserId]
+                             FROM [dbo].[Collection] C
+                                 INNER JOIN #TempStep2 T ON C.[OrganizationId] = T.[OrganizationId]
+                         ) AS source
+                    LEFT JOIN [dbo].[CollectionUser] AS target
+                        ON target.[CollectionId] = source.[CollectionId] AND target.[OrganizationUserId] = source.[OrganizationUserId]
+                    WHERE target.[CollectionId] IS NULL;
```
eliykat, line 111: 💭 Why doesn't this follow the same pattern/structure as Step 1? I thought it would be pretty similar because it's the same thing but for orgUsers instead of Groups. However, this uses target/source naming and has nested SELECT queries. Is there some difference here that I'm missing or is this just a result of removing the batching logic?

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,160 @@
...
+                -- Step 3
+                    -- Update [dbo].[CollectionUser] with [Manage] = 1 using the temporary table
+                    UPDATE cu
+                    SET cu.[ReadOnly] = 0,
+                        cu.[HidePasswords] = 0,
+                        cu.[Manage] = 1
+                    FROM [dbo].[CollectionUser] cu
+                    INNER JOIN #TempStep3 temp ON cu.[OrganizationUserId] = temp.[OrganizationUserId];
+
+                    -- Insert rows to [dbo].[CollectionUser] with [Manage] = 1 using the temporary table
```
eliykat, line 128: I'd like a bit more explanation here:

```suggestion
                    -- Insert rows to [dbo].[CollectionUser] with [Manage] = 1 using the temporary table
                    -- This is for orgUsers who are Managers / EditAssignedCollections but have access via a group
                    -- We cannot give the whole group Manage permissions so we have to give them a direct assignment
```

eliykat (changes requested):
I know this has been a lot of work and revisions, thanks for putting the time in to be the expert on these changes. I have a few questions below.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,160 @@
+CREATE PROCEDURE [dbo].[Organization_EnableCollectionEnhancements]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    -- Step 1: AccessAll migration for Groups
+        -- Create a temporary table to store the groups with AccessAll = 1
+        SELECT [Id] AS [GroupId], [OrganizationId]
+        INTO #TempStep1
```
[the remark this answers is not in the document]
r-tome: That makes sense, I'm all for readability. I've renamed them.

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,160 @@
...
+            -- Step 2
+            SELECT [OrganizationUserId]
+            FROM #TempStep2
+
+            UNION
+
+            -- Step 3
+            SELECT [OrganizationUserId]
+            FROM #TempStep3
+        ) AS CombinedOrgUsers;
```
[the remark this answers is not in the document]
eliykat: TIL! thanks for confirming

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
r-tome: That makes sense and its easier to read. I have rewritten those parts here, what do you think?

```
UPDATE target
SET
    target.[ReadOnly] = 0,
    target.[HidePasswords] = 0,
    target.[Manage] = 0
FROM [dbo].[CollectionUser] AS target
INNER JOIN [dbo].[Collection] AS C ON target.[CollectionId] = C.[Id]
INNER JOIN #TempUsersAccessAll AS TU ON C.[OrganizationId] = TU.[OrganizationId] AND target.[OrganizationUserId] = TU.[OrganizationUserId];
```

```
INSERT INTO [dbo].[CollectionUser] ([CollectionId], [OrganizationUserId], [ReadOnly], [HidePasswords], [Manage])
SELECT C.[Id] AS [CollectionId], TU.[OrganizationUserId], 0, 0, 0
FROM [dbo].[Collection] C
INNER JOIN #TempUsersAccessAll TU ON C.[OrganizationId] = TU.[OrganizationId]
LEFT JOIN [dbo].[CollectionUser] target
    ON target.[CollectionId] = C.[Id] AND target.[OrganizationUserId] = TU.[OrganizationUserId]
WHERE target.[CollectionId] IS NULL;
```

@rkac-bw will perform some testing to see if there isn't any performance loss.

rkac-bw: LGTM

eliykat (dismissed):

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,153 @@
...
+                    INNER JOIN #TempUserManagers TUM ON GU.[OrganizationUserId] = TUM.[OrganizationUserId]
+                    WHERE NOT EXISTS (
+                        SELECT 1 FROM [dbo].[CollectionUser] CU
+                        WHERE CU.[CollectionId] = CG.[CollectionId] AND CU.[OrganizationUserId] = TUM.[OrganizationUserId]
+                    );
+
+                    -- Update [dbo].[OrganizationUser] to migrate all OrganizationUsers with Manager role to User role
+                    UPDATE OU
+                    SET OU.[Type] = 2 -- User
+                    FROM [dbo].[OrganizationUser] OU
```
rkac-bw, line 135: Do we need tu update revisiondate here?

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -0,0 +1,153 @@
...
+                    FROM [dbo].[Collection] C
+                    INNER JOIN #TempUsersAccessAll TU ON C.[OrganizationId] = TU.[OrganizationId]
+                    LEFT JOIN [dbo].[CollectionUser] target
+                        ON target.[CollectionId] = C.[Id] AND target.[OrganizationUserId] = TU.[OrganizationUserId]
+                    WHERE target.[CollectionId] IS NULL;
+
+                    -- Update OrganizationUser to clear AccessAll flag
+                    UPDATE OU
+                    SET [AccessAll] = 0, [RevisionDate] = GETUTCDATE()
+                    FROM [dbo].[OrganizationUser] OU
```
rkac-bw, line 109: If don't need to update organizationuser below maybe do not have to do it here

r-tome: I've updated the other missing `RevisionDate` update

rkac-bw (approved):
looks good to me

PR #3441, opened 2023-11-13
[PM-4767] update grant_save procedure

rkac-bw:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

### **Objective**
Optimize the grant_save sql and uses launchDarkly to use the original or new stored procedure which changes a merge to update insert. Should lock less and increase concurrency leading to not pinning cpu

* add new procedure to be called by AsyncSave depending on LaunchDarkly variable
* add constant and set default for LaunchDarkly variable
* add the ability to pass LaunchDarkly variable to AsyncSave and override depending if dapper or EF
* Add sql script to migration and Auth SQL scripts

### **Code changes**

* **src/Core/Auth/Repositories/IGrantRepository.cs:** Adds new entry to interface that allows SaveAsync to accept grant object and boolean
* **src/Core/Constants.cs:** Adds Boolean Constant GrantSaveOptimization set in LaunchDarkly, set to False in selfHosted
* **src/Identity/IdentityServer/PersistedGrantStore.cs:** Adds two private fields for and constructor for bring in Launchadarkly variable, adds second parameter for variable to await _grantRepository.SaveAsync
* **src/Infrastructure.Dapper/Auth/Repositories/GrantRepository.cs:**  Modifies method  method, SaveAsync, that performs asynchronous database operations related to a Grant object to accept boolean from LaunchDarkly
* **src/Infrastructure.EntityFramework/Auth/Repositories/GrantRepository.cs:** Override method to allow 2 variables even though not used
* **src/Sql/Auth/dbo/Stored Procedures/Grant_Save_withInsertUpdate.sql:** Changes merge command to deafult to update and if nothing updated inesrts
* **util/Migrator/DbScripts/2023-11-13_00_AuthGrantSaveWithInsertUpdate.sql:** Adds the optimized grant_save_WIthInserttUpdate 

## Before you submit

### DIFF
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -66,6 +66,7 @@ public static class FeatureFlagKeys
     public const string BillingPlansUpgrade = "billing-plans-upgrade";
     public const string BillingStarterPlan = "billing-starter-plan";
     public const string KeyRotationImprovements = "key-rotation-improvements";
+    public const string GrantSaveOptimization = "grant-save-optimization";
 
     public static List<string> GetAllKeys()
     {
@@ -81,7 +82,8 @@ public static class FeatureFlagKeys
         return new Dictionary<string, string>()
         {
             { TrustedDeviceEncryption, "true" },
-            { Fido2VaultCredentials, "true" }
+            { Fido2VaultCredentials, "true" },
+            { GrantSaveOptimization, "false"}
         };
     }
 }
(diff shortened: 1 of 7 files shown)


src/Core/Constants.cs
```diff
@@ -81,7 +82,8 @@ public static List<string> GetAllKeys()
         return new Dictionary<string, string>()
         {
             { TrustedDeviceEncryption, "true" },
-            { Fido2VaultCredentials, "true" }
+            { Fido2VaultCredentials, "true" },
+            { GrantSaveOptimization, "false"}
```
withinfocus, line 86: ❌ This is a strange section of code, but is for overrides where we need to force non-default usage of feature flags for use cases like self-hosting. You won't need this change.

ike-kottlowski (approved):

withinfocus (changes requested):
Robert's adjusting a few things in the background.

PR #3436, opened 2023-11-10
[PM-3797 Part 3] Add vault domains to key rotation

jlf0dev:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Previous PR: #3434
Adds ciphers and folders to the new key rotation.

## Code changes

* **file.ext:** Description of what was changed and why

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3442, opened 2023-11-13
[PM-3797 Part 4] Add Sends to new Key Rotation

jlf0dev:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Add sends to new key rotation.

Previous PR: #3436
See Part 1 here for background and context: #3425

## Code changes

## Before you submit

### DIFF
--- a/src/Api/Tools/Validators/SendRotationValidator.cs
+++ b/src/Api/Tools/Validators/SendRotationValidator.cs
@@ -1,0 +1,49 @@
+﻿using Bit.Api.Auth.Validators;
+using Bit.Api.Tools.Models.Request;
+using Bit.Core.Entities;
+using Bit.Core.Exceptions;
+using Bit.Core.Tools.Entities;
+using Bit.Core.Tools.Repositories;
+using Bit.Core.Tools.Services;
+
+namespace Bit.Api.Tools.Validators;
+
+public class SendRotationValidator : IRotationValidator<IEnumerable<SendWithIdRequestModel>, IEnumerable<Send>>
+{
+    private readonly ISendService _sendService;
+    private readonly ISendRepository _sendRepository;
+
+    public SendRotationValidator(ISendService sendService, ISendRepository sendRepository)
+    {
+        _sendService = sendService;
+        _sendRepository = sendRepository;
+    }
+
+    public async Task<IEnumerable<Send>> ValidateAsync(User user, IEnumerable<SendWithIdRequestModel> sends)
+    {
+        var result = new List<Send>();
+        if (sends == null || !sends.Any())
+        {
+            return result;
+        }
+
+        var existingSends = await _sendRepository.GetManyByUserIdAsync(user.Id);
+        if (existingSends == null || !existingSends.Any())
+        {
+            return result;
+        }
+
+        foreach (var existing in existingSends)
+        {
+            var send = sends.FirstOrDefault(c => c.Id == existing.Id);
+            if (send == null)
+            {
+                throw new BadRequestException("All existing folders must be included in the rotation.");
+            }
+
+            result.Add(send.ToSend(existing, _sendService));
+        }
+
+        return result;
+    }
+}
--- a/src/Api/Tools/Validators/SendRotationValidator.cs
+++ b/src/Api/Tools/Validators/SendRotationValidator.cs
@@ -8,18 +8,26 @@ using Bit.Core.Tools.Services;
 
 namespace Bit.Api.Tools.Validators;
 
-public class SendRotationValidator : IRotationValidator<IEnumerable<SendWithIdRequestModel>, IEnumerable<Send>>
+/// <summary>
+/// Send implementation for <see cref="IRotationValidator{T,R}"/>
+/// </summary>
+public class SendRotationValidator : IRotationValidator<IEnumerable<SendWithIdRequestModel>, IReadOnlyList<Send>>
 {
     private readonly ISendService _sendService;
     private readonly ISendRepository _sendRepository;
 
+    /// <summary>
+    /// Instantiates a new <see cref="SendRotationValidator"/>
+    /// </summary>
+    /// <param name="sendService">Enables conversion of <see cref="SendWithIdRequestModel"/> to <see cref="Send"/></param>
+    /// <param name="sendRepository">Retrieves all user <see cref="Send"/>s</param>
     public SendRotationValidator(ISendService sendService, ISendRepository sendRepository)
     {
         _sendService = sendService;
         _sendRepository = sendRepository;
     }
 
-    public async Task<IEnumerable<Send>> ValidateAsync(User user, IEnumerable<SendWithIdRequestModel> sends)
+    public async Task<IReadOnlyList<Send>> ValidateAsync(User user, IEnumerable<SendWithIdRequestModel> sends)
     {
         var result = new List<Send>();
         if (sends == null || !sends.Any())
--- a/src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
+++ b/src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Identity;
 
 namespace Bit.Core.Auth.UserFeatures.UserKey.Implementations;
 
+/// <inheritdoc />
 public class RotateUserKeyCommand : IRotateUserKeyCommand
 {
     private readonly IUserService _userService;
@@ -19,6 +20,17 @@ public class RotateUserKeyCommand : IRotateUserKeyCommand
     private readonly IPushNotificationService _pushService;
     private readonly IdentityErrorDescriber _identityErrorDescriber;
 
+    /// <summary>
+    /// Instantiates a new <see cref="RotateUserKeyCommand"/>
+    /// </summary>
+    /// <param name="userService">Master password hash validation</param>
+    /// <param name="userRepository">Updates user keys and re-encrypted data if needed</param>
+    /// <param name="cipherRepository">Provides a method to update re-encrypted cipher data</param>
+    /// <param name="folderRepository">Provides a method to update re-encrypted folder data</param>
+    /// <param name="sendRepository">Provides a method to update re-encrypted send data</param>
+    /// <param name="emergencyAccessRepository">Provides a method to update re-encrypted emergency access data</param>
+    /// <param name="pushService">Logs out user from other devices after successful rotation</param>
+    /// <param name="errors">Provides a password mismatch error if master password hash validation fails</param>
     public RotateUserKeyCommand(IUserService userService, IUserRepository userRepository,
         ICipherRepository cipherRepository, IFolderRepository folderRepository, ISendRepository sendRepository,
         IEmergencyAccessRepository emergencyAccessRepository,
--- a/src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
+++ b/src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
@@ -2,6 +2,7 @@
 using Bit.Core.Entities;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
+using Bit.Core.Tools.Repositories;
 using Bit.Core.Vault.Repositories;
 using Microsoft.AspNetCore.Identity;
 
@@ -13,12 +14,13 @@ public class RotateUserKeyCommand : IRotateUserKeyCommand
     private readonly IUserRepository _userRepository;
     private readonly ICipherRepository _cipherRepository;
     private readonly IFolderRepository _folderRepository;
+    private readonly ISendRepository _sendRepository;
     private readonly IEmergencyAccessRepository _emergencyAccessRepository;
     private readonly IPushNotificationService _pushService;
     private readonly IdentityErrorDescriber _identityErrorDescriber;
 
     public RotateUserKeyCommand(IUserService userService, IUserRepository userRepository,
-        ICipherRepository cipherRepository, IFolderRepository folderRepository,
+        ICipherRepository cipherRepository, IFolderRepository folderRepository, ISendRepository sendRepository,
         IEmergencyAccessRepository emergencyAccessRepository,
         IPushNotificationService pushService, IdentityErrorDescriber errors)
     {
@@ -26,6 +28,7 @@ public class RotateUserKeyCommand : IRotateUserKeyCommand
         _userRepository = userRepository;
         _cipherRepository = cipherRepository;
         _folderRepository = folderRepository;
+        _sendRepository = sendRepository;
         _emergencyAccessRepository = emergencyAccessRepository;
         _pushService = pushService;
         _identityErrorDescriber = errors;
@@ -64,6 +67,12 @@ public class RotateUserKeyCommand : IRotateUserKeyCommand
             {
                 saveEncryptedDataActions.Add(_folderRepository.UpdateForKeyRotation(user.Id, model.Folders));
             }
+
+            if (model.Sends.Any())
+            {
+                saveEncryptedDataActions.Add(_sendRepository.UpdateForKeyRotation(user.Id, model.Sends));
+            }
+
             if (model.EmergencyAccessKeys.Any())
             {
                 saveEncryptedDataActions.Add(
--- a/src/Core/Tools/Repositories/ISendRepository.cs
+++ b/src/Core/Tools/Repositories/ISendRepository.cs
@@ -1,5 +1,6 @@
-﻿#nullable enable
+#nullable enable
 
+using Bit.Core.Auth.UserFeatures.UserKey;
 using Bit.Core.Repositories;
 using Bit.Core.Tools.Entities;
 
@@ -33,4 +34,12 @@ public interface ISendRepository : IRepository<Send, Guid>
     /// The task's result contains the loaded <see cref="Send"/>s.
     /// </returns>
     Task<ICollection<Send>> GetManyByDeletionDateAsync(DateTime deletionDateBefore);
+
+    /// <summary>
+    /// Updates encrypted data for sends during a key rotation
+    /// </summary>
+    /// <param name="userId">The user that initiated the key rotation</param>
+    /// <param name="sends">A list of sends with updated data</param>
+    UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
+        IEnumerable<Send> ciphers);
 }
--- a/src/Core/Tools/Repositories/ISendRepository.cs
+++ b/src/Core/Tools/Repositories/ISendRepository.cs
@@ -1,4 +1,4 @@
-#nullable enable
+﻿#nullable enable
 
 using Bit.Core.Auth.UserFeatures.UserKey;
 using Bit.Core.Repositories;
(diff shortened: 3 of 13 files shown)


jlf0dev:
Current dependencies on/for this PR:
* master
  * **PR #3434** <a href="https://app.graphite.dev/github/pr/bitwarden/server/3434?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #3436** <a href="https://app.graphite.dev/github/pr/bitwarden/server/3436?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #3442** <a href="https://app.graphite.dev/github/pr/bitwarden/server/3442?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
        * **PR #3445** <a href="https://app.graphite.dev/github/pr/bitwarden/server/3445?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/bitwarden/server/3442?utm_source=stack-comment).

ike-kottlowski (changes requested):
🎨 : Could you add happy path tests? Ensuring it meets the failure criteria is important, but I think the happy path tests also build confidence in the codebase.

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
```diff
@@ -69,4 +70,28 @@ public async Task<ICollection<Core.Tools.Entities.Send>> GetManyByUserIdAsync(Gu
...
+            var userSends = await GetDbSet(dbContext)
+                .Where(s => s.UserId == userId)
+                .ToListAsync();
+            var validSends = userSends
+                .Where(send => newSends.Any(newSend => newSend.Id == send.Id));
+            foreach (var send in validSends)
+            {
+                var updateSend = newSends.First(newSend => newSend.Id == send.Id);
+                send.Key = updateSend.Key;
+            }
```
audrey-jensen, line 91: 🎨 If you turn newSends into a dictionary, this would be a key lookup instead of a linear search.

```cs
var newSends = sends.ToDictionary(s => s.Id);

// ... in foreach
send.Key = newSends[send.Id].Key
```

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using Bit.Test.Common.AutoFixture.Attributes;
+using NSubstitute;
+using Xunit;
+
+namespace Bit.Api.Test.Tools.Validators;
+
+[SutProviderCustomize]
+public class SendRotationValidatorTests
+{
+    [Theory, BitAutoData]
```
audrey-jensen, line 17: ❓ Where do I go to find the test data this is running?

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -48,4 +50,56 @@ public async Task<ICollection<Send>> GetManyByDeletionDateAsync(DateTime deletio
...
+                bulkCopy.DestinationTableName = "#TempSend";
+                var sendsTable = sends.ToDataTable();
+                foreach (DataColumn col in sendsTable.Columns)
+                {
+                    bulkCopy.ColumnMappings.Add(col.ColumnName, col.ColumnName);
+                }
+
+                sendsTable.PrimaryKey = new DataColumn[] { sendsTable.Columns[0] };
+                await bulkCopy.WriteToServerAsync(sendsTable);
+            }
```
audrey-jensen, line 81: 👍🏻 Been a long time since I've seen plain old ADO.Net in use, and this brings me back. I used to have this whole toolkit for splicing together strong types with the type erased world of ADO, and it had lots of code that looked like this, including the lambda-returning business for being able to splice together various things. It also included fancy bits like attributes to signal primary keys, an in-built dynamic pivot engine, and so on. I had a lot of time on my hands back then.

Super happy I picked up this review. 😄

audrey-jensen (changes requested):
- 👍 (:+1:) for great changes
- 📝 (:memo:) for notes or general info
- ❓ (:question:) for questions
- 🎨 (:art:) for suggestions / improvements
- ⚠️ (:warning:) for concerns needing attention
- 🌱 (:seedling:) for future improvements or indications of technical debt

src/Api/Tools/Validators/SendRotationValidator.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public SendRotationValidator(ISendService sendService, ISendRepository sendRepository)
+    {
+        _sendService = sendService;
+        _sendRepository = sendRepository;
+    }
+
+    public async Task<IEnumerable<Send>> ValidateAsync(User user, IEnumerable<SendWithIdRequestModel> sends)
+    {
+        var result = new List<Send>();
```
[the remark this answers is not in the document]
jlf0dev: Done!

src/Api/Tools/Validators/SendRotationValidator.cs
```diff
@@ -0,0 +1,49 @@
...
+        foreach (var existing in existingSends)
+        {
+            var send = sends.FirstOrDefault(c => c.Id == existing.Id);
+            if (send == null)
+            {
+                throw new BadRequestException("All existing folders must be included in the rotation.");
+            }
+
+            result.Add(send.ToSend(existing, _sendService));
+        }
```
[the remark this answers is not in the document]
jlf0dev: I can't think of a great way to accomplish this since the lazy evaluation means the exception isn't actually being thrown

src/Api/Tools/Validators/SendRotationValidator.cs
```diff
@@ -0,0 +1,49 @@
...
+    private readonly ISendService _sendService;
+    private readonly ISendRepository _sendRepository;
+
+    public SendRotationValidator(ISendService sendService, ISendRepository sendRepository)
+    {
+        _sendService = sendService;
+        _sendRepository = sendRepository;
+    }
+
+    public async Task<IEnumerable<Send>> ValidateAsync(User user, IEnumerable<SendWithIdRequestModel> sends)
```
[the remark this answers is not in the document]
jlf0dev: Added!

src/Core/Tools/Repositories/ISendRepository.cs
```diff
@@ -33,4 +34,12 @@ public interface ISendRepository : IRepository<Send, Guid>
     /// The task's result contains the loaded <see cref="Send"/>s.
     /// </returns>
     Task<ICollection<Send>> GetManyByDeletionDateAsync(DateTime deletionDateBefore);
+
+    /// <summary>
+    /// Updates encrypted data for sends during a key rotation
+    /// </summary>
+    /// <param name="userId">The user that initiated the key rotation</param>
+    /// <param name="sends">A list of sends with updated data</param>
+    UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
```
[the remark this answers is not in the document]
jlf0dev: I don't throw anything special in here, not sure what to call out.

src/Infrastructure.Dapper/Tools/Helpers/SendHelpers.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Data;
+using Bit.Core.Tools.Entities;
+
+namespace Bit.Infrastructure.Dapper.Tools.Helpers;
+
+public static class SendHelpers
```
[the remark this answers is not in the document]
jlf0dev: Good callout, added

src/Infrastructure.Dapper/Tools/Helpers/SendHelpers.cs
```diff
@@ -0,0 +1,33 @@
...
+            (nameof(Send.Key), typeof(string), c => c.Key),
+            (nameof(Send.Password), typeof(string), c => c.Password),
+            (nameof(Send.MaxAccessCount), typeof(int), c => c.MaxAccessCount),
+            (nameof(Send.AccessCount), typeof(int), c => c.AccessCount),
+            (nameof(Send.CreationDate), typeof(DateTime), c => c.CreationDate),
+            (nameof(Send.RevisionDate), typeof(DateTime), c => c.RevisionDate),
+            (nameof(Send.ExpirationDate), typeof(DateTime), c => c.ExpirationDate),
+            (nameof(Send.DeletionDate), typeof(DateTime), c => c.DeletionDate),
+            (nameof(Send.Disabled), typeof(bool), c => c.Disabled),
+            (nameof(Send.HideEmail), typeof(bool), c => c.HideEmail),
```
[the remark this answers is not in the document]
jlf0dev: Yep, been thinking the same. Just felt like this kind of enhancement isn't warranted yet. Cool sample though!

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
jlf0dev: Good catch!

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
jlf0dev: The data is provided by AutoFixture through the attributes. Which means its auto-generated with some customization provided. AutoFixture has been pretty confusing for me to work with and I think we're going to move away from it in the future.
Most of the files for this is in `test/Common/AutoFixture`

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
jlf0dev: Sounds like quite the package! Most of this is pretty old code that I'm keeping around because it works for our use case (if it ain't broke...)

src/Api/Tools/Validators/SendRotationValidator.cs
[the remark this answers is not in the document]
audrey-jensen: Methods that accept and return IEnumerable<T> should be assumed to be lazy. If you want to eagerly evaluate it, then return an `IReadOnlyList<T>` or `IReadOnlyCollection<T>` from the method. That way the consumer knows its eagerly evaluated and they can do things like index it.

src/Core/Tools/Repositories/ISendRepository.cs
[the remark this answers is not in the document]
audrey-jensen: Does it throw the `BadRequestException`? Based on the name I'd assume it does, albeit indirectly.

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
audrey-jensen: Without the data, there's no way to understand what is being tested. In a worst case scenario, the customizations end up replicating the computation under test.  Using a fixed data set is a crucial part of the test design when the system under test is deterministic. If the system is not deterministic or this is a specialized case--a fuzz test, for example--then automatically generated data is completely cool. 

⚠️ As written, I cannot understand what these tests are actually testing. We should probably huddle about how to address that instead of continuing here.

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
audrey-jensen: > if it ain't broke... 

Then you probably aren't looking deep enough. ;)

src/Infrastructure.Dapper/Tools/Helpers/SendHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+/// Dapper helper methods for Sends
+/// </summary>
+public static class SendHelpers
+{
+    /// <summary>
+    /// Converts an IEnumerable of Sends to a DataTable
+    /// Note: Contains a hardcoded list of properties and must be updated with model
+    /// </summary>
+    /// <param name="sends">List of sends</param>
+    /// <returns><see cref="DataTable"/></returns>
```
audrey-jensen, line 16: ⛏️ I'd be more descriptive. "A data table matching the schema of `dbo.Send` containing one row mapped from the items in `sends`" (with the appropriate refs ofc.)

src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
```diff
@@ -2,30 +2,34 @@
...
     private readonly IUserService _userService;
     private readonly IUserRepository _userRepository;
     private readonly ICipherRepository _cipherRepository;
     private readonly IFolderRepository _folderRepository;
+    private readonly ISendRepository _sendRepository;
     private readonly IEmergencyAccessRepository _emergencyAccessRepository;
     private readonly IPushNotificationService _pushService;
     private readonly IdentityErrorDescriber _identityErrorDescriber;
 
     public RotateUserKeyCommand(IUserService userService, IUserRepository userRepository,
```
audrey-jensen, line 23: 📝 Here's a place not in the tools code where a public member isn't documented...

src/Api/Tools/Validators/SendRotationValidator.cs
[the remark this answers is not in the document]
jlf0dev: Should have specified that the `ValidateAsync` inherits the documentation from the interface. Adding documentation for each of these feels overkill since its purpose is basically the same.

Not sure I see the benefit of adding documentation to a constructor in this situation, since DI handles the instantiation.

src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
jlf0dev: Same answer here about constructors, not sure about the value gained

src/Api/Tools/Validators/SendRotationValidator.cs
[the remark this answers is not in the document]
jlf0dev: Documentation added to constructor

src/Api/Tools/Validators/SendRotationValidator.cs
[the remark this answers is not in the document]
jlf0dev: Added. We don't need those kinds of operations for the rest of the rotation, but I can see the logic. The rest of the domains will be changed in future PRs.

src/Infrastructure.Dapper/Tools/Helpers/SendHelpers.cs
jlf0dev: Can't figure out how to reference the file, but changed

src/Infrastructure.Dapper/Tools/Helpers/SendHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Bit.Infrastructure.Dapper.Tools.Helpers;
+
+/// <summary>
+/// Dapper helper methods for Sends
+/// </summary>
+public static class SendHelpers
+{
+    /// <summary>
+    /// Converts an IEnumerable of Sends to a DataTable
+    /// Note: Contains a hardcoded list of properties and must be updated with model
```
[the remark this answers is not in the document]
jlf0dev: Done!

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
jlf0dev: Rewritten the tests to be deterministic. Please consider alerting the rest of the engineering team of these requirements for Tools code.

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        var sendRepository = Substitute.For<ISendRepository>();
+
+        var sut = new SendRotationValidator(
+            sendService,
+            sendRepository
+        );
+
+        var user = new User { Id = new Guid() };
+        var sends = CreateSends();
+        var userSends = sends.Select(s =>
```
audrey-jensen, line 34: ⚠️  This line is performing a computation, and then used as part of an assertion on line 42. Use a hard-coded expected result instead.

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            new Send { Id = s.Id.Value, Type = s.Type, Data = "{}" }).ToList();
+        userSends.Add(new Send { Id = Guid.NewGuid(), Data = "{}" });
+        sendRepository.GetManyByUserIdAsync(user.Id).Returns(userSends);
```
audrey-jensen, line 62: 🎨 This could be more efficiently arranged by using an empty set and simply checking to see if any result causes ValidateAsync to throw. If at least one item is necessary, then just set up a single send to be returned. This also eliminates the need to perform computation that relies on the input data in the test.

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        var sut = new SendRotationValidator(
+            sendService,
+            sendRepository
+        );
+
+        var user = new User { Id = new Guid() };
+        var sends = CreateSends();
+        var userSends = sends.Select(s =>
+            new Send { Id = s.Id.Value, Type = s.Type, Data = "{}" }).ToList();
+        userSends.RemoveAt(0);
```
audrey-jensen, line 84: ⛏️  This is not checking whether the send belongs to the user, since its userId hasn't been set up to be different than the input userId--in fact, none of the sends have their userid set. Instead, it's checking that `GetManyByUserId` no longer lists the send, which could happen for many reasons--for example, the send expired and was deleted. The name of the test case should be updated to reflect the actual state of the test: that the send is no longer in the system.

⚠️ The mock data should not be derived from the input data. In a best-case scenario, the test will be brittle and fail when the code changes. In a worst-case, it'll flag a false-positive and the validation will allow records it shouldn't. Consider hard-coding the stub's return values instead of deriving them from the input.

test/Api.Test/Tools/Validators/SendRotationValidatorTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                DeletionDate = DateTime.UtcNow.AddDays(20),
+                Disabled = false,
+                Id = new Guid("9a65bbfb-8138-4aa5-a572-e5c0a41b540e"),
+                Key = "Send3Key",
+                Name = "Send 3",
+                Type = SendType.File,
+                File = new SendFileModel(new SendFileData("File name", "Notes", "File name here")),
+                HideEmail = true
+            }
```
audrey-jensen, line 127: 📝  This data set is being used as a test input, to derive the mock output, and in one case to derive the expected result of the test. These three cases should be treated separately. More precise guidance for each case is provided above.

audrey-jensen (changes requested):
The unit tests need to be revised so that the mocks, inputs, and outputs are not derived from each other.

👍🏻 That said, I'd like to point out that this behavior was also in the original PR, but it wasn't clear to me that it was problematic until the autofixture logic was removed from the test. So, major props for following through on embedding the test data. Without that I'd not have been able to see the antipattern. :)

src/Infrastructure.Dapper/Tools/Helpers/SendHelpers.cs
audrey-jensen: I don't think there's a way to ref the file, unfortunately. :(

audrey-jensen (approved):
Thank you for working with me to improve the documentation and unit tests! There was a bit of back and forth, and I appreciate you stuck with it! 😄

ike-kottlowski (approved):

PR #3446, opened 2023-11-14
Prefer compounds for preLaunchTasks

MGibson1:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Currently, our configurations define `preLaunchTasks`, which are then run concurrently for compound launch configurations. `dotnet` does not do well with parallel builds due to common dependencies that lock files and cause intermittent failures on deploy of `preLauchTasks`. This results in running a `full server` configurations, some fraction of the projects failing to build, and ending up with only `SSO` and `Billing` running. Very annoying.

This rework creates `configurations` that have no build task associated to just run existing code, hides them from the debugging interface, then builds compound tasks that both build and run each project. It also defined `tasks` that sequentially builds each project for a given compound.

I also got rid of the order part of our configurations to prefer alphabetical. This is really just the order the compounds are defined, but I put them in alphabetical order.

## Before you submit

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,94 +7,251 @@
         {
             "name": "Min Server",
             "configurations": [
-                "Identity",
-                "API"
+                "run-Identity",
+                "run-API"
             ],
             "presentation": {
                 "hidden": false,
                 "group": "AA_compounds",
                 "order": 1
             },
+            "preLaunchTask": "buildIdentityApi",
             "stopAll": true
         },
         {
             "name": "Admin, API, Identity",
             "configurations": [
-                "Admin",
-                "API",
-                "Identity"
+                "run-Admin",
+                "run-API",
+                "run-Identity"
             ],
             "presentation": {
                 "hidden": false,
                 "group": "AA_compounds",
                 "order": 3
             },
+            "preLaunchTask": "buildIdentityApiAdmin",
             "stopAll": true
         },
         {
             "name": "Full Server",
             "configurations": [
-                "Admin",
-                "API",
-                "EventsProcessor",
-                "Identity",
-                "Sso",
-                "Icons",
-                "Billing",
-                "Notifications"
+                "run-Admin",
+                "run-API",
+                "run-EventsProcessor",
+                "run-Identity",
+                "run-Sso",
+                "run-Icons",
+                "run-Billing",
+                "run-Notifications"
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -7,94 +7,251 @@
...
+            ],
+            "presentation": {
+                "hidden": false,
+                "group": "self-host",
+            },
+            "preLaunchTask": "buildSso",
+        },
+    ],
+    "configurations": [
+        // Configurations represent run-only scenarios so that they can be used in multiple compounds
```
withinfocus, line 246: ❓ Invalid to have a comment here right?

.vscode/launch.json
```diff
@@ -7,94 +7,251 @@
...
+                "run-Sso"
+            ],
+            "presentation": {
+                "hidden": false,
+                "group": "cloud",
+            },
+            "preLaunchTask": "buildSso",
+        },
+        {
+            "name": "Admin Self Host",
```
withinfocus, line 179: ❓ Is there anything VS provides to group self-host vs. not perhaps?

.vscode/tasks.json
```diff
@@ -1,6 +1,65 @@
...
+                "buildIdentity",
+                "buildAPI",
+                "buildAdmin"
+            ],
+            "problemMatcher": [
+                "$msCompile"
+            ]
+        },
+        {
+            "label": "buildFullServer",
```
withinfocus, line 28: ❓ I may not be understanding the problem fully, but could it be valuable to always build the whole solution and then run components? What downstream complications might happen with partial builds?

.vscode/launch.json
MGibson1: for json, technically, yes, but vscode is fine with comments in json.

.vscode/launch.json
MGibson1: These will be grouped in the UI due to the `presentation` field, but there is no grouping at the json level, no.

.vscode/tasks.json
MGibson1: The same complications as running partial servers. There's no point in building a project that you don't run. Dependencies of projects will still be built, it's just that, for example, SSO is left out of the build when you don't start it up.

withinfocus (approved):

PR #3453, opened 2023-11-16
[AC-1782] Import can manage

aj-rosado:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
Users should also be able to import into some collections

## Code changes

* **ImportCipherControllers.cs:** Added validation to verify if an user can create a collection in the FlexibleCollections context

## Before you submit


rkac-bw (approved):
Looks good to me

src/Api/Tools/Controllers/ImportCiphersController.cs
```diff
@@ -64,14 +74,46 @@ public async Task PostImport([FromBody] ImportCiphersRequestModel model)
...
+        var orgCollectionIds =
+            (await _collectionRepository.GetManyByOrganizationIdAsync(orgId))
+            .Select(c => c.Id)
+            .ToHashSet();
+
+        //We need to verify if the user is trying to import into existing collections
+        var existingCollections = collections.Where(tc => orgCollectionIds.Contains(tc.Id));
+
+        //When importing into existing collection, we need to verify if the user has permissions
+        if (existingCollections.Count() > 0 && !(await _authorizationService.AuthorizeAsync(User, existingCollections, BulkCollectionOperations.ImportCiphers)).Succeeded)
```
djsmith85, line 112: This should be using `.Any()`instead of `.Count()`

src/Api/Tools/Controllers/ImportCiphersController.cs
```diff
@@ -64,14 +74,46 @@ public async Task PostImport([FromBody] ImportCiphersRequestModel model)
...
         var orgId = new Guid(organizationId);
-        if (!await _currentContext.AccessImportExport(orgId))
+        var collections = model.Collections.Select(c => c.ToCollection(orgId)).ToList();
+        var orgFlexibleCollections = (await _organizationService.GetByIdAsync(orgId)).FlexibleCollections;
+
+
+        //An User is allowed to import if CanCreate Collections or has AccessToImportExport
+        var authorized = orgFlexibleCollections
+            ? await CheckOrgImportPermission(collections, orgId) || await _currentContext.AccessImportExport(orgId)
+            : await _currentContext.AccessImportExport(orgId);
```
djsmith85, line 84: This can be simplified by pushing the `_currentContext.AccessImportExport(orgId)`-check into `CheckOrgImportPermission(collections, orgId)` and using some early returns.

djsmith85 (changes requested):

src/Api/Tools/Controllers/ImportCiphersController.cs
```diff
@@ -64,14 +74,46 @@ public async Task PostImport([FromBody] ImportCiphersRequestModel model)
...
             throw new NotFoundException();
         }
 
         var userId = _userService.GetProperUserId(User).Value;
-        var collections = model.Collections.Select(c => c.ToCollection(orgId)).ToList();
         var ciphers = model.Ciphers.Select(l => l.ToOrganizationCipherDetails(orgId)).ToList();
         await _cipherService.ImportCiphersAsync(collections, ciphers, model.CollectionRelationships, userId);
     }
+
+    private async Task<bool> CheckOrgImportPermission(List<Collection> collections, Guid orgId)
```
vincentsalucci, line 96: Hmm. I feel like this method should be contained within the authorization handler check for being able to `ImportCiphers`.  As of right now, that only covers the `editing` of collections if we're importing into existing collections. 

@eliykat @shane-melton @djsmith85 :
Does it make more sense for the `ImportCiphers` operation to handle all necessary conditions (ex: creating new collections **and** editing collections (if necessary)). It should also remove the necessity to check the `CurrentContext.AccessImportExport` since that should be flattened into the above operation as well.

aj-rosado: The `AuthorizationHandler` needs to receive collections otherwise it fails and import only have existing collections when an user is trying to import into existing collections.
If the user is importing and creating new collections or importing without collections, it fails.
This is the reason I could not use the `AuthorizationHandler` for everything

aj-rosado: Adding some more context to my previous commment.
On the `BulkCollectionAuthorizationHandler.HandleRequirementAsync`

`
if (resources == null || !resources.Any())
        {
            context.Fail();
            return;
        }
`

If we don't pass any collection as parameter (or an empty list) it will fail.

We have 3 different cases to check here.

1. From AdminConsole importer is possible to import ciphers without collections, that cases would return success on the first validation
`await _currentContext.AccessImportExport(orgId)` 

2. If the user fails that validation, he might still be able to import in case os FlexibleCollections if he can Create new Collections. So we check the `BulkCollectionOperations.Create`. We pass all the collections as we only check if he can create, if there is no collection, this should fail.

3. If that user is trying to import into existing collections, he is not creating them but adding items into them. Se we need to check if the user is importing into collections that he manages. We can't pass all the collections as if they don't exist, it would fail on the manage permission for that collections.

I initially tried to have all validations on the `BulkCollectionAuthorizationHandler` but wasn't able to find a way to validate all this conditions. 
What might could work, could be to create a `ImportAuthorizationHandler`? But it still would need to call the `BulkCollectionAuthorizationHandler`

r-tome (commented):
I left a comment with a minor suggestion, otherwise it looks great!

r-tome (approved):
LGTM!

djsmith85 (approved):

Jingo88 (approved):

PR #3379, opened 2023-10-25
[AC-1122] Add AllowAdminAccessToAllCollectionItems setting to Organizations

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective
Create a new collection management setting to allow owner/admins access to all collections and items (`AllowAdminAccessToAllCollectionItems`). The setting will default to enabled for existing organizations to preserve current behavior. Only once the flexible collections feature flag is enabled will newly created organizations default to disabled.

This PR only adds the column to the database and adds API support for toggling the setting; it does enact any behavior changes.

Related Clients PR:

## Code changes

* **src/Core/Entities/Organization.cs:** Add the new property to the Organization entity
* **src/Core/Services/Implementations/OrganizationService.cs**: Ensure newly created orgs have the correct default value for the setting depending on the feature flag
* **Organization Sprocs and Views:** Add the new property to the Create/Update sprocs and Details View
* **util/Migrator/DbScripts/2023-10-24_04_AdminCollectionItemAccess.sql:** Add the migration script
* **EF Migrations**
* **API Request/Response Models:** Add the new property
* **src/Core/Context/CurrentContextOrganization.cs:** Add the new property so its available for future logic to enforce the setting.

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3458, opened 2023-11-20
[AC-1741] Include owners/admins can manage all collections setting in license file

gbubemismith:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Collection Management Settings should be added to the license file and should be updated in the self-hosted instance when the license file is updated.

This PR is dependent on https://github.com/bitwarden/server/pull/3379

## Code changes

* **src/Api/AdminConsole/Controllers/OrganizationsController.cs:** Updated endpoint to allow only cloud access.
* **src/Core/Entities/Organization.cs:** Added `AllowAdminAccessToAllCollectionItems` from license.
* **src/Core/Models/Business/OrganizationLicense.cs:** Added `AllowAdminAccessToAllCollectionItems` property to OrganizationLicense and incremented version numbering
* **src/Core/Services/Implementations/OrganizationService.cs** Sets property if the feature flag is enabled or defaults to true

## Before you submit


eliykat:
Billing Team should own license files so I've requested a review from them, please treat them as codeowners as well.

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -145,10 +147,10 @@ public OrganizationLicense()
     /// </summary>
     /// <remarks>Intentionally set one version behind to allow self hosted users some time to update before
```
cyprain-okeke, line 148: Is this comment still validate or we need to update it?

cyprain-okeke (commented):
Looks Good

src/Api/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -782,6 +781,7 @@ public async Task<OrganizationSsoResponseModel> PostSso(Guid id, [FromBody] Orga
 
     [HttpPut("{id}/collection-management")]
     [RequireFeature(FeatureFlagKeys.FlexibleCollections)]
+    [SelfHosted(NotSelfHostedOnly = true)]
```
cyprain-okeke, line 784: should the user be able to change this setting on the self-hosted instance?

gbubemismith: Only cloud access is required of this

src/Core/Models/Business/OrganizationLicense.cs
gbubemismith: This is still valid

eliykat:
This needs to wait until #3388 is merged, which adds the first collection management setting (limit collection creation/deletion) to the license file. Currently, you're both trying to bump the license version at the same time. And both PRs should wait until unit tests in #3474 are merged.

util/Migrator/DbScripts/2023-10-24_04_AdminCollectionItemAccess.sql
eliykat, line 1: Looks like this was accidentally included

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -348,6 +352,11 @@ public bool VerifyData(Organization organization, IGlobalSettings globalSettings
             //     valid = organization.LimitCollectionCreationDeletion == LimitCollectionCreationDeletion;
             // }
 
+            if (valid && Version >= 14)
+            {
+                valid = organization.AllowAdminAccessToAllCollectionItems == AllowAdminAccessToAllCollectionItems;
+            }
```
eliykat, line 358: Should be commented our per the version 13 validation above.

This is because self-hosted orgs do not use this value unless the feature flag is turned on in the self-hosted instance, therefore we expect this check to fail in some circumstances. (e.g. cloud has the flag on, client updates the setting and then updates their license file, but the flag is off in self-hosted. The license value will be ignored, the org value won't match the license file, so this validation would otherwise fail.)

We can do this because validation of this property is not particularly important either (it's not billable).

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -137,6 +138,7 @@ public OrganizationLicense()
     public int? SmSeats { get; set; }
     public int? SmServiceAccounts { get; set; }
     public bool LimitCollectionCreationDeletion { get; set; } = true;
+    public bool AllowAdminAccessToAllCollectionItems { get; set; }
```
eliykat, line 141: Needs to default to `true`. This covers the case where a client loads an old license file which doesn't have this value.

eliykat (changes requested):

cturnbull-bitwarden (approved):

src/Core/AdminConsole/Entities/Organization.cs
```diff
@@ -268,5 +268,6 @@ public void UpdateFromLicense(OrganizationLicense license, bool flexibleCollecti
         SmSeats = license.SmSeats;
         SmServiceAccounts = license.SmServiceAccounts;
         LimitCollectionCreationDeletion = !flexibleCollectionsIsEnabled || license.LimitCollectionCreationDeletion;
+        AllowAdminAccessToAllCollectionItems = !flexibleCollectionsIsEnabled || license.AllowAdminAccessToAllCollectionItems;
```
eliykat, line 271: The naming here isn't very clear, sorry. `flexibleCollectionsIsEnabled` refers to the MVP feature flag. But this is a v1 feature. So you'll need to pass in both:

```cs
public void UpdateFromLicense(OrganizationLicense license, bool flexibleCollectionsMvpIsEnabled, bool flexibleCollectionsV1IsEnabled)
{
   // snip
       LimitCollectionCreationDeletion = !flexibleCollectionsMvpIsEnabled || license.LimitCollectionCreationDeletion;
       AllowAdminAccessToAllCollectionItems = !flexibleCollectionsV1IsEnabled || license.AllowAdminAccessToAllCollectionItems;
```

It's awkward but we'll clean it up after release.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -601,7 +601,8 @@ private async Task ValidateSignUpPoliciesAsync(Guid ownerId)
             UseSecretsManager = license.UseSecretsManager,
             SmSeats = license.SmSeats,
             SmServiceAccounts = license.SmServiceAccounts,
-            LimitCollectionCreationDeletion = !flexibleCollectionsIsEnabled || license.LimitCollectionCreationDeletion
+            LimitCollectionCreationDeletion = !flexibleCollectionsIsEnabled || license.LimitCollectionCreationDeletion,
+            AllowAdminAccessToAllCollectionItems = !flexibleCollectionsIsEnabled || license.AllowAdminAccessToAllCollectionItems
```
eliykat, line 605: Same here, `AllowAdminAccessToAllCollectionItems` should use the v1 feature flag, and please rename existing variable to clearly distinguish between them.

`LimitCollectionCreationDeletion` should continue to use the mvp feature flag.

eliykat (changes requested):
The license file part of it looks good, thanks for taking the time to figure out those tests.

Just noticed this, then we're good to go.

eliykat (approved):

cturnbull-bitwarden (approved):

PR #3463, opened 2023-11-21
[AC-1809] Make LimitCollectionCreationDeletion in Current Context

vincentsalucci:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Code changes
* **src/Core/Context/CurrentContext.cs**: Injected `ApplicationCacheService` and retrieved/applied `LimitCollectionCreationDeletion` from the cached orgs.
* **src/Core/Context/ICurrentContext.cs**: Updated signature method in interface
* **src/Core/Models/Data/Organizations/OrganizationAbility.cs**: Added `LimitCollectionCreationDeletion` property

## Before you submit

### DIFF
--- a/src/Core/Context/CurrentContext.cs
+++ b/src/Core/Context/CurrentContext.cs
@@ -7,7 +7,9 @@ using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Identity;
 using Bit.Core.Models.Data;
+using Bit.Core.Models.Data.Organizations;
 using Bit.Core.Repositories;
+using Bit.Core.Services;
 using Bit.Core.Settings;
 using Bit.Core.Utilities;
 using Microsoft.AspNetCore.Http;
@@ -22,6 +24,7 @@ public class CurrentContext : ICurrentContext
     private bool _builtClaimsPrincipal;
     private IEnumerable<ProviderOrganizationProviderDetails> _providerOrganizationProviderDetails;
     private IEnumerable<ProviderUserOrganizationDetails> _providerUserOrganizations;
+    private readonly IApplicationCacheService _applicationCacheService;
 
     public virtual HttpContext HttpContext { get; set; }
     public virtual Guid? UserId { get; set; }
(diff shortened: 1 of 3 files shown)


src/Core/Context/CurrentContext.cs
shane-melton, line 1: though/question: `cachedOrgs` is a `IDictionary<>` of every organization in the database and this is doing a lot of `.containsKey()` and `cachedOrgs[Guid]` accesses. I could potentially see a performance issue if this is being called repeatedly for every current context construction.

If it is a problem, maybe we can grab all of the Orgs ids from the claims and do the `cachedOrgs` lookup once per org? Or maybe its not a problem at all?

src/Core/Context/CurrentContext.cs
```diff
@@ -184,6 +190,7 @@ private List<CurrentContextOrganization> GetOrganizations(Dictionary<string, IEn
                     Id = new Guid(c.Value),
                     Type = OrganizationUserType.Owner,
                     AccessSecretsManager = accessSecretsManager.ContainsKey(c.Value),
+                    LimitCollectionCreationDeletion = !cachedOrgs.ContainsKey(new Guid(c.Value)) || cachedOrgs[new Guid(c.Value)].LimitCollectionCreationDeletion,
```
r-tome, line 193: This might be a little more efficient because we're only accessing the dictionary once
```suggestion
                    LimitCollectionCreationDeletion = cachedOrgs[new Guid(c.Value)]?.LimitCollectionCreationDeletion ?? false,
```

eliykat (commented):
As discussed - I think that we can probably just call `applicationCacheService` from the collection handler without having to incorporate it into `CurrentContext`.

I did say that you could wait until further PRs are merged (as I suspect all related logic will end up moving into the handler), but to avoid delaying this fix until then, the simplest way forward right now is to just swap any reference to `CurrentContextOrganization.LimitCollectionCreationDeletion` with a call to `applicationCacheService` + checking the result. You can even do this in CurrentContext itself; the point is that you call it when you need it rather than constructing it in every request (and adding more responsibilities to CurrentContext).

vincentsalucci:
Thanks again everybody for the conversation. I'm going to close this PR and start fresh to re-align with our selected solution!

PR #3473, opened 2023-11-22
[AC-1827] Revert Change that Granted Premium to Families 2019

cyprain-okeke:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Revert Change that Granted Premium to Families 2019

## Code changes

## Before you submit


util/Migrator/DbScripts/2023-11-22_00_Revert2019FamilyPlanFeatureUpgrade.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN TRY
```
rkac-bw, line 1: I noticed we're using explicit transactions in this script. Since we've got IMPLICIT_TRANSACTIONS on, SQL Server's already handling transactions for us at the start of data-changing statements like INSERT, UPDATE, etc. So, our BEGIN TRANSACTION, COMMIT, and ROLLBACK might be doing a double job here for a single update.

If we plan to add more updates or changes that need to happen all in one go, then these explicit transactions are necessary. For example, if there were two UPDATEs that should either both succeed or fail together to keep our data from going haywire, we'd totally need this explicit wrap.

cyprain-okeke: resolved

rkac-bw (approved):
Looks good to me

eliykat (approved):

alsikora (approved):

PR #3474, opened 2023-11-23
[AC-1839] Add OrganizationLicense unit tests

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Add unit tests to protect against breaking existing installs when updating the organization license class.

I've added coverage for the current license version and the previous license version as examples.

## Code changes

`OrganizationLicenseStaticVersions` contains static json examples of the current and previous license file versions. The unit tests load up these stored values into the current `OrganizationLicense` class (which you may have modified) and checks that (1) the hash has not changed and (2) it still validates against the organization.

This is basically trying to replicate the experience of a new self-hosted instance loading up an existing license file after upgrading.

When you increment the license file version, you must generate a new json string to store in `OrganizationLicenseStaticVersions`. There's a helper to do this which is a bit awkward (running a fake test and copying the value out of the debugger) but also fairly simple.

We should make more fundamental improvements to this class, but this puts up some guard rails in the meantime.

## Before you submit


cturnbull-bitwarden (approved):

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -145,7 +145,7 @@ public OrganizationLicense()
     /// </summary>
     /// <remarks>Intentionally set one version behind to allow self hosted users some time to update before
     /// getting out of date license errors</remarks>
-    private const int CURRENT_LICENSE_FILE_VERSION = 12;
+    public const int CURRENT_LICENSE_FILE_VERSION = 12;
```
withinfocus, line 148: ⛏️ This should follow coding standards if it's now used elsewhere so I'd assume it'll rename to `CurrentLicenseFileVersion`.

eliykat:
Feedback has been addressed, I also tweaked naming and comments a little to arrange the code better. e.g.
* `OrganizationLicenseFileFixtures` is a more descriptive name instead of `StaticVersions`
* Moved the `CreateOrganization` method to the fixtures class so it's with the rest of the test data, renamed to `OrganizationFactory`
* slightly more helpful xmldoc comments to help devs who come across this

cturnbull-bitwarden (approved):

PR #3492, opened 2023-11-30
[AC-1871] Register IFeatureService in Events project

eliykat:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

#3372 started using IFeatureService in Events, but it wasn't registered. This PR registers it as a singleton, as it is in the other projects.

I double checked the other projects while I was here, everyone else either uses SharedWeb which registers this service, or they don't use the service at all.

## Before you submit

### DIFF
--- a/src/Events/Startup.cs
+++ b/src/Events/Startup.cs
@@ -70,7 +70,6 @@ public class Startup
             services.AddSingleton<IEventWriteService, RepositoryEventWriteService>();
         }
 
-        // To be removed in AC-1723 provided feature flags are not otherwise being used in this project
         services.AddSingleton<IFeatureService, LaunchDarklyFeatureService>();
 
         // Mvc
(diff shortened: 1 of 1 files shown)


src/Events/Startup.cs
```diff
@@ -70,6 +70,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<IEventWriteService, RepositoryEventWriteService>();
         }
 
+        // To be removed in AC-1723 provided feature flags are not otherwise being used in this project
```
withinfocus, line 73: ❓ Is this necessary? Doesn't hurt to load the service in my opinion, for future use.

djsmith85 (dismissed):

src/Events/Startup.cs
eliykat: I wasn't sure how aggressive to be about that, but I agree. Comment removed.

withinfocus (approved):

withinfocus:
We have to require builds and testing!

PR #3493, opened 2023-11-30
[Flexible collections] Experiment: add CreateNewCollections to JWT claims

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Collection management settings affect what roles can do:
* LimitCollectionCreationDeletion affects whether everyone can create a new collection, or only users with admin permissions
* LimitCollectionCreationDeletion also affects whether a user with Can Manage permissions for a collection can delete that collection (or only users with admin permissions can)
* AllowAdminAccessToAllCollectionItems affects whether admins and owners can view items in unassigned collections

Therefore, it is no longer enough to just know the role, we also need to know the value of these settings to know what a given role can do.

We have 3 options:
1. Include the setting directly in the claims - rejected in #3433 
2. Calculate the more granular permission and include it in the claims (this PR) - suggested by @withinfocus
3. Leave it out of the claims completely and expand the OrgAbilities cache to include collection management settings (to still avoid a database read)

This PR tries out no. 2, adding granular permissions to the claims. I am not endorsing this approach necessarily.

Advantages:
* more authorization logic can make decisions based on a "dumb" data object (`CurrentContextOrganization`) rather than having to fetch data from multiple sources
* avoids async calls to database or cache service
* centralises this `CanCreateCollections` logic in 1 place, otherwise it's repeated quite a few times across authorization handlers after the changes in #3360. Of course we could always centralise the logic elsewhere, but it's a nice side-effect.
* probably the easiest, based on the changeset in this PR.

Disadvantages:
* Microsoft suggests that claims should be based on identity (e.g. "this user is an owner"), not ability (e.g. "this user can create collections"). This violates that rule. ([docs](https://learn.microsoft.com/en-us/aspnet/core/security/authorization/claims?view=aspnetcore-8.0))
* adding more weight to the JWT which is sent with every request to the server. What's our limit here - is there a risk we keep adding to the JWT instead of investing in other caching solutions?
* inconsistent with how we handle other permissions, which are calculated in the code based on role, not calculated upfront and stored in the JWT
* can be slow to update after db value changes (need to wait until a token refresh)

If we don't go with 2, then we go with 3, adding the collection management settings to the org abilities cache. A collection management setting is not really an org "ability", so it's expanding the purposes of that cache, but that might be OK given that the cache is storing commonly accessed org properties. That would leave it to the authorization code to decide what to do based on the value of `orgAbility.LimitCollectionCreationDeletion`.

## Code changes

## Before you submit

### DIFF
--- a/src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
+++ b/src/Api/Vault/AuthorizationHandlers/Collections/CollectionAuthorizationHandler.cs
@@ -83,10 +83,7 @@ public class CollectionAuthorizationHandler : BulkAuthorizationHandler<Collectio
     {
         // If the limit collection management setting is disabled, allow any user to create collections
         // Otherwise, Owners, Admins, and users with CreateNewCollections permission can always create collections
-        if (org is
-        { LimitCollectionCreationDeletion: false } or
-        { Type: OrganizationUserType.Owner or OrganizationUserType.Admin } or
-        { Permissions.CreateNewCollections: true })
+        if (org is {CreateNewCollections: true})
         {
             context.Succeed(requirement);
             return;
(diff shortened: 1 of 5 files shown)


eliykat:
After further internal discussions, we decided to continue with the OrgAbilities route (option 3).

PR #3504, opened 2023-12-01
Make development easier

justindbaur:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The objective is to remove a few steps from the setup process to get `server` up and running. We [currently instruct](https://contributing.bitwarden.com/getting-started/server/guide#generate-certificates) ([permalink](https://github.com/bitwarden/contributing-docs/blob/3c0043fecff96eaf0a0b46a23482e45ccde05234/docs/getting-started/server/guide.md#generate-certificates)) developers to run a script to generate them 2 certificates. One for data protection and one for identity server. 

The certificate for data protection isn't used today and has been unnecessary for some time (forever?). [Here](https://github.com/bitwarden/server/blob/6e4a057d55b3af878e4fcb38532beab389e737eb/src/SharedWeb/Utilities/ServiceCollectionExtensions.cs#L475-L479) you can see the method return early if the environment is `Development`. The `GlobalSettings.DataProtection.CertificateThumbprint` that we instruct users to set, never gets used. 

The certificate for identity server is used to sign json web tokens from the server. It being persisted across restarts has value and is what a lot of developers here have come to expect so I added a new development setting to both `Identity` and `Sso` so that they know where to look for shared storage. Now, on the first run Identity Server will generate a key and persist it to disk. There is a slight chance for a race condition where `Identity` and `Sso` startup at nearly the same time and get to the code to persist their key to disk at the same time but that didn't happen for me locally and most of the time developers will run `Min Server` at first which is just `Identity` and `Api` so `Identity` will persist the key and then you will never have an issue again. There is one other use of the `IdentityServerCertificate` which is for SAML SSO. It's an [optional](https://github.com/bitwarden/server/blob/6e4a057d55b3af878e4fcb38532beab389e737eb/bitwarden_license/src/Sso/Utilities/DynamicAuthenticationSchemeProvider.cs#L375-L378) addition and our local IDP we recommend for SAML works just fine without. Our other development SSO option, Okta, does not use it at all. For the reason that some people may want to use it sometimes, I have left the parts of the scripts where we help the creation of those certificates around but I want to change the docs to only mention running that script to the SSO section of the docs vs being in the getting started guide. 

## Code changes

* **.devcontainer/{community_dev|internal_dev}/postCreateCommand.sh:** Fully remove all the code dealing with certificates. 
* **appsettings.Development.json:** Add relative paths to our dev folder so that a shared key can be stored there.
* **create_certificates_*:** Removed code that generated data protection certificates, but left in the code to generate Identity Server certificates.
* **src/SharedWeb/Utilities/ServiceCollectionExtensions.cs:** Added code for preferring a persisted development key over a fully ephemeral, in memory one. Bitwarden developers who have setup their identity server certificate & thumbprint already will not hit this portion of the code. 
* **test/IntegrationTestCommon/Factories/WebApplicationFactoryBase.cs:** Forced integration tests to use in memory key
* **.devcontainer/internal_dev/devcontainer.json:** Open up the mail catcher UI port outside of the dev container. 
* **.vscode/tasks.json:** Added a couple helpers to VS Code tasks that I found useful when setting up my dev container.

## Before you submit


src/Core/Settings/GlobalSettings.cs
```diff
@@ -80,6 +80,7 @@ public virtual string LicenseDirectory
     public virtual IPasswordlessAuthSettings PasswordlessAuth { get; set; } = new PasswordlessAuthSettings();
     public virtual IDomainVerificationSettings DomainVerification { get; set; } = new DomainVerificationSettings();
     public virtual ILaunchDarklySettings LaunchDarkly { get; set; } = new LaunchDarklySettings();
+    public virtual string DevelopmentDirectory { get; set; }
```
withinfocus, line 83: 🎨 Can you support this in the interface too?

withinfocus:
What are some suggested migration / cleanup steps for engineers in the current state?

justindbaur:
@withinfocus 

> What are some suggested migration / cleanup steps for engineers in the current state?

There are no steps _required_, all current setups should work just fine, their data protection cert was never used but also never hurt and their identity server certificate will still take precedence over the signingkey.jwk file. It will continue to get a non-null cert [from here](https://github.com/bitwarden/server/blob/6e4a057d55b3af878e4fcb38532beab389e737eb/src/SharedWeb/Utilities/ServiceCollectionExtensions.cs#L509C51-L513) and then use that for identity server. If they want to use the other one they can just delete `globalSettings > identityServer > certificateThumbprint`, restart server, and accept that all existing sessions will be expired.

withinfocus (approved):
Great quality of life improvement!

.vscode/tasks.json
```diff
@@ -211,6 +211,42 @@
...
+            "default": "true",
+            "description": "Whether or not to clear existing secrets",
+            "options": [
+                {
+                    "label": "true",
+                    "value": "true"
+                },
+                {
+                    "label": "false",
+                    "value": "false"
```
jlf0dev, line 241: I know this is probably how it works but serious
```
if (label == true) {
    return true;
} else {
    return false;
}
```
vibes here 😂

jlf0dev (approved):
Love this!

PR #3506, opened 2023-12-02
[PM-4168] update keys for WebAuthnLoginCredential

ike-kottlowski:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

[Graphite Link](https://app.graphite.dev/github/pr/bitwarden/server/3506/PM-4168-update-keys-for-WebAuthnLoginCredential)

[corresponding client PR](https://github.com/bitwarden/clients/pull/7074)

## Objective

Adding necessary methods and requests to enable passkey assertion and updating passkey `KeySets`.

### New Files
* **WebAuthnLoginCredentialUpdateRequestModel.cs :** Model used by the `[PUT] /webauthn` endpoint.

### Updated Files
* **WebAuthnController.cs :** 
  * Added Assertion method for user to fetch assertion options and create a token to ensure the correct flow is followed.
  * Updated `/Options` endpoint to `/Attestation-options` to be clearer. 
  * Added `[PUT] /webauthn/`. It accepts `WebAuthnLoginCredentialUpdateRequestModel`
* **WebAuthnLoginCredentialCreatelRequestModel.cs :** rename.
* **WebAuthnLoginAssertionOptionsScope.cs :** Added entry `UpdateKeySet`
* **IWebAuthnCredentialRepository.cs :** Added method for updating `WebAuthnCredential`.
* **WebAuthnCredentialRepository.cs :** Implemented new update method.
* **WebAuthnControllerTests.cs :** 
  * Changed some tests to be more specific by including `Attestation` instead of `Post`
  * Updated `Assert` calls in a `POST` method to verify mock methods were called.
  * Added various tests to account for the new `[PUT] /webauthn` endpoint. 

## Before you submit


src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -64,8 +69,24 @@ public async Task<WebAuthnCredentialCreateOptionsResponseModel> PostOptions([Fro
         };
     }
 
+    [HttpPost("assertion-options")]
+    public async Task<WebAuthnLoginAssertionOptionsResponseModel> AssertionOptions([FromBody] SecretVerificationRequestModel model)
+    {
+        var user = await VerifyUserAsync(model);
+        var options = _userService.StartWebAuthnLoginAssertion();
+
+        var tokenable = new WebAuthnLoginAssertionOptionsTokenable(WebAuthnLoginAssertionOptionsScope.PrfRegistration, options);
```
jlf0dev, line 78: Probably shouldn't just assume we're doing Prf Registration here

jlf0dev (commented):
So far so good 👍

src/Api/Auth/Controllers/WebAuthnController.cs
ike-kottlowski: No, I think you're right. There should be a difference between the states of the Token. `login != registration != update`. We can use the token to control for which state the request is in. This way we can ensure "single responsibility" for tokens.

src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -93,6 +114,29 @@ private async Task ValidateRequireSsoPolicyDisabledOrNotApplicable(Guid userId)
         }
     }
 
+    [HttpPut()]
+    public async Task UpdateCredential([FromBody] WebAuthnLoginCredentialUpdateRequestModel model)
+    {
+        var tokenable = _assertionOptionsDataProtector.Unprotect(model.Token);
+        if (!tokenable.TokenIsValid(WebAuthnLoginAssertionOptionsScope.PrfRegistration))
+        {
+            throw new BadRequestException("The token associated with your request is expired. A valid token is required to continue.");
```
coroiu, line 123: nit(non-blocking): Technically they could've sent a token with another scope

```suggestion
            throw new BadRequestException("The token associated with your request is invalid. A valid token is required to continue.");
```

```suggestion
            throw new BadRequestException("The token associated with your request is invalid or has expired. A valid token is required to continue.");
```

src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -93,6 +114,29 @@ private async Task ValidateRequireSsoPolicyDisabledOrNotApplicable(Guid userId)
...
+        }
+
+        var (_, credential) = await _userService.CompleteWebAuthLoginAssertionAsync(tokenable.Options, model.DeviceResponse);
+        if (credential == null)
+        {
+            throw new BadRequestException("Unable to update WebAuthnLoginCredential.");
```
coroiu, line 129: nit: maybe make this a bit more user friendly since this will probably be shown to the user

```suggestion
            throw new BadRequestException("Unable to update credential.");
```

src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -93,6 +114,29 @@ private async Task ValidateRequireSsoPolicyDisabledOrNotApplicable(Guid userId)
...
+        }
+
+        // assign new keys to credential
+        credential.EncryptedUserKey = model.EncryptedUserKey;
+        credential.EncryptedPrivateKey = model.EncryptedPrivateKey;
+        credential.EncryptedPublicKey = model.EncryptedPublicKey;
```
coroiu, line 135: suggestion: verify this somehow? Maybe using `credential.PrfStatus === Enabled` to make sure that all keys are set and valid

src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -93,6 +114,29 @@ private async Task ValidateRequireSsoPolicyDisabledOrNotApplicable(Guid userId)
         }
     }
 
+    [HttpPut()]
+    public async Task UpdateCredential([FromBody] WebAuthnLoginCredentialUpdateRequestModel model)
+    {
+        var tokenable = _assertionOptionsDataProtector.Unprotect(model.Token);
+        if (!tokenable.TokenIsValid(WebAuthnLoginAssertionOptionsScope.PrfRegistration))
```
coroiu, line 121: suggestion: add a test case for this

src/Api/Auth/Models/Request/WebAuthn/WebAuthnLoginCredentialCreatelRequestModel.cs
```diff
@@ -29,5 +29,8 @@ public class WebAuthnCredentialRequestModel
     [EncryptedString]
     [EncryptedStringLength(2000)]
     public string EncryptedPrivateKey { get; set; }
+
+    public (string userKey, string publicKey, string privateKey) GetKeys()
+        => (EncryptedUserKey, EncryptedPublicKey, EncryptedPrivateKey);
```
coroiu, line 34: question: I like, but it's not used? WIP maybe? 😄

src/Api/Auth/Controllers/WebAuthnController.cs
coroiu: Ah right, I just saw that your tests are WIP, nevermind 😄

coroiu (commented):
Same as client PR, looking good, nothing unexpected!

src/Api/Auth/Models/Request/WebAuthn/WebAuthnLoginCredentialCreatelRequestModel.cs
ike-kottlowski: I think it was vestigial from a previous implementation I was working on. 🤔

src/Api/Auth/Controllers/WebAuthnController.cs
ike-kottlowski: I added a check to ensure that `credential.SupportsPRf == true`, that what you mean by _valid_?

coroiu: I really meant checking `credential.PrfStatus == Enabled`, but I definitely wasn't clear on why.

The reason is that for PRF to be enabled you need all of those keys, so it doesn't really make sense to allow someone to call this endpoint without providing all of the keys. I realize now however that you might solve this issue by enforcing the fields in the api model 🤔 

Also adding `SupportsPrf === true` was actually a good idea, hadn't even crossed my mind but it doesn't really makes sense to set PRF keys for a credential that doesn't support PRF from the beginning!

coroiu: I really meant checking `credential.PrfStatus == Enabled`, but I definitely wasn't clear on why.

The reason is that for PRF to be enabled you need all of those keys, so it doesn't really make sense to allow someone to call this endpoint without providing all of the keys. I realize now however that you might solve this issue by enforcing the fields in the api model 🤔 

Also adding `SupportsPrf === true` was actually a good idea, hadn't even crossed my mind but it doesn't really makes sense to set PRF keys for a credential that doesn't support PRF from the beginning!

coroiu (commented):
looks good, added a response to an older comment

src/Infrastructure.Dapper/Auth/Repositories/WebAuthnCredentialRepository.cs
```diff
@@ -44,4 +44,15 @@ public async Task<ICollection<WebAuthnCredential>> GetManyByUserIdAsync(Guid use
             return results.ToList();
         }
     }
+
+    public async Task<bool> UpdateAsync(WebAuthnCredential credential)
+    {
+        using var connection = new SqlConnection(ConnectionString);
+        var affectedRows = await connection.ExecuteAsync(
```
jlf0dev, line 51: nit: variable isn't used

jlf0dev (approved):
Added some nits but overall I think this is good!

src/Infrastructure.Dapper/Auth/Repositories/WebAuthnCredentialRepository.cs
ike-kottlowski: Yep vestigial from when I was returning a `bool`

src/Api/Auth/Controllers/WebAuthnController.cs
ike-kottlowski: I added `[Required]` attribute to the request model.

jlf0dev (approved):

PR #3513, opened 2023-12-05
[AC-1753] Automatically assign provider's pricing to new organizations

cyprain-okeke:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

For MSP Providers that partnered with Bitwarden prior to Nov 6, 2023, they should retain 2020 pricing for all existing and new organizations. When a 2020 MSP creates a new organizations or adds an existing organization, the 2020 price should be automatically assigned and shown in the UI.

## Code changes

## Before you submit


cturnbull-bitwarden (dismissed):
Looks great, thank you sir

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -393,6 +397,94 @@ public async Task AddOrganizationsToReseller(Guid providerId, IEnumerable<Guid>
...
+        {
+            if (mapping.Value.IndexOf(organization.Plan, StringComparison.Ordinal) != -1)
+            {
+                organization.PlanType = mapping.Key;
+                organization.Plan = mapping.Value;
+                return organization.PlanType;
+            }
+        }
+
+        throw new BadRequestException("Invalid PlanType selected");
```
addisonbeck, line 476: This should throw an `ArgumentException` instead of a `BadRequestException` since at this stage we're dealing with our own pre-existing data and not something from a user request.

https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception?view=net-8.0

src/Admin/Views/Shared/_OrganizationForm.cshtml
```diff
@@ -83,14 +83,24 @@
                     @{
                         var planTypes = Enum.GetValues<PlanType>()
                             .Where(p =>
-                                Model.Provider == null ||
+                                (Model.Provider == null && !p.ToString().Contains("2020")) ||
                                 (Model.Provider != null
-                                 && p is >= PlanType.TeamsMonthly2019 and <= PlanType.EnterpriseAnnually2019 or >= PlanType.TeamsMonthly2020 and <= PlanType.EnterpriseAnnually)
+                                 && (
+                                     (Model.Provider.CreationDate < new DateTime(2023, 11, 6) && p.ToString().Contains("2020"))
```
addisonbeck, line 89: This is another good bit of evidence to support that we need some kind of central log/configuration with dates around pricing. Hardcoding the value all around is going to hurt us later when we inevitably have to adjust this all again.

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -393,6 +397,94 @@ public async Task AddOrganizationsToReseller(Guid providerId, IEnumerable<Guid>
...
+            }
+        }
+        catch (Exception e)
+        {
+            throw new Exception("Unable to update existing plan on stripe");
+        }
+
+    }
+
+    private static PlanType GetPlanTypeFromPlan(string plan, Organization organization)
```
addisonbeck, line 456: I'm confused by what this method is for. It seems like it takes and organization plan string (the display name) and maps it back to the enum value - but the Organization object has a PlanType value on it already.

addisonbeck (changes requested):

addisonbeck:
I think this came up elsewhere, but some unit tests would be appreciated here, as well as maybe a screenshot of the UI changes to the admin portal.

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -393,6 +397,94 @@ public async Task AddOrganizationsToReseller(Guid providerId, IEnumerable<Guid>
...
+        if (provider.CreationDate >= new DateTime(2023, 11, 6))
+        {
+            return;
+        }
+
+        foreach (var orgId in organizationIds)
+        {
+            var organization = await _organizationRepository.GetByIdAsync(orgId);
+            var subscriptionItem = await GetSubscriptionItemAsync(organization.GatewaySubscriptionId, GetPlanId(organization.PlanType));
+            var newPlanType = GetPlanTypeFromPlan(organization.Plan, organization);
```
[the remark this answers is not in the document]
cyprain-okeke: we get  the org plan and get his 2020 plan equivalent. and set the org plan to 2020 plan

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -380,7 +384,7 @@ public async Task AddOrganizationsToReseller(Guid providerId, IEnumerable<Guid>
         {
             throw new BadRequestException("Provider must be of type Reseller in order to assign Organizations to it.");
         }
-
+        await UpdatePlanForExistingProviderAsync(organizationIds, provider);
```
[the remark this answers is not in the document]
addisonbeck: I think generalizing it makes sense to do. `UpdatePlanForExistingProviderAsync` doesn't just handle these grandfathered plans. Right now the way it reads is:

1. Every subscription setup event calls the method, no matter the special rules that may apply to them.
2. `if` there is no special rate applied `then` ✅, we're all good, return the subscription as a normal request. A price rate is still applied, it's just the default.
3. `else if` there *is* a special rate applied: apply it, and return. 

So, while only existing providers get updated, *all* providers go through the check. There is also a high likelihood this gets adjusted in the future based on any later rate changes that happen. 

I don't want to hold up any work based on opinions about naming, and I trust your judgement, so this isn't a blocking thing for me, but I'd still like to make the suggestion.

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -381,18 +386,108 @@ public async Task AddOrganizationsToReseller(Guid providerId, IEnumerable<Guid>
...
+            if (mapping.Value.IndexOf(organization.Plan, StringComparison.Ordinal) != -1)
+            {
+                organization.PlanType = mapping.Key;
+                organization.Plan = mapping.Value;
+                return organization.PlanType;
+            }
+        }
+
+        throw new ArgumentException("Invalid PlanType selected");
+    }
```
addisonbeck, line 480: I'm still confused by what this method is doing. It takes in an organization as input, runs through some logic to check if that plan is a 2020 plan, and then returns the same plan that was passed in? It also sets some values on the `organization` object, but it looks like those values are the same ones that were originally passed in, and then nothing happens with the `organization` after that anyway since the `mapping.Key` gets returned right after the values are set. What's this method's purpose?

addisonbeck (dismissed):

cturnbull-bitwarden (approved):

cturnbull-bitwarden (dismissed):

addisonbeck (approved):

addisonbeck (approved):

cturnbull-bitwarden (approved):

PR #3514, opened 2023-12-05
[AC-1889] Fix ManageGroups custom permission not getting all collections

eliykat:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

A custom user with the Manage Groups permission cannot see collections assigned to groups (In Admin Console -> Groups) if the user is not otherwise assigned to those collections.

This was initially marked as a new regression, but on further attempts to reproduce, it does appear to already be present in prod. And from my perspective, I couldn't see any recent code changes that would've caused it.

The problem is that `collectionService.GetOrganizationCollections` does not return all collections for `ManageGroups` users, only those they are assigned to.

## Code changes

If the user can manage groups, return all collections from the `organizations/{id}/collections` endpoint.

Per the comment, I didn't push this down to the service call because that is used in various places and could unintentionally open up other actions to these users. This is being replaced by new logic in #3360 anyway, so a targeted (if ugly) fix seemed the best solution for now.

## Before you submit


shane-melton (dismissed):

Jingo88 (approved):

src/Api/Controllers/CollectionsController.cs
```diff
@@ -135,7 +135,22 @@ public async Task<ListResponseModel<CollectionAccessDetailsResponseModel>> GetMa
...
     public async Task<ListResponseModel<CollectionResponseModel>> Get(Guid orgId)
     {
-        IEnumerable<Collection> orgCollections = await _collectionService.GetOrganizationCollectionsAsync(orgId);
+        IEnumerable<Collection> orgCollections = null;
+        if (await _currentContext.ManageGroups(orgId))
+        {
+            // ManageGroups users need to see all collections to manage other users' collection access.
+            // This is not added to collectionService.GetOrganizationCollectionsAsync as that may have
+            // unintended consequences on other logic that also uses that method.
+            // This is a quick fix but it will be properly fixed by permission changes in Flexible Collections.
```
vincentsalucci, line 144: If this is going to be fixed down the road, should we have an explicit task in our `Post Release` epic to properly track/update/remove this code? Or do we think an existing task encompasses the necessary clean-up?

eliykat: #3360 introduces a v2 version of this endpoint, so this code will be deleted when we remove the feature flags and clean up all the legacy code.

vincentsalucci (approved):

PR #3516, opened 2023-12-05
[DEVOPS-1654] Tag Server images in master with git commit

michalchecinski:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Tag Server images in master with git commit.

## Code changes

* **.github/workflows/build.yml:** Add another docker image name when on master.

## Before you submit

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -308,10 +308,10 @@ jobs:
           SHA: ${{ github.sha }}
         run: |
           NAMES="${_AZ_REGISTRY}/${PROJECT_NAME}:${IMAGE_TAG}"
-          # if [[ "${IMAGE_TAG}" == "dev" ]]; then
-          SHORT_SHA=$(git rev-parse --short ${SHA})
-          NAMES=$NAMES",${_AZ_REGISTRY}/${PROJECT_NAME}:dev-${SHORT_SHA}"
-          # fi
+          if [[ "${IMAGE_TAG}" == "dev" ]]; then
+            SHORT_SHA=$(git rev-parse --short ${SHA})
+            NAMES=$NAMES",${_AZ_REGISTRY}/${PROJECT_NAME}:dev-${SHORT_SHA}"
+          fi
           echo "names=$NAMES" >> $GITHUB_OUTPUT
 
       - name: Get build artifact
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -308,10 +308,10 @@ jobs:
           SHA: ${{ github.sha }}
         run: |
           NAMES="${_AZ_REGISTRY}/${PROJECT_NAME}:${IMAGE_TAG}"
-          if [[ "${IMAGE_TAG}" == "dev" ]]; then
-            SHORT_SHA=$(echo "${SHA}" | cut -c1-7)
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -305,7 +305,13 @@ jobs:
...
           IMAGE_TAG: ${{ steps.tag.outputs.image_tag }}
           PROJECT_NAME: ${{ steps.setup.outputs.project_name }}
-        run: echo "name=${_AZ_REGISTRY}/${PROJECT_NAME}:${IMAGE_TAG}" >> $GITHUB_OUTPUT
+          SHA: ${{ github.sha }}
+        run: |
+          NAME="${_AZ_REGISTRY}/${PROJECT_NAME}:${IMAGE_TAG}"
+          if [[ "${IMAGE_TAG}" == "dev" ]]; then
+            NAME=$NAME",${_AZ_REGISTRY}/${PROJECT_NAME}:${SHA}"
+          fi
+          echo "name=$NAME" >> $GITHUB_OUTPUT
```
vgrassia, line 314: If we're going to add multiple tags I think we should update the name of the steps related to be more accurate. Instead of `Generate image full name` we should have something like `Generate tag(s)`.  Also, do we want the full `SHA` by itself as a tag, or something like `dev-a1b2c3`?

vgrassia (changes requested):

.github/workflows/build.yml
michalchecinski: I think `dev` with short commit would be more readable, but from the card I assumed it should be a full commit hash.

vgrassia: Let's use the full hash, but still include the tag name so it's easily identifiable. `${_AZ_REGISTRY}/${PROJECT_NAME}:dev-${SHA}`

mimartin12: I didn't specify the commit has in the planned work, because I wanted it to be up for discussion. @michalchecinski makes a good point, a short commit would be easier to read, however, GitHub provides the `GITHUB_SHA` as a full commit. There will need to be assumptions and logic to shorten that when looking/pulling for an image as well.

.github/workflows/build.yml
```diff
@@ -300,12 +300,19 @@ jobs:
...
+        id: image-names
         env:
           IMAGE_TAG: ${{ steps.tag.outputs.image_tag }}
           PROJECT_NAME: ${{ steps.setup.outputs.project_name }}
-        run: echo "name=${_AZ_REGISTRY}/${PROJECT_NAME}:${IMAGE_TAG}" >> $GITHUB_OUTPUT
+          SHA: ${{ github.sha }}
+        run: |
+          NAMES="${_AZ_REGISTRY}/${PROJECT_NAME}:${IMAGE_TAG}"
+          # if [[ "${IMAGE_TAG}" == "dev" ]]; then
+          SHORT_SHA=$(echo "${SHA}" | cut -c1-7)
```
mimartin12, line 312: Something like `git rev-parse --short` here would also work I think, and not be subject to possible changes in arguments via the `cut` command.

.github/workflows/build.yml
michalchecinski: I made it short with `dev` prefix. Let me know if that's okay.

mimartin12 (approved):

vgrassia (approved):

PR #3530, opened 2023-12-08
[AC-1389] [AC-1919] check AllowAdminAccessToAllCollectionItems when saving Collection

willmartian:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Don't throw a `BadRequestException` when saving a Collection if `AllowAdminAccessToAllCollectionItems` is `true`.

## Code changes

* **src/Core/Services/Implementations/CollectionService.cs:** Don't throw a `BadRequestException` when saving a Collection if `AllowAdminAccessToAllCollectionItems` is `true`
* **test/Core.Test/Services/CollectionServiceTests.cs:** Set `AllowAdminAccessToAllCollectionItems` to `false` in test

## Before you submit


src/Core/Services/Implementations/CollectionService.cs
```diff
@@ -60,7 +60,7 @@ public class CollectionService : ICollectionService
         {
             var groupHasManageAccess = groupsList?.Any(g => g.Manage) ?? false;
             var userHasManageAccess = usersList?.Any(u => u.Manage) ?? false;
-            if (!groupHasManageAccess && !userHasManageAccess)
+            if (!groupHasManageAccess && !userHasManageAccess && !org.AllowAdminAccessToAllCollectionItems)
             {
                 throw new BadRequestException(
                     "At least one member or group must have can manage permission.");
```
eliykat, line 66: You also need to update the feature flag check on line 59 to use the v1 flag instead of the MVP flag. (to match the client-side change)

eliykat (changes requested):

src/Core/Services/Implementations/CollectionService.cs
eliykat: We discussed this in the sync, but just closing the loop - we decided it would be safest to leave it in to cover a rollback scenario, where the flag is turned off again after db values have been modified. It also makes life easier for QA who are messing around with their data. (I like that you're thinking critically about flags though, aside from those edge cases I would agree.)

src/Core/Services/Implementations/CollectionService.cs
```diff
@@ -56,11 +56,11 @@ public class CollectionService : ICollectionService
         var usersList = users?.ToList();
 
         // If using Flexible Collections - a collection should always have someone with Can Manage permissions
-        if (_featureService.IsEnabled(FeatureFlagKeys.FlexibleCollections, _currentContext))
+        if (_featureService.IsEnabled(FeatureFlagKeys.FlexibleCollectionsV1, _currentContext))
         {
             var groupHasManageAccess = groupsList?.Any(g => g.Manage) ?? false;
             var userHasManageAccess = usersList?.Any(u => u.Manage) ?? false;
-            if (!groupHasManageAccess && !userHasManageAccess)
+            if (!groupHasManageAccess && !userHasManageAccess && !org.AllowAdminAccessToAllCollectionItems)
```
eliykat, line 63: Just clarified with Priya and this error does _not_ depend on `org.AllowAdminAccessToAllCollectionItems`. But at least you've updated the feature flag reference while you're here 😁

willmartian: Understood! Reverted that. Should the other reference to the `FeatureFlagKeys.FlexibleCollections` [here ](https://github.com/bitwarden/server/pull/3530/files#diff-f4e0ad94f5988defbed21fff3fc24821c161ae60bb5ed330ce80ebd7370ce931R46-R47)also be updated to V1?

eliykat: Nope, that reference is used [here](https://github.com/bitwarden/server/blob/5b1a4150388d1d707a0dde0f006328ecdb1b93d4/src/Core/Services/Implementations/CollectionService.cs#L128-L129) which is MVP behaviour, so that's fine.

eliykat (commented):
See threaded comment above

eliykat (changes requested):
As discussed, this error depends on both the v1 flag and on `!org.AllowAdminAccessToAllCollectionItems`. This is not what I said previously, sorry for the confusion.

This should be fixed by reverting this commit: https://github.com/bitwarden/server/pull/3530/commits/302984dd7e6dd72a553d26892807589b87bc384f

Please also see previous feedback left above.

willmartian:
> sorry for the confusion

No worries at all! @eliykat

eliykat (approved):
Fixed tests for you, otherwise lgtm!

PR #3571, opened 2023-12-13
[AC-1809] Update OrganizationAbility with Collection Management Settings

vincentsalucci:
## Type of change

- [X] Bug fix
- [X] New feature development
- [ ] Other

## Objective

> In order to deliver collection management settings to the authorization handlers, an update needs to be made to the `OrganizationAbility` pipeline to include the above. In order to prevent extra database calls during API usage and `CurrentContext` creation. Helper methods were added to each handler to necessitate the transfer of information. 

## Code changes

**.../AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs**
**.../AuthorizationHandlers/Groups/GroupAuthorizationHandler.cs**
**.../AuthorizationHandlers/OrganizationUsers/OrganizationUserAuthorizationHandler.cs**

* Injected `ApplicationCacheService`. 
* Retrieved `OrganizationAbility` on demand with helper method. 
* Restructured permissions conditionals to handle separation of collection management settings from `CurrentContextOrganization` claims generated permission object.

**src/Core/AdminConsole/Context/CurrentContextOrganization.cs**

* Removed properties of collection management settings in order to reduce confusion about where those values should be retrieved from

**src/Core/AdminConsole/Models/Data/Organizations/OrganizationAbility.cs**

* Added properties for collection management settings

**src/Core/Context/CurrentContext.cs**

* Injected `ApplicationCacheService`
* Updated `ViewAssignedCollections` method with cache to retrieve `LimitCollecitonCreationDeletion`. This method is temporary and will be removed with V1 Flexible Collections work. This is merely a placeholder to maintain backwards compatibility.

**src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs**

* Updated Entity Framework representation of `OrganizationAbility` list retrieval by adding the collection management settings

**src/Sql/dbo/Stored Procedures/Organization_ReadAbilities.sql**

* Updated sproc by adding the collection management settings

**util/Migrator/DbScripts/2023-12-01_00_OrganizationAbilitiesWithCollectionManagementSettings.sql**

* Created migration script

**test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs**
**test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs**
**test/Api.Test/Vault/AuthorizationHandlers/OrganizationUserAuthorizationHandlerTests.cs**

* Updated all tests to reflect changes made to the storage/retrieval of collection management settings

## Before you submit


src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -247,19 +257,32 @@ public class BulkCollectionAuthorizationHandler : BulkAuthorizationHandler<BulkC
...
+        if (_organizationAbility != null)
+        {
+            return _organizationAbility;
+        }
+
+        (await _applicationCacheService.GetOrganizationAbilitiesAsync())
+            .TryGetValue(_targetOrganizationId, out var organizationAbility);
+
+        return _organizationAbility = organizationAbility;
+    }
```
eliykat, line 287: I'm not sure about the `_organizationAbility` private property here. Presumably it avoids repeated calls to the `applicationCacheService`, but that service is already a cache - it's OK to call it multiple times. (It's just an in-memory dict if you look inside.)

There's also the chance that we check different orgs in the same request (not today afaik but also not impossible). In that case, you'd have the wrong `_organizationAbility` stored here. (That's solvable by comparing orgIds, but as I've said, I don't think it's necessary.)

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -247,19 +257,32 @@ public class BulkCollectionAuthorizationHandler : BulkAuthorizationHandler<BulkC
 
     private async Task<bool> IsAssignedToCollectionsAsync(
         ICollection<Collection> targetCollections,
-        CurrentContextOrganization org,
+        CurrentContextOrganization? org,
```
eliykat, line 260: The org shouldn't be nullable by this point. (If the org is null, they're not a member and there's no point doing the db lookup.)

Did you hit any errors with it not being nullable, or are you just addressing IDE suggestions? I'd rather leave it not nullable here and ensure that callers check for null or `HasValue` first.

(As an aside - we use a lot of tricky pattern matching just to avoid explicit null checks. I wonder whether it's easier to just check for null - but I don't actually know.)

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -96,17 +101,22 @@ public class BulkCollectionAuthorizationHandler : BulkAuthorizationHandler<BulkC
     private async Task CanCreateAsync(AuthorizationHandlerContext context, IAuthorizationRequirement requirement,
         CurrentContextOrganization? org)
     {
-        // If the limit collection management setting is disabled, allow any user to create collections
-        // Otherwise, Owners, Admins, and users with CreateNewCollections permission can always create collections
+        // Owners, Admins, and users with CreateNewCollections permission can always create collections
```
eliykat, line 104: This kinda makes the logic easier to understand as well! It separates out the user role vs. the organization setting.

src/Core/Context/CurrentContext.cs
```diff
@@ -382,7 +385,9 @@ public async Task<bool> ViewAssignedCollections(Guid orgId)
         var org = GetOrganization(orgId);
         if (org != null)
         {
-            canCreateNewCollections = !org.LimitCollectionCreationDeletion || org.Permissions.CreateNewCollections;
+            (await _applicationCacheService.GetOrganizationAbilitiesAsync())
+                .TryGetValue(org.Id, out var organizationAbility);
+            canCreateNewCollections = !organizationAbility?.LimitCollectionCreationDeletion ?? org.Permissions.CreateNewCollections;
```
eliykat, line 390: If you look a little earlier in this method, this isn't used by FC. I think it was initially, then as we replaced more of the collections controller, FC-specific logic was added to an authorization handler.

Therefore, you should be able to remove the `limitCollectionCreationDeletion` references from here entirely, along with the cache service.

test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
```diff
@@ -134,7 +144,6 @@ public class BulkCollectionAuthorizationHandlerTests
         CurrentContextOrganization organization)
     {
         organization.Type = userType;
-        organization.LimitCollectionCreationDeletion = true;
```
eliykat, line 137: This line (whether true or false) has been removed from a lot of tests with no replacement. Is that intentional? In some cases, I think it's required to make sure that the test is testing what we want it test. (test test test) They're not failing, but we also need to think about false negatives.

I suggest adding a private method that tests can call to arrange this value.

eliykat (changes requested):
I think this works!

test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
vincentsalucci: This is intentional. However, I went back through all of these tests revolving around needing the `LimitCollectionCreationDeletion` property and added some explicit calls to make sure we never get flaky tests.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
vincentsalucci: That's fair. Its also the case where we should only be hitting it once per request anyway. 👍

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
vincentsalucci: I wasn't 100% sure on this. The reason its even questionable is because technically from the `CurrentContext` the method `GetOrganization` might not have the organization within the claims, which would return `null`. However, I've reviewed the change and to match the flow of the other requirement checks, I've added a `null` check for the org in that `CanDeleteAsync` conditional and reverted the optional operator from the `IsAssignedToCollectionsAsync` signature and method.

src/Core/Context/CurrentContext.cs
vincentsalucci: Ah - I think you're right. It seems like this method should only be called when the feature flag for MPV is **off**. Will remove with next commit.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -96,17 +100,22 @@ public class BulkCollectionAuthorizationHandler : BulkAuthorizationHandler<BulkC
...
             return;
         }
 
+        // If the limit collection management setting is disabled, allow any user to create collections
+        if (await GetOrganizationAbilityAsync() is { LimitCollectionCreationDeletion: false })
+        {
+            context.Succeed(requirement);
+            return;
+        }
+
```
eliykat, line 118: Just noticed this - you also have to check `org != null` here, otherwise you're calling `context.Succeed` only on the basis of the collection management setting, without actually knowing whether or not the user is a member of that org. It might be worth adding a comment noting this as well, it's easy to overlook.

Please also check any other usages, including in other handlers, to make sure this is covered.

Maybe you could pass the `currentContextOrganization` into `GetOrganizationAbilityAsync`, just to make this link more explicit.

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
```diff
@@ -26,7 +28,6 @@ public class GroupAuthorizationHandlerTests
         CurrentContextOrganization organization)
     {
         organization.Type = userType;
-        organization.LimitCollectionCreationDeletion = true;
```
eliykat, line 29: Continuing on from my last comment about this - you've restored some of the test arrangements, but not all. The concern isn't necessarily flaky tests, it's also the risk that tests pass when they should be failing.

In this case, for example, a user can read all groups if (1) they can create collections or (2) they're an owner or admin. This test is meant to be testing (2), but if this value isn't expressly set to `true` to limit collection creation, it'll always pass because of (1) instead - without effectively testing (2).

(Another side effect of having `true` as a default value instead of `false`! alas)

I think that you need to have a helper method which creates the OrgAbilities object and sets up the mock return on the applicationCache dependency, then replace each deleted line here with a call to that helper method.

eliykat (changes requested):

test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
eliykat: left a further comment about this

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
vincentsalucci: Great catch - thank you!

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
vincentsalucci: Hmm - I'm following. Updated tests and cleaned up existing usages of inline dictionary creation via a helper arrange method.

eliykat (dismissed):

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
eliykat: Looks good! And looks like you removed it from tests that didn't depend on it, thanks for going through it closely.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -96,17 +100,23 @@ public class BulkCollectionAuthorizationHandler : BulkAuthorizationHandler<BulkC
...
         { Type: OrganizationUserType.Owner or OrganizationUserType.Admin } or
         { Permissions.CreateNewCollections: true })
         {
             context.Succeed(requirement);
             return;
         }
 
+        // Check for non-null org here: the user must be apart of the organization for this setting to take affect
+        // If the limit collection management setting is disabled, allow any user to create collections
+        if (org is not null && await GetOrganizationAbilityAsync() is { LimitCollectionCreationDeletion: false })
```
shane-melton, line 114: suggestion (non-blocking): Could we instead pass the `org` into `GetOrganizationAbilityAsync()` to be extra explicit and make these `if ()` branches slightly less complicated and reduce their repetition in the other operations.

We can have `GetOrganizationAbilityAsync()` check for a null `org` and return `null` so we can still use the `await GetOrganizationAbilityAsync() is {...}` branch.

I also imagine we could do the same in the other handlers.

shane-melton (approved):
This looks good to go to me!

I did have one suggestion that I think would improve the  the `GetOrganizationAbilityAsync()` helper and remove some repeated null checks. But, I don't think it should be blocking and maybe we can revisit another time if desired.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
eliykat: This should be done now!

src/Api/Vault/AuthorizationHandlers/OrganizationUsers/OrganizationUserAuthorizationHandler.cs
```diff
@@ -76,10 +79,33 @@ public class OrganizationUserAuthorizationHandler : AuthorizationHandler<Organiz
             return;
         }
 
+        // Check for non-null org here: the user must be apart of the organization for this setting to take affect
+        // If the limit collection management setting is disabled, allow any user to read all organization users
+        if (org is not null && await GetOrganizationAbilityAsync(org) is { LimitCollectionCreationDeletion: false })
```
shane-melton, line 84: @eliykat I think we might have missed removing this extra `org is not null` check

shane-melton (commented):
Thanks for making that change @eliykat! Just saw one small thing that I believe was missed.

rkac-bw (dismissed):
looks good to me

shane-melton (approved):

rkac-bw (approved):
Looks good to me

eliykat:
All changes are flagged. Only exception is adding the new properties to OrgAbility sproc and class, however they are straightforward, should not impact current behaviour, and I've smoke tested them locally.

eliykat:
d'oh - realised that I now need an AC Team review and it should probably not be me now that I've made changes as well. Asked @r-tome to help out.

r-tome (approved):

PR #3574, opened 2023-12-14
[AC-1817] Remove AccessAll logic from miscellaneous sprocs

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Continue to remove AccessAll logic in sprocs. This is a misc. grouping of sprocs that was left over after my last PRs.

## Code changes

The following sprocs have been updated:

**Cipher sprocs**
- Cipher_ReadCanEditByIdUserId.sql
- Cipher_UpdateCollections.sql - in turn called by
  - Cipher_CreateWithCollections
  - Cipher_UpdateWithCollections
  - CipherDetails_CreateWithCollections

**Misc sprocs**
- Folder_DeleteById.sql
- CollectionUser_ReadByOrganizationUserIds.sql
- OrganizationUserUserDetails_ReadWithCollectionsById.sql

The following sprocs have been deleted:
- OrganizationUser_ReadWithCollectionsById.sql (this was unused - see comment in migration file for more details)

The corresponding EF queries have also been updated accordingly.

Calling locations are updated to pass in the `FlexibleCollections` flag value, as in previous PRs.

## Before you submit


src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
```diff
@@ -18,11 +18,12 @@ public interface IOrganizationUserRepository : IRepository<OrganizationUser, Gui
...
     Task<OrganizationUser> GetByOrganizationAsync(Guid organizationId, Guid userId);
-    Task<Tuple<OrganizationUser, ICollection<CollectionAccessSelection>>> GetByIdWithCollectionsAsync(Guid id);
     Task<OrganizationUserUserDetails> GetDetailsByIdAsync(Guid id);
     Task<Tuple<OrganizationUserUserDetails, ICollection<CollectionAccessSelection>>>
-        GetDetailsByIdWithCollectionsAsync(Guid id);
-    Task<ICollection<OrganizationUserUserDetails>> GetManyDetailsByOrganizationAsync(Guid organizationId, bool includeGroups = false, bool includeCollections = false);
+        GetDetailsByIdWithCollectionsAsync(Guid id, bool flexibleCollectionsIsEnabled);
+    Task<ICollection<OrganizationUserUserDetails>> GetManyDetailsByOrganizationAsync(Guid organizationId);
+    Task<ICollection<OrganizationUserUserDetails>> GetManyDetailsByOrganizationAsync(Guid organizationId,
+        bool includeGroups, bool includeCollections, bool flexibleCollectionsIsEnabled);
```
eliykat, line 26: This interface has been split so that we now use overloads instead of optional parameters.

This is useful because the feature flag logic is only required when `includeCollections == true`, and once we make this an overload, we see that this argument is only passed in by one calling location. This is way less to update.

src/Core/Vault/Repositories/IFolderRepository.cs
```diff
@@ -9,6 +9,10 @@ public interface IFolderRepository : IRepository<Folder, Guid>
     Task<Folder> GetByIdAsync(Guid id, Guid userId);
     Task<ICollection<Folder>> GetManyByUserIdAsync(Guid userId);
 
+    [Obsolete("Not used while FlexibleCollections feature flag is in use - use the overload to pass in the flag value instead.")]
+    new Task DeleteAsync(Folder folder);
+    Task DeleteAsync(Folder folder, bool useFlexibleCollections);
```
eliykat, line 14: This is a bit awkward, but the base repository class provides an implementation of `DeleteAsync`, which we want to prevent use of (because you need to use the overload to pass in the flag value).

eliykat:
No longer required.

PR #3575, opened 2023-12-14
[AC-1981] Fix CollectionsController.Get auth check by just checking collections for the requested orgId

r-tome:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
`CollectionsController.Get` would throw a `NotFoundException` if the user was assigned to collections in multiple organizations.

## Code changes

## Before you submit

### DIFF
--- a/src/Api/Controllers/CollectionsController.cs
+++ b/src/Api/Controllers/CollectionsController.cs
@@ -611,11 +611,6 @@ public class CollectionsController : Controller
         {
             var assignedCollections = await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value, FlexibleCollectionsIsEnabled);
             orgCollections = assignedCollections.Where(c => c.OrganizationId == orgId && c.Manage).ToList();
-            var readAuthorized = (await _authorizationService.AuthorizeAsync(User, orgCollections, BulkCollectionOperations.Read)).Succeeded;
-            if (!readAuthorized)
-            {
-                throw new NotFoundException();
-            }
         }
 
         var responses = orgCollections.Select(c => new CollectionResponseModel(c));
--- a/src/Api/Controllers/CollectionsController.cs
+++ b/src/Api/Controllers/CollectionsController.cs
@@ -610,7 +610,7 @@ public class CollectionsController : Controller
         else
         {
             var assignedCollections = await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value, FlexibleCollectionsIsEnabled);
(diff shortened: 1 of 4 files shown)


src/Api/Controllers/CollectionsController.cs
```diff
@@ -609,13 +609,10 @@ private async Task<ListResponseModel<CollectionResponseModel>> GetByOrgId_vNext(
...
-            var readAuthorized = (await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.Read)).Succeeded;
-            if (readAuthorized)
-            {
-                orgCollections = collections.Where(c => c.OrganizationId == orgId);
-            }
-            else
+            var assignedCollections = await _collectionRepository.GetManyByUserIdAsync(_currentContext.UserId.Value, FlexibleCollectionsIsEnabled);
+            orgCollections = assignedCollections.Where(c => c.OrganizationId == orgId).ToList();
+            var readAuthorized = (await _authorizationService.AuthorizeAsync(User, orgCollections, BulkCollectionOperations.Read)).Succeeded;
+            if (!readAuthorized)
```
eliykat, line 615: I still think this is problematic. `orgCollections` will be all collections the user is assigned to for that org, but not necessarily all they have manage permissions for. So the authorizationService call could still fail.

The idea of this endpoint is "give me all the collections the user is allowed to read". If they can't read all, that basically means just the ones they have manage permissions for. If they don't have any, then it returns an empty ListResponseModel. It should basically never throw.

Therefore I think you just want to do `orgCollections = assignedCollections.Where(c => c.OrganizationId == orgId && c.Manage).ToList();` without any further call to `AuthorizationService`.

Yes it's unfortunate that this is effectively repeating auth logic that we're trying to centralize in the handler, but handlers don't do filtering. It's breaking the abstraction a little, but I think we gotta.

Tagging @shane-melton as he was in these discussions as well.

If (if!) my suggestions here are correct, I think it would also apply to the other "filtering" endpoint here.

r-tome: That sounds about right. We did have this discussion for [this other endpoint here](https://github.com/bitwarden/server/blob/b77ee017e379104011c4e01096955e42d34b1b7f/src/Api/Controllers/CollectionsController.cs#L586C14-L586C14). I went ahead and did those modifications, let me know what you think.

eliykat (approved):
lgtm! Thanks for following up on this

shane-melton (approved):

PR #3577, opened 2023-12-14
[PM-5293] Redis for Grants

justindbaur:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Store grants in Redis using MessagePack serialization. 

## Code changes

* **perf/MicroBenchmarks/Identity/IdentityServer/RedisPersistedGrantStoreTests.cs:** Benchmark to compare SQL vs Redis updates
* **src/Core/Settings/GlobalSettings.cs:** Add setting just for grants redis connection string
* **src/Identity/IdentityServer/RedisPersistedGrantStore.cs:** Persisted grant store that stores grants primarily in Redis but can fallback to another store.
* **src/Identity/Utilities/ServiceCollectionExtensions.cs:** Add redis services conditionally behind new config value

## Before you submit


src/Identity/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -45,11 +47,33 @@ public static class ServiceCollectionExtensions
...
+            // Add the original persisted grant store via it's implementation type
+            // so we can inject it right after.
+            services.AddSingleton<PersistedGrantStore>();
+
+            services.AddSingleton<IPersistedGrantStore>(sp =>
+            {
+                return new RedisPersistedGrantStore(
+                    ConnectionMultiplexer.Connect(globalSettings.Grants.RedisConnectionString),
+                    sp.GetRequiredService<ILogger<RedisPersistedGrantStore>>(),
+                    sp.GetRequiredService<PersistedGrantStore>() // Fallback grant store
```
kspearrin, line 67: Why not do this in the ctor?

justindbaur: Specifically the `ConnectionMultiplexer.Connect`? Because if I were to write unit tests for this I could instantiate `RedisPersistedGrantStore` with a mocked version of `IConnectionMultiplexer`. But I don't have such tests and I don't think I would write tests that mock redis behavior vs just running it against redis locally so if you want me to move it there I absolutely can.

kspearrin: This just looks dirty instead of letting DI do its thing. What if we create and register a new IGrantConnectionMultiplexer or something and register that?

perf/MicroBenchmarks/Identity/IdentityServer/RedisPersistedGrantStoreTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            Type = "refresh_token",
+            SessionId = null,
+            ClientId = "web",
+            Expiration = creationTime.AddHours(1),
+        };
+    }
+
+    [Params("Redis", "SQL")]
```
withinfocus, line 70: ⛏️ These two strings could be constants and reused.

src/Core/Settings/GlobalSettings.cs
```diff
@@ -57,6 +57,7 @@ public virtual string LicenseDirectory
     public virtual ConnectionStringSettings Events { get; set; } = new ConnectionStringSettings();
     public virtual IConnectionStringSettings Redis { get; set; } = new ConnectionStringSettings();
     public virtual NotificationsSettings Notifications { get; set; } = new NotificationsSettings();
+    public virtual GrantsSettings Grants { get; set; } = new GrantsSettings();
```
withinfocus, line 60: ❓ Candidate for also being in the interface? Seems random what is or isn't.

withinfocus, line 60: ⚠️ Do we want to overdesign a little here? Could reference this as some kind of "entity caching" and have sub-objects. Alternatively, this could just go in the IdentityServer configuration block.

src/Identity/Utilities/ServiceCollectionExtensions.cs
justindbaur: I can do that but in my opinion the real fix here is to use the new [keyed services](https://andrewlock.net/exploring-the-dotnet-8-preview-keyed-services-dependency-injection-support/) in .NET 8. We can name a specific instance of `IConnectionMultiplexer` to be used for `"grants"`. Then we don't need a special marker interface.

kspearrin: I agree. I don't care either way if we feel it might be better to just wait for that.

kspearrin (dismissed):

JaredSnider-Bitwarden (dismissed):
LGTM after discussion on call to understand original issue and solution!

kspearrin (approved):

JaredSnider-Bitwarden (approved):
LGTM!

justindbaur:
Merging to `main` and then a PR picking this will be made into `hotfix-rc` for testing.

PR #3578, opened 2023-12-14
[SM-1048] Fix linq2db Regression

Thomas-Avery:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The purpose of this PR is to fix a known regression with linq2db 5.3.1 which is implicitly installed by linq2db.EntityFrameworkCore 7.6.0.  

The recommend fix is explicitly adding linq2db 5.3.2 as a dependency.

https://github.com/linq2db/linq2db.EntityFrameworkCore/issues/368

## Code changes

* **src/Infrastructure.EntityFramework/Infrastructure.EntityFramework.csproj:** 
Explicitly add linq2db 5.3.2 as a dependency.

## Before you submit

### DIFF
--- a/src/Infrastructure.EntityFramework/Infrastructure.EntityFramework.csproj
+++ b/src/Infrastructure.EntityFramework/Infrastructure.EntityFramework.csproj
@@ -2,6 +2,7 @@
 
     <ItemGroup>
       <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="12.0.1" />
+      <PackageReference Include="linq2db" Version="5.3.2" />
       <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="7.0.14" />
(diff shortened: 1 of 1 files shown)


coltonhurst (approved):
Looks good, thanks for linking the relevant discussion. Only two questions:

1. You were able to test this works locally?
2. If we don't already have it, can we create an internal tech debt ticket to remove the dependency once the new version is out with the fix? (as mentioned [here](https://github.com/linq2db/linq2db.EntityFrameworkCore/issues/368#issuecomment-1805963057))

Thomas-Avery:
> Looks good, thanks for linking the relevant discussion. Only two questions:
> 
> 1. You were able to test this works locally?
> 2. If we don't already have it, can we create an internal tech debt ticket to remove the dependency once the new version is out with the fix? (as mentioned [here](https://github.com/linq2db/linq2db.EntityFrameworkCore/issues/368#issuecomment-1805963057))

Thank you.

1. Yes 
2. SM-1049

PR #3586, opened 2023-12-16
[PM-2740] Add null check on base64-encoded values on knowndevice query

trmartin4:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
In diagnosing some server exceptions, I noticed that 99% of the 500 error responses from our API service were due to null reference exceptions on the `knowndevice` endpoint.

The query-string implementation had a `null` check on both email and identifier, so I replicated that on this implementation as well.

This should result in a more informative `400` error with a message in it, but primarily the benefit is to allow us to have better visibility of true, unexpected `500` status error messages in our logs.

```
System.NullReferenceException: Object reference not set to an instance of an object.
at Bit.Core.Utilities.CoreHelpers.Base64UrlDecode(String input) in /home/runner/work/server/server/src/Core/Utilities/CoreHelpers.cs:line 370
at Bit.Core.Utilities.CoreHelpers.Base64UrlDecodeString(String input) in /home/runner/work/server/server/src/Core/Utilities/CoreHelpers.cs:line 338
at Bit.Api.Controllers.DevicesController.GetByIdentifierQuery(String email, String deviceIdentifier) in /home/runner/work/server/server/src/Api/Controllers/DevicesController.cs:line 212
```

📓 **Note**
This will still cause 500 error responses if an invalid base64-encoded value is passed in, instead of nothing at all.  The change to address that is larger in scope and will be in a subsequent PR.

## Code changes
- **KnownDeviceRequestModel.cs:** Added new model to represent inputs for known device call.
- **DevicesController.cs:** Modified known device query to use the model for input.

## Before you submit

### DIFF
--- a/src/Api/Controllers/DevicesController.cs
+++ b/src/Api/Controllers/DevicesController.cs
@@ -1,4 +1,5 @@
-﻿using Bit.Api.Auth.Models.Request;
+﻿using Api.Models.Request;
+using Bit.Api.Auth.Models.Request;
 using Bit.Api.Auth.Models.Request.Accounts;
 using Bit.Api.Models.Request;
 using Bit.Api.Models.Response;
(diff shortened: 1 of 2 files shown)


ike-kottlowski (approved):

justindbaur (approved):
I don't think I had ever seen `[FromHeader]` applied on a model property but I like it!

PR #3615, opened 2023-12-23
[PM-5435] Handle Fido2VerificationException on WebAuthn 2FA

trmartin4:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
When there is an error validating the RPID during WebAuthn 2FA, the `Fido2NetLib` library throws a `Fido2VerificationException`.  This exception was not being caught, and it was causing a 500 `Unhandled exception` response to the client.

## Code changes
* **WebAuthnTokenProvider:** Add try/catch around the verification call.  Return `false` (not verified) if there is a verification exception.

## Before you submit

### DIFF
--- a/src/Core/Auth/Identity/WebAuthnTokenProvider.cs
+++ b/src/Core/Auth/Identity/WebAuthnTokenProvider.cs
@@ -103,19 +103,26 @@ public class WebAuthnTokenProvider : IUserTwoFactorTokenProvider<User>
         // established ownership in this context.
         IsUserHandleOwnerOfCredentialIdAsync callback = (args, cancellationToken) => Task.FromResult(true);
 
-        var res = await _fido2.MakeAssertionAsync(clientResponse, options, webAuthCred.Item2.PublicKey, webAuthCred.Item2.SignatureCounter, callback);
+        try {
+            var res = await _fido2.MakeAssertionAsync(clientResponse, options, webAuthCred.Item2.PublicKey, webAuthCred.Item2.SignatureCounter, callback);
 
-        provider.MetaData.Remove("login");
+            provider.MetaData.Remove("login");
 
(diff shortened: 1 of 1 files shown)


src/Core/Auth/Identity/WebAuthnTokenProvider.cs
```diff
@@ -103,19 +103,27 @@ public async Task<bool> ValidateAsync(string purpose, string token, UserManager<
...
+            var providers = user.GetTwoFactorProviders();
+            providers[TwoFactorProviderType.WebAuthn].MetaData[webAuthCred.Item1] = webAuthCred.Item2;
+            user.SetTwoFactorProviders(providers);
+            await userService.UpdateTwoFactorProviderAsync(user, TwoFactorProviderType.WebAuthn, logEvent: false);
+
+            return res.Status == "ok";
+        }
+        catch (Fido2VerificationException)
+        {
+            return false;
```
JaredSnider-Bitwarden, line 124: Question: what does the user see when this returns false now?

trmartin4: Returning `false` here will get passed from the `WebAuthnTokenProvider` back to `VerifyTwoFactor` through the call to `VerifyTwoFactorTokenAsync()` on the `UserManager`, as you can see [here](https://github.com/aspnet/AspNetIdentity/blob/b7826741279450c58b230ece98bd04b4815beabf/src/Microsoft.AspNet.Identity.Core/UserManager.cs#L1439).

As a result, the 2FA response will be considered invalid, a new "needs 2FA" response will be sent back to the client, and the user will be asked to try again or select a new method.

JaredSnider-Bitwarden (approved):

PR #3616, opened 2023-12-23
[PM-5437] Handle client_credentials clientId that is not a valid GUID

trmartin4:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When using the `client_credentials` grant type, the `ClientStore` is used to retrieve the client information by `clientId`.

We were missing a check to make sure that the `clientId` is a valid GUID before instantiating the new GUID object.  This would throw an exception and return a 500 response to users if the provided `clientId` could not be parsed.

## Code changes

* **ClientStore.cs:** Use `TryParse` to make sure the `clientId` is a valid GUID.  Return `null` if not.

## Before you submit


src/Identity/IdentityServer/ClientStore.cs
```diff
@@ -90,7 +90,12 @@ public async Task<Client> FindClientByIdAsync(string clientId)
 
     private async Task<Client> CreateApiKeyClientAsync(string clientId)
     {
-        var apiKey = await _apiKeyRepository.GetDetailsByIdAsync(new Guid(clientId));
+        if (!Guid.TryParse(clientId, out var guid))
+        {
+            return null;
```
JaredSnider-Bitwarden, line 95: Question: What happens / what does the user see when `null` is returned?

trmartin4: My thinking here was to replicate the behavior in the other client type parsing.  See https://github.com/bitwarden/server/blob/0f487f69005b092554d4b178645a03d4ef9a3478/src/Identity/IdentityServer/ClientStore.cs#L229 and https://github.com/bitwarden/server/blob/0f487f69005b092554d4b178645a03d4ef9a3478/src/Identity/IdentityServer/ClientStore.cs#L259, for example.

Also, checking the source code for IdentityServer4 you can see in [this](https://github.com/IdentityServer/IdentityServer4/blob/4dc10e665f5ede63274427036c25cdc216130eb9/src/EntityFramework.Storage/test/IntegrationTests/Stores/ClientStoreTests.cs#L34) test that `null` is the expected response when the client does not exist - as it would be in this case.

PR #3634, opened 2024-01-02
[PM-5645] Cosmos DB Grant Storage

kspearrin:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Migrate Grant data from Redis to CosmosDB storage. Cosmos implementation should fallback to Redis and SQL if Grant is not found in Cosmos. This will allow us to migrate storage services without issue.

https://bitwarden.atlassian.net/browse/PM-5645

## Code changes

perf/MicroBenchmarks/Identity/IdentityServer/RedisPersistedGrantStoreTests.cs

- Updated test to add Cosmos

src/Core/Auth/Entities/Grant.cs

- Add IGrant interface. Also created ctor instead of building the grant from the store class.

src/Core/Auth/Models/Data/GrantItem.cs

- Model representing the Cosmos item.

src/Core/Auth/Models/Data/IGrant.cs

- Interface for grants since the SQL and Cosmos models differ in some ways.

src/Core/Auth/Repositories/Cosmos/Base64IdStringConverter.cs

- This converter is needed to serialize the Base64 grant, which is used as the cosmos `id`. `id` fields cannot contain certain base64 characters, like `\`, so we use base64 URL encoding, instead.

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs

- Cosmos repo implementation

src/Core/Auth/Repositories/IGrantRepository.cs

- Change repo interface to use IGrant.

src/Core/Core.csproj

- Formatting, I guess.

src/Core/Settings/GlobalSettings.cs

- New connection string for Identity cosmos db.

src/Core/Utilities/CoreHelpers.cs

- Refactored the Base64 encoder helpers to separate the string transformations from the byte[] encoding step. This allows for usage of just the transform step, which I needed for the `Base64IdStringConverter`.

src/Core/Utilities/SystemTextJsonCosmosSerializer.cs

- By default, CosmosDB uses Newtonsoft JSON serializer. I added this so we can use `System.Text.Json`, as is standard in our codebase now.

src/Identity/IdentityServer/PersistedGrantStore.cs

- Add fallback store option step to the Get method.

src/Identity/Utilities/ServiceCollectionExtensions.cs

- Added Cosmos option to grant store registration. Also moved all the class instantiations to helper methods to make things easier to read and more DRY.

src/Infrastructure.Dapper/Auth/Repositories/GrantRepository.cs

- Move to IGrant interface.

src/Infrastructure.EntityFramework/Auth/Repositories/GrantRepository.cs

- Move to IGrant interface.

## Before you submit


src/Core/Auth/Models/Data/GrantTableEntity.cs
```diff
@@ -0,0 +1,65 @@
...
+    public string ClientId { get; set; }
+    public string Description { get; set; }
+    public DateTime CreationDate { get; set; } = DateTime.UtcNow;
+    public DateTime? ExpirationDate { get; set; }
+    public DateTime? ConsumedDate { get; set; }
+    public string Data { get; set; }
+
+    public static (string partitionKey, string rowKey) CreateTableEntityKeys(string grantKey)
+    {
+        // TODO: How do we want to partition? Or do we? refs:
```
withinfocus, line 60: I usually look at what we index, but per https://github.com/bitwarden/server/blob/main/src/Sql/Auth/dbo/Tables/Grant.sql there's not much valuable there. The subject ID is technically nullable but never is for us and we put that user ID in there. Separately, would it be valuable to partition on the client ID (which is our type)?

withinfocus: ⚠️ We want to try a null / empty row key right?

src/Core/Auth/Models/Data/IGrant.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Bit.Core.Auth.Models.Data;
+
+public interface IGrant
```
withinfocus, line 3: ℹ️  Our SQL storage of grants now has an integer ID as a primary key -- may want to represent this here.

src/Core/Auth/Repositories/TableStorage/GrantRepository.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public Task DeleteManyAsync(string subjectId, string sessionId, string clientId, string type) => throw new NotImplementedException();
+
+    private CloudTable GetTableShard(IGrant grant)
+    {
+        return GetTableShard(grant.Key);
+    }
+
+    private CloudTable GetTableShard(string key)
```
withinfocus, line 83: 🌱 Might be better to test without this at first, to avoid the complexity if we can.

kspearrin: We can test with just 1 connection string configured and it would be the same

src/Core/Auth/Models/Data/IGrant.cs
kspearrin: I left that on the Grant entity use for SQL since it didn't seem necessary on other implementations

src/Core/Auth/Models/Data/GrantTableEntity.cs
kspearrin: yes

src/Identity/IdentityServer/PersistedGrantStore.cs
```diff
@@ -47,28 +58,11 @@ public async Task RemoveAsync(string key)
 
     public async Task StoreAsync(PersistedGrant pGrant)
     {
-        var grant = ToGrant(pGrant);
+        var grant = _toGrant(pGrant);
         await _grantRepository.SaveAsync(grant);
```
withinfocus, line 62: ❓ No fallback storage? I suppose that's moot.

src/Identity/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -51,31 +52,43 @@ public static class ServiceCollectionExtensions
...
         services.AddTransient<ICorsPolicyService, CustomCorsPolicyService>();
         return identityServerBuilder;
     }
+
+    private static PersistedGrantStore BuildCosmosStore(IServiceProvider sp, GlobalSettings globalSettings)
+    {
+        return new PersistedGrantStore(
+            new Core.Auth.Repositories.Cosmos.GrantRepository(globalSettings),
+            g => new Core.Auth.Models.Data.GrantItem(g),
+            fallbackGrantStore: BuildRedisStore(sp, globalSettings));
```
withinfocus, line 77: 💭 Perhaps a non-issue, but this assumes that if you've configured a Cosmos store that you also have a Redis store.

bitwarden-server.sln
```diff
@@ -94,27 +93,27 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IntegrationTestCommon", "te
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Scim", "bitwarden_license\src\Scim\Scim.csproj", "{BC3B3F8C-621A-4CB8-9563-6EC0A2C8C747}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SqlServerEFScaffold", "util\SqlServerEFScaffold\SqlServerEFScaffold.csproj", "{2F2E8BB0-6838-48DA-B581-71B9F13DE364}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlServerEFScaffold", "util\SqlServerEFScaffold\SqlServerEFScaffold.csproj", "{2F2E8BB0-6838-48DA-B581-71B9F13DE364}"
```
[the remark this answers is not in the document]
kspearrin: Yea, I could probably revert this whole change. It happened when I added a new project for the load tests, which I didn't even commit here.

src/Core/Auth/Models/Data/GrantItem.cs
```diff
@@ -0,0 +1,77 @@
...
+    public DateTime CreationDate { get; set; } = DateTime.UtcNow;
+    [JsonPropertyName("exp")]
+    public DateTime? ExpirationDate { get; set; }
+    [JsonPropertyName("con")]
+    public DateTime? ConsumedDate { get; set; }
+    [JsonPropertyName("data")]
+    public string Data { get; set; }
+    // https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-time-to-live?tabs=dotnet-sdk-v3#set-time-to-live-on-an-item-using-an-sdk
+    [JsonPropertyName("ttl")]
+    public int Ttl { get; set; } = -1;
```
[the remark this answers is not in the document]
kspearrin: It needs to be an actual property that get's mapped with the JSON serializers, so not sure I could make it some other type like Lazy

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            Serializer = new SystemTextJsonCosmosSerializer(new JsonSerializerOptions
+            {
+                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+                WriteIndented = false,
+            })
+        };
+        _client = new CosmosClient(cosmosConnectionString, options);
+        _database = _client.GetDatabase("identity");
+        _container = _database.GetContainer("grant");
```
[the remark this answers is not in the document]
kspearrin: We could, but I am not sure why we would name them something different. The connection string for the Cosmos account to use is config driven already. Is there a use case you can think of that would benefit from this coming from a config prop?

Not sure I understand the property names point you made.

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+    public Task<ICollection<IGrant>> GetManyAsync(string subjectId, string sessionId, string clientId, string type)
+        => throw new NotImplementedException();
+
+    public async Task SaveAsync(IGrant obj)
+    {
+        if (obj is not GrantItem item)
+        {
+            item = new GrantItem(obj);
+        }
+        item.SetTtl();
```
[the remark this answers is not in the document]
kspearrin: I don't think I have to set this here since it is handled in the GrantItem ctor, but I am not sure what things might cause the ExpirationDate to change by the time the ctor is called and this save method is called. That's why I added it here too, to make sure it was computed again just before being saved.

src/Core/Auth/Repositories/Cosmos/SystemTextJsonCosmosSerializer.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Text.Json;
+using Azure.Core.Serialization;
+using Microsoft.Azure.Cosmos;
+
+namespace Bit.Core.Auth.Repositories.Cosmos;
+
+// ref: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs
+public class SystemTextJsonCosmosSerializer : CosmosSerializer
```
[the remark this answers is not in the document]
kspearrin: Perhaps. Do you have a suggestion? We would want to use this for any Cosmos implementations using our own mapped models.

src/Identity/IdentityServer/PersistedGrantStore.cs
kspearrin: I don't think so. The fallback in this case is to just handle the migration to a new system where we need to read from the fallback until not needed anymore.

src/Identity/Utilities/ServiceCollectionExtensions.cs
kspearrin: True, I can add some logic to set a null fallback if redis is not configured.

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs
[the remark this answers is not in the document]
withinfocus: I'm probably just overthinking things for a change here via config we would likely never do, like a need to move to multiple containers.

My other comment is about how proprietary we make the name of a configuration property e.g. "Cosmos" when maybe we could leave it more ambiguous for the future should:

- it not be called Cosmos anymore
- we support other document or table database types behind this
- something else.

This is more of an OCD thing of mine for future-proofing and again it could just not happen or we'd move to a whole new configuration naming.

src/Core/Auth/Repositories/Cosmos/SystemTextJsonCosmosSerializer.cs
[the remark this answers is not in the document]
withinfocus: I dislike it being in a team-owned (Auth) space vs. open to all, and we'd want reuse. It's not great, but I think `Utilities` is a better place.

withinfocus (commented):
Couple areas for potential improvement / change which are the unresolved comments. I don't love where TTL logic is but I am not sure I have a better suggestion.

Something transient with the failed tests? This can go as-is as I see it.

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs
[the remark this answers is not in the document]
kspearrin: I think this is overthinking it a bit. This implementation is pretty specific to cosmos and trying to genericize it would make it less clear what is going on here, IMO.

withinfocus (dismissed):
Not sure what's up with the tests though.

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Net;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Bit.Core.Auth.Models.Data;
+using Bit.Core.Settings;
+using Bit.Core.Utilities;
+using Microsoft.Azure.Cosmos;
```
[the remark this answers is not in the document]
kspearrin: I can follow up with a new PR after this.

JaredSnider-Bitwarden (commented):
The addition of CosmosDB for grant storage is looking excellent overall. Appreciate the tests and refactors! Just one question 👇

src/Core/Auth/Repositories/Cosmos/GrantRepository.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        var options = new CosmosClientOptions
+        {
+            Serializer = new SystemTextJsonCosmosSerializer(new JsonSerializerOptions
+            {
+                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+                WriteIndented = false
+            })
+        };
+        _client = new CosmosClient(cosmosConnectionString, options);
```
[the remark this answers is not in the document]
kspearrin: The Cosmos grant repository is a dependency of the `PersistedGrantStore`, which is registered as a singleton.

We can't register the cosmos grant repository to DI as IGrantRepository since we need multiple instances of `IGrantRepository`. Until we move to .NET 8 with keyed services, I don't think there is a better solution that what we're doing right now. Same with CosmosClient. I think we need to be able to key that. I can add some comments.

withinfocus (approved):

JaredSnider-Bitwarden (approved):
LGTM!

ike-kottlowski (approved):

PR #3640, opened 2024-01-03
Eliminate in-app purchase logic

withinfocus:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


src/Api/Models/Response/SubscriptionResponseModel.cs
```diff
@@ -42,6 +42,7 @@ public SubscriptionResponseModel(User user, UserLicense license = null)
     public BillingSubscription Subscription { get; set; }
     public UserLicense License { get; set; }
     public DateTime? Expiration { get; set; }
+    [Obsolete("In-app purchase payment method is not allowed.")]
     public bool UsingInAppPurchase { get; set; }
```
withinfocus, line 46: ❓ Can I get some info on where this might be used, if anywhere? May not need the obsolete attribute.

src/Core/Enums/PaymentMethodType.cs
```diff
@@ -17,8 +17,10 @@ public enum PaymentMethodType : byte
     [Display(Name = "Wire Transfer")]
     WireTransfer = 5,
     [Display(Name = "Apple In-App Purchase")]
+    [Obsolete("In-app purchase payment method is not allowed.")]
     AppleInApp = 6,
     [Display(Name = "Google In-App Purchase")]
+    [Obsolete("In-app purchase payment method is not allowed.")]
```
withinfocus, line 23: ❓ Same -- do we feel OK with just removing these? Then I can remove the logic checks too, if we believe this type will never come in.

src/Core/Services/IUserService.cs
```diff
@@ -54,11 +54,10 @@ public interface IUserService
     Task<Tuple<bool, string>> SignUpPremiumAsync(User user, string paymentToken,
         PaymentMethodType paymentMethodType, short additionalStorageGb, UserLicense license,
         TaxInfo taxInfo);
-    Task IapCheckAsync(User user, PaymentMethodType paymentMethodType);
     Task UpdateLicenseAsync(User user, UserLicense license);
     Task<string> AdjustStorageAsync(User user, short storageAdjustmentGb);
     Task ReplacePaymentMethodAsync(User user, string paymentToken, PaymentMethodType paymentMethodType, TaxInfo taxInfo);
-    Task CancelPremiumAsync(User user, bool? endOfPeriod = null, bool accountDelete = false);
```
withinfocus, line 61: ⚠️ Would like some expert 👀 on this one. The `accountDelete` parameter was intermixed with in-app purchases. It seemed unused but makes me uncomfortable.

src/Core/Services/Implementations/StripePaymentService.cs
```diff
@@ -1128,8 +1050,7 @@ public async Task<string> PayInvoiceAfterSubscriptionChangeAsync(ISubscriber sub
         return paymentIntentClientSecret;
     }
 
-    public async Task CancelSubscriptionAsync(ISubscriber subscriber, bool endOfPeriod = false,
```
withinfocus, line 1131: ⚠️ The other side of my comment earlier. This is not "account delete"-related.

JaredSnider-Bitwarden (approved):
Approving the removal of the `PostIapCheck` from the auth owned `AccountsController` after verifying it was not used in clients as well.  

Did not review rest of PR.

src/Api/Models/Response/SubscriptionResponseModel.cs
amorask-bitwarden: @withinfocus The only place it's used, as far as I can tell, is in the `user-subscription.component` in `clients` [here](https://github.com/bitwarden/clients/blob/28de9439beb87133c8683434df952a0c0be94100/apps/web/src/app/billing/individual/user-subscription.component.ts#L72), [here](https://github.com/bitwarden/clients/blob/28de9439beb87133c8683434df952a0c0be94100/apps/web/src/app/billing/individual/user-subscription.component.ts#L109) and [here](https://github.com/bitwarden/clients/blob/28de9439beb87133c8683434df952a0c0be94100/apps/web/src/app/billing/individual/user-subscription.component.ts#L172). It just opens certain dialogs restricting the user from managing their subscription a certain way if they're using in-app purchases.

withinfocus: Aha, I see now. I killed those off in https://github.com/bitwarden/clients/pull/7433/commits/18768bbf840132ba7f71aa2f0a31eea1c536f4b9. I can probably remove this now that it's gone from there.

src/Core/Services/Implementations/StripePaymentService.cs
cturnbull-bitwarden: Agreed, this seems to be a non-issue given that you're removing the single use of this parameter.

src/Core/Services/IUserService.cs
cturnbull-bitwarden: Yup, you're in the clear here.

src/Core/Services/Implementations/StripePaymentService.cs
```diff
@@ -377,21 +366,6 @@ private async Task ChangeOrganizationSponsorship(Organization org, OrganizationS
 
         if (user.Gateway == GatewayType.Stripe && !string.IsNullOrWhiteSpace(user.GatewayCustomerId))
         {
-            if (!string.IsNullOrWhiteSpace(paymentToken))
-            {
-                try
-                {
-                    await UpdatePaymentMethodAsync(user, paymentMethodType, paymentToken, true, taxInfo);
```
cturnbull-bitwarden, line 384: 🤔  It's not immediately clear to me why this is getting deleted. What's your thought behind it?

cturnbull-bitwarden (approved):
Thank you! Left one question, but overall approved.

src/Core/Services/Implementations/StripePaymentService.cs
withinfocus: The way I read this, since this code block explicitly calls `UpdatePaymentMethodAsync` with `allowInAppPurchases = true`, it's only relevant for those types. I can see how this might be unsafe however and other payment updates with a token -- are there any? -- might still need this, but then should the exception be suppressed?

cturnbull-bitwarden: I went through and was able to reproduce this. It seems that that flag is only used to throw the `GatewayException`. This code is still hit when a user who previously paid for premium outside of an organization lets their subscription cancel, then resubscribes after it's already cancelled. In that case, the `user.GatewayCustomerId` will still have a value, and the `paymentToken` will be the payment method ID for the card that they set up on the second signup. All that to say, this should probably stay, just sans the boolean!

withinfocus: OK, lemme know if https://github.com/bitwarden/server/pull/3640/commits/83509538b2777585b9755c2b76c828b2b0c44621 works then.

cturnbull-bitwarden (approved):
Thank you!

JaredSnider-Bitwarden (approved):

PR #3643, opened 2024-01-04
[AC-2026] Add UI and endpoint to set organization property

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Add an endpoint and repository method for existing orgs to opt-in to flexible collections.

Client PR here: https://github.com/bitwarden/clients/pull/7443

Data migrations will be added here: https://github.com/bitwarden/server/pull/3437

## Code changes

* **file.ext:** Description of what was changed and why

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3644, opened 2024-01-04
[AC-2027] Update Flexible Collections logic to use organization property

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

We are changing to using an organization property, organization.flexibleCollections, to track whether an org is using the flexible collections changes.

This PR updates the optionality logic in server to refer to this new organization property instead of the feature flag from configService.

As with the collection management settings, we generally access this via the `applicationCacheService` to minimise db calls, unless the org object is already within scope.

Note that we are also using feature flags in a different way:
* `flexible-collections-signup` determines whether **new** organizations are automatically using FC. It sets the `org.FlexibleCollections` property when the org is created (no data migrations required)
* `flexible-collections-migration` determines whether **existing** organizations are able to opt-in to FC. This opt-in process will be done via a new endpoint which will be added in #3643, which will run data migrations and then set `org.FlexibleCollections`.

So we are now gating the entrance to the feature, rather than toggling the feature itself.

The old `flexible-collections` flag will be removed.

There are a few other opinionated changes which I've commented on below.

## Code changes

## Before you submit


src/Api/Controllers/CollectionsController.cs
```diff
@@ -311,6 +306,18 @@ public async Task PostBulkCollectionAccess([FromBody] BulkCollectionAccessReques
...
+        if (collections.All(c => c.OrganizationId != orgId))
+        {
+            throw new BadRequestException("All collections must belong to the same organization");
+        }
+
+        // Authorization logic assumes flexible collections is enabled
+        if (!await FlexibleCollectionsIsEnabledAsync(orgId))
+        {
+            throw new NotFoundException();
+        }
```
eliykat, line 319: This is basically replicating the `[RequireFeature(FeatureFlagKeys.FlexibleCollections)]` guard, except using the org property.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -397,27 +404,17 @@ public async Task DeleteUser(Guid orgId, Guid id, Guid orgUserId)
         await _collectionService.DeleteUserAsync(collection, orgUserId);
     }
 
-    private void DeprecatedPermissionsGuard()
-    {
-        if (FlexibleCollectionsIsEnabled)
-        {
-            throw new FeatureUnavailableException("Flexible Collections is ON when it should be OFF.");
-        }
-    }
```
eliykat, line 406: I propose removing this for a couple of reasons
1. it was probably overkill - we ended up separating the current endpoints and the `vNext` endpoints pretty cleanly, and this controller has reasonable test coverage, so we have some confidence that the correct code paths are being run
2. now that we're doing our data migration earlier, there's less risk in the wrong authorization code being run. Previously, the manager enum was still in the database, so the wrong code path might authorize an unauthorized action. Now managers really will be downgraded to users, even if the old authorization logic is accidentally called, the worst case is that they'll get rejected - which is what this guard did anyway.

Therefore I didn't worry about calling the orgAbilities cache and making all this async - I just removed it. Open for discussion though, it's not a huge lift to restore it.

src/Core/Constants.cs
```diff
@@ -96,12 +96,17 @@ public static class FeatureFlagKeys
     public const string VaultOnboarding = "vault-onboarding";
     public const string AutofillV2 = "autofill-v2";
     public const string BrowserFilelessImport = "browser-fileless-import";
-    public const string FlexibleCollections = "flexible-collections";
+    /// <summary>
+    /// Deprecated - never used, do not use. Will always default to false. Will be deleted as part of Flexible Collections cleanup
+    /// </summary>
+    public const string FlexibleCollections = "flexible-collections-disabled-do-not-use";
```
eliykat, line 102: This is now only being used in the repository calls to use the `v2` sprocs, which don't contain AccessAll logic. However, we're now relying on setting AccessAll to `0`, rather than changing out the sprocs. This can always default to false to use the existing sprocs.

For safety, I intentionally broke the key here so that it can't be turned on. Really it should be deleted but I didn't want to trigger additional codeowner review - I'll clean up the remaining references in https://bitwarden.atlassian.net/browse/AC-2061

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -603,8 +614,12 @@ private async Task ValidateSignUpPoliciesAsync(Guid ownerId)
             UseSecretsManager = license.UseSecretsManager,
             SmSeats = license.SmSeats,
             SmServiceAccounts = license.SmServiceAccounts,
-            LimitCollectionCreationDeletion = !flexibleCollectionsMvpIsEnabled || license.LimitCollectionCreationDeletion,
-            AllowAdminAccessToAllCollectionItems = !flexibleCollectionsV1IsEnabled || license.AllowAdminAccessToAllCollectionItems
+            LimitCollectionCreationDeletion = license.LimitCollectionCreationDeletion,
+            AllowAdminAccessToAllCollectionItems = license.AllowAdminAccessToAllCollectionItems,
```
eliykat, line 618: I simplified this logic as it was too difficult to reason about and explain. It's much easier to just always update from the license file values. At the end of the day, `limitCollectionCreationDeletion` won't do anything unless `organization.FlexibleCollections == true` (which it won't unless we've turned on the migration and/or signup flag) and `allowAdminAccessToAllCollections` won't do anything unless the v1 flag is turned on. So let's just keep this logic simple and rely on the flags to control behaviour like we do everywhere else.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -2007,11 +2027,6 @@ public async Task ValidateOrganizationUserUpdatePermissions(Guid organizationId,
         {
             throw new BadRequestException("Custom users can only grant the same custom permissions that they have.");
         }
-
-        if (FlexibleCollectionsIsEnabled && newType == OrganizationUserType.Manager && oldType is not OrganizationUserType.Manager)
-        {
-            throw new BadRequestException("Manager role is deprecated after Flexible Collections.");
-        }
```
eliykat, line 2014: TODO

eliykat (commented):
A few comments for the reviewer.

test/Core.Test/AdminConsole/Services/OrganizationServiceTests.cs
r-tome, line 1: These tests removed here should also be put back in, right?

r-tome (commented):
Looks great, I just left some minor suggestions, other than that it looks ready.

cyprain-okeke (commented):
The changes around billing looks good. Although is the a request to add a unit test for this change

src/Api/Controllers/CollectionsController.cs
```diff
@@ -311,6 +307,18 @@ public async Task PostBulkCollectionAccess([FromBody] BulkCollectionAccessReques
...
         }
 
+        var orgId = collections.First().OrganizationId;
+        if (collections.All(c => c.OrganizationId != orgId))
+        {
+            throw new BadRequestException("All collections must belong to the same organization");
+        }
+
+        // Authorization logic assumes flexible collections is enabled
+        if (!await FlexibleCollectionsIsEnabledAsync(orgId))
```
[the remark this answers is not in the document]
eliykat: It's because the orgId is taken from the collection objects - which means we need to get the collections from the db, and also check they all belong to the same org (even though this is also covered in the handler).

I rewrote it slightly to use the orgId parameter, it should still have the same effect but addresses your feedback.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -311,6 +307,18 @@ public async Task PostBulkCollectionAccess([FromBody] BulkCollectionAccessReques
             throw new NotFoundException("One or more collections not found.");
         }
 
+        var orgId = collections.First().OrganizationId;
```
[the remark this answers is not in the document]
eliykat: Should be addressed with the previous comment

eliykat:
@cyprain-okeke I added tests for UpdateOrganizationLicenseCommand, it checks that the license file is saved correctly and the org is updated from the license file properties (including the ones we're modifying here). It follows the example in [LicensingServiceTests](https://github.com/bitwarden/server/blob/c2484cd50f8e1b3111397629187914982a4d0bd0/test/Core.Test/Services/LicensingServiceTests.cs) where we have the sut create a real file in a tempdir and then clean it up afterwards.

r-tome (approved):

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -418,6 +416,9 @@ public async Task VerifyBankAsync(Guid organizationId, int amount1, int amount2)
         }
     }
 
+    /// <summary>
+    /// Create a new organization in a cloud environment
```
shane-melton, line 420: praise: Thanks for the summaries clarifying the target environment!

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -30,6 +30,13 @@ public class InMemoryApplicationCacheService : IApplicationCacheService
         return _orgAbilities;
     }
 
+    public async Task<OrganizationAbility> GetOrganizationAbilityAsync(Guid organizationId)
```
shane-melton, line 33: thought/question (non-blocking): Would marking this explicitly as nullable (`Task<OrganizationAbility?>`) be beneficial? We do so in the authorization handler helpers, so it may be useful to continue to communicate that it may be null to any consumers.

shane-melton (dismissed):
This looks good to me, nice work! 

I did have one non-blocking question about the `InMemoryApplicationCacheService`.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
shane-melton: Sounds good, thank you!

shane-melton (approved):

r-tome (approved):

cyprain-okeke (approved):
LGTM

PR #3647, opened 2024-01-05
Improve local migrations performance

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

We have many more migration scripts than we used to and they're taking longer to run after you pull from `main`. For each script, we:
1. query the migrations db to decide if it needs to be run
2. (if required) run it against the vault db
3. (if required) record it against the migrations db

This makes it O(n) for the number of migration scripts we have in the dir, because we have to run the first query for _all_ scripts. Note that we only have to run 2 and 3 for net new migration scripts.

I've refactored this so that step 1 is a single query that is only executed once, regardless of the number of migration scripts in the dir. Now we only have to loop over net new scripts to execute queries 2 and 3.

Using a "no migrations required" scenario as a baseline:

Old: 27 seconds
New: 2 seconds

## Code changes

## Before you submit

### DIFF
--- a/dev/helpers/mssql/.gitignore
+++ b/dev/helpers/mssql/.gitignore
@@ -1,1 +1,0 @@
-all_migrations.txt
--- a/dev/helpers/mssql/.gitignore
+++ b/dev/helpers/mssql/.gitignore
@@ -1,0 +1,1 @@
+all_migrations.txt
--- a/dev/helpers/mssql/ReadRequiredMigrations.sql
+++ b/dev/helpers/mssql/ReadRequiredMigrations.sql
@@ -6,6 +6,8 @@ BEGIN
 
     CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
 
+    -- This is only run locally in dev environments so we are not concerned with SQL injection attacks here.
+    -- This should not be reused in any other context.
     DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @MigrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
     EXEC(@bulkInsertSql);
 
--- a/dev/helpers/mssql/ReadRequiredMigrations.sql
+++ b/dev/helpers/mssql/ReadRequiredMigrations.sql
@@ -6,7 +6,7 @@ BEGIN
 
     CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
 
-    DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @migrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
+    DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @MigrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
     EXEC(@bulkInsertSql);
 
     -- Select migrations that do not appear in the [dbo].[migrations] table
--- a/dev/helpers/mssql/ReadRequiredMigrations.sql
+++ b/dev/helpers/mssql/ReadRequiredMigrations.sql
@@ -1,5 +1,5 @@
 CREATE OR ALTER PROCEDURE ReadRequiredMigrations
-    @migrationsFile NVARCHAR(MAX)
+    @MigrationsFile NVARCHAR(MAX)
 AS
 BEGIN
     SET NOCOUNT ON;
@@ -9,7 +9,7 @@ BEGIN
     DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @migrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
     EXEC(@bulkInsertSql);
 
-    -- Select migrations that do not appear in the [dbo].[Migrations] table
+    -- Select migrations that do not appear in the [dbo].[migrations] table
     SELECT IM.[Filename]
     FROM [#InputMigrations] IM
     LEFT JOIN [dbo].[migrations] M ON IM.[Filename] = M.[Filename]
--- a/dev/helpers/mssql/read_required_migrations.sql
+++ b/dev/helpers/mssql/read_required_migrations.sql
@@ -1,0 +1,21 @@
+-- Create a stored procedure
+CREATE OR ALTER PROCEDURE ReadRequiredMigrations
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    -- Create a temporary table to store the input Filenames
+    CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
+
+    BULK INSERT #InputMigrations FROM "/mnt/helpers/all_migrations.txt" WITH (FIELDTERMINATOR = ';', ROWTERMINATOR = ';')
+
+    -- Select migrations that do not appear in the inputMigrations table using a LEFT JOIN
+    SELECT IM.[Filename]
+    FROM [#InputMigrations] IM
+    LEFT JOIN [dbo].[migrations] M ON IM.[Filename] LIKE '%' + M.[Filename] + '%'
+    WHERE M.[Filename] IS NULL
+    ORDER BY [Filename] ASC
+
+    -- Drop the temporary table
+    DROP TABLE #InputMigrations
+END
--- a/dev/helpers/mssql/read_required_migrations.sql
+++ b/dev/helpers/mssql/read_required_migrations.sql
@@ -1,16 +1,18 @@
 CREATE OR ALTER PROCEDURE ReadRequiredMigrations
+    @migrationsFile NVARCHAR(MAX)
 AS
 BEGIN
     SET NOCOUNT ON;
 
     CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
 
-    BULK INSERT #InputMigrations FROM "/mnt/helpers/all_migrations.txt" WITH (FIELDTERMINATOR = ';', ROWTERMINATOR = ';')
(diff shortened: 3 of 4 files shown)


dev/helpers/mssql/.gitignore
```diff
@@ -0,0 +1 @@
+all_migrations.txt
```
withinfocus, line 1: 💭 How about just cleaning up, or generating a temp file e.g. `tempfile` instead of this?

dev/helpers/mssql/read_required_migrations.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR ALTER PROCEDURE ReadRequiredMigrations
+    @migrationsFile NVARCHAR(MAX)
```
withinfocus, line 2: ⛏️ Might as well follow our SQL naming practices with a leading capital letter.

dev/helpers/mssql/read_required_migrations.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR ALTER PROCEDURE ReadRequiredMigrations
```
withinfocus, line 1: ⛏️ Filename on disk should be `ReadRequiredMigrations.sql` I'd say.

dev/helpers/mssql/read_required_migrations.sql
```diff
@@ -0,0 +1,20 @@
...
+    @migrationsFile NVARCHAR(MAX)
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
+
+    DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @migrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
+    EXEC(@bulkInsertSql);
```
withinfocus, line 10: ℹ️ SQL injection attack 😜 ... not really a thing.

dev/helpers/mssql/read_required_migrations.sql
```diff
@@ -0,0 +1,20 @@
...
+
+    -- Select migrations that do not appear in the [dbo].[Migrations] table
```
withinfocus, line 12: Super ⛏️ The table name is actually lowercase right? But the comment capitalizes.

dev/helpers/mssql/run_migrations.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
 
-# There seems to be [a bug with docker-compose](https://github.com/docker/compose/issues/4076#issuecomment-324932294) 
+# There seems to be [a bug with docker-compose](https://github.com/docker/compose/issues/4076#issuecomment-324932294)
 # where it takes ~40ms to connect to the terminal output of the container, so stuff logged to the terminal in this time is lost.
 # The best workaround seems to be adding tiny delay like so:
 sleep 0.1;
 
 MIGRATE_DIRECTORY="/mnt/migrator/DbScripts"
+HELPERS_DIRECTORY="/mnt/helpers"
```
withinfocus, line 9: 💭 Won't need this if you use a temp file.

dev/helpers/mssql/read_required_migrations.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR ALTER PROCEDURE ReadRequiredMigrations
+    @migrationsFile NVARCHAR(MAX)
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
+
+    DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @migrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
```
rkac-bw, line 9: this requires the sql server to have access to the file which may not always be possible. Might be safer to parse the text file into json app side and pass in the json:

```
CREATE OR ALTER PROCEDURE ReadRequiredMigrations
    @migrationsJson NVARCHAR(MAX)
AS
BEGIN
    SET NOCOUNT ON;

    -- Create a table to hold the JSON data
    CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX));

    -- Insert JSON data into the temporary table
    INSERT INTO #InputMigrations (Filename)
    SELECT [value] FROM OPENJSON(@migrationsJson);

    -- The rest of your procedure remains the same
    SELECT IM.[Filename]
    FROM [#InputMigrations] IM
    LEFT JOIN [dbo].[migrations] M ON IM.[Filename] = M.[Filename]
    WHERE M.[Filename] IS NULL
    ORDER BY [Filename] ASC;

    DROP TABLE #InputMigrations;
END;
```

withinfocus (changes requested):
Few quality of life improvements.

@rkac-bw these scripts and the various migration processes are areas where we can invest a lot and you can get your hands into.

dev/helpers/mssql/run_migrations.sh
```diff
@@ -1,11 +1,12 @@
...
 SERVER='mssql'
 DATABASE="vault_dev"
 USER="SA"
```
rkac-bw, line 12: Can avoid file creation by just using shell variable I think:
```

# Navigate to the migrations directory and create a JSON array of all SQL files
jsonArray=$(cd $MIGRATE_DIRECTORY && ls -1 *.sql | sed 's/^/"/' | sed 's/$/"/' | paste -sd, -)

# Format the JSON array
formattedJsonArray="[$jsonArray]"

# Escape single quotes in JSON content
escapedJsonContent=$(echo $formattedJsonArray | sed "s/'/''/g")

# Prepare the SQL query
query="EXEC ReadRequiredMigrations '$escapedJsonContent'"

# Execute the query using sqlcmd
result=$(/opt/mssql-tools/bin/sqlcmd -S $SERVER -d migrations_$DATABASE -U $USER -P $PASSWD -I -Q "$query" -W -h-1)
echo $result

```

eliykat: This is similar to my initial approach, but I get this error (using your code snippet above):

```bashSqlcmd: '/mnt/helpers/read_required_migrations.sql': Invalid filename.
Msg 105, Level 15, State 1, Server e82f038a69b6, Line 1 Unclosed quotation mark after the character string '["2017-08-19_00_InitialSetup.sql","2017-08-22_00_LicenseCheckScripts.sql","2017-08-30_00_CollectionWriteOnly.sql","2017-09-06_00_CipherDetails.sql","2017-09-08_00_OrgUserCounts.sql","2017-10-25_00_OrgUserUpdates.sql","2017-11-06_00_FamilyPlanAdjustments.sql","2017-11-13_00_IndexTuning.sql","2017-11-24_00_UpdateProcs.sql","2017-12-12_00_Events.sql","2018-02-28_00_LoginUris.sql","2018-03-12_00_FixLoginUris.sql","2018-03-21_00_AdminPortal.sql","2018-04-02_00_Org2fa.sql","2018-04-24_00_CipherQueryTuning.sql","2018-06-11_00_WebVaultUpdates.sql","2018-07-28_00_DbTuning.sql","2018-08-14_00_UserKdf.sql","2018-08-28_00_PremiumOrgAbilities.sql","2018-09-25_00_OrgPurge.sql","2018-10-17_00_ManagerRole.sql","2018-12-19_00_OrgUserTwoFactorEnabled.sql","2019-01-31_00_Transactions.sql","2019-03-01_00_OrgApi.sql","2019-05-01_00_CipherOrgDetailsFix.sql","2019-05-15_00_GroupNameFixes.sql","2019-05-28_00_CollectionCipherImpr
```

There is not actually any unclosed quotation mark, after some googling my best guess was that it's hitting the max command length allowed by the environment and therefore truncating the json string. Hence I switched to importing from a file which can be an arbitrary size/length.

I'm open to suggestions if you know how to solve this error though.

dev/helpers/mssql/read_required_migrations.sql
eliykat: Yes I figured it's not a risk here, but I should probably add a comment to this effect otherwise we'll be paying out a bounty to someone who thinks they've hacked our systems

dev/helpers/mssql/ReadRequiredMigrations.sql
```diff
@@ -0,0 +1,22 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    CREATE TABLE #InputMigrations (Filename NVARCHAR(MAX))
+
+    -- This is only run locally in dev environments so we are not concerned with SQL injection attacks here.
+    -- This should not be reused in any other context.
+    DECLARE @bulkInsertSql NVARCHAR(4000) = 'BULK INSERT #InputMigrations FROM ''' + @MigrationsFile + ''' WITH (FIELDTERMINATOR = '';'', ROWTERMINATOR = '';'')';
+    EXEC(@bulkInsertSql);
```
eliykat, line 12: @withinfocus @rkac-bw is this reasonable or should we just escape it and avoid the explanation?

dev/helpers/mssql/run_migrations.sh
eliykat: I added the temp file but it's still used in 2 locations, seems it useful to keep.

dev/helpers/mssql/.gitignore
eliykat: This is done

dev/helpers/mssql/run_migrations.sh
rkac-bw: This is caused by a broken version of sqlcmd on the tools container that forces textsize session variable of 4096 in the sqlcmd version 13 that it uses. Why not issue the sql commands from powershell like the ef commands but using;
```
 # Load the assembly
Add-Type -AssemblyName "System.Data.SqlClient"
```

dev/helpers/mssql/.gitignore
withinfocus: I could have been more clear here. What you have now works, but I meant the literal `tempfile` command that happens to be OS-specific but more generally I am thinking about `mktemp`: https://www.gnu.org/software/autogen/mktemp.html -- this is usually the approach I've seen.

dev/helpers/mssql/ReadRequiredMigrations.sql
withinfocus: I think it's fine.

eliykat:
Closing this as we'll likely use dbup instead of spending more time here.

PR #3653, opened 2024-01-08
[AC 2035] When a subscription is cancelled, the customer cannot self-server any subscription changes.

cyprain-okeke:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When a subscription is cancelled, any attempt to add Secrets (or change anything else with your subscription) will not work.

## Code changes

## Before you submit


amorask-bitwarden (approved):

src/Billing/Controllers/StripeController.cs
```diff
@@ -156,6 +156,12 @@ public async Task<IActionResult> PostWebhook([FromQuery] string key)
                 // org
                 if (organizationId != Guid.Empty)
                 {
+                    if (subCanceled)
+                    {
+                        var organization = await _organizationRepository.GetByIdAsync(organizationId);
+                        organization.GatewaySubscriptionId = null;
```
cturnbull-bitwarden, line 162: ❓ Shouldn't this apply to `Gateway` and `GatewayCustomerId` as well?

cturnbull-bitwarden (changes requested):
This solution doesn't appear to address the core of the issue. By removing the reference to Stripe, the status does appear to not be "cancelled." However, the user won't be able to upgrade as the upgrade flow won't be able to locate the subscription, given the ID to it was deleted. If we go the route of removing the customer ID and gateway as well, then the user will lose all the invoices they previously paid in Stripe. We'll need a different approach that doesn't involve deleting the reference to Stripe.

PR #3655, opened 2024-01-10
SwaggerUI - persist authorization in dev mode

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

I found an option that will persist your SwaggerUI authorization between page refreshes. This is useful for development because you can change your code, restart your api project, and refresh the page without having to reauthorize every time.

I have enabled it for development only, keeping the stricter behaviour for production environments.

## Code changes

## Before you submit

### DIFF
--- a/src/Api/Startup.cs
+++ b/src/Api/Startup.cs
@@ -288,6 +288,12 @@ public class Startup
                     "Bitwarden Public API");
                 config.OAuthClientId("accountType.id");
                 config.OAuthClientSecret("secretKey");
+
+                // Persist authorization on page refresh - for development use only
+                if (Environment.IsDevelopment())
+                {
+                    config.EnablePersistAuthorization();
+                }
             });
         }
 


src/Api/Startup.cs
```diff
@@ -288,6 +288,12 @@ public void ConfigureServices(IServiceCollection services)
                     "Bitwarden Public API");
                 config.OAuthClientId("accountType.id");
                 config.OAuthClientSecret("secretKey");
+
+                // Persist authorization on page refresh - for development use only
+                if (Environment.IsDevelopment())
```
withinfocus, line 293: ❓ This is already in a conditional for development or self-host, and is there a risk with the latter? I don't know why self-host would want Swagger on in the first place, but this seems like a reasonable addition for all Swagger usages.

eliykat: The authorization persists between page reloads and even after exiting the browser. Therefore I assume it's saving the session token (at best) or unencrypted org api key (at worst) in the browser local storage. I don't care about that for a development environment but I didn't think it was a reasonable default for self-hosted instances, which are production environments.

eliykat: Also worth noting that this authorization gives you broad admin/owner privileges over the organization, including updating subscription details and adding/removing users. So I'd be careful about it in production.

withinfocus (approved):

PR #3656, opened 2024-01-10
[AC-1880] - Public API - Update collection permission associations with Manage property

vincentsalucci:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Code changes
* **src/Api/AdminConsole/Public/Models/AssociationWithPermissionsBaseModel.cs** Added `Manage` property with XMLdoc explaining its purpose
* **src/Api/AdminConsole/Public/Models/Request/AssociationWithPermissionsRequestModel.cs** Added `Manage` property to the request model - defaults to `false` if not set
* **src/Api/AdminConsole/Public/Models/Response/AssociationWithPermissionsResponseModel.cs** Added `Manage` property to the response model

## Before you submit


r-tome (approved):

src/Api/AdminConsole/Public/Models/Request/AssociationWithPermissionsRequestModel.cs
```diff
@@ -10,7 +10,8 @@ public CollectionAccessSelection ToCollectionAccessSelection()
         {
             Id = Id.Value,
             ReadOnly = ReadOnly.Value,
-            HidePasswords = HidePasswords.GetValueOrDefault()
+            HidePasswords = HidePasswords.GetValueOrDefault(),
+            Manage = Manage.GetValueOrDefault()
```
eliykat, line 14: I suggest that we only want to set this if `org.FlexibleCollections == true`. Otherwise I think it'll show up as a blank label in the web vault, because FC changes won't be enabled. And in any case, we don't want to let users start using FC features until their org has been migrated.

eliykat: After thinking about it, you should probably throw instead, just to provide unambiguous feedback to the user.

src/Api/AdminConsole/Public/Controllers/MembersController.cs
```diff
@@ -119,7 +122,9 @@ public async Task<IActionResult> List()
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     public async Task<IActionResult> Post([FromBody] MemberCreateRequestModel model)
     {
-        var associations = model.Collections?.Select(c => c.ToCollectionAccessSelection());
+        (await _applicationCacheService.GetOrganizationAbilitiesAsync())
+            .TryGetValue(_currentContext.OrganizationId.Value, out var organizationAbility);
```
eliykat, line 126: I added a `GetOrganizationAbilityAsync(orgId)` method you can use here instead. 😁 and anywhere else in this PR that you've called this method.

src/Api/AdminConsole/Public/Models/Request/AssociationWithPermissionsRequestModel.cs
```diff
@@ -1,17 +1,27 @@
...
         };
+
+        // Throws if the org has not migrated to use FC but has passed in a Manage value in the request
+        if (!migratedToFlexibleCollections && Manage.HasValue)
+        {
+            throw new BadRequestException(
+                "Your organization has not migrated to use Flexible Collections and cannot make use of the Manage property.");
+        }
+
+        collectionAccessSelection.Manage = Manage.GetValueOrDefault();
```
eliykat, line 24: You could still do this where you had it in the `CollectionAccessSelection` initializer. That way all the property assignments are together, and you still throw if it's not allowed.

src/Api/AdminConsole/Public/Models/Request/AssociationWithPermissionsRequestModel.cs
```diff
@@ -1,17 +1,27 @@
...
-            HidePasswords = HidePasswords.GetValueOrDefault(),
-            Manage = Manage.GetValueOrDefault()
+            HidePasswords = HidePasswords.GetValueOrDefault()
         };
+
+        // Throws if the org has not migrated to use FC but has passed in a Manage value in the request
+        if (!migratedToFlexibleCollections && Manage.HasValue)
+        {
+            throw new BadRequestException(
+                "Your organization has not migrated to use Flexible Collections and cannot make use of the Manage property.");
```
eliykat, line 21: Suggested alternative: "Your organization must be using the latest collection enhancements to use the Manage property." This aligns slightly more with the error message wording I added in https://github.com/bitwarden/server/pull/3671 (although that's still in code review so feedback is welcome).

In particular, flexible collections is an internal name only. If you like you can just swap out "Flexible Collections" for "collection enhancements" and that would also work.

eliykat (changes requested):

src/Api/AdminConsole/Public/Controllers/MembersController.cs
vincentsalucci: Sneaky! I like it - thank you!

src/Api/AdminConsole/Public/Models/Request/AssociationWithPermissionsRequestModel.cs
vincentsalucci: Fair!

src/Api/AdminConsole/Public/Models/Request/AssociationWithPermissionsRequestModel.cs
vincentsalucci: Totally fine with getting on the same page with our messaging. Thank you!

eliykat (approved):

PR #3657, opened 2024-01-10
[PM-5665] use proper Grant entity model for Dapper mapping

kspearrin:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Dapper can't map to an interface, so use proper entity  class when issuing dapper queries.

## Code changes

* **GrantRepository:** Change dapper mapping to proper class.

## Before you submit


ike-kottlowski (approved):
I would have thought Dapper would have handled Interfaces.

PR #3661, opened 2024-01-11
[AC-1843] Automate PM discount for SM Trial

cturnbull-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

When a customer signs up for a Secrets Manager trial or plan, they will automatically subscribe to Password Manager. The Password Manager subscription should be automatically discounted. The sm-standalone discount in Stripe will be used to comp everything but Secrets Manager.

## Code changes

* **OrganizationCreateRequestModel.cs:** Added `IsFromSecretsManagerTrial` bool which determines if the discount should be applied
* **SubscriptionResponseModel.cs:** Added `AppliesTo` which tells the frontend which `productId`s a discount/coupon applies to. Also added `ProductId` to the subscription item so that it can be compared to the `AppliesTo` array on the frontend
* **OrganizationService.cs:** Now passing in `signup.IsFromSecretsManagerTrial` to `_paymentService.PurchaseOrganizationAsync`
* **OrganizationUpgrade.cs:** Added `IsFromSecretsManagerTrial`, which will determine whether the password manager discount should be applied
* **SubscriptionInfo.cs:** Added `AppliesTo` and `ProductId` similar to above
* **IPaymentService.cs:** Added optional parameter `signupIsFromSecretsManagerTrial` to `PurchaseOrganizationAsync`
* **StripePaymentService.cs:** Added constant containing the discount ID for the secrets manager discount. Added optional parameter `signupIsFromSecretsManagerTrial` to `PurchaseOrganizationAsync` and now applying the coupon if its value is `true`

## Before you submit


amorask-bitwarden (commented):
Source code looks good to me, but I think you have some tests that need to be cleaned up.

amorask-bitwarden (approved):

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -507,7 +507,7 @@ public async Task VerifyBankAsync(Guid organizationId, int amount1, int amount2)
             await _paymentService.PurchaseOrganizationAsync(organization, signup.PaymentMethodType.Value,
```
vincentsalucci, line 507: Non-blocking: might be nice to refactor this method to take an object created from the `signup` object/properties instead of the slow-n-steady bloat of method signature size.

vincentsalucci (approved):
Only reviewed files owned by ac team. One small non-blocking nit. Otherwise, ship! 👍

PR #3662, opened 2024-01-11
[PM-5519] [PM-5526] [PM-5624] [PM-5600] Tweak EF settings for MySQL grant auto-increment

withinfocus:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Use manual SQL to create `Grant`'s new `Id` as a primary key and auto-incrementing column. We don't really work with auto-incrementing integers here and while the new column works fine for MSSQL (with defaults of 1 as the seed and increment by 1) our MySQL library can't generate the migration correctly for when data is present (see https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1828).

## Before you submit

### DIFF
--- a/src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
+++ b/src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
@@ -14,8 +14,7 @@ public class GrantEntityTypeConfiguration : IEntityTypeConfiguration<Grant>
 
         builder
             .Property(s => s.Id)
-            .UseIdentityColumn()
-            .ValueGeneratedOnAdd();
+            .UseIdentityColumn();
 
         builder
             .HasIndex(s => s.Key)
--- a/src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
+++ b/src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
@@ -14,7 +14,8 @@ public class GrantEntityTypeConfiguration : IEntityTypeConfiguration<Grant>
 
         builder
             .Property(s => s.Id)
-            .UseIdentityColumn();
+            .UseIdentityColumn()
+            .ValueGeneratedOnAdd();
 
         builder
             .HasIndex(s => s.Key)
--- a/src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
+++ b/src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
@@ -13,6 +13,10 @@ public class GrantEntityTypeConfiguration : IEntityTypeConfiguration<Grant>
             .IsClustered();
 
         builder
+            .Property(s => s.Id)
+            .UseIdentityColumn();
+
+        builder
             .HasIndex(s => s.Key)
             .IsUnique(true);
 
--- a/src/Sql/Auth/dbo/Tables/Grant.sql
+++ b/src/Sql/Auth/dbo/Tables/Grant.sql
@@ -1,6 +1,6 @@
 ﻿CREATE TABLE [dbo].[Grant]
 (
-    [Id] INT NOT NULL IDENTITY,
+    [Id] INT NOT NULL IDENTITY(1,1),
     [Key] NVARCHAR (200) NOT NULL,
     [Type] NVARCHAR (50) NOT NULL,
     [SubjectId] NVARCHAR (200) NULL,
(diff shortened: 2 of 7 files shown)


util/MySqlMigrations/Migrations/20231214162533_GrantIdWithIndexes.cs
```diff
@@ -78,7 +78,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
             table: "Grant",
             type: "int",
             nullable: false,
-            defaultValue: 0)
+            defaultValue: 1)
```
withinfocus, line 81: ℹ️ Override for what's generated (0). 0 is not allowed here. Same goes for the other two.

util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
```diff
@@ -488,7 +489,8 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                 {
                     b.Property<int>("Id")
                         .ValueGeneratedOnAdd()
-                        .HasColumnType("int");
+                        .HasColumnType("int")
+                        .HasAnnotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn);
```
withinfocus, line 493: ℹ️ Another override. The generated code put a Postgres annotation here instead which doesn't make sense. It also did it for SQLite.

util/SqliteMigrations/Migrations/DatabaseContextModelSnapshot.cs
```diff
@@ -486,7 +486,8 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                 {
                     b.Property<int>("Id")
                         .ValueGeneratedOnAdd()
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("INTEGER")
+                        .HasAnnotation("Sqlite:Autoincrement", true);
```
withinfocus, line 490: ℹ️ Manual addition to match the migration.

util/MySqlMigrations/Migrations/20231214162533_GrantIdWithIndexes.cs
rkac-bw: I don't think defaultvalue as a line is needed, the identity will always provide value

rkac-bw: `migrationBuilder.AddColumn<int>(
    name: "Id",
    table: "Grant",
    type: "int",
    nullable: false)
    .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn);
`

src/Infrastructure.EntityFramework/Auth/Configurations/GrantEntityTypeConfiguration.cs
```diff
@@ -12,6 +12,10 @@ public void Configure(EntityTypeBuilder<Grant> builder)
             .HasKey(s => s.Id)
             .IsClustered();
 
+        builder
+            .Property(s => s.Id)
+            .UseIdentityColumn();
+
```
withinfocus, line 18: ℹ️ This is for posterity and has no effect but I wanted to point it out explicitly.

src/Sql/Auth/dbo/Tables/Grant.sql
```diff
@@ -1,6 +1,6 @@
 ﻿CREATE TABLE [dbo].[Grant]
 (
-    [Id] INT NOT NULL IDENTITY,
+    [Id] INT NOT NULL IDENTITY(1,1),
```
withinfocus, line 3: ℹ️ Same as the above -- MSSQL handles this automatically and it's implied. This migration worked fine for us in our cloud.

util/MySqlMigrations/Migrations/20231214162533_GrantIdWithIndexes.cs
```diff
@@ -73,13 +72,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
             .OldAnnotation("MySql:CharSet", "utf8mb4");
 
-        migrationBuilder.AddColumn<int>(
-            name: "Id",
-            table: "Grant",
-            type: "int",
-            nullable: false,
-            defaultValue: 0)
-            .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn);
+        migrationBuilder.Sql("ALTER TABLE `Grant` ADD COLUMN `Id` INT AUTO_INCREMENT UNIQUE;");
```
withinfocus, line 75: ℹ️ Note that I make this a unique index for the moment, then add the named primary key. I couldn't find a way to one-line the primary key creation _with a name_ -- it could be created but not named.

ike-kottlowski (approved):

rkac-bw (approved):
looks good to me

PR #3671, opened 2024-01-16
[AC-2052] Block Manager role and AccessAll if using FlexibleCollections

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

If an organization is using Flexible Collections, they should not be able to:
* set `Group.AccessAll = true`
* set `OrganizationUser.AccessAll = true`
* set `OrganizationUser.Type = OrganizationUserType.Manager` (this was already implemented but has been tweaked in this PR)

## Code changes

* `CreateGroupCommand` - checks `Group.AccessAll` on creation
* `UpdateGroupCommand` - checks `Group.AccessAll` on update
* `OrganizationService.SaveUsersSendInvitesAsync` - checks AccessAll and user type on creation (invite). Note that all creation flows end up at this private method, so I put the logic here to ensure it's always executed
* `OrganizationService.SaveUserAsync` - checks AccessAll and user type on update

I also found a couple of places where `AccessAll` was still being set to `true` on org creation. That now conditionally sets that flag based on whether the org is using FC or not.

Note that these methods are used by the private and public API, so we get public API changes for free.

## Before you submit


test/Core.Test/AdminConsole/Services/OrganizationServiceTests.cs
```diff
@@ -378,7 +378,7 @@ public async Task SignUpAsync_InvalidateServiceAccount_ShouldThrowException(Orga
 
     [Theory]
     [OrganizationInviteCustomize(InviteeUserType = OrganizationUserType.User,
-         InvitorUserType = OrganizationUserType.Owner), BitAutoData]
+         InvitorUserType = OrganizationUserType.Owner), OrganizationCustomize(FlexibleCollections = false), BitAutoData]
```
eliykat, line 381: I've updated all existing tests to set FlexibleCollections to `false` explicitly, to avoid flaky tests accidentally tripping this new code.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -2027,15 +2052,6 @@ public async Task ValidateOrganizationUserUpdatePermissions(Guid organizationId,
...
         }
-
-        // TODO: pass in the whole organization object when this is refactored into a command/query
-        // See AC-2036
-        var organizationAbility = await _applicationCacheService.GetOrganizationAbilityAsync(organizationId);
-        var flexibleCollectionsEnabled = organizationAbility?.FlexibleCollections ?? false;
-        if (flexibleCollectionsEnabled && newType == OrganizationUserType.Manager && oldType is not OrganizationUserType.Manager)
-        {
-            throw new BadRequestException("Manager role is deprecated after Flexible Collections.");
-        }
```
eliykat, line 2038: This was in the wrong place - this method checks that the current user has permissions to invite/update the target user. Therefore it's only executed when there is a current user, which is not always the case (e.g. when updating via the public api)

r-tome (dismissed):
LGTM!

vincentsalucci (approved):

vincentsalucci (approved):

PR #3676, opened 2024-01-16
[AC-1124] Restrict admins from accessing items in Collections tab

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

Restrict admins from accessing organization ciphers when the organization collection setting is disabled.

See corresponding client PR: https://github.com/bitwarden/clients/pull/7537

### `src/Api/Vault/Controllers/CiphersController.cs`

- Update `/ciphers/organization-details` endpoint to always return all collections for admins when flexible collections V1 is enabled. Otherwise, continue with existing behavior.
  - In flexible collections V1, the endpoint will become only available to admin/owners/providers. Others should use the new `/ciphers/organization-details/assigned` endpoint.
- Add `/ciphers/organization-details/assigned` endpoint to return all ciphers the requesting user has access to for an organization. This includes unassigned ciphers for admin/owners/providers.
- Add access helpers that will be moved to their own service / auth handler in AC-2062

### `src/Core/Vault/Models/Data/CipherDetails.cs`
- Add new `CipherDetailsWithCollections` domain model to include both cipher permissions and collectionIds in the new endpoint response. Normally the admin endpoints have excluded the permission info as it was unnecessary and available via sync data. The new `/ciphers/organization-details/assigned` endpoint is now expected to include this info so the appropriate  edit/delete options can be presented to the user in the admin console when admins are restricted from editing ciphers. 

### `src/Core/Vault/Queries/IOrganizationCiphersQuery.csc`

- Introduce the `IOrganizationCiphersQuery` service to group cipher queries related to a specific organization. This should ultimately replace organizational cipher query methods in the `CipherService` once flexible collections is fully release.

### `src/Core/Vault/VaultServiceCollectionExtensions.cs`
- Introduce `AddVaultServices` service collection extension method to add additional Vault services (currently only the `IOrganizationCiphersQuery`).

### `src/Core/Vault/Repositories/ICipherRepository.cs`
- Add new `GetManyUnassignedOrganizationDetailsByOrganizationIdAsync` method to retrieved organization ciphers that are **not** assigned to any collections.

### `EntityFramework/Vault/Repositories/Queries/CipherOrganizationDetailsReadByOrganizationIdQuery.cs`
- Update the EF query to support querying unassigned ciphers only to match the functionality added via the new sproc for Dapper.

### `src/Sql/Vault/dbo/Stored Procedures/Cipher/CipherOrganizationDetails_ReadUnassignedByOrganizationId.sql`
- New MSSQL sproc for querying unassigned organization ciphers.

## Before you submit


Jingo88 (approved):

util/Migrator/DbScripts/2024-01-16_00_AddUnassignedCiphersQuery.sql
```diff
@@ -0,0 +1,31 @@
+CREATE OR ALTER PROCEDURE [dbo].[CipherOrganizationDetails_ReadUnassignedByOrganizationId]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
```
rkac-bw, line 6: It would be more efficient to not use the not in clause and do where:

```

SELECT
    C.*,
    CASE
        WHEN O.[UseTotp] = 1 THEN 1
        ELSE 0
    END [OrganizationUseTotp]
FROM
    [dbo].[CipherView] C
LEFT JOIN
    [dbo].[OrganizationView] O ON O.[Id] = C.[OrganizationId]
LEFT JOIN
    [dbo].[CollectionCipher] CC ON C.[Id] = CC.[CipherId]
LEFT JOIN
    [dbo].[Collection] S ON S.[Id] = CC.[CollectionId]
    AND S.[OrganizationId] = C.[OrganizationId]
WHERE
    C.[UserId] IS NULL
    AND C.[OrganizationId] = @OrganizationId
    AND CC.[CipherId] IS NULL
```

shane-melton: Thanks for the suggestion! Updated in [db7da53](https://github.com/bitwarden/server/pull/3676/commits/db7da53a99292dc3b8560d4aa9945ab4661258f5)

rkac-bw (dismissed):
Looks good to me

Jingo88 (approved):

rkac-bw (approved):
Looks good to me

Jingo88 (approved):

PR #3679, opened 2024-01-17
Establish IFeatureService as scoped

withinfocus:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Migrates `IFeatureService` to be scoped and independently registers a singleton for the LaunchDarkly client.

## Code changes

* **src/Core/Services/IFeatureService.cs:** Removal of `ICurrentContext` as a method parameter in several places as it can now be injected.
* **src/Core/Services/Implementations/LaunchDarklyFeatureService.cs:** Use of the injected context and client. Exposes logic that used to be in the constructor as a static method so it can be contained here and not pollute service extension methods.
* **src/Events/Startup.cs / src/SharedWeb/Utilities/ServiceCollectionExtensions.cs:** Use of the new registration pattern.
* **test/Core.Test/Services/LaunchDarklyFeatureServiceTests.cs:** Use of the context and client.

The remaining changes are just cleanup given the interface dropping the context parameter.

## Before you submit

### DIFF
--- a/src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
+++ b/src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
@@ -4,16 +4,25 @@ using Bit.Core.Utilities;
 using LaunchDarkly.Logging;
 using LaunchDarkly.Sdk.Server;
 using LaunchDarkly.Sdk.Server.Integrations;
+using LaunchDarkly.Sdk.Server.Interfaces;
 
 namespace Bit.Core.Services;
 
-public class LaunchDarklyFeatureService : IFeatureService, IDisposable
+public class LaunchDarklyFeatureService : IFeatureService
 {
-    private readonly LdClient _client;
+    private readonly ILdClient _client;
+    private readonly ICurrentContext _currentContext;
     private const string _anonymousUser = "25a15cac-58cf-4ac0-ad0f-b17c4bd92294";
 
     public LaunchDarklyFeatureService(
-        IGlobalSettings globalSettings)
+        ILdClient client,
+        ICurrentContext currentContext)
+    {
+        _client = client;
+        _currentContext = currentContext;
+    }
+
+    public static Configuration GetConfiguredClient(GlobalSettings globalSettings)
     {
         var ldConfig = Configuration.Builder(globalSettings.LaunchDarkly?.SdkKey);
         ldConfig.Logging(Components.Logging().Level(LogLevel.Error));
@@ -64,7 +73,7 @@ public class LaunchDarklyFeatureService : IFeatureService, IDisposable
             ldConfig.Offline(true);
         }
 
-        _client = new LdClient(ldConfig.Build());
+        return ldConfig.Build();
     }
 
     public bool IsOnline()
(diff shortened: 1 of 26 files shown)


src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
```diff
@@ -4,16 +4,25 @@
 using LaunchDarkly.Logging;
 using LaunchDarkly.Sdk.Server;
 using LaunchDarkly.Sdk.Server.Integrations;
+using LaunchDarkly.Sdk.Server.Interfaces;
 
 namespace Bit.Core.Services;
 
-public class LaunchDarklyFeatureService : IFeatureService, IDisposable
```
withinfocus, line 10: ℹ️ This concerns me as it's against guidance from LaunchDarkly, but it's not available when using `ILdClient`. Any alternatives?

src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
```diff
@@ -4,16 +4,25 @@
...
     public LaunchDarklyFeatureService(
-        IGlobalSettings globalSettings)
+        ILdClient client,
+        ICurrentContext currentContext)
+    {
+        _client = client;
+        _currentContext = currentContext;
+    }
+
+    public static Configuration GetConfiguredClient(GlobalSettings globalSettings)
```
justindbaur, line 25: I would normally saw that this type of thing should be in the service factory but since this is more code than would normally be in a factory I do like that it stays here.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -727,4 +727,15 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
             });
         });
     }
+
+    public static void AddOptionality(this IServiceCollection services,
+        GlobalSettings globalSettings)
+    {
+        services.AddSingleton<ILdClient>(_ =>
+        {
+            return new LdClient(LaunchDarklyFeatureService.GetConfiguredClient(globalSettings));
```
justindbaur, line 736: I would get `GlobalSettings` out of the `IServiceProvider` that is currently discarded. This gets rid of a closure and makes the `AddOptionality` method not need to supply it.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -727,4 +727,15 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
             });
         });
     }
+
+    public static void AddOptionality(this IServiceCollection services,
```
justindbaur, line 731: Returning the `IServiceCollection` is generally standard so calls can be chained. We don't consistently follow it but I prefer it.

justindbaur (changes requested):
Thank you!

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
withinfocus: Nice, I like it. There's a bunch of these extension methods not doing that today.

justindbaur (approved):

eliykat (approved):
Nice quality of life improvement, thanks!

r-tome (approved):

jlf0dev (approved):
Nice!

gbubemismith (approved):
LGTM

PR #3686, opened 2024-01-19
[PM-5788] Ensure Collection Service respects Flexible Collections falg

shane-melton:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The `CollectionService.SaveAsync()` method assumed that if the V1 FC flag was enabled for a user it implied that FC MVP was also enabled. However, now that we moved FC MVP to opt-in and moved the feature flag to the organization table – that may not always be the case. So this adds an additional check that the organization also has FC enabled before enforcing FC rules when saving a Collection.

## Code changes

* **src/Core/Services/Implementations/CollectionService.cs:** Add check for `org.FlexibleCollections`

## Before you submit


eliykat (approved):
Good catch! I didn't think of that but you're right.

PR #3304, opened 2023-09-28
[PM-3565] Enforce higher minimum KDF

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Extract KDF logic into a new Range class. Increase minimum iterations for PBKDF.

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #3687, opened 2024-01-19
[PM-3777[PM-3633] Update minimum KDF iterations when creating new User record

trmartin4:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
When a TDE user that doesn't have a master password has their MP created via admin account recovery, the KDF iterations for their `User` record on the server are used to generate the master key.  

In all other cases in which a user creates an account, the KDF iterations are defined and passed in from the client, so this server-side default is overwritten.  

We can't do that in this case because the user generating the password is not the user whose account it is; the person generating the admin account recovery has to use the KDF iterations defined in the database for the user.

This change brings the default created on new `User` table entries up to the default, using the range structure that was introduced in #3304 .

## Code changes
- `RegisterRequestModel.cs` and `User.cs`:
  - Updated initialization of `KdfIterations` to use the default value instead of 5000
- Tests
  - Updated tests to use the default value

## Before you submit


ike-kottlowski (approved):
Excellent, removing hard coded values is always the right move, thanks!

PR #3691, opened 2024-01-22
[PM-5819] `user.displayName` should be empty string and not null if no account name has been set

coroiu:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Set `user.DisplayName` to empty string if `null` to comply with WebAuthn spec

## Before you submit


jlf0dev (approved):
Assuming we don't need to deal with existing entries, approved!

jlf0dev (approved):
Assuming we don't need to deal with existing entries, approved!

PR #3694, opened 2024-01-23
Cosmos DB for Events

kspearrin:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


src/Core/Models/Data/EventItem.cs
```diff
@@ -0,0 +1,78 @@
...
+        SecretId = e.SecretId;
+        ServiceAccountId = e.ServiceAccountId;
+    }
+
+    [JsonPropertyName("id")]
+    public string Id { get; set; }
+
+    [JsonPropertyName("type")]
+    public EventType Type { get; set; }
+    [JsonPropertyName("ip")]
```
withinfocus, line 40: 🎨 Trying to save space with some of these? I would think we'll never need to look at the DB and always use this mapping so maybe even more compressed? Reminds me of a JWT token.

src/Core/Repositories/Cosmos/EventRepository.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            Serializer = new SystemTextJsonCosmosSerializer(new JsonSerializerOptions
+            {
+                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+                WriteIndented = false
+            }),
+            // ref: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/tutorial-dotnet-bulk-import
+            AllowBulkExecution = true
+        };
+        // TODO: Perhaps we want to evaluate moving this to DI as a keyed service singleton in .NET 8
```
withinfocus, line 33: 🎨 I'd assume 8 for this change. We are really close to finishing the upgrade.

src/Core/Repositories/Cosmos/EventRepository.cs
```diff
@@ -0,0 +1,152 @@
...
+        _client = new CosmosClient(cosmosConnectionString, options);
+        // TODO: Better naming here? Seems odd
+        _database = _client.GetDatabase("events");
+        _container = _database.GetContainer("events");
```
withinfocus, line 37: 💭 https://learn.microsoft.com/en-us/azure/cosmos-db/resource-model

Could we use containers for organizations? Something else?

Or: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/model-partition-example

Are events then the container, and the database is something company-level for us?

PR #3696, opened 2024-01-23
[PM-3561] Clean the return url of any whitespace

kspearrin:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Resolves potential security issues related to whitespace allowing open redirects with the SSO return URL.

## Code changes

* **AccountController.cs:** Clean the return url of any whitespace characters, such as space and tab.

## Before you submit


bitwarden_license/src/Sso/Controllers/AccountController.cs
```diff
@@ -209,6 +209,8 @@ public IActionResult ExternalChallenge(string scheme, string returnUrl, string s
             returnUrl = "~/";
         }
 
+        // Clean the returnUrl of any whitespace characters
+        returnUrl = returnUrl.Replace(" ", string.Empty).Replace("\t", string.Empty);
```
ike-kottlowski, line 213: I did some quick sleuthing, but nothing conclusive on whether or not "\r" or "\n" also need to be addressed?

ike-kottlowski (commented):
One question about other whitespace encodings.

bitwarden_license/src/Sso/Controllers/AccountController.cs
withinfocus: 🎨 To pile on a bit, I think it's a good idea to put this in a utility method somewhere instead of just here.

Few approaches here: https://stackoverflow.com/questions/6219454/efficient-way-to-remove-all-whitespace-from-string

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -868,4 +869,9 @@ public static string GetEmailDomain(string email)
 
         return null;
     }
+
+    public static string ReplaceWhiteSpace(string input, string newValue)
```
withinfocus, line 873: 🌱 Valid Codecov comment -- would be nice to add a handful of tests.

withinfocus (dismissed):
Solid, but maybe throw in some tests.

ike-kottlowski (approved):

withinfocus (approved):

ike-kottlowski (approved):

PR #3702, opened 2024-01-25
[AC-2077] Set a minimum number of seats for the tested Organization

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Fix some issues with tests related to inviting new users to an Organization. These problems sometimes occur because AutoFixture sets the total number of seats below the required count.

## Code changes

* **test/Core.Test/AdminConsole/AutoFixture/OrganizationFixtures.cs:** Set the seat number to be 100 minumum.

## Before you submit


test/Core.Test/AdminConsole/AutoFixture/OrganizationFixtures.cs
```diff
@@ -24,12 +24,14 @@ public void Customize(IFixture fixture)
     {
         var organizationId = Guid.NewGuid();
         var maxCollections = (short)new Random().Next(10, short.MaxValue);
+        var seats = (short)new Random().Next(100, short.MaxValue);
```
eliykat, line 27: 100 is a bit arbitrary, the minimum seat value is 1 so I suggest using that.

eliykat (changes requested):
This isn't the only reason they're failing. For example, see this run on `main` just now: https://github.com/bitwarden/server/runs/21286479623#r4s99

> Bit.Core.Test.Services.OrganizationServiceTests ► InviteUser_SsoOrgWithNullSsoConfig_Passes(organization: Organization { AllowAdminAccessToAllCollectionItems = False, BillingEmail = "07cbdbc2-4052-4418-b00a-b6850f964209", BusinessAddress1 = "63115829-6...
>
>Failed test found in:
  test/Core.Test/TestResults/oss-test-results.trx
>Error:
  System.AggregateException : One or more errors occurred while inviting users. (Plan does not allow additional seats.)
  ---- Bit.Core.Exceptions.BadRequestException : Plan does not allow additional seats.

So I think you also need to review what plan(s) are used for these tests and fix them to values we know will work. Preferably using this OrganizationCustomization.

test/Core.Test/AdminConsole/AutoFixture/OrganizationFixtures.cs
r-tome: @eliykat I have added a new property to configure the `PlanType` on `OrganizationCustomization`. It defaults to `PlanType.EnterpriseAnnually` as that plan works for all the tests that use that customization.

test/Core.Test/AdminConsole/AutoFixture/OrganizationFixtures.cs
```diff
@@ -19,17 +19,24 @@ public class OrganizationCustomization : ICustomization
...
     public bool FlexibleCollections { get; set; }
+    public PlanType PlanType { get; set; }
 
     public void Customize(IFixture fixture)
     {
         var organizationId = Guid.NewGuid();
         var maxCollections = (short)new Random().Next(10, short.MaxValue);
+        var plan = StaticStore.Plans.FirstOrDefault(p => p.Type == PlanType);
+        var seats = (short)new Random().Next(plan.PasswordManager.BaseSeats, plan.PasswordManager.MaxSeats ?? short.MaxValue);
+        var smSeats = (short)new Random().Next(plan.SecretsManager.BaseSeats, plan.SecretsManager.MaxSeats ?? short.MaxValue);
```
eliykat, line 30: `plan.SecretsManager` can be null if the plan doesn't offer it - e.g. families. This will work fine with the default (Enterprise) but potentially not otherwise.

Suggestion only:

```cs
var smSeats = plan.SupportsSecretsManager 
  ? (short)new Random().Next(plan.SecretsManager.BaseSeats, plan.SecretsManager.MaxSeats ?? short.MaxValue) 
  : null;
```

eliykat (dismissed):
Nice solution! Just one potential issue but otherwise good to go.

test/Core.Test/AdminConsole/AutoFixture/OrganizationFixtures.cs
r-tome: Big oversight on my part, thanks for spotting this!

eliykat (approved):

PR #3719, opened 2024-01-29
[AC-2106] Add check for providers and additional check for null response

shane-melton:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When permission details were add to the collection `PUT/POST` response in #3658 an additional check for provider users was missed and would cause an exception to be thrown because provider users are not explicitly assigned to the managed organization's collections. 

This PR modifies that logic to exclude trying to fetch the permission details when a provider user is making the request on behalf of the managed organization.

## Code changes

* **src/Api/Controllers/CollectionsController.cs:** Don't fetch additional collection permissions for Provider users as the permissions will not be available. Also add a null check when returning the response to avoid any unexpected null reference exceptions.

## Before you submit

### DIFF
--- a/src/Api/Controllers/CollectionsController.cs
+++ b/src/Api/Controllers/CollectionsController.cs
@@ -251,7 +251,7 @@ public class CollectionsController : Controller
 
         await _collectionService.SaveAsync(collection, groups, users);
 
-        if (!_currentContext.UserId.HasValue)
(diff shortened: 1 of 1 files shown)


shane-melton:
Closed in favor of a combined solution with https://github.com/bitwarden/server/pull/3721

PR #3720, opened 2024-01-30
[AC-2068] Allows Users to read all users/groups when Flexible Collections is enabled

shane-melton:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Flexible collections allows any user to manage their own collections. In order to share their collection with other members/groups of the organization they need to be able to list all other users/groups in the organization. This drastically simplifies the `OrganizationUser` and `Group` authorization handlers so that they only need to check that the acting user is a member of the target organization or is a provider user.

## Code changes

* **GroupAuthorizationHandler.cs** and **OrganizationUserAuthorizationHandler.cs:**
  *  Remove extra permission and organization setting checks and replace with a simple null check to ensure the acting user is a member of the target organization. 
* Update the tests to reflect the new behavior.

## Before you submit


r-tome (approved):
Great work! The logic is clear and way simpler

gbubemismith (approved):

PR #3725, opened 2024-01-31
[AC-2114] Custom role should be converted to a user when beta is turned on

r-tome:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
`Delete assigned collections` should be granted `Can Manage` to collections 
`Edit assigned collections` should be granted `Can Edit` access to collections
Custom users with no other permissions should be migrated to User type

## Code changes

* **src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql:** Update the sproc to correctly migrate the users
* **util/Migrator/DbScripts/2024-01-31_00_Organization_EnableCollectionEnhancements_MigrateCustomUsers.sql:** Migration script

## Before you submit

### DIFF
--- a/src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
+++ b/src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
@@ -1,6 +1,5 @@
 CREATE PROCEDURE [dbo].[Organization_EnableCollectionEnhancements]
     @OrganizationId UNIQUEIDENTIFIER
-WITH RECOMPILE
 AS
 BEGIN
     SET NOCOUNT ON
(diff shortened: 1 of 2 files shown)


eliykat (commented):
Your description here does not match the requirements in the ticket.

You say:

> Delete assigned collections should be granted Can Manage to collections
> Edit assigned collections should be granted Can Edit access to collections
> Custom users with no other permissions should be migrated to User type

But the requirements in the ticket say:

> manage assigned collections should be granted Can Manage to collections with a user role
> delete assigned collections should retain collection access they’ve been directly assigned to from the collection tab with a user role
> edit assigned collections should be granted Can Edit access to collections with a user role

However, I also think those requirements misunderstand the current custom permissions:
1. Manage Assigned Collections is only a combination of the create/edit/delete custom permissions ([ref](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts#L217-L220)). Given that it's basically just a shortcut to ticking those 3 boxes, I don't think we should have any migration logic based on that.
2. Edit Any Collection is equivalent to the Can Manage collection permission (because you can edit the collection itself and its access), not the Can Edit permission (which only lets you edit the collection items)

Lastly, and this is more subjective, I think that going to the trouble of converting custom users back to regular users is not worth it; this is already an intensive migration script, do we really need to add more JSON processing to it? If a user is left as a custom user with no permissions ticked, that is already equivalent to a regular User, so frankly I wasn't going to worry about it. Might be worth discussing how important this is.

rkac-bw (dismissed):
Looks good to me, runs slower with the json parsing but looks to be necessary

r-tome:
@eliykat the ticket requirements were changed after I created this PR.

> However, I also think those requirements misunderstand the current custom permissions:
> 
> 1. Manage Assigned Collections is only a combination of the create/edit/delete custom permissions ([ref](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts#L217-L220)). Given that it's basically just a shortcut to ticking those 3 boxes, I don't think we should have any migration logic based on that.
> 2. Edit Any Collection is equivalent to the Can Manage collection permission (because you can edit the collection itself and its access), not the Can Edit permission (which only lets you edit the collection items)

On the second point I think you meant "Edit Assigned Collections". If so then I agree with you.

Regarding the JSON parsing, as @rkac-bw pointed out above, it does make the migrations run slower (more than double the time) so I reckon its not worth. I'll talk to product about it.

r-tome:
@eliykat I have talked with Product and the only change required here is to migrate Custom users (with only 'Edit/Delete assigned collections') to the standard User role

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
```diff
@@ -21,15 +21,34 @@ BEGIN
...
         WHERE ou.[OrganizationId] = @OrganizationId
-            AND (ou.[Type] = 3 OR (ou.[Permissions] IS NOT NULL
-                AND ISJSON(ou.[Permissions]) > 0 AND JSON_VALUE(ou.[Permissions], '$.editAssignedCollections') = 'true'));
+              AND (ou.[Type] = 3 OR
+                   (ou.[Type] = 4 AND
+                    ou.[Permissions] IS NOT NULL AND
+                    ISJSON(ou.[Permissions]) > 0 AND
+                    (JSON_VALUE(ou.[Permissions], '$.editAssignedCollections') = 'true' OR
+                     JSON_VALUE(ou.[Permissions], '$.deleteAssignedCollections') = 'true')
+               ));
```
eliykat, line 51: This appears to parse the JSON at least 4 times:
* 3x calls to JSON_VALUE
* 1x call to OPENJSON

...plus probably some level of parsing involved in `ISJSON` in order to validate the format.

My guess is that SQL probably isn't smart enough to cache this (a quick Google [seems to confirm this](https://dba.stackexchange.com/questions/332558/performance-implications-of-openjson-in-view)). This may be why it had such an outsized impact on performance.

Is there a way to rewrite this to minimise JSON queries; can you use OPENJSON to parse it once and then query those results? e.g. parse it and then JOIN it to your `OrganizationUser` query, instead of parsing it in nested queries for each value?

Alternatively, you could make this change to the code instead - per your closed server PR - and avoid parsing JSON in SQL altogether. At this stage the opt-in is pure SQL migrations so I'm OK to keep the changes here, just mentioning it for completeness.

The performance is probably not noticeable for individual opt-in but seems like it will cause more work when running it in bulk.

eliykat (commented):
Some ideas on performance

PS. 

> On the second point I think you meant "Edit Assigned Collections". If so then I agree with you.

Yes, exactly what I meant 😁

src/Sql/dbo/Stored Procedures/Organization_EnableCollectionEnhancements.sql
r-tome: Took your suggestion and created new PR here https://github.com/bitwarden/server/pull/3770

PR #3760, opened 2024-02-07
[PM-6137] Fix invalid Swagger generation in knowndevice

dani-garcia:
## Type of change
```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Using `[FromHeader]` without a header name on a request parameter generates an invalid Swagger schema, so we need to lift the parameters out of the class to make sure the generated schema is correct.


JaredSnider-Bitwarden (approved):
LGTM. Seems weird that Swagger can't figure out `[FromHeader]` on the request model and on each property within it though.

dani-garcia:
Yeah, been an open issue in Swashbuckle for a while https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1938

Thankfully this is the only place where FromHeader is used this way.

JaredSnider-Bitwarden:
Thank you for linking the issue. Interesting stuff.

PR #3762, opened 2024-02-07
[PM 5864] Resolve root cause of double-charging customers with implementation of PM-3892

cyprain-okeke:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Modify the invoicing system to introduce a dollar threshold. For customers on monthly subscriptions, update the logic to generate invoices only when the accumulated cost of seat increases exceeds the threshold amount. Adjust the invoicing process to consolidate multiple seat increases within a single invoice if they meet the dollar threshold.

Threshold (aligned with billing CS team) = $500

## Code changes

## Before you submit

https://github.com/bitwarden/server/assets/108260115/a69b2bec-6117-47a9-859a-c9c1c831c23b

https://github.com/bitwarden/server/assets/108260115/898f151e-911c-44cc-8c6b-2e3f87df744b


src/Core/Services/Implementations/StripePaymentService.cs
```diff
@@ -762,16 +764,44 @@ private async Task ChangeOrganizationSponsorship(Organization org, OrganizationS
         var daysUntilDue = sub.DaysUntilDue;
         var chargeNow = collectionMethod == "charge_automatically";
         var updatedItemOptions = subscriptionUpdate.UpgradeItemsOptions(sub);
+        var isPm5864DollarThresholdEnabled = _featureService.IsEnabled(FeatureFlagKeys.PM5864DollarThreshold);
 
         var subUpdateOptions = new Stripe.SubscriptionUpdateOptions
         {
             Items = updatedItemOptions,
-            ProrationBehavior = "always_invoice",
+            ProrationBehavior = isPm5864DollarThresholdEnabled && invoiceNow
```
cturnbull-bitwarden, line 772: ⚠️ This doesn't quite look right. We want to `always_invoice` if the feature flag isn't true _or_ if `invoiceNow` is true.

src/Core/Services/Implementations/StripePaymentService.cs
```diff
@@ -762,16 +764,44 @@ private async Task ChangeOrganizationSponsorship(Organization org, OrganizationS
...
+            var immediatelyInvoice = upcomingInvoiceWithChanges.AmountRemaining >= 5000;
+
+            subUpdateOptions.BillingCycleAnchor = immediatelyInvoice
+                ? SubscriptionBillingCycleAnchor.Now
+                : SubscriptionBillingCycleAnchor.Unchanged;
+        }
+
+        if (invoiceNow && isPm5864DollarThresholdEnabled)
+        {
+            subUpdateOptions.BillingCycleAnchor = SubscriptionBillingCycleAnchor.Now;
```
cturnbull-bitwarden, line 802: ⛏️ This appears to be a duplicate of what you have on line `795`. Also, the "invoice now" behavior is handled already on `773` by setting the proration behavior to `always_invoice`, so there should be no need to add this block here.

src/Core/Services/Implementations/StripePaymentService.cs
```diff
@@ -762,16 +764,44 @@ private async Task ChangeOrganizationSponsorship(Organization org, OrganizationS
...
+            {
+                Customer = storableSubscriber.GatewayCustomerId,
+                Subscription = storableSubscriber.GatewaySubscriptionId,
+                SubscriptionItems = ToInvoiceSubscriptionItemOptions(updatedItemOptions),
+                SubscriptionProrationBehavior = Constants.CreateProrations,
+                SubscriptionProrationDate = prorationDate,
+                SubscriptionBillingCycleAnchor = SubscriptionBillingCycleAnchor.Now
+            });
+
+            // also check to see if any of the plan IDs are for an annual plan, or if the update is for a plan upgrade
```
cturnbull-bitwarden, line 792: :pick: This comment can be removed

cturnbull-bitwarden (changes requested):
Looks good overall! Left a few smaller bits of feedback

src/Core/Services/Implementations/StripePaymentService.cs
```diff
@@ -762,16 +764,44 @@ private async Task ChangeOrganizationSponsorship(Organization org, OrganizationS
...
+            var immediatelyInvoice = upcomingInvoiceWithChanges.AmountRemaining >= 5000;
```
cturnbull-bitwarden, line 793: If the threshold is supposed to be $500, then this constant should be `50000` instead.

src/Core/Services/Implementations/StripePaymentService.cs
cyprain-okeke: resolved

src/Core/Services/Implementations/StripePaymentService.cs
cyprain-okeke: resolved

src/Core/Services/Implementations/StripePaymentService.cs
cyprain-okeke: resolved

src/Core/Services/Implementations/StripePaymentService.cs
cyprain-okeke: resolved

cturnbull-bitwarden (approved):
:tada:

amorask-bitwarden (changes requested):
I'm not sure this is actually achieving the desired effect. Are you able to include screenshots or a video of the end result in Stripe for an update that doesn't hit the dollar threshold? 

What I'm seeing when I run this locally is that, regardless of whether the dollar threshold is hit and we update the subscription accordingly, [this block](https://github.com/bitwarden/server/blob/pm-5864-resolve-root-cause-of-double-charging/src/Core/Services/Implementations/StripePaymentService.cs#L843-L851) is still invoked, which tries to pay for the invoice. 

Then, when we attempt to pay the invoice in `PayInvoiceAfterSubscriptionChangeAsync`, this [invocation](https://github.com/bitwarden/server/blob/pm-5864-resolve-root-cause-of-double-charging/src/Core/Services/Implementations/StripePaymentService.cs#L1046-L1049) actually throws an exception saying the invoice has already been finalized. Also the `cardPaymentMethodId` is null, for some reason. 

This is what the resulting upcoming invoice looks like in Stripe, which doesn't seem right to me.
image

amorask-bitwarden (approved):
Looks good! Thanks for resolving that and adding videos.

PR #3766, opened 2024-02-08
[AC-2184] Force sync after opting in to Flexible Collections

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When an org opts in to Flexible Collections, the user who triggered the migration has their vault synced so that they see the changes immediately. However, that isn't done for other owners and admins, so if they have the web vault open they will continue to see the old layouts until they sync their vault or refresh the page.

We can use the push notification system to make them do a full sync. We don't currently do this en masse, but I figure this is limited to owners/admins, and will presumably only actually trigger a sync for those who are actually logged in. Nonetheless I'd like @withinfocus's confirmation that this is OK.

This is fairly low stakes in my opinion. It's a fairly brief edge case (you have to be sitting there with the Admin Console open around the time of migration);  server-side validation will prevent the user from doing anything not permitted; and if this solution isn't acceptable, we can just update the opt-in copy with a warning instead. However I did think this was the better solution that was worth presenting.

## Code changes

## Before you submit


kspearrin (approved):

withinfocus (dismissed):

src/Api/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -846,6 +849,14 @@ public async Task EnableCollectionEnhancements(Guid id)
 
         organization.FlexibleCollections = true;
         await _organizationService.ReplaceAndUpdateCacheAsync(organization);
+
+        // Force a vault sync for all owners and admins of the organization so that changes show immediately
+        // Custom users are intentionally not handled as they are likely to be less impacted and we want to limit simultaneous syncs
+        var orgUsers = await _organizationUserRepository.GetManyByOrganizationAsync(id, null);
```
eliykat, line 855: Previously this used a db call that returned `OrganizationUserUserDetails`, which includes a bunch of unnecessary information from the `User` table. This is a simpler call as it just returns `OrganizationUser` which is a single table. It doesn't filter by `minStatus` so we need to filter the results ourselves but I still think that's likely to be more efficient.

eliykat:
Added tests and used a simpler db call.

PR #3770, opened 2024-02-08
[AC-2114] Downgrade Custom roles to User if flexible collections are enabled and only active permissions are 'Edit/Delete assigned collections'

r-tome:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Downgrade Custom roles to User if flexible collections are enabled and only active permissions are 'Edit/Delete assigned collections'

## Code changes

* **src/Api/AdminConsole/Controllers/OrganizationUsersController.cs:** Check if Flexible Collections is enabled for Organization and if true then downgrade the user role
* **src/Api/AdminConsole/Models/Response/ProfileOrganizationResponseModel.cs:** Check if Flexible Collections is enabled for Organization and if true then downgrade the user role

## Before you submit


eliykat (dismissed):
I like this approach because we don't have to add JSON parsing to the migration script, and we don't have to change the migration script when it's already been through testing. I wasn't sure when I mentioned it but seeing it I think it looks good.

This generally lgtm except that we also need it for the public api which returns organizationUser information. See `MembersController`.

Also, I've created AC-2188 as a follow-up task after full release to clean up the data and remove this transitional code.

vincentsalucci (approved):

PR #3772, opened 2024-02-08
[PM-6177] Persistent distributed cache using Cosmos

kspearrin:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
We have various short-lived items that need persistence, such as TOTP code, identity tickets, etc. Today they use Redis, which by default does not provide the persistence that we need and is arguably not the right technology for this use case.

Add a new type of IDistributedCache that can be used for the persistent things using Cosmos with TTL as the backing store.

A follow-up PR will implement this new IDistributedCache in the appropriate places.

## Before you submit

### DIFF
--- a/src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
+++ b/src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
@@ -48,6 +48,8 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc.Localization;
+using Microsoft.Azure.Cosmos.Fluent;
+using Microsoft.Extensions.Caching.Cosmos;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -693,16 +695,33 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
...
+        if (!string.IsNullOrEmpty(globalSettings.DistributedCache?.Cosmos?.ConnectionString))
         {
-            options.Configuration = globalSettings.Redis.ConnectionString;
-        });
+            services.AddKeyedSingleton<IDistributedCache>("persistent", (s, _) =>
+                new CosmosCache(new CosmosCacheOptions
+                {
+                    DatabaseName = "cache",
+                    ContainerName = "default",
+                    CreateIfNotExists = false,
```
withinfocus, line 717: ❓ What's the impact of turning this on?

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -693,16 +695,33 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
...
-        services.AddStackExchangeRedisCache(options =>
+        if (!string.IsNullOrEmpty(globalSettings.DistributedCache?.Cosmos?.ConnectionString))
         {
-            options.Configuration = globalSettings.Redis.ConnectionString;
-        });
+            services.AddKeyedSingleton<IDistributedCache>("persistent", (s, _) =>
+                new CosmosCache(new CosmosCacheOptions
+                {
+                    DatabaseName = "cache",
+                    ContainerName = "default",
```
withinfocus, line 716: 🌱 I want us to keep thinking about this and its planning for ideal expansion of use cases before this merges.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
kspearrin: It automagically creates it. I'd rather us be explicit about how it is created and configured.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
kspearrin: I think this structure allows for expansion by implementing additional keyed services for IDistributedCache if we need/want to. Each of these could have their own database and/or container.

withinfocus: Something about these names is too ... straightforward / default (quite literally) for me, but I guess I don't have a ... default alternative.

withinfocus (approved):

PR #3784, opened 2024-02-09
[PM-5518] Refactor Email Token Providers

kspearrin:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Refactor email token providers to use persistent `IDistributedCache` as the backing store for randomly generated tokens. This replaces existing TOTP functionality for email token, which has been determined to be inappropriate for our use cases of emailed tokens.

## Code changes

* **EmailTokenProvider.cs:** Get rid of calls into user service and keep all implementation details inside the token provider. Generate and validate randomly generated alpha and/or numeric strings using persistent `IDistributedCache`. Remove code related to the two-factor use case.
* **EmailTwoFactorTokenProvider.cs:** Inherited implementation of `EmailTokenProvider` for two factor use cases. Does true token generation and validation instead of faking generation results with redacted emails like before.
* **UserService.cs:** Call the `UserManager` (base class) generate/validate two-factor token methods, which in turn calls into our new `EmailTwoFactorTokenProvider`.
* **CoreHelpers.cs:** Moved email redaction code into here.
* **BaseRequestValidator.cs:** Generate redacted email from core helper rather than getting it from token provider results.
* **appsettings.Development.json:** Add mail config to Identity dev project environment. This mirrors what we do in Api. Need this to send 2FA emails on login.
* **ServiceCollectionExtensions.cs:** Register our custom token providers for `CoreHelpers.CustomProviderName(TwoFactorProviderType.Email)` and `TokenOptions.DefaultEmailProvider` instead of using the ASP.NET `EmailTokenProvider`.

## Before you submit


src/Core/Auth/Identity/EmailTokenProvider.cs
```diff
@@ -10,100 +9,55 @@ namespace Bit.Core.Auth.Identity;
...
+        var code = CoreHelpers.SecureRandomString(TokenLength, TokenAlpha, true, false, TokenNumeric, false);
+        var cacheKey = string.Format(CacheKeyFormat, user.Id, user.SecurityStamp, purpose);
+        await _distributedCache.SetAsync(cacheKey, Encoding.UTF8.GetBytes(code), _distributedCacheEntryOptions);
+        return code;
     }
 
     public async Task<bool> ValidateAsync(string purpose, string token, UserManager<User> manager, User user)
     {
-        var cacheKey = string.Format(CacheKeyFormat, user.Id, token);
+        var cacheKey = string.Format(CacheKeyFormat, user.Id, user.SecurityStamp, purpose);
```
kspearrin, line 47: Mirroring some of the implementation details from [ASP.NET Core's EmailTokenProvider](https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Extensions.Core/src/EmailTokenProvider.cs), which includes security stamp as a parameter in the token generation. We can include it as part of the cache key to get similar benefits.

kspearrin:
One issue that I thought about here is self hosted customers who are using memory-based IDistributedCache (the default) will run into issues when token validation is needed across projects. For example, when using the "Send verification code email again" option, which happens from the API, but validation is done in Identity.

kspearrin:
> One issue that I thought about here is self hosted customers who are using memory-based IDistributedCache (the default) will run into issues when token validation is needed across projects. For example, when using the "Send verification code email again" option, which happens from the API, but validation is done in Identity. Perhaps we should move that API endpoint to Identity? Or expect customers to start using a persistent cache implementation. Perhaps we could use SQL Server as a backing store.

This concern is being addressed with https://github.com/bitwarden/server/pull/3791

kspearrin:
> This is looking good to me so far, just some failing tests

Fixed tests

jlf0dev (approved):

PR #3791, opened 2024-02-12
[PM-5518] Sql-backed IDistributedCache

kspearrin:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Self-hosted customers and dev environments need a Distributed Cache store. Use SQL, with the ofifical Microsoft library for SQL Server and a custom implementation for EF. EF implementation copies the official SQL Server one pretty closely.

## Code changes

See inline self-review comments.

## Before you submit


src/Infrastructure.EntityFramework/Models/Cache.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Bit.Infrastructure.EntityFramework.Models;
+
+public class Cache
+{
+    [StringLength(499)]
```
withinfocus, line 7: :x: Different value than the table definition.

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
```diff
@@ -72,6 +73,7 @@ protected override void OnModelCreating(ModelBuilder builder)
 
         // Going forward use `IEntityTypeConfiguration` in the Configurations folder for managing
         // Entity Framework code first database configurations.
+        var eCache = builder.Entity<Cache>();
```
withinfocus, line 76: 🎨 Per the above comment let's use the separate configuration files. There's a few examples e.g. `Grant`.

src/Sql/dbo/Tables/Cache.sql
```diff
@@ -0,0 +1,14 @@
+﻿CREATE TABLE [dbo].[Cache] (
+    [Id] [nvarchar](449) NOT NULL,
+    [Value] [varbinary](max) NOT NULL,
+    [ExpiresAtTime] [datetimeoffset](7) NOT NULL,
+    [SlidingExpirationInSeconds] [bigint] NULL,
+    [AbsoluteExpiration] [datetimeoffset](7) NULL,
```
withinfocus, line 6: ⛏️ Minor casing and syntax differences from what we normally do.

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
```diff
@@ -0,0 +1,71 @@
...
+        _deleteExpiredCachedItemsDelegate = DeleteExpiredCacheItems;
+    }
+
+    public byte[] Get(string key)
+    {
+        ArgumentNullException.ThrowIfNull(key);
+        var cache = _databaseContext.Cache
+            .Where(c => c.Id == key && c.ExpiresAtTime >= _systemClock.UtcNow)
+            .SingleOrDefault();
+        ScanForExpiredItemsIfRequired();
```
withinfocus, line 31: 💭 Strange placement to me. I guess this has its protections in place to not run too often but I wonder where it can run, and more async perhaps?

kspearrin: Yes, see implementation below. It runs as a Task only every 30 minutes (`_expiredItemsDeletionInterval`). This is mostly copying https://github.com/dotnet/aspnetcore/blob/main/src/Caching/SqlServer/src/SqlServerCache.cs

src/Sql/dbo/Tables/Cache.sql
kspearrin: Yes, this just matches the generated table expected for SqlServer IDistCache when you run the `dotnet sql-cache create` command. See https://learn.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-8.0#distributed-sql-server-cache

src/Infrastructure.EntityFramework/Models/Cache.cs
kspearrin: Matches the SqlServer table definition below.

withinfocus: One is 449, other is 499.

kspearrin: Totally missed that!

src/Infrastructure.EntityFramework/Configurations/CacheEntityTypeConfiguration.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Bit.Infrastructure.EntityFramework.Models;
```
kspearrin, line 1: Explanation: Configure the Cache table in EF.

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
```diff
@@ -0,0 +1,304 @@
+﻿using Bit.Infrastructure.EntityFramework.Models;
```
kspearrin, line 1: Explanation: IDistributedCache implementation for EF. Mostly copies the SqlServerCache implementation from Microsoft here: https://github.com/dotnet/aspnetcore/blob/main/src/Caching/SqlServer/src/SqlServerCache.cs

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -68,41 +68,7 @@ public static class ServiceCollectionExtensions
...
-                    break;
-            }
-        }
-        else
-        {
-            // Default to attempting to use SqlServer connection string if globalSettings.DatabaseProvider has no value.
-            connectionString = globalSettings.SqlServer.ConnectionString;
-        }
-
+        var (provider, connectionString) = GetDatabaseProvider(globalSettings);
```
kspearrin, line 71: I needed this logic for determining what database was being used down below for IDistCache registration, so moved this logic into a method that I can reused.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -716,7 +695,7 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
         }
         else
         {
-            services.AddKeyedSingleton<IDistributedCache, MemoryDistributedCache>("persistent");
+            services.AddKeyedSingleton("persistent", (s, _) => s.GetRequiredService<IDistributedCache>());
```
kspearrin, line 698: Use whatever was registered for the non-keyed service instead of registering a new instance.

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
```diff
@@ -0,0 +1,304 @@
...
+        {
+            return pgEx.SqlState == "23505";
+        }
+        // Sqlite
+        else if (e.InnerException is Microsoft.Data.Sqlite.SqliteException liteEx)
+        {
+            return liteEx.SqliteErrorCode == 19 && liteEx.SqliteExtendedErrorCode == 1555;
+        }
+        return false;
+    }
```
justindbaur, line 303: Oh boy 😆, not saying you can do it in a better way but just oh boy...

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
```diff
@@ -0,0 +1,304 @@
...
+
+    private void ScanForExpiredItemsIfRequired()
+    {
+        lock (_mutex)
+        {
+            var utcNow = DateTime.UtcNow;
+            if ((utcNow - _lastExpirationScan) > _expiredItemsDeletionInterval)
+            {
+                _lastExpirationScan = utcNow;
+                Task.Run(_deleteExpiredCachedItemsDelegate);
```
justindbaur, line 261: This concerns me a bit, I know this is how the SqlServer one does it but a DbContext is only supposed to be accessed from a single thread at a time and it has safety mechanisms to detect that. Read the bullet point three in [this section](https://learn.microsoft.com/en-us/ef/core/performance/advanced-performance-topics?tabs=with-di%2Cexpression-api-with-constant#reducing-runtime-overhead). We might be okay because it's a different instance from a different scope but I'd really love some tests making sure this works.

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
```diff
@@ -28,6 +28,7 @@ public DatabaseContext(DbContextOptions<DatabaseContext> options)
     public DbSet<UserServiceAccountAccessPolicy> UserServiceAccountAccessPolicy { get; set; }
     public DbSet<GroupServiceAccountAccessPolicy> GroupServiceAccountAccessPolicy { get; set; }
     public DbSet<ApiKey> ApiKeys { get; set; }
+    public DbSet<Cache> Cache { get; set; }
```
justindbaur, line 31: Most of these use the plural name of the item for the property (other than the SM ones) but I think `Cache` _is_ the plural, or at least `Caches` sound weird to me.

justindbaur (changes requested):

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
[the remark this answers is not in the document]
kspearrin: Whats the different in TimeProvider and ISystemClock? I noticed the SQL Server cache implementation used ISystemClock for this reason.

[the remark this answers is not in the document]
kspearrin: Looks like TimeProvider is new and recommended over ISystemClock https://github.com/aspnet/Announcements/issues/505

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
kspearrin: Cache seems appropriate to me still

rkac-bw (dismissed):
lgtm

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
kspearrin: There's a whole bunch of tests from MS on this here: https://github.com/dotnet/aspnetcore/tree/main/src/Caching/SqlServer/test

Do you see a way we can bring these over for our implementation?

withinfocus:
@justindbaur @rkac-bw this is ready for review again!

rkac-bw (dismissed):
lgtm

justindbaur (approved):
I was slightly concerned about the expired items scanning, so I added a few tests. None of the tests show any issues though. So looks good to me. I will annotate where I made a couple minor tweaks to the code to facilitate the tests.

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
```diff
@@ -0,0 +1,315 @@
...
+
+public class EntityFrameworkCache : IDistributedCache
+{
+#if DEBUG
+    // Used for debugging in tests
+    public Task scanTask;
+#endif
+    private static readonly TimeSpan _defaultSlidingExpiration = TimeSpan.FromMinutes(20);
+    private static readonly TimeSpan _expiredItemsDeletionInterval = TimeSpan.FromMinutes(30);
+    private DateTimeOffset _lastExpirationScan;
```
justindbaur, line 17: I removed the `static` off this, this matches what the `SqlServerCache` implementation does. 

https://github.com/dotnet/aspnetcore/blob/091da452aab1838fd30f6de1decd921326736146/src/Caching/SqlServer/src/SqlServerCache.cs#L27

src/Infrastructure.EntityFramework/EntityFrameworkCache.cs
```diff
@@ -0,0 +1,315 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Bit.Infrastructure.EntityFramework;
+
+public class EntityFrameworkCache : IDistributedCache
+{
+#if DEBUG
+    // Used for debugging in tests
+    public Task scanTask;
+#endif
```
justindbaur, line 14: Made it so the scanTask is available in tests so we can actually make sure that it doesn't end in a fault state. Since the Task is largely fired off and not handled it could error and we never get it reported. This could still happen in production code though.

withinfocus:
@justindbaur awesome tests! @rkac-bw can you give a final review stamp?

rkac-bw (approved):
lgtm

PR #3793, opened 2024-02-13
Vault/ac 2084/include permissions for admin endpoints

shane-melton:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
We’ve run into multiple places where we need the current user’s collection permissions after fetching them from the admin endpoints. Those responses do not include the manage, readonly, hidePasswords flags so we’ve been forced to merge the permissions from the collection sync data with the admin endpoint response multiple times. 

This PR adds those permission details to those admin endpoint responses. Doing so required creating new SQL queries to fetch collections and permission information even if there is no relationship (all existing queries that included permission info automatically filter out collections that have no existing User/Group relationship).

#### New `CollectionAdminDetails` Model
Add a new Domain model that represents collection details that are intended to be used for collection management/administrative purposes. The model includes permission details for a given user, a flag for if that user is explicitly assigned to the collection, and optional lists for User/Group access relationships.

#### `ICollectionRepository.cs`
Add additional documentation to existing methods and add the two new `*WithPermissions` queries. By their inherent nature, these queries do not perform our normal database level authorization checks, so care should be taken to ensure the requesting user has access to the organization. (Our existing queries quietly remove any collection resources the user does not have access to.)
- `GetManyByOrganizationIdWithPermissionsAsync()`
  - Fetches all of collections that belong to an organization, with permission details for a provided userId. Can also optionally include User/Group access relationships.
- `GetByIdWithPermissionsAsync()`
  - Fetches a single collection by Id with permission details for the provided userId. Can also optionally include User/Group access relationships.

## Before you submit

### DIFF
--- a/src/Api/Controllers/CollectionsController.cs
+++ b/src/Api/Controllers/CollectionsController.cs
@@ -533,47 +533,45 @@ public class CollectionsController : Controller
     private async Task<CollectionAccessDetailsResponseModel> GetDetails_vNext(Guid id)
     {
         // New flexible collections logic
-        var (collection, access) = await _collectionRepository.GetByIdWithAccessAsync(id);
-        var authorized = (await _authorizationService.AuthorizeAsync(User, collection, BulkCollectionOperations.ReadWithAccess)).Succeeded;
+        var collectionDetails =
+            await _collectionRepository.GetByIdWithPermissionsAsync(id, _currentContext.UserId, true);
+
+        var authorized = (await _authorizationService.AuthorizeAsync(User, collectionDetails, BulkCollectionOperations.ReadWithAccess)).Succeeded;
         if (!authorized)
         {
             throw new NotFoundException();
         }
 
-        return new CollectionAccessDetailsResponseModel(collection, access.Groups, access.Users);
+        return new CollectionAccessDetailsResponseModel(collectionDetails);
     }
 
     private async Task<ListResponseModel<CollectionAccessDetailsResponseModel>> GetManyWithDetails_vNext(Guid orgId)
     {
-        // We always need to know which collections the current user is assigned to
-        var assignedOrgCollections = await _collectionRepository
-            .GetManyByUserIdWithAccessAsync(_currentContext.UserId.Value, orgId, true);
+        var allOrgCollections = await _collectionRepository.GetManyByOrganizationIdWithPermissionsAsync(
+            orgId, _currentContext.UserId.Value, true);
 
         var readAllAuthorized =
             (await _authorizationService.AuthorizeAsync(User, CollectionOperations.ReadAllWithAccess(orgId))).Succeeded;
         if (readAllAuthorized)
         {
-            // The user can view all collections, but they may not always be assigned to all of them
-            var allOrgCollections = await _collectionRepository.GetManyByOrganizationIdWithAccessAsync(orgId);
-
-            return new ListResponseModel<CollectionAccessDetailsResponseModel>(allOrgCollections.Select(c =>
-                new CollectionAccessDetailsResponseModel(c.Item1, c.Item2.Groups, c.Item2.Users)
-                {
-                    // Manually determine which collections they're assigned to
-                    Assigned = assignedOrgCollections.Any(ac => ac.Item1.Id == c.Item1.Id)
-                })
+            return new ListResponseModel<CollectionAccessDetailsResponseModel>(
+                allOrgCollections.Select(c => new CollectionAccessDetailsResponseModel(c))
             );
         }
 
-        // Filter the assigned collections to only return those where the user has Manage permission
-        var manageableOrgCollections = assignedOrgCollections.Where(c => c.Item1.Manage).ToList();
+        // Filter collections to only return those where the user has Manage permission
+        var manageableOrgCollections = allOrgCollections.Where(c => c.Manage).ToList();
+
+        var authorized = (await _authorizationService.AuthorizeAsync(User, manageableOrgCollections, BulkCollectionOperations.ReadWithAccess)).Succeeded;
+
+        if (!authorized)
+        {
+            throw new NotFoundException();
+        }
 
         return new ListResponseModel<CollectionAccessDetailsResponseModel>(manageableOrgCollections.Select(c =>
-            new CollectionAccessDetailsResponseModel(c.Item1, c.Item2.Groups, c.Item2.Users)
-            {
-                Assigned = true // Mapping from manageableOrgCollections implies they're all assigned
-            })
-        );
+            new CollectionAccessDetailsResponseModel(c)
+        ));
     }
 
     private async Task<ListResponseModel<CollectionResponseModel>> GetByOrgId_vNext(Guid orgId)
--- a/src/Api/Models/Response/CollectionResponseModel.cs
+++ b/src/Api/Models/Response/CollectionResponseModel.cs
@@ -50,6 +50,17 @@ public class CollectionAccessDetailsResponseModel : CollectionResponseModel
         Users = users.Select(g => new SelectionReadOnlyResponseModel(g));
     }
 
+    public CollectionAccessDetailsResponseModel(CollectionAdminDetails collection)
+        : base(collection, "collectionAccessDetails")
+    {
+        Assigned = collection.Assigned;
+        ReadOnly = collection.ReadOnly;
+        HidePasswords = collection.HidePasswords;
+        Manage = collection.Manage;
+        Groups = collection.Groups?.Select(g => new SelectionReadOnlyResponseModel(g)) ?? Enumerable.Empty<SelectionReadOnlyResponseModel>();
+        Users = collection.Users?.Select(g => new SelectionReadOnlyResponseModel(g)) ?? Enumerable.Empty<SelectionReadOnlyResponseModel>();
+    }
+
     public IEnumerable<SelectionReadOnlyResponseModel> Groups { get; set; }
     public IEnumerable<SelectionReadOnlyResponseModel> Users { get; set; }
 
@@ -57,4 +68,8 @@ public class CollectionAccessDetailsResponseModel : CollectionResponseModel
     /// True if the acting user is explicitly assigned to the collection
     /// </summary>
     public bool Assigned { get; set; }
+
+    public bool ReadOnly { get; set; }
+    public bool HidePasswords { get; set; }
+    public bool Manage { get; set; }
 }
--- a/src/Core/Models/Data/CollectionAdminDetails.cs
+++ b/src/Core/Models/Data/CollectionAdminDetails.cs
@@ -1,0 +1,17 @@
+﻿#nullable enable
+namespace Bit.Core.Models.Data;
+
+/// <summary>
+/// Collection information that includes permission details for a particular user along with optional
+/// access relationships for Groups/Users. Used for collection management.
+/// </summary>
+public class CollectionAdminDetails : CollectionDetails
+{
+    public IEnumerable<CollectionAccessSelection>? Groups { get; set; } = new List<CollectionAccessSelection>();
+    public IEnumerable<CollectionAccessSelection>? Users { get; set; } = new List<CollectionAccessSelection>();
+
+    /// <summary>
+    /// Flag for whether the user has been explicitly assigned to the collection either directly or through a group.
+    /// </summary>
+    public bool Assigned { get; set; }
+}
--- a/src/Core/Repositories/ICollectionRepository.cs
+++ b/src/Core/Repositories/ICollectionRepository.cs
@@ -6,13 +6,48 @@ namespace Bit.Core.Repositories;
 public interface ICollectionRepository : IRepository<Collection, Guid>
 {
     Task<int> GetCountByOrganizationIdAsync(Guid organizationId);
+
+    /// <summary>
+    /// Returns a collection and fetches group/user associations for the collection.
+    /// </summary>
     Task<Tuple<Collection, CollectionAccessDetails>> GetByIdWithAccessAsync(Guid id);
+
+    /// <summary>
+    /// Returns a collection with permission details for the provided userId and fetches group/user associations for
+    /// the collection.
+    /// If the user does not have a relationship with the collection, nothing is returned.
+    /// </summary>
     Task<Tuple<CollectionDetails, CollectionAccessDetails>> GetByIdWithAccessAsync(Guid id, Guid userId, bool useFlexibleCollections);
+
+    /// <summary>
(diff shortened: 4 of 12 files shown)


src/Api/Controllers/CollectionsController.cs
```diff
@@ -533,47 +533,45 @@ private async Task<CollectionResponseModel> Get_vNext(Guid collectionId)
...
+                allOrgCollections.Select(c => new CollectionAccessDetailsResponseModel(c))
             );
         }
 
-        // Filter the assigned collections to only return those where the user has Manage permission
-        var manageableOrgCollections = assignedOrgCollections.Where(c => c.Item1.Manage).ToList();
+        // Filter collections to only return those where the user has Manage permission
+        var manageableOrgCollections = allOrgCollections.Where(c => c.Manage).ToList();
+
+        var authorized = (await _authorizationService.AuthorizeAsync(User, manageableOrgCollections, BulkCollectionOperations.ReadWithAccess)).Succeeded;
```
shane-melton, line 565: question for reviewers: This may not be actually be necessary? We already implicitly know the user has `Manage` permission in the `Where()` statement above, so we can likely return the `manageableOrgCollections` without the extra authorization check.

eliykat (commented):
Started looking at this today just to get my head around it. Will continue tomorrow.

src/Sql/Vault/dbo/Stored Procedures/Collections/Collection_ReadByIdWithPermissions.sql
```diff
@@ -0,0 +1,54 @@
...
+	LEFT JOIN
+	    [dbo].[CollectionGroup] CG ON CG.[CollectionId] = C.[Id] AND CG.[GroupId] = GU.[GroupId]
+	WHERE
+	    C.[Id] = @CollectionId
+
+   IF (@IncludeAccessRelationships = 1)
+    BEGIN
+        EXEC [dbo].[CollectionGroup_ReadByCollectionId] @CollectionId
+        EXEC [dbo].[CollectionUser_ReadByCollectionId] @CollectionId
+	END
```
eliykat, line 53: I'm curious to hear @rkac-bw's opinion on whether it's better to call these queries from within this sproc, or as entirely separate db calls from the repository code (like we do [here](https://github.com/bitwarden/server/blob/cd2e3e0df9d0e2ebeaf026b9229213b1833b2a5f/src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationUserRepository.cs#L239-L254) for example).

rkac-bw: This is fine in my opinion, these will both generate their own plan and be optimized in their own execution be it from client or stored procedure.

src/Core/Models/Data/CollectionAdminDetails.cs
```diff
@@ -0,0 +1,17 @@
+﻿#nullable enable
+namespace Bit.Core.Models.Data;
+
+/// <summary>
+/// Collection information that includes permission details for a particular user along with optional
+/// access relationships for Groups/Users. Used for collection management.
+/// </summary>
+public class CollectionAdminDetails : CollectionDetails
```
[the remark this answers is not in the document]
eliykat: I think we discussed this at the time and decided the current name was fine.

src/Api/Models/Response/CollectionResponseModel.cs
```diff
@@ -50,11 +105,26 @@ public CollectionAccessDetailsResponseModel(Collection collection, IEnumerable<C
...
     public IEnumerable<SelectionReadOnlyResponseModel> Users { get; set; }
 
     /// <summary>
     /// True if the acting user is explicitly assigned to the collection
     /// </summary>
     public bool Assigned { get; set; }
+
+    public bool ReadOnly { get; set; }
+    public bool HidePasswords { get; set; }
+    public bool Manage { get; set; }
```
eliykat, line 129: Could this extend the `CollectionAccessDetailsResponseModel`? It's very similar, and the idea seems to be "CollectionDetails but with Access", which suggests a subclass.

src/Api/Models/Response/CollectionResponseModel.cs
```diff
@@ -26,19 +26,74 @@ public CollectionResponseModel(Collection collection, string obj = "collection")
...
     }
 
     public bool ReadOnly { get; set; }
     public bool HidePasswords { get; set; }
     public bool Manage { get; set; }
+
+    /// <summary>
+    /// Flag for whether the user has been explicitly assigned to the collection either directly or through a group.
+    /// </summary>
+    public bool Assigned { get; set; }
```
eliykat, line 79: I think the `Assigned` property is starting to collapse the different classes here. So far you've kept the old & new uses separate: `CollectionDetails` is for user vault data (always assigned) and `CollectionAdminDetails` is for admin actions (maybe assigned). I expected to see a flow like:

* CollectionDetails -> CollectionDetailsResponseModel
* CollectionAdminDetails -> CollectionAdminDetailsResponseModel (which includes optional group/member access details)

but where it would not be possible to do, e.g. CollectionDetails -> CollectionAdminDetailsResponseModel.

The benefit of this is that the type system enforces the separate uses, rather than being careful about calling `FromAssigned` or `FromUnassigned`. And if you're using `FromUnassigned`, then you have a response class where 4 properties are effectively unused (`assigned` bool + the 3 permissions), which suggests a different use of classes & subclasses.

That said, I know this has been tricky to integrate with existing code so this may be a compromise or there may be other uses I'm missing.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -555,47 +555,45 @@ private async Task<CollectionResponseModel> Get_vNext(Guid collectionId)
     private async Task<CollectionAccessDetailsResponseModel> GetDetails_vNext(Guid id)
     {
         // New flexible collections logic
-        var (collection, access) = await _collectionRepository.GetByIdWithAccessAsync(id);
-        var authorized = (await _authorizationService.AuthorizeAsync(User, collection, BulkCollectionOperations.ReadWithAccess)).Succeeded;
+        var collectionDetails =
```
eliykat, line 558: The different models are a little hard to keep track of, so I suggest being verbose with the names - `collectionAdminDetails` here.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -555,47 +555,45 @@ private async Task<CollectionResponseModel> Get_vNext(Guid collectionId)
...
-                })
+            return new ListResponseModel<CollectionAccessDetailsResponseModel>(
+                allOrgCollections.Select(c => new CollectionAccessDetailsResponseModel(c))
             );
         }
 
-        // Filter the assigned collections to only return those where the user has Manage permission
-        var manageableOrgCollections = assignedOrgCollections.Where(c => c.Item1.Manage).ToList();
+        // Filter collections to only return those where the user has Manage permission
+        var manageableOrgCollections = allOrgCollections.Where(c => c.Manage).ToList();
```
eliykat, line 585: I was going to suggest this, but I guess if they're not assigned then Manage will be false anyway?

```suggestion
        var manageableOrgCollections = allOrgCollections.Where(c => c.Assigned && c.Manage).ToList();
```

src/Api/Controllers/CollectionsController.cs
eliykat: I agree this isn't necessary, which is why we didn't do it on the previous implementation. It breaks the abstraction in a way that I don't like but it is what it is for now.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -648,19 +646,16 @@ private async Task<CollectionResponseModel> Post_vNext(Guid orgId, [FromBody] Co
 
         if (!_currentContext.UserId.HasValue || await _currentContext.ProviderUserForOrgAsync(orgId))
         {
-            return new CollectionResponseModel(collection);
+            return CollectionDetailsResponseModel.FromUnassigned(collection);
         }
```
eliykat, line 650: I'm not sure if this is a bug - but you can be a providerUser for an org AND a member. It should probably check that they're not a member and are a provider.

eliykat (changes requested):
I've reviewed the SQL sprocs, models, and controller methods to make sure they're in line with our discussions and AC Team's requirements, but I'll leave the rest (mostly EF queries and test logic I think) to the Vault reviewer as codeowner.

Thank you again for tackling this!

src/Api/Controllers/CollectionsController.cs
eliykat, line 1: There are build errors in this file: https://github.com/bitwarden/server/actions/runs/8274332589/job/22639547174?pr=3793

src/Api/Controllers/CollectionsController.cs
shane-melton: Good catch, fixed in https://github.com/bitwarden/server/pull/3793/commits/1a4047b8db2f1a5db2ed40754c500560dc053c45

src/Api/Controllers/CollectionsController.cs
shane-melton: Fixed in [859ab5c](https://github.com/bitwarden/server/pull/3793/commits/859ab5cb97f7a5f11090d11a7edfc2b18102f93d)

src/Api/Models/Response/CollectionResponseModel.cs
shane-melton: Very good points, thanks for bringing this up. Approaching it this way allowed me to clean up the response models quite a bit and make it easier to understand. 

Removed the `FromAssigned` and `FromUnassigned` static constructors and the `Assigned` property from `CollectionDetails` (as it is already implied as assigned). Commit [859ab5c](https://github.com/bitwarden/server/pull/3793/commits/859ab5cb97f7a5f11090d11a7edfc2b18102f93d)

src/Api/Controllers/CollectionsController.cs
shane-melton: Good call, fixed in [2f18bc8](https://github.com/bitwarden/server/pull/3793/commits/2f18bc8c0bd1e07cb8fbc30dcda537381996e351)

src/Api/Controllers/CollectionsController.cs
shane-melton: Yep, `Managed` will never be true if `!Assigned`

src/Api/Controllers/CollectionsController.cs
shane-melton: Removed redundant check in [16bb5f0](https://github.com/bitwarden/server/pull/3793/commits/16bb5f06b1d1f173c45dbc8ca5f7b851e10835cb)

src/Core/Repositories/ICollectionRepository.cs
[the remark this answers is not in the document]
shane-melton: Agreed!

src/Api/Models/Response/CollectionResponseModel.cs
eliykat: Looks much better, nice work.

eliykat (dismissed):
lgtm!

Jingo88 (approved):

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -427,7 +484,64 @@ public async Task<ICollection<CollectionDetails>> GetManyByUserIdAsync(Guid user
...
+                        c.ExternalId
+                    }).Select(collectionGroup => new CollectionAdminDetails
+                    {
+                        Id = collectionGroup.Key.Id,
+                        OrganizationId = collectionGroup.Key.OrganizationId,
+                        Name = collectionGroup.Key.Name,
+                        CreationDate = collectionGroup.Key.CreationDate,
+                        RevisionDate = collectionGroup.Key.RevisionDate,
+                        ExternalId = collectionGroup.Key.ExternalId,
+                        ReadOnly = Convert.ToBoolean(collectionGroup.Min(c => Convert.ToInt32(c.ReadOnly))),
```
eliykat, line 510: The conversions are unnecessary:

* `collectionGroup.All(c => c.ReadOnly)` (equivalent of MIN, it will return false if there is any `ReadOnly = false`)
* `collectionGroup.Any(c => c.Manage)` (equivalent of MAX, it will return true if there is any `Manage = true`)

However I can see this is how we do it in existing queries, and that does more closely track the SQL equivalent, so I'll leave it up to you as to which you prefer.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -427,7 +484,64 @@ public async Task<ICollection<CollectionDetails>> GetManyByUserIdAsync(Guid user
         {
             var dbContext = GetDatabaseContext(scope);
             var query = CollectionAdminDetailsQuery.ByCollectionId(collectionId, userId).Run(dbContext);
-            var collectionDetails = await query.FirstOrDefaultAsync();
+
+            CollectionAdminDetails collectionDetails;
+
+            // SQLite does not support the GROUP BY clause
```
eliykat, line 490: Boo :( If this is the case, should we just have the single implementation? I guess the Sqlite version here is arguably less performant because it doesn't use delayed execution. But in practice I'm not sure it would make a significant difference, and it avoids having divergent EF implementations for different databases. Would be interested in what @justindbaur thinks.

eliykat (approved):
Love the tests, a couple of comments below but nothing blocking.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
shane-melton: I'm not too attached either way (a single implementation or split). We do have a few other repository calls here that have similar behavior, so its not entirely new, if still pretty ugly. 

I'm also not too familiar with the magnitude of the performance difference, but I do expect this to be called fairly frequently in the AC until we have better caching there, so its something to keep in mind.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
shane-melton: That's a fair point. I had gone with what already existed in the other queries, as I figured there was some underlying implementation detail forcing the extra conversions (e.g. maybe one provider doesn't behave properly with `Any` or `All`).

I think I'll opt to leave it alone for now since we know it works as expected and it keeps consistent with the other queries.

eliykat (approved):

rkac-bw (approved):

Jingo88 (approved):

PR #3795, opened 2024-02-13
[DEVOPS-1218] Add dryrun mode to MsSqlMigratorUtility

michalchecinski:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add dryrun mode to MsSqlMigratorUtility. It's needed to perform migration introduction database: https://bitwarden.atlassian.net/browse/DEVOPS-1218

## Code changes

* **util/MsSqlMigratorUtility/Program.cs** Add dryrun flag to MsSqlMigratorUtility
* **util/Migrator/DbMigrator.cs:** `Add dry run parameter to the DbMigrator which will only list SQL scripts that will be ran.

## Before you submit


rkac-bw (approved):
Looks good to me

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -20,31 +20,35 @@ private static int Main(string[] args)
         [Option('r', "repeatable", Description = "Mark scripts as repeatable")]
         bool repeatable = false,
         [Option('f', "folder", Description = "Folder name of database scripts")]
-        string folderName = MigratorConstants.DefaultMigrationsFolderName) => MigrateDatabase(databaseConnectionString, verbose, repeatable, folderName);
+        string folderName = MigratorConstants.DefaultMigrationsFolderName,
+        [Option("dryrun", Description = "Dry run migrations. It'll print the scripts that will be run without actually running them")]
```
withinfocus, line 24:
```suggestion
        [Option("dry-run", Description = "Print the scripts that will be applied without actually executing them")]
```

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -20,31 +20,35 @@ private static int Main(string[] args)
...
+        bool dryRun = false
+        ) => MigrateDatabase(databaseConnectionString, verbose, repeatable, folderName);
```
withinfocus, line 26: ❌ You need to pass the parameter in, right?

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -20,31 +20,35 @@ private static int Main(string[] args)
...
 
     private static void WriteUsageToConsole()
     {
         Console.WriteLine("Usage: MsSqlMigratorUtility <database-connection-string>");
         Console.WriteLine("Usage: MsSqlMigratorUtility <database-connection-string> -v|--verbose (for verbose output of migrator logs)");
+        Console.WriteLine("Usage: MsSqlMigratorUtility <database-connection-string> --dryrun (for dry run migrations. It'll print the scripts that will be run without actually running them)");
```
withinfocus, line 32:
```suggestion
        Console.WriteLine("Usage: MsSqlMigratorUtility <database-connection-string> --dry-run (print the scripts that will be applied without actually executing them)");
```

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -20,31 +20,35 @@ private static int Main(string[] args)
...
         Console.WriteLine("Usage: MsSqlMigratorUtility <database-connection-string> -v|--verbose (for verbose output of migrator logs) -r|--repeatable (for marking scripts as repeatable) -f|--folder <folder-name-in-migrator-project> (for specifying folder name of scripts)");
     }
 
-    private static bool MigrateDatabase(string databaseConnectionString, bool verbose = false, bool repeatable = false, string folderName = "")
+    private static bool MigrateDatabase(string databaseConnectionString, bool verbose = false, bool repeatable = false, string folderName = "", bool dryRun = false)
     {
         var logger = CreateLogger(verbose);
 
-        logger.LogInformation($"Migrating database with repeatable: {repeatable} and folderName: {folderName}.");
+        logger.LogInformation($"Migrating database with repeatable: {repeatable}, folderName: {folderName} and dry-run: {dryRun}.");
```
withinfocus, line 41:
```suggestion
        logger.LogInformation($"Migrating database with repeatable: {repeatable}, folderName: {folderName}, dry-run: {dryRun}.");
```

util/Setup/Program.cs
```diff
@@ -195,7 +195,7 @@ private static void MigrateDatabase(int attempt = 1)
 
         migrator.MigrateMsSqlDatabaseWithRetries(log);
 
-        migrator.MigrateMsSqlDatabaseWithRetries(log, true, MigratorConstants.TransitionMigrationsFolderName);
+        migrator.MigrateMsSqlDatabaseWithRetries(log, true, false, MigratorConstants.TransitionMigrationsFolderName);
```
withinfocus, line 198: ❌ Unnecessary to provide this as it's optional.

util/Migrator/DbMigrator.cs
```diff
@@ -130,6 +132,17 @@ public DbMigrator(string connectionString, ILogger<DbMigrator> logger)
...
 
         var upgrader = builder.Build();
+
+        if (dryRun)
+        {
+            var scriptsToExec = upgrader.GetScriptsToExecute();
+            foreach (var script in scriptsToExec)
+            {
+                _logger.LogInformation(Constants.BypassFiltersEventId, script.Name);
+            }
```
withinfocus, line 142: 🎨 Let's not use a log statement for every script. Instead, create a `StringBuilder` earlier and append to it here, while also including `Environment.Newline` at the end. This will render much better.

withinfocus (changes requested):

util/Migrator/DbMigrator.cs
michalchecinski: I used string builder here.

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -20,31 +20,35 @@ private static int Main(string[] args)
...
         Console.WriteLine("Usage: MsSqlMigratorUtility <database-connection-string> -v|--verbose (for verbose output of migrator logs) -r|--repeatable (for marking scripts as repeatable) -f|--folder <folder-name-in-migrator-project> (for specifying folder name of scripts)");
     }
 
-    private static bool MigrateDatabase(string databaseConnectionString, bool verbose = false, bool repeatable = false, string folderName = "")
+    private static bool MigrateDatabase(string databaseConnectionString, bool verbose = false, bool repeatable = false, string folderName = "", bool dryRun = false)
     {
         var logger = CreateLogger(verbose);
 
-        logger.LogInformation($"Migrating database with repeatable: {repeatable} and folderName: {folderName}.");
+        logger.LogInformation($"Migrating database with repeatable: {repeatable}, folderName: {folderName}, dry-run: {dryRun}.");
```
withinfocus, line 41:
```suggestion
        logger.LogInformation($"Migrating database with repeatable: {repeatable}, folderName: {folderName}, dryRun: {dryRun}.");
```

util/Setup/Program.cs
withinfocus: I would instead write `folderName: MigratorConstants.TransitionMigrationsFolderName`. It's still better to avoid passing the default. This is often why it's best to add new optional parameters _only_ to the end of method signatures.

util/Migrator/DbMigrator.cs
```diff
@@ -130,6 +133,19 @@ public DbMigrator(string connectionString, ILogger<DbMigrator> logger)
         }
 
         var upgrader = builder.Build();
+
+        if (dryRun)
+        {
+            var scriptsToExec = upgrader.GetScriptsToExecute();
+            var stringBuilder = new StringBuilder();
```
withinfocus, line 140: 🎨 Before your loop put a line in here with "Scripts that will be applied:" so it's clear what it's outputting.

withinfocus (changes requested):
Small wins, but really close.

withinfocus (commented):
Given https://github.com/bitwarden/server/pull/3910 and its priority for client needs I still want to do this too but would like for it to wait until that's done given the decent refactor there.

util/Migrator/DbMigrator.cs
michalchecinski: Done.

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -15,21 +15,23 @@ private static int Main(string[] args)
         [Option('r', "repeatable", Description = "Mark scripts as repeatable")]
         bool repeatable = false,
         [Option('f', "folder", Description = "Folder name of database scripts")]
-        string folderName = MigratorConstants.DefaultMigrationsFolderName)
-        => MigrateDatabase(databaseConnectionString, repeatable, folderName);
+        string folderName = MigratorConstants.DefaultMigrationsFolderName,
+        [Option("dry-run", Description = "Print the scripts that will be applied without actually executing them")]
```
withinfocus, line 19: 💭 We use shorthand for the others ... would `s` work? I see in the APT docs at least that "s" means "simulate" and then has an alias `dry-run`. Just trying to be consistent here. We can call it `dryRun` everywhere lower down.

michalchecinski: Maybe the `d` alias will suite here beter?

withinfocus (approved):
🤷 on the exact shorthand to use, but we can do what we want.

PR #3799, opened 2024-02-14
[AC-2206] Fix assigning Manage access to default collection

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Fix #3776 which did not account for `orgUser` being `null` in the private `OrganizationService.SignUp` method. This occurs when a provider is creating an organization, the org is created without an owner and then an invite is sent out separately (see `ProviderService.CreateOrganizationAsync`).

## Code changes

* return the `defaultCollection` from the private `OrganizationService.SignUp` method
* to make this less ugly, change to using named value tuples in these return signatures
* if `orgUser` is null, skip assigning access to the default collection
* in `ProviderService.CreateOrganizationAsync`, take the returned collection and give the owner Manage access when inviting them

I don't like how this is handled in multiple places, but I am avoiding any larger refactors until we split up OrganizationService into separate classes (which we have tech debt tickets for).

## Before you submit

### DIFF
--- a/bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
+++ b/bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
@@ -496,7 +496,7 @@ public class ProviderService : IProviderService
     {
         ThrowOnInvalidPlanType(organizationSignup.Plan);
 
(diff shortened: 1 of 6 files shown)


r-tome (approved):

vincentsalucci:
Merging on behalf of @eliykat

PR #3800, opened 2024-02-14
[AC-2212] Fix Flexible Collections block in Public API

eliykat:
Only throw if collection.Manage is true

## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The Public API Members controller was throwing if any value was provided for `manage` without Flexible Collections being enabled. However, this is a problem for scripts that GET a member (which includes `manage: false` for any collections) and then use that result in their PUT request. Really we should only block the action if `manage` is set to `true`.

## Code changes

## Before you submit


vincentsalucci (approved):

vincentsalucci:
Merging on behalf of @eliykat

PR #3805, opened 2024-02-14
[AC-2214] Defect - provider reseller org creation when fc signup flag enabled

vincentsalucci:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Code changes

* **src/Admin/Controllers/ProvidersController.cs**: Injected `featureService` and passed along the value of the `flexible-collection-signup` flag to the organization model creation
* **src/Admin/Models/OrganizationEditModel.cs**: Assigned the fc signup enabled state to the `Organization.FlexibleCollections` property that will be used to create the db entity for the pending org.

## Before you submit

### DIFF
--- a/src/Admin/Controllers/ProvidersController.cs
+++ b/src/Admin/Controllers/ProvidersController.cs
@@ -1,6 +1,7 @@
 ﻿using Bit.Admin.Enums;
 using Bit.Admin.Models;
 using Bit.Admin.Utilities;
+using Bit.Core;
 using Bit.Core.AdminConsole.Entities.Provider;
 using Bit.Core.AdminConsole.Enums.Provider;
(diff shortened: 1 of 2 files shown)


r-tome (dismissed):
Looks good to me!

eliykat (changes requested):
Generally looks good, but just to repost our Slack discussions here for visibility:
1. we also need to set the default Collection Management Settings - refer to OrganizationService.SignUp (the public methods) for the logic
2. we need to consider any orgs created via this method that didn't have these settings set properly - they should default to true until we turn flags on, but because this code path wasn't covered these orgs would have defaulted to false (enabling new functionality)

eliykat (approved):

PR #3814, opened 2024-02-16
[PM-6343] Fix 415 response for cancel subscription

amorask-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

In a previous update, I had tried to combine both a FF on case and FF off case in one endpoint. However, one case required a request body and one did not. This resulted in the case where the FF was off responding with a 415 Unsupported Media type. 

This PR resolves the issue by simply splitting the endpoint in two; one for the case where the FF is on and one for the case where the FF is off. The web client uses the FF to determine which endpoint to invoke. 

## Code changes

## Before you submit

### DIFF
--- a/src/Api/AdminConsole/Controllers/OrganizationsController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationsController.cs
@@ -464,50 +464,52 @@ public class OrganizationsController : Controller
         await _organizationService.VerifyBankAsync(orgIdGuid, model.Amount1.Value, model.Amount2.Value);
     }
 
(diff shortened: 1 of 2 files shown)


jlf0dev (approved):
Just want to note we should likely move these to a billing controller at some point 👍

amorask-bitwarden:
@jlf0dev Agreed - we'd definitely like to. It's just hard to delineate what exactly should belong to Auth or AC vs. Billing because the endpoints are mixed responsibility.

r-tome (approved):

PR #3829, opened 2024-02-20
Include all projects in coverage

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Not all of our server projects had associated test projects which caused them to be omitted from the code coverage. Added projects to ensure the coverage gets reported accurately.

## Before you submit


test/Admin.Test/UnitTest1.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Admin.Test;
```
withinfocus, line 1: 💭 What can we put here that prompts inquiry for adding tests? I am thinking we rename the file, class, and initial method to something implying "placeholder" or "temporary", perhaps even with a build warning.

Hinton: I can rename the tests to `PlaceholderUnitTest` if we want.

PR #3832, opened 2024-02-20
Consolidated scanning

withinfocus:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Consolidates scanning for several tools into one unified workflow. This performs three things:

* SAST scanning: with our Checkmarx tool of choice based on a past assessment. This runs PR scans in an incremental mode and finishes within 2-3 minutes vs. some very long run times of 45+ minutes for a full scan; full scans will still occur on push events. Since we now drive the scanning, after this merges we will turn off webhook-triggered scans for every repo using this.
* Quality scanning: with our SonarCloud tool of choice based on a past assessment. This tool will also have "automatic analysis" disabled since we run it in CI now. CodeScene will be disabled after this merges.

These two tools ship in one workflow file, and this will likely move to our template repository. Configuration will be minimal to nothing for other repositories.

The third:

* Container scanning: with Grype. Popular tool that does well vs. others and is free. This needs to be close to container generation and given how intertwined our Docker builds are I left it nearby. This allows the above scanner workflow to ship independently too, and we'll selectively add container scanning where it makes sense.

SAST and container scan results are exported as SARIF and uploaded to the GitHub Advanced Security interface for internal review.

## Code changes

* **.github/workflows/build.yml:** Container scanning additions and some minor cleanup.
* **.github/workflows/scan.yml:** New workflow for SAST and quality scans.

## Before you submit


vgrassia (approved):

.github/workflows/scan.yml
MGibson1, line 1: I don't like that this takes us further into the situation where arbitrary PRs require privileged secrets access.

It's a serious blocker to out goals of faster development times and increased community involvement.

.github/workflows/scan.yml
```diff
@@ -0,0 +1,60 @@
...
+      - "rc"
+      - "hotfix-rc"
+  pull_request:
+
+permissions: read-all
+
+jobs:
+  sast:
+    name: SAST scan
+    runs-on: ubuntu-22.04
```
Hinton, line 17: Can we add a condition to exclude forks?

```yml
if: "! github.event.pull_request.head.repo.fork"
```

.github/workflows/scan.yml
```diff
@@ -0,0 +1,60 @@
+name: Scan
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+      - "rc"
+      - "hotfix-rc"
```
Hinton, line 9: Are we not running full checks on `main`?

Hinton (commented):
I would prefer we stick with only checkmarx until all the teams are onboarded.

Once that's done we can start evaluating code quality tools like sonar but we need to have the capacity to triage the existing backlog before those tools provide any value. Prior to that it's just noise and people start to quickly ignore them.

.github/workflows/scan.yml
withinfocus: We'll eventually want this running in all cases. See chat but we can at some point in time look at a total shift here to support community contributions. I think this has to be status quo for the moment.

.github/workflows/scan.yml
withinfocus: See https://github.com/bitwarden/server/pull/3832/files#diff-246cd0c2f7db532638dd80a92ac011f49b3d26038983a4c0169ea8f8a5c39280R28 -- this will run full checks outside PRs.

withinfocus:
> I would prefer we stick with only checkmarx until all the teams are onboarded.
> 
> Once that's done we can start evaluating code quality tools like sonar but we need to have the capacity to triage the existing backlog before those tools provide any value. Prior to that it's just noise and people start to quickly ignore them.

At the same time I want to build up real findings so we can adjust direction if needed. We've tested these tools in the past and reports are light, but I still want to see it really working. I do agree that we want to avoid noise, and given that we'll disable CodeScene (that is that to me) this could be a net win immediately.

PR #3843, opened 2024-02-26
[BEEEP] begin 2fa integration tests for identity

jlf0dev:
- fix org mappings and query

## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Begins some integration testing for 2fa in our token endpoint so we can pull that logic out.

Ran into some EF issues in this with the `GetManyByUserIdAsync()` that were resolved by getting rid of an inner IEnumerable created by `Select`. I changed that to `SelectMany` and added an `Include` to [eagerly load](https://learn.microsoft.com/en-us/ef/core/querying/related-data/eager) the relationships in order to access the `Organization` on the `OrganizationUser` returned.

While troubleshooting this issue, I found that AutoMapper has a test for validating mappings, `AssertConfigurationIsValid()`. I added a unit test and discovered that the additional properties added to the `Organization` model were causing it to be invalid. While this _probably_ worked regardless of the changes I made to ignore these properties, I'm leaving this in here as a proposition to add a unit test with `AssertConfigurationIsValid()` for all our EF repositories.

## Code changes

## Before you submit


jlf0dev:
@withinfocus @justindbaur tagging to see if ya'll had any thoughts on requiring `AssertConfigurationIsValid()` for validating our AutoMapper configurations. I'm not super familiar with the state of our EF repo but I know we have some unified work coming up that might benefit from it

jlf0dev:
Great! I'll leave this the way it is for now since validation of all the mappings likely requires a much greater change.

PR #3859, opened 2024-02-29
DEVOPS-1840  - Automatic Version Bump Calculation

vgrassia:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This PR updates the Version Bump workflow so that if a version override is not specified, the next version number is calculated automatically.

## Code changes

* **.github/workflows/version-bump.yml:** Change input `version_number` to `version_number_override`. Add an output to the `bump_version` job for use in the `cut_rc` job. Change branch naming since we don't have the version number to use. Add checks to see if we call the `version-bump` action with or without specifying a version. Use `set-job-output` to set the `version` output for the `bump_version` job since you cannot have two steps have identical `id` values. Formatting.

## Before you submit

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Bump version props - Version Override
         if: ${{ inputs.version_number_override != '' }}
         id: bump-version-override
-        uses: bitwarden/gh-actions/version-bump@main
+        uses: bitwarden/gh-actions/version-bump@a162eac2907814117e049408c15bfb6740522d59
         with:
           version: ${{ inputs.version_number_override }}
           file_path: "Directory.Build.props"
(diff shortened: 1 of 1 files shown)


.github/workflows/version-bump.yml
```diff
@@ -1,13 +1,12 @@
...
 
 on:
   workflow_dispatch:
     inputs:
-      version_number:
-        description: "New version (example: '2024.1.0')"
-        required: true
+      version_number_override:
+        description: "New version override (leave blank for automatic calculation, example: '2024.1.0')"
+        required: false
```
mimartin12, line 9: ⛏️  It would be nice to have some validation on this input, simply to ensure that we fail to version bump if the version number provided isn't what we expect.

vgrassia: I have an action I created for input validation [here](https://github.com/bitwarden/gh-actions/pull/262). Let me know what you think.

mimartin12 (approved):
LGTM

PR #3864, opened 2024-03-03
[AC-1682] Flexible collections: EF data migrations for deprecated permissions

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective
With Flexible Collections, the Manager role and 'EditAssignedCollections' permission are deprecated. Therefore, we must explicitly grant access to those users for the collections they previously had access to, but now with the new Manage permission. Additionally, users and groups with the AccessAll permission will now have access to all collections within their organizations.

## Code changes
This PR only contains EF data migrations, SQL migrations are [here](https://github.com/bitwarden/server/pull/3437).

* **src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs:** The EF data migration written in linq that is equivalent to the SQL data migration script
* **test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationEnableCollectionEnhancementTests.cs:** Modified the integration tests to also run for EF databases
## Before you submit


src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                    })
+                    .ToList();
+                dbContext.CollectionUsers.AddRange(newCollectionUsers);
+            }
+
+            // Update OrganizationUser to clear AccessAll flag
+            dbContext.OrganizationUsers
+                .Where(ou => orgUserIdsWithAccessAll.Contains(ou.Id))
+                .ToList()
+                .ForEach(ou => ou.AccessAll = false);
```
r-tome, line 371: Same thing here about the `RevisionDate` internal setter.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                dbContext.CollectionUsers.AddRange(newCollectionUsersWithManage);
+            }
+
+            // Update OrganizationUser to migrate Managers to User role
+            dbContext.OrganizationUsers
+                .Where(ou =>
+                    managerOrgUsersIds.Contains(ou.Id) &&
+                    ou.Type == OrganizationUserType.Manager)
+                .ToList()
+                .ForEach(ou => ou.Type = OrganizationUserType.User);
```
r-tome, line 424: `RevisionDate` setter also is internal.

util/MySqlMigrations/Factories.cs
```diff
@@ -36,7 +36,7 @@ public DatabaseContext CreateDbContext(string[] args)
             .GlobalSettings;
 
         var optionsBuilder = new DbContextOptionsBuilder<DatabaseContext>();
-        var connectionString = globalSettings.MySql?.ConnectionString;
+        var connectionString = globalSettings.MySql?.ConnectionString ?? Environment.GetEnvironmentVariable("CONN_STR");
```
r-tome, line 39: I needed to add this for the Git "Database tests" action

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                    cg.Manage = false;
+                });
+
+            // Insert new rows into CollectionGroup
+            foreach (var group in groupIdsWithAccessAll)
+            {
+                var newCollectionGroups = dbContext.Collections
+                    .Where(c =>
+                        c.OrganizationId == organizationId &&
+                        !dbContext.CollectionGroups.Any(cg => cg.CollectionId == c.Id && cg.GroupId == group))
```
eliykat, line 307: I think this could be

```suggestion
                        !c.CollectionGroups.Any(cg.GroupId == group))
```

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                        HidePasswords = false,
+                        Manage = false
+                    })
+                    .ToList();
+                dbContext.CollectionGroups.AddRange(newCollectionGroups);
+            }
+
+            // Update Group to clear AccessAll flag
+            dbContext.Groups
+                .Where(g => groupIdsWithAccessAll.Contains(g.Id))
```
eliykat, line 322: I suggest you select the groups to begin with rather than their ids, i.e. `groupsWithAccessAll` instead of `groupIdsWithAccessAll`. Then you don't need to look them up again here.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                .ToList()
+                .ForEach(g => g.AccessAll = false);
```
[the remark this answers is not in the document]
eliykat: Please check with @addisonbeck but I assume EF will update it for us.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+
+            // Save changes for Step 1
+            await dbContext.SaveChangesAsync();
```
eliykat, line 327: Please double check how this works, but I suspect that `dbContext.SaveChangesAsync` will work against your transaction commit/rollback statements. That is, it'll write your changes to the db so you can no longer roll them back - I think you can only use one or the other. Is there any need to commit our work partway through?

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                    cu.Manage = false;
+                });
+
+            // Insert new rows into CollectionUser
+            foreach (var organizationUser in orgUserIdsWithAccessAll)
+            {
+                var newCollectionUsers = dbContext.Collections
+                    .Where(c =>
+                        c.OrganizationId == organizationId &&
+                        !dbContext.CollectionUsers.Any(cu => cu.CollectionId == c.Id && cu.OrganizationUserId == organizationUser))
```
eliykat, line 354: This could also be simplified:

```suggestion
                        !c.CollectionUsers.Any(cu => cu.OrganizationUserId == organizationUser))
```

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
eliykat, line 371: Same here as with groups, just hold on to the orgUsers to begin with.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
eliykat, line 424: Same as above

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                                                              cu.CollectionId == cg.CollectionId &&
+                                                              cu.OrganizationUserId == manager)
+                                                    select new CollectionUser
+                                                    {
+                                                        CollectionId = cg.CollectionId,
+                                                        OrganizationUserId = manager,
+                                                        ReadOnly = false,
+                                                        HidePasswords = false,
+                                                        Manage = true
+                                                    }).Distinct().ToList();
```
eliykat, line 414: I wasn't sure about mixing the linq query and method syntax, but this is a bit more complex so I can understand why you've done it. Some things are easier to express in the query syntax.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,199 @@ into grouped
...
+                var userToUpdate = dbContext.Users
+                    .FirstOrDefault(u => dbContext.OrganizationUsers.Any(ou =>
+                        ou.UserId == u.Id &&
+                        ou.Id == organizationUserId &&
+                        ou.Status == OrganizationUserStatusType.Confirmed));
+
+                if (userToUpdate != null)
+                {
+                    userToUpdate.AccountRevisionDate = DateTime.UtcNow;
+                }
```
eliykat, line 454: You can use `dbContext.UserBumpAccountRevisionDateByOrganizationUserIdsAsync`.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationEnableCollectionEnhancementTests.cs
```diff
@@ -11,7 +11,7 @@ namespace Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories;
 
 public class OrganizationEnableCollectionEnhancementTests
 {
-    [DatabaseTheory, MssqlDatabaseData]
```
eliykat, line 14: Please delete the `MssqlDatabaseData` class, it was really just a hack because we didn't have an EF implementation.

eliykat (changes requested):
Great work! In addition to changes below, I suggest splitting this PR like so:

1. This PR should only add the EF repository method, and make the minor changes to the tests and Factories.cs. This can get merged to main and allow QA to manually test the repository method via opt-in (not how it'll actually be used but probably more convenient for manual testing than the migration itself). That also lets us focus on finalising the logic.
2. Create a new PR for the self-host release. In that PR, we have the migrations for **all** db providers (including mssql), and we set the `FlexibleCollectionsSignup` flag to `true`. That PR should be relatively straightforward because the work has already been done. And when we merge that PR, we know everything is set for the self-hosted release.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: The data is only actually persisted after we commit the transaction (see here: https://learn.microsoft.com/en-us/ef/core/saving/transactions#controlling-transactions).


> Is there any need to commit our work partway through?

Yes to retrieve the dirty data on subsequent queries.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: I don't like mixing them either but I'm always for readability first. I think method syntax works best for simple queries whereas query syntax is better for more complex queries (maybe because its similar to SQL).

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: I ended up changing it to this 
```suggestion
                       c.CollectionGroups.All(cg => cg.GroupId != groupId))
```

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: That is a good idea but I had to keep the `groupIdsWithAccessAll` variable otherwise the linq queries wouldn't work. If we put complex objects in the where clauses then it can't translate the query properly.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: I've implemented it this way
```suggestion
                        c.CollectionUsers.All(cu => cu.OrganizationUserId != organizationUser))
```

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
eliykat: That's fine, I still think it's an improvement!

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
eliykat: Thanks for the link, makes sense to me.

eliykat:
@justindbaur could you please take a look at the changes to `Factories.cs` and confirm they're OK?

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,185 @@ into grouped
...
+            await dbContext.SaveChangesAsync();
+
+            // Step 3: For all OrganizationUsers with Manager role or 'EditAssignedCollections' permission
+            // update their existing CollectionUser rows and insert new rows with [Manage] = 1
+            // and finally update all OrganizationUsers with Manager role to User role
+            var managerOrgUsers = dbContext.OrganizationUsers
+                .Where(ou =>
+                    ou.OrganizationId == organizationId &&
+                    (ou.Type == OrganizationUserType.Manager ||
+                     (ou.Permissions != null && EF.Functions.Like(ou.Permissions, "%\"editAssignedCollections\":true%"))))
```
eliykat, line 381: I'm not sure if `ou.Permissions` is cleaned up if they change type, to be safe you should also check for `ou.Type == OrganizationUserType.Custom`. That's also what we do in the mssql sproc.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -270,8 +271,185 @@ into grouped
...
+                                                              cu.CollectionId == cg.CollectionId &&
+                                                              cu.OrganizationUserId == manager.Id)
+                                                    select new CollectionUser
+                                                    {
+                                                        CollectionId = cg.CollectionId,
+                                                        OrganizationUserId = manager.Id,
+                                                        ReadOnly = false,
+                                                        HidePasswords = false,
+                                                        Manage = true
+                                                    }).Distinct().ToList();
```
eliykat, line 414: I'm not sure `Distinct()` is going to have the desired effect here - it'll use the default equality comparer which I assume will be based on reference equality, not value equality.

Can you please check this and also add/update a test to cover the situation where 2 groups provide a manager access to the same collection. (I believe this is the situation it intends to cover?)

eliykat (changes requested):
I closely compared it to the mssql sproc and have a few minor comments, otherwise I think we're good to go.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: They're not cleaned up, good point!

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: Distinct() in LINQ compares the property values. I had to include this because we already had one test that failed otherwise `Migrate_CustomUser_WithEditAssignedCollections_WithoutAccessAll_InGroupWithAccessAll_GivesCanManageAccessToAllCollections`.
I have also updated `Migrate_Manager_WithoutAccessAll_GivesCanManageAccess_ToGroupAssignedCollections` to test this.

r-tome:
> @justindbaur could you please take a look at the changes to `Factories.cs` and confirm they're OK? Not really sure how they were working before without this.

I have cut those changes from this PR as the EF migrations will be on another upcoming PR.

rkac-bw (approved):
LGTM

PR #3869, opened 2024-03-04
[AC-1864] Event type for initiation path

cyprain-okeke:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

We need an Upgrade Path event type reference event in the database. They will track:

2-person org → Families, Teams Starter, Teams, Enterprise

Families → Teams Starter, Teams, Enterprise

Teams Starter → Teams, Enterprise

Teams → Enterprise

## Code changes

## Before you submit


src/Core/OrganizationFeatures/OrganizationSubscriptions/UpgradeOrganizationPlanCommand.cs
```diff
@@ -295,6 +298,25 @@ public class UpgradeOrganizationPlanCommand : IUpgradeOrganizationPlanCommand
...
+            [(ProductType.Free, ProductType.TeamsStarter)] = "2-person org → Teams Starter",
+            [(ProductType.Free, ProductType.Teams)] = "2-person org → Teams",
+            [(ProductType.Free, ProductType.Enterprise)] = "2-person org → Enterprise",
+            [(ProductType.Families, ProductType.TeamsStarter)] = "Families → Teams Starter",
+            [(ProductType.Families, ProductType.Teams)] = "Families → Teams",
+            [(ProductType.Families, ProductType.Enterprise)] = "Families → Enterprise",
+            [(ProductType.TeamsStarter, ProductType.Teams)] = "Teams Starter → Teams",
+            [(ProductType.TeamsStarter, ProductType.Enterprise)] = "Teams Starter → Enterprise",
+            [(ProductType.Teams, ProductType.Enterprise)] = "Teams → Enterprise"
+        };
```
audrey-jensen, line 315: ⚠️ Could this mapping be pulled into a static variable? That'd prevent it from being initialized every time this method is invoked.

🤔 This might also be better structured as a lookup for each member of the tuple.

```csharp
var memberships = new Dictionary<ProductType> {
  [ProductType.Free] = "2-person org",
  [ProductType.Families] = "Families",
  [ProductType.TeamsStarter] = "Teams Starter",
  // ...
};

if(memberships.TryGetValue(oldProductType, out oldDescription) &&
   memberships.TryGetValue(newProductType, out newDescription))
{
  return $"{oldDescription} → {newDescription}";
} else {
  return null;
}
```

src/Core/Tools/Models/Business/ReferenceEvent.cs
```diff
@@ -243,4 +243,12 @@ public ReferenceEvent(ReferenceEventType type, IReferenceable source, ICurrentCo
     /// the value should be <see langword="null" />.
     /// </value>
     public string SignupInitiationPath { get; set; }
+
+    /// <summary>
+    /// The will hold the details around the plan cusomer is and the plan they are migrating to. For example, "Teams Starter → Teams, Enterprise".
```
audrey-jensen, line 248: 🎨 Just a suggestion to improve the clarity of these docs.

```suggestion
    /// The upgrade applied to an account. The current plan is listed first,
    /// followed by the plan they are migrating to. For example, 
    /// "Teams Starter → Teams, Enterprise".
```

src/Core/Tools/Models/Business/ReferenceEvent.cs
```diff
@@ -243,4 +243,12 @@ public ReferenceEvent(ReferenceEventType type, IReferenceable source, ICurrentCo
...
+    /// </summary>
+    /// <value>
+    /// <see langword="null"/> when the event was not originated by an application.
```
audrey-jensen, line 251: I think the following is accurate, unless the event isn't filed at all when a downgrade occurs.

```suggestion
    /// <see langword="null"/> when the event was not originated by an application,
    /// or when a downgrade occurred.
```

audrey-jensen (changes requested):
⚠️ Requested change
🤔 Thoughts that could benefit from discussion
🎨 Suggestion

src/Core/Tools/Models/Business/ReferenceEvent.cs
cyprain-okeke: added recommended comment

src/Core/Tools/Models/Business/ReferenceEvent.cs
cyprain-okeke: added recommended comment

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpgradeOrganizationPlanCommand.cs
cyprain-okeke: You are right, I have the refactored the code  as suggested

src/Core/Tools/Models/Business/ReferenceEvent.cs
```diff
@@ -245,10 +245,13 @@ public ReferenceEvent(ReferenceEventType type, IReferenceable source, ICurrentCo
...
     /// <summary>
-    /// The will hold the details around the plan cusomer is and the plan they are migrating to. For example, "Teams Starter → Teams, Enterprise".
+    /// The upgrade applied to an account. The current plan is listed first,
+    /// followed by the plan they are migrating to. For example,
+    /// "Teams Starter → Teams, Enterprise".
     /// </summary>
     /// <value>
-    /// <see langword="null"/> when the event was not originated by an application.
+    /// <see langword="null"/> when the event was not originated by an application,
+    /// or when a downgrade occurred.
```
audrey-jensen, line 254: ⚠️ Now that the new and old product types are looked up separately, the field tracks downgrades, so this line isn't correct. It will be null, however, when either the new or old product description isn't tracked.

I see two ways to resolve this:

1. If this behavior is intended, update the docs.
2. If an upgrade should always be tracked, provide a fallback. Delete "or when a downgrade occurred" from the docs.

Here's one way the fallback could work. This falls back to the enum's name when there's no description.

```csharp
var oldDescription = _upgradePath.TryGetValue(oldProductType, out oldDescription) ? oldDescription : $"{oldProductType:G}";
```

Sorry for not recognizing this when I recommended the other code edit. 😓

audrey-jensen (changes requested):
⚠️ requested change

src/Core/Tools/Models/Business/ReferenceEvent.cs
cyprain-okeke: thank you so much, for pointing it out

audrey-jensen (approved):

addisonbeck (approved):

PR #3873, opened 2024-03-04
[PM-6631] Handle Fido2VerificationException during passkey attestation and assertion

trmartin4:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

In https://github.com/bitwarden/server/pull/3615 we handled the `Fido2VerificationException` when asserting a WebAuthn credential for 2FA.

In this PR, we address the `MakeNewCredentialAsync` methods similarly, as well as the `MakeAssertionAsync` when asserting a WebAuthn credential for login, which was missed in #3615 .

📓 We have https://bitwarden.atlassian.net/browse/PM-4172 in the backlog to consolidate the implementations, at which point we should consider an abstraction.

## Code changes
* **AssertWebAuthnLoginCredentialCommand:** Added try/catch around assertion that returns a `BadRequestException` instead of the unhandled exception returned previously.  This will be handled on the client, as it is the pattern already established in the class for communicating assertion errors.
* **CreateWebAuthnLoginCredentialCommand:** Added try/catch around attestation that returns `false` along with a log message. I did this instead of throwing a `BadRequestException` as this is the pattern already established in this command for handling invalid data.  I added a log here as returning `false` gives no indication of the root cause.
* **UserService:** Added try/catch around attestation that returns `false` along with a log message. I did this instead of throwing a `BadRequestException` as this is the pattern already established in this command for handling invalid data.  I added a log here as returning `false` gives no indication of the root cause.

## Before you submit


trmartin4:
@coroiu do you recall why the `Assert...` command handles validation issues by throwing a `BadRequestException`, whereas the `Create...` command returns `false`?  Is it because we are already returning additional data in the assertion, and it was more flexible to throw the exceptions?

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/AssertWebAuthnLoginCredentialCommand.cs
```diff
@@ -46,8 +46,17 @@ public async Task<(User, WebAuthnCredential)> AssertWebAuthnLoginCredential(Asse
...
+        Fido2NetLib.Objects.AssertionVerificationResult assertionVerificationResult = null;
+        try
+        {
+            assertionVerificationResult = await _fido2.MakeAssertionAsync(
+                assertionResponse, options, credentialPublicKey, (uint)credential.Counter, callback);
+        }
+        catch (Fido2VerificationException)
+        {
+            throw new BadRequestException("Unable to verify credential.");
+        }
```
coroiu, line 59: Did I write this?? I must've completely missed the fact that `_fido2.MakeAssertionAsync` can throw exceptions... Good catch!

❓ **issue:** So one thing we need to be careful with here is that we don't allow credential and/or user enumeration. If we have unique exception messages here then the caller will be able to differentiate between 

- `Invalid Credential -> Not found in database`
- `Unable to verify credential -> Found in database, but invalid assertion`

trmartin4: I've adjusted the messages with [`bd46adc` (#3873)](https://github.com/bitwarden/server/pull/3873/commits/bd46adce7727750045f31dda692f0caf6d5b50e2).

trmartin4:
I had to make the `CreateWebAuthnLoginCredentialCommand` `public` in order for the tests to pass, given that I had a dependency on `ILogger<T>`.  I'm not quite sure why, which doesn't give me a good feeling about it - happy to change it back if you have any ideas about another solution.

coroiu (approved):
A blast from the past :)

JaredSnider-Bitwarden (commented):
Really nice error handling refactors!  Do you think we should add tests for these commands / user service scenarios while we are in here?

PR #3894, opened 2024-03-13
[PM-3176] Extract IOrganizationService.SaveUserAsync to a command

r-tome:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Extract `OrganizationService.SaveUserAsync` to a command.

## Code changes

* **src/Api/AdminConsole/Controllers/OrganizationUsersController.cs:** Replaced `IOrganizationService.SaveUserAsync` with `IUpdateOrganizationUserCommand.UpdateUserAsync`
* **src/Api/AdminConsole/Public/Controllers/MembersController.cs:** Replaced `IOrganizationService.SaveUserAsync` with `IUpdateOrganizationUserCommand.UpdateUserAsync`
* **src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IUpdateOrganizationUserCommand.cs:** Command interface
* **src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommand.cs:** Command with code extracted from `OrganizationService`
* **src/Core/AdminConsole/Services/IOrganizationService.cs:** Deleted method `SaveUserAsync` that was extracted
* **src/Core/AdminConsole/Services/Implementations/OrganizationService.cs:** Deleted method `SaveUserAsync` that was extracted
* **src/Core/OrganizationFeatures/OrganizationServiceCollectionExtensions.cs:** Registered `UpdateOrganizationUserCommand` for DI
* **test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommandTests.cs:** Added unit tests for `UpdateOrganizationUserCommand`
* **test/Core.Test/AdminConsole/Services/OrganizationServiceTests.cs:** Removed tests for the extracted method and added missing tests for other methods

## Before you submit


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IUpdateOrganizationUserCommand.cs
eliykat, line 1: Please enable nullable and mark `groups` as nullable.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommand.cs
```diff
@@ -0,0 +1,114 @@
...
+        if (user.Id.Equals(default(Guid)))
+        {
+            throw new BadRequestException("Invite the user first.");
+        }
+
+        var originalUser = await _organizationUserRepository.GetByIdAsync(user.Id);
+        if (user.Equals(originalUser))
+        {
+            throw new BadRequestException("Please make changes before saving.");
+        }
```
eliykat, line 50: Does this actually work? I assume it's comparing object references, which will always be false, but the intent is presumably to check value equality. (Note that the unit test uses the same object reference, which you'd never get in practice.)

If it doesn't work, I suggest just removing it as it's effectively dead code and I don't think this check is that important.

eliykat (changes requested):
A couple of minor comments, otherwise looks great.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/UpdateOrganizationUserCommand.cs
r-tome: It does not, I just tested it to be sure.

eliykat (approved):
Great work!

PR #3899, opened 2024-03-13
DEVOPS-1790 - Add Cleanup RC Branch workflow

vgrassia:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This PR adds the `Cleanup RC Branch` workflow which automatically runs when a new release is tagged.  It checks for the presence of the `hotfix-rc` branch first and deletes it if it exists, otherwise it checks for the `rc` branch and deletes it if it exists.

## Code changes

* **.github/workflows/cleanup-rc-branch.yml:** Add workflow.

## Before you submit

### DIFF
--- a/.github/workflows/cleanup-rc-branch.yml
+++ b/.github/workflows/cleanup-rc-branch.yml
@@ -34,7 +34,7 @@ jobs:
         run: |
           hotfix_rc_branch_check=$(git ls-remote --heads origin hotfix-rc | wc -l)
           rc_branch_check=$(git ls-remote --heads origin rc | wc -l)
-          
+
           if [[ "${hotfix_rc_branch_check}" -gt 0 ]]; then
             echo "hotfix-rc branch exists." | tee -a $GITHUB_STEP_SUMMARY
             echo "name=hotfix-rc" >> $GITHUB_OUTPUT
--- a/.github/workflows/cleanup-rc-branch.yml
+++ b/.github/workflows/cleanup-rc-branch.yml
@@ -1,0 +1,53 @@
+---
+name: Cleanup RC Branch
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-rc-branch.yml
```diff
@@ -0,0 +1,53 @@
...
+          hotfix_rc_branch_check=$(git ls-remote --heads origin hotfix-rc | wc -l)
+          rc_branch_check=$(git ls-remote --heads origin rc | wc -l)
+
+          if [[ "${hotfix_rc_branch_check}" -gt 0 ]]; then
+            echo "hotfix-rc branch exists." | tee -a $GITHUB_STEP_SUMMARY
+            echo "name=hotfix-rc" >> $GITHUB_OUTPUT
+          elif [[ "${rc_branch_check}" -gt 0 ]]; then
+            echo "rc branch exists." | tee -a $GITHUB_STEP_SUMMARY
+            echo "name=rc" >> $GITHUB_OUTPUT
+          fi
```
mimartin12, line 44: ❓ This logic assumes that we will always have either `hotfix-rc` or `rc` exist when a tag has been created (trigger for the workflow). Is this always the case?

vgrassia: This is the case for the way we currently deploy.  We can't delete `hotfix-rc` or `rc` before we publish the release in GitHub, which creates the tag.  This would be the first thing to run as soon as we hit the publish button.  That being said, the logic there is designed so it won't fail, just silently continue.  I chose that because there may be a scenario where someone might manually tag a release and we don't want to assume we can delete any branches then.

mimartin12 (approved):

PR #3910, opened 2024-03-18
[PM-6909] Centralize database migration logic

withinfocus:
## Objective

Centralizes database migration logic.

## Before you submit


util/Migrator/DbMigrator.cs
```diff
@@ -12,39 +12,34 @@ public class DbMigrator
 {
     private readonly string _connectionString;
     private readonly ILogger<DbMigrator> _logger;
-    private readonly string _masterConnectionString;
 
-    public DbMigrator(string connectionString, ILogger<DbMigrator> logger)
```
withinfocus, line 17: ℹ️ The logger is now created internally. Three callsites with no customization and two without DI.

util/Migrator/DbMigrator.cs
```diff
@@ -12,39 +12,34 @@ public class DbMigrator
...
-        CancellationToken cancellationToken = default(CancellationToken))
+        CancellationToken cancellationToken = default)
     {
         var attempt = 1;
-
         while (attempt < 10)
         {
             try
             {
+                PrepareDatabase(cancellationToken);
```
withinfocus, line 32: ℹ️ Separation of "preparation" tasks from the actual migration. The tasks in here really don't have to run any time but once, but not going after that here.

util/Migrator/DbMigrator.cs
```diff
@@ -56,17 +51,14 @@ public DbMigrator(string connectionString, ILogger<DbMigrator> logger)
         return false;
     }
 
-    public bool MigrateDatabase(bool enableLogging = true,
-        bool repeatable = false,
-        string folderName = MigratorConstants.DefaultMigrationsFolderName,
-        CancellationToken cancellationToken = default(CancellationToken))
+    private void PrepareDatabase(CancellationToken cancellationToken = default)
     {
-        if (_logger != null)
```
withinfocus, line 64: ℹ️ This is now lower down.

❓ There was a confusing mix of logger usages in this class so I just went with "use it or not". The only project not logging is `Setup` -- why? Do we want to suppress setup log messages?

util/Migrator/DbMigrator.cs
```diff
@@ -119,20 +126,13 @@ public DbMigrator(string connectionString, ILogger<DbMigrator> logger)
...
         if (enableLogging)
         {
-            if (_logger != null)
-            {
-                builder.LogTo(new DbUpLogger(_logger));
-            }
-            else
-            {
-                builder.LogToConsole();
-            }
```
withinfocus, line 129: ℹ️ This seems to be what our logger will do ...

util/Migrator/SqlServerDbMigrator.cs
```diff
@@ -1,109 +1,21 @@
...
-using Microsoft.Extensions.Logging;
 
 namespace Bit.Migrator;
 
 public class SqlServerDbMigrator : IDbMigrator
 {
-    private readonly string _connectionString;
-    private readonly ILogger<SqlServerDbMigrator> _logger;
-    private readonly string _masterConnectionString;
+    private readonly DbMigrator _migrator;
```
withinfocus, line 8: ❓ Everything in this class was duplicated by the core migrator, so now it just uses that like `Setup` and the MSSQL utility. Am I missing anything? Why was this copied?

util/MsSqlMigratorUtility/Program.cs
```diff
@@ -15,60 +12,26 @@ private static int Main(string[] args)
     public void Execute(
         [Operand(Description = "Database connection string")]
         string databaseConnectionString,
-        [Option('v', "verbose", Description = "Enable verbose output of migrator logs")]
```
withinfocus, line 18: ❓ Nothing emitted debug messages so this seemed unnecessary. Am I missing something here? Do we use this anywhere, like with automation?

util/Setup/Program.cs
```diff
@@ -185,17 +186,19 @@ private static void PrintEnvironment()
...
-    private static void MigrateDatabase(int attempt = 1)
+    private static void PrepareAndMigrateDatabase()
     {
         var vaultConnectionString = Helpers.GetValueFromEnvFile("global",
             "globalSettings__sqlServer__connectionString");
-        var migrator = new DbMigrator(vaultConnectionString, null);
+        var migrator = new DbMigrator(vaultConnectionString);
 
-        var log = false;
+        var enableLogging = false;
```
withinfocus, line 195: ℹ️ If we determine this is unnecessary I can kill off the whole parameter down the chain.

withinfocus:
Added a few reviewers directly with historical knowledge and for any gotchas.

util/MsSqlMigratorUtility/Program.cs
michalchecinski: It's used in the automations. If the `verbose` flag is set during execution, the Migrator logs the SQL scripts applied during the run. If we want to remove the `verbose` option, I would like to set the `enableLogging` param in the `MigrateMsSqlDatabaseWithRetries` method to always be `true`, like you did in lines 28 and 32. This is helpful to see which scripts are run in the automation (to confirm if all needed have been applied).

withinfocus: OK, so even though you invoked this as "verbose", what it was actually doing is simply logging. It configured a debug-level log emission, but nothing logs at that level. I think this is fine then and at some point the automations could omit that parameter once this merges.

You can see in the Action run logs for this PR that you still get those migration messages. Please review the database testing run if you don't mind to ensure this works as you (DevOps) expect.

michalchecinski: I just checked it, and it works as expected 😃

michalchecinski: @withinfocus We need to ensure that this change is backwards compatible. If we use the `-v` flag after removing the verbose option, will the `MsSqlMigratoeUtility` run as expected, or will it throw an error stating that it does not know the `-v` option?

withinfocus: Sigh, I hacked around on this and can't get `Unrecognized option '-v'` to leave. I pushed https://github.com/bitwarden/server/pull/3910/commits/32b0801a6ebfbb8b54dc901126822af3a277d966 just now for this -- is that all that's needed?

rkac-bw (dismissed):
LGTM

util/MsSqlMigratorUtility/Program.cs
michalchecinski: The `${@}` should stay, as it passes other args, e.g. the `--repeatable` or `--folder`.

util/MsSqlMigratorUtility/Dockerfile
```diff
@@ -5,4 +5,4 @@ LABEL com.bitwarden.product="bitwarden"
 WORKDIR /app
 COPY obj/build-output/publish .
 
-ENTRYPOINT ["sh", "-c", "dotnet /app/MsSqlMigratorUtility.dll \"${MSSQL_CONN_STRING}\" -v ${@}", "--" ]
+ENTRYPOINT ["sh", "-c", "dotnet /app/MsSqlMigratorUtility.dll \"${MSSQL_CONN_STRING}\"", "--" ]
```
michalchecinski, line 8:
```suggestion
ENTRYPOINT ["sh", "-c", "dotnet /app/MsSqlMigratorUtility.dll \"${MSSQL_CONN_STRING}\" ${@}", "--" ]
```

michalchecinski (changes requested):

rkac-bw (dismissed):

michalchecinski (dismissed):

util/Migrator/DbMigrator.cs
justindbaur: `SqlServerDbMigrator` got theirs from DI which could be customized, how about leave a constructor that takes it but offer one without which will just create the default one?

withinfocus: Lemme know if https://github.com/bitwarden/server/pull/3910/commits/7d1eee7ae3a3f8ac5c45ce73263e86eeb60a5f37 works for you.

justindbaur: Perfect

rkac-bw (approved):
LGTM

michalchecinski (approved):

PR #3915, opened 2024-03-19
[PM-6934] Prevent enabling two step login policy if any Org member has no master password and no 2FA set up

r-tome:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/src/Core/AdminConsole/Services/Implementations/PolicyService.cs
+++ b/src/Core/AdminConsole/Services/Implementations/PolicyService.cs
@@ -124,14 +124,15 @@ public class PolicyService : IPolicyService
                 switch (policy.Type)
                 {
                     case PolicyType.TwoFactorAuthentication:
-                        foreach (var orgUser in removableOrgUsers)
+                        var orgUsersWithoutMasterPassword = orgUsers.Where(ou =>
+                            ou.Status != OrganizationUserStatusType.Invited &&
+                            ou.Status != OrganizationUserStatusType.Revoked &&
+                            !ou.HasMasterPassword);
+                        foreach (var orgUser in orgUsersWithoutMasterPassword)
                         {
                             if (!await userService.TwoFactorIsEnabledAsync(orgUser))
                             {
-                                await organizationService.DeleteUserAsync(policy.OrganizationId, orgUser.Id,
-                                    savingUserId);
-                                await _mailService.SendOrganizationUserRemovedForPolicyTwoStepEmailAsync(
-                                    org.DisplayName(), orgUser.Email);
+                                throw new BadRequestException("Policy could not be enabled. Members of your organization would lose access to their accounts if this policy were enabled.");
                             }
                         }
                         break;


eliykat (changes requested):
Per Alison's latest comment in Slack (after you'd logged off):

> we just need to block it if there are users that both don’t have MP and 2FA, we shouldn’t change the existing process for everyone else.

So if a user has an MP but not 2FA, they still need to be removed.

I suggest looping through `removableOrgUsers` to build a list of `removableOrgUsers` without 2FA enabled. (Only do that once because the `UserService` call is expensive.)

Then you can check `removableOrgUsersWithout2FA.Any(ou => !ou.HasMasterPassword)` and throw the error if required.

After that, you can have the existing code that removes users without 2FA.

Please also make sure you add tests (I know you probably just ran out of time.)

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
```diff
@@ -341,6 +360,75 @@ await sutProvider.GetDependency<IPolicyRepository>().Received()
...
+        SetupOrg(sutProvider, policy.OrganizationId, org);
+
+        sutProvider.GetDependency<IPolicyRepository>()
+            .GetByIdAsync(policy.Id)
+            .Returns(new Policy
+            {
+                Id = policy.Id,
+                Type = PolicyType.TwoFactorAuthentication,
+                Enabled = false,
+            });
```
eliykat, line 385: It's confusing that this isn't returning the `policy` object from the test parameters.

Either:
* use `Policy policy` in the test parameters and return that from your mock db call, or
* use a `Guid policyId` in the test parameters and create your policy object in the body of the test

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
```diff
@@ -341,6 +360,75 @@ await sutProvider.GetDependency<IPolicyRepository>().Received()
...
+            UserId = Guid.NewGuid(),
+            HasMasterPassword = false
+        };
+
+        sutProvider.GetDependency<IOrganizationUserRepository>()
+            .GetManyDetailsByOrganizationAsync(policy.OrganizationId)
+            .Returns(new List<Core.Models.Data.Organizations.OrganizationUsers.OrganizationUserUserDetails>
+            {
+                orgUserDetailUser,
+            });
```
eliykat, line 404: I would like this to return multiple orgUsers, with some being eligible for removal (i.e. MP + no 2FA) and some not being eligible (i.e. no MP + no 2FA), in a semi-random order. That way we can make assert that **none** are removed. The idea is to make sure that your `OrderBy` statement in the SUT works as intended.

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
```diff
@@ -341,6 +360,75 @@ await sutProvider.GetDependency<IPolicyRepository>().Received()
         Assert.True(policy.RevisionDate - utcNow < TimeSpan.FromSeconds(1));
     }
 
+    [Theory, BitAutoData]
+    public async Task SaveAsync_ExistingPolicy_UpdateTwoFactor_WithoutMasterPasswordOr2FA_ThrowsBadRequest(
+        [AdminConsoleFixtures.Policy(PolicyType.TwoFactorAuthentication)] Policy policy, SutProvider<PolicyService> sutProvider)
+    {
+        // If the policy that this is updating isn't enabled then do some work now that the current one is enabled
```
eliykat, line 367: I don't understand this comment, is it still current?

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
```diff
@@ -341,6 +360,75 @@ await sutProvider.GetDependency<IPolicyRepository>().Received()
...
+        sutProvider.GetDependency<IPolicyRepository>()
+            .GetByIdAsync(policy.Id)
+            .Returns(new Policy
+            {
+                Id = policy.Id,
+                Type = PolicyType.TwoFactorAuthentication,
+                Enabled = false,
+            });
+
+        var orgUserDetailUser = new Core.Models.Data.Organizations.OrganizationUsers.OrganizationUserUserDetails
```
eliykat, line 387: nit, seeing as how I'm already asking for changes: please bring this into scope with `using` instead of the super long chain of namespaces.

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
```diff
@@ -124,10 +124,16 @@ public class PolicyService : IPolicyService
                 switch (policy.Type)
                 {
                     case PolicyType.TwoFactorAuthentication:
-                        foreach (var orgUser in removableOrgUsers)
+                        foreach (var orgUser in removableOrgUsers.OrderBy(ou => !ou.HasMasterPassword))
```
eliykat, line 127: This appears to be sorting the list in the reverse order that you expect. You can remove the `!` which will sort it in ascending order by default (false values will come first). I would also recommend adding a comment to explain why the order is important.

eliykat (dismissed):

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
r-tome: You're absolutely right, the unit tests confirmed that!

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
r-tome: No, it was left from a copying and pasting another unit test.

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
r-tome: Same thing here, this was copied from another test. Removed it as it doesn't matter if its new or updating.

test/Core.Test/AdminConsole/Services/PolicyServiceTests.cs
r-tome: Updated to reflect that!

vincentsalucci (approved):

PR #3916, opened 2024-03-20
Include Context information in config response

MGibson1:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Adds feature flag context to config response. This is useful for debugging purposes.

Question: Is there a reason to consider this sensitive? It uses the same bearer token to, say, retrieve full sync data, so all information is retrievable through other endpoints.

## Before you submit

### DIFF
--- a/src/Core/Services/IFeatureService.cs
+++ b/src/Core/Services/IFeatureService.cs
@@ -1,5 +1,11 @@
 ﻿namespace Bit.Core.Services;
 
+public struct FeatureFlagContext
+{
+    public Guid? UserId { get; init; }
+    public Guid[] OrganizationIds { get; init; }
+}
+
 public interface IFeatureService
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/Core/Services/IFeatureService.cs
```diff
@@ -1,5 +1,11 @@
 ﻿namespace Bit.Core.Services;
 
+public struct FeatureFlagContext
```
withinfocus, line 3: 🎨 I am still hung up on this being associated to feature flags. My opinion is that context is a platform-level concept and should be there and not in these classes.

MGibson1: Good thought, I was thinking of context in terms of the LD context, but that's entirely derived from `CurrentContext`. A safe report of that context should fully describe anything that any given Feature Flag provider could use to key off of

PR #3934, opened 2024-03-28
[AC-2172] Member modal - limit admin access

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Back-end changes for bitwarden/clients#8343. 

~Stacked on #3793.~

Adds additional authorization logic on the following endpoints:
* OrganizationUser.Invite
* OrganizationUser.Put

Also updates authorization handlers for v1 logic.

## Code changes

* **OrganizationUsersController**
  * Invite - check that the saving user can modify access for the collections they're trying to give the invited user
  * Put - this is more complex. Remember that the access-selector does not emit a value for disabled rows, so the client will only send values for the collections the user can modify. This method now (1) ensures that the user can modify the collection access sent by the client, and (2) concats them with collection access that the user _can't_ change before saving to the database. This ensures we don't overwrite collection access that we don't want to change.

* **BulkCollectionAuthorizationHandler**
  * cache the managed collections so that we can call this in a loop in the Put method, above. (Now that I've done this, I think we could remove the "bulk" implementation of this, which would greatly simplify things - but I'll leave that for another day.)
  * change the `Update` method to take into account v1 flag and collection management setting.

* **OrganizationUser_ReadWithCollectionsById** - maybe the last sproc that didn't have the Manage property 😁 

## Before you submit


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -183,16 +184,29 @@ public async Task<OrganizationUserResetPasswordDetailsResponseModel> GetResetPas
     }
 
     [HttpPost("invite")]
-    public async Task Invite(string orgId, [FromBody] OrganizationUserInviteRequestModel model)
+    public async Task Invite(Guid orgId, [FromBody] OrganizationUserInviteRequestModel model)
```
vincentsalucci, line 187: Seems like we're starting to get some more business logic in the controller. Do you have any plans/concern here?

vincentsalucci (approved):
Besides my question, I think this is good!

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
eliykat: I agree, but it is authorization logic still - you only have to sort the groups and collections in this way because you're accessing it via this endpoint as a user. It doesn't apply to using the public api, for example. This is reflected by the fact that it requires `AuthorizationService` for this logic, which isn't available in the core layer. So I think it stays here for now, although I do agree that the endpoint itself is on the longer side.

eliykat:
Adding the hold label to remind myself not to merge yet - it's stacked on another feature branch.

Jingo88 (dismissed):

eliykat:
I realised that these changes weren't feature flagged properly - PUT did not check for `org.FlexibleCollections`. PUT and POST both contained a v1 flag check inside the authorization handler, but I think it's better if we do it expressly in the controller as well. Therefore:
* POST now checks the v1 flag explicitly (as well as `org.FlexibleCollections`)
* PUT has been split into a separate vNext method which is only called if `org.FlexibleCollections` and v1 flag is enabled.

eliykat:
I've rebased this on main so that it's not blocked by https://github.com/bitwarden/server/pull/3793. It can't be QA'd without it, but it can be reviewed and merged, and new code is feature flagged so it won't do any harm in the meantime.

rkac-bw (approved):
lgtm

Jingo88 (approved):

vincentsalucci (approved):

PR #3938, opened 2024-03-28
[AC-2211] SM Changes

cturnbull-bitwarden:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Other
```

## Objective

Client side PR: https://github.com/bitwarden/clients/pull/8531

## Code changes

## Before you submit


src/Core/Enums/PlanType.cs
```diff
@@ -28,14 +28,24 @@ public enum PlanType : byte
...
+    TeamsAnnually2023 = 13,
+    [Display(Name = "Enterprise (Monthly) 2023")]
+    EnterpriseMonthly2023 = 14,
+    [Display(Name = "Enterprise (Annually) 2023")]
+    EnterpriseAnnually2023 = 15,
+    [Display(Name = "Teams Starter 2023")]
+    TeamsStarter2023 = 16,
     [Display(Name = "Teams (Monthly)")]
-    TeamsMonthly = 12,
+    TeamsMonthly = 17,
```
amorask-bitwarden, line 42: 💭 I thought this wrong for a second, but this is how we grandfather in existing organizations, right? They're DB column will still point to the 2023 ints; clever.

amorask-bitwarden (approved):
Ez pz. We should probably save this PR in Confluence or pin it in our dev channel as an example of how to do a backwards compatible plan update. It's been the easiest 1 to digest so far that I've seen.

amorask-bitwarden (dismissed):

src/Core/Enums/PlanType.cs
cturnbull-bitwarden: Yeah, exactly. This will only affect new organizations

amorask-bitwarden (approved):

PR #3966, opened 2024-04-08
[AC-2420] Fix customer discount ID and SM invite validation

amorask-bitwarden:
## Type of change

```
- [X] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

This PR includes some changes that ensure an organization that is on the Secrets Manager Standalone Trial cannot just invite Password Manager members to their organization. 

The client-side will check whether the organization is a Standalone Secrets Manager organization and, if it is, it will remove the option to "Include Secrets Manager" on a user invite and always default it to `true`. 

On the server-side, in this PR, we just need to fix an existing bug that was preventing the inviting of a single user who had both PM and SM when the organization currently had 1 PM and 1 SM seat. 

## Code changes

* **[src/Api/Models/Response/SubscriptionResponseModel.cs](https://github.com/bitwarden/server/pull/3966/files#diff-a2fd9d9d7b376a7afe3eca901e053a3848dc7c29d124fc029f4cbe43820922d5):** Modernize
* **[src/Core/Models/Business/SubscriptionInfo.cs](https://github.com/bitwarden/server/pull/3966/files#diff-e4ffeaf04a929eb8c49ed3fd3742205ad44a5faab349fc7f42d863705d3a7873):** Set `BillingCustomerDiscount.Id` equal to the Stripe Coupon ID rather than Discount ID.
* **[src/Core/AdminConsole/Services/Implementations/OrganizationService.cs](https://github.com/bitwarden/server/pull/3966/files#diff-443d95443018d2d3daf6bd19935e90c1d57489f788a64e7d790bfb8f5a68b021):** We need to auto-scale SM seats **after** we add the PM seats. If we don't, we'll throw an exception saying we can't scale SM seats past our PM seat count because the updated SM seats are being validated against the organization's current seat count; not the newly updated seat count like it should be. This removes the need for pre-validation since the actual update does the validation already.
* **[...nFeatures/OrganizationSubscriptions/Interface/IUpdateSecretsManagerSubscriptionCommand.cs](https://github.com/bitwarden/server/pull/3966/files#diff-d6728828a2484bebc5949c7b73d60b7810d6b1d685e91a04780f3591b7eba637):** No longer needed
* **[...OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs](https://github.com/bitwarden/server/pull/3966/files#diff-981e28d73c7b87a173a56b7f6e1f4366bac2f9d081dfe5f70459d025777f4914):** Method can be private.


r-tome (approved):

r-tome (commented):
Is it possible to test the bugfix with a unit test?

amorask-bitwarden:
> Is it possible to test the bugfix with a unit test?

Hi @r-tome, thank you for the review! I took a look and unfortunately it seems like this method's call depth is very large and remarkably difficult to mock start to finish. If this is a blocker I can continue working at it, but since the only change is moving an invocation that would be mocked anyway, I thought it would be easier to have QA thoroughly test invites.

PR #3967, opened 2024-04-10
[AC-2436] Show unassigned items banner

eliykat:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add a server endpoint that returns `true` if the current user is an owner or admin of any organization that has unassigned ciphers. Otherwise return `false`.

The clients will use this to show a banner: client PR is https://github.com/bitwarden/clients/pull/8655

This is limited to orgs using flexible collections because self-host has not yet been migrated and it would be premature to show them the banner.

## Code changes

Add endpoint in CiphersController. Notes:
* this uses `_cipherRepository.GetManyUnassignedOrganizationDetailsByOrganizationIdAsync` to determine whether an org has any unassigned ciphers. This is behind a feature flag elsewhere, can @shane-melton please confirm this is OK to use here? I see nothing that strictly ties it to flexible collections v1.
* looped database calls are obviously not great but (a) this is going out as a hotfix, (b) the response will be cached client-side and (c) most people have 0-1 organizations. I opted to use what we already have rather than adding database changes to the hotfix lift.

## Before you submit

### DIFF
--- a/src/Api/Vault/Controllers/CiphersController.cs
+++ b/src/Api/Vault/Controllers/CiphersController.cs
@@ -1118,8 +1118,11 @@ public class CiphersController : Controller
     [HttpGet("has-unassigned-ciphers")]
     public async Task<bool> HasUnassignedCiphers()
     {
+        var orgAbilities = await _applicationCacheService.GetOrganizationAbilitiesAsync();
+
         var adminOrganizations = _currentContext.Organizations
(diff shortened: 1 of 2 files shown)


r-tome (approved):

shane-melton (approved):
This looks good to me! 

> This is behind a feature flag elsewhere, can @shane-melton please confirm this is OK to use here? I see nothing that strictly ties it to flexible collections v1.

Yes, this is okay to use without the V1 flag.

PR #3990, opened 2024-04-15
[AC-2427] update discount logic for complimentary password manager

cturnbull-bitwarden:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Other
```

## Objective

For Secrets Manager free trials, this change will remove the 100% off discount for password manager if Secrets Manager is removed, either by removing it yourself, upgrading to a new org without it, or if CS removes it in the Stripe dashboard.
 
I also took the opportunity to refactor a bit. The diff sucks, but the gist is that the business logic inside of each event type's `if` block was moved to a separate private handler method. It's one small step that'll make moving to commands and queries, or event handlers, easier when we do that. Take a look at the last commit to see the relevant business logic changes.

## Code changes

## Before you submit


amorask-bitwarden (approved):
Looks awesome, nice work.

PR #3991, opened 2024-04-15
[AC-2403] Don't subtract grace period from expiration date when license is in trial

amorask-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This PR fixes a bug where, in the `OrganizationSubscriptionResponse` for self-hosted, the expiration date for a subscription in trial was having the grace period deducted from it when it never had it appended on the cloud side in the first place. This resulted in the appearance that the trial expiration was almost 2 months in the past for self-hosted users.

Screenshot 2024-04-15 at 3 20 21 PM


eliykat (commented):
Now the helper text is inaccurate - it says "after your subscription expires, you have 60 days to apply an updated license file", but the grace period ends on the same day as the subscription expiration.

Do trial users get a grace period? If so, it looks like further changes are required to make this work. If not, I assume the helper text should be hidden or updated.

amorask-bitwarden:
> Now the helper text is inaccurate - it says "after your subscription expires, you have 60 days to apply an updated license file", but the grace period ends on the same day as the subscription expiration.
> 
> Do trial users get a grace period? If so, it looks like further changes are required to make this work. If not, I assume the helper text should be hidden or updated.

Hi @eliykat, thanks for catching that and apologies - that was sloppy. I missed that helper text. I spoke with Product & CS and they confirmed that a subscription in Trial does not get a grace period. As such, I'll be hiding the helper text in an associated `clients` PR.

eliykat:
No problem at all - the client-side fix looks good.

PR #4003, opened 2024-04-18
Properly handle new policy enrollments in the public API

addisonbeck:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This commit patches a bug in the public API for policy management. Currently when applying a policy for the first time through the public API the desired configuration is ignored, and a default enablement of the `TwoFactorAuthentication` policy is performed instead. This usually results in no policy being applied and users being booted from the org. Check Jira for a recreation video from QA.

When applying a policy that does not exist some code gets called that looks like this:

```cs
if (existingPolicy == null)
{
    policyToApply = model.ToPolicy(_currentContext.OrganizationId.Value);
}
```

This model construction looks like this:

```cs
return ToPolicy(new Policy
{
    OrganizationId = orgId
});
```

This doesn't set any information other than the `OrganizationId`, so the default policy of `0` (or `TwoFactorAuthentication`) is what ends up being sent to the service to be applied. It needs to look more like this:

```cs
return ToPolicy(new Policy
{
    OrganizationId = orgId,
    Enabled = Enabled.GetValueOrDefault(),
    Data = Data != null ? JsonSerializer.Serialize(Data) : null,
    Type = type
});
```

Changing the model in this way resolved the issue for me during local testing.

## Code changes

- `PoliciesControllerTests`: This is a new file. I just wrote one test for the direct bug.
- `PoliciesController`: pass in the `type` input parameter to the policy constructor for new policies. I also changed the name of the query parameter from `id` to `type` to be more clear and aligned with other methods, but this change is unrelated and functionally doesn't do anything.
- `PolicyUpdateRequestModel`: Properly build out the new policy object being applied.

## Screenshots

Below is an example of me calling the policies endpoint with a `7` (`SendOptions`) policy type. The policy has not been applied to this organization before and applies successfully.

https://github.com/bitwarden/server/assets/15897251/d74c74f1-55f7-42e1-a843-091ba2409268

## References

- [Jira ticket for this work](https://bitwarden.atlassian.net/browse/AC-577)
- [A set of refactors to `PolicyService` developed while debugging this issues](https://github.com/bitwarden/server/pull/4001)

### DIFF
--- a/src/Api/AdminConsole/Public/Controllers/PoliciesController.cs
+++ b/src/Api/AdminConsole/Public/Controllers/PoliciesController.cs
@@ -93,7 +93,7 @@ public class PoliciesController : Controller
             _currentContext.OrganizationId.Value, type);
         if (policy == null)
         {
-            policy = model.ToPolicy(_currentContext.OrganizationId.Value);
+            policy = model.ToPolicy(_currentContext.OrganizationId.Value, type);
         }
         else
         {
--- a/src/Api/AdminConsole/Public/Controllers/PoliciesController.cs
+++ b/src/Api/AdminConsole/Public/Controllers/PoliciesController.cs
@@ -83,7 +83,7 @@ public class PoliciesController : Controller
     /// </remarks>
     /// <param name="type">The type of policy to be updated.</param>
     /// <param name="model">The request model.</param>
-    [HttpPut("{id}")]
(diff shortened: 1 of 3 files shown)


eliykat (approved):
Nice find

PR #4012, opened 2024-04-24
[AC-2379] Stripe webhook update provider status

cturnbull-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

Updates the Stripe webhook to enable or disable a given Provider if there are changes to the Provider's subscription status.

## Code changes

* **ProviderService.cs:** Two new methods, `EnableAsync` and `DisableAsync`, have been added. These methods change the Enabled property of a Provider and update the RevisionDate to the current UTC. They also handle updating the provider repository and application cache
* **Billing.csproj:** Updated to include a reference to `Commercial.Core.csproj` so that `IProviderService` can be added as a dependency
* **Startup.cs:** Added services to the DI container for IProviderService
* **StripeController:** Disabling or enabling a provider if its subscription status is either "unpaid" or "incomplete expired," or "active" respectively
* **Provider.cs:** Updated to allow the `RevisionDate` property to be set externally, enabling updates to this property from the new methods in `ProviderService.cs`
* **IProviderService.cs:** Updated to include declarations for the `EnableAsync` and `DisableAsync` methods
* **NoopProviderService.cs:** Dummy implementations of `EnableAsync` and `DisableAsync` have been added, throwing `NotImplementedException` when called

## Before you submit


amorask-bitwarden (approved):

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -647,6 +647,38 @@ public async Task DeleteAsync(Provider provider)
...
+    public async Task DisableAsync(Guid providerId)
+    {
+        var provider = await _providerRepository.GetByIdAsync(providerId);
+
+        if (provider is not { Enabled: true })
+        {
+            return;
+        }
+
+        provider.Enabled = true;
```
vincentsalucci, line 675: This should be `false`

vincentsalucci (changes requested):

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
cturnbull-bitwarden: Thanks for catching this! Just pushed an update

amorask-bitwarden (dismissed):

vincentsalucci (dismissed):
Non-blocking: only thing missing from this PR are a handful of tests for the new service methods added. Please follow-up at some point to get that coverage added.

amorask-bitwarden (approved):

vincentsalucci (approved):

cturnbull-bitwarden:
Close this PR as we learned that this `Enabled` flag doesn't actually do anything. In the future, we'll just use the subscription status to decide if the Provider is enabled or not.

PR #4011, opened 2024-04-23
[AC-2489] Resolve SM Standalone issues with SCIM & Directory Connector

amorask-bitwarden:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

When organization users were invited through SCIM or through a Directory Connector import, we weren't checking whether the organization was using an SM Standalone trial.

This PR makes that check and, if the organization is on SM Standalone, ensures each invite's `AccessSecretsManager` property is set to `true`.

https://github.com/bitwarden/server/assets/144709477/efb778d6-9cc1-4b75-a334-a5c80292f55d

https://github.com/bitwarden/server/assets/144709477/e1781549-486c-4871-9a16-be2fa36a6641

PR #4022, opened 2024-04-26
[AC-292] Public Api - allow configuration of custom permissions

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Allow users to configure custom permissions via the Public API.

This also involved refactoring the `OrganizationService` invite user methods so that I wasn't adding to existing tech debt. I also considered extracting this to a command, however this turned out to be non-trivial, so I think it's best tackled separately.

## Code changes

Following the commits in order:
* https://github.com/bitwarden/server/commit/b183325e6e6cead9a8ead607387196e8100ffab7 - `OrganizationService` had too many very similar methods for inviting users - `InviteUserAsync` (x2 overloads), `SaveUserSendInviteAsync`, `InviteUsersAsync` (x2 overloads), and `SaveUsersSendInvitesAsync`. It was difficult to keep track of the control flow and understand where changes needed to happen. I remember we did this originally to avoid exposing nullable parameters to outside callers, however I think the result has overall been worse for it. I combined several of these methods so that we just have `InviteUserAsync` (single) -> `InviteUsersAsync` (multiple) -> private `SaveUsersSendInvitesAsync`.
  * I think a better solution here is to have a `LogSubject` wrapper which can represent an OrgUser or an EventUser and that's only unwrapped once it reaches EventService - but out of scope here.
* https://github.com/bitwarden/server/commit/82fb3f86c487ab1beca312e2c018d8cc7ef000c0 - `InviteUserAsync` took an increasing number of parameters and I didn't want to add `Permissions` here. Refactor it to take an `OrganizationUserInvite` object, similar to `InviteUsersAsync`. Refactor calling locations to create this object (often from an existing request object) and pass it in.
* https://github.com/bitwarden/server/commit/4e65cb99d2fd210868711cba4d7e8863126edea9 - finally :) add the custom permissions to the Public API request and response models. It is optional in all cases because we don't want to break existing scripts that aren't using it (either by throwing errors or by blasting away existing values).
* https://github.com/bitwarden/server/commit/32a352591a61da06948609a7323bec5a842c065f - if a custom user is being invited, but Permissions have not been specified, use a default Permissions object. This was prompted by [AC-2318](https://bitwarden.atlassian.net/browse/AC-2318) which pointed out that our logic does not gracefully handle nulls here. That only occurred due to manipulation of the private API, however now that we're exposing it to the Public API we need to make sure we handle this case.
* add tests, minor tweaks

## Before you submit

[AC-2318]: https://bitwarden.atlassian.net/browse/AC-2318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


r-tome (commented):
Everything looks great to me! I noticed we don't have integration tests for the public `MembersController`. Is it feasible to add tests on this PR for the `Post` action?

eliykat:
I had to reapply changes from #4011 in the last merge commit. `PostUserCommand` now sets `invite.AccessSecretsManager` rather than adding another parameter to the invite user method. @r-tome please just look at that carefully to ensure I've maintained that fix.

r-tome (dismissed):
Excellent work on these integration tests!
Suggestion: on the teardown we could delete the inserted data

eliykat:
All test data is created with unique GUIDS, I think deleting data from the database would just make the tests run longer (and integration tests are already slow).

I fixed a bug where it was checking the `OrganizationUserType` before the transitional logic for deprecated permissions. Now it checks it after the user type is translated, so if a custom user is being mapped to a regular User, it won't return a permissions object.

r-tome (approved):

PR #4032, opened 2024-04-30
[AC-2538] Member modal - limit admin access - fix ManageUsers custom permission

eliykat:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Fix bug for Flexible Collections v1: the `BulkCollectionAuthorizationHandler` did not authorize the "Manage Users" or "Manage Groups" custom permissions, which broke the flow for #3934 (and would've for #3998).

This requires that we split the `ModifyAccess` operation into `ModifyUserAccess` and `ModifyGroupAccess`, to match the granularity of the custom permissions, and use these new operations where required.

## Code changes

- **src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionOperations.cs** - delete `ModifyAccess` and replace it with `ModifyUserAccess` and `ModifyGroupAccess`
- **src/Api/AdminConsole/Controllers/OrganizationUsersController.cs** - update to use `ModifyUserAccess`, which is always what you're doing via the `OrganizationUsersController`
- **src/Api/Controllers/CollectionsController.cs** - update to use the new operations as appropriate (depending on what the endpoint is updating)
- **src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs** - the biggest changes:
  - make all private methods return a bool and call `context.Succeed` from the top-level method only. I really like how this simplified the handler
  - add private methods for the 2x new operations/requirements. They are subsets of the more general `Update` operation, e.g. if you can `Update` then you can `ModifyUserAccess`. There's an argument for splitting these out completely (so that `Update` only refers to the collection and not its access) but I decided to pull up for now.
- update tests.

## Before you submit

### DIFF
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -199,7 +199,7 @@ public class OrganizationUsersController : Controller
         {
             var collections = await _collectionRepository.GetManyByManyIdsAsync(model.Collections.Select(a => a.Id));
             var authorized =
-                (await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.ModifyAccess))
+                (await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.ModifyUserAccess))
                 .Succeeded;
             if (!authorized)
             {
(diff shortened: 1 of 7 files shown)


vincentsalucci (approved):
I like centralizing the authorized boolean as well.

shane-melton (approved):
This looks great! I really like that we can simply return a boolean for the helpers and don't need to pass in the requirement everywhere!

PR #4040, opened 2024-05-01
[AC-2513] Scaling PM seat count with SM seat count

cturnbull-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

For organizations who have signed up for the Secrets Manager free trial via https://vault.bitwarden.com/#/register?org=teams&layout=secretsManager, this PR scales the organization's Password Manager seat count in step with adjustments to the organization's Secrets Manager seat count. We know when an organization has signed up to trial secrets manager because we're already applying a discount to the Password Manager subscription item. By checking for that discount, we can decide if we should scale Password Manager seats along with Secrets Manager.

This is due to a couple of factors:
1. On the frontend, we're now hiding the Password Manager subscription adjustment form from the "subscriptions" tab in the Admin Console for organizations who are trialing Secrets Manager in this way
2. We have a requirement where Secrets Manager seats cannot exceed Password Manager seats
3. We don't want to remove this requirement yet

We'll be splitting Secrets Manager to be a standalone product later this year, such that a user won't have to sign up for Password Manager in order to sign up for Secrets Manager. Until we do that, we need to do some goofy things like this to simulate that they're currently different products.

In addition, I've separated out endpoints from the Admin Console OrganizationsController.cs that are billing related into a new Billing owned OrganizationsController (same routes). As a result, the diff will be hard to read. I recommend first looking at the commit history, as I've committed in logical steps to the above description. The first commit shows the business change as requested by product, the subsequent commits are the codeowner changes. Feel free to comment on any of it.

Client-side PR: https://github.com/bitwarden/clients/pull/9011

## Code changes

* **src\Api\AdminConsole\Controllers\OrganizationsController.cs:** Moved Billing endpoints to the billing-owned controller of the same route
* **src\Api\Billing\Controllers\OrganizationBillingController.cs:** Moved `GetBilling` endpoint to this controller since it matches the base route
* **src\Api\Billing\Controllers\OrganizationsController.cs:** New billing-owned controller to house organization-level endpoints that relate to billing, like `GET {id}/billing-status`. Added new business logic to `POST {id}/sm-subscription` that scales PM seat counts with SM seat changes for SM standalone organizations
* **test\Api.Test\AdminConsole\Controllers\OrganizationsControllerTests.cs:** Moved billing unit tests to new billing-owned test file
* **test\Api.Test\Billing\Controllers\OrganizationsControllerTests.cs:** New billing-owned test file that tests organization controller

https://github.com/bitwarden/server/assets/133619638/50f68c43-fc34-484b-a648-e66d70423201

## Before you submit


src/Api/Billing/Controllers/OrganizationsController.cs
r-tome, line 1: Non-blocker: maybe we could take this opportunity to get rid of all the manual Guid parsing `var orgIdGuid = new Guid(id);`

src/Api/Billing/Controllers/OrganizationsController.cs
```diff
@@ -0,0 +1,394 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(organization.GatewayCustomerId) ||
+            string.IsNullOrWhiteSpace(organization.GatewaySubscriptionId) ||
+            model.SeatAdjustment == 0)
+        {
+            return organization;
+        }
+
+        var subscriptionInfo = await paymentService.GetSubscriptionAsync(organization);
+        if (subscriptionInfo?.CustomerDiscount?.Id != "sm-standalone")
```
r-tome, line 385: `sm-standalone` is present on `StripeConstants.cs`, could we use that here? I lean towards centralizing these types of codes within a static class so we only have to change it in one place if we ever need to.

r-tome (approved):
Great work! I left a few suggestions but none are blockers.
Also, thanks for suggesting to look through the commit history, that helped to review this 😀

src/Api/Billing/Controllers/OrganizationsController.cs
cturnbull-bitwarden: Good suggestions! I went ahead and implemented both of these suggestions in the latest two commits

r-tome (approved):
Thanks for making these changes!

amorask-bitwarden (approved):

PR #4050, opened 2024-05-03
DEVOPS-1901 Fix error: az login again to refresh permissions

urbinaalex17:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Split the remove docker images steps per registry since that's what causing the issue:

```bash
WARNING: Unable to get AAD authorization tokens with message: 2024-03-22 14:54:44.177986 An error occurred: CONNECTIVITY_REFRESH_TOKEN_ERROR
Access to registry 'bitwardenqa.azurecr.io' was denied. Response code: 403. Please try running 'az login' again to refresh permissions.
```

## Code changes

* **.github/workflows/cleanup-after-pr.yml:** Separate the removal of docker images per registry instead of using a list of registries.

## Before you submit

### DIFF
--- a/.github/workflows/cleanup-after-pr.yml
+++ b/.github/workflows/cleanup-after-pr.yml
@@ -22,7 +22,7 @@ jobs:
         run: az acr login -n $_AZ_REGISTRY --only-show-errors
 
       ########## Remove Docker images ##########
-      - name: Remove the Docker image from bitwardenprod ACR
+      - name: Remove the Docker image from ACR
         env:
           REF: ${{ github.event.pull_request.head.ref }}
           REGISTRY: bitwardenprod
--- a/.github/workflows/cleanup-after-pr.yml
+++ b/.github/workflows/cleanup-after-pr.yml
@@ -25,7 +25,6 @@ jobs:
       - name: Remove the Docker image from ACR
         env:
           REF: ${{ github.event.pull_request.head.ref }}
-          REGISTRY: bitwardenprod
           SERVICES: |
             services:
               - Admin
--- a/.github/workflows/cleanup-after-pr.yml
+++ b/.github/workflows/cleanup-after-pr.yml
@@ -19,22 +19,11 @@ jobs:
       - name: Log in to Azure ACR
         run: az acr login -n bitwardenqa
 
-      - name: Log in to Azure - production subscription
-        uses: Azure/login@92a5484dfaf04ca78a94597f4f19fea633851fa2 # v1.4.7
-        with:
-          creds: ${{ secrets.AZURE_PROD_KV_CREDENTIALS }}
-
-      - name: Log in to Azure ACR
-        run: az acr login -n bitwardenprod
-
       ########## Remove Docker images ##########
-      - name: Remove the Docker image from ACR
+      - name: Remove the Docker image from bitwardenqa ACR
         env:
           REF: ${{ github.event.pull_request.head.ref }}
-          REGISTRIES: |
-            registries:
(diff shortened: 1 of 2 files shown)


.github/workflows/cleanup-after-pr.yml
```diff
@@ -19,6 +19,47 @@ jobs:
...
+            )
+
+            if [[ "$TAG_EXISTS" == "true" ]]; then
+              echo "[*] Tag exists. Removing tag"
+              az acr repository delete --name $REGISTRY --image $SERVICE_NAME:$IMAGE_TAG --yes
+            else
+              echo "[*] Tag does not exist. No action needed"
+            fi
+          done
+
```
mimartin12, line 62: This registry isn't even used, only `bitwardenprod` https://github.com/bitwarden/server/blob/5a5e5c50580055f3dd681cf763627b0324d73e82/.github/workflows/build.yml#L13-L14. I think this, along with other references, can be removed from this workflow. Feel free to handle what is outside of scope in a different change, but this specific one can be removed.

.github/workflows/cleanup-after-pr.yml
```diff
@@ -28,13 +69,10 @@ jobs:
...
       ########## Remove Docker images ##########
-      - name: Remove the Docker image from ACR
+      - name: Remove the Docker image from bitwardenprod ACR
         env:
           REF: ${{ github.event.pull_request.head.ref }}
-          REGISTRIES: |
-            registries:
-              - bitwardenprod
-              - bitwardenqa
+          REGISTRY: bitwardenprod
```
mimartin12, line 75: 🌱  There are multiple references to `bitwardenprod` and some duplication, you could improve this by using a global environment variable. Example https://github.com/bitwarden/server/blob/5a5e5c50580055f3dd681cf763627b0324d73e82/.github/workflows/build.yml#L13-L14

mimartin12 (changes requested):
Some of the fixes include references to a container registry that just isn't used, as shown in the build workflows. We don't need it and can remove.

urbinaalex17:
@mimartin12 , is the qa registry the one that we are not using?

mimartin12:
> @mimartin12 , is the qa registry the one that we are not using?

Yes, the build workflow only pushes to prod. https://github.com/bitwarden/server/blob/5a5e5c50580055f3dd681cf763627b0324d73e82/.github/workflows/build.yml#L13-L14

.github/workflows/cleanup-after-pr.yml
urbinaalex17: Thanks for the suggestion!

.github/workflows/cleanup-after-pr.yml
urbinaalex17: Suggestion addressed. Thanks, @mimartin12!

.github/workflows/cleanup-after-pr.yml
```diff
@@ -5,36 +5,27 @@ on:
...
 
       ########## Remove Docker images ##########
       - name: Remove the Docker image from ACR
         env:
           REF: ${{ github.event.pull_request.head.ref }}
-          REGISTRIES: |
-            registries:
-              - bitwardenprod
-              - bitwardenqa
+          REGISTRY: bitwardenprod
```
mimartin12, line 28:
```suggestion
```
❓ Still needed?

mimartin12 (changes requested):

mimartin12 (approved):

PR #4053, opened 2024-05-03
Allow for bulk updating `AuthRequest` database objects

addisonbeck:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

> 🧪 **NOTE**
> Despite what the bots want you to think: I did write an integration test for this.

## Objective

To facilitate a new bulk device login request approval workflow in the admin
console we need to update `IAuthRequestRepisitory` to include an
`UpdateManyAsync()` method. It should accept a list of `AuthRequest` table
objects, and implementations will do a very simple 1:1 update of the passed
in data. 

This pull request implements this behavior in the repository and database
layers of the bitwarden server.

## Code changes

See the [commits tab](https://github.com/bitwarden/server/pull/4053/commits)

## References

* [AC-2301](https://bitwarden.atlassian.net/browse/AC-2301) is the Jira ticket for this work.
* This PR is extended by [#4064](https://github.com/bitwarden/server/pull/4064), which adds a command layer for applying business logic 
to the more general and high level functions written in this PR for the repository layer.
* There is also [#4077](https://github.com/bitwarden/server/pull/4077) which exposes an API endpoint that calls through to the methods implanted here.

[AC-2301]: https://bitwarden.atlassian.net/browse/AC-2301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


util/Migrator/DbScripts/2024-05-05_00_UpdateManyAuthRequests.sql
```diff
@@ -0,0 +1,61 @@
...
+    SELECT
+        *
+    FROM
+        sys.types
+    WHERE
+        [Name] = 'AuthRequestType' AND
+        is_user_defined = 1
+)
+BEGIN
+    CREATE TYPE [dbo].[AuthRequestType] AS TABLE(
```
rkac-bw, line 11: In discussions with @withinfocus, we would like to move away from creating type objects in sql server  and use something a little more generic like json for bulk inserts

src/Sql/dbo/Stored Procedures/AuthRequest_UpdateMany.sql
```diff
@@ -0,0 +1,30 @@
+CREATE PROCEDURE [dbo].[AuthRequest_UpdateMany]
```
ike-kottlowski, line 1: ❓ : I don't think this accomplishes the same "add or update" methodology of the EF implementation. This method only updates the Auth Requests, it doesn't also add them if they don't exist. Should Auth Requests even be approved if they don't exist?

ike-kottlowski (changes requested):
I have some concerns with the Entity Framework and Stored Procedure implementations not matching in function.

src/Sql/dbo/Stored Procedures/AuthRequest_UpdateMany.sql
addisonbeck: This is a little tricky. I don't *want* this method to be able to create auth requests, for the reasons you laid out and others. So, the Entity Framework implementation is the problem here. *But* I don't know of a way to just do a simple update in EF without the baked in creation logic - do you? 

The EF implementation does include [some handling logic](https://github.com/bitwarden/server/pull/4053/files#diff-7884ebbcbc79ad5dcaa797917eb746ee51e59df458f2424f9ec9094f59519d3fR81-R88) that checks for default IDs on incoming objects and filters them out, which should stop most creations from being possible without a serious logic error somewhere downstream. I also added an assert on https://github.com/bitwarden/server/pull/4053/commits/a8e53d844cd214b35dd440917fb753fe771cf95c to validate that objects without set ids don't make it to the update call.

This won't stop somebody from running `id = new Guid()` and getting created by `UpdateRange`, but I'm not sure how to stop that without doing a costly hit to check all passed in IDs to see if they exist. We can go that route if you want to, or do something else, or leave this as is. Let me know!

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
```diff
@@ -69,4 +69,29 @@ public async Task<ICollection<OrganizationAdminAuthRequest>> GetManyPendingByOrg
             return orgUserAuthRequests;
         }
     }
+
+    public async Task UpdateManyAsync(IEnumerable<Core.Auth.Entities.AuthRequest> authRequests)
```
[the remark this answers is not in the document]
addisonbeck: It's not supposed to do more than updating. I've left the name for now and continued this conversation below ⬇️

util/Migrator/DbScripts/2024-05-05_00_UpdateManyAuthRequests.sql
addisonbeck: Implemented as suggest on https://github.com/bitwarden/server/pull/4053/commits/cea0589d554ff1ec1a353450bda5c338c2459aec and then with some fixes on https://github.com/bitwarden/server/pull/4053/commits/632e3fa1c3d7b7a582c86c4bd90e520300e836e3 that are separated because I merged in between them.

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
ike-kottlowski, line 1: How are you ensuring that the test target's Entity Framework?

src/Sql/dbo/Stored Procedures/AuthRequest_UpdateMany.sql
ike-kottlowski: Entity Framework support is still in Beta so I'm not sure it's worth the trouble to ensure complete accuracy at this moment. 

I did notice that the `dbcontext.UpdateRange(object[])` [does check for default values](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.updaterange?view=efcore-8.0#microsoft-entityframeworkcore-dbcontext-updaterange(system-collections-generic-ienumerable((system-object)))) already in the primary key field and ignore entities that have a `default` value in the PK field. 

I don't think it's worth the effort to address at this time.

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
```diff
@@ -72,6 +72,114 @@ public class AuthRequestRepositoryTests
...
+            TrimMilliseconds(x.CreationDate) == TrimMilliseconds(y.CreationDate) &&
+            TrimMilliseconds(x.ResponseDate) == TrimMilliseconds(y.ResponseDate) &&
+            TrimMilliseconds(x.AuthenticationDate) == TrimMilliseconds(y.AuthenticationDate) &&
+            x.OrganizationId == y.OrganizationId;
+        }
+
+        // Assert that the unchanged auth request is still unchanged
+        var skippedAuthRequest = await authRequestRepository.GetByIdAsync(authRequestNotToBeUpdated.Id);
+        Assert.True(AuthRequestEquals(skippedAuthRequest, authRequestNotToBeUpdated));
+        //Assert.Equal(skippedAuthRequest.CreationDate, authRequestNotToBeUpdated.CreationDate);
```
ike-kottlowski, line 169: is this line meant to be commented?

addisonbeck: No, that's a leftover. Removed on https://github.com/bitwarden/server/pull/4053/commits/c6660d8de710d18a762a8fa57978f799b12639e7

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
addisonbeck: Database providers are registered with the test classes through the `[DatabaseTheory]` attribute. The database testing job sets up a bunch that get run in [this config](https://github.com/bitwarden/server/blob/489f6246b1922960f23902ab3db823d456503c42/.github/workflows/test-database.yml#L82-L98). You can usually see the output of the last test runs targeting EF for these tests in the output of the action, but that's complaining right now in a way I missed. Let me get that fixed up.

addisonbeck: Okay, [this workflow report](https://github.com/bitwarden/server/pull/4053/checks?check_run_id=25224171311) shows 8 tests passed in this class from the latest build. There are only two tests in the class, and the variance is that they run on different databases configured in the way described above. Hopefully this clears things up, but let me know if it doesn't!

ike-kottlowski (approved):

rkac-bw (approved):
LGTM

PR #4057, opened 2024-05-06
[AC-2313] Add Gateway fields to Provider edit in Admin

amorask-bitwarden:
## Type of change

```
- [ ] Bug fix
- [X] New feature development
- [ ] Other
```

## Objective

This PR does some refactoring around creating and modifying Pricing Plans in Admin. It also adds input fields to the Provider Edit screen so CS can view and update the Gateway fields for a Provider who has Consolidated Billing.

https://github.com/bitwarden/server/assets/144709477/2db8047d-7d14-4efc-8113-4f33e36d4761


src/Admin/AdminConsole/Controllers/ProvidersController.cs
```diff
@@ -177,14 +177,19 @@ public async Task<IActionResult> Edit(Guid id)
...
         {
             return RedirectToAction("Index");
         }
 
-        model.ToProvider(provider);
+        provider.BillingEmail = model.BillingEmail?.ToLowerInvariant().Trim();
+        provider.BillingPhone = model.BillingPhone?.ToLowerInvariant().Trim();
+        provider.Gateway = model.Gateway;
+        provider.GatewayCustomerId = model.GatewayCustomerId;
+        provider.GatewaySubscriptionId = model.GatewaySubscriptionId;
```
eliykat, line 191: What's wrong with `model.ToProvider`? :(
(To expand slightly - I like putting this kind of mapping/update logic inside models, it makes sense to me from a general OOP perspective, and it removes boilerplate from the controller or service that uses it.)

src/Admin/AdminConsole/Controllers/ProvidersController.cs
```diff
@@ -213,6 +219,13 @@ public async Task<IActionResult> Edit(Guid id, ProviderEditModel model)
         {
             foreach (var providerPlan in providerPlans)
             {
+                providerPlan.SeatMinimum = providerPlan.PlanType switch
+                {
+                    PlanType.TeamsMonthly => model.TeamsMonthlySeatMinimum,
+                    PlanType.EnterpriseMonthly => model.EnterpriseMonthlySeatMinimum,
+                    _ => providerPlan.SeatMinimum
+                };
```
eliykat, line 227: 🎨  This is a style nit only, and fairly subjective, so it's ultimately your decision: I think the switch statement makes this harder to read than just if/else, and it makes you handle a redundant default case (`providerPlan.SeatMinimum = providerPlan.SeatMinimum`). (I know this is often a suggestion by Rider, I just disagree with it.)

src/Admin/AdminConsole/Views/Providers/Edit.cshtml
```diff
@@ -48,103 +48,142 @@
...
+                <div class="form-group">
+                    <label asp-for="GatewaySubscriptionId"></label>
+                    <div class="input-group">
+                        <input type="text" class="form-control" asp-for="GatewaySubscriptionId">
+                        <div class="input-group-append">
+                            <button class="btn btn-secondary" type="button" id="gateway-subscription-link">
+                                <i class="fa fa-external-link"></i>
+                            </button>
+                        </div>
+                    </div>
```
eliykat, line 98: Optional: given that this is used in multiple locations, you could extract it into a nested view. See `Views/Providers/_ViewInformation.cshtml` as an example.

eliykat (changes requested):
A few comments, nothing major :)

src/Admin/AdminConsole/Controllers/ProvidersController.cs
amorask-bitwarden: I personally just find the signature `public virtual Provider ToProvider(Provider existingProvider)` and the practice of passing around mutable object properties in general somewhat confusing, but I've reset it back to its original state and added the new Gateway fields to the existing method.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
amorask-bitwarden: Changed to `if/else`.

src/Admin/AdminConsole/Views/Providers/Edit.cshtml
amorask-bitwarden: Going to pass on this one for now. Despite having the same inputs, the two partials between providers and organizations would not exactly be the same and I think aligning them is outside the scope of this ticket.

cturnbull-bitwarden (approved):

src/Admin/AdminConsole/Views/Providers/Edit.cshtml
eliykat: I hadn't compared them that closely, so that's fair enough. It's only used in 2 places anyway; we can revisit if it's ever needed in a third.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
eliykat: That's reasonable - do you think it would be better if it was on the object being mutated instead, e.g. `existingProvider.Update(editProviderModel)`? That way the object is updating itself and not mutating its argument. (I suspect we've done it the current way to avoid importing api models into the core layer, but you could avoid this by making `Update` an api-only extension method on `Provider`.)

Anyway - none of that needs to be done here - just thinking aloud.

eliykat (approved):
Thanks!

src/Admin/AdminConsole/Controllers/ProvidersController.cs
amorask-bitwarden: Yeah I think that could be a bit more readable. Even just using a different signature for the same approach might make it easier to reason through at first glance as well. 

For instance, it makes perfect sense for `CreateProviderModel` to have a `ToProvider` method that returns a `Provider` entity that will be inserted into the DB. But for the `EditProviderModel`, we already have an existing `Provider`, so if we're going to update in place, there's no reason to return anything and the method name could be more along the lines of `EditProviderModel.UpdateEntity(Provider)` or even just `EditProviderModel.Update(Provider)`. If AC is open to that, I can definitely start making those changes as we make more Admin updates!

eliykat: I agree with `EditProviderModel.Update(Provider)`, returning `void`!

PR #4074, opened 2024-05-10
[AC-2447] Update PutCollection to return Unavailable cipher when last Can Manage Access is Removed

Jingo88:
```
- [ X ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When a user removes the final Can Manage access from a cipher, we will make their update request, and then return the cipher with an `Unavailable` value for the client to read

## Screen Recording

https://github.com/bitwarden/server/assets/8302660/5d0bda01-88b9-4f26-9db1-116430d1c873


shane-melton (changes requested):
I talked a bit about this solution during the FC dev sync and we came up with a potentially cleaner solution that avoids the extra repository call and adding the `Unavailable` property to every cipher details response.

We can create an `OptionalCipherDetailsResponseModel` that has a nested response model and the unavailable flag. We can then use that response type instead of the `CipherDetailsResponseModel` for the endpoint return type.

```c#
public class OptionalCipherDetailsResponseModel : ResponseModel
{
    public bool Unavailable { get; set; }
    
    public CipherDetailsResponseModel? Cipher { get; set; }
    
    private OptionalCipherDetailsResponseModel()
        : base("optionalCipherDetails") { }
}
```


Also,  I do believe we'll need to duplicate this endpoint with a `vNext` version to maintain compatibility with other clients that use this same endpoint and will not be expecting the new model until they're updated.

Let me know what you think @Jingo88!

shane-melton (changes requested):
Looking good! Just a couple of test suggestions

shane-melton (approved):

PR #4087, opened 2024-05-15
[AC-240] - BUG - Confirm Admin/Owners to org when excluded from Single Org Policy

vincentsalucci:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

> Owners and Admins were incorrectly being denied confirmation to organizations where a Single Org requirement was in place. Align the `confirm` policy checks with the `invite/accepted` pattern for excluding `Owner/Admin` user types where necessary.

## Code changes

* **OrganizationService.cs**: Removed `policyRepository` call and replaced it with a method used to grab all policies being applied to a given user. This aligns our pattern with how we check against applied policies during the `invite/accept` stages. Updated the `CheckPolicies` method to handle the new flow.
* **OrganizationUserPolicyDetailsFixtures.cs**: Helper fixture that creates an `OrganizationUserPolicyDetails` object with the given `PolicyType` to help reduce customization of the objects within the testing suite.
* **OrganizationServiceTests.cs**: Removed references to the `policyRepository` calls within the `ConfirmUser(s)` tests. Updated all altered flows with the new policy objects being returned. Added a few tests to round out extended use cases. 

## Before you submit


src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1449,22 +1448,27 @@ public async Task AutoAddSeatsAsync(Organization organization, int seatsToAdd, D
         }
     }
 
-    private async Task CheckPolicies(ICollection<Policy> policies, Guid organizationId, User user,
+    private async Task CheckPolicies(Guid organizationId, User user,
```
r-tome, line 1451: I would add the user type as a parameter here

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1449,22 +1448,27 @@ public async Task AutoAddSeatsAsync(Organization organization, int seatsToAdd, D
...
-        var usingTwoFactorPolicy = policies.Any(p => p.Type == PolicyType.TwoFactorAuthentication && p.Enabled);
-        if (usingTwoFactorPolicy && !await userService.TwoFactorIsEnabledAsync(user))
+        // Enforce Two Factor Authentication Policy for this organization
+        var orgRequiresTwoFactor = (await _policyService.GetPoliciesApplicableToUserAsync(user.Id, PolicyType.TwoFactorAuthentication)).Any(p => p.OrganizationId == organizationId);
+        if (orgRequiresTwoFactor && !await userService.TwoFactorIsEnabledAsync(user))
         {
             throw new BadRequestException("User does not have two-step login enabled.");
         }
 
-        var usingSingleOrgPolicy = policies.Any(p => p.Type == PolicyType.SingleOrg && p.Enabled);
```
r-tome, line 1461: And here only check this condition if the confirmed user type is not owner or admin

vincentsalucci:
> The old approach read the database once whereas this new approach reads multiple times. Also, correct me if I'm wrong, but we need to check the active policies for the accepting Org, not any other Orgs the user belongs to.

While I agree with this, and actually implemented your alternate approach **first**: I went back and reviewed the `accept` flows and this is how we handle the policy checks there. I thought it was better to align the patterns throughout the user status flows instead of implementing a one-off solution where if (for some reason) in the future we change the exclusion strategy it has to be updated in multiple places and could be skipped where this becomes an issue (again).

When it comes to the single org policy itself, we have to check not only the organization you're trying to join, but also if you're apart of other orgs where that policy applies to you.

r-tome (approved):
@vincentsalucci thanks for explaining that, looks good to me then. Makes sense to reuse the existing logic on `PolicyService.GetPoliciesApplicableToUserAsync`

PR #4097, opened 2024-05-17
[AC-2604] Fix aggregation of CollectionGroup permissions

eliykat:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Fix the following bug: when a user is in multiple groups, and those groups have varying levels of access to the same collection (including at least 1 group with Can Manage access), the user does not always receive Can Manage access to that collection.

This is a bug that @shane-melton picked up earlier when implementing #3793. For the CollectionGroup permissions `HidePasswords` and `ReadOnly`, we resolve multiple groups by taking the `MIN` value - representing the most generous permissions for the user (i.e. `0`, indicating that they can view passwords and edit). However, we also applied this to `Manage`, which gives them the _least_ generous permissions (`0`, indicating that they cannot manage).

Change `MIN([Manage])` to `MAX([Manage])` so that if the user has Manage permissions from 1 group, that will override the others.

## Code changes

## Before you submit


eliykat:
As this is a db change, let's merge after `rc` is cut on Monday.

shane-melton (approved):
Thanks for fixing these!

rkac-bw (approved):
lgtm

PR #4109, opened 2024-05-21
[AC-2521] Remove FlexibleCollectionsSignUp feature flag

eliykat:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Remove the `FlexibleCollectionsSignUp` feature flag.

This was only ever used server-side to set `Organization` properties on creation, so it doesn't need to be persisted for any clients.

## Code changes

* set `Organization.FlexibleCollections` to `true`
* set `Organization.LimitCollectionCreationDeletion` to `false` (thereby enabling new FC behaviour) - this is done by actually removing the explicit assignment, which defaults to `false`
* remove references from test code, including an old test which will now never run. Update existing test to include this assertion

## Before you submit

### DIFF
--- a/src/Admin/AdminConsole/Models/OrganizationEditModel.cs
+++ b/src/Admin/AdminConsole/Models/OrganizationEditModel.cs
@@ -162,7 +162,7 @@ public class OrganizationEditModel : OrganizationViewModel
                 { "baseServiceAccount", p.SecretsManager.BaseServiceAccount }
             });
 
-    public Organization CreateOrganization(Provider provider, bool flexibleCollectionsSignupEnabled, bool flexibleCollectionsV1Enabled)
+    public Organization CreateOrganization(Provider provider, bool flexibleCollectionsV1Enabled)
     {
         BillingEmail = provider.BillingEmail;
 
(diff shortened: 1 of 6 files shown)


src/Admin/AdminConsole/Models/OrganizationEditModel.cs
```diff
@@ -162,19 +162,18 @@ public OrganizationEditModel(Provider provider)
...
             });
 
-    public Organization CreateOrganization(Provider provider, bool flexibleCollectionsSignupEnabled, bool flexibleCollectionsV1Enabled)
+    public Organization CreateOrganization(Provider provider, bool flexibleCollectionsV1Enabled)
     {
         BillingEmail = provider.BillingEmail;
 
         var newOrg = new Organization
         {
             // This feature flag indicates that new organizations should be automatically onboarded to
```
vincentsalucci, line 171: Suggestion: give these comments an update to match the new actions (or lack thereof)

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -575,12 +568,11 @@ public async Task<(Organization organization, OrganizationUser organizationUser,
 
             // This feature flag indicates that new organizations should be automatically onboarded to
```
vincentsalucci, line 569: Same as above.

src/Admin/AdminConsole/Models/OrganizationEditModel.cs
eliykat: I'm not sure exactly what you had in mind here but I updated the comments to be more specific, let me know what you think.

I also considered making the `Organization.FlexibleCollections` property read-only and always defaulting to `true`, but it's still used by the migration code which hasn't been deleted yet, so I maintained the current approach for now.

vincentsalucci (approved):
Thanks for the comment updates (and sorry for not providing more direction) I was just looking for them to be current with the code they're reflecting since the MVP features have been fully realeased, etc.

PR #4071, opened 2024-05-09
[PM-5157] Update duo metadata

ike-kottlowski:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Rename Duo properties to match Duo version 4 updates.

⭐ _Side Quest_: Added entity Framework migration for `Providers.GatewayType`.

## Code changes

* Various files have been changed adding `ClientId` and `ClientSecret` to duo responses and requests.
  * v2 changes will be removed once the database is in a stable state to handle Duo SDK v4 params.
* Added Entity Framework migrations for `Provider.GatewayType`

## Before you submit

- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- If making database changes - make sure you also update Entity Framework queries and/or migrations
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #4126, opened 2024-05-24
[PM-5216] User and Organization Duo Request and Response Model refactor

ike-kottlowski:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Adding support for both Duo SDK v2 and Duo SDK v4.

Keep in mind that this is a transitional phase for the EDD and backwards support. Once this is release into the wild, I will begin work on removing the Duo SDK v2 metadata and targeting only the Duo SDK v4 metadata.

I tried to test the `DuoApi` model, but it was too much of a hassle since we are removing it in the future anyway. I will include tests for the `TwoFactorController` when we make the changes for only supporting Duo SDK v4.

Also EF addition of Gateway Type Column rename for Admin Console. It was already there So I thought I'd leave it.

## Code changes

* **src/Api/Auth/Controllers/TwoFactorController.cs:** Added support for `ClientId` and `ClientSecret` when validating Duo Config; Minor refactoring.
* **src/Api/Auth/Models/Request/TwoFactorRequestModels.cs:** Added support for `ClientId` and `ClientSecret` while maintaining `SKey` and `IKey` for backwards compatibility.
* **src/Api/Auth/Models/Response/TwoFactor/TwoFactorDuoResponseModel.cs:** Added support for `ClientId` and `ClientSecret` while maintaining `SKey` and `IKey` for backwards compatibility.
* **src/Core/Auth/Identity/TemporaryDuoWebV4SDKService.cs:**: Changed to target `ClientId` and `ClientSecret` when validating metadata.
  * This class will become the new DuoTokenProvider.
* **src/Identity/IdentityServer/BaseRequestValidator.cs:** We no longer support Duo SDK V2, so we are only returning `AuthUrl` for Duo Universal Prompt.
* **utils/*:** directory: This is to bring Entity Framework implementations up to date with current schema. This is a holdover from previous iteration of this same task.

## Before you submit


ike-kottlowski:
This PR replaces and addresses concerns raised in [this PR](https://github.com/bitwarden/server/pull/4071)

test/Api.Test/Auth/Models/Request/TwoFactorDuoRequestModelValidationTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            Host = "api-12345abc.duosecurity.com",
+            ClientId = "clientId",
+            ClientSecret = "clientSecret",
+        };
+
+        // Act
+        var result = model.Validate(new ValidationContext(model));
+
+        // Assert
+        Assert.NotNull(result);
```
JaredSnider-Bitwarden, line 65: ⚠️ asserting not null doesn't confirm that there are no validation errors. We should `Assert.Empty(result)` instead (I think)

JaredSnider-Bitwarden (approved):
Nice work!

test/Api.Test/Auth/Models/Request/TwoFactorDuoRequestModelValidationTests.cs
ike-kottlowski: Good catch, if there was a validation error then the return value would be `notnull` as well.

rkac-bw (approved):
LGTM
the migrations are to keep in sync with prod that already has the migration applied

PR #4144, opened 2024-05-30
[PM-5963] Fix tde offboarding vault corruption

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-5963
https://github.com/bitwarden/clients/pull/9480

## 📔 Objective

Fixes tde to master password flow. When switching from tde to MP, the server instantly stops sending the decryption options for TDE, leaving the client no way to get the user key. This PR fixes that by checking, whether a user has no masterpassword *but a trusted device*. In this case it sends the trusted device info along with a flag notifying the client it has to set a password due to TDE offboarding. Updating the MP uses a new route, since neither a normal mp change (requires old mp), or temp mp (would send an unrelated email about the mp change) fit. This change is event logged as a new tde offboarding mp event.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
```diff
@@ -130,6 +133,12 @@ private async Task BuildTrustedDeviceOptions()
             hasManageResetPasswordPermission = await _currentContext.ManageResetPassword(_ssoConfig!.OrganizationId);
         }
 
+        // If tde was disabled the user can reset master pw
+        if (_user != null && !_user.HasMasterPassword() && _device != null && _device.IsTrusted())
+        {
+            hasManageResetPasswordPermission = true;
```
jlf0dev, line 139: We may want a different flag. The client needs to communicate this flow differently from the TDE offboarding one.

quexten: Yeah, that makes sense. It does seem to work correctly this way though, so at least the flow/setting the keys seems correct. Just need to make it more robust in the code.

quexten: Added "isTdeOffboarding" in 89979ad6eaee1304dc8a7a276d61945b8e7490f9. Needs client support aswell, but I think this should separate the cases clearly.

quexten: Also event logging this change now.

src/Core/Services/Implementations/UserService.cs
```diff
@@ -827,6 +827,31 @@ public async Task<IdentityResult> UpdateTempPasswordAsync(User user, string newM
         return IdentityResult.Success;
     }
 
+    public async Task<IdentityResult> UpdateTdeOffboardingPasswordAsync(User user, string newMasterPassword, string key, string hint)
```
jlf0dev, line 830: This should probably be [its own command](https://contributing.bitwarden.com/architecture/server/)

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -939,6 +939,29 @@ public async Task PutUpdateTempPasswordAsync([FromBody] UpdateTempPasswordReques
         throw new BadRequestException(ModelState);
     }
 
+    [HttpPut("update-tde-offboarding-password")]
+    public async Task PutUpdateTdePasswordAsync([FromBody] UpdateTdeOffboardingPasswordRequestModel model)
```
jlf0dev, line 943: I think we need some validation here to prevent misuse of this endpoint. Like making sure that this request is coming from the intended device and that the organization actually is offboarding from TDE.

src/Api/Auth/Models/Request/Accounts/UpdateTdeOffboardingPasswordRequestModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Api.Models.Request.Organizations;
+
+namespace Bit.Api.Auth.Models.Request.Accounts;
+
+public class UpdateTdeOffboardingPasswordRequestModel : OrganizationUserResetPasswordRequestModel
```
jlf0dev, line 6: I don't really like us extending an Admin Console model for this, can we just create our own?

jlf0dev (commented):
Awesome work! Really excited to see this being done and thanks for thinking of the Events! Got some feedback for you.

src/Api/Auth/Controllers/AccountsController.cs
quexten: Added validation and also unit tests for at least some of the validation. I'm not sure if all of it is required.

src/Core/Services/Implementations/UserService.cs
quexten: Done

src/Api/Auth/Models/Request/Accounts/UpdateTdeOffboardingPasswordRequestModel.cs
quexten: Agreed, I pulled in the properities from the `OrganizationUserResetPasswordRequestModel`, so now there is no coupling to it anymore.

src/Core/Auth/UserFeatures/TdeOffboardingPassword/TdeOffboardingPasswordCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        _userService = userService;
+        _userRepository = userRepository;
+        _eventService = eventService;
+        _organizationUserRepository = organizationUserRepository;
+        _ssoUserRepository = ssoUserRepository;
+        _ssoConfigRepository = ssoConfigRepository;
+        _pushService = pushService;
+    }
+
+    public async Task<IdentityResult> UpdateTdeOffboardingPasswordAsync(User user, string newMasterPassword, string key, string hint)
```
jlf0dev, line 41: Really appreciate the validation here 👏 . Can we add a check for `newMasterPassword` and `key`? I know those are required in the request model, but this command can technically be called outside of that.

test/Core.Test/Auth/UserFeatures/TdeOffboarding/TdeOffboardingPasswordCommandTests.cs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+
+    [Theory]
+    [BitAutoData]
+    public async Task TdeOffboardingPasswordCommand_RejectWithMasterPassword(SutProvider<TdeOffboardingPasswordCommand> sutProvider,
+        User user, string masterPassword, string key, string hint)
+    {
+        await Assert.ThrowsAsync<BadRequestException>(() => sutProvider.Sut.UpdateTdeOffboardingPasswordAsync(user, masterPassword, key, hint));
+    }
```
jlf0dev, line 96: question: I'm not sure I understand what this is testing, or how its throwing. Can you add some clarity on this one?

src/Core/Auth/UserFeatures/TdeOffboardingPassword/Interfaces/ITdeOffboardingPasswordCommand.cs
```diff
@@ -0,0 +1,19 @@
...
+
+/// <summary>
+/// <para>Manages the setting of the initial master password for a <see cref="User"/> in an organization.</para>
+/// <para>This class is primarily invoked in two scenarios:</para>
+/// <para>1) In organizations configured with Single Sign-On (SSO) and master password decryption:
+/// just in time (JIT) provisioned users logging in via SSO are required to set a master password.</para>
+/// <para>2) In organizations configured with SSO and trusted devices decryption:
+/// Users who are upgraded to have admin account recovery permissions must set a master password
+/// to ensure their ability to reset other users' accounts.</para>
+/// </summary>
```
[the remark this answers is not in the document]
quexten: Hah, this is left from when I copied the setinitialmasterpasswordcommand to start this class from... It has nothing to do with TDEOffboarding https://github.com/bitwarden/server/blob/a60f70dde5d44de08cadc923f394fb575cf56b1d/src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/ISetInitialMasterPasswordCommand.cs#L6-L14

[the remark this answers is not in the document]
quexten: 5b0224f

src/Core/Auth/UserFeatures/TdeOffboardingPassword/TdeOffboardingPasswordCommand.cs
quexten: 14d0176

test/Core.Test/Auth/UserFeatures/TdeOffboarding/TdeOffboardingPasswordCommandTests.cs
quexten: Added a comment explaining this in e58b4c4. The idea is that TDE offboarding should only be possible when the user is actually a passwordless TDE user, that is, has no master password set. If there is a MP set, then the request should be rejected.

jlf0dev (approved):

src/Api/Auth/Models/Request/Accounts/UpdateTdeOffboardingPasswordRequestModel.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Bit.Api.Auth.Models.Request.Accounts;
+
+public class UpdateTdeOffboardingPasswordRequestModel
+{
+    [Required]
+    [StringLength(300)]
+    public string NewMasterPasswordHash { get; set; }
+    [Required]
+    public string Key { get; set; }
```
jlf0dev, line 11: nit: I try to avoid naming things `Key` since we have a lot of keys 😁

jlf0dev (approved):

src/Api/Auth/Models/Request/Accounts/UpdateTdeOffboardingPasswordRequestModel.cs
quexten: I generally agree, but in this case update-password and update-temp-password also use "key" so I'd rather have consistency between them and rename it only in tde offboarding.

PR #4154, opened 2024-06-03
Add a master password hash check to account recovery enrollment

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/AC-2465
https://github.com/bitwarden/clients/pull/8770

## 📔 Objective

This PR adds a master password hash check to the endpoint used to enroll in account recovery. It pairs with the client PR attached above.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -456,6 +456,11 @@ public class OrganizationUsersController : Controller
             throw new UnauthorizedAccessException();
         }
 
+        if (!string.IsNullOrWhiteSpace(model.ResetPasswordKey) && !await _userService.VerifySecretAsync(user, model.Secret))
+        {
+            throw new BadRequestException("Incorrect password");
+        }
+
         var callingUserId = user.Id;
         await _organizationService.UpdateUserResetPasswordEnrollmentAsync(
             orgId, userId, model.ResetPasswordKey, callingUserId);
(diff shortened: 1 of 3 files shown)


vincentsalucci (approved):

eliykat (commented):
I didn't have time to investigate it today, but what endpoint is used by auto-enrollment (on accepting an org invite)? Will it still work after these changes?

This otherwise looks good to me.

addisonbeck:
Thanks for bringing this up again, I forgot we were concerned about this. I checked it out and this actually isn't an issue. Auto-enrollment actually happens in the `accept` endpoint [here](https://github.com/bitwarden/server/blob/fa4dc4aaf2f6b47cd4c6adc39ada24a0d6195bf3/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs#L281-L305).

I checked and everything still works fine. Here's a demo: 

https://github.com/bitwarden/server/assets/15897251/da38d8b5-65af-4863-ab5d-2387b5d339bd


This endpoint *also* doesn't have any kind of explicit password hash checking. Maybe we care about that? We could pass it along from the account creation/login flows as we discussed. What do you think @eliykat?

Regardless, this shouldn't be be a blocking concern for this work. I'm going to go ahead and send this to QA.

eliykat:
@addisonbeck thanks for confirming that, I agree that other endpoint is non-blocking for this PR.

The accept user flow is authenticated via the email token already (a different secret), so I think that's OK. What do _you_ think? 😁 if you're satisfied with it then I think we're done with password reset.

addisonbeck:
> The accept user flow is authenticated via the email token already (a different secret), so I think that's OK. What do you think? 

Sorry, I missed this, but I do agree.

PR #4157, opened 2024-06-05
[PM-4371] Implement PRF key rotation

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-4371
https://github.com/bitwarden/clients/pull/9517

## 📔 Objective

Implement key rotation for PRF keys. Fixes prf login causing a temporary, local, vault corruption. For this, we extend the webauthncredentialresponsemodel to include the encrypted public and user key, because the client needs these to rotate the keysets. The private key is not rotated (because it is encrypted with the inaccessible prf key from each registered webauthn prf credential).

Further, we extend the regular key rotation command with the webauthn credentials, containing exactly the new userkey encrypted public key, and the new public key encrypted user key.

A validator and unit tests are included.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


ike-kottlowski (approved):

src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
```diff
@@ -99,6 +102,17 @@ public async Task<IdentityResult> RotateUserKeyAsync(User user, RotateUserKeyDat
                     _organizationUserRepository.UpdateForKeyRotation(user.Id, model.OrganizationUsers));
             }
 
+            if (model.WebAuthnKeys.Any())
+            {
+                foreach (var webAuthnCredential in model.WebAuthnKeys)
```
ike-kottlowski, line 107: 🎨 : I think this should be changed to use the `delegate` convention as above. It speeds up the database query since it sends a list.

You'll need to update the `ICredentialRepository` to include 
`UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(UserId id, IEnumerable<WebAuthnCredential> credential)`. 

[Interface example](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs#L53)

[Implementation example](https://github.com/bitwarden/server/blob/97b3f3e7eeb0cfb9d94ba04b54b451dac791a8e1/src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationUserRepository.cs#L541)

CC: @jlf0dev

ike-kottlowski (changes requested):
I think the call to update the keys in the command needs to be modified.

src/Core/Auth/UserFeatures/UserKey/Implementations/RotateUserKeyCommand.cs
quexten: Also, do we in general collect numbers on the performance of specific routes? I'd imagine user key rotation is comparatively rather rare, so it both shouldn't affect the database load much, nor should it be a problem if it takes a few milliseconds more for the user. (I still plan to make this change, trying to establish the process around optimizations though)

ike-kottlowski: I think our goal is to use [Open Telemetry ](https://contributing.bitwarden.com/architecture/adr/observability-with-opentelemetry) but I don't know at what level it's implemented.

quexten: Need to test the queries a bit more on my end, but I think this is implemented correctly now.

src/Sql/dbo/User Defined Types/WebauthnCredentialRotationDataType.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TYPE [dbo].[WebauthnCredentialRotationDataType] AS TABLE(
```
withinfocus, line 1: @rkac-bw can help work with you on this, but we'd like to stop using custom SQL types.

quexten: So, something more like is done for ciphers?:
https://github.com/bitwarden/server/blob/a1d609b208f37f45bd8cf8d7c22769cd0455b281/src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs#L327-L381

withinfocus: Doesn't even need to be that advanced with bulk copy. There's a JSON input that can be used. Robert will have a recent example.

quexten: Adjusted the implementation based on Robert's example. Thanks!

rkac-bw (approved):
LGTM

ike-kottlowski (dismissed):

jlf0dev (dismissed):
Fantastic work! Thank you grabbing this ticket 👏 

I want to mention a minor nit that I would love to see if you have time. We're currently trying to separate the concepts of `WebauthnTwoFactor` and `WebauthnLogin`, we even have a [tech debt ticket](https://bitwarden.atlassian.net/browse/PM-4172?atlOrigin=eyJpIjoiNGQzODI5NWQyODZjNDAyYmI3OWFkYjk4MzY2YWNkMDYiLCJwIjoiaiJ9) to eventually go through the server and address it. If possible, it would be great to have these also renamed as well. Definitely not blocking though!

quexten:
@jlf0dev Renamed the webauthn named types to WebAuthnLogin in the latest commit. (No changes otherwise).

jlf0dev (approved):
Thanks!

jlf0dev (changes requested):
Whoops, some tests are failing

rkac-bw (approved):
lgtm

jlf0dev (approved):

PR #4178, opened 2024-06-12
[AC-1943] Implement provider client invoice report

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/AC-1943

## 📔 Objective

This PR does the following:
- Updates the `ProviderInvoiceItem` SQL objects to fix some missed business logic. Despite these objects not being used in Production yet, I've tried to make the migration backwards compatible and added comments to explain why the changes were necessary. @bitwarden/dept-dbops  Let me know if you see anything funky there. 
- Updates our handler for the `invoice.created` Stripe event to now record `ProviderInvoiceItems` if the event is related to a Provider.
- Adds a handler for the `invoice.finalized` Stripe event to update the `ProviderInvoiceItems` created in the step above with the newly generated invoice number.
- Adds a method to the `ProviderBillingController` that, under the hood, uses [CsvHelper](https://joshclose.github.io/CsvHelper/), to generate the export necessary for download on the client.

Associated `clients` PR: https://github.com/bitwarden/clients/pull/9520

## 📸 Screenshots

Provider creation process and invoice finalization:

https://github.com/bitwarden/server/assets/144709477/d71f9db3-504d-44e7-8bc3-5a802805bef8

Downloading a provider client invoice export for a provider with multiple clients:

https://github.com/bitwarden/server/assets/144709477/20134d73-f6de-4a0c-86fb-320f97478f41

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Sql/Billing/Stored Procedures/ProviderInvoiceItem_Create.sql
+++ b/src/Sql/Billing/Stored Procedures/ProviderInvoiceItem_Create.sql
@@ -7,7 +7,8 @@ CREATE PROCEDURE [dbo].[ProviderInvoiceItem_Create]
     @PlanName NVARCHAR (50),
     @AssignedSeats INT,
     @UsedSeats INT,
-    @Total MONEY
+    @Total MONEY,
+    @Created DATETIME2 (7)
 AS
 BEGIN
     SET NOCOUNT ON
@@ -36,6 +37,6 @@ BEGIN
         @AssignedSeats,
         @UsedSeats,
         @Total,
-        GETUTCDATE()
+        @Created
     )
 END
--- a/src/Sql/Billing/Stored Procedures/ProviderInvoiceItem_Create.sql
+++ b/src/Sql/Billing/Stored Procedures/ProviderInvoiceItem_Create.sql
@@ -8,10 +8,12 @@ CREATE PROCEDURE [dbo].[ProviderInvoiceItem_Create]
     @AssignedSeats INT,
     @UsedSeats INT,
     @Total MONEY,
-    @Created DATETIME2 (7)
+    @Created DATETIME2 (7) = NULL
 AS
 BEGIN
     SET NOCOUNT ON
+
+    SET @Created = COALESCE(@Created, GETUTCDATE())
 
     INSERT INTO [dbo].[ProviderInvoiceItem]
     (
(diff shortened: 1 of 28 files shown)


src/Sql/Billing/Stored Procedures/ProviderInvoiceItem_Create.sql
```diff
@@ -7,7 +7,8 @@ CREATE PROCEDURE [dbo].[ProviderInvoiceItem_Create]
     @PlanName NVARCHAR (50),
     @AssignedSeats INT,
     @UsedSeats INT,
-    @Total MONEY
+    @Total MONEY,
+    @Created DATETIME2 (7)
```
rkac-bw, line 11: This would be a breaking change with EDD as we would not be able to deploy this change independent of the code change, trying to call this stored procedure without the extra parameter. The pattern so far has been to create a new version of the procedure and call that with the new release of the code. With a blue green release this would allow the app servers blue green versions to work against the single db. @withinfocus thoughts?

amorask-bitwarden: @rkac-bw That makes sense. The strange thing about this one is that, in the current state, there actually is no code invoking this SPROC. All of that code is added as part of this PR, so I thought I might be able to just re-create the SPROC as is. That being said, I did forget that the code in this PR would be deployed independently of the migration so if something goes South with the DB migration, the code here would be invoking the old version of the SPROC. I can update these to use V2s.

withinfocus: @rkac-bw that's indeed a solution, albeit more advanced. We can just add a `= NULL` and this would be fine.

rkac-bw: Much cleaner and easier, could even have default value of GETUTCDATE(), either is good by me

amorask-bitwarden: Updated to make `@Created` default to `NULL` for both the updated "Create" and "Update" stored procedures. You can't default a SP parameter to `GETUTCDATE`, so I set it with `COALESCE` in the SP body.

https://github.com/bitwarden/server/pull/4178/commits/b7704dfddda36f6e344daca3ee7bde631fc2d21a

rkac-bw (approved):
lgtm

cturnbull-bitwarden (approved):
🎉

PR #4282, opened 2024-06-26
[AC-2809] Remove unused FlexibleCollections feature flag from Cipher Repository

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/AC-2809
## 📔 Objective

**In short:** remove the unused `FlexibleCollections` feature flag from the CipherRepository, along with v2 sprocs that were never used and are no longer required.

**Longer:** `OrganizationUser.AccessAll` and `Group.AccessAll` were flags that gave orgUsers access to all organization items. They were deprecated as part of Flexible Collections, however this was a challenge because the flags are evaluated at the database level. The initial approach was to create v2 versions of all sprocs that used `AccessAll`, and then use a feature flag to switch between the old & new sprocs. This was not workable and was abandoned in favour of a hard data migration - i.e. setting `AccessAll` to `0` and then removing the logic later. The result is that we still have this feature flag being passed around everywhere even though it was never turned on, and v2 sprocs that were never used.

This PR starts removing this tech debt by removing the feature flag code that was intended to switch between the sprocs. After 1 release, the v2 sprocs will be dropped altogether.

Because we ended up migrating the data first, we don't actually need to update these sprocs at all: they all call `UserCipherDetails` which can be updated in place (without copying/versioning). It was only the name change rippling outwards that required these sprocs to be updated.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


ike-kottlowski (approved):
auth changes are good

src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
```diff
@@ -303,12 +303,12 @@ public async Task DeleteDeletedAsync(DateTime deletedDateBefore)
...
     }
 
-    public async Task<CipherDetails> GetByIdAsync(Guid id, Guid userId, bool useFlexibleCollections)
+    public async Task<CipherDetails> GetByIdAsync(Guid id, Guid userId)
     {
         using (var scope = ServiceScopeFactory.CreateScope())
         {
             var dbContext = GetDatabaseContext(scope);
-            var userCipherDetails = new UserCipherDetailsQuery(userId, useFlexibleCollections);
+            var userCipherDetails = new UserCipherDetailsQuery(userId, false);
```
vincentsalucci, line 311: 💭 Just wanted to double check that you wanted to pass in the `false` value here. Initial reaction is we should be passing `true` but if we've already updated the underlying `vNext vs vCurrent` then `false` makes sense.

eliykat: It's intentional - because this flag was never used, this _never_ evaluated to `true`, and these v2 sprocs were never used.

I haven't removed this parameter completely because it's used by other callers that I haven't updated yet.

More detail is in the [tech breakdown](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/900825246/Tech+Breakdown+Flexible+Collections+MVP+cleanup#Remove-AccessAll-from-the-database). These sprocs fall into the "third category" listed in that section. Let's sync up on this soon.

gbubemismith (approved):
Vault changes look good

src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
vincentsalucci: Makes sense! Thanks for explaining.

vincentsalucci (approved):

PR #4388, opened 2024-06-26
[PM-2944] Make Entities Nullable On Unowned Types

justindbaur:
## 📔 Objective

Enable [nullable reference types feature](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) on core unowned entities. This is a foundational step towards enabling it project wide and an ensuring we have accurate request/response models.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Entities/Collection.cs
+++ b/src/Core/Entities/Collection.cs
@@ -1,15 +1,17 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using Bit.Core.Utilities;
 
+#nullable enable
+
 namespace Bit.Core.Entities;
 
 public class Collection : ITableObject<Guid>
 {
     public Guid Id { get; set; }
     public Guid OrganizationId { get; set; }
-    public string Name { get; set; }
+    public string Name { get; set; } = null!;
     [MaxLength(300)]
-    public string ExternalId { get; set; }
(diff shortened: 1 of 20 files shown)


src/Core/Entities/Collection.cs
```diff
@@ -1,15 +1,17 @@
...
+#nullable enable
+
 namespace Bit.Core.Entities;
 
 public class Collection : ITableObject<Guid>
 {
     public Guid Id { get; set; }
     public Guid OrganizationId { get; set; }
-    public string Name { get; set; }
+    public string Name { get; set; } = null!;
```
coroiu, line 12: question: Should we really be doing this when enabling nullabe? Seems like the whole point of having strict null checks?

justindbaur: It's a _good_ solution, we have to promise to the compiler that when this is created through reflection, like how dapper and EF Core do it, that it will be given a non-null value. 

The better solution is to have the `required` keyword put onto the property but unfortunately that causes some errors that I would need to resolve (mostly in tests) but with trying to keep this series of PR's small and only targeting a single team each it wasn't possible right now. I definitely encourage teams to use that going forward.

coroiu (approved):

PR #4389, opened 2024-06-26
[PM-2944] Enable Nullable For Secrets Manager

justindbaur:
## 📔 Objective

Enable [nullable reference types feature](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) on core Secrets Manager entities. This is a foundational step towards enabling it project wide and an ensuring we have accurate request/response models.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/SecretsManager/Entities/ApiKey.cs
```diff
@@ -2,22 +2,24 @@
...
+
 namespace Bit.Core.SecretsManager.Entities;
 
 public class ApiKey : ITableObject<Guid>
 {
     public Guid Id { get; set; }
     public Guid? ServiceAccountId { get; set; }
     [MaxLength(200)]
-    public string Name { get; set; }
+    public string Name { get; set; } = null!;
```
Thomas-Avery, line 14: It seems weird initializing a non-nullable value type as null. 

Have we thought about using the `required` modifier or initializing as an empty string?

Thomas-Avery (changes requested):
It looks like this generates EF migrations to alter the columns to `nullable: false`.

Should these EF migrations be included in this PR?

src/Core/SecretsManager/Entities/ApiKey.cs
justindbaur: It's a pretty common scenario to tell the compiler you know something that it doesn't, in this case that this type is often created through reflection and from data that is guaranteed in another way to be non-null. 

I'm definitely all for using `required` or even constructor parameter which both Dapper and EF will support but unfortunately that caused build errors (mostly in test code) that I felt was out of scope for now. 

I also prefer `null!` to `""` for two reasons. One, the JIT compiler knows to no-op it so it should be faster and two, because it could cover up a problem even more so. You could try and create a `new ApiKey()` and you were supposed to set the name but missed it. Now you might try to do `apiKey.Name.Trim()` and it would work with you not knowing as early that you missed assigning it an actual value. Or since it's not-null it would save into the database correctly. If it was left as null then it would throw an error.

justindbaur:
@Thomas-Avery 

> It looks like this generates EF migrations to alter the columns to nullable: false.
> 
> Should these EF migrations be included in this PR?

My plan is to try and save up the approval on all these PR's, merge them in near unison, and add all the migrations in a single PR. That way I can have only single team PR's.

src/Core/SecretsManager/Entities/ApiKey.cs
Thomas-Avery: Thank you for the detailed explanation!

I prefer `required` but I'm good with moving forward this way to get to enabling this feature project wide.

Thomas-Avery (approved):
LGTM

justindbaur:
@coltonhurst I wasn't going to initially and it could be a team thing to improve it but since you all are all using the integration tests this change started breaking the `NOT NULL` constraints in the database for the tests that weren't setting it to a value. I went ahead and switched these to use `required` and updated the tests.

test/Api.IntegrationTest/SecretsManager/Controllers/ServiceAccountsControllerTests.cs
```diff
@@ -943,6 +929,22 @@ private async Task<(Guid OrganizationId, List<Guid> ServiceAccountIds)> SetupLis
         return (org.Id, serviceAccountIds);
     }
 
+    private static ApiKey CreateTestApiKey(Guid serviceAccountId, string name = "Test Name", DateTime? expiresAt = null)
+    {
+        return new ApiKey
+        {
+            ServiceAccountId = serviceAccountId,
+            Name = name,
+            ExpireAt = expiresAt ?? DateTime.MinValue,
```
Thomas-Avery, line 938:
```suggestion
            ExpireAt = expiresAt,
```

Setting `ExpireAt` to null is valid and is how we set non expiring API keys. 

Having `DateTime.MinValue` would be creating the ApiKey where it is already expired, which I don't think we would want that as the default for tests.

Thomas-Avery (changes requested):
Looks good, thank you!

One minor suggestion and should be good to go for me.

test/Api.IntegrationTest/SecretsManager/Controllers/ServiceAccountsControllerTests.cs
justindbaur: Didn't even look to see that it was nullable. Good catch, fixed!

test/Api.IntegrationTest/SecretsManager/Controllers/ServiceAccountsControllerTests.cs
```diff
@@ -943,6 +929,22 @@ private async Task<(Guid OrganizationId, List<Guid> ServiceAccountIds)> SetupLis
         return (org.Id, serviceAccountIds);
     }
 
+    private static ApiKey CreateTestApiKey(Guid serviceAccountId, string name = "Test Name", DateTime? expiresAt = null)
```
mzieniukbw, line 932:
```suggestion
    private static ApiKey CreateTestApiKey(Guid serviceAccountId, string name = _mockEncryptedString, DateTime? expiresAt = null)
```
Name field is always encoded, this is better default

test/Api.IntegrationTest/SecretsManager/Controllers/ServiceAccountsControllerTests.cs
```diff
@@ -943,6 +929,22 @@ private async Task<(Guid OrganizationId, List<Guid> ServiceAccountIds)> SetupLis
...
+        return new ApiKey
+        {
+            ServiceAccountId = serviceAccountId,
+            Name = name,
+            ExpireAt = expiresAt,
+            Scope = new JsonArray
+            {
+                "api.secrets",
+            }.ToJsonString(),
+            EncryptedPayload = "Test Payload",
```
mzieniukbw, line 943:
```suggestion
            EncryptedPayload = _mockEncryptedString,
```

test/Api.IntegrationTest/SecretsManager/Controllers/ServiceAccountsControllerTests.cs
```diff
@@ -943,6 +929,22 @@ private async Task<(Guid OrganizationId, List<Guid> ServiceAccountIds)> SetupLis
...
+            Key = "Test Key",
```
mzieniukbw, line 944:
```suggestion
            Key = _mockEncryptedString,
```

src/Core/SecretsManager/Entities/ApiKey.cs
```diff
@@ -2,22 +2,24 @@
 using Bit.Core.Entities;
 using Bit.Core.Utilities;
 
+#nullable enable
```
mzieniukbw, line 5: ⛏️ we usually put it at the top, before `using`
To be honest i don't know why 🤷 , do we want to be more consistent on this one ?

mzieniukbw (approved):

Thomas-Avery (approved):

PR #4439, opened 2024-06-28
[AC-2820] Updated org edit form scripts to dynamically update expected values

cturnbull-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/AC-2820

## 📔 Objective

This PR addresses an issue where organizations created via the Bitwarden Portal for a Reseller Provider were being set to have a `MaxStorageGb` of `0` instead of the default `1`. When investigating this issue, I found the same was true when setting `Model.Seats`, so I addressed this as well.

Instead of hardcoding these values as seen with the other form fields being set, I updated the function not to use a switch-case approach on the plan type and setting hard-coded values, but instead to get the selected plan type from the `StaticStore`, then set the relevant fields based on the default values on the given plan. Feel free to disagree with this approach if you prefer the former!

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Admin/AdminConsole/Models/OrganizationEditModel.cs
```diff
@@ -179,13 +179,29 @@ public OrganizationEditModel(Provider provider)
      * This is mapped manually below to provide some type safety in case the plan objects change
```
vincentsalucci, line 179: 💭 Does the comment here still hold true even with the changes to how the helper functions?

util/Migrator/DbScripts/2024-07-01_00_EnsureMaxGBAndSeatsAreSet.sql
```diff
@@ -0,0 +1,8 @@
+UPDATE
```
vincentsalucci, line 1: 💭 Just double checking: we don't ever **allow** `null` values for those properties, right? So we can make the assumption (as your migration is doing here) that they should always be default to `1`.

cturnbull-bitwarden: I did some digging with @rkac-bw to make sure that these were valid. Seats being null doesn't really happen in production, so I dropped setting it in the last commit. MaxStorageGb should be defaulted to 1 for all _premium_ organizations instead of all plan types as I originally had here, so made that update as well

src/Admin/AdminConsole/Models/OrganizationEditModel.cs
vincentsalucci: @cturnbull-bitwarden - non-blocking comment above, but still want to address, if possible!

cturnbull-bitwarden: You're absolutely right to bring this up! My intention wasn't to break the spirit of the comment. As it is, this is no longer safe from rename refactors, so I'll go ahead and update this to not use anonymous type properties.

rkac-bw (approved):
lgtm

vincentsalucci (approved):

PR #4466, opened 2024-07-04
Devcontainer Improvements

justindbaur:
## 📔 Objective

This addresses a few of the pain points I have encountered while developing in these containers.

1. Scared to rebuild, since I hadn't mounted the data protections keys to my host system if I were to rebuild the container I would keep my database but I'd lose access to all my users because their master password hash was protected with keys that did not exist anymore.
2. We don't currently run the container with [`docker-outside-of-docker`](https://github.com/devcontainers/features/tree/9387225ac7ada315d94511848fd490a8519ed591/src/docker-outside-of-docker), while I may revisit that in the future, the line of code in `ef_migrate.ps1` never worked on my machine even though I do have a MySql container available. I would always have to comment out that line to get the script to work. This changes it to only run that line when `docker-compose` is available. Then if the user doesn't have a MySql container available it will just error out on that step.
3. I use VSCode `Run Tasks` to do various things like refresh my secrets and install the dev certificate. The `build*` tasks just fill up the UI with things I never use. I don't think anyone uses these directly and they are for all the `run*` commands in `launch.json`. 
4. C# DevKit will occasionally ask me which solution file I want to use and I always want the one in the root of the repo. This changes us to using a traversal project for running our test files with a OSS and non-OSS configuration. The solution files have the problem of being harder to maintain and easier to accidentally not add a new test project to (which I believe we have done in the past). I think the traversal projects are a better fit for what we want regardless of switching to them making it so C# DevKit doesn't offer the choice. [Read more](https://github.com/microsoft/MSBuildSdks/tree/32403937dc2bb90f74c2156b9e140538b3f25d51/src/Traversal).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.vscode/tasks.json
```diff
@@ -192,20 +207,6 @@
...
-            "presentation": {
-                "echo": true,
-                "reveal": "always",
-                "focus": false,
-                "panel": "shared",
-                "showReuseMessage": true,
-                "clear": false
-            },
-            "problemMatcher": "$msCompile"
-        },
```
justindbaur, line 208: This is built into the C# DevKit now.

.vscode/tasks.json
```diff
@@ -226,12 +227,14 @@
         },
         {
             "label": "Setup Secrets",
```
withinfocus, line 229:
```suggestion
            "label": "Set Up Secrets",
```

⛏️ but can't resist. Doesn't need to actually be applied but noun vs. verb mistakes always get me.

withinfocus (approved):
The usage of `Microsoft.Build.Traversal` is pretty neat and I didn't know we had that flexibility with builds. As long as `dotnet test` like we have at https://github.com/bitwarden/server/actions/runs/9800156276/workflow?pr=4466#L40-L44 still works with this then I'm good.

tangowithfoxtrot (approved):
Very nice! I've been a little afraid to rebuild my local setup as well 😅 It's nice to have this handled now.

LGTM!

withinfocus (approved):

PR #4524, opened 2024-07-17
[AC-1698] Check if a user has 2FA enabled more efficiently

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/AC-1698

## 📔 Objective

The members page loads slowly for large organizations because the database calls to check each user's 2FA status are not optimized.

>Current code flow:
>`OrganizationUsersController.Get` has a looped async call to `userService.TwoFactorIsEnabledAsync` (per user): https://github.com/bitwarden/server/blob/master/src/Api/Controllers/OrganizationUsersController.cs#L93-L94
>that calls `CanAccessPremium` if they have a premium 2FA method (e.g. yubikey)
>that calls `hasPremiumFromOrganization`
>that calls the db to get every orgUser for that user
>So for example, if an org has 500 users with yubikey enabled, it’ll hit the database 500 times for the 1 request.
>We need a more efficient way to check this for many orgUsers at once, preferably starting with their common organization.

This PR improves performance by making a single database call to check 2FA status for multiple users at once.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -81,6 +81,137 @@ public async Task<ICollection<Core.Entities.User>> SearchAsync(string email, int
...
+            var dbContext = GetDatabaseContext(scope);
+            List<DataModel.UserDetails> users;
+
+            if (dbContext.Database.IsNpgsql())
+            {
+                users = await GetDbSet(dbContext)
+                    .Where(e => email == null || EF.Functions.ILike(EF.Functions.Collate(e.Email, "default"), $"{email}%"))
+                    .OrderBy(e => e.Email)
+                    .Skip(skip).Take(take)
+                    .Select(e => new DataModel.UserDetails
```
jlf0dev, line 97: question: can we move all this init logic to the model itself?

src/Core/Repositories/IUserRepository.cs
```diff
@@ -13,12 +13,14 @@ public interface IUserRepository : IRepository<User, Guid>
     Task<User?> GetBySsoUserAsync(string externalId, Guid? organizationId);
     Task<UserKdfInformation?> GetKdfInformationByEmailAsync(string email);
     Task<ICollection<User>> SearchAsync(string email, int skip, int take);
+    Task<ICollection<UserDetails>> SearchDetailsAsync(string email, int skip, int take);
```
jlf0dev, line 16: Question: is this used somewhere? I might be missing it

r-tome: Its not! This was leftover from a previous implementation. I'll remove it.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
r-tome: Yes! Good call.

src/Core/Services/IUserService.cs
```diff
@@ -65,6 +65,8 @@ public interface IUserService
     Task<bool> CheckPasswordAsync(User user, string password);
     Task<bool> CanAccessPremium(ITwoFactorProvidersUser user);
     Task<bool> HasPremiumFromOrganization(ITwoFactorProvidersUser user);
+    Task<IEnumerable<(Guid userId, bool twoFactorIsEnabled)>> TwoFactorIsEnabledAsync(IEnumerable<Guid> userIds);
+    Task<IEnumerable<(T user, bool twoFactorIsEnabled)>> TwoFactorIsEnabledAsync<T>(IEnumerable<T> users) where T : ITwoFactorProvidersUser;
```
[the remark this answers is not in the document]
r-tome: I used a generic type here because it allows the method to return the actual type that was passed in as the input. This means that after the method is called, you still have access to all the properties and methods of the original type, not just those defined in the `ITwoFactorProvidersUser` interface. If we only used `ITwoFactorProvidersUser`, we'd be limited to interacting with the properties and methods defined by that interface.

src/Core/Repositories/IUserRepository.cs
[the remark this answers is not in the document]
r-tome: I have added a description for the new method. Would you like docs for all the other methods too?

shane-melton (approved):
Vault change looks good!

jlf0dev (changes requested):
Thanks you for the great documentation! After thinking about it some more, I'm asking for some bigger architecture changes here. I really like this change though!

src/Core/Models/Data/UserDetails.cs
```diff
@@ -0,0 +1,60 @@
+﻿using Bit.Core.Entities;
+
+namespace Bit.Core.Models.Data;
+
+/// <summary>
+/// Represents a user with additional details.
+/// </summary>
+public class UserDetails : User
```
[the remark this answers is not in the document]
r-tome: I was debating whether to be specific on the name or to use a more general name for future flexibility. However, you're right, the calculated premium field is very specific and could be resource-intensive for the database. I'll go with `UserWithCalculatedPremium`.

src/Core/Services/IUserService.cs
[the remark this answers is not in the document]
r-tome: That makes perfect sense, I'll refactor it!

src/Core/Services/Implementations/UserService.cs
```diff
@@ -1369,6 +1369,71 @@ public async Task<bool> VerifySecretAsync(User user, string secret, bool isSetti
...
+                            break;
+                        }
+                        if (userDetail.HasPremiumAccess)
+                        {
+                            hasTwoFactor = true;
+                            break;
+                        }
+                    }
+                }
+            }
```
jlf0dev, line 1400: One other suggestion, just to be a little more functional:
```suggestion

            // Get all enabled providers
            var enabledProviderKeys = from provider in providers
                where provider.Value?.Enabled ?? false
                select provider.Key;

            // Find the first provider that is enabled and passes the premium check
            var hasTwoFactor = enabledProviderKeys
                .Select(type => userDetail.HasPremiumAccess || !TwoFactorProvider.RequiresPremium(type))
                .FirstOrDefault(twoFactorEnabled => twoFactorEnabled);

```

src/Core/Services/IUserService.cs
```diff
@@ -65,18 +65,6 @@ public interface IUserService
...
-    /// <param name="userIds">The list of user IDs to check.</param>
-    Task<IEnumerable<(Guid userId, bool twoFactorIsEnabled)>> TwoFactorIsEnabledAsync(IEnumerable<Guid> userIds);
-    /// <summary>
-    /// Returns a list of users and whether two factor is enabled for each user.
-    /// </summary>
-    /// <param name="users">The list of users to check.</param>
-    /// <typeparam name="T">The type of user in the list. Must implement <see cref="ITwoFactorProvidersUser"/>.</typeparam>
-    /// <returns></returns>
-    Task<IEnumerable<(T user, bool twoFactorIsEnabled)>> TwoFactorIsEnabledAsync<T>(IEnumerable<T> users) where T : ITwoFactorProvidersUser;
     Task<bool> TwoFactorIsEnabledAsync(ITwoFactorProvidersUser user);
```
jlf0dev, line 68: Can you move the logic for this to the `TwoFactorIsEnabledQuery` as well? You can leave the property on the `UserService` so you don't have to change over all the calls, just add a TODO here for us to clean up later.

src/Core/Services/Implementations/UserService.cs
r-tome: That looks great!

src/Core/Services/IUserService.cs
r-tome: I have marked it as obsolete

src/Core/Auth/UserFeatures/TwoFactorAuth/Interfaces/ITwoFactorIsEnabledQuery.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Returns a list of user IDs and whether two factor is enabled for each user.
+    /// </summary>
+    /// <param name="userIds">The list of user IDs to check.</param>
+    Task<IEnumerable<(Guid userId, bool twoFactorIsEnabled)>> TwoFactorIsEnabledAsync(IEnumerable<Guid> userIds);
+    /// <summary>
+    /// Returns a list of users and whether two factor is enabled for each user.
+    /// </summary>
+    /// <param name="users">The list of users to check.</param>
+    /// <typeparam name="T">The type of user in the list. Must implement <see cref="ITwoFactorProvidersUser"/>.</typeparam>
+    /// <returns></returns>
```
jlf0dev, line 17: nit: empty returns here

jlf0dev (dismissed):
Looks great! Thanks for addressing all my feedback!

util/Migrator/DbScripts/2024-08-02_00_UserReadByIdsWithCalculatedPremium.sql
```diff
@@ -0,0 +1,31 @@
+CREATE OR ALTER PROCEDURE [dbo].[User_ReadByIdsWithCalculatedPremium]
+    @Ids AS [dbo].[GuidIdArray] READONLY
+AS
```
rkac-bw, line 3: Not sure if in scope, but we have been moving away from using data types to pass in arrays, now we would pass  in json :

CREATE OR ALTER PROCEDURE [dbo].[User_ReadByIdsWithCalculatedPremium]
    @Ids NVARCHAR(MAX)
AS
BEGIN
    SET NOCOUNT ON;

    -- Declare a table variable to hold the parsed JSON data
    DECLARE @ParsedIds TABLE (Id UNIQUEIDENTIFIER);

    -- Parse the JSON input into the table variable
    INSERT INTO @ParsedIds (Id)
    SELECT value
    FROM OPENJSON(@Ids);

    -- Check if the input table is empty
    IF (SELECT COUNT(1) FROM @ParsedIds) < 1
    BEGIN
        RETURN(-1);
    END

        -- Main query to fetch user details and calculate premium access
        SELECT
            U.*,
            CASE
                WHEN U.[Premium] = 1
                    OR EXISTS (
                        SELECT 1
                        FROM [dbo].[OrganizationUser] OU
                        JOIN [dbo].[Organization] O ON OU.[OrganizationId] = O.[Id]
                        WHERE OU.[UserId] = U.[Id]
                          AND O.[UsersGetPremium] = 1
                          AND O.[Enabled] = 1
                    )
                    THEN 1
                ELSE 0
            END AS HasPremiumAccess
        FROM
            [dbo].[UserView] U
        WHERE
            U.[Id] IN (SELECT [Id] FROM @ParsedIds);
    
END;

r-tome: I've implemented that change. Thanks for providing the updated sproc!

rkac-bw (approved):
lgtm

vincentsalucci (approved):

shane-melton (dismissed):

jlf0dev (approved):

rkac-bw (approved):

jlf0dev (approved):

shane-melton (approved):

patrick-bitwarden (approved):
Approving for AC after conflict resolution

PR #4549, opened 2024-07-22
[PM-2943] Enable Nullable Repositories in Unowned Files

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-2943

## 📔 Objective

Enable [nullable annotations](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references#null-state-analysis) in our repository layer. This PR only partially enables it in all files that don't have a team owner. After enabling nullable I then tried to satisfy all the warnings in the least obtrusive way possible. 

This PR does have more warnings (111) than the baseline (64) but they are largely in test files that are owned by a team. I plan to address those in those team specific PR's. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Repositories/ICollectionCipherRepository.cs
+++ b/src/Core/Repositories/ICollectionCipherRepository.cs
@@ -2,6 +2,8 @@
 
 namespace Bit.Core.Repositories;
 
+#nullable enable
+
 public interface ICollectionCipherRepository
 {
     Task<ICollection<CollectionCipher>> GetManyByUserIdAsync(Guid userId);
--- a/src/Core/Repositories/TableStorage/EventRepository.cs
+++ b/src/Core/Repositories/TableStorage/EventRepository.cs
@@ -4,6 +4,8 @@ using Bit.Core.Settings;
 using Bit.Core.Utilities;
 using Bit.Core.Vault.Entities;
 
+#nullable enable
+
 namespace Bit.Core.Repositories.TableStorage;
 
 public class EventRepository : IEventRepository
@@ -78,9 +80,9 @@ public class EventRepository : IEventRepository
         await CreateEventAsync(entity);
     }
 
-    public async Task CreateManyAsync(IEnumerable<IEvent> e)
+    public async Task CreateManyAsync(IEnumerable<IEvent>? e)
     {
-        if (!e?.Any() ?? true)
+        if (e is null || !e.Any())
         {
             return;
         }
@@ -91,7 +93,7 @@ public class EventRepository : IEventRepository
             return;
         }
 
-        var entities = e.Where(ev => ev is EventTableEntity).Select(ev => ev as EventTableEntity);
+        var entities = e.OfType<EventTableEntity>();
         var entityGroups = entities.GroupBy(ent => ent.PartitionKey);
         foreach (var group in entityGroups)
         {
@@ -134,7 +136,7 @@ public class EventRepository : IEventRepository
         var result = new PagedResult<IEvent>();
         var query = _tableClient.QueryAsync<AzureEvent>(filter, pageOptions.PageSize);
 
-        await using (var enumerator = query.AsPages(pageOptions?.ContinuationToken,
+        await using (var enumerator = query.AsPages(pageOptions.ContinuationToken,
             pageOptions.PageSize).GetAsyncEnumerator())
         {
             await enumerator.MoveNextAsync();
--- a/src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
+++ b/src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
@@ -7,6 +7,8 @@ using Bit.Core.Settings;
 using Dapper;
 using Microsoft.Data.SqlClient;
 
+#nullable enable
+
 namespace Bit.Infrastructure.Dapper.Repositories;
 
 public class CollectionRepository : Repository<Collection, Guid>, ICollectionRepository
@@ -32,7 +34,7 @@ public class CollectionRepository : Repository<Collection, Guid>, ICollectionRep
         }
     }
 
-    public async Task<Tuple<Collection, CollectionAccessDetails>> GetByIdWithAccessAsync(Guid id)
+    public async Task<Tuple<Collection?, CollectionAccessDetails>> GetByIdWithAccessAsync(Guid id)
     {
         using (var connection = new SqlConnection(ConnectionString))
         {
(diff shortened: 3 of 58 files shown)


src/Core/Repositories/TableStorage/EventRepository.cs
```diff
@@ -78,9 +80,9 @@ public async Task CreateAsync(IEvent e)
         await CreateEventAsync(entity);
     }
 
-    public async Task CreateManyAsync(IEnumerable<IEvent> e)
+    public async Task CreateManyAsync(IEnumerable<IEvent>? e)
     {
-        if (!e?.Any() ?? true)
+        if (e is null || !e.Any())
```
justindbaur, line 85: The old pattern confuses null state analysis and makes it think that e is possibly null even though it shouldn't. This way of writing it helps it understand and in my opinion is more readable. We likely don't even want to allow null to be passed in but as this method is currently written it is meant to gracefully handle null so I kept it. When I go through the service layer I may revisit this to disallow a null collection.

src/Infrastructure.EntityFramework/Repositories/TaxRateRepository.cs
```diff
@@ -17,9 +19,9 @@ public async Task ArchiveAsync(Core.Entities.TaxRate model)
         using (var scope = ServiceScopeFactory.CreateScope())
         {
             var dbContext = GetDatabaseContext(scope);
-            var entity = await dbContext.FindAsync<Core.Entities.TaxRate>(model);
-            entity.Active = false;
-            await dbContext.SaveChangesAsync();
+            await dbContext.TaxRates
+                .Where(tr => tr.Id == model.Id)
+                .ExecuteUpdateAsync(property => property.SetProperty(tr => tr.Active, false));
```
justindbaur, line 24: This method would have completely failed before. `model` isn't the key for `TaxRate`, we should have been passing in `model.Id`. If I just fixed that it would have thrown a `NullReferenceException` when calling `entity.Active` and the given tax rate doesn't exist. This new code behaves much more like the [SQL Server one](https://github.com/bitwarden/server/blob/cca358c88caacfbfbb8030580a567dc8d64ee5a1/src/Sql/dbo/Stored%20Procedures/TaxRate_Archive.sql#L7-L12). It doesn't check at all if a tax rate was actually archived.

...also I don't believe this table gets used at all in SH so this doesn't really matter.

src/Core/Repositories/TableStorage/EventRepository.cs
```diff
@@ -91,7 +93,7 @@ public async Task CreateManyAsync(IEnumerable<IEvent> e)
             return;
         }
 
-        var entities = e.Where(ev => ev is EventTableEntity).Select(ev => ev as EventTableEntity);
+        var entities = e.OfType<EventTableEntity>();
```
justindbaur, line 96: `OfType` does the [same thing](https://github.com/dotnet/runtime/blob/ed53f555311c1cff94153c18aceb3f5997904b8e/src/libraries/System.Linq/src/System/Linq/OfType.cs#L50) we were already doing except by only iterating through the list once and maintaining null analysis.

src/Core/Repositories/TableStorage/EventRepository.cs
dani-garcia: I have to agree that it's way more readable like this, the previous way made me pull out pen and paper to evaluate the order of operations lol

src/Infrastructure.EntityFramework/Repositories/DatabaseContextExtensions.cs
```diff
@@ -1,16 +1,20 @@
...
+#nullable enable
+
 namespace Bit.Infrastructure.EntityFramework.Repositories;
 
 public static class DatabaseContextExtensions
 {
     public static async Task UserBumpAccountRevisionDateAsync(this DatabaseContext context, Guid userId)
     {
         var user = await context.Users.FindAsync(userId);
+        Debug.Assert(user is not null);
```
dani-garcia, line 17: Question: I'm curious why we're using `Debug.Assert` here instead of doing `user!.AccountRevisionDate ...` like in some other places, presumably both options would throw an exception if `user` is null right?

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
```diff
@@ -311,7 +315,9 @@ public async Task<ICollection<CollectionAccessSelection>> GetManyUsersByIdAsync(
 
     public class CollectionWithGroupsAndUsers : Collection
     {
-        public DataTable Groups { get; set; }
-        public DataTable Users { get; set; }
+        [DisallowNull]
+        public DataTable? Groups { get; set; }
```
dani-garcia, line 319: Question: It seems a bit surprising to declare the elements as nullable with `?` and then annotate them with `DisallowNull`, what is the difference between this and just marking them non-nullable?

dani-garcia (approved):
Overall looks pretty good to me, just a couple of small questions

src/Core/Repositories/ICollectionCipherRepository.cs
```diff
@@ -2,6 +2,8 @@
 
 namespace Bit.Core.Repositories;
 
+#nullable enable
```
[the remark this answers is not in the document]
justindbaur: Good catch, yes!

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
justindbaur: This one is tough and can really go multiple ways so happy for some input. My reasoning for this one was because these properties are `null` (albeit very briefly) after the `JsonSerializer.Deserialize<CollectionWithGroupsAndUsers>(JsonSerializer.Serialize(obj))!` calls above. So that is why it feels more honest to say they might be null if someone where to attempt to access it. But then `DisallowNull` makes it so that it will still warn if someone would try to assign null to it. Which is seeming undesirable since we do  this `users != null ? users.ToArrayTVP() : Enumerable.Empty<CollectionAccessSelection>().ToArrayTVP();` whole dance instead of doing `users?.ToArrayTVP()`. 

My bigger preference is to probably not even need this class at all and also skip using json to do a clone-y thing and then there wouldn't be a period of time where the object is in a not good state.

src/Infrastructure.EntityFramework/Repositories/DatabaseContextExtensions.cs
justindbaur: Honestly the other place in this PR _could_ probably use `Debug.Assert` as well. I treat `Debug.Assert`'s as my 0 cost way of making assumptions (and I should have used a better error message which I did change here). Basically the user id should be validated to actually exist in the database by some other way. The other one in this PR should probably actually handle null in that case because as far as I can tell it's a bit more possible.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
dani-garcia: Ah that makes sense, the whole deserialize/serialize thing is definitely a bit suboptimal yeah. I agree with you that `DisallowNull` is better than nothing, though.

dani-garcia (approved):
LGTM!

rkac-bw (approved):
lgtm

PR #4599, opened 2024-08-07
[AC-2614] Member Access Report Endpoint

ttalty:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/AC-2614

## 📔 Objective

Added the OrganizationUserUserDetailsQuery which abstracts the logic from the controller into a query for the GetOrganizationUserUserDetails. 

Created the ReportsController and added the member access logic for creating the report. Including permissions for AccessReports. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


LRNcardozoWDF (approved):

src/Api/Tools/Controllers/ReportsController.cs
```diff
@@ -0,0 +1,150 @@
...
+            });
+
+        // _collectionRepository.GetManyByOrganizationIdWithAccessAsync contains only the group id. 
+        // Create a dictionary to lookup the group names later. 
+        var orgGroups = await _groupRepository.GetManyByOrganizationIdAsync(orgId);
+        var groupNameDictionary = orgGroups.ToDictionary(x => x.Id, x => x.Name);
+
+        // Contains the collections for the organization and the groups/users and permissions
+        var orgCollectionsWithAccess = await _collectionRepository.GetManyByOrganizationIdWithAccessAsync(orgId);
+        var orgItems = await _organizationCiphersQuery.GetAllOrganizationCiphers(orgId);
```
eliykat, line 67: Optimization: group `orgItems` by `collectionId` with their item counts upfront, before entering the loops below.

eliykat (changes requested):
I haven't finished this review, but here's some feedback in the meantime. I'll aim to finish tomorrow - mostly just reviewing the data aggregation in the controller carefully.

src/Api/Tools/Controllers/ReportsController.cs
```diff
@@ -0,0 +1,150 @@
...
+        _applicationCacheService = applicationCacheService;
+    }
+
+    [HttpGet("member-access/{orgId}")]
+    public async Task<IEnumerable<MemberAccessReportModel>> GetMemberAccessReport(Guid orgId)
+    {
+        if (!await _currentContext.AccessReports(orgId))
+        {
+            throw new NotFoundException();
+        }
```
eliykat, line 50: You asked about this in Slack - this authorization check looks good to me.

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
eliykat, line 1: Response models are usually suffixed with `ResponseModel`.

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -0,0 +1,29 @@
...
+    public string GroupName { get; set; }
+    public string CollectionName { get; set; }
+    public int ItemCount { get; set; }
+    public bool ReadOnly { get; set; }
+    public bool HidePasswords { get; set; }
+    public bool Manage { get; set; }
+
+    // internal to not expose 
+    internal ICollection<Guid> CipherIds { get; set; }
+    internal Guid? UserGuid { get; set; }
```
eliykat, line 16: Is this a TODO? If we don't want to expose these to the user then we shouldn't put them in the response model.

In both cases, I think you could track them outside the response model (e.g. in a local var or dict).

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Bit.Api.Tools.Models.Response;
+
+public class MemberAccessReportAccessDetails
```
eliykat, line 3: Some xmldoc comments explaining what these classes represent would be useful.

src/Api/Tools/Controllers/ReportsController.cs
```diff
@@ -0,0 +1,150 @@
...
+        var memberAccessReport = new List<MemberAccessReportModel>();
+        foreach (var user in orgUsers)
+        {
+            var report = new MemberAccessReportModel
+            {
+                UserName = user.OrganizationUserUserDetails.Name,
+                Email = user.OrganizationUserUserDetails.Email,
+                TwoFactorEnabled = user.TwoFactorEnabled,
+                // Both the user's ResetPasswordKey must be set and the organization can UseResetPassword
+                // var orgAbility = await _applicationCacheService.GetOrganizationAbilityAsync(orgId);
```
eliykat, line 123: Old comment can be deleted (the commented-out code).

src/Api/Tools/Controllers/ReportsController.cs
```diff
@@ -0,0 +1,150 @@
...
+            var userAccessDetails = new List<MemberAccessReportAccessDetails>();
+            if (user.OrganizationUserUserDetails.Groups.Any())
+            {
+                var userGroups = accessDetails.Where(x => user.OrganizationUserUserDetails.Groups.Contains(x.GroupId.GetValueOrDefault()));
+                userAccessDetails.AddRange(userGroups);
+            }
+
+            if (user.OrganizationUserUserDetails.Collections.Any())
+            {
+                var userCollections = accessDetails.Where(x => user.OrganizationUserUserDetails.Collections.Any(y => x.CollectionId == y.Id && x.UserGuid == user.OrganizationUserUserDetails.Id));
```
eliykat, line 136: This is a little hard to follow, something that might simplify it is to remove the `collectionId` comparison. I think that if they have an `accessDetails` entry with their userId, we can be sure that they have access to that collection (otherwise it wouldn't have been created in the first place).

src/Api/Tools/Controllers/ReportsController.cs
eliykat, line 1: I think you should consider how this can be unit tested. I suggest that the repository calls are separated from the data manipulation, and you unit test the manipulation on its own. For example, you could have a static method: `MemberAccessReportModel.CreateReport(orgUsers, orgGroups, orgCollectionsWithAccess, orgItems, orgAbility)`, returning `List<MemberAccessReportModel>`. The controller calls the repositories and passes the data directly to the static factory method, which contains the rest of the logic here. Then you can unit test that method without having to mock calls, and you don't have such a thick controller layer.

There might be other ways, but you get the idea.

eliykat (changes requested):
Review completed. I think you've done a great job at figuring out how to arrange and collate the data correctly!

src/Api/Tools/Controllers/ReportsController.cs
ttalty: Great idea.

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -26,4 +43,103 @@ public class MemberAccessReportModel
...
 
+    /// <summary>
+    /// Generates a report for all members of an organization. Containing summary information
+    /// such as item, collection, and group counts. As well as detailed information on the
+    /// user and group collections along with their permissions
+    /// </summary>
+    /// <param name="orgId">Id of the organization to generate the report</param>
+    /// <returns>List of MemberAccessReportResponseModel</returns>
+    public static IEnumerable<MemberAccessReportResponseModel> CreateReport(
+        IEnumerable<OrganizationUserUserDetails> orgUsers,
```
eliykat, line 59: `organizationUsersTwoFactorEnabled` contains the same `OrganizationUserUserDetails` objects as `orgUsers`. I think you can remove the `orgUsers` parameter if you wanted.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserUserDetailsQuery.cs
```diff
@@ -47,9 +44,9 @@ public async Task<IEnumerable<OrganizationUserUserDetailsQueryResponse>> GetOrga
                     userPermissions.DeleteAssignedCollections = false;
                 }
 
-                return orgUser;
+                return o;
             });
-        var responses = await Task.WhenAll(responseTasks);
+        var responses = responseTasks;
```
eliykat, line 49: `responseTasks` is no longer an IEnumerable of Tasks, so its name is no longer accurate, and assigning it between variables here isn't doing anything. You can just `return organizationUsers.Select...` directly.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -132,7 +134,7 @@ public async Task<ListResponseModel<OrganizationUserUserDetailsResponseModel>> G
             throw new NotFoundException();
         }
 
-        var queryResponses = await _organizationUserUserDetailsQuery.GetOrganizationUserUserDetails(
+        var OrganizationUsers = await _organizationUserUserDetailsQuery.GetOrganizationUserUserDetails(
```
eliykat, line 137: Naming convention: should be `organizationUsers` with a small `o`.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -130,28 +134,29 @@ public async Task<ListResponseModel<OrganizationUserUserDetailsResponseModel>> G
...
+            {
+                OrganizationId = orgId,
+                IncludeGroups = includeGroups,
+                IncludeCollections = includeCollections
+            }
+        );
 
-        var organizationUsers = await _organizationUserRepository.GetManyDetailsByOrganizationAsync(orgId, includeGroups, includeCollections);
-        var responseTasks = organizationUsers
+        var organizationUsersTwoFactorEnabled = await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(OrganizationUsers);
```
eliykat, line 146: To be super safe, we should only call this if the feature flag is enabled - in case the new sproc throttles the db or something.

I don't want to waste your time trying to refactor this code too much; I suggest just keeping the existing `Get` and `Get_vNext` split, and just updating both to using the new query. Given that all it requires is changing out the repository call and deleting the user type mapping.

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -0,0 +1,150 @@
...
+    public Guid? GroupId { get; set; }
+    public string GroupName { get; set; }
+    public string CollectionName { get; set; }
+    public int ItemCount { get; set; }
+    public bool ReadOnly { get; set; }
+    public bool HidePasswords { get; set; }
+    public bool Manage { get; set; }
+
+    // internal to not expose 
+    internal Guid? UserGuid { get; set; }
```
eliykat, line 28: My [previous comment](https://github.com/bitwarden/server/pull/4599#discussion_r1714527946) has been partially resolved but I still have the same question for this remaining property.

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -0,0 +1,150 @@
...
+        ICollection<Tuple<Collection, CollectionAccessDetails>> orgCollectionsWithAccess,
+        IEnumerable<CipherOrganizationDetailsWithCollections> orgItems,
+        IEnumerable<(OrganizationUserUserDetails user, bool twoFactorIsEnabled)> organizationUsersTwoFactorEnabled,
+        OrganizationAbility orgAbility)
+    {
+        // Create a dictionary to lookup the group names later. 
+        var groupNameDictionary = orgGroups.ToDictionary(x => x.Id, x => x.Name);
+
+        // Get collection counts into a dictionary
+        var groupCollectionItems = orgItems.SelectMany(x => x.CollectionIds, (x, b) => new { CipherId = x.Id, CollectionId = b }).GroupBy(y => y.CollectionId, (key, g) => new { CollectionId = key, ItemCount = g.Count() });
```
eliykat, line 70: This is called `groupCollectionItems` but it's not related to groups - I assume it refers to the use of `GroupBy`, but it's a confusing name. Maybe just `collectionItemCounts`?

I didn't know that you could use `SelectMany` in this way, which is cool. However this is a super long line, it would help comprehension to break it across multiple lines, e.g. (just because I already did it for review)

```cs
        var groupCollectionItems = orgItems
            .SelectMany(x => x.CollectionIds,
                (x, b) => new { CipherId = x.Id, CollectionId = b })
            .GroupBy(y => y.CollectionId,
                (key, g) => new { CollectionId = key, ItemCount = g.Count() });
```

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -0,0 +1,150 @@
...
+        var collectionCountDictionary = groupCollectionItems.ToDictionary(x => x.CollectionId, x => x.ItemCount);
```
eliykat, line 71: I suggest just chaining this at the end of the preceding variable declaration. As long as each chained operator is on a new line, I think that's easier to follow than declaring additional variables.

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (user.Collections.Any())
+            {
+                var userCollections = accessDetails.Where(x => user.Collections.Any(y => x.UserGuid == user.Id));
+                userAccessDetails.AddRange(userCollections);
+            }
+            report.AccessDetails = userAccessDetails;
+
+            // Distinct items only            
+            var distinctItems = report.AccessDetails.Select(x => x.CollectionId).Distinct();
+            report.TotalItemCount = distinctItems.Select(x => collectionCountDictionary.TryGetValue(x, out var count) ? count : 0).Sum();
```
eliykat, line 142: I'm not sure the `TotalItemCount` still works - an item can be in multiple collections, so just summing the item count of each collection is going to contain duplicates.

bitwarden_license/src/Sso/wwwroot/css/site.css
eliykat, line 1: As mentioned on Slack - remove all these extra files from the diff.

eliykat (changes requested):

src/Api/Tools/Models/Response/MemberAccessReportModel.cs
ttalty: Sounds good. That's always a thing I struggle with deciding. Sometimes I think having a new variable can help make it readable and easier to debug. Sometimes just stringing together is better. I appreciate the input and will change.

eliykat: It's definitely a matter of opinion; certainly it makes debugging harder. Chaining is useful to me because it shows you don't care (and aren't going to use) the intermediate results, you just care about the outcome at the end of the chain. That's particularly useful here where we already have a lot of variables in scope.

You haven't made this change but it's also non-blocking so I'll leave it up to you.

eliykat: I see that you needed to keep the `collectionItems` to count them properly later. That makes sense to me!

eliykat (approved):
Great work and thanks for your patience through multiple rounds of code reviews. It's good to go from my perspective, although certainly the more eyes the better from other reviewers. Also unit tests would still be a good follow-up.

EDIT: although there are still linting errors and test failures.

test/Api.Test/AdminConsole/Queries/OrganizationUserUserDetailsQueryTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        customUser.Permissions = CoreHelpers.ClassToJsonData(new Permissions
+        {
+            EditAssignedCollections = true,
+            DeleteAssignedCollections = true,
+        });
+
+        var response = await sutProvider.Sut.GetOrganizationUserUserDetails(new OrganizationUserUserDetailsQueryRequest { OrganizationId = organizationId });
+
+        var customUserResponse = response.First(r => r.Id == organizationUsers.First().Id);
+        Assert.Equal(OrganizationUserType.User, customUserResponse.Type);
```
eliykat, line 38: Same issue as above - this is now missing key assertions because sut code has been moved back to the controller. Once you've actioned the feedback above, these tests need to be updated so that we keep the same assertions we have today.

eliykat (changes requested):
To summarise my comments below - changing the user type and setting custom permissions to `false` both belong together in the query. You should be able to move the following tests to the query class and have them pass without any change to their assertions:

- Get_DowngradesCustomUsersWithDeprecatedPermissions
- Get_HandlesNullPermissionsObject
- Get_SetsDeprecatedCustomPermissionstoFalse

aj-rosado (approved):

r-tome (approved):

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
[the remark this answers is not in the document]
ttalty: I did not refactor the entire OrgUsersController. I thought we discussed the other work would be resolved in tech debt.

eliykat (approved):

LRNcardozoWDF (approved):

djsmith85 (approved):
Approving for Tools and ready to merge, with previous approval from @aj-rosado and this being guarded behind a feature flag.

PR #4604, opened 2024-08-08
[AC-1722] Deprecate "Edit/Delete Assigned Collections" custom permissions

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10348
https://bitwarden.atlassian.net/browse/PM-10289

## 📔 Objective

Custom users with the `EditAssignedCollections` or `DeleteAssignedCollections` permissions were not migrated earlier, so now they need to be converted to regular Users, and these permissions should be removed from all users' Permissions objects.

This pull request also contains integration tests to test these data migrations.

A new service called IMigrationTesterService has been added to help with integration tests. It detects the database type being tested—if it's SQL, it uses `DbMigrator`; if it's Entity Framework (EF), it finds and runs the appropriate migration.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


util/Migrator/DbScripts/2024-08-07_00_FinalFlexibleCollectionsDataMigrations.sql
r-tome, line 1: @rkac-bw , could you please check if this script is too demanding on the database? I’m concerned because most, if not all, users have Permissions blobs, and parsing the JSON might be resource-intensive.

r-tome:
I did some testing and discovered an issue when the migration we’re testing isn’t the latest one. Here’s what happened:
- I added the migration `FinalFlexibleCollectionsDataMigrations`.
- I then added another migration called `Test`.
- This resulted in the table `__EFMigrationsHistory` having a record for `FinalFlexibleCollectionsDataMigrations`, followed by `Test`.
- However, when I ran the test specifically for `FinalFlexibleCollectionsDataMigrations`, the `Test` record was deleted from `__EFMigrationsHistory`.

This could mean that the integration tests for a specific migration should be deleted when adding a next migration.

r-tome:
@rkac-bw has assisted me with optimizing the SQL data migrations.

eliykat:
@justindbaur I'd super appreciate your feedback on this one as it's extending the database testing attributes you wrote. My main uncertainty with it (discussed with @r-tome) is that unlike other tests, this is a point-in-time test only; it won't be updated with code changes or subsequent changes to the db schema. For example, if we later removed the `Permissions` column, this test would no longer pass or even compile.

It's kind of like state migrations in clients, which have to maintain their own copies of all models they reference. I'm not sure if it's worth going to that extent, or whether we just delete migration integration tests after release, given that migration scripts should never be modified after the fact.

justindbaur (approved):
This is some very awesome infrastructure and very good tests. I definitely agree with @eliykat, these are very point-in-time tests, and the only way to make sure they could pass all into the future would be to do a lot of raw sql for each provider...which would really suck. It does seem to only use `GetByIdAsync` and `CreateAsync` which are the least likely to have major changes but I think I might lean towards the tests just being deleted after a release. 

As a reviewer I would definitely appreciate knowing that tests were written and tested but as a maintainer if they were left around they could break on me far into the future and I wouldn't have context as to why. I think that's why it makes sense to me that they should exist but they should probably be short lived. But I without a doubt want to keep all the plumbing that makes it possible around so that the next person that wants to test their migrations can do it even easier.

test/Infrastructure.IntegrationTest/Services/Implementations/EfMigrationTesterService.cs
```diff
@@ -0,0 +1,59 @@
...
+                deleteCommand = "DELETE FROM \"__EFMigrationsHistory\" WHERE \"MigrationId\" LIKE @migrationName";
+                parameter = new NpgsqlParameter("@migrationName", "%" + _migrationName);
+                break;
+            case SupportedDatabaseProviders.Sqlite:
+                parameter = new SqliteParameter("@migrationName", "%" + _migrationName);
+                break;
+        }
+
+        _databaseContext.Database.ExecuteSqlRaw(deleteCommand, parameter);
+    }
```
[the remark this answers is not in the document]
r-tome: I tried that now but it requires that we add the `HistoryRow` class as part of the `DbContext` and I don't think we should have that table accessible, right?

r-tome:
@justindbaur thank you for the feedback! I completely agree that these tests should be deleted after a release. Their main purpose is to support the data migrations development process and provide confidence to reviewers that the scripts work across all DB providers. However, the infrastructure that supports these tests is essential for future data migrations and should be kept to make testing easier for the next person.

test/Infrastructure.IntegrationTest/AdminConsole/Migrations/FinalFlexibleCollectionsDataMigrationsTests.cs
eliykat, line 1: Please add tests for:

1. a non-custom user with a JSON blob; they should have the old permissions removed but their type should not change
2. a user whose JSON blob already has these properties removed, mostly to make sure it doesn't throw an error

util/Migrator/DbScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.sql
eliykat, line 1: @rkac-bw I know you've optimized this, but do you have an idea of how long it will take to run when deployed? Just so we don't cause any issues with long-running migrations.

If there's any concern with how long this will take, or the complexity of this migration, we can clear a lot of these JSON values first in a separate migration because they're not needed anyway. But, if it's not a blocker, then all the better and we can get to that other cleanup later.

util/Migrator/DbScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.sql
```diff
@@ -0,0 +1,118 @@
...
+
+    SET @RowsAffected = @@ROWCOUNT;
+
+    IF @RowsAffected = 0
+        BREAK;
+END
+
+-- Remove 'editAssignedCollections' and 'deleteAssignedCollections' properties from Permissions
+    -- Step 1: Create a temporary table to store the IDs and parsed JSON values
+    CREATE TABLE #TempIds (
```
eliykat, line 58: Indentation

test/Infrastructure.IntegrationTest/AdminConsole/Migrations/FinalFlexibleCollectionsDataMigrationsTests.cs
```diff
@@ -0,0 +1,156 @@
...
+        // Assert that the user was migrated to a User type with null permissions
+        var migratedOrgUser = await organizationUserRepository.GetByIdAsync(orgUser.Id);
+        Assert.NotNull(migratedOrgUser);
+        Assert.Equal(orgUser.Id, migratedOrgUser.Id);
+        Assert.Equal(OrganizationUserType.User, migratedOrgUser.Type);
+        Assert.Null(migratedOrgUser.Permissions);
+    }
+
+    [DatabaseTheory, DatabaseData(MigrationName = "FinalFlexibleCollectionsDataMigrations")]
+    public async Task RunMigration_WithoutAssignedCollectionsPermissions_WithCustomUserType_RemovesAssignedCollectionsPermissions(
```
eliykat, line 82: Would be good to set some other permissions to `true` and then make sure their values are preserved.

util/PostgresMigrations/HelperScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.psql
```diff
@@ -0,0 +1,38 @@
...
+UPDATE "OrganizationUser"
+SET
+    "Type" = 2,
+    "Permissions" = NULL
+WHERE
+    "Type" = 4
+    AND jsonb_typeof("Permissions"::jsonb) = 'object'
+    AND (
+        COALESCE(("Permissions"::jsonb)->>'editAssignedCollections', 'false') = 'true'
+        OR COALESCE(("Permissions"::jsonb)->>'deleteAssignedCollections', 'false') = 'true'
```
eliykat, line 11: I assume `COALESCE(..., 'false')` is required in case the value is missing/null?

util/PostgresMigrations/HelperScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.psql
```diff
@@ -0,0 +1,38 @@
...
+    "Permissions" = "Permissions"::jsonb - 'editAssignedCollections' - 'deleteAssignedCollections'
+WHERE
+    "Permissions" IS NOT NULL
+    AND (
+        SELECT
+            CASE
+                WHEN jsonb_typeof("Permissions"::jsonb) = 'object' THEN TRUE
+                ELSE FALSE
+            END
+    );
```
eliykat, line 38: I'm not great at postgres. But is this `SELECT` and `CASE` statement doing anything? You could just use `AND jsonb_typeof("Permissions"::jsonb) = 'object'` in the first UPDATE statement above.

eliykat (changes requested):
Great work on this! Did a closer review now that the integration test questions are mostly worked out.

As discussed on Slack, you also need to:
* fix sqlite migration errors
* bump date on your migration scripts (that should resolve the `AccessAll` null errors)

util/Migrator/DbScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.sql
r-tome: > First update: 299 Rows
> Second update: 783,906 rows
> The way it is written now it takes an hour to update and each loop locks for up to 10 seconds.
> this version runs in about 10 minutes and each loop is around a second

These are the benchmarks @rkac-bw gave me along with the optimized script

test/Common/AutoFixture/Attributes/CustomAutoDataAttribute.cs
```diff
@@ -11,7 +11,7 @@ public CustomAutoDataAttribute(params Type[] iCustomizationTypes) : this(iCustom
 
     public CustomAutoDataAttribute(params ICustomization[] customizations) : base(() =>
     {
-        var fixture = new Fixture();
+        var fixture = new Fixture().WithAutoNSubstitutions();
```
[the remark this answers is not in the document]
r-tome: It was necessary before, some integration tests are broken on `main` (ie: GetBySsoUserAsync_Works_DataMatches).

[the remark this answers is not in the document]
r-tome: An `ILogger<T>` was introduced to a few repositories and since then the integration tests stopped working:

> AutoFixture was unable to create an instance from Microsoft.Extensions.Logging.ILogger`1[Bit.Infrastructure.Dapper.Repositories.OrganizationRepository] because it's an interface. There's no single, most appropriate way to create an object implementing the interface, but you can help AutoFixture figure it out.

test/Infrastructure.EFIntegration.Test/AutoFixture/EntityFrameworkRepositoryFixtures.cs
```diff
@@ -91,6 +92,8 @@ public object Create(object request, ISpecimenContext context)
                 })
             .CreateMapper()));
 
+            fixture.Customize<ILogger<T>>(x => x.FromFactory(() => Substitute.For<ILogger<T>>()));
```
r-tome, line 95: See comment above.

util/Migrator/DbMigrator.cs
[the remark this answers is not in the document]
r-tome: I've refactored it to not use `DbMigrator`. I originally wrote it that way but then thought about reusing `DbMigrator` since it's already proven to work. However, you're right, it's best not to modify production code for tests.

test/Infrastructure.IntegrationTest/AdminConsole/Migrations/FinalFlexibleCollectionsDataMigrationsTests.cs
r-tome: Added `accessEventLogs = true`

util/PostgresMigrations/HelperScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.psql
r-tome: Yes

util/PostgresMigrations/HelperScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.psql
r-tome: You're absolutely right, the `SELECT CASE` is not needed. This was leftover from my earlier experiments. I'm not great at Postgres so its a lot of trial and error 😅

util/PostgresMigrations/HelperScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.psql
r-tome: Actually, I've just tested the script without `COALESCE` and it still works.

test/Infrastructure.IntegrationTest/Services/Implementations/EfMigrationTesterService.cs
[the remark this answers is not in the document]
justindbaur: Yeah, I think you are right @r-tome, I saw [this](https://github.com/dotnet/efcore/blob/78400fcec978162059afbf744775d3db720820ca/src/EFCore.Relational/Migrations/HistoryRepository.cs#L218-L224) and thought it might be implicitly part of the model but now that I look a bit closer they are creating their own `ModelBuilder` for that [here](https://github.com/dotnet/efcore/blob/78400fcec978162059afbf744775d3db720820ca/src/EFCore.Relational/Migrations/HistoryRepository.cs#L89-L95). 

I can definitely live with how it is now, was just hoping we could make it a little more future proof.

test/Infrastructure.IntegrationTest/Services/Implementations/SqlMigrationTesterService.cs
```diff
@@ -0,0 +1,60 @@
...
+        var resourceName = assembly.GetManifestResourceNames()
+            .FirstOrDefault(r => r.EndsWith($"{scriptName}.sql"));
+
+        if (resourceName == null)
+        {
+            throw new FileNotFoundException($"SQL migration script file for '{scriptName}' was not found.");
+        }
+
+        using var stream = assembly.GetManifestResourceStream(resourceName);
+        using var reader = new StreamReader(stream);
```
justindbaur, line 57: You've already validated that this is a valid resource with the given name so you can override the compiler warning telling you it might be null.

```suggestion
        using var reader = new StreamReader(stream!);
```

test/Infrastructure.IntegrationTest/AdminConsole/Migrations/FinalFlexibleCollectionsDataMigrationsTests.cs
```diff
@@ -0,0 +1,253 @@
...
+        Assert.Equal(orgUser.Id, migratedOrgUser.Id);
+        Assert.Equal(OrganizationUserType.Custom, migratedOrgUser.Type);
+        Assert.NotEqual(orgUser.Permissions, migratedOrgUser.Permissions);
+        Assert.NotNull(migratedOrgUser.Permissions);
+        Assert.Contains("accessEventLogs", orgUser.Permissions);
+        Assert.Contains("editAssignedCollections", orgUser.Permissions);
+        Assert.Contains("deleteAssignedCollections", orgUser.Permissions);
+
+        Assert.Contains("accessEventLogs", migratedOrgUser.Permissions);
+        Assert.True(migratedOrgUser.GetPermissions().AccessEventLogs);
```
justindbaur, line 114: Can you take care of both of these `Dereference of a possibly null reference.` warnings? I'm fine with that being just `migratedOrgUser.GetPermissions()!.AccessEventLogs` if you are sure `GetPermissions` won't return null in this context. Or the longer way would be do do something like:
```suggestion
        var migratedOrgUserPermissions = migratedOrgUser.GetPermissions();
        Assert.NotNull(migratedOrgUserPermissions);
        Assert.True(migratedOrgUserPermissions.AccessEventLogs);
```

justindbaur (changes requested):
Just a few things around the compiler warnings, other than that it all looks great.

util/PostgresMigrations/HelperScripts/2024-08-21_00_FinalFlexibleCollectionsDataMigrations.psql
```diff
@@ -0,0 +1,32 @@
...
+    AND COALESCE(("Permissions"::jsonb)->>'accessReports', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'createNewCollections', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'editAnyCollection', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'deleteAnyCollection', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'manageGroups', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'managePolicies', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'manageSso', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'manageUsers', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'manageResetPassword', 'true') = 'false'
+    AND COALESCE(("Permissions"::jsonb)->>'manageScim', 'true') = 'false';
```
eliykat, line 24: You removed some but not all of the `COALESCE`s in this file - I'm thinking these are also unnecessary now.

util/Migrator/DbScripts/2024-08-12_00_FinalFlexibleCollectionsDataMigrations.sql
eliykat: OK cool thank you - maybe just give BRE a heads up that this script will take about 10 mins. (That runtime seems acceptable but is still on the long side, so just to avoid any concern on their end.)

eliykat (changes requested):
Great work, 1 minor comment below but otherwise good to go.

Please give QA a heads up before merging to `main`. Is there an orgUser they know is still a custom user in the database, so they can manually confirm a successful migration (both in the db and regression test in the app)? Once this is merged to `main` it'll be too late for QA to check this.

eliykat (dismissed):

eliykat (approved):

test/Infrastructure.IntegrationTest/AdminConsole/Migrations/FinalFlexibleCollectionsDataMigrationsTests.cs
```diff
@@ -0,0 +1,334 @@
...
+        await organizationUserRepository.ReplaceAsync(orgUser);
+
+        // Run data migration
+        migrationTester.ApplyMigration();
+
+        // Assert that the user kept the Admin type and no changes were made to the permissions
+        var migratedOrgUser = await organizationUserRepository.GetByIdAsync(orgUser.Id);
+        Assert.NotNull(migratedOrgUser);
+        Assert.Equal(orgUser.Id, migratedOrgUser.Id);
+        Assert.Equal(orgUser.Type, migratedOrgUser.Type);
```
eliykat, line 190: Generally you should avoid comparing 2 mutable values like this. Both may have changed without us realising and this could pass. Better to break it out into separate assertions using a literal value, e.g.

```cs
        Assert.Equal(OrganizationUserType.Custom, orgUser.Type);
        Assert.Equal(OrganizationUserType.Custom, migratedOrgUser.Type);
```

util/PostgresMigrations/HelperScripts/2024-08-26_00_FinalFlexibleCollectionsDataMigrations.psql
```diff
@@ -0,0 +1,35 @@
+﻿-- Migrate Custom users who only have 'editAssignedCollections' and/or 'deleteAssignedCollections' custom permissions to the User type.
+UPDATE "OrganizationUser"
+SET
+    "Type" = 2,
+    "Permissions" = NULL
+WHERE
+    "Type" = 4
+    AND "Permissions" IS NOT NULL
+    AND "Permissions" ~ '^\s*\{.*\}\s*$'  -- Check if Permissions is a valid JSON object
```
eliykat, line 9: This is rough. But I see that postgres didn't support Is_Json until [relatively recently](https://modern-sql.com/caniuse/is-json) and I don't think we've defined any version support for our EF databases.

Looking at what we've done previously in Postgres, I found this: https://github.com/bitwarden/server/blob/main/util/PostgresMigrations/HelperScripts/2021-09-21_01_SplitManageCollectionsPermission.psql#L1. In that case it doesn't look like we guarded against invalid json at all, so this already seems better as a rough check.

I think this is a fine solution for now.

eliykat (approved):

rkac-bw (approved):
LGTM, we will have to manual deploy the migration in prod

PR #4623, opened 2024-08-13
[PM-10589] Update database defaults for collection management settings

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10589

## 📔 Objective

> When we added LimitCollectionCreationDeletion and AllowAdminAccessToAllCollectionItems, we set them to default to 1, being the transitional setting.
> 
> That no longer applies because all orgs are being created with these set to 0 by the code. We should probably update this at the db level to reflect that, in particular because AllowAdminAccessToAllCollectionItems = 1 is the more permissive setting which is not intended.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


eliykat (changes requested):
Thanks for doing this! There is also a default constraint on the column: https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Tables/Organization.sql#L55

r-tome:
> Thanks for doing this! There are also default constraints on the columns: https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Tables/Organization.sql#L54-L55

I forgot to check those, thank you!

util/Migrator/DbScripts/2024-08-13_00_UpdateCollectionManagementSettingsDbDefaults.sql
```diff
@@ -0,0 +1,312 @@
+-- Objective: 'LimitCollectionCreationDeletion' and 'AllowAdminAccessToAllCollectionItems' columns were set to 1 by
+-- default as a transitional measure. This script updates the default values to 0 to mirror the values set in code.
+
+-- Step 1: Drop existing default constraints
+ALTER TABLE [dbo].[Organization] DROP CONSTRAINT [DF_Organization_LimitCollectionCreationDeletion];
+GO
+ALTER TABLE [dbo].[Organization] DROP CONSTRAINT [DF_Organization_AllowAdminAccessToAllCollectionItems];
+GO
```
eliykat, line 8: We should check if the constraint exists before trying to drop it: [example from our code](https://github.com/bitwarden/server/blob/main/util/Migrator/DbScripts/2024-06-11_00_FixProviderInvoiceItem.sql#L9-L15), [related SO answer](https://stackoverflow.com/a/13186918). Although that doesn't discriminate between tables.

@rkac-bw what's your recommendation?

util/Migrator/DbScripts/2024-08-13_00_UpdateCollectionManagementSettingsDbDefaults.sql
```diff
@@ -0,0 +1,312 @@
...
+
+-- Step 2: Add new default constraints with default value of 0
+ALTER TABLE [dbo].[Organization] ADD CONSTRAINT [DF_Organization_LimitCollectionCreationDeletion] DEFAULT (0) FOR [LimitCollectionCreationDeletion];
+GO
+ALTER TABLE [dbo].[Organization] ADD CONSTRAINT [DF_Organization_AllowAdminAccessToAllCollectionItems] DEFAULT (0) FOR [AllowAdminAccessToAllCollectionItems];
+GO
```
eliykat, line 14: Similarly, need to check that the column exists before adding the constraint.

eliykat (changes requested):
Mostly lgtm, just a few questions around our usual practice of checking that things exist before changing them.

eliykat (approved):
Looks good to me, thanks!

util/Migrator/DbScripts/2024-08-13_00_UpdateCollectionManagementSettingsDbDefaults.sql
rkac-bw: It's much cleaner to check if exists

rkac-bw (approved):
lgtm

PR #4630, opened 2024-08-14
[AC-2971] Draft: Policy strategy pattern

eliykat:
## 🎟️ Tracking

TBA

## 📔 Objective

This is a proof of concept for refactoring the server-side PolicyService.

`PolicyService.SaveAsync` takes up most of the class, and it has the following problems:
* Lots of conditional logic and generally a bit hard to follow
* Bloated, increases in size as we add more policy types
* Because all policies use the same class, high risk of regression when changing any one policy
* Requires a high degree of knowledge about all policies, generally has too many responsibilities

This PR suggests implementing a [strategy pattern](https://refactoring.guru/design-patterns/strategy) to handle enabling and disabling policies. Each `PolicyType` should have a corresponding `IPolicyStrategy` implementation, which can be used to handle actions when the policy is changed from enabled -> disabled or vice versa. `PolicyService` then injects all strategies and simply delegates the work to the matching strategy.

This also allows for easier unit testing, which can be done for each strategy, rather than the entire PolicyService class.

This PR shows how this could work, using the Single Org policy as an example. Please provide feedback to help us plan future tech debt work.

**Future improvements - but out of scope for now**

These strategies could also be used for other policy-specific logic, such as:
* [reconciling multiple policies](https://github.com/bitwarden/server/blob/f21d43f252920190e31e36a0f98f308d59230fdf/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L88-L89)
* [determining which users are exempt from the policy](https://github.com/bitwarden/server/blob/f21d43f252920190e31e36a0f98f308d59230fdf/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L134-L135)

Separating strategies from the service could also let us have different code ownership, e.g. teams own policies that affect their own domains (something we've talked about for a while now).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
@@ -1,0 +1,15 @@
+﻿using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Implementations;
+using Bit.Core.AdminConsole.Services;
+using Bit.Core.AdminConsole.Services.Implementations;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies;
+
+public static class PolicyServiceCollectionExtensions
+{
+    public static void AddPolicyServices(this IServiceCollection services)
+    {
+        services.AddScoped<IPolicyService, PolicyService>();
+        services.AddScoped<IPolicyStrategy, SingleOrgPolicyStrategy>();
+    }
+}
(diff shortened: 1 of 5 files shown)


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies;
+
+public static class PolicyServiceCollectionExtensions
+{
+    public static void AddPolicyServices(this IServiceCollection services)
+    {
+        services.AddScoped<IPolicyService, PolicyService>();
+        services.AddScoped<IPolicyStrategy, SingleOrgPolicyStrategy>();
+    }
```
eliykat, line 14: Other strategy implementations would also be registered here under the `IPolicyStrategy` interface.

eliykat:
Implemented by #4877 🎉

PR #4632, opened 2024-08-14
Log events from the import organization flow

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10811
https://bitwarden.atlassian.net/browse/PM-10810

Client PR: https://github.com/bitwarden/clients/pull/10660

## 📔 Objective

Users removed by directory connector syncs do not log events to organization event logs. This commit adds this missing logging.

## 📸 Screenshots

https://github.com/user-attachments/assets/bbd05f9c-77d5-44ab-a8a2-d78d85c5c12f

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1800,16 +1800,19 @@ public async Task ImportAsync(Guid organizationId,
...
-
-            await _organizationUserRepository.DeleteManyAsync(removeUsersSet
+            var removeUsersSet = new HashSet<string>(removeUserExternalIds)
                 .Except(newUsersSet)
                 .Where(u => existingUsersDict.ContainsKey(u) && existingUsersDict[u].Type != OrganizationUserType.Owner)
-                .Select(u => existingUsersDict[u].Id));
+                .Select(u => existingUsersDict[u]);
+
+            await _organizationUserRepository.DeleteManyAsync(removeUsersSet.Select(u => u.Id));
+            events.AddRange(removeUsersSet.Select(u => (new OrganizationUser { Id = u.Id, UserId = u.UserId, OrganizationId = u.OrganizationId }, EventType.OrganizationUser_Removed, (DateTime?)DateTime.UtcNow)));
```
eliykat, line 1815: We should pass in the real, complete `OrganizationUser` object here. It looks like you're constructing an incomplete object based on the properties that `EventService` currently uses, but that could change in the future without this calling location knowing about it.

We should get the full `OrganizationUser` object from the database - although we already have `OrganizationUserUserDetails` in scope, so that is a bit duplicative. Perhaps instead we could change the `IEventService.LogOrganizationUserEventAsync` interface to accept a interface rather than the concrete `OrganizationUser` object, then you can use the `OrganizationUserUserDetails` you already have without writing mapping code.

eliykat: Also, on a _much_ less substantive note: breaking this line into multiple lines would help readability.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1953,6 +1957,7 @@ await UpdateUsersAsync(group, groupsDict[group.ExternalId].ExternalUserIds,
             }
         }
 
+        await _eventService.LogOrganizationUserEventsAsync(events);
```
eliykat, line 1960: There's an [overload](https://github.com/bitwarden/server/blob/3d7fe4f8afaa65add13c58c13d4351ae86d0f22c/src/Core/Services/IEventService.cs#L22-L23) that lets you specify an `EventSystemUser`. The idea is that it identifies "who" (or what) made the change when it's not tied to a specific user account - currently just used for SCIM and the domain verification job.

I suggest we add an `EventSystemUser.PublicApi` enum value, so that this displays in the Admin Console logs as a change made via the public api (technically it's not limited to directory connector). Note that this would also require updating the [client-side enum](https://github.com/bitwarden/clients/blob/main/libs/common/src/enums/event-system-user.enum.ts#L2) and [translation mapping](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/admin-console/organizations/manage/events.component.ts#L22-L25) but otherwise should "just work".

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
eliykat, line 1: Looking at what other logging may be required, it looks like this method creates and updates groups but does not log any events for it.

eliykat (changes requested):
This is really good work to tighten up our logging 🎉

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
addisonbeck: Implemented as suggested on https://github.com/bitwarden/server/pull/4632/commits/aa362e441cf72d05fd20dedbff46820a31f3db61

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
addisonbeck: Implemented as suggested on https://github.com/bitwarden/server/pull/4632/commits/89cdfffde2bbe078e13e20dfeb1f386de85ed34b

Client PR is https://github.com/bitwarden/clients/pull/10660

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
addisonbeck: Implemented as suggested on https://github.com/bitwarden/server/pull/4632/commits/67e2776fc9b5b4b0800d2125234c2d3809255ae1

addisonbeck:
@eliykat It's worth mentioning that I sort of hate the way all of the changes from this feedback are implemented. They're all important changes, but implementing them without doing a redesign of the import function was rough. I'm open to any suggestions here and would appreciate it if you have any.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1930,19 +1930,20 @@ public async Task ImportAsync(Guid organizationId,
...
 
-                await _groupRepository.CreateAsync(group);
+                savedGroups.Add(await _groupRepository.CreateAsync(group));
                 await UpdateUsersAsync(group, groupsDict[group.ExternalId].ExternalUserIds,
                     existingExternalUsersIdDict);
             }
 
             await _eventService.LogGroupEventsAsync(
-                newGroups.Select(g => (g, EventType.Group_Created, eventSystemUser, (DateTime?)DateTime.UtcNow)));
+                savedGroups.Select(g => (g, EventType.Group_Created, eventSystemUser, (DateTime?)DateTime.UtcNow)));
```
eliykat: Won't `savedGroups` always duplicate `newGroups`?

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
eliykat: I really like how this turned out!

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1913,17 +1930,21 @@ await _organizationUserRepository.DeleteManyAsync(removeUsersSet
 
             var newGroups = groups
                 .Where(g => !existingExternalGroupsDict.ContainsKey(g.Group.ExternalId))
-                .Select(g => g.Group);
+                .Select(g => g.Group).ToList();
 
+            var savedGroups = new List<Group>();
```
eliykat, line 1935: It took me a moment to figure out why you're compiling this separate list. I assume it's because `groupRepository.CreateAsync` could return a new object?

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1913,17 +1930,21 @@ await _organizationUserRepository.DeleteManyAsync(removeUsersSet
...
 
-                await _groupRepository.CreateAsync(group);
+                savedGroups.Add(await _groupRepository.CreateAsync(group));
                 await UpdateUsersAsync(group, groupsDict[group.ExternalId].ExternalUserIds,
                     existingExternalUsersIdDict);
             }
 
+            await _eventService.LogGroupEventsAsync(
+                savedGroups.Select(g => (g, EventType.Group_Created, eventSystemUser, (DateTime?)DateTime.UtcNow)));
+
```
eliykat, line 1947: You also need to log the `Group_Updated` event for the next code block.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1953,6 +1974,14 @@ await UpdateUsersAsync(group, groupsDict[group.ExternalId].ExternalUserIds,
...
         }
 
+        if (eventSystemUser != null)
+        {
+            await _eventService.LogOrganizationUserEventsAsync(events.Select(e => (e.ou, e.e, eventSystemUser.Value, e.d)));
+        }
+        else
+        {
+            await _eventService.LogOrganizationUserEventsAsync(events);
+        }
```
eliykat, line 1984: Why does this need the `if` statement but `LogGroupEvents` doesn't?

eliykat (changes requested):

eliykat:
> @eliykat It's worth mentioning that I sort of hate the way all of the changes from this feedback are implemented. They're all important changes, but implementing them without doing a redesign of the import function was rough. I'm open to any suggestions here and would appreciate it if you have any.

There are a few issues here that I can see, which I have some suggestions for. All of them increase the scope of this ticket so it's your decision, but let me know if any would make you feel better about the changes to this method.

---

1. `ImportAsync` should be extracted to another class, presumably a command. Obviously some of these discussions are ongoing, however it might be simple enough that we could do it now, if you wanted. The method is largely self contained within `OrganizationService`. It only calls other services/repositories (which can be injected into the command), `UpdateUsersAsync` (which is only used by this method so can be moved with it), and the public `InviteUsersAsync` (which we can fudge and just call from the command for now, still seems like a net improvement). If you do this, please do it in a separate PR first.

2. It's awkward to have 2 different types of "users" - `OrganizationUser`s and `EventSystemUser`s - which end up getting tacked on to every method definition. I think we probably need to have a wrapper object that can take either/or and which can be passed around and then unwrapped right at the end in `EventService`. for example:

```cs
class LogUser
{
  public OrganizationUser? OrganizationUser { get; } // can only be set in the ctor
  public EventSystemUser? EventSystemUser { get; }

  public LogUser(OrganizationUser organizationUser)
  {
    OrganizationUser = organizationUser;
  }

  public LogUser(EventSystemUser eventSystemUser)
  {
    EventSystemUser = eventSystemUser;
  }
}
```

Then refactor every interface all the way through the call chain to only take a `LogUser`, and unwrap it in EventService. If you choose to do this you could start only with the interfaces you're using here.

3. I actually didn't realise we have a private api that calls this method. Directory Connector uses the public api. I can't see anywhere that uses the private api - there is api code in `clients` but it appears to be unused. We could delete the private api so it's only called by the public api, then require an `EventSystemUser` to be passed in, and then it's non-nullable and that simplifies some of the flow without doing (2) above.

src/Core/Services/IEventService.cs
```diff
@@ -18,10 +19,10 @@ public interface IEventService
     Task LogGroupEventAsync(Group group, EventType type, EventSystemUser systemUser, DateTime? date = null);
     Task LogGroupEventsAsync(IEnumerable<(Group group, EventType type, EventSystemUser? systemUser, DateTime? date)> events);
     Task LogPolicyEventAsync(Policy policy, EventType type, DateTime? date = null);
-    Task LogOrganizationUserEventAsync(OrganizationUser organizationUser, EventType type, DateTime? date = null);
-    Task LogOrganizationUserEventAsync(OrganizationUser organizationUser, EventType type, EventSystemUser systemUser, DateTime? date = null);
-    Task LogOrganizationUserEventsAsync(IEnumerable<(OrganizationUser, EventType, DateTime?)> events);
-    Task LogOrganizationUserEventsAsync(IEnumerable<(OrganizationUser, EventType, EventSystemUser, DateTime?)> events);
+    Task LogOrganizationUserEventAsync<T>(T organizationUser, EventType type, DateTime? date = null) where T : IOrganizationUser;
```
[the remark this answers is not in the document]
addisonbeck: C# does not like this conversion for existing references. We could cast it, but I think generics are more appropriate for the language. I could be missing an easy fix, though.

Screenshot 2024-08-23 at 4 18 49 PM

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
addisonbeck: It's because this
```c#
            var newGroups = groups
                .Where(g => !existingExternalGroupsDict.ContainsKey(g.Group.ExternalId))
                .Select(g => g.Group).ToList();

            foreach (var group in newGroups)
            {
                group.CreationDate = group.RevisionDate = DateTime.UtcNow;
                newGroups[newGroups.IndexOf(group)] = await _groupRepository.CreateAsync(group);
                await UpdateUsersAsync(group, groupsDict[group.ExternalId].ExternalUserIds,
                    existingExternalUsersIdDict);
            }
```

Does this

```
    System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
         at System.Collections.Generic.List`1.Enumerator.MoveNext()
         at Bit.Core.Services.OrganizationService.ImportAsync(Guid organizationId, Nullable`1 importingUserId, IEnumerable`1 groups, IEnumerable`1 newUsers, IEnumerable`1 removeUserExternalIds, Boolean overwriteExisting, Nullable`1 eventSystemUser) in /Users/me/bitwarden/server/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs:line 1936

```

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
addisonbeck: `LogGroupEvents` doesn't have an overload without an `EventSystemUser` argument, it just handles a null value for one.

```c#
    Task LogGroupEventsAsync(IEnumerable<(Group group, EventType type, EventSystemUser? systemUser, DateTime? date)> events);
```

`LogOragnizationUserEvents` uses an overload to exclude the property instead
```cs
    Task LogOrganizationUserEventsAsync<T>(IEnumerable<(T, EventType, DateTime?)> events) where T : IOrganizationUser;
    Task LogOrganizationUserEventsAsync<T>(IEnumerable<(T, EventType, EventSystemUser, DateTime?)> events) where T : IOrganizationUser;
```

src/Core/Services/IEventService.cs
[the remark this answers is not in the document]
eliykat: I didn't understand this error so I tried making the change myself on this branch, and I didn't receive any error. However I suspect the difference is only stylistic at most, so it doesn't matter particularly. And if I am wrong it's definitely not worth casting for.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
eliykat: Thanks for the explanation!

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
addisonbeck: Implemented on https://github.com/bitwarden/server/pull/4632/commits/e7d1f7eb2cff926994f61b91c88e553fe1bc8b6f

addisonbeck:
Implemented #3 and #2 on https://github.com/bitwarden/server/pull/4632/commits/ec980a05cf43af084bb402172e610bd7ff6a374d and https://github.com/bitwarden/server/pull/4632/commits/b15ec81f782f1d905893b47ed8c48a7a20df7822

I've also updated the `clients` PR to remove the `import` endpoint reference in `api.service` that is unused.

src/Api/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -312,31 +311,6 @@ await _providerBillingService.ScaleSeats(
         await _organizationService.DeleteAsync(organization);
     }
 
-    [HttpPost("{id}/import")]
-    public async Task Import(string id, [FromBody] ImportOrganizationUsersRequestModel model)
```
eliykat, line 316: You can also delete `ImportOrganizationUsersRequestModel`, which doesn't appear to be used anywhere else. (The public api has its own model.)

eliykat (changes requested):
1 minor comment below, also tests are still failing.

Otherwise I think this is great and takes a few steps in the right direction for this area of the code 👍

addisonbeck:
Tests are green now

eliykat (approved):

PR #4705, opened 2024-08-27
[PM-5450] Add check for admin/org access for events

nick-livefront:
## 🎟️ Tracking

[PM-5450](https://bitwarden.atlassian.net/browse/PM-5450)

## 📔 Objective

Issue: When a user only has access to a cipher from the `AllowAdminAccessToAllCollectionItems` policy on an organization, events aren't collected.
Fix: Add check for events to account for organizations that have `AllowAdminAccessToAllCollectionItems` enabled and fetch the cipher without the users id.

Respective [Client side PR](https://github.com/bitwarden/clients/pull/10678)

## 📸 Screenshots

https://github.com/user-attachments/assets/c4932d05-60ae-44a0-8eea-37c93c2ccc34

## 🦮 Reviewer guidelines

[PM-5450]: https://bitwarden.atlassian.net/browse/PM-5450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Events/Controllers/CollectController.cs
+++ b/src/Events/Controllers/CollectController.cs
@@ -1,4 +1,4 @@
-using Bit.Core.Context;
+﻿using Bit.Core.Context;
 using Bit.Core.Enums;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
@@ -19,19 +19,22 @@ public class CollectController : Controller
     private readonly ICipherRepository _cipherRepository;
     private readonly IOrganizationRepository _organizationRepository;
     private readonly IFeatureService _featureService;
+    private readonly IApplicationCacheService _applicationCacheService;
 
     public CollectController(
         ICurrentContext currentContext,
         IEventService eventService,
         ICipherRepository cipherRepository,
         IOrganizationRepository organizationRepository,
-        IFeatureService featureService)
+        IFeatureService featureService,
+        IApplicationCacheService applicationCacheService)
     {
         _currentContext = currentContext;
         _eventService = eventService;
         _cipherRepository = cipherRepository;
         _organizationRepository = organizationRepository;
         _featureService = featureService;
+        _applicationCacheService = applicationCacheService;
     }
 
     [HttpPost]
@@ -84,10 +87,16 @@ public class CollectController : Controller
                             continue;
                         }
 
-                        var org = await _organizationRepository.GetByIdAsync(eventModel.OrganizationId.Value);
+                        var org = _currentContext.GetOrganization(eventModel.OrganizationId.Value);
+                        var orgAbility = await _applicationCacheService.GetOrganizationAbilityAsync(eventModel.OrganizationId.Value);
                         cipher = await _cipherRepository.GetByIdAsync(eventModel.CipherId.Value);
 
-                        if (!org.AllowAdminAccessToAllCollectionItems || cipher == null)
+                        var cipherBelongsToOrg = cipher.OrganizationId == eventModel.OrganizationId;
+                        var customOrgAllowsAnyCollection = org is { Type: OrganizationUserType.Custom, Permissions.EditAnyCollection: true };
+                        var orgAllowsOwnerAdminAccess = orgAbility is { AllowAdminAccessToAllCollectionItems: true } && org is
+                        { Type: OrganizationUserType.Admin or OrganizationUserType.Owner };
+
+                        if (!cipherBelongsToOrg || !(customOrgAllowsAnyCollection || orgAllowsOwnerAdminAccess))
                         {
                             continue;
                         }
--- a/src/Events/Controllers/CollectController.cs
+++ b/src/Events/Controllers/CollectController.cs
@@ -1,4 +1,4 @@
-﻿using Bit.Core.Context;
+using Bit.Core.Context;
 using Bit.Core.Enums;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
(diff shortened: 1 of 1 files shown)


src/Events/Controllers/CollectController.cs
```diff
@@ -77,7 +77,15 @@ public async Task<IActionResult> Post([FromBody] IEnumerable<EventModel> model)
                     }
                     if (cipher == null)
                     {
-                        continue;
+                        // When the user cannot access the cipher directly, check if the organization allows for
+                        // admin/owners access to all collections and the user can access the cipher from that perspective.
+                        var org = await _organizationRepository.GetByIdAsync(eventModel.OrganizationId.Value);
```
shane-melton, line 82: 🎨 We have a service `IApplicationCacheService` that provides a method called `GetOrganizationAbilityAsync()` that accepts and `organizationId`. It uses a cache to reduce the number of queries against the database for commonly used properties on organizations.

We should try and use that here instead of the `organizationRepoistory` if possible.

shane-melton: ⛏️ We should have a check in case the `organizationId` is missing, as is this looks like it will throw an exception trying to access the `.Value`.

src/Events/Controllers/CollectController.cs
```diff
@@ -77,7 +77,15 @@ public async Task<IActionResult> Post([FromBody] IEnumerable<EventModel> model)
...
+                        cipher = await _cipherRepository.GetByIdAsync(eventModel.CipherId.Value);
+
+                        if (!org.AllowAdminAccessToAllCollectionItems || cipher == null)
```
shane-melton, line 85: ⛏️ We should also have a "sanity" check to be sure the returned cipher belongs to the organization we checked. If they don't match something is wrong and we should skip.

shane-melton, line 85: ⚠️ We don't actually know if the current context User belongs to the `eventModel.OrganizationId` and is an Admin to know if the `AllowAdminAccessToAllCollectionItems` settings applies to them. The `_currentContext` has a method called `GetOrganization(orgId)` that will return the org if the user is a member in some capacity. It returns an object that we can check the user `Type` of to be sure they're and Admin/Owner. 

Note: We **cannot** use the `_currentContext.OrganizationAdmin()` helper because it will return `true` for providers who should *not* have access to organization ciphers.

A good example to reference is here: https://github.com/bitwarden/server/blob/0a6c35e56c0e0ffa4368ec81fbfb0eb9eeb9cd8d/src/Api/Vault/Controllers/CiphersController.cs#L352-L367

shane-melton (changes requested):
Thanks for jumping on this! I left some comments on how to lock this down with our current permission practices. As always, if you have any questions please let me know!

src/Events/Controllers/CollectController.cs
nick-livefront: 👍 [7fbd875](https://github.com/bitwarden/server/pull/4705/commits/7fbd875a6e01ba1d150df36275412174fc0d902e)

src/Events/Controllers/CollectController.cs
nick-livefront: >We should have a check in case the organizationId is missing, as is this looks like it will throw an exception trying to access the .Value.

👍 [7fbd875](https://github.com/bitwarden/server/pull/4705/commits/7fbd875a6e01ba1d150df36275412174fc0d902e)

src/Events/Controllers/CollectController.cs
nick-livefront: Great context, thanks! I believe I have addressed the both this comment and your [above comment](https://github.com/bitwarden/server/pull/4705#discussion_r1733239111) following the same pattern you linked here. Let me know if I missed anything!

[7c2d18a](https://github.com/bitwarden/server/pull/4705/commits/7c2d18ae7af58fc217c1aab38295e529d546b8d8)

shane-melton (approved):
Looks good, thank you!

addisonbeck (dismissed):

src/Events/Controllers/CollectController.cs
```diff
@@ -77,7 +80,20 @@ public async Task<IActionResult> Post([FromBody] IEnumerable<EventModel> model)
...
+                        // admin/owners access to all collections and the user can access the cipher from that perspective.
+                        if (!eventModel.OrganizationId.HasValue)
+                        {
+                            continue;
+                        }
+
+                        cipher = await _cipherRepository.GetByIdAsync(eventModel.CipherId.Value);
+                        var cipherBelongsToOrg = cipher.OrganizationId == eventModel.OrganizationId;
+
+                        if (!cipherBelongsToOrg || cipher == null)
```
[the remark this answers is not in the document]
shane-melton: Ideally, we could re-use the authorization helpers in the `CiphersController` (we have a ticket to make these more accessible [PM-10296](https://bitwarden.atlassian.net/browse/PM-10296)). 

But, for the time being, we still need to ensure the user belongs to the cipher's organization by checking that `_currentContext.GetOrganization(eventModel.OrganizationId.Value)` returns a value.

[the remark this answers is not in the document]
nick-livefront: Thanks for keeping me inline, @shane-melton! [50ce7a6](https://github.com/bitwarden/server/pull/4705/commits/50ce7a635e4004b78f5efbb5c6bb1def213b4918)

shane-melton (approved):
Looks good!

PR #4701, opened 2024-08-27
[PM-10365] Drop OrganizationUser AccessAll

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10365
## 📔 Objective

Final step in removing OrganizationUser.AccessAll column:

* MSSQL: (1) add a default column value, (2) update sprocs to remove the column, (3) drop the column
  * also drop some old sprocs no longer in use and the TVP they used
* EF: remove shadow property configuration and run migrations

**WARNING:** this includes an index rebuild on the `OrganizationUsers` table, because the index previously included the `AccessAll` column. I have dropped & recreated it without this column on the assumption it's still useful for the remaining columns in the index. However we may need to coordinate this index rebuild with BRE.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #4709, opened 2024-08-28
Split `LimitCollectionCreationDeletion` into two database columns

addisonbeck:
### 📚 Stacked PRs

1. `server`: https://github.com/bitwarden/server/pull/4709
    ⬆️ __YOU ARE HERE__
1. `server`: https://github.com/bitwarden/server/pull/4730
1. `clients`: https://github.com/bitwarden/clients/pull/11223
1. `server`: https://github.com/bitwarden/server/pull/4808
1. `server`: https://github.com/bitwarden/server/pull/4809
1. `clients`: https://github.com/bitwarden/clients/pull/11258
1. `server`: https://github.com/bitwarden/server/pull/4810

### 🤳 The Bigger Picture 

There is an option in the Admin Console that removes collection creation and
deletion controls from all organization members that are not in the Owner and
Admin roles. This PR stack splits this conjoined option into two separate
controls: one to block `create` operations and one to block `delete`
operations. This is a part of an effort to make these settings better align
with what customers expect it to do.

### 👉 This Pull Request 👍

This pull request submits an initial migration & a transition migration for
introducing a `dbo.Organization.LimitCollectionCreation` column and a
`dbo.Organization.LimitCollectionDeletion` column that will split
`dbo.Organization.LimitCollectionCreationDeletion` into two individual data
points.

I've include the data-syncing `UPDATE` script that copies the existing
column's data over to the new column as a separate script from the initial
migration. This `UPDATE` script seems like it might fit the criteria of a
"transition" phase migration in EDD. I think it would be fine to run with the
initial migration, but I am unsure if (quoting the contributing docs)

> [the script would] be too slow or put too much load on the database, or
> otherwise make it unsuitable for the Initial migration

It seems like it wouldn't be too much to me, but let me know!

The same updates have been made to the entity framework database
implementations. The entity fields are synced by a hand written migration,
and are `get` and `set` to pull from `LimitCollectionCreationDeletion` until
the next phase of this work that will fade that field out of use in the
server project.

### 🧮 Side Effects

- I deleted a test class that was meant to be ephemeral and was blocking new
  EF migrations from being properly handled by tests.
- I changed the default value of `LimitCollectionCreationDeletion` from
  `true` to `false` in Entity Framework. This is already the default in
  MSSQL, and I think it was missed in EF.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10863]: https://bitwarden.atlassian.net/browse/PM-10863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


addisonbeck:
@r-tome @eliykat @vincentsalucci This isn't ready for merge yet. I want to get a bit further and manually test. That said: if anyone finds time for a review I'd appreciate it. I haven't done a formally processed EDD migration yet, and I do not want to miss anything. Thanks!

eliykat (commented):
You need to refresh `OrganizationView` after adding the column; but this otherwise looks right to me, including EDD considerations.

To confirm, I expect the timeline would be:
1. Merge this PR
-- rc cut --
2. Update the code to use the new column
-- rc cut --
3. Drop the old column

addisonbeck:
Thanks! I added a refresh to `OranizationView` on `725dbfd`

r-tome (dismissed):
LGTM!

src/Core/AdminConsole/Entities/Organization.cs
```diff
@@ -94,6 +94,9 @@ public class Organization : ITableObject<Guid>, IStorableSubscriber, IRevisable,
     /// they have Can Manage permissions for.
     /// </summary>
     public bool LimitCollectionCreationDeletion { get; set; }
+    public bool LimitCollectionCreation { get => LimitCollectionCreationDeletion; set => LimitCollectionCreationDeletion = value; }
+    public bool LimitCollectionDeletion { get => LimitCollectionCreationDeletion; set => LimitCollectionCreationDeletion = value; }
```
eliykat, line 98: I don't think this is necessary because it's handled at the db layer. The values are `COALESCE`d with `LimitCollectionCreationDeletion` on both read and write operations, and I think that's best to do in one place only.

src/Infrastructure.EntityFramework/AdminConsole/Configurations/OrganizationEntityTypeConfiguration.cs
```diff
@@ -12,9 +12,18 @@ public void Configure(EntityTypeBuilder<Organization> builder)
             .Property(o => o.Id)
             .ValueGeneratedNever();
 
+        // Deprecated https://bitwarden.atlassian.net/browse/PM-10863
         builder.Property(c => c.LimitCollectionCreationDeletion)
             .ValueGeneratedNever()
-            .HasDefaultValue(true);
+            .HasDefaultValue(false);
```
eliykat, line 18: I thought EF generated its properties based on the `Entity` class - what does this configuration class do? Is it required to set the default value?

src/Sql/dbo/Stored Procedures/Organization_Create.sql
```diff
@@ -168,7 +175,9 @@ BEGIN
         @MaxAutoscaleSmSeats,
         @MaxAutoscaleSmServiceAccounts,
         @SecretsManagerBeta,
-        @LimitCollectionCreationDeletion,
+        COALESCE(@LimitCollectionCreation, @LimitCollectionDeletion, 0), -- Deprecated https://bitwarden.atlassian.net/browse/PM-10863)
```
eliykat, line 178: Thinking this through to make sure I understand:

* combination of current code + new db: `@LimitCollectionCreation` and `@LimitCollectionDeletion` are NULL, but they both take the value of `LimitCollectionCreationDeletion` due to the `COALESCE` above. This line ends up writing the current value correctly, even though it doesn't reference `@LimitCollectionCreationDeletion` directly. (And we also set the new properties accordingly.)

* combination of new code + new db: `@LimitCollectionCreationDeletion` will effectively continue to be updated based on the value of `@LimitCollectionCreation`. It won't be used in the new code, so it's only being updated in case there's a rollback scenario. In that case it's not really clear which value it should track given that it doesn't map to either perfectly, but I guess that's still better than it being `NULL`ed out by default.

I think this is OK but please correct me if I'm missing anything. I think this is worth a regression sweep from QA after merging to `main` in any case just to make sure.

eliykat (changes requested):
Great work on this! Mostly questions for my own understanding / confirmation, then I think it's ready for Robert's review.

src/Core/AdminConsole/Entities/Organization.cs
addisonbeck: They are `COALESCE`d in the Dapper implementation, but without the use of stored procedures in the Entity Framework implementation we need to keep the values in sync some other way. This seemed like the least intrusive way to do that, to me.

src/Infrastructure.EntityFramework/AdminConsole/Configurations/OrganizationEntityTypeConfiguration.cs
addisonbeck: Yeah, it seems like this is mostly just to set the default value. I don't think it's necessary if the default is `false` - so I've removed these on the latest changes.

src/Sql/dbo/Stored Procedures/Organization_Create.sql
addisonbeck: 💯

> In that case it's not really clear which value it should track 

The way it is written now this scenario would result in `LimitCollectionCreationDeletion` being "rolled back" to the value of `LimitCollectionCreation` specifically since it is the first written conditional in the `COALESCE` statement.

src/Infrastructure.EntityFramework/AdminConsole/Configurations/OrganizationEntityTypeConfiguration.cs
eliykat: Now the column is non-nullable with no default value though - that's going to throw errors when you remove it from the code but haven't removed it from the db yet. (Even in EF we persist the column for an additional release as a shadow property after it's removed from the Entity class.) Bools will default to `false` in C# but I think you still need the default value for the database itself.

src/Core/AdminConsole/Entities/Organization.cs
eliykat: This comment made me realise - usually I do my MSSQL migrations 1 release before the code changes, but I do the EF migrations at the same time as the code changes, given that it's code-first and doesn't have to support EDD. That would mean there are no changes to the C# code at all (including this class), just silently adding the MSSQL column and keeping it in sync.

I think this approach should also work but I haven't tried it myself. As long as EF uses the getters and setters properly (no tricky reflection) then I agree it should stay in sync.

src/Infrastructure.EntityFramework/AdminConsole/Configurations/OrganizationEntityTypeConfiguration.cs
addisonbeck: That isn't something this code was doing. [The migrations that included these configuration references](https://github.com/bitwarden/server/commit/5e231f770d7c87a9b543463b8ba20614160def63#diff-66eb2e465dc54617072b9061c488289da865a0e77a0dbd22d384ea0c719351ceR13-R35) still resulted in `nullable: false` database columns. I'm not sure we really want nullable database columns here - that is never really a valid state for the data in EF land.

To keep the column around for a release event after we remove the field from code we would want a shadow property, but that should still go on the entity. On that change I'll just reverse the way [these values](https://github.com/bitwarden/server/pull/4709/files#diff-dc77aef59f6aee693dac228d795b9af1254ad275970d861bffbc4c5888d52effR96-R97) mirror each other so that the old field gets/sets from the new one. I'd also move the declaration of the properties to the EF specific entity model to avoid still trying to pass it to Dapper. The release after will remove the old property from the entity (removing it from EF) and remove it from SQL. 

Does that sound right?

src/Core/AdminConsole/Entities/Organization.cs
addisonbeck: That makes sense. I did the same thing at first, but decided to try this instead to keep the database types as 1:1 as possible across releases. We can cut it out and just do the SQL here if that's anybody else's preference.

eliykat: Nope, I think your approach here is a valid one as well!

src/Infrastructure.EntityFramework/AdminConsole/Configurations/OrganizationEntityTypeConfiguration.cs
eliykat: I think it's a good move to use shadow properties for this change - it confirms that the first stage of the EDD cycle contains db changes only.

I don't think I wanted it to be nullable, more that I thought it needed to have a default value. But I see how you can control that through the shadow property definition, which is neat. So you've addressed my concerns here.

eliykat (approved):
This looks good to me! Thanks for taking the time.

addisonbeck:
I rebased this on `main` to update migration script dates and keep everything in sync for branches further down the tree than this one like https://github.com/bitwarden/server/pull/4730

No functional changes were made.

addisonbeck:
I'd also like to ignore the [failing SonarCloud action run](https://github.com/bitwarden/server/pull/4709/checks?check_run_id=30669364006) on this PR. It seems mostly upset about tests, but this entirely database operations.

withinfocus:
I opened up the Sonar report finding, and there's one thing listed:

https://sonarcloud.io/project/issues?severities=BLOCKER&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=bitwarden_server&open=AZH6gLSYkzQSuElBzVms

It suggests a `WHERE` with your update.

addisonbeck:
Thank you @withinfocus, that's helpful and did give me an idea. I _do_ want to update every record on the table, so a `WHERE` clause isn't wrong logically. But, it cuts down the amount of records accessed to only check for rows that don't match the default value of the column being updated. That sounds like a good enough reason to add one. 

@eliykat this does mark a small functional change since review - please give a review of the very small change on https://github.com/bitwarden/server/pull/4709/commits/c37f77f645877b7ad96c3cd53e61c5e06df270a4

eliykat (approved):
LGTM, also note my recent database changes in https://github.com/bitwarden/server/pull/4701 just in case you need to re-do any of your migrations. (You may have already done this)

rkac-bw (approved):
lgtm

PR #4712, opened 2024-08-29
[PM-8445] Update trial initiation UI

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8445

## 📔 Objective

The goal is to implement a feature flag that makes the payment step of the trial flow optional, display a visual indicator showing remaining trial time, allow the addition of payment details before the trial ends, and enhance the UI for organisation’s that are disabled due to failed or non-payment.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Billing/Services/Implementations/OrganizationSubscriptionService.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace Bit.Core.Billing.Services.Implementations;
+
+public class OrganizationSubscriptionService(
+    IGlobalSettings globalSettings,
+    ILogger<SubscriberService> logger,
+    IStripeAdapter stripeAdapter) : IOrganizationSubscriptionService
+{
+    private const string SecretsManagerStandaloneDiscountId = "sm-standalone";
+
+    public async Task<string> PurchaseOrganizationNoPaymentMethod(Organization org, StaticStore.Plan plan,
```
amorask-bitwarden, line 21: ❌ Just mentioning this here even though we discussed this on Slack, but this new service won't be necessary. The idea behind the `OrganizationBillingService` is that we should completely centralize all of the logic around purchasing organizations so that our team can manage the contract and logic. 

After [this](https://github.com/bitwarden/server/pull/4800) is merged, all of this can be accomplished with the following:
```csharp
var sale = OrganizationSale.From(organization, signup);
await organizationBillingService.Finalize(sale);
```
This will already handle all of the logic implemented here.

amorask-bitwarden (changes requested):

r-tome (commented):
Looks good to me, I'll review it again after you make the changes requested by @amorask-bitwarden

src/Api/AdminConsole/Models/Request/Organizations/OrganizationNoPaymentCreateRequest.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Bit.Api.AdminConsole.Models.Request.Organizations;
+
+public abstract class OrganizationNoPaymentCreateRequest : OrganizationCreateRequestBase
```
amorask-bitwarden, line 3: ❓ What is the purpose of this request model? It doesn't appear to ever be used. If we're not using this request model, then it's unclear why we need an `OrganizationCreateRequestBase` and `OrganizationCreateRequestModel`.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -591,9 +591,9 @@ await _referenceEventService.RaiseEventAsync(
             else
             {
                 await _paymentService.PurchaseOrganizationAsync(organization, signup.PaymentMethodType.Value,
```
amorask-bitwarden, line 593: ❌  My understanding of this feature is that we want to be able to start an organization subscription without a payment method. The underlying code for the `OrganizationBillingService.Finalize(Sale)` handles that, but when the Stripe deprecation feature flag is off, I don't think the `StripePaymentService` will handle that. The [method signature](https://github.com/bitwarden/server/blob/3c86ec6a35fd6b4995214124e0d5b8eb356ffd5c/src/Core/Services/Implementations/StripePaymentService.cs#L54-L57) actually demands a non-nullable `PaymentMethodType`, which is why it's being passed `signup.PaymentMethodType.Value`. But if your request is no longer validating that the incoming `PaymentMethodType` is actually provided, by design, then you're going to get a null reference exception here.

amorask-bitwarden (changes requested):

src/Api/AdminConsole/Models/Request/Organizations/OrganizationNoPaymentCreateRequest.cs
cyprain-okeke: when a client make a request for an organization that does not require payment method, it will be used while  OrganizationCreateRequestModel  works for existing flow where payment method is required and validated.

amorask-bitwarden (dismissed):
Billing stuff looks good, just request @bitwarden/team-admin-console review and you're ready to go.

src/Api/AdminConsole/Models/Request/Organizations/OrganizationCreateRequestModel.cs
```diff
@@ -14,42 +14,56 @@ public class OrganizationCreateRequestModel : IValidatableObject
...
+
     public string BillingAddressCity { get; set; }
+
     public string BillingAddressState { get; set; }
+
     public string BillingAddressPostalCode { get; set; }
-    [StringLength(2)]
-    public string BillingAddressCountry { get; set; }
+
+    [StringLength(2)] public string BillingAddressCountry { get; set; }
```
r-tome, line 57: ⛏️ Optional: Please break a line on the data annotations so the format is the same as everywhere else

r-tome (dismissed):
Looks good, just a minor optional nitpick

r-tome (approved):

amorask-bitwarden (approved):

PR #4726, opened 2024-09-03
[PM-10316] Add Command to Remove User and Delete Data for Organization-Managed Users

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10316
https://bitwarden.atlassian.net/browse/PM-10337

## 📔 Objective

Create a command that does the following:
1. Takes an organizationId and an organizationUserId as inputs.
2. Checks if the user, identified by organizationUserId, is managed by the organization identified by organizationId.
3. If the user is managed by the organization, remove the user from the organization.
4. Delete the user's account data.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


addisonbeck (dismissed):

addisonbeck (approved):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
eliykat, line 1: The validation logic in this command needs to be built out a little more.

It should cover all the same cases as removing an organization user, specifically:
* you cannot remove/delete yourself
* a non-owner cannot remove/delete an owner
* you cannot delete the last owner of an organization

See:
https://github.com/bitwarden/server/blob/e5c77d5f90efa399e632dd53988b73cf416d0a61/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1623-L1623

(To clarify, I'm not suggesting that you call this code exactly, I think it's OK to repeat the checks even though it's not totally DRY.)

Additionally, as discussed on Slack, you cannot delete a user if they're in the Invited status. This is incidentally covered by `if (!orgUser.UserId.HasValue)` but should be explicit.

Plus the checks you already have here (e.g. managed status).

I suggest putting these in a private method that can be called by both public methods, to avoid duplication and make sure validation is consistent.

src/Core/AdminConsole/Enums/EventType.cs
```diff
@@ -58,6 +58,8 @@ public enum EventType : int
     OrganizationUser_Restored = 1512,
     OrganizationUser_ApprovedAuthRequest = 1513,
     OrganizationUser_RejectedAuthRequest = 1514,
+    OrganizationUser_Deleted = 1515, // Both user and organization user data were deleted
+    OrganizationUser_Left = 1516,    // User voluntarily left the organization
```
eliykat, line 62: Is this getting wired up in a separate PR?

eliykat (changes requested):
I think this worked much better as a separate command 👍

src/Core/AdminConsole/Enums/EventType.cs
r-tome: Yes, I'll do that here https://bitwarden.atlassian.net/browse/PM-10338

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -13,87 +14,102 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
...
+                results.Add((orgUserId, e.Message));
             }
         }
 
         return results;
     }
+
+    private async Task RepositoryDeleteUserAsync(Guid organizationId, OrganizationUser orgUser, Guid? deletingUserId, IDictionary<Guid, bool> managementStatus = null)
+    {
+        if (deletingUserId.HasValue && orgUser.UserId.Value == deletingUserId.Value)
```
eliykat, line 78: Possible null reference exception on `orgUser.UserId.Value`. You should be able to fix this by moving it below the next `if` block (which checks for `HasValue`).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -13,87 +14,102 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
...
+        {
+            if (deletingUserId.HasValue && !await _currentContext.OrganizationOwner(organizationId))
+            {
+                throw new BadRequestException("Only owners can delete other owners.");
+            }
+
+            if (!await _organizationService.HasConfirmedOwnersExceptAsync(organizationId, new[] { orgUser.Id }, includeProvider: true))
+            {
+                throw new BadRequestException("Organization must have at least one confirmed owner.");
+            }
```
eliykat, line 98: `HasConfirmedOwnersExceptAsync` contains a database call, this is going to hit the db once per user when it's called in a loop. Also, it doesn't work when it's called separately like this - if you're deleting all 3 owners of an organization, it will always think there's another 2 owners. It should be called once, with all the users you want to delete.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -13,87 +14,102 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
...
+        }
+
+        if (!managementStatus.TryGetValue(orgUser.Id, out var isManaged) || !isManaged)
```
eliykat, line 101: Possible null reference exception on `managementStatus`. It has a default value of `null` which doesn't look like it's needed. Would the query ever return `null`?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
eliykat, line 1: Please use `#nullable enable`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -13,87 +14,102 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
...
+        }
+
+        var userToDelete = await _userService.GetUserByIdAsync(orgUser.UserId.Value);
+        if (userToDelete == null)
+        {
+            throw new NotFoundException("User not found.");
+        }
+
+        await _userService.DeleteAsync(userToDelete);
+        await _eventService.LogOrganizationUserEventAsync(orgUser, EventType.OrganizationUser_Deleted);
```
eliykat, line 113: `eventService` has an override for multiple organizationUser events, we should use that in our bulk method.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -13,87 +14,102 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
...
+                throw new BadRequestException("Organization must have at least one confirmed owner.");
+            }
+        }
+
+        if (!managementStatus.TryGetValue(orgUser.Id, out var isManaged) || !isManaged)
+        {
+            throw new BadRequestException("User is not managed by the organization.");
+        }
+
+        var userToDelete = await _userService.GetUserByIdAsync(orgUser.UserId.Value);
```
eliykat, line 106: There's a bulk repository method: `UserRepository.GetManyAsync`.

In fact, `UserService.GetUserByIdAsync` is meant to get the current user only and cache it in `CurrentContext`, so it definitely shouldn't be used here.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -13,87 +14,102 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
...
             {
-                results.Add((orgUser, e.Message));
+                results.Add((orgUserId, e.Message));
             }
         }
 
         return results;
     }
+
+    private async Task RepositoryDeleteUserAsync(Guid organizationId, OrganizationUser orgUser, Guid? deletingUserId, IDictionary<Guid, bool> managementStatus = null)
```
eliykat, line 76: This method uses exception handling for control flow. I suggest it returns a `string` instead, containing any possible error message. That also means you don't need the try/catch block above, you can always add the return value into the `results` dict.

eliykat (changes requested):
To summarise my comments below, I think `RepositoryDeleteUserAsync` isn't quite the right abstraction - it doesn't work when called in a loop for each user.

I was going to suggest a method like `ValidateAsync(Guid organizationId, IEnumerable<Guid> organizationUserIds, Guid? deletingUserId)` which validates multiple at once, avoiding the issues with validating them one-by-one.  However, it may be tricky to return individualised validation errors that way.

In any case, it seems like there needs to be some unbundling of the `RepositoryDeleteUserAsync` method, so that the bulk public method can make use of bulk methods on other services and repositories.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (deletingUserId.HasValue && orgUser.UserId.Value == deletingUserId.Value)
+        {
+            throw new BadRequestException("You cannot delete yourself.");
+        }
+
+        if (!orgUser.UserId.HasValue || orgUser.Status == OrganizationUserStatusType.Invited)
+        {
+            throw new BadRequestException("You cannot delete a user with Invited status.");
```
eliykat, line 85: Sukhleen provided updated wording here: https://bitwarden.atlassian.net/browse/PM-10315?focusedCommentId=71066

> Cannot delete members with the Invited status

eliykat: Please also update all other error wording with `member` rather than `user`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
r-tome: I chose to keep throwing exceptions since it aligns with our usual practice for single methods.

r-tome:
@eliykat great feedback, thank you! I have addressed your suggestions and reduced the number of database calls.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
eliykat: True! I still think it's awkward but it doesn't need to be a blocker.

eliykat (approved):
Great work @r-tome, thanks for working through the extra changes!

PR #4730, opened 2024-09-03
Split `Organization.LimitCollectionCreationDeletion` into two separate business rules

addisonbeck:
### 📚 Stacked PRs

1. `server`: https://github.com/bitwarden/server/pull/4709
1. `server`: https://github.com/bitwarden/server/pull/4730
    ⬆️ __YOU ARE HERE__
1. `clients`: https://github.com/bitwarden/clients/pull/11223
1. `server`: https://github.com/bitwarden/server/pull/4808
1. `server`: https://github.com/bitwarden/server/pull/4809
1. `clients`: https://github.com/bitwarden/clients/pull/11258
1. `server`: https://github.com/bitwarden/server/pull/4810

### 🤳 The Bigger Picture 

There is an option in the Admin Console that removes collection creation and
deletion controls from all organization members that are not in the Owner and
Admin roles. This PR stack splits this conjoined option into two separate
controls: one to block `create` operations and one to block `delete`
operations. This is a part of an effort to make these settings better align
with what customers expect it to do.

### 👉 This Pull Request 👍

In the previous PR we added two new database columns to facilitate a rename &
split operation on an existing column. In this PR we go about implementing
the new columns in `server` models and implementing the desired functionality
of those new properties throughout business logic and tests.

This PR and the two following it encompass all of the functional changes to
the system implemented by this PR stack. The next PR implements the feature
in the admin console UI in `clients`, and the one after that enables the
feature for self hosted installations. The first and second PRs are
maintenance tasks to support evolutionary database design and entity
framework.

### 🧮 Side Effects

There are several side effects involved in this PR. This includes:

1. Updates to the front end of the admin portal to support the new
   organization properties.
1. Updates to license files. We're removing these settings and
   `AllowAdminsAccessToAllCollections` from the license file as part of this
   work.
1. A lot of difficult to follow updates to tests - mostly around licensing
   backwards compatibility. 

## 📸 Screenshots

Several recorded demos can be found in Jira

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10863]: https://bitwarden.atlassian.net/browse/PM-10863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Api/AdminConsole/Controllers/OrganizationsController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationsController.cs
@@ -514,7 +514,6 @@ public class OrganizationsController : Controller
     }
 
     [HttpPut("{id}/collection-management")]
-    [SelfHosted(NotSelfHostedOnly = true)]
     public async Task<OrganizationResponseModel> PutCollectionManagement(Guid id, [FromBody] OrganizationCollectionManagementUpdateRequestModel model)
     {
         var organization = await _organizationRepository.GetByIdAsync(id);
--- a/src/Api/AdminConsole/Controllers/OrganizationsController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationsController.cs
@@ -535,7 +535,7 @@ public class OrganizationsController : Controller
             throw new NotFoundException();
         }
 
-        await _organizationService.UpdateAsync(model.ToOrganization(organization), eventType: EventType.Organization_CollectionManagement_Updated);
+        await _organizationService.UpdateAsync(model.ToOrganization(organization, _featureService), eventType: EventType.Organization_CollectionManagement_Updated);
         return new OrganizationResponseModel(organization);
     }
 }
--- a/src/Api/AdminConsole/Controllers/OrganizationsController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationsController.cs
@@ -516,6 +516,14 @@ public class OrganizationsController : Controller
     [HttpPut("{id}/collection-management")]
     public async Task<OrganizationResponseModel> PutCollectionManagement(Guid id, [FromBody] OrganizationCollectionManagementUpdateRequestModel model)
     {
+        if (
+          _globalSettings.SelfHosted &&
+          !_featureService.IsEnabled(FeatureFlagKeys.LimitCollectionCreationDeletionSplit)
+        )
+        {
+            throw new BadRequestException("Only allowed when not self hosted.");
+        }
+
         var organization = await _organizationRepository.GetByIdAsync(id);
         if (organization == null)
         {
--- a/src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
+++ b/src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
@@ -55,6 +55,9 @@ public class OrganizationResponseModel : ResponseModel
         SmServiceAccounts = organization.SmServiceAccounts;
         MaxAutoscaleSmSeats = organization.MaxAutoscaleSmSeats;
         MaxAutoscaleSmServiceAccounts = organization.MaxAutoscaleSmServiceAccounts;
+        LimitCollectionCreation = organization.LimitCollectionCreation;
+        LimitCollectionDeletion = organization.LimitCollectionDeletion;
+        // Deperectated: https://bitwarden.atlassian.net/browse/PM-10863
         LimitCollectionCreationDeletion = organization.LimitCollectionCreationDeletion;
         AllowAdminAccessToAllCollectionItems = organization.AllowAdminAccessToAllCollectionItems;
     }
@@ -98,6 +101,9 @@ public class OrganizationResponseModel : ResponseModel
     public int? SmServiceAccounts { get; set; }
     public int? MaxAutoscaleSmSeats { get; set; }
     public int? MaxAutoscaleSmServiceAccounts { get; set; }
+    public bool LimitCollectionCreation { get; set; }
+    public bool LimitCollectionDeletion { get; set; }
+    // Deperectated: https://bitwarden.atlassian.net/browse/PM-10863
     public bool LimitCollectionCreationDeletion { get; set; }
     public bool AllowAdminAccessToAllCollectionItems { get; set; }
 }
--- a/src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
+++ b/src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
@@ -57,7 +57,7 @@ public class OrganizationResponseModel : ResponseModel
         MaxAutoscaleSmServiceAccounts = organization.MaxAutoscaleSmServiceAccounts;
         LimitCollectionCreation = organization.LimitCollectionCreation;
         LimitCollectionDeletion = organization.LimitCollectionDeletion;
-        // Deperectated: https://bitwarden.atlassian.net/browse/PM-10863
+        // Deprecated: https://bitwarden.atlassian.net/browse/PM-10863
         LimitCollectionCreationDeletion = organization.LimitCollectionCreationDeletion;
         AllowAdminAccessToAllCollectionItems = organization.AllowAdminAccessToAllCollectionItems;
     }
--- a/src/Api/Models/Request/Organizations/OrganizationCollectionManagementUpdateRequestModel.cs
+++ b/src/Api/Models/Request/Organizations/OrganizationCollectionManagementUpdateRequestModel.cs
@@ -4,11 +4,17 @@ namespace Bit.Api.Models.Request.Organizations;
 
 public class OrganizationCollectionManagementUpdateRequestModel
 {
+    public bool LimitCollectionCreation { get; set; }
+    public bool LimitCollectionDeletion { get; set; }
+    // Deprecated: https://bitwarden.atlassian.net/browse/PM-10863
     public bool LimitCreateDeleteOwnerAdmin { get; set; }
     public bool AllowAdminAccessToAllCollectionItems { get; set; }
 
     public virtual Organization ToOrganization(Organization existingOrganization)
     {
+        existingOrganization.LimitCollectionCreation = LimitCollectionCreation;
+        existingOrganization.LimitCollectionDeletion = LimitCollectionDeletion;
+        // Deprecated: https://bitwarden.atlassian.net/browse/PM-10863
         existingOrganization.LimitCollectionCreationDeletion = LimitCreateDeleteOwnerAdmin;
         existingOrganization.AllowAdminAccessToAllCollectionItems = AllowAdminAccessToAllCollectionItems;
         return existingOrganization;
--- a/src/Api/Models/Request/Organizations/OrganizationCollectionManagementUpdateRequestModel.cs
+++ b/src/Api/Models/Request/Organizations/OrganizationCollectionManagementUpdateRequestModel.cs
@@ -1,4 +1,6 @@
-﻿using Bit.Core.AdminConsole.Entities;
+﻿using Bit.Core;
+using Bit.Core.AdminConsole.Entities;
+using Bit.Core.Services;
 
 namespace Bit.Api.Models.Request.Organizations;
 
@@ -10,12 +12,18 @@ public class OrganizationCollectionManagementUpdateRequestModel
     public bool LimitCreateDeleteOwnerAdmin { get; set; }
     public bool AllowAdminAccessToAllCollectionItems { get; set; }
 
-    public virtual Organization ToOrganization(Organization existingOrganization)
+    public virtual Organization ToOrganization(Organization existingOrganization, IFeatureService featureService)
     {
-        existingOrganization.LimitCollectionCreation = LimitCollectionCreation;
-        existingOrganization.LimitCollectionDeletion = LimitCollectionDeletion;
-        // Deprecated: https://bitwarden.atlassian.net/browse/PM-10863
-        existingOrganization.LimitCollectionCreationDeletion = LimitCreateDeleteOwnerAdmin;
+        if (featureService.IsEnabled(FeatureFlagKeys.LimitCollectionCreationDeletionSplit))
+        {
+            existingOrganization.LimitCollectionCreation = LimitCollectionCreation;
+            existingOrganization.LimitCollectionDeletion = LimitCollectionDeletion;
+        }
+        else
+        {
+            existingOrganization.LimitCollectionCreationDeletion = LimitCreateDeleteOwnerAdmin || LimitCollectionCreation || LimitCollectionDeletion;
+        }
+
         existingOrganization.AllowAdminAccessToAllCollectionItems = AllowAdminAccessToAllCollectionItems;
         return existingOrganization;
     }
--- a/src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
+++ b/src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
@@ -126,9 +126,9 @@ public class BulkCollectionAuthorizationHandler : BulkAuthorizationHandler<BulkC
 
         var organizationAbility = await GetOrganizationAbilityAsync(org);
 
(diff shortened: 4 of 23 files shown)


src/Core/AdminConsole/Entities/Organization.cs
```diff
@@ -300,7 +313,6 @@ public void UpdateFromLicense(OrganizationLicense license)
         UseSecretsManager = license.UseSecretsManager;
         SmSeats = license.SmSeats;
         SmServiceAccounts = license.SmServiceAccounts;
-        LimitCollectionCreationDeletion = license.LimitCollectionCreationDeletion;
```
eliykat, line 303: This method is used when **updating** an existing org from a license file. You've removed the `LimitCollectionCreationDeletion` assignment, but not `AllowAdminAccessToAllCollectionItems`, which is inconsistent.

Also, the `OrganizationService.SignUpAsync` method (which **creates** a new org from a license file) has this changed feature flagged, but here it's not feature flagged.

eliykat: I'm also not sure about the value of feature flagging self-hosted code given that we can't dynamically control it. It ends up being a code change to turn the feature flag on anyway.

src/Core/AdminConsole/Entities/Organization.cs
```diff
@@ -93,7 +93,20 @@ public class Organization : ITableObject<Guid>, IStorableSubscriber, IRevisable,
...
+    // `LimitCollectionDeletion`.
+    public bool LimitCollectionCreationDeletion
+    {
+        get => LimitCollectionCreation || LimitCollectionDeletion;
+        set
+        {
+            LimitCollectionCreation = value;
+            LimitCollectionDeletion = value;
+        }
+    }
```
eliykat, line 109: I think this is a clever way to handle this!

src/Core/Models/Business/OrganizationLicense.cs
eliykat, line 1: Please explain to Priya how this will work for older self-hosted customers. Not everyone upgrades immediately, there is likely to be a period where they see 1 setting in self-host but the 2 new settings in cloud, and still need to update it via the license file. Your mapping code will handle this by `OR`ing the 2 new values, but we may have to answer some questions about it in the meantime.

eliykat (changes requested):
There's a bit of complexity to this that I think you've handled really well 👍

src/Api/Models/Request/Organizations/OrganizationCollectionManagementUpdateRequestModel.cs
eliykat, line 1: Note that the endpoint for this (`OrganizationsController.PutCollectionManagement`) is still blocked for self-hosted. Does this need to be changed in this PR?

src/Core/Models/Business/OrganizationLicense.cs
eliykat: Relatedly - how are we timing the rollout? If we enable the feature flag in cloud first, then self-hosted customers will be in this state at least until the next self-hosted release. That's probably unavoidable but worth discussing.

src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
```diff
@@ -55,6 +55,9 @@ public OrganizationResponseModel(Organization organization, string obj = "organi
         SmServiceAccounts = organization.SmServiceAccounts;
         MaxAutoscaleSmSeats = organization.MaxAutoscaleSmSeats;
         MaxAutoscaleSmServiceAccounts = organization.MaxAutoscaleSmServiceAccounts;
+        LimitCollectionCreation = organization.LimitCollectionCreation;
+        LimitCollectionDeletion = organization.LimitCollectionDeletion;
+        // Deperectated: https://bitwarden.atlassian.net/browse/PM-10863
```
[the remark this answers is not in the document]
addisonbeck: So do I, but I did correct it anyway.

__Commit:__ [Fix mispelling](https://github.com/bitwarden/server/pull/4730/commits/c29f9f691ddcc450c9b2e0b3e7528842b66f88f4)

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -367,12 +381,6 @@ to the Organization object. It's validated as part of the hash but does not need
                         organization.SmServiceAccounts == SmServiceAccounts;
             }
 
-            /*
-             * Version 14 added LimitCollectionCreationDeletion and Version 15 added AllowAdminAccessToAllCollectionItems,
-             * however these are just user settings and it is not worth failing validation if they mismatch.
-             * They are intentionally excluded.
-             */
-
```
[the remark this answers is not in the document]
addisonbeck: Added as written

__Commit:__ [Re-add contextual comment regarding dropped license properties](https://github.com/bitwarden/server/pull/4730/commits/2433312f1c6a755949828b19913d7dd985357efd)

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepositoryTests.cs
```diff
@@ -253,7 +253,8 @@ public async Task GetManyDetailsByUserAsync_Works(IUserRepository userRepository
         Assert.Equal(orgUser1.Permissions, result.Permissions);
         Assert.Equal(organization.SmSeats, result.SmSeats);
         Assert.Equal(organization.SmServiceAccounts, result.SmServiceAccounts);
-        Assert.Equal(organization.LimitCollectionCreationDeletion, result.LimitCollectionCreationDeletion);
```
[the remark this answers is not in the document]
addisonbeck: Yes. I've added this back.

__Commit:__ [Add back deleted assertion for deprecated property](https://github.com/bitwarden/server/pull/4730/commits/d79cf59f4073215e9d53d15a04e7046d62e68d10)

test/Core.Test/Models/Business/OrganizationLicenseFileFixtures.cs
```diff
@@ -111,7 +111,5 @@ public static Organization OrganizationFactory() =>
             SmServiceAccounts = 8,
             MaxAutoscaleSmSeats = 101,
             MaxAutoscaleSmServiceAccounts = 102,
-            LimitCollectionCreationDeletion = true,
-            AllowAdminAccessToAllCollectionItems = true,
```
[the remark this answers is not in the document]
addisonbeck: Added back!

__Commit:__ [Add back removed fixture property assignment](https://github.com/bitwarden/server/pull/4730/commits/97961e1f440d990e79fc677e86fb5dcc2e2d7438)

src/Core/AdminConsole/Entities/Organization.cs
addisonbeck: I've added this removed property assignment back, and feature toggled both it and the `AllowAdminAccessToAllCollectionItems` assignment. This should look the same between `Create` and `Update` operations now.

__Commit:__ [Improve feature toggling scenerios for self hosted org creation/update](https://github.com/bitwarden/server/pull/4730/commits/aeeb796aed06565dfbba62a90d9676fcabef7eb7)

src/Core/AdminConsole/Entities/Organization.cs
addisonbeck: Thank you!

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -123,8 +124,14 @@ private async Task<bool> CanCreateAsync(CurrentContextOrganization? org)
             return true;
         }
 
+        var organizationAbility = await GetOrganizationAbilityAsync(org);
+
+        var limitCollectionCreationEnabled = !(_featureService.IsEnabled(FeatureFlagKeys.LimitCollectionCreationDeletionSplit)
+            ? organizationAbility is { LimitCollectionCreation: false }
+            : organizationAbility is { LimitCollectionCreationDeletion: false });
```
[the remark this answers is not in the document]
addisonbeck: This conditional logic is really messy, but changing it too much breaks tests. An exact implementation of your suggestion makes logical sense, and is what I tried to do at first, but something here isn't right and it seemed out of scope for me to spend a lot of time on it.

You can see this by looking at the Github Actions runs on the PR. They pass right before [this commit](https://github.com/bitwarden/server/pull/4730/commits/0a80eb70b997b1ee80cbed32feb1631162ebe9a5) which is an exact implementation of your suggestion that fails.

As for this specifically:

> Also the variable is called enabled but it's set to true if the setting is disabled.

That's not true. The variable is set to true if the setting is enabled. The right side of this equality operator is completely wrapped in an inversion, so the output of the whole ternary is inverted. The ternary evaluated to `true` if the setting is `false` and then that whole thing is flipped before the assignment to the variable.

Again, I know this is not very pleasant. I'm open to suggestions on how to improve this without getting off track too much from the work being done.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
```diff
@@ -256,8 +263,15 @@ private async Task<bool> CanDeleteAsync(ICollection<Collection> resources, Curre
...
-        var canDeleteManagedCollections = organizationAbility is { LimitCollectionCreationDeletion: false } ||
-                                          org is { Type: OrganizationUserType.Owner or OrganizationUserType.Admin };
+
+        var limitCollectionDeletionEnabled = !(_featureService.IsEnabled(FeatureFlagKeys.LimitCollectionCreationDeletionSplit)
+            ? organizationAbility is { LimitCollectionDeletion: false }
+            : organizationAbility is { LimitCollectionCreationDeletion: false });
+
+        var canDeleteManagedCollections =
+            !limitCollectionDeletionEnabled ||
+            org is { Type: OrganizationUserType.Owner or OrganizationUserType.Admin };
```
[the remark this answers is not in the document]
addisonbeck: Responded above

src/Core/Models/Business/OrganizationLicense.cs
addisonbeck: I'm planning to record some demos. I'll include one of an out-of-date self hosted combination and attach in Jira with an explanation. As for the rollout: what you said is what I was thinking. I don't know what else we could do other than release to self host with the feature on alongside cloud and hope for the best. Seems better to lag a little and explain.

src/Api/Models/Request/Organizations/OrganizationCollectionManagementUpdateRequestModel.cs
addisonbeck: I've unblocked this endpoint. Thanks for catching that.

__Commit:__ [Unblock PutCollectionManagement for self host](https://github.com/bitwarden/server/pull/4730/commits/d9703a0b01b0ba31fc9c02d3b28f21ea94f7068a)

src/Core/AdminConsole/Entities/Organization.cs
eliykat: This looks good to me now, it's all following the feature flag.

src/Api/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -514,7 +514,6 @@ public async Task<OrganizationSsoResponseModel> PostSso(Guid id, [FromBody] Orga
     }
 
     [HttpPut("{id}/collection-management")]
-    [SelfHosted(NotSelfHostedOnly = true)]
```
eliykat, line 517: This change isn't feature flagged - strictly speaking you'd have to check `!featureFlag && globalSettings.SelfHosted` in the body of the method and throw an error if required. However if we're just going to rely on the client-side flag for the ~1 release that this code is here but inactive, I think that's OK.

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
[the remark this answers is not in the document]
eliykat: You're right, I misread the inversion. I think the issue here is that the `is` pattern matching contains an implicit null check, which actually tests something pretty important - that the user is a member of the org at all.

My suggestion will work if you wrap it in a null check for `CurrentContextOrganization`. This is also somewhat special knowledge, but that is constructed from the `UserPrincipal` claims, so it'll only be `not null` if the user is a member of the org. This prevents the null check / inversion from producing unexpected results and is still a bit more explicit than the current logic.

```cs
        var isMember = org is not null;
        if (isMember)
        {
            var organizationAbility = await GetOrganizationAbilityAsync(org);

            var limitCollectionCreationEnabled = _featureService.IsEnabled(FeatureFlagKeys.LimitCollectionCreationDeletionSplit)
                ? organizationAbility is { LimitCollectionCreation: true }
                : organizationAbility is { LimitCollectionCreationDeletion: true };

            // If the limit collection management setting is disabled, allow any user to create collections
            if (!limitCollectionCreationEnabled)
            {
                return true;
            }
        }
```

Although still not ideal I think this is an improvement.

eliykat (commented):
See also https://github.com/bitwarden/server/pull/4730#discussion_r1789485266.

src/Api/AdminConsole/Controllers/OrganizationsController.cs
addisonbeck: I did decide in the moment that it was okay not to feature flag this, and in some ways was maybe even more risky since that involves modifying existing logic a bit, but I did go ahead and add this as suggested.

__Commit:__ [Feature toggle route unblocking](https://github.com/bitwarden/server/pull/4730/commits/2894dab5b3ed7828aaf03d6d09433fc5a15fa1f9)

eliykat (approved):

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
eliykat, line 1: This all looks good to me, my last request is some unit tests that run the new logic - particularly because we had to give it a few revisions. I expect this would just look like a copy + paste of the current create & delete tests, but using the new `orgAbility` properties, and mocking the feature flag to be on.

Then when you remove the feature flag later, you can delete the old tests and leave the new ones.

eliykat (changes requested):

src/Api/Vault/AuthorizationHandlers/Collections/BulkCollectionAuthorizationHandler.cs
addisonbeck: Implemented as suggested

__Commit:__ [Create tests](https://github.com/bitwarden/server/pull/4730/commits/506d3a0fade1d3b908888795a2f3199703647225)

These caught a bug! 

__Commit:__ [Fix bug caught by tests](https://github.com/bitwarden/server/pull/4730/commits/10c4aa724023a581fc8c4d05c7e221f8cdc6cc56)

src/Core/Models/Business/OrganizationLicense.cs
```diff
@@ -150,7 +157,14 @@ public OrganizationLicense(Organization org, SubscriptionInfo subscriptionInfo,
...
     /// <remarks>Intentionally set one version behind to allow self hosted users some time to update before
-    /// getting out of date license errors</remarks>
+    /// getting out of date license errors
+    /// </remarks>
+    /// <remarks>
+    /// ⚠ This is important! We are on this license version because of new
+    /// properties that were added for versions 14 and 15. These properties
+    /// were later removed! When you increment to version 16 please delete
+    /// this comment.
+    /// </remarks>
```
eliykat, line 167: I don't understand the purpose of this comment. Maybe it no longer applies given that we haven't actually removed the properties. (only marked them as deprecated)

eliykat (dismissed):

src/Core/Models/Business/OrganizationLicense.cs
addisonbeck: Yeah, this should be removed. Done!

__Commit:__ [Remove remark about license](https://github.com/bitwarden/server/pull/4730/commits/768a362b15c181adeeefef220a40038772650a60)

Jingo88 (approved):

eliykat (approved):

amorask-bitwarden (approved):

PR #4740, opened 2024-09-05
Use `ILogger.BeginScope` For Better Scopes

justindbaur:
## 📔 Objective

Using `LogContext.PushProperty()` does not translate to additional scopes when using `Microsoft.Extensions.Logging` but when using `Serilog.Extensions.Logging` `ILogger.BeginScope` will translate to `LogContext.Push()` so we should prefer to use the version in MEL for better compatibility. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/SharedWeb/Utilities/RequestLoggingMiddleware.cs
+++ b/src/SharedWeb/Utilities/RequestLoggingMiddleware.cs
@@ -1,0 +1,106 @@
+using System.Collections;
+using Bit.Core.Settings;
+using Bit.Core.Utilities;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+
+#nullable enable
+
+namespace Bit.SharedWeb.Utilities;
+
+public sealed class RequestLoggingMiddleware
+{
+    private readonly RequestDelegate _next;
+    private readonly ILogger<RequestLoggingMiddleware> _logger;
+    private readonly GlobalSettings _globalSettings;
+
+    public RequestLoggingMiddleware(RequestDelegate next, ILogger<RequestLoggingMiddleware> logger, GlobalSettings globalSettings)
+    {
+        _next = next;
+        _logger = logger;
+        _globalSettings = globalSettings;
+    }
+
+    public Task Invoke(HttpContext context)
+    {
+        using (_logger.BeginScope(
+          new RequestLogScope(context.GetIpAddress(_globalSettings),
+            GetHeaderValue(context, "user-agent"),
+            GetHeaderValue(context, "device-type"),
+            GetHeaderValue(context, "device-type"))))
+        {
+            return _next(context);
+        }
+
+        static string? GetHeaderValue(HttpContext httpContext, string header)
+        {
+            if (httpContext.Request.Headers.TryGetValue(header, out var value))
+            {
+                return value;
+            }
+
+            return null;
+        }
+    }
+
+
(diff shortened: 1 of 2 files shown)


src/SharedWeb/Utilities/RequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,106 @@
...
+        using (_logger.BeginScope(
+          new RequestLogScope(context.GetIpAddress(_globalSettings),
+            GetHeaderValue(context, "user-agent"),
+            GetHeaderValue(context, "device-type"),
+            GetHeaderValue(context, "device-type"))))
+        {
+            return _next(context);
+        }
+
+        static string? GetHeaderValue(HttpContext httpContext, string header)
```
withinfocus, line 35: 🌱 As an aside I could see this being a shared utility method. Tons of potential uses: https://github.com/search?q=repo%3Abitwarden%2Fserver%20httpContext.Request.Headers&type=code

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -537,31 +535,7 @@ public static GlobalSettings AddGlobalSettingsServices(this IServiceCollection s
...
-        app.Use(async (httpContext, next) =>
-        {
-            using (LogContext.PushProperty("IPAddress", httpContext.GetIpAddress(globalSettings)))
-            using (LogContext.PushProperty("UserAgent", GetHeaderValue(httpContext, "user-agent")))
-            using (LogContext.PushProperty("DeviceType", GetHeaderValue(httpContext, "device-type")))
-            using (LogContext.PushProperty("Origin", GetHeaderValue(httpContext, "origin")))
-            {
-                httpContext.Response.OnStarting((state) =>
-                {
-                    httpContext.Response.Headers.Append("Server-Version", AssemblyHelpers.GetVersion());
```
withinfocus, line 559: ⚠️ We lost this as well -- is that handled elsewhere?

justindbaur: No it's not, I could not see anything that was consuming it and we have the version on the `/config` endpoint anyway. Happy to add it back though.

src/SharedWeb/Utilities/RequestLoggingMiddleware.cs
justindbaur: I'm not so sure, most of those uses want to do something only when there is a value. They'd likely be better served by using `TryGetValue` themselves, vs calling this helper and then having to do a null check on if we have back a value. 

I'd love for us to enable this rule as a warning: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1854

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
withinfocus: Worth asking all of engineering I think. Clients might be benefitting from it perhaps?

justindbaur: No red alarms [brought up here](https://bitwarden.slack.com/archives/GRN5AAZ6Y/p1725659416925649). Want to put it's removal behind a feature flag so we can quickly turn it back on if anything goes wrong?

withinfocus (approved):

withinfocus:
Conflict ...

withinfocus (approved):

PR #4752, opened 2024-09-10
[PM-3478] Refactor OrganizationUser api

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-3478
## 📔 Objective

Introduce a new OrganizationUsers endpoint which provides a listing of all members' basic details - name, email address, status, role. Restrict access to the current endpoint which returns full details.

The new endpoint and models are called OrganizationUserUser**Mini**Details, following the naming convention of CipherMiniDetails. Also because naming is hard, and "mini" is a relatively clear convention.

The corresponding clients PR refactors most api calls to use this new endpoint: https://github.com/bitwarden/clients/pull/10949.

**Vault team:** I think the only change here for you is that I'm moving some OrganizationUser authorization code out of your ownership. Not sure why it was there to begin with, I assume it was by mistake.

**Notes about implementation**
* authorization checks are implemented using handlers
* the existing `OrganizationUserUserDetails` handler was in the api project and owned by Vault (not sure why) - I have moved it into AC Team ownership in Core
* I never liked putting the organization id in the authorization requirement - it results in a lot of boilerplate. Instead, I've added a strongly-typed wrapper around an organization Id `Guid`, called `OrganizationIdResource`. We can use this as the "resource" for `AuthorizationService`, and keep the "requirement" definitions simple and consistent.
* the query for mini details uses the same repository call as the full details; this could be optimized with its own repository call, but it's unclear to me whether that would make any practical difference for the cost of maintaining the repository queries
* the tests have more boilerplate than I'd like, I want to investigate potential use of the `MemberData` attribute to further streamline them but decided that I didn't have time here

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Authorization/OrganizationScopeResource.cs
r-tome, line 1: Maybe this could be a [record](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record)? I haven't used them much, but it seems like a good fit here since the object is immutable after creation. Records also simplify comparing instances, as two records with the same data (in this case, the same ID) are considered equal.

eliykat: Done!

r-tome (approved):
Great work!

eliykat:
@r-tome I made a couple of changes to simplify this.

I want us to keep authorization checks at the controller level, so I moved that to the controller endpoint. However, that leaves a query class that is basically just a repository call and a bunch of mapping code. So I deleted the query class and the data model, and we just call the repository directly. I think this is OK for simple queries where a separate class only adds overhead (tests, interface definition, data models).

eliykat:
@r-tome I've also fixed a bug in https://github.com/bitwarden/server/pull/4752/commits/ac44bde0bdfc73707bbbaf261d65f1d7282f5115 where an ordinary user could not manage collection access if `LimitCollectionCreationDeletion == true`. Turns out all org users need access to this endpoint regardless of the collection management settings. The handler and its tests have been updated accordingly.

r-tome (approved):
Thanks for providing the context. Changes look good

eliykat:
I've added a feature flag to manage these changes:
* the new endpoint requires the feature flag to be turned on
* the authorization check on the old endpoint now pivots based on the value of the feature flag (less permissive vs. more permissive)

I've also added tests for both feature flag states and done a quick manual test.

Corresponding changes have been made to clients.

r-tome (dismissed):

r-tome (approved):

shane-melton (approved):

PR #4758, opened 2024-09-10
Support overriding organization context for feature flags

withinfocus:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-6014

## 📔 Objective

Allows for server code to pass an overriding organization ID in to force context assessments to just that organization.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
+++ b/src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
@@ -92,28 +92,28 @@ public class LaunchDarklyFeatureService : IFeatureService
         return _client.Initialized && !_client.IsOffline();
     }
 
-    public bool IsEnabled(string key, bool defaultValue = false)
+    public bool IsEnabled(string key, bool defaultValue = false, Guid? overridingOrganizationId = null)
     {
-        return _client.BoolVariation(key, BuildContext(), defaultValue);
+        return _client.BoolVariation(key, BuildContext(overridingOrganizationId), defaultValue);
     }
 
-    public int GetIntVariation(string key, int defaultValue = 0)
+    public int GetIntVariation(string key, int defaultValue = 0, Guid? overridingOrganizationId = null)
     {
-        return _client.IntVariation(key, BuildContext(), defaultValue);
+        return _client.IntVariation(key, BuildContext(overridingOrganizationId), defaultValue);
     }
 
-    public string GetStringVariation(string key, string defaultValue = null)
+    public string GetStringVariation(string key, string defaultValue = null, Guid? overridingOrganizationId = null)
     {
-        return _client.StringVariation(key, BuildContext(), defaultValue);
+        return _client.StringVariation(key, BuildContext(overridingOrganizationId), defaultValue);
     }
 
-    public Dictionary<string, object> GetAll()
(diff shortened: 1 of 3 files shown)


src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
```diff
@@ -190,15 +195,24 @@ void SetCommonContextAttributes(ContextBuilder builder)
 
             case Identity.ClientType.Organization:
                 {
-                    if (_currentContext.OrganizationId.HasValue)
+                    ContextBuilder ldOrg;
```
withinfocus, line 198: 💭 Not sure if this override belongs here. If someone was in an organization context already this doesn't really make sense to use, but would code be passing the override in? I guess that would have no effect right?

src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
```diff
@@ -214,9 +228,21 @@ void SetCommonContextAttributes(ContextBuilder builder)
                         builder.Add(ldServiceAccount.Build());
                     }
 
-                    if (_currentContext.OrganizationId.HasValue)
+                    if (overridingOrganizationId.HasValue || _currentContext.OrganizationId.HasValue)
```
withinfocus, line 231: ⚠️ Line 223 above is something I came upon and worries me because I can't recall why this was done. There isn't really a use case for service account contexts with targeting so far. Regardless, there's no service account ID in context, so should we use the user ID with a kind (that's the LD term for the type) of "service account"? I'd like to straighten this out.

src/Core/Services/Implementations/LaunchDarklyFeatureService.cs
```diff
@@ -214,9 +228,21 @@ void SetCommonContextAttributes(ContextBuilder builder)
...
                     {
-                        var ldOrg = LaunchDarkly.Sdk.Context.Builder(_currentContext.OrganizationId.Value.ToString());
+                        ContextBuilder ldOrg;
+                        if (overridingOrganizationId.HasValue)
```
withinfocus, line 234: 💭 Same question -- should the override be respected here? There's an organization ID tied to the service account.

withinfocus:
@eliykat we talked about doing this a long time ago -- can you share some thoughts on the change?

It would help the server at least, but not sure we can help the client calls via API at the moment since this would have to somehow be passed into the config endpoint I suppose? Seems messy.

eliykat:
> @eliykat we talked about doing this a long time ago -- can you share some thoughts on the change?
> 
> It would help the server at least, but not sure we can help the client calls via API at the moment since this would have to somehow be passed into the config endpoint I suppose? Seems messy.

I agree it would have to be exposed to the client to be maximum useful. It could be included in the `serverConfig` object that is already synced to the client - so the client receives both user-scoped feature flags and organization-scoped feature flags for each organization they're a member of - and exposes that via `ConfigService`.

Alternatively you could request a set of org-specific flags via a separate api call, but that probably undermines the work we did to cache this object client-side.

withinfocus:
I'm inclined to abandon this change given the complications. While there is a potential gap with use today, I think the management of multiple organizations' feature states locally is messier. It would be handy to have this automatically in context but that's also complicated.

PR #4688, opened 2024-08-21
[PM-10560] Create notification database storage

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10560

## 📔 Objective

The purpose of this PR is to add the database and repository layer for the notification center work.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #4769, opened 2024-09-12
PM-10560: Adding Cascades back to Notification Center

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10560

## 📔 Objective

- Fix inability to drop organization or user, due to missing cascades.
In context of https://github.com/bitwarden/server/pull/4688#discussion_r1752672235
- Added missing foreign key in NotificationStatus for Notification with cascades

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
+++ b/util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
@@ -854,7 +854,7 @@ namespace Bit.MySqlMigrations.Migrations
 
                     b.HasIndex("GroupId");
 
-                    b.ToTable("CollectionGroups", (string)null);
+                    b.ToTable("CollectionGroups");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.Models.CollectionUser", b =>
@@ -878,7 +878,7 @@ namespace Bit.MySqlMigrations.Migrations
 
                     b.HasIndex("OrganizationUserId");
 
-                    b.ToTable("CollectionUsers", (string)null);
+                    b.ToTable("CollectionUsers");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.Models.Device", b =>
@@ -1922,7 +1922,7 @@ namespace Bit.MySqlMigrations.Migrations
 
                     b.HasIndex("SecretsId");
 
-                    b.ToTable("ProjectSecret", (string)null);
+                    b.ToTable("ProjectSecret");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.SecretsManager.Models.GroupProjectAccessPolicy", b =>
--- a/util/PostgresMigrations/Migrations/DatabaseContextModelSnapshot.cs
+++ b/util/PostgresMigrations/Migrations/DatabaseContextModelSnapshot.cs
@@ -859,7 +859,7 @@ namespace Bit.PostgresMigrations.Migrations
 
                     b.HasIndex("GroupId");
 
-                    b.ToTable("CollectionGroups", (string)null);
+                    b.ToTable("CollectionGroups");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.Models.CollectionUser", b =>
@@ -883,7 +883,7 @@ namespace Bit.PostgresMigrations.Migrations
 
                     b.HasIndex("OrganizationUserId");
 
-                    b.ToTable("CollectionUsers", (string)null);
+                    b.ToTable("CollectionUsers");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.Models.Device", b =>
@@ -1928,7 +1928,7 @@ namespace Bit.PostgresMigrations.Migrations
 
                     b.HasIndex("SecretsId");
 
-                    b.ToTable("ProjectSecret", (string)null);
+                    b.ToTable("ProjectSecret");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.SecretsManager.Models.GroupProjectAccessPolicy", b =>
--- a/util/SqliteMigrations/Migrations/DatabaseContextModelSnapshot.cs
+++ b/util/SqliteMigrations/Migrations/DatabaseContextModelSnapshot.cs
@@ -843,7 +843,7 @@ namespace Bit.SqliteMigrations.Migrations
 
                     b.HasIndex("GroupId");
 
-                    b.ToTable("CollectionGroups", (string)null);
+                    b.ToTable("CollectionGroups");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.Models.CollectionUser", b =>
@@ -867,7 +867,7 @@ namespace Bit.SqliteMigrations.Migrations
 
                     b.HasIndex("OrganizationUserId");
 
-                    b.ToTable("CollectionUsers", (string)null);
+                    b.ToTable("CollectionUsers");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.Models.Device", b =>
@@ -1911,7 +1911,7 @@ namespace Bit.SqliteMigrations.Migrations
 
                     b.HasIndex("SecretsId");
 
-                    b.ToTable("ProjectSecret", (string)null);
+                    b.ToTable("ProjectSecret");
                 });
 
             modelBuilder.Entity("Bit.Infrastructure.EntityFramework.SecretsManager.Models.GroupProjectAccessPolicy", b =>


src/Sql/NotificationCenter/dbo/Tables/NotificationStatus.sql
```diff
@@ -5,5 +5,6 @@ CREATE TABLE [dbo].[NotificationStatus]
     [ReadDate] DATETIME2 (7) NULL,
     [DeletedDate] DATETIME2 (7) NULL,
     CONSTRAINT [PK_NotificationStatus] PRIMARY KEY CLUSTERED ([NotificationId] ASC, [UserId] ASC),
-    CONSTRAINT [FK_NotificationStatus_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id])
+    CONSTRAINT [FK_NotificationStatus_Notification] FOREIGN KEY ([NotificationId]) REFERENCES [dbo].[Notification] ([Id]) ON DELETE CASCADE,
+    CONSTRAINT [FK_NotificationStatus_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id]) ON DELETE NO ACTION
```
mzieniukbw, line 9: This `ON DELETE CASCADE` creates cycles and is unecessary, since the cascade flow is as follows: `delete user -> cascade delete notification > cascade delete notification status`

rkac-bw: Implement application-level cascading:
Handle the deletion of related records in your application code rather than relying on database-level cascades

withinfocus: @rkac-bw the cascade pattern is all over the place for account deletion though, so the intention is to just keep going with that for now.

rkac-bw: Well it may be better to pull it back because SQL Server is preventing you from adding another cascade delete from User to NotificationStatus due to multiple cascade paths

rkac-bw: These are multipath deletes to same record:
User -> Notification -> NotificationStatus
User -> NotificationStatus


Deleting a User could delete NotificationStatus directly.
It could also delete Notification, which would then cascade to delete NotificationStatus.

Not allowed in SQ Server< Notification had ON DELETE CASCADE for UserId
NotificationStatus had ON DELETE CASCADE for NotificationId
You wanted to add ON DELETE CASCADE for UserId on NotificationStatus

Options:
- Switch to handling cascade deletes in code (my preference)
- Use triggers on user for on delete
- Use ON DELETE SET NULL and clean up process later (we should have a clean up process to clean up old and stale records regardless)
- Redesign the database schema to avoid multiple paths.

rkac-bw: Still not sure why we would delete the notification if there is no user to request it, and in the denormalized view would just be flagged as global but have the userid for receiving, Creating a  record for every user when it is global is an interesting issue in both scenarios.

withinfocus: Alright, after some background chats and research I feel conflicted about what our strategy really is with cascades. We do need to deal with deletions, but I see us using them in conjunction with logic in the https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Stored%20Procedures/User_DeleteById.sql proc. At the end of the day here and given how these optional FKs operate, let's just invest in the proc.

rkac-bw: Chatting about this and it looks like it would be best to have deletes caused by user to be run from here:

https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Stored%20Procedures/User_DeleteById.sql

mzieniukbw:
I have changed the Primary key columns order to be consistent with what we have in SQL Server, by changing EF src/Infrastructure.EntityFramework/NotificationCenter/Configurations/NotificationStatusEntityTypeConfiguration.cs
But now MySql migration does not work correctly, due to:
```
Applying migration '20240909133252_NotificationCenter'.
Applying migration '20240916193337_UpdateNotificationCenterOnDeleteCascade'.
Failed executing DbCommand (3ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
CALL POMELO_BEFORE_DROP_PRIMARY_KEY(NULL, 'NotificationStatus');
ALTER TABLE `NotificationStatus` DROP PRIMARY KEY;
MySqlConnector.MySqlException (0x80004005): Cannot drop index 'PRIMARY': needed in a foreign key constraint
```
Should i revert it ? But then we have inconsistencies between Sql Server and EF ?
Or maybe we need some manual migration for EF MySQL ?
Or maybe i change the order for Sql server ?
Any idea ?

src/Sql/NotificationCenter/dbo/Tables/NotificationStatus.sql
mzieniukbw: Fixed in f1844e3

withinfocus:
There shouldn't be cascades at all now, since your procs for MSSQL and EF logic elsewhere handles it.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -199,6 +199,14 @@ await dbContext.ApiKeys.Where(ak => ak.ServiceAccount.OrganizationId == organiza
             await dbContext.ServiceAccount.Where(sa => sa.OrganizationId == organization.Id)
                 .ExecuteDeleteAsync();
 
+            var notificationStatuses = from ns in dbContext.NotificationStatuses
+                                       join n in dbContext.Notifications on ns.NotificationId equals n.Id
+                                       where n.OrganizationId == organization.Id
+                                       select ns;
+            dbContext.NotificationStatuses.RemoveRange(notificationStatuses);
```
Thomas-Avery, line 206:
```suggestion
        await dbContext.NotificationStatuses.Where(ns => ns.Notification.OrganizationId == organization.Id).ExecuteDeleteAsync();
```

Can we use `ExecuteDeleteAsync` here?

`.RemoveRange()` Will query for all notification status that match the filter, load it into application memory, then submit a delete statement for each.

More information here https://learn.microsoft.com/en-us/ef/core/saving/execute-insert-update-delete#executedelete

src/Infrastructure.EntityFramework/NotificationCenter/Configurations/NotificationStatusEntityTypeConfiguration.cs
```diff
@@ -10,9 +10,13 @@ public class NotificationStatusEntityTypeConfiguration : IEntityTypeConfiguratio
     public void Configure(EntityTypeBuilder<NotificationStatus> builder)
     {
         builder
-            .HasKey(ns => new { ns.UserId, ns.NotificationId })
+            .HasKey(ns => new { ns.NotificationId, ns.UserId })
```
Thomas-Avery, line 13: I know ordering can have performance implications with some database providers. 

In the main query, we join on both properties but have `UserId` in the Where clause.

We should probably test which way is more performant or if it affects performance at all before making this change in light of MySQL migrations breaking.

It might make sense to have it be its own separate ticket and scoping it out of this one.

Thomas-Avery (changes requested):

src/Infrastructure.EntityFramework/NotificationCenter/Configurations/NotificationStatusEntityTypeConfiguration.cs
mzieniukbw: I will revert this for now. My worry is that in EF we have `UserId, NotificationId` as columns order, while SQL Server have it as `NotificationId, UserId`, so it's a bit inconsistent. Also a secondary (non-primary) index is needed to match on the second column (i.e. delete by UserId only or join on NotificationId only for deletes).

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
mzieniukbw: Nice, i will make that change.
It's still blurry to me where and how EF code is translated to SQ and which parts will it optimize. I will remember this one in the future.

src/Infrastructure.EntityFramework/NotificationCenter/Configurations/NotificationStatusEntityTypeConfiguration.cs
mzieniukbw: Fixed in 1c48155

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
mzieniukbw: Fixed in 1c48155

util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
```diff
@@ -854,7 +854,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.HasIndex("GroupId");
 
-                    b.ToTable("CollectionGroups");
+                    b.ToTable("CollectionGroups", (string)null);
```
Thomas-Avery, line 857: ❓ I don't get these database context model snapshot changes when running migrations locally?

I don't think we need any EF migrations for this now.

Thomas-Avery (changes requested):

util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
mzieniukbw: I will have to edit this file manually, since i removed the migration with `dotnet ef migrations remove --force`, instead of reverting the changes in commits and this is what i was left with 🤷

mzieniukbw: Fixed in 70b31f2

mzieniukbw: Fixed in 70b31f2

Thomas-Avery (approved):
LGTM

r-tome (approved):

rkac-bw (approved):
lgtm

PR #4770, opened 2024-09-12
PM 12001 - Fix Empty User Search 500

jrmccannon:
## 🎟️ Tracking

JIRA: https://bitwarden.atlassian.net/browse/PM-12001

## 📔 Objective

When searching for a user that does not exist in the given organization, the serialization fails for sending an empty `IEnumerable<ValueTuple<T1, T2>>`. This is a quick fix to not set the temp data if the enumerable is empty. Also added some null checking to the razor template.

I suspect the DefaultTempDataSerializer has an issue because if an `IEnumerable<ValueTuple<T1, T2>>` with a value in it is returned, it will serialize the data no problem. It only has an issue with an empty enumerable. I did try converting the value over to a `Dictionary<Guid, bool>` and it failed when evaluating during the render. There seems to be a weird interaction between there.

Ideally, we would create a new model to represent the `User` instead of the database model and populate the value on that.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Admin/Controllers/UsersController.cs
+++ b/src/Admin/Controllers/UsersController.cs
@@ -68,7 +68,10 @@ public class UsersController : Controller
         {
             var user2Fa = (await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(users.Select(u => u.Id))).ToList();
             // TempDataSerializer is having an issue serializing an empty IEnumerable<Tuple<T1,T2>>, do not set if empty.
-            if (user2Fa.Count != 0) TempData["UsersTwoFactorIsEnabled"] = user2Fa;
+            if (user2Fa.Count != 0)
+            {
+                TempData["UsersTwoFactorIsEnabled"] = user2Fa;
+            }
         }
 
         return View(new UsersModel
--- a/src/Admin/Controllers/UsersController.cs
+++ b/src/Admin/Controllers/UsersController.cs
@@ -66,7 +66,9 @@ public class UsersController : Controller
 
         if (_featureService.IsEnabled(FeatureFlagKeys.MembersTwoFAQueryOptimization))
         {
-            TempData["UsersTwoFactorIsEnabled"] = await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(users.Select(u => u.Id));
+            var user2Fa = (await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(users.Select(u => u.Id))).ToList();
+            // TempDataSerializer is having an issue serializing an empty IEnumerable<Tuple<T1,T2>> so sending null if empty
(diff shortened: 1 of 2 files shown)


addisonbeck (dismissed):

src/Admin/Controllers/UsersController.cs
```diff
@@ -66,7 +66,9 @@ public async Task<IActionResult> Index(string email, int page = 1, int count = 2
 
         if (_featureService.IsEnabled(FeatureFlagKeys.MembersTwoFAQueryOptimization))
         {
-            TempData["UsersTwoFactorIsEnabled"] = await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(users.Select(u => u.Id));
+            var user2Fa = (await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(users.Select(u => u.Id))).ToList();
+            // TempDataSerializer is having an issue serializing an empty IEnumerable<Tuple<T1,T2>>, do not set if empty.
+            if (user2Fa.Count != 0) TempData["UsersTwoFactorIsEnabled"] = user2Fa;
```
eliykat, line 71: Code style: please don't use single-line `if` statements.

src/Admin/Views/Users/Index.cshtml
```diff
@@ -72,8 +72,8 @@
                             }
                             @if (featureService.IsEnabled(Bit.Core.FeatureFlagKeys.MembersTwoFAQueryOptimization))
                             {
-                                var usersTwoFactorIsEnabled = TempData["UsersTwoFactorIsEnabled"] as IEnumerable<(Guid userId, bool twoFactorIsEnabled)>;
-                                @if(usersTwoFactorIsEnabled.FirstOrDefault(tuple => tuple.userId == user.Id).twoFactorIsEnabled)
+                                @if(TempData["UsersTwoFactorIsEnabled"] is IEnumerable<(Guid userId, bool twoFactorIsEnabled)> usersTwoFactorIsEnabled
+                                    && usersTwoFactorIsEnabled.FirstOrDefault(tuple => tuple.userId == user.Id).twoFactorIsEnabled)
```
eliykat, line 76: If we're using `FirstOrDefault` then we should handle the `null` case (I assume default to `false`). Here we _should_ have a 2FA result for each user in the result set, so I think just `First` would be fine. That said, I haven't tested that locally, and I would be OK if you wanted to take the more cautious approach.

eliykat (dismissed):
Some very minor feedback, but overall looks good to me!

src/Admin/Views/Users/Index.cshtml
jrmccannon: Added a couple checks. The Roslyn analyzer says that the default value of `ValueTuple<Guid, bool>` isn't null, so I broke up the `if` statement into separate variables and used pattern for null checking.

addisonbeck (approved):

PR #4796, opened 2024-09-23
[PM-12526] Can Reduce Org's PM seats to be lower than SM seats

jonashendrickx:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12526

## 📔 Objective

Orgs cannot have more SM seats than PM seats.

When creating an org, or when increasing the number of SM seats, the check for that is properly performed. However, this check is not performed when decreasing the number of PM seats on an org. Thereby letting you have an org whose PM seats < SM seats.

Steps to Reproduce:

1. Create an org that has an SM subscription
2. Set their PM and SM seats to be the same
3. Go to the Org’s subscription page
4. Decrease the number of PM seats and save

- Expected: It is not saved properly and an error toast is displayed to the user that you cannot have more SM seats than PM seats in an org.
- Actual: It is saved properly

## 📸 Screenshots

Validated scenarios:

- Downgrading seats with new PM seat count < SM seat count
- Upgrading seats with new PM seat count == SM seat count
- Upgrading seats with new PM seat count > SM seat count
- Downgrading seats with new PM seat count < SM seat count after SM has been deactivated or unsubscribed.
- Downgrading PM seats

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


cturnbull-bitwarden (approved):
Looks good! Left one nitpick.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -241,6 +241,14 @@ public async Task UpdateSubscription(Guid organizationId, int seatAdjustment, in
             throw new NotFoundException();
         }
 
+        if (organization.UseSecretsManager)
+        {
+            if (organization.Seats + seatAdjustment < organization.SmSeats)
```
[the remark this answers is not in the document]
jonashendrickx: @cturnbull-bitwarden  done

eliykat (changes requested):
I suggest that the `AdjustSeatsAsync` method (called by this method) is the better place for this logic, as it contains all the other validation when adjusting seats for PM.

`AdjustSeatsAsync` is called by a few other methods as well, which is good because then we're doing this validation in all flows. The only potential issue I can see is that when [autoscaling is reverted](https://github.com/bitwarden/server/blob/576be63d2a8373dd3edbea2489512d556b2d1769/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1178-L1178), we revert PM seats and then SM seats - we'd need to reverse the order to ensure that we don't accidentally trip this rule. (Please double check me to make sure this otherwise works.)

amorask-bitwarden (dismissed):

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1176,20 +1181,41 @@ await _referenceEventService.RaiseEventAsync(
...
+            var smSeatAdjustment = initialSmSeatCount.HasValue && currentOrganization.SmSeats.HasValue
+                ? initialSmSeatCount.Value - currentOrganization.SmSeats.Value
+                : 0;
 
-            // Revert SmSeat autoscaling
-            if (initialSmSeatCount.HasValue && currentOrganization.SmSeats.HasValue &&
-                currentOrganization.SmSeats.Value != initialSmSeatCount.Value)
+            // There is a possibility that the initial state may have a lower PM seat count than SM seat count.
+            // In this case we need to revert in the opposite order
+            if (seatAdjustment != 0 && smSeatAdjustment != 0 && initialSeatCount.Value > currentOrganization.SmSeats.Value)
```
eliykat, line 1194: I don't think there's any need for this conditional logic, unless you found an issue in your testing.

I just meant swap the current code blocks, i.e.

```cs
            // Revert SmSeat autoscaling
            // Do this first so that SmSeats never exceed PM seats (due to current billing requirements)
            if (initialSmSeatCount.HasValue && currentOrganization.SmSeats.HasValue &&
                currentOrganization.SmSeats.Value != initialSmSeatCount.Value)
            {
                var smSubscriptionUpdateRevert = new SecretsManagerSubscriptionUpdate(currentOrganization, false)
                {
                    SmSeats = initialSmSeatCount.Value
                };
                await _updateSecretsManagerSubscriptionCommand.UpdateSubscriptionAsync(smSubscriptionUpdateRevert);
            }

            // Revert PM autoscaling
            if (initialSeatCount.HasValue && currentOrganization.Seats.HasValue && currentOrganization.Seats.Value != initialSeatCount.Value)
            {
                await AdjustSeatsAsync(organization, initialSeatCount.Value - currentOrganization.Seats.Value);
            }
```

eliykat (changes requested):

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -358,6 +358,11 @@ private async Task<string> AdjustSeatsAsync(Organization organization, int seatA
             }
         }
 
+        if (organization.UseSecretsManager && organization.Seats + seatAdjustment < organization.SmSeats)
+        {
+            throw new BadRequestException("You cannot have more Password Manager seats than Secrets Manager seats.");
```
eliykat, line 363: Just spotted this. I think this logic is correct, but the error message is around the wrong way. You cannot have more SM seats than PM seats. (Every SM user must also be a PM user.)

cturnbull-bitwarden (approved):

eliykat (approved):
LGTM, but please send me a calendar invite if you still want to sync on this!

PR #4797, opened 2024-09-23
[PM-12074] - Refactored `Index` to use `UserModel`

jrmccannon:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-12074

## 📔 Objective
Instead of using the entity model for the User which is directly tied to the database, I switched out the model for a lightweight view model that can be used to hold only what that page needs.  This will allow it to be expanded for properties that aren't directly on the database model.  This also removes the feature flag code in the controller and view.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Admin/Controllers/UsersController.cs
```diff
@@ -63,20 +56,11 @@ public async Task<IActionResult> Index(string email, int page = 1, int count = 2
 
         var skip = (page - 1) * count;
         var users = await _userRepository.SearchAsync(email, skip, count);
-
-        if (_featureService.IsEnabled(FeatureFlagKeys.MembersTwoFAQueryOptimization))
```
r-tome, line 67: The feature flag lookup, should be kept here. We only want to execute that query if its enabled

src/Admin/Models/UserModel.cs
r-tome, line 1: Instead of creating a new model can we refactor `UserViewModel`?

test/Admin.Test/Models/UserModelTests.cs
r-tome, line 1: Use `[SutProviderCustomize]` to automatically create the fixture and `[BitAutoData]` to create test data

r-tome (changes requested):

src/Admin/Models/UserModel.cs
jrmccannon: Yep! Also had to tweak edit model and all the `.cshtml` files to reflect this.

r-tome (approved):
Looks good to me! Thanks for making the changes I had requested.

Ensure that QA tests the editing of all User fields that should be modifiable

PR #4804, opened 2024-09-25
[PM-12630] support for ping identity SCIM provisioning

kspearrin:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-12630

## 📔 Objective

SCIM for Ping Identity kind of worked already out of the box, but we needed to add support for identifying Ping as a SCIM provider and allowing it to use Group PUT calls to update group/user membership.

### DIFF
--- a/bitwarden_license/src/Scim/Context/ScimContext.cs
+++ b/bitwarden_license/src/Scim/Context/ScimContext.cs
@@ -13,7 +13,7 @@ public class ScimContext : IScimContext
     private bool _builtHttpContext;
 
     // See IP list from Ping in docs: https://support.pingidentity.com/s/article/PingOne-IP-Addresses
-    private readonly HashSet<string> _pingIpAddresses =
+    private static readonly HashSet<string> _pingIpAddresses =
     [
         "18.217.152.87",
         "52.14.10.143",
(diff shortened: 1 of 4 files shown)


bitwarden_license/src/Scim/Context/ScimContext.cs
```diff
@@ -4,13 +4,40 @@
...
+        "18.184.65.93",
+        "52.57.244.92",
+        "18.195.7.252",
+        "108.128.67.71",
+        "34.246.158.102",
+        "108.128.250.27",
+        "52.63.103.92",
+        "13.54.131.18",
+        "52.62.204.36"
+    ];
```
kspearrin, line 39: I am not sure the best place for this static list. Open to suggestions if this isn't right.

bitwarden_license/src/Scim/Context/ScimContext.cs
```diff
@@ -4,13 +4,40 @@
...
```
kspearrin, line 39: I am not sure the best place for this static list. Open to suggestions if this isn't right.

jrmccannon: I'd say maybe move it to a separate file. Maybe a new folder for `Providers/` and then have this be in `PingIdentityConstants`.  But maybe something to consider after we have to store things for other providers.

jrmccannon (approved):

Issue #4034, opened 2024-04-30
SCIM /Users listing does not return any users unless the `startIndex` and `count` is specified


dougedey-sf:
### Steps To Reproduce

1. Enable SCIM Provisioning on an account
2. Request the users listing using the provided SCIM Url and add `/Users` to the end (i.e.: `https://scim.bitwarden.com/v2/<Tenant ID>/Users`

### Expected Result

You should get a list of users, unfiltered.

### Actual Result

No data is returned

### Additional Context

I believe this is a bug, but it may be intentional, based on the third check on line 58 in [GetUsersListQuery](https://github.com/bitwarden/server/blob/ccaee0b719158ff0d983ffe11f9ae827e2b353a9/bitwarden_license/src/Scim/Users/GetUsersListQuery.cs#L37-L67), it requires the `startIndex` and `count` to be set, these are defined as [optional parameters](https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2.4) in the SCIM Paginaiton RFC, and should default to `startIndex` of 1 and the server defined `count`

[Query Resources](https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2) does not state that `startIndex` and `count` are required query variables, and the example given:

```
GET /Users?attributes=userName
Host: example.com
Accept: application/scim+json
Authorization: Bearer h480djs93hd8
```

Also does not work with the BitWarden API.

I believe the fix here is to add an:

```c#
else
{
  userList = orgUsers.OrderBy(ou => ou.Email)
    .Take(10) # I am unsure what the real default should be here
    .ToList();
  totalResults = orgUsers.Count;
}
```

### Build Version

2024.4.1

### Environment

Cloud (bitwarden.com)

### Issue Tracking Info

- [X] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

cbbit:
Hi there,

Thank you for your report!

I have flagged this to our engineering team.

If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time - our engineering team will be happy to review these.

Thanks once again!

PR #4812, opened 2024-09-27
[PM-12777] Fixed Issue #4034, API endpoint now handles optional parameters

BensonB12:
## 🎟️ Tracking

https://github.com/bitwarden/server/issues/4034

## 📔 Objective

The API endpoint: 'https://scim.bitwarden.com/v2/{Tenant-ID}/Users' now accepts null values for 'count' and 'startingIndex'. It now has a default value of count 50 and startingIndex 1. We made an integration test (based on a previously made one) that now tests that function. 

User: https://github.com/dougedey-sf said that this new function is the wanted behavior. We did not double check that, but we did notice that https://github.com/cbbit commented without contradicting the issue.

## 🦮 Reviewer guidelines


bitwarden_license/src/Scim/Users/GetUsersListQuery.cs
```diff
@@ -18,6 +18,9 @@ public GetUsersListQuery(IOrganizationUserRepository organizationUserRepository)
         string emailFilter = null;
         string usernameFilter = null;
         string externalIdFilter = null;
+        count ??= 50;
+        startIndex ??= 1;
```
jrmccannon, line 22: Could you please move these defaults up to a request model in the controller?

jrmccannon (changes requested):
First off, thank you for contributing to Bitwarden. This is a good initial pass and thank you for including a test. To improve on this, you could maybe add a new request object that uses some data annotations that limits the values that could be passed in to that endpoint.  You can then provide that object as a parameter in `UserController#Get` with the `[FromQuery]` attribute.

That class could look something like
```
public class GetUsersQueryParam
{
        public string Filter { get; init; } = string.Empty;

        [Range(1, int.MaxValue)]
        public int Count { get; init; } = 50;

        [Range(1, int.MaxValue)]
        public int StartIndex { get; init; } = 1;
}
```

This would provide callers with helpful error messages in the event they accidentally provide invalid values. Then you could refactor the method signature and syntax of `GetUsersListQuery#GetUserListAsync()`  to take`int`s rather than `int?`s.

Thanks again!

jrmccannon (approved):

jrmccannon:
Sorry for the delay.  This has passed through QA and will be included in the 2024.10.1 release.  Thank you again for your contribution.

PR #4848, opened 2024-10-03
PM-11123: Notification Status Details view

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-11123

## 📔 Objective

Get Notifications with Statuses as one database query

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Sql/NotificationCenter/dbo/Views/NotificationStatusDetailsView.sql
```diff
@@ -0,0 +1,13 @@
...
+    N.*,
+    NS.UserId AS NotificationStatusUserId,
+    NS.ReadDate,
+    NS.DeletedDate
+FROM
+    [dbo].[Notification] AS N
+LEFT JOIN
+    [dbo].[NotificationStatus] as NS
+ON
+    N.[Id] = NS.[NotificationId]
```
withinfocus, line 13: ⛏️ Newlines at the end.

src/Core/NotificationCenter/Entities/NotificationStatusDetails.cs
```diff
@@ -0,0 +1,26 @@
...
+    public ClientType ClientType { get; set; }
+    public Guid? UserId { get; set; }
+    public Guid? OrganizationId { get; set; }
+    [MaxLength(256)]
+    public string? Title { get; set; }
+    public string? Body { get; set; }
+    public DateTime CreationDate { get; set; }
+    public DateTime RevisionDate { get; set; }
+    // Notification Status fields
+    public Guid? NotificationStatusUserId { get; set; }
```
withinfocus, line 23: ❓ This is here just to aid in querying right? Can we restructure the join usage to avoid this being defined in the view and response objects? I find it confusing to be there when it would be set as the above `UserId`.

Thomas-Avery: What about the case of a notification that is targeted at an organization? 

```
Id : 61c208d3-74ae-4d5f-9c92-150ab941d262
ClientType: All
UserId: NULL
OrganizationId: 44674c0f-a6af-4c09-9c65-03ab89094764
NotificationStatusUserId: <target-userId>
ReadDate: 2025-09-18 19:00:52.0000000
DeletedDate: 2025-09-18 19:00:52.0000000
```

withinfocus: I thought about that, but ... does the caller care? A user is calling this. We could arguably drop both user ID properties.

src/Core/NotificationCenter/Entities/NotificationStatusDetails.cs
Thomas-Avery, line 1: The pattern is to put entities that match DB tables in `/Core/<team>/Entities`. 

This is a data model for a view which typically would go in `/Core/NotificationCenter/Models/Data`

Thomas-Avery (changes requested):

src/Core/NotificationCenter/Entities/NotificationStatusDetails.cs
Thomas-Avery: True, in these cases it's implied that the notification is for the given user.

mzieniukbw: I can drop it from the `NotificationStatusDetails.cs`. Like you Both said, it does not matter.
But it does from `NotificationStatusDetailsView` and `Notification_ReadByUserIdAndStatus`.
We need to filter on `NotificationStatus.UserId`, so we only return current user statuses. This needs to be done from stored procedure, so the view have to expose it.
We also need `Notification.UserId`, so we only return notifications that match current user (assuming it's not null).
Both are nullable. I just can't see how we can do this if we only have one or another (or combined with ISNULL, also does not work).
Alternatively I can get rid of the `NotificationStatusDetailsView`, that solves this problem.

withinfocus: The easiest path forward might be to keep these in the model and view then, but I don't want them to be seen in the API response model at least.

mzieniukbw: I think it's fine to delete the `NotificationStatusUserId` field from `NotificationStatusDetails.cs` - so the only occurrence is in SQL files.

src/Sql/NotificationCenter/dbo/Views/NotificationStatusDetailsView.sql
mzieniukbw: Fixed in 74b1fcdd

src/Core/NotificationCenter/Entities/NotificationStatusDetails.cs
mzieniukbw: Fixed in 5c936d09

src/Core/NotificationCenter/Entities/NotificationStatusDetails.cs
mzieniukbw: Fixed in 88e1edd5

withinfocus (commented):
I am jumping on a plane so I'll let @Thomas-Avery get final review in.

withinfocus (approved):
(but my part seems good, and we'll see the real changes next)

Thomas-Avery (approved):
LGTM

PR #4853, opened 2024-10-04
[PM-11404] Account Management: Prevent a verified user from purging their vault

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-11404

## 📔 Objective

If the Account Deprovisioning feature flag is enabled then prevent organization managed users from purging their vault.

Clients PR: https://github.com/bitwarden/clients/pull/11411

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


JaredSnider-Bitwarden (dismissed):
Auth changes LGTM!

cturnbull-bitwarden (approved):

src/Core/Services/Implementations/UserService.cs
```diff
@@ -31,6 +31,8 @@
 using File = System.IO.File;
 using JsonSerializer = System.Text.Json.JsonSerializer;
 
+#nullable enable
```
eliykat, line 34: Please don't enable nullable unless you've reviewed the code to make sure its null handling is correct. In this case, you've enabled it on this entire class, but if you look at it in the Github diff, there are null warnings everywhere because the existing code doesn't handle nullable types properly.

You should enable it just for your new methods by putting `#nullable enable` immediately before the block of new code, and then `#nullable disable` immediately after. Then double check the Github diff to find and fix any null warnings.

Same for any other files here.

src/Api/AdminConsole/Models/Response/ProfileOrganizationResponseModel.cs
```diff
@@ -122,4 +128,13 @@ public ProfileOrganizationResponseModel(OrganizationUserOrganizationDetails orga
...
     public bool AllowAdminAccessToAllCollectionItems { get; set; }
+    /// <summary>
+    /// Indicates if the organization manages the active user.
+    /// </summary>
+    /// <remarks>
+    /// An organization manages a user if the user's email domain is verified by the organization and the user is a member of it.
+    /// The organization must be enabled and able to have verified domains.
+    /// This property is null if the Account Deprovisioning feature flag is disabled.
+    /// </remarks>
+    public bool? ManagesActiveUser { get; set; }
```
eliykat, line 139: I think "active" is confusing here. This class doesn't have any concept of whether the user it relates to is "active" in the client-side state or not. 

This class (and `Organization.cs` on the client-side) already relates to a specific user - it has a `UserId` and an `OrganizationUserId`. So there is no need to qualify who the user is - it's not necessarily the "active" user, it's the user that matches the `UserId` in this class.

I suggest taking the "active" wording out of this. Maybe `UserIsManagedByOrganization`?

eliykat (changes requested):
I have a couple of comments based on our recent discussions, but I'll still defer to @jrmccannon for the full review.

src/Api/AdminConsole/Models/Response/ProfileOrganizationResponseModel.cs
r-tome: That is fair, I'll rename it to `UserIsManagedByOrganization`, thanks for suggesting it.
I named it "active" because when we return this object its always in relation to the active user. You're right that it doesn't have to be though.

eliykat (commented):
My feedback has been resolved, thanks!

eliykat (approved):

jrmccannon (approved):
Overall, I don't see any problems with this and it is consistent with the rest of the code base. I did have one :pick:, but its a non-blocking change.

jrmccannon (dismissed):

src/Core/Services/Implementations/UserService.cs
```diff
@@ -1263,21 +1263,29 @@ public async Task<bool> IsLegacyUser(string userId)
...
 
-    public async Task<bool> IsManagedByAnyOrganizationAsync(Guid userId)
+#nullable enable
+    public async Task<bool?> IsManagedByAnyOrganizationAsync(Guid userId)
     {
-        var managingOrganization = await GetOrganizationManagingUserAsync(userId);
-        return managingOrganization != null;
+        var managingOrganizations = await GetOrganizationsManagingUserAsync(userId);
+        return managingOrganizations?.Any();
     }
```
eliykat, line 1271: Apologies for the late suggestion, but this should return a `bool`, not a `bool?`. The user is either managed or they're not.

src/Core/Services/Implementations/UserService.cs
```diff
@@ -1263,21 +1263,29 @@ public async Task<bool> IsLegacyUser(string userId)
...
 
-    public async Task<Organization> GetOrganizationManagingUserAsync(Guid userId)
+    public async Task<IEnumerable<Organization>?> GetOrganizationsManagingUserAsync(Guid userId)
```
eliykat, line 1273: I suggest that this is also made non-nullable. You can just return an empty list if the feature flag is off.

r-tome: I changed it to return an empty list, cleanup will be easier later on.

cturnbull-bitwarden (dismissed):

eliykat (approved):

JaredSnider-Bitwarden (approved):

gbubemismith (dismissed):
Vault changes look good

cturnbull-bitwarden (approved):

JaredSnider-Bitwarden (approved):

gbubemismith (approved):

jrmccannon (approved):

PR #4873, opened 2024-10-09
[PM-11249] Update cipher revision date when an attachment is added or deleted

nick-livefront:
## 🎟️ Tracking

[PM-11249](https://bitwarden.atlassian.net/browse/PM-11249)

## 📔 Objective

**Current Issue:** When an attachment is added or deleted, web/browser clients are not being updated of the change.

**Cause:** The logic within the [CoreSyncService](https://github.com/bitwarden/clients/blob/e2275ad0bc138bf2fd0c2577206f488a756839b5/libs/common/src/platform/sync/core-sync.service.ts#L126) on the client side uses the cipher revision date to determine if a sync is needed. Server side the cipher revision date isn't updated.

**Proposed Fix:** 
- Update the revision date on the cipher if an attachment is added or deleted and store in the DB
- In order to support this, the delete attachment endpoint needs to return the cipher as clients will need to know of the new revision date on the cipher.
- [client PR](https://github.com/bitwarden/clients/pull/11758) for the parallel changes to support this.

## 📸 Screenshots

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/589514eb-f074-49f5-8097-e64450f5e24a" />|<video src="https://github.com/user-attachments/assets/861733f2-3b2f-405e-a9b3-23c667eea4b6" />|

## 🦮 Reviewer guidelines

[PM-11249]: https://bitwarden.atlassian.net/browse/PM-11249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shane-melton:
❓ Are you still able to modify the cipher after uploading an attachment? I'm worried that we don't return the updated cipher in the attachment endpoint so the `cipher.revisionDate` will be stale client side, which will prevent users from making changes until they manually re-sync or refresh their vault.

nick-livefront:
> ❓ Are you still able to modify the cipher after uploading an attachment? I'm worried that we don't return the updated cipher in the attachment endpoint so the `cipher.revisionDate` will be stale client side, which will prevent users from making changes until they manually re-sync or refresh their vault (on the client that uploaded the attachment).

Nice catch, I only tested on the client that received the update which in turn would have the updated cipher.

On the _same_ client you are not able to. 


<video src="https://github.com/user-attachments/assets/e395f837-6cbf-43a0-8442-94eef404d196" />

nick-livefront:
@shane-melton I order to support this change for deleting an attachment, the UI needed to know the updated revision timestamp. I added logic to return the updated cipher to both the [client](https://github.com/bitwarden/server/pull/4873) and here [13acab5](https://github.com/bitwarden/server/pull/4873/commits/13acab5b8c329703234c3ca23b79dff20275df1e)

gbubemismith (approved):
Changes look good

nick-livefront:
Moving to a draft as this work is being tabled for a couple weeks why we wait for the extension refresh to roll out.

Issue #2013, opened 2022-05-24
Potential errors in the project


NikiPn:
### Steps To Reproduce

Hello!

I’d like to show you several issues found in your project.

### Potential error in [BillingInfo.cs](https://github.com/bitwarden/server/blob/master/src/Core/Models/Business/BillingInfo.cs)
The property is assigned values twice.

Below is a code fragment with this issue:
```
    public BillingInvoice(Invoice inv)
    {
         Amount = inv.AmountDue / 100M;    // <=
         Date = inv.Created;
         Url = inv.HostedInvoiceUrl;
         PdfUrl = inv.InvoicePdf;
         Number = inv.Number;
         Paid = inv.Paid;
         Amount = inv.Total / 100M;        // <=
    }
```

The result of the inv.AmountDue / 100 expression is [assigned](https://github.com/bitwarden/server/blob/master/src/Core/Models/Business/BillingInfo.cs#L142) to Amout. However, there’s a similar operation five lines below but this time inv.Total / 100M is [assigned](https://github.com/bitwarden/server/blob/master/src/Core/Models/Business/BillingInfo.cs#L148).

### Potential error in [AppleIapService.cs](https://github.com/bitwarden/server/blob/master/src/Core/Services/Implementations/AppleIapService.cs).
Perhaps the '??' operator doesn’t work as expected.

Here’s this code fragment:
```
    throw new Exception("Failed verifying Apple IAP after too many attempts. Last attempt status: " +
    lastReceiptStatus?.Status ?? "null");
```

The '??' operator’s priority is lower than the '+' operator’s priority. Therefore, the value of the Status property is added to the string first, and after that the null-coalescing operator snaps into action.

[Link](https://github.com/bitwarden/server/blob/master/src/Core/Services/Implementations/AppleIapService.cs#L95) to the line with error.

### Potential error in [FreshdeskController.cs](https://github.com/bitwarden/server/blob/master/src/Billing/Controllers/FreshdeskController.cs).
Erroneous postfix notation.

The fragment with a potential error:
```
    private async Task<HttpResponseMessage> CallFreshdeskApiAsync(...., int retriedCount = 0)
    {
         ....
         return await CallFreshdeskApiAsync(request, retriedCount++);             // <=
    }
```

During the postfix incrementation the current value of the variable is returned first and only after that this value is increased.

[Link](https://github.com/bitwarden/server/blob/master/src/Billing/Controllers/FreshdeskController.cs#L167) to the line with error.

[...]

PR #4896, opened 2024-10-15
[PM-13693] Fixed null warnings

MJebran:
## 🎟️ Tracking
github issue link: 
https://github.com/bitwarden/server/issues/2013

link to 14 specified issues mentioned in the above link: 
https://pvs-studio.com/en/blog/posts/csharp/0947/ 

## 📔 Objective

We fixed 4 issues that could have cased null error, issues were warnings themselves

Changed a method that was not Undernet test, looking for conformation to add tests.

Collaborators (mob programmed)
https://github.com/luris26 
https://github.com/BensonB12

## 🦮 Reviewer guidelines


src/Infrastructure.EntityFramework/AdminConsole/Repositories/Queries/OrganizationUserUserViewQuery.cs
```diff
@@ -28,9 +28,9 @@ from su in su_g.DefaultIfEmpty()
             SsoExternalId = x.su.ExternalId,
             Permissions = x.ou.Permissions,
             ResetPasswordKey = x.ou.ResetPasswordKey,
-            UsesKeyConnector = x.u != null && x.u.UsesKeyConnector,
+            UsesKeyConnector = x.u.UsesKeyConnector,
             AccessSecretsManager = x.ou.AccessSecretsManager,
-            HasMasterPassword = x.u != null && !string.IsNullOrWhiteSpace(x.u.MasterPassword)
+            HasMasterPassword = !string.IsNullOrWhiteSpace(x.u.MasterPassword)
```
justindbaur, line 33: Generated SQLite from before change

```sql
SELECT "o"."Id",
  "o"."UserId",
  "o"."OrganizationId",
  "u"."Name",
  COALESCE("u"."Email", "o"."Email") AS "Email",
  "u"."AvatarColor",
  "u"."TwoFactorProviders",
  "u"."Premium",
  "o"."Status",
  "o"."Type",
  "o"."ExternalId",
  "s"."ExternalId" AS "SsoExternalId",
  "o"."Permissions",
  "o"."ResetPasswordKey",
  "u"."Id" IS NOT NULL AND "u"."UsesKeyConnector" AS "UsesKeyConnector",
  "o"."AccessSecretsManager",
  "u"."Id" IS NOT NULL AND "u"."MasterPassword" IS NOT NULL AND trim("u"."MasterPassword") <> 'P|CfDJ8KUSA3v4Y1dFg8Rnf6FjA5fdvqaglKtClSQ1I3diP2vxzTdwq2_dCgDHF-chiyxdEEkJZ_XZnvZFhIpQGoqQL4f35sOzasRWeL8HXy0O59nSBVpX2WOp565OeY4fYRwuNg' AS "HasMasterPassword"
FROM "OrganizationUser" AS "o"
LEFT JOIN "User" AS "u" ON "o"."UserId" = "u"."Id"
LEFT JOIN "SsoUser" AS "s" ON "o"."UserId" = "s"."UserId" AND "o"."OrganizationId" = "s"."OrganizationId"
```

After change:
```sql
SELECT "o"."Id",
  "o"."UserId",
  "o"."OrganizationId",
  "u"."Name",
  COALESCE("u"."Email",
  "o"."Email") AS "Email",
  "u"."AvatarColor",
  "u"."TwoFactorProviders",
  "u"."Premium",
  "o"."Status",
  "o"."Type",
  "o"."ExternalId",
  "s"."ExternalId" AS "SsoExternalId",
  "o"."Permissions",
  "o"."ResetPasswordKey",
  "u"."UsesKeyConnector",
  "o"."AccessSecretsManager",
  "u"."MasterPassword" IS NOT NULL AND trim("u"."MasterPassword") <> 'P|CfDJ8KUSA3v4Y1dFg8Rnf6FjA5dJYW3Rjd8aosV-K01yVWmZVkVbItCW4oOuqXMvKO2PD9DIw0XKjy8sqgWtIKF1q7cT1JKgc3F2mwImAtRqP5WfrlURovVo9WdKmv2sI9lHUg' AS "HasMasterPassword"
FROM "OrganizationUser" AS "o"
LEFT JOIN "User" AS "u" ON "o"."UserId" = "u"."Id"
LEFT JOIN "SsoUser" AS "s" ON "o"."UserId" = "s"."UserId" AND "o"."OrganizationId" = "s"."OrganizationId"
```

And the equivalent hand wrote SQL:
```sql
SELECT
    OU.[Id],
    OU.[UserId],
    OU.[OrganizationId],
    U.[Name],
    ISNULL(U.[Email], OU.[Email]) Email,
    U.[AvatarColor],
    U.[TwoFactorProviders],
    U.[Premium],
    OU.[Status],
    OU.[Type],
    OU.[AccessSecretsManager],
    OU.[ExternalId],
    SU.[ExternalId] SsoExternalId,
    OU.[Permissions],
    OU.[ResetPasswordKey],
    U.[UsesKeyConnector],
    CASE WHEN U.[MasterPassword] IS NOT NULL THEN 1 ELSE 0 END AS HasMasterPassword
FROM
    [dbo].[OrganizationUser] OU
LEFT JOIN
    [dbo].[User] U ON U.[Id] = OU.[UserId]
LEFT JOIN
    [dbo].[SsoUser] SU ON SU.[UserId] = OU.[UserId] AND SU.[OrganizationId] = OU.[OrganizationId]
```

So this seems like a good change to me.

justindbaur (commented):
Just one note, thanks for the contribution!

src/Infrastructure.EntityFramework/AdminConsole/Repositories/Queries/OrganizationUserUserViewQuery.cs
eliykat: I also double checked this because I was concerned about orgUsers in the invited state that are not linked to a user, but they're excluded by the inner join on `User.Id`. So we should always have a User available. And as has been pointed out, we access `x.u` before this anyway, so clearly we do.

eliykat (approved):
Thanks for your contribution @MJebran ! The code owned by `team-admin-console-dev` looks good to me.

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -678,15 +678,14 @@ await _collectionCipherRepository.UpdateCollectionsForAdminAsync(cipher.Id,
         await _pushService.PushSyncCipherUpdateAsync(cipher, collectionIds);
     }
 
-    public async Task ImportCiphersAsync(
+    public async Task ImportFolderCiphersAsync(
```
djsmith85, line 681: Note to self: These should likely be moved into a tools-owned import service leveraging methods from the CipherService/Repository

djsmith85 (approved):
Approving for Tools

eliykat (approved):

cturnbull-bitwarden (dismissed):
Billing changes look good. Thank you for your contribution!

eliykat (commented):
There are some CI failures that are blocking merge.

1. formatting - please run `dotnet format` from the root of the repository and then commit the result.
2. tests failures - for some reason, removing the null check on `UsesKeyConnector` in `OrganizationUserUserViewQuery` has broken the SCIM integration tests. It throws a null reference exception even though when I debug everything seems to have a value. However, this also made me notice that we're using it in the SCIM `PostUserCommand` to return an invited user's details, however I think the inner join in this query would exclude invited users. I will find some time to look into this further.

amorask-bitwarden (approved):

shane-melton (approved):

jonashendrickx (approved):

src/Infrastructure.EntityFramework/AdminConsole/Repositories/Queries/OrganizationUserUserViewQuery.cs
```diff
@@ -28,9 +28,9 @@ from su in su_g.DefaultIfEmpty()
             SsoExternalId = x.su.ExternalId,
             Permissions = x.ou.Permissions,
             ResetPasswordKey = x.ou.ResetPasswordKey,
-            UsesKeyConnector = x.u != null && x.u.UsesKeyConnector,
+            UsesKeyConnector = x.u.UsesKeyConnector,
```
eliykat, line 31: This is causing the current failure in the CI test, because the [left join](https://github.com/bitwarden/server/pull/4896/files#r1823392614) on the `User` table can result in `UsesKeyConnector` being `NULL` - so it can't be assigned to a non-nullable value type. I suggest you revert this line or convert the null in some other way (e.g. `x.u.UsesKeyConnector == true`).

eliykat (changes requested):
Hi @MJebran, thanks for your patience with this one. I came back to it today and it still took way to long to figure out what was happening under the hood here. EF can be counterintuitive. Please address the issue below, merge in the latest `main`, and then we should be good to go.

PR #4903, opened 2024-10-16
[PM-13446] Add 'IsMultiOrgEnterprise' column to 'Provider' table

jonashendrickx:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13446

## 📔 Objective

Add an IsMultiOrgEnterprise column to the Provider table that will be used to differentiate between standard MSP’s and those who are multi-organization enterprises.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


util/Migrator/DbScripts/2024-10-16_00__AddIsMultiOrgEnterpriseColumn.sql
```diff
@@ -0,0 +1,175 @@
+-- Table
+IF OBJECT_ID('[dbo].[ClientOrganizationMigrationRecord]') IS NULL
```
amorask-bitwarden, line 2: ❌ It looks like you might have forgot to update the SQL Server migration script. This is just creating an unrelated table that already exists. Yours will look more like [this](https://github.com/bitwarden/server/blob/9f7e05869e2a0040b5b5e6197398f82c71e75b13/util/Migrator/DbScripts/2024-03-07_00_SetupProviderBilling.sql), but just with the new `IsMultiOrgEnterprise` column.

util/PostgresMigrations/Migrations/20241016140839_AddIsMultiOrgEnterpriseColumn.cs
```diff
@@ -0,0 +1,1739 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
+
+#nullable disable
+
+namespace Bit.PostgresMigrations.Migrations;
+
+/// <inheritdoc />
+public partial class AddIsMultiOrgEnterpriseColumn : Migration
```
amorask-bitwarden, line 9: ❌ Looks like something got messed up with your EF Migration here. It appears to be rebuilding the entire database in this file. Check out your [MySql migration file](https://github.com/bitwarden/server/pull/4903/files#diff-bd6759498181b48264485e5081fd40a5f9d754c4c0bcafa83f5ab288ddc8b707). That's what we want.

util/SqliteMigrations/Migrations/20241016140849_AddIsMultiOrgEnterpriseColumn.cs
```diff
@@ -0,0 +1,1733 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Bit.SqliteMigrations.Migrations;
+
+/// <inheritdoc />
+public partial class AddIsMultiOrgEnterpriseColumn : Migration
```
amorask-bitwarden, line 8: ❌ Same thing here as with Postgres.

amorask-bitwarden (changes requested):
Just a few things with your migration files.

eliykat (commented):
This seems like a perfect case for adding a new `ProviderType` enum. Is there a reason we can't use that, or a particular need for adding a new column instead?

amorask-bitwarden:
> This seems like a perfect case for adding a new `ProviderType` enum. Is there a reason we can't use that, or a particular need for adding a new column instead?

Hey @eliykat, we primarily decided against this because Multi-Enterprise Organizations are a subset of MSPs. Their functionality will be almost exactly the same as MSPs, but they will have different plan options during setup in the Bitwarden Portal. As such, it was my estimation that they don't actually represent a new provider type functionality-wise, but just an MSP that needs an indicator for a marginal functionality difference. This also allows us to re-use all of the existing Consolidated Billing functionality without having to add a new Enum check throughout the platform.

eliykat:
I think I have two main concerns about this:

1. MSP has a semantic meaning in this domain: it's a managed service provider business. That is just a different thing than a single enterprise with multiple Bitwarden organizations. Calling a multi-org provider something it's not (an MSP) and then relying on a different property to define what it _actually_ is, is confusing and a likely source of tech debt and bugs in the future. Especially if their functionality starts to diverge in the future. (See also - the Provider Information page in the Bitwarden Portal describes it as a type - I think the code should reflect the domain more closely.)
2. This design allows invalid combinations: you could have a `Reseller` provider type with `IsMultiOrgEnterprise` enabled, which (from what I can tell) would be an invalid combination. If another provider type is added in the future, is it a new `ProviderType` enum or another `bool` column? Either allows additional invalid combinations which shouldn't be possible just as a matter of data / schema design. This is exactly why we added `ProviderType` previously.

amorask-bitwarden:
> I think I have two main concerns about this:
> 
> 
> 
> 1. MSP has a semantic meaning in this domain: it's a managed service provider business. That is just a different thing than a single enterprise with multiple Bitwarden organizations. Calling a multi-org provider something it's not (an MSP) and then relying on a different property to define what it _actually_ is, is confusing and a likely source of tech debt and bugs in the future. (See also - the Provider Information page in the Bitwarden Portal describes it as a type - I think the code should reflect the domain more closely.)
> 
> 2. This design allows invalid combinations: you could have a `Reseller` provider type with `IsMultiOrgEnterprise` enabled, which (from what I can tell) would be an invalid combination. If another provider type is added in the future, is it a new `ProviderType` enum or another `bool` column? Either allows additional invalid combinations which shouldn't be possible just as a matter of data / schema design. This is exactly why we added `ProviderType` previously.
> 
> 
> 
> I admit I am not familiar with the consolidated billing code, but being able to have a simple enum check seems better to me than `enum && bool`, _or_ an implicit assumption that "MSP includes multi-org enterprises unless stated otherwise".

Understood, that seems valid and reasonable. We'll adjust the approach here accordingly. @jonashendrickx hold off on additional work here as we'll close this ticket and re-work the next one as the starting point.

PR #4912, opened 2024-10-17
[PM-14818]  Update `migrate.ps1` to support test database used by integration tests

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10863

## 📔 Objective

The integration tests for databases use different `secrets.json` properties than the ones used by locally running servers. This is to support not corrupting folk's local test data. 

Currently these test databases are not supported by `migrate.ps1` and other tooling. This PR updates `migrate.ps1` with a `--test` switch, and adds logic for connecting to and migrating these databases.

The `--all` switch was modified to include running migrations against test databases as well.

### 🧮 Side Effects

As part of this work I had to modify an old MySql migration because of an error in its manually written logic. The script in question checks for the existence of a column from system tables before trying to drop it. This doesn't check for a specific databases, and so can cause false positives if you try to have multiple databases on the same server.

## 📸 Screenshots

In this demo I build fresh databases from scratch and run some migrations using `migrate.ps1`. It: 

1. starts with a `pwsh migrate.ps1 --all` command that skips test databases because no connection strings are found.
2. Runs `pwsh migrate.ps1 --all` again, this time with connection strings set
3. Shows examples of `pwsh migrate.ps1 --test --{DATABASE}` for targeting specific test databases.
4. Shows the newly created databases in Azure Data Studio.

https://github.com/user-attachments/assets/32ee43c0-3c27-41d4-9c27-7f5a540e642e

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/dev/migrate.ps1
+++ b/dev/migrate.ps1
@@ -7,11 +7,13 @@ param(
   [switch]$mysql,
   [switch]$mssql,
   [switch]$sqlite,
-  [switch]$selfhost
+  [switch]$selfhost,
+  [switch]$test
 )
 
 # Abort on any error
 $ErrorActionPreference = "Stop"
+$currentDir = Get-Location
 
 if (!$all -and !$postgres -and !$mysql -and !$sqlite) {
   $mssql = $true;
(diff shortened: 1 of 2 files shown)


util/MySqlMigrations/Migrations/20231214162533_GrantIdWithIndexes.cs
eliykat, line 1: This makes sense to me but requesting a review from @rkac-bw to confirm this retroactive change is OK.

dev/migrate.ps1
```diff
@@ -25,36 +27,62 @@ if ($all -or $postgres -or $mysql -or $sqlite) {
...
+  }
 }
 
-$currentDir = Get-Location
-
-Foreach ($item in @(@($mysql, "MySQL", "MySqlMigrations"), @($postgres, "PostgreSQL", "PostgresMigrations"), @($sqlite, "SQLite", "SqliteMigrations"))) {
+Foreach ($item in @(
+    @($mysql, "MySQL", "MySqlMigrations", "mySql", 2),
+    @($postgres, "PostgreSQL", "PostgresMigrations", "postgreSql", 0),
+    @($sqlite, "SQLite", "SqliteMigrations", "sqlite", 1)
```
eliykat, line 65: The test attributes (`DatabaseData`, `DatabaseTheory`) allow these providers to be defined in any order, but this script assumes that postgres is 0, sqlite is 1, mySql is 2 and mssql is 3. That won't be the case for everyone, depending on how they've set up their user secrets.

You could add logic here iterating over the user secrets to find the index for each database type, but it might make more sense to update the structure of the user secrets themselves to adopt a more conventional key-value format, like `test:database:mysql:connectionString`. That would also require updating the test attributes, so it's a little in the weeds, but it would make sure this Just Works for everyone.

eliykat (commented):
Thanks for taking the time to follow this one up.

util/MySqlMigrations/Migrations/20231214162533_GrantIdWithIndexes.cs
rkac-bw: Yes, the change for MySQL makes sense.  The existence check for the Id column in the Grant table is scoped to the current database schema. This prevents false positives when multiple databases are present on the same server, avoiding unintended changes to other databases.

rkac-bw (approved):
LGTM

dev/migrate.ps1
justindbaur: We can't change the format to be `test:database:mysql:connectionString` where `mysql` is treated as the `SupportedDatabaseProviders` enum. This would limit you from being able to setup the same provider multiple times which these tests were explicitly made to support. See mariadb being [added to the pipeline here](https://github.com/bitwarden/server/pull/4989). I would also want it to be able to run against different versions of the same database. If we want to change it to where it's `databases:[pretty name]:connectionString` but then there is still `databases:[pretty name]:type` then I'd be fine with that but that doesn't really solve making this easier. 

Also, for what it's worth I am planning a pretty big overhaul to these tests when xUnit v3 releases. The main thing being it's added a `AssemblyFixtureAttribute` which:
> can be applied at the assembly-level to add an assembly-wide fixture. Fixtures at this level are created before any test is run, and cleaned up after all tests have finished running.
This is what I have been waiting for to make auto instrumentation possible, so the hope is that we will automatically create a database and run migrations for all possible database types. And when it's not able to run on certain types it will show those types as skipped so that it's a bit more obvious that you are missing some configuration.

eliykat: That makes sense! I hadn't considered multiple providers of the same type. In the meantime, maybe we just keep the fixed array index numbers in the script, or add some logic to identify which type each array entry is?

eliykat (approved):
Thanks for following up on this!

PR #4920, opened 2024-10-21
[PM-13447] Add Multi Org Enterprise providers to Admin Console

jonashendrickx:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13447

## 📔 Objective

- Add the ability to create a multi-organization enterprise provider in the admin console. This is only available for current enterprise plans (not the legacy plans).

- Provider types are ordered using the 'Order' property on the 'DisplayAttribute'. Product wanted to have them grouped together based on whether a provider portal would be used for the specific provider type.

- Refactored: Provider creation has been refactored, so we can reduce code duplication and keep the front-end code more simple. If we go with Blazor or Razor Components we could make it easier to handle multiple forms on one page. Life-cycle management with ASP.NET MVC seems to make things unnecessarily complicated.

The [Settings](https://github.com/bitwarden/passwordless-server/blob/main/src/AdminConsole/Components/Pages/App/Settings/Settings.razor) page for an application at Passwordless.dev is a good example of a page with multiple forms in Blazor. 

## 📸 Screenshots

Redesigned provider selection, with ordering changes:
image

Flow showing separate screens:
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Admin/AdminConsole/Models/CreateMultiOrganizationEnterpriseProviderModel.cs
+++ b/src/Admin/AdminConsole/Models/CreateMultiOrganizationEnterpriseProviderModel.cs
@@ -1,0 +1,62 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Core.AdminConsole.Entities.Provider;
+using Bit.Core.AdminConsole.Enums.Provider;
+using Bit.Core.Billing.Enums;
+using Bit.SharedWeb.Utilities;
+
+namespace Bit.Admin.AdminConsole.Models;
+
+public class CreateMultiOrganizationEnterpriseProviderModel : IValidatableObject
+{
+    [Display(Name = "Provider Type")]
+    public ProviderType Type { get; set; }
+
+    [Display(Name = "Owner Email")]
+    public string OwnerEmail { get; set; }
+
+    [Display(Name = "Name")]
+    public string Name { get; set; }
+
+    [Display(Name = "Business Name")]
+    public string BusinessName { get; set; }
+
+    [Display(Name = "Primary Billing Email")]
+    public string BillingEmail { get; set; }
+
+
+    [Display(Name = "Enterprise Seat Minimum")]
+    public int EnterpriseSeatMinimum { get; set; }
+
+    [Display(Name = "Plan")]
+    public PlanType Plan { get; set; }
+
+    public virtual Provider ToProvider()
+    {
+        return new Provider
+        {
+            Type = Type,
+            Name = Name,
+            BusinessName = BusinessName,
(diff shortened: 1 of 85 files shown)


src/Admin/AdminConsole/Models/CreateMultiOrganizationEnterpriseProviderModel.cs
```diff
@@ -0,0 +1,43 @@
...
+public class CreateMultiOrganizationEnterpriseProviderModel : IValidatableObject
+{
+    [Display(Name = "Owner Email")]
+    public string OwnerEmail { get; set; }
+
+    [Display(Name = "Enterprise Seat Minimum")]
+    public int EnterpriseSeatMinimum { get; set; }
+
+    [Display(Name = "Plan")]
+    [Required]
```
jonashendrickx, line 17: This improves the error message returned when no plan was selected. Otherwise you get: "The value '' is invalid". Now it returns a proper error message.

src/Admin/AdminConsole/Views/Providers/Create.cshtml
```diff
@@ -1,80 +1,36 @@
...
+
 @model CreateProviderModel
+
 @inject Bit.Core.Services.IFeatureService FeatureService
+
 @{
     ViewData["Title"] = "Create Provider";
-}
 
-@section Scripts {
```
jonashendrickx, line 10: There's no longer the necessity to toggle the visibility of all sections as they're now using their own pages to keep sanity with maintainability having multiple forms on this page otherwise.

amorask-bitwarden (approved):
Looks great from the Billing side, nice work.

src/Admin/Enums/HtmlHelperExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿
+using Bit.SharedWeb.Utilities;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.AspNetCore.Mvc.Rendering;
+
+public static class HtmlHelper
+{
+    public static IEnumerable<SelectListItem> GetEnumSelectList<T>(this IHtmlHelper htmlHelper, IEnumerable<T> values)
```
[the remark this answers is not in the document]
jonashendrickx: The issue I had with it was that I needed to pass the type. But then it would return everything. Now I just pass the enum values we want converted for the drop down.

BTreston (approved):

BTreston (changes requested):
Looking at it again I feel like this could use some unit test coverage. The changes look good otherwise

BTreston (approved):

PR #4940, opened 2024-10-24
[PM-13014] - Add CanToggleStatus property to PolicyRepsonseModel based on Policy Validators

jrmccannon:
## 🎟️ Tracking
[PM-13014](https://bitwarden.atlassian.net/browse/PM-13014)

## 📔 Objective
This will add a `CanToggleStatus` flag to the `PolicyResponseModel` to inform clients on whether or not they will be allowed to change the status of the policy (enable/disable). This is using the new `PolicyValidator#ValidateAsync` methods to determine if it can be changed. This is limited to `Single Org Policy` for now.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-13014]: https://bitwarden.atlassian.net/browse/PM-13014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -52,6 +59,14 @@ public PoliciesController(
...
 
         _orgUserInviteTokenDataFactory = orgUserInviteTokenDataFactory;
+        _featureService = featureService;
+
+        var dictionary = new Dictionary<PolicyType, IPolicyValidator>();
+        foreach (var validator in validators)
+        {
+            dictionary.TryAdd(validator.Type, validator);
+        }
+        _policyValidators = dictionary;
```
jonashendrickx, line 69:
```suggestion
                _policyValidators = validators.ToDictionary(v => v.Type, v => v);
```

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -68,6 +83,22 @@ public async Task<PolicyResponseModel> Get(string orgId, int type)
             throw new NotFoundException();
         }
 
+        if (_featureService.IsEnabled(FeatureFlagKeys.AccountDeprovisioning))
```
eliykat, line 86:
```suggestion
        if (_featureService.IsEnabled(FeatureFlagKeys.AccountDeprovisioning && policy.Type == PolicyType.SingleOrg))
```

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -81,7 +112,30 @@ public async Task<ListResponseModel<PolicyResponseModel>> Get(string orgId)
         }
 
         var policies = await _policyRepository.GetManyByOrganizationIdAsync(orgIdGuid);
-        var responses = policies.Select(p => new PolicyResponseModel(p));
```
eliykat, line 84: Correct me if I'm overlooking something, but I can't see that we need this property on this endpoint. The policy add-edit dialog only uses the single GET endpoint. Each policy validator can make arbitrary database calls, so I don't want to call them in a loop here unnecessarily.

I suggest splitting the response models so that the single GET endpoint is returning an extended response model containing the new property, and this endpoint is unchanged.

src/Api/AdminConsole/Public/Controllers/PoliciesController.cs
eliykat, line 1: The new property is only used by our UI, I don't think it needs to be added to the public API.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidator.cs
```diff
@@ -92,10 +100,20 @@ public async Task<string> ValidateAsync(PolicyUpdate policyUpdate, Policy? curre
     {
         if (policyUpdate is not { Enabled: true })
         {
+            var resultString = new StringBuilder();
+
             var ssoConfig = await _ssoConfigRepository.GetByOrganizationIdAsync(policyUpdate.OrganizationId);
-            return ssoConfig.ValidateDecryptionOptionsNotEnabled([MemberDecryptionType.KeyConnector]);
+            resultString.Append(ssoConfig.ValidateDecryptionOptionsNotEnabled([MemberDecryptionType.KeyConnector]));
```
eliykat, line 106: I'm not sure we have a way of displaying multiple errors in the UI, and in any case this would just return one big long string. Better to just return the first validation error that occurs I think. Although you could certainly follow up with design to see if there's some way to display all errors up-front.

eliykat (changes requested):

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -68,6 +83,22 @@ public async Task<PolicyResponseModel> Get(string orgId, int type)
...
+                    .ValidateAsync(
+                        new PolicyUpdate
+                        {
+                            Data = policy.Data,
+                            Enabled = !policy.Enabled,
+                            OrganizationId = policy.OrganizationId,
+                            Type = policy.Type
+                        }, policy));
+
+            return new PolicyResponseModel(policy, canToggle);
```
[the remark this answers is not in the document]
eliykat: See also my questions in `#team-eng-adminconsole` about this pattern. If we disable the UI for other reasons, I'm not sure the user will understand why it's disabled. That is a question for Product/Design, just mentioning it here for completeness.

src/Api/AdminConsole/Public/Controllers/PoliciesController.cs
jrmccannon: It could be added to the API to let the users who consume the public API that their request will probably fail. 🤷 I can remove it though if its unnecessary.

src/Api/AdminConsole/Models/Response/Helpers/PolicyDetailResponses.cs
```diff
@@ -0,0 +1,19 @@
...
+public static class PolicyDetailResponses
+{
+    public static async Task<PolicyDetailResponseModel> GetSingleOrgPolicyDetailResponseAsync(Policy policy, IOrganizationHasVerifiedDomainsQuery hasVerifiedDomainsQuery)
+    {
+        if (policy.Type is not PolicyType.SingleOrg)
+        {
+            throw new ArgumentException($"'{nameof(policy)}' must be of type '{nameof(PolicyType.SingleOrg)}'.", nameof(policy));
+        }
+
+        return new PolicyDetailResponseModel(policy, await hasVerifiedDomainsQuery.HasVerifiedDomainsAsync(policy.OrganizationId));
```
eliykat, line 17: I think the bool needs to be inverted:
- has verified domains => can't toggle state
- does not have verified domains => can toggle state

(Probably a drawback of a bool that defaults to `true`, I didn't realise that when I suggested the naming, although it's not a huge deal)

src/Api/AdminConsole/Models/Response/Helpers/PolicyDetailResponses.cs
eliykat, line 1: No changes required, but I am interested in the decision to use an extension method here vs. a static method on the class itself. I can make some guesses but would like to hear your thoughts.

src/Core/AdminConsole/Models/Api/Response/PolicyDetailResponseModel.cs
eliykat, line 1: Api request/response models live in the `Api` project.

eliykat (changes requested):

src/Api/AdminConsole/Models/Response/Helpers/PolicyDetailResponses.cs
jrmccannon: Personally, I like to have my models/types/etc pretty plain. The extension method here is more a mapping function. Take a policy and turn it into a PolicyDetailResponseModel with the IOrganizationHasVerifiedDomainsQuery.  

That was just my thinking about it.  Having this class exist to hold each policies mapping from an entity to the response with its own logic makes for a nice place to organize them all. There could eventually be a factory method that handles all policies and instantiates them based on type, but for now this is pretty explicit.

eliykat: Thanks for the explanation!

eliykat: I also missed that it's an extension of the domain class rather than the response. Which helps keep Api concerns out of Core, which I like.

eliykat (approved):

Jingo88 (approved):

ike-kottlowski (approved):

PR #4943, opened 2024-10-25
[PM-11408] domain verification stat in portal and add cs delete permission

BTreston:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-11408

## 📔 Objective

Adds the domain verification state in the admin portal when viewing a user, gated behind a feature flag. Adds the User.delete permission to the cs role. 

## 📸 Screenshots
Screenshot 2024-10-24 at 12 39 52 PM

Screenshot 2024-10-24 at 12 40 08 PM

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Admin/Controllers/UsersController.cs
+++ b/src/Admin/Controllers/UsersController.cs
@@ -4,6 +4,7 @@ using Bit.Admin.Enums;
 using Bit.Admin.Models;
 using Bit.Admin.Services;
 using Bit.Admin.Utilities;
+using Bit.Core;
 using Bit.Core.Auth.UserFeatures.TwoFactorAuth.Interfaces;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
@@ -25,6 +26,7 @@ public class UsersController : Controller
     private readonly IAccessControlService _accessControlService;
     private readonly ITwoFactorIsEnabledQuery _twoFactorIsEnabledQuery;
     private readonly IOrganizationRepository _organizationRepository;
+    private readonly IFeatureService _featureService;
 
     public UsersController(
         IUserRepository userRepository,
@@ -33,7 +35,8 @@ public class UsersController : Controller
         GlobalSettings globalSettings,
         IAccessControlService accessControlService,
         ITwoFactorIsEnabledQuery twoFactorIsEnabledQuery,
-        IOrganizationRepository organizationRepository)
+        IOrganizationRepository organizationRepository,
+        IFeatureService featureService)
     {
         _userRepository = userRepository;
         _cipherRepository = cipherRepository;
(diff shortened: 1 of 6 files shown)


src/Admin/Controllers/UsersController.cs
```diff
@@ -99,7 +107,9 @@ public async Task<IActionResult> Edit(Guid id)
         var billingInfo = await _paymentService.GetBillingAsync(user);
         var billingHistoryInfo = await _paymentService.GetBillingHistoryAsync(user);
         var isTwoFactorEnabled = await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(user);
-        return View(new UserEditModel(user, isTwoFactorEnabled, ciphers, billingInfo, billingHistoryInfo, _globalSettings));
+        var organizationsWithVerifiedUserEmailDomain = await _organizationRepository.GetByVerifiedUserEmailDomainAsync(id);
+        var verifiedDomain = organizationsWithVerifiedUserEmailDomain.Count > 0;
```
r-tome, line 111: Same thing here, use https://github.com/bitwarden/server/blob/109ba14cf40025f36ef11c46c2a3fdde0d823538/src/Core/Services/IUserService.cs#L99

src/Admin/Controllers/UsersController.cs
```diff
@@ -153,4 +163,10 @@ public async Task<IActionResult> Delete(Guid id)
 
         return RedirectToAction("Index");
     }
+
+    /* FEATURE FLAG CHECK, TO BE REMOVED IN THE FUTURE*/
+    public bool accountDeprovisioningEnabled()
```
r-tome, line 168: This method should be `private` as its not used anywhere else and use PascalCase for naming.

I would change this to return a nullable boolean like so:
```

public async Task<bool?> IsUserDomainVerifiedAsync(Guid userId)
    {
        return _featureService.IsEnabled(FeatureFlagKeys.AccountDeprovisioning)
            ? await _userService.IsManagedByAnyOrganizationAsync(userId)
            : null;
    }
```
With this approach, there's no need to pass the feature flag to the ViewModels. If the value is `null`, assume the feature is disabled and omit it from the view.

src/Admin/Views/Users/_ViewInformation.cshtml
```diff
@@ -12,6 +12,12 @@
     <dt class="col-sm-4 col-lg-3">Email Verified</dt>
     <dd class="col-sm-8 col-lg-9">@(Model.EmailVerified ? "Yes" : "No")</dd>
 
+    @* FEATURE FLAG TO CONTROLL IF THIS SHOULD ACTUALLY SHOW *@
+    @if(Model.AccountDeprovisioningEnabled){
```
r-tome, line 16: If the property is nullable then check it here
```suggestion
    @if(Model.DomainVerified.HasValue){
```

r-tome (changes requested):
I think this can be simplified a bit by not having to pass the feature flag into the ViewModels. Cleanup later will also be easier

r-tome (commented):
Just a minor nitpick

r-tome (approved):

PR #4959, opened 2024-10-30
[PM-12479] - Adding group-details endpoint

jrmccannon:
## 🎟️ Tracking
[PM-12479](https://bitwarden.atlassian.net/browse/PM-12479)

## 📔 Objective
This is step one in restricting the data provided by `organizations/{id}/groups`.  If the feature flag is enabled, it will return only the group value with the collections null. In the future,  the collections property will be removed. This also includes a new endpoint `organizations/{id}/group-details` which will return the group and collections only for admins, owners, providers and users allowed to manage groups or members.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-12479]: https://bitwarden.atlassian.net/browse/PM-12479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
eliykat, line 1: This is **optional** depending on how deep you want to go with testing. But xUnit supports an attribute called `MemberData` which can be used to parameterize tests with complex objects: https://hamidmosalla.com/2017/02/25/xunit-theory-working-with-inlinedata-memberdata-classdata/ (scroll down to the Member Data heading). All these tests are quite similar, I wonder if you could combine almost all of them by defining an object containing data for all test cases (e.g. all combinations of role, custom permissions, expectedResult) and consuming it with the `MemberData` attribute in a single test.

We have a wrapper around it called [BitMemberAutoDataAttribute](https://github.com/bitwarden/server/blob/2abd37d2d7f19c6fbe1ec5efedf6e915dafd1746/test/Common/AutoFixture/Attributes/BitMemberAutoDataAttribute.cs#L1-L1) which may provide some examples.

eliykat (changes requested):

src/Api/AdminConsole/Controllers/GroupsController.cs
```diff
@@ -16,7 +17,7 @@
 
 namespace Bit.Api.AdminConsole.Controllers;
 
-[Route("organizations/{orgId}/groups")]
+[Route("organizations/{orgId}")]
```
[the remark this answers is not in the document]
jrmccannon: The top level was removed since the new route would be `organizations/{id}/group-details` since group details isn't following `/groups`. I believe the only way to override that would be to put a `/` at the start of the route, but then you'd override any additional route pathing that was set up higher in the stack.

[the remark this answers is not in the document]
eliykat: In that case I suggest leaving `groups` in the top-level route and using `details` at the method level, i.e. `organizations/{id}/groups/details`. (We already do this for collection details.) My main concern is that all our controllers use a controller-level route, this would be an outlier that is easily missed. This also keeps all groups-related resources under the `groups` path which I think might be better REST design (although that might be arguable).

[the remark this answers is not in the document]
eliykat: I see I specified `group-details` in the Jira ticket, that's my bad!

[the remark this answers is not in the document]
jrmccannon: No worries. Easy fix. I'll move group back to the top and make this one details. I'll update the client code accordingly as well.

src/Api/Api.csproj
eliykat, line 1: nit: introduced some random whitespace here

src/Core/AdminConsole/OrganizationFeatures/Groups/Authorization/GroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+using Bit.Core.AdminConsole.OrganizationFeatures.Shared.Authorization;
+using Bit.Core.Context;
+using Bit.Core.Enums;
+using Microsoft.AspNetCore.Authorization;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Groups.Authorization;
+
+/// <summary>
+/// Handles authorization logic for Group operations.
+/// This uses new logic implemented in the Flexible Collections initiative.
```
eliykat, line 11: You can delete this line referring to Flexible Collections - maybe this whole comment if you don't think it's useful.

src/Core/AdminConsole/OrganizationFeatures/Groups/Authorization/GroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+/// </summary>
+public class GroupAuthorizationHandler(ICurrentContext currentContext)
+    : AuthorizationHandler<GroupOperationRequirement, OrganizationScope>
+{
+    protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context,
+        GroupOperationRequirement requirement, OrganizationScope organizationScope)
+    {
+        var authorized = requirement switch
```
eliykat, line 19: Nice use of this switch syntax

src/Core/AdminConsole/OrganizationFeatures/Groups/Authorization/GroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private async Task<bool> CanReadAllAsync(OrganizationScope organizationScope) =>
+        currentContext.GetOrganization(organizationScope) is not null
+        || await currentContext.ProviderUserForOrgAsync(organizationScope);
+
+    private async Task<bool> CanViewGroupDetailsAsync(OrganizationScope organizationScope) =>
+        currentContext.GetOrganization(organizationScope) is { Type: OrganizationUserType.Owner } or { Type: OrganizationUserType.Admin }
+            or { Permissions: { ManageGroups: true } or { ManageUsers: true } }
+        || await currentContext.ProviderUserForOrgAsync(organizationScope);
```
eliykat, line 41: We have some [code style guidelines](https://contributing.bitwarden.com/contributing/code-style/csharp#conditionals) that say boolean operators go at the end of lines. I don't feel strongly about it but I feel obliged to mention it.

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
```diff
@@ -86,40 +89,142 @@ public async Task CanReadAllAsync_WhenMissingOrgAccess_NoSuccess(
...
+    {
+        organization.Type = OrganizationUserType.User;
+        organization.Permissions = new Permissions();
+
+        var context = new AuthorizationHandlerContext(
+            new[] { GroupOperations.ReadAll },
+            new ClaimsPrincipal(),
+            scope);
+
+        sutProvider.GetDependency<ICurrentContext>().UserId.Returns(userId);
```
eliykat, line 222: I assume this is copy/paste from the existing tests. The new sut doesn't check `currentContext.UserId` so this is unnecessary.

eliykat (changes requested):
Great work, just a few minor things then good to go.

test/Api.Test/Vault/AuthorizationHandlers/GroupAuthorizationHandlerTests.cs
jrmccannon: This was leftover from an older iteration. I'll remove it. Good catch.

src/Core/AdminConsole/OrganizationFeatures/Groups/Authorization/GroupAuthorizationHandler.cs
jrmccannon: While I disagree and prefer the opposite, I will bow to the whims of the majority. 😄

eliykat (approved):

vincentsalucci (approved):
Approving on behalf of @eliykat. Only new change was moving related test file for the `GroupAuthorizationHandler` into AC team scope (matches changes to other `GroupAuthorization` files)

PR #5034, opened 2024-11-13
[PM-13026] Refactor remove and bulkremove methods to throw error if user is managed by an organization

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13026

## 📔 Objective

> Verified members should not be removed. Instead, verified members should go through the deletion flow to ensure all their individual data is properly deleted. This change will prevent accidental removals and enforce consistent deletion practices for verified members.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
```diff
@@ -106,21 +114,33 @@ public async Task<List<Tuple<OrganizationUser, string>>> RemoveUsersAsync(Guid o
                     throw new BadRequestException("Only owners can delete other owners.");
                 }
 
+                if (managementStatus.TryGetValue(orgUser.Id, out var isManaged) && isManaged)
+                {
+                    throw new BadRequestException("Managed members cannot be simply removed, their entire individual account must be deleted.");
```
eliykat, line 119: Please check with Product/Design about this wording if you haven't already.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
```diff
@@ -106,21 +114,33 @@ public async Task<List<Tuple<OrganizationUser, string>>> RemoveUsersAsync(Guid o
...
+                }
```
eliykat, line 120: This command is also used by SCIM which can continue to remove managed users (integrations are excluded from the delete behavior until we figure out how to do it safely). So there needs to be some conditionality here. Maybe stealing `actingUser` from Jared's PR would be useful so it can encapsulate `deletingUserId` or the `SystemUser` (and in that case you can exempt SCIM from this check).

I don't love it but I think my problem is with the diverging business logic. Handling it in the command like this seems reasonable.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
r-tome: Thanks for reminding me! I did check and changed the copy.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
r-tome: I have refactored this so that we don't check managed users when an `EventSystemUser` is passed.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IRemoveOrganizationUserCommand.cs
eliykat, line 1: As general feedback: I think this command has too many overloads, I appreciate the xmldoc but it remains confusing. I think getting rid of the `UserId` overload (per the comment above) and then using the `ActingUser` class (which should be merged shortly) will help combine some of these. Not required in this PR though.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
```diff
@@ -34,123 +47,107 @@ public RemoveOrganizationUserCommand(
...
 
-    public async Task RemoveUserAsync(Guid organizationId, Guid userId)
+    public async Task RemoveUserAsync(Guid organizationId, Guid organizationUserId, EventSystemUser eventSystemUser)
     {
-        var organizationUser = await _organizationUserRepository.GetByOrganizationAsync(organizationId, userId);
+        var organizationUser = await _organizationUserRepository.GetByIdAsync(organizationUserId);
         ValidateDeleteUser(organizationId, organizationUser);
 
-        await RepositoryDeleteUserAsync(organizationUser, null);
+        await RepositoryDeleteUserAsync(organizationUser, null, eventSystemUser);
```
eliykat, line 80: Please use a named argument when passing in `null`s to these private methods to avoid mixups, e.g. `await RepositoryDeleteUserAsync(organizationUser, deletingUserId: null, eventSystemUser);`

eliykat (changes requested):
This is bordering on how much refactoring I feel comfortable with without it being feature flagged. However it _mostly_ seems like a lift-and-shift of existing code so I think it's fine for now, especially given that it's fixing the UserId bug. Please just get QA to regression test all these flows when they test that bug.

I think there are more improvements we could make to the structure of this command (e.g. reducing duplicate logic) but again that can wait until another time when we can better feature flag it. For now I think it works, I just have a few minor bits of feedback.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IRemoveOrganizationUserCommand.cs
```diff
@@ -1,14 +1,53 @@
...
 namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
 
 public interface IRemoveOrganizationUserCommand
 {
+    /// <summary>
+    /// Removes a user from an organization.
+    /// </summary>
+    /// <param name="organizationId">The ID of the organization.</param>
+    /// <param name="userId">The ID of the user to remove.</param>
+    Task RemoveUserAsync(Guid organizationId, Guid userId);
```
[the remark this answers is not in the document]
r-tome: Its not really a bug though -- its different from the other methods because it retrieves the `OrganizationUser` object by filtering by both `OrganizationId` **and** `UserId`.

I can create the follow-up ticket. I didn't want to refactor too much here, my intention was just to tidy this up a bit whilst adding the new error if the user is managed.

eliykat (approved):
Looks good, thanks!

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IRemoveOrganizationUserCommand.cs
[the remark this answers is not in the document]
eliykat: My bad! This showed up on the diff as a new line so I thought you were adding this interface, which would've meant that existing callers in `main` today were calling one of the other methods with a `UserId`. But you just moved it. Disregard what I said about there being a bug - but please still create a follow up ticket :)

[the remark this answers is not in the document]
r-tome: Ticket created [here](https://bitwarden.atlassian.net/browse/PM-15427)

PR #5035, opened 2024-11-13
[PM-13013] add delete many async method to i user repository and i user service for bulk user deletion

BTreston:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13013

## 📔 Objective

Creates the method `DeleteManyAsync` and adds `User_DeleteByIds` stored procedure to enable bulk deletion of users

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BTreston:
WIP: I am yet to add this functionality to `IUserService`.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -167,7 +167,19 @@ public override async Task DeleteAsync(User user)
         {
             await connection.ExecuteAsync(
                 $"[{Schema}].[{Table}_DeleteById]",
```
BTreston, line 169: needs to be `$"[{Schema}].[{Table}_DeleteByIds]"`. Same for below

src/Sql/dbo/Stored Procedures/User_DeleteById.sql
r-tome, line 1: It may be better to avoid changing this stored procedure and instead create a new one called `User_DeleteManyByIds`. This approach could simplify maintenance by allowing the database to be updated first without needing to worry about request failures if the API hasn't been updated yet.

Then, in a later release, we can switch to just using the new sproc and delete the old one.

src/Sql/dbo/Stored Procedures/User_DeleteById.sql
```diff
@@ -1,5 +1,5 @@
-﻿CREATE PROCEDURE [dbo].[User_DeleteById]
-    @Id UNIQUEIDENTIFIER
+﻿CREATE PROCEDURE [dbo].[User_DeleteByIds]
+    @Ids [dbo].[GuidIdArray]
```
r-tome, line 2: We're no longer using `GuidIdArray`, instead we parse a json string. Example: https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Stored%20Procedures/User_ReadByIdsWithCalculatedPremium.sql

test/Infrastructure.EFIntegration.Test/Repositories/UserRepositoryTests.cs
r-tome, line 1: `Infrastructure.EFIntegration.Test` is deprecated. Database testing should be done in `Infrastructure.Integration.Test`.

r-tome (changes requested):
Looking great! We'll sync up later today and discuss the suggestions I've made

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -167,7 +167,19 @@ public override async Task DeleteAsync(User user)
...
-                new { Id = user.Id },
+                new { Ids = new List<Guid> { user.Id }.ToGuidIdArrayTVP() },
```
r-tome, line 170: Revert this change to pass the single Id. The failing integration test will pass then.

src/Sql/dbo/Stored Procedures/User_DeleteByIds.sql
```diff
@@ -0,0 +1,158 @@
...
+    BEGIN
+        RETURN(-1);
+    END
+
+    DECLARE @BatchSize INT = 100
+
+    -- Delete ciphers
+    WHILE @BatchSize > 0
+    BEGIN
+        BEGIN TRANSACTION User_DeleteById_Ciphers
```
r-tome, line 26: I think we may want a big transaction that scopes deleting everything related to each single user. @rkac-bw will have much better insights

r-tome (changes requested):
Great work! There's just a tiny change to revert on the single delete method.

test/Infrastructure.IntegrationTest/Auth/Repositories/UserRepoistoryTests.cs
eliykat, line 1: Typo in the filename: `UserRepoistory`

src/Core/Services/Implementations/UserService.cs
```diff
@@ -32,6 +32,8 @@
 using File = System.IO.File;
 using JsonSerializer = System.Text.Json.JsonSerializer;
 
+#nullable enable
```
eliykat, line 35: This is probably enabling nullable for too much of the file - check out all of the Github annotations below in the diff view. Only enable nullable on code that expects it. (i.e. you've written or reviewed it to make sure it handles nullability properly)

src/Core/Services/Implementations/UserService.cs
eliykat, line 1: Please make sure to add / update tests on this file.

test/Infrastructure.IntegrationTest/Auth/Repositories/UserRepoistoryTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        await userRepository.DeleteAsync(user);
+
+        var newUser = await userRepository.GetByIdAsync(user.Id);
+        Assert.NotNull(newUser);
+        Assert.NotEqual(newUser.AccountRevisionDate, user.AccountRevisionDate);
+    }
+
+    [DatabaseTheory, DatabaseData]
+    public async Task DeleteManyAsync_Works(IUserRepository userRepository)
```
eliykat, line 28: I would also add a third user who is not deleted, then assert that they are still present in the db.
You could also add a few relations, e.g. ciphers or orgUsers, to make sure that the delete cascades properly.

eliykat (commented):
My review was automatically requested but I can see that @r-tome is already reviewing this one, so I've just left a couple of high level comments and will let him handle the rest.

src/Core/Services/IUserService.cs
```diff
@@ -45,6 +45,7 @@ Task<IdentityResult> ChangeKdfAsync(User user, string masterPassword, string new
     Task<string> GenerateUserTokenAsync(User user, string tokenProvider, string purpose);
     Task<IdentityResult> DeleteAsync(User user);
     Task<IdentityResult> DeleteAsync(User user, string token);
+    Task<IEnumerable<(Guid UserId, string? ErrorMessage)>> DeleteManyAsync(IEnumerable<User> users);
```
r-tome, line 48: You need to enable nullable reference types

```suggestion
#nullable enable
        Task<IEnumerable<(Guid UserId, string? ErrorMessage)>> DeleteManyAsync(IEnumerable<User> users);
#nullable disable
```

src/Core/Services/Implementations/UserService.cs
```diff
@@ -223,59 +223,39 @@ public async Task SaveUserAsync(User user, bool push = false)
...
         {
             return IdentityResult.Failed(new IdentityError
             {
-                Description = "Cannot delete this user because it is the sole owner of at least one provider. Please delete these providers or upgrade another user.",
+                Description = results
             });
         }
+    }
+
+    public async Task<IEnumerable<(Guid UserId, string? ErrorMessage)>> DeleteManyAsync(IEnumerable<User> users)
```
r-tome, line 246: Also enable nullable reference types for this method

src/Core/Services/Implementations/UserService.cs
```diff
@@ -223,59 +223,39 @@ public async Task SaveUserAsync(User user, bool push = false)
...
+    {
+        var results = await ValidateDeleteUsersAsync(users.ToList());
 
-        if (!string.IsNullOrWhiteSpace(user.GatewaySubscriptionId))
+        await _userRepository.DeleteManyAsync(users);
```
r-tome, line 250: You should only delete users that have passed validation. Something like
```suggestion
        var usersToDelete = users.Where(u => string.IsNullOrEmpty(u.ErrorMessage);
        await _userRepository.DeleteManyAsync(usersToDelete);
        foreach (var user in usersToDelete)
```

src/Core/Services/Implementations/UserService.cs
```diff
@@ -288,6 +268,61 @@ public async Task<IdentityResult> DeleteAsync(User user, string token)
         return await DeleteAsync(user);
     }
 
+    private async Task<IEnumerable<(Guid UserId, string? ErrorMessage)>> ValidateDeleteUsersAsync(List<User> users)
```
r-tome, line 271: Enable nullable reference types

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -36,6 +38,7 @@ public DeleteManagedOrganizationUserAccountCommand(
         _userRepository = userRepository;
         _currentContext = currentContext;
         _hasConfirmedOwnersExceptQuery = hasConfirmedOwnersExceptQuery;
+        _featureService = featureService;
     }
 
     public async Task DeleteUserAsync(Guid organizationId, Guid organizationUserId, Guid? deletingUserId)
```
r-tome, line 44: I'd prefere if you create a new version of this method for when the account deprovisioning flag is enabled. So check the flag first thing in this method and redirect to the new one if enabled
```suggestion
    public async Task DeleteUserAsync(Guid organizationId, Guid organizationUserId, Guid? deletingUserId)
    {
        if _featureService.IsEnabled(FeatureFlagKeys.AccountDeprovisioning))
        {
            return await DeleteUserAsync_vNext(organizationId, organizationUserId, deletingUserId);
        }
```

r-tome (changes requested):

eliykat (commented):
My feedback has been resolved, thanks! Over to @r-tome.

src/Core/Services/Implementations/UserService.cs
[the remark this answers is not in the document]
eliykat: Should the bulk delete even be in the `UserService`? The only time it will ever be used is when an admin is doing the action, and the validation isn't a perfect fit. It made sense for us to reuse the existing `UserService` method when doing an individual delete initially, but maybe we revisit this if we're at the stage of writing separate implementations anyway.

src/Core/Services/Implementations/UserService.cs
```diff
@@ -288,6 +272,63 @@ public async Task<IdentityResult> DeleteAsync(User user, string token)
...
+        foreach (var user in users)
+        {
+            // Check if user is the only owner of any organizations.
+            var onlyOwnerCount = await _organizationUserRepository.GetCountByOnlyOwnerAsync(user.Id);
+            if (onlyOwnerCount > 0)
+            {
+                var deletedOrg = false;
+                var orgs = await _organizationUserRepository.GetManyDetailsByUserAsync(user.Id,
+                    OrganizationUserStatusType.Confirmed);
+                if (orgs.Count == 1)
```
eliykat, line 289: For example: if we're deleting a user, and they're the sole owner of an organization, then they must belong to at least 2 organizations (our org + the org they own), so this will always evaluate to false. I don't think this would do any harm, but as I said, it's not a great fit.

eliykat (commented):
Looked at this again and had some concerns about the changes in UserService.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
eliykat: But this command is only used by account deprovisioning?

r-tome: Yes, I was confused! I cleared this up with @BTreston and forgot to edit this.

src/Core/Services/Implementations/UserService.cs
```diff
@@ -259,6 +259,11 @@ public override async Task<IdentityResult> DeleteAsync(User user)
             return IdentityResult.Failed(new IdentityError
             {
                 Description = "Cannot delete this user because it is the sole owner of at least one provider. Please delete these providers or upgrade another user.",
+
```
r-tome, line 262: Empty lines added

r-tome (changes requested):

r-tome (dismissed):
LGTM! Thanks for making the changes I requested. Great work on this!

src/Core/Services/Implementations/UserService.cs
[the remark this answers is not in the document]
JaredSnider-Bitwarden: @eliykat , yes, our team has been moving methods out of services into commands whenever possible to adhere to [ADR-0008](https://contributing.bitwarden.com/architecture/server/#cqrs-adr-0008). 

I am fine with y'all owning the `DeleteManyCommand` as well since it is an org admin only action.

eliykat (commented):
The feedback [here](https://github.com/bitwarden/server/pull/5035#discussion_r1866691587) suggests that `UserService.DeleteManyAsync` doesn't belong in UserService and should be moved into a command under AC Team ownership.

Given that this code is only used for an admin deleting claimed users, I suggest;
1. it is moved into the existing `DeleteManagedOrganizationUserCommand` as a private method
2. its validation logic is simplified, mostly `// Check if user is the only owner of any organizations`. At the moment it's a straight copy-paste of the UserService logic, but as I mentioned [above](https://github.com/bitwarden/server/pull/5035/files#r1858004088) not all of the code applies (even though the overall business requirement still does).
3. the existing code in the command is updated to call this instead of UserService.

Patrick-Pimentel-Bitwarden (dismissed):
Tests and code looks good 👍

BTreston:
@eliykat I've implemented the requested changes, though I'm not 100% sure on if the onlyProviderCount and gateway validation checks are necessary or why they are unnecessary if not. Probably my inexperience with the product, but the checks seem to make sense to me.

rkac-bw (approved):
lgtm

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -98,6 +114,7 @@ public async Task DeleteUserAsync(Guid organizationId, Guid organizationUserId,
             }
         }
 
+        await DeleteManyAsync(users);
```
eliykat, line 117: `DeleteManyAsync` returns a result object which is unused. It needs to add to the `results` variable we already have. Also, its validation logic should be better incorporated into the validation steps above.

I suggest the simplest way to do this is:
- refactor the private `ValidateDeleteUsersAsync` you added to take a single user. The method is just a loop anyway. Maybe rename it something to differentiate it, e.g. `ValidateUserMembershipAndPremium`
- also refactor it to throw its error message
- call it from within the loop above, just before `results.Add((orgUserId, string.Empty));`. That way any error will be caught and added to the existing results object
- then by the time `DeleteManyAsync` is called, we know all validation has passed

eliykat, line 117: Also, `DeleteManyAsync` should only receive users that have passed validation. (i.e. where they have an empty string in their `results` entry.)

test/Core.Test/Services/UserServiceTests.cs
eliykat, line 1: These tests aren't really testing anything (no real assertions) and also aren't required now that we're not making changes to this class.

eliykat (changes requested):
I think the command could be better structured but we can leave that for a follow-up ticket. For now I think this will get it in good order. Let me know if you want to discuss or pair on it.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -114,7 +117,9 @@ public async Task DeleteUserAsync(Guid organizationId, Guid organizationUserId,
             }
         }
 
-        await DeleteManyAsync(users);
+        var usersToDelete = users.Where(user => results.Select(r => r.OrganizationUserId == user.Id && string.IsNullOrEmpty(r.ErrorMessage)).Count() > 1);
```
eliykat, line 120: Non-blocking, fyi only: the idiomatic way to write this is:
```suggestion
        var usersToDelete = users.Where(user => results.Any(r => r.OrganizationUserId == user.Id && string.IsNullOrEmpty(r.ErrorMessage)));
```

It's the equivalent of `Array.some` in JS/TS.

eliykat (dismissed):
LGTM, @r-tome do you mind giving it another look over to double check me?

EDIT: I created a ticket here with some suggestions for follow-up refactors to complete the loop here: https://bitwarden.atlassian.net/browse/PM-15621. However they can be post-release.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -19,15 +23,24 @@ public class DeleteManagedOrganizationUserAccountCommand : IDeleteManagedOrganiz
     private readonly IUserRepository _userRepository;
     private readonly ICurrentContext _currentContext;
     private readonly IHasConfirmedOwnersExceptQuery _hasConfirmedOwnersExceptQuery;
-
+    private readonly IFeatureService _featureService;
```
r-tome, line 26: `IFeatureService` is not being used, it can be removed

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommandTests.cs
```diff
@@ -235,7 +235,7 @@ await sutProvider.GetDependency<IEventService>().Received(0)
 
     [Theory]
     [BitAutoData]
-    public async Task DeleteManyUsersAsync_WithValidUsers_DeletesUsersAndLogsEvents(
+    public async Task DeleteManyUsersAsync_WhenFeatureFlagEnabled_WithValidUsers_DeletesUsersAndLogsEvents(
```
r-tome, line 238: Feature flag is not used
```suggestion
    public async Task DeleteManyUsersAsync_WithValidUsers_DeletesUsersAndLogsEvents(
```

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommandTests.cs
```diff
@@ -257,15 +257,62 @@ public async Task DeleteManyUsersAsync_WithValidUsers_DeletesUsersAndLogsEvents(
...
+        await sutProvider.GetDependency<IUserService>().Received(0).DeleteAsync(Arg.Any<User>());
+        await sutProvider.GetDependency<IEventService>().Received(1).LogOrganizationUserEventsAsync(
+            Arg.Is<IEnumerable<(OrganizationUser, EventType, DateTime?)>>(events =>
+                events.Count(e => e.Item1.Id == orgUser1.Id && e.Item2 == EventType.OrganizationUser_Deleted) == 1
+                && events.Count(e => e.Item1.Id == orgUser2.Id && e.Item2 == EventType.OrganizationUser_Deleted) == 1));
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task DeleteManyUsersAsync_WhenFeatureFlagDisabled_WithValidUsers_DeletesUsersAndLogsEvents(
```
r-tome, line 281: This can be deleted because the feature flag is not used

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -98,6 +117,9 @@ public async Task DeleteUserAsync(Guid organizationId, Guid organizationUserId,
             }
         }
 
+        var usersToDelete = users.Where(user => results.Select(r => r.OrganizationUserId == user.Id && string.IsNullOrEmpty(r.ErrorMessage)).Count() > 1);
+
+        await DeleteManyAsync(usersToDelete);
```
r-tome, line 122:
```suggestion
        if (usersToDelete.Any())
        {
            await DeleteManyAsync(usersToDelete);
        }
```

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommandTests.cs
```diff
@@ -274,11 +321,38 @@ await sutProvider.GetDependency<IEventService>().Received(1).LogOrganizationUser
 
     [Theory]
     [BitAutoData]
-    public async Task DeleteManyUsersAsync_WhenUserNotFound_ReturnsErrorMessage(
+    public async Task DeleteManyUsersAsync_WhenFeatureFlagEnabled_WhenUserNotFound_ReturnsErrorMessage(
```
r-tome, line 324:
```suggestion
    public async Task DeleteManyUsersAsync_WhenUserNotFound_ReturnsErrorMessage(
```

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommandTests.cs
```diff
@@ -274,11 +321,38 @@ await sutProvider.GetDependency<IEventService>().Received(1).LogOrganizationUser
...
+        Assert.Single(result);
+        Assert.Equal(orgUserId, result.First().Item1);
+        Assert.Contains("Member not found.", result.First().Item2);
+        await sutProvider.GetDependency<IEventService>().Received(0)
+            .LogOrganizationUserEventsAsync(Arg.Any<IEnumerable<(OrganizationUser, EventType, DateTime?)>>());
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task DeleteManyUsersAsync_WhenFeatureFlagDisabled_WhenUserNotFound_ReturnsErrorMessage(
```
r-tome, line 346: Can be deleted

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommandTests.cs
```diff
@@ -274,11 +321,38 @@ await sutProvider.GetDependency<IEventService>().Received(1).LogOrganizationUser
...
+        sutProvider.GetDependency<IFeatureService>()
+            .IsEnabled(FeatureFlagKeys.AccountDeprovisioning).Returns(true);
+
+        // Act
+        var result = await sutProvider.Sut.DeleteManyUsersAsync(organizationId, new[] { orgUserId }, null);
+
+        // Assert
+        Assert.Single(result);
+        Assert.Equal(orgUserId, result.First().Item1);
+        Assert.Contains("Member not found.", result.First().Item2);
```
r-tome, line 339: Assert that nothing was deleted
```suggestion
        Assert.Contains("Member not found.", result.First().Item2);
        await sutProvider.GetDependency<IUserRepository>()
            .DidNotReceiveWithAnyArgs()
            .DeleteManyAsync(default);
```

r-tome (changes requested):
Looking good, just left a few suggestions and then its good to go

r-tome (approved):
Great work!

rkac-bw (approved):
lgtm

JaredSnider-Bitwarden (approved):
LGTM

PR #5037, opened 2024-11-13
[PM-10319] - Revoke Non Complaint Users for 2FA and Single Org Policy Enablement

jrmccannon:
## 🎟️ Tracking
[PM-10319](https://bitwarden.atlassian.net/browse/PM-10319)

## 📔 Objective
This change revokes non-compliant users when 2FA or Single Org policies are enabled.  This adds the a `RevokeNonCompliantOrganizationUserCommand` along with a bulk set status stored procedure. It also adds two new email templates to be sent when a user is revoked from an organization.

## 📸 Screenshots
Two FA Policy Revocation Email
image

Single Org Policy Email
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10319]: https://bitwarden.atlassian.net/browse/PM-10319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


r-tome (commented):
Looks good, left some suggestions

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Requests/RevokeOrganizationUser.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Bit.Core.AdminConsole.Models.Data;
+using Bit.Core.Models.Data.Organizations.OrganizationUsers;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Requests;
+
+public record RevokeOrganizationUsers(
+    Guid OrganizationId,
+    IEnumerable<OrganizationUserUserDetails> OrganizationUsers,
```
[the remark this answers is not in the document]
jrmccannon: I would need to add Status to IOrganizationUser and update all the spots where it is referenced.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidator.cs
```diff
@@ -80,17 +85,42 @@ private async Task RemoveNonCompliantUsersAsync(Guid organizationId)
...
-                await _removeOrganizationUserCommand.RemoveUserAsync(organizationId, orgUser.Id,
-                    savingUserId);
+                throw new BadRequestException(string.Join(", ", commandResult.ErrorMessages));
+            }
 
-                await _mailService.SendOrganizationUserRemovedForPolicySingleOrgEmailAsync(
-                    org.DisplayName(), orgUser.Email);
+            await Task.WhenAll(revocableUsers.Select(x =>
+                _mailService.SendOrganizationUserRevokedForPolicySingleOrgEmailAsync(org.DisplayName(), x.Email)));
+        }
```
eliykat, line 109: The command does not allow a mixed success / fail state, right? I was initially concerned that if any member could not be revoked, then nobody gets their email. But looking at the command, it validates everyone upfront and then doesn't proceed if there are any errors, so I think that's OK.

That said, if an admin revokes multiple users from the UI, we _do_ allow a success or failure state per user, so we may need to accommodate both behaviors if the command is expanded to replace the OrganizationService methods.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidator.cs
eliykat, line 1: Same comments as other validator

src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
```diff
@@ -58,4 +58,6 @@ UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
     /// Returns a list of OrganizationUsers with email domains that match one of the Organization's claimed domains.
     /// </summary>
     Task<ICollection<OrganizationUser>> GetManyByOrganizationWithClaimedDomainsAsync(Guid organizationId);
+
+    Task SetOrganizationUsersStatusAsync(IEnumerable<Guid> userIds, OrganizationUserStatusType status);
```
eliykat, line 62: Not all `OrganizationUserStatusType` changes are so simple. Changing from Invited -> Accepted and Accepted -> Confirmed requires other updates to the data (e.g. linking the Organization User to the User, doing key exchanges), so this cannot be used for those transitions. At the moment it would be limited to revoking users and restoring users. Which is fine, but how can we limit this interface so it's not misused? Maybe the SQL is general purpose, but the repository interface remains specific to the revoke action?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    private async Task<CommandResult> ValidateAsync(RevokeOrganizationUsersRequest request)
+    {
+        if (!PerformedByIsAnExpectedType(request.ActionPerformedBy))
+        {
+            return new CommandResult([ErrorRequestedByWasNotValid]);
+        }
+
+        if (request.ActionPerformedBy is StandardUser loggableUser
+            && request.OrganizationUsers.Any(x => x.UserId == loggableUser.UserId))
```
eliykat, line 68: Outdated reference to loggableUser?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    private static (OrganizationUserUserDetails organizationUser, EventType eventType, EventSystemUser eventSystemUser, DateTime? time) GetRevokedUserEventBySystemUserTuple(
+        OrganizationUserUserDetails organizationUser, EventSystemUser systemUser, DateTimeOffset dateTimeOffset) => new(organizationUser,
+        EventType.OrganizationUser_Revoked, systemUser, dateTimeOffset.UtcDateTime);
+
+    private async Task<CommandResult> ValidateAsync(RevokeOrganizationUsersRequest request)
+    {
+        if (!PerformedByIsAnExpectedType(request.ActionPerformedBy))
+        {
+            return new CommandResult([ErrorRequestedByWasNotValid]);
```
eliykat, line 64: Seems like an overload accepting a single error message would be useful.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+    private static bool IsAlreadyRevoked(OrganizationUserUserDetails organizationUser) =>
+        organizationUser is { Status: OrganizationUserStatusType.Revoked };
+
+    private static bool NonOwnersCannotRevokeOwners(OrganizationUserUserDetails organizationUser,
+        IActingUser actingUser) =>
+        actingUser is StandardUser { IsOrganizationOwner: false } && organizationUser.Type == OrganizationUserType.Owner;
+
+    private static bool IsActingUserAllowedToRevokeOwner(OrganizationUserUserDetails organizationUser,
+        StandardUser requestingOrganizationUser) => organizationUser is { Type: OrganizationUserType.Owner }
+                                                    && requestingOrganizationUser.IsOrganizationOwner;
```
eliykat, line 114: I think this is unused (and duplicates the previous method).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        if (validationResult.HasErrors)
+        {
+            return validationResult;
+        }
+
+        await organizationUserRepository.SetOrganizationUsersStatusAsync(request.OrganizationUsers.Select(x => x.Id),
+            OrganizationUserStatusType.Revoked);
+
+        if (request.ActionPerformedBy.GetType() == typeof(StandardUser))
```
eliykat, line 36: Do we need to use reflection here? You use pattern matching everywhere else.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+            await eventService.LogOrganizationUserEventsAsync(
+                request.OrganizationUsers.Select(x => GetRevokedUserEventTuple(x, timeProvider.GetUtcNow())));
+        }
+        else if (request.ActionPerformedBy is SystemUser { SystemUserType: not null } loggableSystem)
+        {
+            await eventService.LogOrganizationUserEventsAsync(
+                request.OrganizationUsers.Select(x =>
+                    GetRevokedUserEventBySystemUserTuple(x, loggableSystem.SystemUserType.Value,
+                        timeProvider.GetUtcNow())));
+        }
```
eliykat, line 47: This will log slightly different times. Usually we call `GetUtcNow` once and then use that 1 reference in the `Select` callback for consistency.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
```diff
@@ -0,0 +1,115 @@
...
+        return validationResult;
+    }
+
+    private static (OrganizationUserUserDetails organizationUser, EventType eventType, DateTime? time) GetRevokedUserEventTuple(
+        OrganizationUserUserDetails organizationUser, DateTimeOffset dateTimeOffset) => new(organizationUser,
+        EventType.OrganizationUser_Revoked, dateTimeOffset.UtcDateTime);
+
+    private static (OrganizationUserUserDetails organizationUser, EventType eventType, EventSystemUser eventSystemUser, DateTime? time) GetRevokedUserEventBySystemUserTuple(
+        OrganizationUserUserDetails organizationUser, EventSystemUser systemUser, DateTimeOffset dateTimeOffset) => new(organizationUser,
+        EventType.OrganizationUser_Revoked, systemUser, dateTimeOffset.UtcDateTime);
```
eliykat, line 58: This seems like a lot of boilerplate when you could just pass in a tuple directly. Is it doing much for you?

src/Core/Services/Implementations/HandlebarsMailService.cs
```diff
@@ -295,6 +294,20 @@ public async Task SendOrganizationUserRemovedForPolicyTwoStepEmailAsync(string o
         await _mailDeliveryService.SendEmailAsync(message);
     }
 
+    public async Task SendOrganizationUserRevokedForTwoFactoryPolicyEmailAsync(string organizationName, string email)
+    {
+        var message = CreateDefaultMessage($"You have been revoked from {organizationName}", email);
```
eliykat, line 299: Does the org name need to be sanitized here? (@r-tome you did all the sanitization stuff I think?)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Requests/RevokeOrganizationUser.cs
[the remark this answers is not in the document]
r-tome: I didn't realise it was so much effort!

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -721,4 +721,14 @@ public UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(
...
     }
+
+    public async Task SetOrganizationUsersStatusAsync(IEnumerable<Guid> userIds, OrganizationUserStatusType status)
+    {
+        using var scope = ServiceScopeFactory.CreateScope();
+
+        var dbContext = GetDatabaseContext(scope);
+
+        await dbContext.OrganizationUsers.Where(x => userIds.Contains(x.OrganizationId))
+            .ExecuteUpdateAsync(s => s.SetProperty(x => x.Status, status));
```
eliykat, line 732: You also need to bump the account revision dates here. See https://github.com/bitwarden/server/blob/e16cad50b16a8bed011f86c039e6892c4ebfe9cf/src/Infrastructure.EntityFramework/Repositories/DatabaseContextExtensions.cs#L122-L123.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidatorTests.cs
```diff
@@ -126,4 +127,72 @@ await sutProvider.GetDependency<IMailService>()
...
+
+        await sutProvider.Sut.OnSaveSideEffectsAsync(policyUpdate, policy);
+
+        await sutProvider.GetDependency<IRevokeNonCompliantOrganizationUserCommand>()
+            .DidNotReceive()
+            .RevokeNonCompliantOrganizationUsersAsync(Arg.Any<RevokeOrganizationUsersRequest>());
+    }
+    // TODO feature flag is enabled and revoke returns errors, we throw
+    // TODO feature flag is enabled and revoke returns no errors, no throw
+    // TODO feature flag is disabled we don't call command
```
eliykat, line 197: TODOs?

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidatorTests.cs
```diff
@@ -201,9 +206,99 @@ public async Task OnSaveSideEffectsAsync_UsersToBeRemovedDontHaveMasterPasswords
...
+
+        await sutProvider.Sut.OnSaveSideEffectsAsync(policyUpdate, policy);
+
+        await sutProvider.GetDependency<IRevokeNonCompliantOrganizationUserCommand>()
+            .DidNotReceive()
+            .RevokeNonCompliantOrganizationUsersAsync(Arg.Any<RevokeOrganizationUsersRequest>());
+    }
+
+    [Theory, BitAutoData]
+    public async Task OnSaveSideEffectsAsync_GivenUpdateTo2faPolicy_WhenAccountProvisioningIsEnabled_ThenRevokeUserCommandShouldNotBeCalled(
```
eliykat, line 265: The test name is not descriptive here - I think the case you're testing, more specifically, is that it won't revoke a user without a master password.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidator.cs
```diff
@@ -55,33 +66,66 @@ private async Task RemoveNonCompliantUsersAsync(Guid organizationId)
...
+        {
+            if (DoMembersNotHaveAPasswordWithTwoFactorAuthenticationEnabled(revocableOrgUsers, organizationUsersTwoFactorEnabled))
+            {
+                throw new BadRequestException(NonCompliantMembersWillLoseAccess);
+            }
+
+            var result = await _revokeNonCompliantOrganizationUserCommand.RevokeNonCompliantOrganizationUsersAsync(
+                new RevokeOrganizationUsersRequest(organizationId, revocableOrgUsers,
+                    new StandardUser(savingUserId ?? Guid.Empty,
+                        await _currentContext.OrganizationOwner(organizationId))));
```
eliykat, line 89: So here's a fun one. `currentContext.OrganizationOwner` implicitly includes providers. This is... _somewhat_ known when dealing with `currentContext`. But I recommend avoiding this ambiguity in your new object - e.g. adding an xmldoc noting that `IsOwner` can include providers in this context.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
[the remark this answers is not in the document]
jrmccannon: I think it could, but I would want to do that change in a separate PR. I also don't want to change the name until then so we know its safe to be the de facto RevokeOrganizationUserCommand

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
jrmccannon: Idk. It removes noise from where the actual action is happening. I'd rather have simple method signature in the select than `new ValueTuple<OrganizationUserUserDetails, EventType, DateTime?>(organizationUser,
            EventType.OrganizationUser_Revoked, dateTimeOffset.UtcDateTime);`.

src/Core/Services/Implementations/HandlebarsMailService.cs
eliykat: Thanks!

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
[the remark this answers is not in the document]
eliykat: That's cool, can you please make a follow-up tech debt ticket to look into this?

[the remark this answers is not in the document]
jrmccannon: Created:
https://bitwarden.atlassian.net/browse/PM-15052

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidator.cs
jrmccannon: I'll do you one better. I'll change it to `IsOwnerOrProvider`

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidator.cs
jrmccannon: The complicated thing about this command is all the validation logic.  What we could do is break out the validation into a validator class, then have two commands (one for the 2FA/Single Org policy enablement and one for a user explicitly revoking a group of users). Then each command can make the decision about enforcing the validation result.

Unless we just want the command to just only reject invalid responses. Then we would make a Response object to extend CommandResult.  Maybe something like
```
public class RevokeOrganizationUserCommandResponse : CommandResult 
{
...
public Guid[] SuccessfullyRevokedUsers { get; } 
public Guid[] FailedToRevokeUsers { get; }
...
}

```

That way we should show the partial success of the action.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidatorTests.cs
```diff
@@ -126,4 +127,69 @@ await sutProvider.GetDependency<IMailService>()
             .SendOrganizationUserRemovedForPolicySingleOrgEmailAsync(organization.DisplayName(),
                 "user3@example.com");
     }
+
+    [Theory, BitAutoData]
+    public async Task OnSaveSideEffectsAsync_GivenUpdatedSingleOrgPolicy_WhenAccountDeprovisioningIsDisabled_Then(
```
r-tome, line 132: I would either rename this to something clearer like `OnSaveSideEffectsAsync_WhenAccountDeprovisioningDisabled_DoesNotRevokeUsers ` or I would update the test above `OnSaveSideEffectsAsync_RemovesNonCompliantUsers` to assert that it does not revoke users

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidatorTests.cs
r-tome, line 1: Its missing testing when AccountDeprovisioning is enabled and users are revoked

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidatorTests.cs
r-tome, line 1: We're missing a test for the happy path when account deprovisioning is enabled and users have master passwords (should call RevokeNonCompliantOrganizationUsersAsync)

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidatorTests.cs
```diff
@@ -201,9 +206,99 @@ public async Task OnSaveSideEffectsAsync_UsersToBeRemovedDontHaveMasterPasswords
...
 
-        Assert.Contains("Policy could not be enabled. Non-compliant members will lose access to their accounts. Identify members without two-step login from the policies column in the members page.", badRequestException.Message, StringComparison.OrdinalIgnoreCase);
+        Assert.Equal(TwoFactorAuthenticationPolicyValidator.NonCompliantMembersWillLoseAccess, badRequestException.Message);
 
         await sutProvider.GetDependency<IRemoveOrganizationUserCommand>().DidNotReceiveWithAnyArgs()
             .RemoveUserAsync(organizationId: default, organizationUserId: default, deletingUserId: default);
     }
+
+    [Theory, BitAutoData]
+    public async Task OnSaveSideEffectsAsync_GivenUpdateTo2faPolicy_WhenAccountProvisioningIsDisabled_ThenRevokeUserCommandShouldNotBeCalled(
```
r-tome, line 216: `OnSaveSideEffectsAsync_RemovesNonCompliantUsers` could verify both the removal of users AND that revoke isn't called and then this test isn't required

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
```diff
@@ -63,19 +67,24 @@ public PolicyService(
...
+        _currentContext = currentContext;
     }
 
-    public async Task SaveAsync(Policy policy, Guid? savingUserId)
+    public async Task SaveAsync(Policy policy, Guid? savingUserId, EventSystemUser? eventSystemUser = null)
     {
         if (_featureService.IsEnabled(FeatureFlagKeys.Pm13322AddPolicyDefinitions))
         {
             // Transitional mapping - this will be moved to callers once the feature flag is removed
+            // TODO make sure to populate with SystemUser if not an actual user
```
r-tome, line 78: TODO no longer needed?

src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
```diff
@@ -58,4 +58,6 @@ UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
     /// Returns a list of OrganizationUsers with email domains that match one of the Organization's claimed domains.
     /// </summary>
     Task<ICollection<OrganizationUser>> GetManyByOrganizationWithClaimedDomainsAsync(Guid organizationId);
+
+    Task RevokeOrganizationUserAsync(IEnumerable<Guid> userIds);
```
r-tome, line 62: Please add xmldoc here. Also, I would suggest renaming this to fit in with our usual naming scheme:
```suggestion
    Task RevokeManyByUserIdAsync(IEnumerable<Guid> userIds);
```

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommandTests.cs
```diff
@@ -0,0 +1,175 @@
...
+            .HasConfirmedOwnersExceptAsync(organizationId, Arg.Any<IEnumerable<Guid>>())
+            .Returns(true);
+
+        var result = await sutProvider.Sut.RevokeNonCompliantOrganizationUsersAsync(command);
+
+        await sutProvider.GetDependency<IOrganizationUserRepository>()
+            .Received(1)
+            .RevokeOrganizationUserAsync(Arg.Any<IEnumerable<Guid>>());
+
+        Assert.True(result.Success);
```
r-tome, line 166: Verify event logging for the revoked user

r-tome (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/VerifyOrganizationDomainCommand.cs
```diff
@@ -12,124 +14,114 @@
...
-        _policyService = policyService;
-        _featureService = featureService;
-        _logger = logger;
-    }
 
 
     public async Task<OrganizationDomain> UserVerifyOrganizationDomainAsync(OrganizationDomain organizationDomain)
     {
-        var domainVerificationResult = await VerifyOrganizationDomainAsync(organizationDomain);
+        var actingUser = new StandardUser(currentContext.UserId ?? Guid.Empty, await currentContext.OrganizationOwner(organizationDomain.OrganizationId));
```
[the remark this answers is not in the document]
jrmccannon: I'd rather not have null. I could throw at this point instead of using Guid.Empty. I was just defaulting to Guid.Empty in the event that a caller hasn't initialized the CurrentContext.

Do we know which applications would call this method without the CurrentContext populated with a user?

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
jrmccannon: This TODO will get addressed with a subsequent story.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidator.cs
eliykat: All good ideas. I've asked Priya for feedback in the team channel about whether we can align this behavior. No other changes required for now.

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/VerifyOrganizationDomainCommand.cs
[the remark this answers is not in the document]
eliykat: I'm not sure about _this method_ specifically, but generally:
- the Public API and SCIM - `CurrentContext.UserId` is null because it's the organization that's authenticated, not the user
- a system call (jobs) - also not tied to a user

A regular call from a client through the private api will always have UserId populated (to my knowledge) because CurrentContext is always initialized in the middleware and the private api always requires user authentication.

What this means for you here I'm not sure. This method seems specifically designed to be called by a user, so throwing if there is no UserId available seems reasonable. And if the UserId is null for some reason, then `_currentContext.OrganizationOwner` is probably not going to be populated either (as this is also user specific).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeNonCompliantOrganizationUserCommand.cs
eliykat: Oh, it's explicitly a `ValueTuple` rather than just a (named) tuple. Yep then I agree that's cumbersome otherwise.

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
eliykat: I can do this when I remove the policy definition feature flag shortly. Callers will have to construct their own `PolicyUpdate` objects, so they'll need to make sure they're specifying the right actingUser.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidator.cs
```diff
@@ -56,10 +56,58 @@ public async Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? curr
...
+        var organization = await _organizationRepository.GetByIdAsync(organizationId);
+
+        var currentActiveRevocableOrganizationUsers =
+            (await _organizationUserRepository.GetManyDetailsByOrganizationAsync(organizationId))
+            .Where(ou => ou.Status != OrganizationUserStatusType.Invited &&
+                         ou.Status != OrganizationUserStatusType.Revoked &&
+                         ou.Type != OrganizationUserType.Owner &&
+                         ou.Type != OrganizationUserType.Admin &&
+                         performedBy is StandardUser stdUser &&
+                         stdUser.UserId != ou.UserId)
```
eliykat, line 83: Same here

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/TwoFactorAuthenticationPolicyValidator.cs
```diff
@@ -56,10 +56,58 @@ public async Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? curr
...
+
+        if (currentActiveRevocableOrganizationUsers.Count == 0)
+        {
+            return;
+        }
+
+        var organizationUsersTwoFactorEnabled =
+            await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(currentActiveRevocableOrganizationUsers);
+
+        if (DoMembersNotHaveAPasswordWithTwoFactorAuthenticationEnabled(currentActiveRevocableOrganizationUsers, organizationUsersTwoFactorEnabled))
```
eliykat, line 94: I'm not sure this method name is accurate; shouldn't it be "**Without**TwoFactorAuthenticationEnabled"? Or maybe the double negative is throwing me. How about `if (NonCompliantMembersWillLoseAccess)`?

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SingleOrgPolicyValidatorTests.cs
```diff
@@ -117,19 +121,40 @@ public async Task OnSaveSideEffectsAsync_RemovesNonCompliantUsers(
...
+        else
+        {
+            await sutProvider.GetDependency<IRemoveOrganizationUserCommand>()
+                .Received(1)
+                .RemoveUserAsync(policyUpdate.OrganizationId, nonCompliantUser.Id, savingUserId);
+            await sutProvider.GetDependency<IMailService>()
+                .Received(1)
+                .SendOrganizationUserRemovedForPolicySingleOrgEmailAsync(organization.DisplayName(),
+                    "user3@example.com");
+        }
```
eliykat, line 152: Not a blocker, but I generally avoid conditional assertions in tests; this suggests to me that you should just have 2 tests. That's easier to clean up as well, because you can just delete the whole test.

eliykat (changes requested):
Looking good, just a couple of things.

Also see my [feedback above](https://github.com/bitwarden/server/pull/5037#discussion_r1851307500) re: empty guid vs. throwing.

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/VerifyOrganizationDomainCommand.cs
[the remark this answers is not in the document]
jrmccannon: Added exception throw redirecting caller to use the other method. Not sure which exception to throw, so just going with BadRequest for now.

eliykat (approved):

r-tome (dismissed):
Great work!

rkac-bw (dismissed):
lgtm

eliykat (approved):

rkac-bw (approved):
lgtm

PR #5051, opened 2024-11-20
[PM-11360] Remove export permission for providers

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-11360
## 📔 Objective

Remove provider export permissions. Provider users will not have the ability to access an organization’s export tab or export a managed organization’s items.

These changes have a few side effects which are listed below.

### Code changes
1. add feature flag
2. add `vNext` version of `OrganizationExportController` endpoint
    - this has been rewritten per our current patterns to use authorizationService and queries
3. add an authorization handler and separate authorization logic from query logic
    - this handler no longer permits providers to export (the functional change)
    - this also fixes the confusing and contradictory authorization logic in the current service methods
4. mark old service methods as obsolete (these are not used anywhere else and can be removed with feature flag cleanup)
5. implicit removal of backwards compatibility code for pre-2023.1.0 clients (this is in the current controller method but is now outside of our support window so I haven't replicated it in the new method)

Note that the managed collection export code path is only used by CLI. This is because the GUI clients use their sync data for this export, not the endpoint, but CLI was never updated to do this. I have decided to support this functionality on the endpoint so that (1) I didn't have to feature flag a non-web client, and (2) I didn't break existing CLI clients who may have been using this.

@djsmith85 I suggest that Tools Team could revisit the CLI later and switch it over to using sync data (to be consistent with the other clients), then you can also provide a proper lead time of 3 major releases before removing the old code from the server so as not to break existing workflows.

Clients PR (already merged): https://github.com/bitwarden/clients/pull/12062
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Tools/Authorization/VaultExportAuthorizationHandler.cs
eliykat, line 1: @djsmith85 I've put this under Tools ownership given that it relates to export functionality, but I've raised it for discussion in `#dev-tech-leads` in case other teams think that AC should own these.

src/Api/Tools/Authorization/VaultExportAuthorizationHandler.cs
eliykat, line 1: @djsmith85 I've put this under Tools ownership given that it relates to export functionality, but I've raised it for discussion in `#dev-tech-leads` in case other teams think that AC should own these.

eliykat:
I decided to keep the export managed collections codepath. I've updated the description with this info.

JimmyVo16:
I'm taking a look at this now.

shane-melton (approved):
Vault changes look good!

JimmyVo16 (approved):

src/Api/Tools/Authorization/VaultExportAuthorizationHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+    : AuthorizationHandler<VaultExportOperationRequirement, OrganizationScope>
+{
+    protected override Task HandleRequirementAsync(AuthorizationHandlerContext context,
+        VaultExportOperationRequirement requirement, OrganizationScope organizationScope)
+    {
+        var org = currentContext.GetOrganization(organizationScope);
+
+        var authorized = requirement switch
+        {
+            not null when requirement == VaultExportOperations.ExportWholeVault =>
```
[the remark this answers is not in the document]
eliykat: This is using the ASP.NET [resource-based authorization](https://learn.microsoft.com/en-us/aspnet/core/security/authorization/resourcebased?view=aspnetcore-8.0) framework. I agree that the interfaces are very general - for example, you can pass any combination of operation & resource to AuthorizationService even if you don't have a handler defined for it - but the benefit is that it maintains a loose coupling between the controller and the authorization logic. It's basically a [chain of responsibility pattern](https://refactoring.guru/design-patterns/chain-of-responsibility). Hit me up if you want to discuss further!

djsmith85 (approved):
Approving for Tools! @eliykat Thanks for working on this. We'll follow up with the CLI work and I've seen some more areas that could use some clean up

Issue #4845, opened 2024-10-03
Emergency Access E-Mails use Wrong Encoding for Names


windmueller:
### Steps To Reproduce

1. Configure a name for your account using special characters like a German umlaut
2. Invite someone for emergency access
3. Confirm the invatation on both sides

### Expected Result

The e-mails should display the special characters correctly.

### Actual Result

The actual e-mails look like:

```
You have been invited to become an emergency contact for Stephan Windm&#252;ller.
If you do not wish to become an emergency contact for Stephan Windm&#252;ller, you can safely ignore this email.
```

and

```
This email is to notify you that you have been confirmed as an emergency access contact for Stephan Windm&#252;ller.
```

### Build Version

2024.9.2

### Environment

Cloud (bitwarden.com)

### Issue Tracking Info

- [X] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

Krychaz:
Hi there,

Thank you for your report! 

I was able to reproduce this issue, and I have flagged this to our engineering team.

If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time - our engineering team will be happy to review these.

Thanks once again!

closebot-bw:
⚠️ **Stale Issue Notice**

This issue has been automatically marked as stale due to inactivity. It will be closed in **2 weeks** (October 15, 2025) if no further activity occurs.

If this issue is still relevant and you would like to keep it open, please:
- Comment on this issue to show continued interest
- Provide any additional information or updates
- Confirm that the issue still exists in the latest version

Thank you for your contribution to this project! 🙏

windmueller:
This issue has been reproduced and I would still like to see it fixed.

PR #5082, opened 2024-11-25
[PM-15392] Decode Foreign Encodings at User Creation

BensonB12:
## 🎟️ Tracking

https://github.com/bitwarden/server/issues/4845#issue-2563235760)(https://github.com/bitwarden/server/issues/4845

## 📔 Objective

Solve the issue of coded foreign characters getting into emails. Was: 'Windm&#252;ller' Now: 'Windmüller'

Notes:
Our test are probably in the wrong file. We tried to find the best fitting file we could.

We tried to find the place where the string gets coded to '&#252;' and other foreign characters, but we ended up assuming that it was coming from the api call into the controller. Our solution does not work if that is not the case. 

Co-authors
- https://github.com/MJebran
- https://github.com/luris26

## 🦮 Reviewer guidelines


Patrick-Pimentel-Bitwarden:
👍

Patrick-Pimentel-Bitwarden (dismissed):
👍  Reviewed code, lgtm

trmartin4:
Hello @BensonB12,

First, thank you for your contribution, and for also including tests for your fix.

After reviewing this change, we feel like this is likely not the appropriate fix for the underlying problem.  We believe that the scope of https://github.com/bitwarden/server/issues/4845 may be larger than just emergency access emails (as you likely identified), but we need to audit the code to assess the proper place for making the encoding change so that it is updated at all points at which a user's email is changed, and also at the point that has the least possible other side effects in the application.

We are also in the process of migrating to a new registration flow, which - if registration is the appropriate place to adjust - would require adding code there as well.

We will be closing this PR and evaluating the fix for https://github.com/bitwarden/server/issues/4845 internally, but if you would like to continue iterating on a solution please feel free to re-open.

PR #5089, opened 2024-11-26
[PM-14376] Add GET tasks endpoint

gbubemismith:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Vault/Controllers/SecurityTaskController.cs
+++ b/src/Api/Vault/Controllers/SecurityTaskController.cs
@@ -1,0 +1,32 @@
+﻿using Bit.Api.Models.Response;
+using Bit.Api.Vault.Models.Response;
+using Bit.Core.Services;
+using Bit.Core.Vault.Enums;
+using Bit.Core.Vault.Queries;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Bit.Api.Vault.Controllers;
+
+[Route("tasks")]
+[Authorize("Application")]
+public class SecurityTaskController :  Controller
+{
+    private readonly IUserService _userService;
+    private readonly IGetTaskDetailsForUserQuery _getTaskDetailsForUserQuery;
+
+    public SecurityTaskController(IUserService userService, IGetTaskDetailsForUserQuery getTaskDetailsForUserQuery)
+    {
+        _userService = userService;
+        _getTaskDetailsForUserQuery = getTaskDetailsForUserQuery;
+    }
+
+    [HttpGet("")]
+    public async Task<ListResponseModel<SecurityTasksResponseModel>> Get([FromQuery] IEnumerable<SecurityTaskStatus> status = null)
+    {
+        var userId = _userService.GetProperUserId(User).Value;
+        var securityTasks = await _getTaskDetailsForUserQuery.GetTaskDetailsForUserAsync(userId, status);
+        var response = securityTasks.Select(x => new SecurityTasksResponseModel(x)).ToList();
+        return new ListResponseModel<SecurityTasksResponseModel>(response);
+    }
+}
--- a/src/Infrastructure.Dapper/Vault/Repositories/SecurityTaskRepository.cs
+++ b/src/Infrastructure.Dapper/Vault/Repositories/SecurityTaskRepository.cs
@@ -40,7 +40,7 @@ public class SecurityTaskRepository : Repository<SecurityTask, Guid>, ISecurityT
 
         var results = await connection.QueryAsync<SecurityTask>(
             $"[{Schema}].[SecurityTask_ReadByUserIdStatus]",
-            new {UserId = userId, Statuses = statusTable.AsTableValuedParameter("dbo.SecurityTaskStatusArray")},
+            new {UserId = userId, Status = statusTable.AsTableValuedParameter("dbo.SecurityTaskStatusArray")},
             commandType: CommandType.StoredProcedure);
 
         return results.ToList();
--- a/src/Infrastructure.Dapper/Vault/Repositories/SecurityTaskRepository.cs
+++ b/src/Infrastructure.Dapper/Vault/Repositories/SecurityTaskRepository.cs
@@ -1,5 +1,6 @@
 ﻿using Bit.Core.Settings;
 using Bit.Core.Vault.Entities;
+using Bit.Core.Vault.Enums;
 using Bit.Core.Vault.Repositories;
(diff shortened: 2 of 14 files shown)


src/Sql/dbo/User Defined Types/SecurityTasksStatusArray.sql
shane-melton, line 1: 🎨 We could probably make this more generic for `byte` enums so we can avoid creating unique array types for other `byte` enums we have in the future. We could rename this to `ByteEnumArray` or something similar.  What do you think?

src/Infrastructure.Dapper/Vault/Repositories/SecurityTaskRepository.cs
```diff
@@ -1,18 +1,48 @@
...
         : base(connectionString, readOnlyConnectionString)
-    { }
+    {
+    }
 
+    /// <inheritdoc />
+    public async Task<ICollection<SecurityTask>> GetManyByUserIdStatusAsync(Guid userId,
+        IEnumerable<SecurityTaskStatus> status = null)
+    {
+        var statusTable = new DataTable();
```
shane-melton, line 28: 🎨 If we go the route of making this a more generic `ByteEnumArray` we could add a new DapperHelper like `ToByteArrayTVP(this IEnumerable<byte> values)` to do this instead. [See here for more context.](https://github.com/bitwarden/server/blob/329eef82cdd68d1fa65bd883212bfcaaba7a325f/src/Infrastructure.Dapper/DapperHelpers.cs#L156-L176)

src/Sql/Vault/dbo/Stored Procedures/SecurityTask/SecurityTask_ReadByUserIdStatus.sql
```diff
@@ -0,0 +1,45 @@
...
+    WHERE
+        OU.[UserId] = @UserId
+        AND OU.[Status] = 2 -- Ensure user is confirmed
+        AND O.[Enabled] = 1
+        AND (
+            ST.[CipherId] IS NULL
+            OR (
+                C.[Id] IS NOT NULL
+                AND (
+                    CU.[Manage] = 1
```
shane-melton, line 35: ⚠️ I think I missed updating this ticket description, but we need to filter so that users with `Edit` access to the cipher see the tasks (including password visibility). That way any user can mark them as completed.

src/Sql/Vault/dbo/Stored Procedures/SecurityTask/SecurityTask_ReadByUserIdStatus.sql
```diff
@@ -0,0 +1,45 @@
...
+                    OR CG.[Manage] = 1
+                )
+            )
+        )
+        AND (
+            NOT EXISTS (SELECT 1 FROM @Status)
```
shane-melton, line 41: ❓ Could we simplify this SQL by having the query or repository method ensure the `Status` parameter always has default values if none are provided? I believe it would allow us to simply always filter by `ST.[Status] IN (SELECT [Value] FROM @Status)` and avoid the `NOT EXISTS (SELECT 1 FROM @Status)`.

src/Infrastructure.Dapper/Vault/Repositories/SecurityTaskRepository.cs
```diff
@@ -1,18 +1,48 @@
...
+        statusTable.Columns.Add("Vaule", typeof(byte));
```
shane-melton, line 29: ⛏️ Typo "Value"

src/Infrastructure.EntityFramework/Vault/Repositories/Queries/SecurityTaskReadByUserIdStatusQuery.cs
```diff
@@ -0,0 +1,84 @@
...
+                        ou.UserId == _userId &&
+                        ou.Status == OrganizationUserStatusType.Confirmed &&
+                        o.Enabled &&
+                        (
+                            st.CipherId == null ||
+                            (
+                                c != null &&
+                                cc != null &&
+                                (
+                                    (cu != null && cu.Manage) || (cg != null && cg.Manage)
```
shane-melton, line 63: ⚠️ Similar to the SQL query, we need to filter on `Can Edit` so that users can see/complete any cipher security task  they have password edit access to.

shane-melton (changes requested):

src/Sql/Vault/dbo/Stored Procedures/SecurityTask/SecurityTask_ReadByUserIdStatus.sql
gbubemismith: Won't we lose the ability to retrieve all tasks regardless of their status if we use this approach?

src/Sql/dbo/User Defined Types/SecurityTasksStatusArray.sql
gbubemismith: 👍🏼 nice name

src/Sql/Vault/dbo/Stored Procedures/SecurityTask/SecurityTask_ReadByUserIdStatus.sql
shane-melton: My thinking was we would have the repository method always provide a default value of every status type so that all tasks could be included. With that said, I don't feel too strongly about it either way, just thought it may make the SQL a little easier to follow.

src/Api/Vault/Controllers/SecurityTaskController.cs
```diff
@@ -0,0 +1,37 @@
...
+using Bit.Core.Vault.Enums;
+using Bit.Core.Vault.Queries;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Bit.Api.Vault.Controllers;
+
+[Route("tasks")]
+[Authorize("Application")]
+public class SecurityTaskController : Controller
```
shane-melton, line 13: 🎨 We should add a `[RequireFeature(FeatureFlagKeys.SecurityTasks)]` to the controller to hide it behind the feature flag.

shane-melton (changes requested):
Looks great! Just one suggestion to hide the controller behind the feature flag.

shane-melton (approved):

rkac-bw (approved):
lgtm, might have to look at query against security task view closer when data set gets large

PR #5122, opened 2024-12-05
[PM-15608] Create more KDF defaults for prelogin

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15608

## 📔 Objective

Add more KDF default options for prelogin endpoint.


kspearrin (commented):
This test still needs to be fixed, but I am not sure what the right way to go about this is. Any suggestions? https://github.com/bitwarden/server/pull/5122/files#diff-c1f1ab5b303cfba32ccd21519e9dfec8702038d2c4bc9371ab8efc2ace0ef3f5R102

src/Identity/Controllers/AccountsController.cs
```diff
@@ -69,6 +107,25 @@ IDataProtectorTokenFactory<RegistrationEmailVerificationTokenable> registrationE
...
+    {
+        // Add some random defaults as well
+        var random = new Random();
+        for (var i = 0; i < 2; i++)
+        {
+            _defaultKdfResults.Add(new()
+            {
+                Kdf = KdfType.PBKDF2_SHA256,
+                KdfIterations = random.Next(100_000, 1_000_000),
+            });
```
JaredSnider-Bitwarden, line 127: Does the randomness provide any additional protection? Doesn't it provide information if a given email corresponds to the same index position (of one of the random KDF iterations) across n requests. The caller would see the information change across n requests, right? If so, then they could extrapolate that the user doesn't exist. 

Per a discussion with @justindbaur , the random values would be different across each identity node as well which I believe would reduce the effectiveness of including random data. 

Further, if we are going to include random data, we probably should normalize it to 100k increments (as recommended in our [docs](https://bitwarden.com/help/kdf-algorithms/#pbkdf2)) so that the KDF iterations aren't something like "159,549" which it is extremely unlikely a normal person would pick.

JaredSnider-Bitwarden (commented):
Thank you for tackling this! We spent some time talking through the scenarios here, and we think there might be more discussion to be had regarding the random values.

src/Identity/Controllers/AccountsController.cs
```diff
@@ -217,11 +274,24 @@ public async Task<PreloginResponseModel> PostPrelogin([FromBody] PreloginRequest
...
+                var hmacMessage = Encoding.UTF8.GetBytes(model.Email.Trim().ToLowerInvariant());
+                using var hmac = new System.Security.Cryptography.HMACSHA256(_defaultKdfHmacKey);
+                var hmacHash = hmac.ComputeHash(hmacMessage);
+                // Convert the hash to a number
+                var hashHex = BitConverter.ToString(hmacHash).Replace("-", string.Empty).ToLowerInvariant();
+                var hashFirst8Bytes = hashHex.Substring(0, 16);
+                var hashNumber = long.Parse(hashFirst8Bytes, System.Globalization.NumberStyles.HexNumber);
+                // Find the default KDF value for this hash number
+                var hashIndex = (int)(Math.Abs(hashNumber) % _defaultKdfResults.Count);
+                kdfInformation = _defaultKdfResults[hashIndex];
```
justindbaur, line 293: I'd recommend using `XxHash32` for this instead of `HMACSHA256`, not only is is much faster:

```
| Method            | Mean        |
|------------------ |------------:|
| HMACSHA256        | 1,602.88 ns |
| XxHash32          |    39.02 ns |

Tested with the email: " John.smith@gmail.com"
```

but since we won't be truncating any of it, it's much more dispersion tested. 

Here is the code for it:

```C#
var hmacMessage = Encoding.UTF8.GetBytes(_cases[EmailCase].Trim().ToLowerInvariant());
Span<byte> hash = stackalloc byte[sizeof(uint)];
XxHash32.Hash(hmacMessage, hash, _seed /* this replaces `_defaultKdfHmaxKey` */);
// Convert the hash to a number
var hashNumber = BitConverter.ToUInt32(hash);
// Find the default KDF value for this hash number
var hashIndex = (int)(hashNumber % _defaultKdfResults.Count);
return _defaultKdfResults[hashIndex];
```

Since I use a `stackalloc` this would have to move into a non-async method and then just called from the controller action.

kspearrin: Nice. I have never used XxHash, but I am not sure the perf savings of HMAC vs XxHash are really substantial in the grand scheme of things here.

src/Identity/Controllers/AccountsController.cs
kspearrin: Yea, I didn't think about the various nodes in a load balanced setup. I think we should probably just revert the random values then.

kspearrin: Done

JaredSnider-Bitwarden (approved):
LGTM! Thank you for tackling this!

PR #5124, opened 2024-12-06
[PM-15547] Revoke managed user on 2FA removal if enforced by organization policy

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15547

## 📔 Objective

When the two factor organization policy is enabled, then user should be revoked when they remove their 2FA provider if the policy applies to them.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Enums/EventSystemUser.cs
```diff
@@ -6,4 +6,5 @@ public enum EventSystemUser : byte
     SCIM = 1,
     DomainVerification = 2,
     PublicApi = 3,
+    TwoFactorDisabling = 4,
```
r-tome, line 9: I'm not sure about this name, I welcome suggestions

JimmyVo16: Maybe `TwoFactorDisabled`? `Disabled` sounds more like a `state`, but I don't feel strongly about this. The name you choose isn't a blocker for me.

src/Core/Services/Implementations/UserService.cs
```diff
@@ -1355,13 +1361,27 @@ public void SetTwoFactorProvider(User user, TwoFactorProviderType type, bool set
...
         var twoFactorPolicies = await _policyService.GetPoliciesApplicableToUserAsync(user.Id, PolicyType.TwoFactorAuthentication);
+        var organizationsManagingUser = await GetOrganizationsManagingUserAsync(user.Id);
 
         var removeOrgUserTasks = twoFactorPolicies.Select(async p =>
         {
-            await _removeOrganizationUserCommand.RemoveUserAsync(p.OrganizationId, user.Id);
             var organization = await _organizationRepository.GetByIdAsync(p.OrganizationId);
-            await _mailService.SendOrganizationUserRemovedForPolicyTwoStepEmailAsync(
-                organization.DisplayName(), user.Email);
+            if (_featureService.IsEnabled(FeatureFlagKeys.AccountDeprovisioning) && organizationsManagingUser.Any(o => o.Id == p.OrganizationId))
```
JimmyVo16, line 1369: I was thinking of moving organizationsManagingUser into the if statement so it's only called if the feature check is on, but your way is cleaner, and if performance isn't an issue, I think it's the better option.

JimmyVo16 (dismissed):

src/Core/Services/Implementations/UserService.cs
r-tome: If the feature flag is disabled then `GetOrganizationsManagingUserAsync` simply returns an empty list.
ref: https://github.com/bitwarden/server/blob/170836aba1f87ee4a5c5a0b33d76dfa487ee94d1/src/Core/Services/Implementations/UserService.cs#L1295

src/Core/AdminConsole/Enums/EventSystemUser.cs
r-tome: Makes sense to me, I'll rename it

PR #5204, opened 2024-12-31
[PM-16603] Add userkey rotation v2

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16603
Client PR: https://github.com/bitwarden/clients/pull/12646

## 📔 Objective

Adds a new command for user-key rotation that includes the KDF settings too, and updates the password. Full description of why this change is made is on the client PR.

In short, adds a new endpoint `rotate-user-account-keys` which performs userkey rotation, while updating the masterpassword, and kdf parameters, to ensure a consistent update.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -377,6 +380,9 @@ public async Task PostKdf([FromBody] KdfRequestModel model)
         throw new BadRequestException(ModelState);
     }
 
+    /**
+     * @deprecated
```
JaredSnider-Bitwarden, line 384: ⚠️  We should use `[Obsolete("<message explaining deprecation + pointing to new endpoint>")]` to mark endpoints as deprecated like we do elsewhere.

src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs
JaredSnider-Bitwarden, line 1: 🎨 : I would highly recommend writing a test suite for this important command similar to the `RotateUserKeyCommandTests`

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -220,30 +220,13 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
         try
         {
             // Update user
-            await using (var cmd = new SqlCommand("[dbo].[User_UpdateKeys]", connection, transaction))
```
JaredSnider-Bitwarden, line 223: ❓ : This looks like a good change to me, but there is a small bit of functionality difference where `User_UpdateKeys` used to make `AccountRevisionDate` default to `RevisionDate` if null. 

If that's not an issue, can you investigate what our proper stored proc deprecation processes are as I don't believe `User_UpdateKeys` is used anywhere else now?

JaredSnider-Bitwarden (changes requested):
This is an excellent change. Thank you for drastically improving our key rotation processes. I have just a few comments / questions below:

src/Infrastructure.Dapper/Repositories/UserRepository.cs
quexten: Interesting, on the entityframework repository this does not have the same fallback behavior.

(Also, my changes just ignored the entity framework, so luckily caught that here).

src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs
quexten: Added tests! Thanks for reminding me!

JaredSnider-Bitwarden (dismissed):
Nice work! LGTM!

quexten:
Forgot to add the feature flag server-side; added it now.

util/Migrator/DbScripts_finalization/2025-01-21_UserkeyRotationV2.sql
```diff
@@ -0,0 +1,7 @@
+--- Remove old updatekeys procedure that is now unused ---
+
+IF OBJECT_ID('[dbo].[User_UpdateKeys]') IS NOT NULL
```
mzieniukbw, line 3: Do you wlaos want to delete this SP from repo ? Location: https://github.com/bitwarden/server/blob/7e74695afc4257a63b8dd280359277cd885cfc48/src/Sql/dbo/Stored%20Procedures/User_UpdateKeys.sql#L1

mzieniukbw (changes requested):

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -219,31 +219,19 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
         await using var transaction = connection.BeginTransaction();
         try
         {
-            // Update user
-            await using (var cmd = new SqlCommand("[dbo].[User_UpdateKeys]", connection, transaction))
+            if (user.AccountRevisionDate == null)
```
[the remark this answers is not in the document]
quexten: :| I wonder why the SP had handling for this then:

https://github.com/bitwarden/server/blob/7e74695afc4257a63b8dd280359277cd885cfc48/src/Sql/dbo/Stored%20Procedures/User_UpdateKeys.sql#L20

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -219,31 +219,19 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
...
-                cmd.Parameters.Add("@AccountRevisionDate", SqlDbType.DateTime2).Value =
-                    user.AccountRevisionDate;
-                cmd.Parameters.Add("@LastKeyRotationDate", SqlDbType.DateTime2).Value =
-                    user.LastKeyRotationDate;
-                cmd.ExecuteNonQuery();
+                user.AccountRevisionDate = user.RevisionDate;
             }
 
+            // Update user
+            await ReplaceAsync(user);
```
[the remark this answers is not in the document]
quexten: Replaced with:
```
            ProtectData(user);
            await connection.ExecuteAsync(
                $"[{Schema}].[{Table}_Update]",
                user,
                transaction: transaction,

``` now

File: src/Api/Auth/Controllers/AccountsController.cs [hunk 6]
```diff
@@ -415,6 +419,54 @@ public async Task PostKey([FromBody] UpdateKeyRequestModel model)
         throw new BadRequestException(ModelState);
     }
 
+    [HttpPost("rotate-user-account-keys")]
+    public async Task PostUserAccountKeys([FromBody] RotateUserAccountKeysModel model)
```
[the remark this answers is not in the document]
quexten: Added unit tests and integration tests

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -194,6 +194,51 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(Core.Entities.User user,
...
+            }
+
+            await transaction.CommitAsync();
+        }
+        catch
+        {
+            await transaction.RollbackAsync();
+            throw;
+        }
+
```
Thomas-Avery, line 239:
```suggestion
            // Update user
            var userEntity = await dbContext.Users.FindAsync(user.Id);
            if (userEntity == null)
            {
                throw new ArgumentException("User not found", nameof(user));
            }

            userEntity.SecurityStamp = user.SecurityStamp;
            userEntity.Key = user.Key;
            userEntity.MasterPassword = user.MasterPassword;
            userEntity.PrivateKey = user.PrivateKey;
            userEntity.LastKeyRotationDate = user.LastKeyRotationDate;
            userEntity.AccountRevisionDate = user.AccountRevisionDate;
            userEntity.RevisionDate = user.RevisionDate;

            await dbContext.SaveChangesAsync();

            //  Update re-encrypted data
            foreach (var action in updateDataActions)
            {
                // connection and transaction aren't used in EF
                await action();
            }

            await transaction.CommitAsync();
        
```
:pick: Optional 

With the default behavior, you get rollbacks without having to do the try catch https://learn.microsoft.com/en-us/ef/core/saving/transactions#default-transaction-behavior

test/Api.Test/KeyManagement/Controllers/AccountsKeyManagementControllerTests.cs
```diff
@@ -93,4 +105,46 @@ await sutProvider.GetDependency<IRegenerateUserAsymmetricKeysCommand>().Received
                 Arg.Is(orgUsers),
                 Arg.Is(accessDetails));
     }
+
```
Thomas-Avery, line 108: Would expect unit tests for failure cases. 

`_userService.GetUserByPrincipalAsync` returning null and throwing `UnauthorizedAccessException`.

`   var result = await _rotateUserAccountKeysCommand.RotateUserAccountKeysAsync(user, dataModel);` returning `result.Errors` and throwing `BadRequestException`

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -254,6 +254,48 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
...
+        try
+        {
+            if (user.AccountRevisionDate == null)
+            {
+                user.AccountRevisionDate = user.RevisionDate;
+            }
+
+            ProtectData(user);
+            await connection.ExecuteAsync(
+                $"[{Schema}].[{Table}_Update]",
```
Thomas-Avery, line 274: Seems odd to be doing a complete replacement of the user entity when this is called `UpdateUserKeyAndEncryptedDataV2Async`.  If anything changes on that user between the controller fetch and calling this repository method, we will overwrite it.

Can we do the following or is there something I'm missing?: 
```
            if (user.AccountRevisionDate == null)
            {
                user.AccountRevisionDate = user.RevisionDate;
            }
            
            // Update user values using data protection
            ProtectData(user);
            
              // Update user columns needed for key rotation
                   await connection.ExecuteAsync(
               $"[{Schema}].[User_UpdateKeys]",
               new
               {
                   user.Id,
                   user.SecurityStamp,
                   user.Key,
                   user.PrivateKey,
                   user.RevisionDate,
                   user.AccountRevisionDate,
                   user.LastKeyRotationDate
               },

            //  Update re-encrypted data
            foreach (var action in updateDataActions)
            {
                await action(connection, transaction);
            }      
```

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -254,6 +254,48 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
...
+        catch
+        {
+            transaction.Rollback();
+            throw;
+        }
+        finally
+        {
+            connection.Close();
+            UnprotectData(user);
+        }
```
Thomas-Avery, line 295:
```suggestion
```

We don't need to `UnprotectData` the user entity because this method doesn't return anything.

Thomas-Avery (changes requested):
Looking good, a few minor things to take a look at. 

An overall suggestion for net new code, can we use `#nullable enable` at the top of the files to enable nullable reference types? This is currently a mix of using that feature in some places and not in others, which I find confusing.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
quexten: The idea for this was that it is unclear if the caller expects the reference they pass in to be the same after calling. I guess since we are the only caller in rotateuseraccountkeyscommand, this should be safe. But it would be unexpected from a caller's point of view for the passed in user object to be mutated.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
quexten: The above is missing at least the masterpassword hash and masterpassword hint. 

As it is implemented right now, we do our  sanity/safety checks regarding the: public key, email, kdf settings, before calling `UpdateUserKeyAndEncryptedDataV2Async`, with one user object that has been validated.

If we just update the above, it could be the case that f.e the kdf setting changes after the sanity check, but before the above update, leading to a key that does not match the kdf settings being stored. 

So we would either need to include:
public key, email, kdf settings, in the update too, replacing them?

src/Infrastructure.Dapper/Repositories/UserRepository.cs
Thomas-Avery: In C# arguments are passed by value by default. You need to use the `ref` modifier to turn things into pass by reference.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters

Typically, for a repository layer, you have `Task` to indicate you will be saving the object as provided, or you return a value to indicate changes.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
Thomas-Avery: Looks like we're missing updating those fields in the EF code too then? 

I'm confused here.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
Thomas-Avery: This can be disregarded, followed up in slack.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
quexten: Hmm, that means that the EF code is wrong. I think that means the EF code needs better test coverage in this case because this could have slipped through.

Thomas-Avery: I know we recently published some new contributing docs info on EF integration tests. https://contributing.bitwarden.com/getting-started/server/database/ef/#testing-ef-changes

Might be something to look into here.

test/Api.Test/KeyManagement/Controllers/AccountsKeyManagementControllerTests.cs
quexten: Added!

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -254,6 +254,46 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
...
+            }
+            transaction.Commit();
+        }
+        catch
+        {
+            connection.Close();
+            UnprotectData(user);
+            throw;
+        }
+        connection.Close();
```
Thomas-Avery, line 292: Do we need the `connection.Close()` here?

I thought the using statements handle disposing and closing those things in dapper?

test/Core.Test/KeyManagement/UserKey/RotateUserAccountKeysCommandTests.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        user.Email = model.MasterPasswordUnlockData.Email;
+        sutProvider.GetDependency<IUserService>().CheckPasswordAsync(user, model.OldMasterKeyAuthenticationHash)
+            .Returns(false);
+
+        var result = await sutProvider.Sut.RotateUserAccountKeysAsync(user, model);
+
+        Assert.NotEqual(IdentityResult.Success, result);
+    }
+
```
Thomas-Avery, line 28:
```suggestion
  [Theory, BitAutoData]
    public async Task ThrowsWhenUserIsNull(SutProvider<RotateUserAccountKeysCommand> sutProvider,
        RotateUserAccountKeysData model)
    {
        await Assert.ThrowsAsync<ArgumentNullException>(async () => await sutProvider.Sut.RotateUserAccountKeysAsync(null, model));
    }
```

Thomas-Avery (changes requested):

src/Api/KeyManagement/Models/Requests/UserDataRequestModel.cs
```diff
@@ -0,0 +1,12 @@
...
+using Bit.Api.Tools.Models.Request;
+using Bit.Api.Vault.Models.Request;
+
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+public class AccountDataRequestModel
+{
+    public required IEnumerable<CipherWithIdRequestModel> Ciphers { get; set; }
+    public required IEnumerable<FolderWithIdRequestModel> Folders { get; set; }
+    public required IEnumerable<SendWithIdRequestModel> Sends { get; set; }
```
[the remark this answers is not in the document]
quexten: That is my expectation, yes.

src/Api/Auth/Models/Request/Accounts/MasterPasswordUnlockDataModel.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    public required KdfType KdfType { get; set; }
+    public required int KdfIterations { get; set; }
+    public int? KdfMemory { get; set; }
+    public int? KdfParallelism { get; set; }
+
+    public required string Email { get; set; }
+    [StringLength(300)]
+    public required string MasterKeyAuthenticationHash { get; set; }
+    public required string MasterKeyEncryptedUserKey { get; set; }
```
Thomas-Avery, line 19:
```suggestion
    [EncryptedString] public required string MasterKeyEncryptedUserKey { get; set; }
```
Should some of these have the `[EncryptedString]` validator?

src/Api/KeyManagement/Models/Requests/AccountKeysRequestModel.cs
```diff
@@ -0,0 +1,8 @@
+﻿#nullable enable
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+public class AccountKeysRequestModel
+{
+    public required string UserKeyEncryptedAccountPrivateKey { get; set; }
```
Thomas-Avery, line 6:
```suggestion
    [EncryptedString] public required string UserKeyEncryptedAccountPrivateKey { get; set; }
```

src/Api/Auth/Models/Request/Accounts/MasterPasswordUnlockDataModel.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Bit.Api.Auth.Models.Request.Accounts;
+
+public class MasterPasswordUnlockDataModel : IValidatableObject
+{
+    public required KdfType KdfType { get; set; }
+    public required int KdfIterations { get; set; }
+    public int? KdfMemory { get; set; }
+    public int? KdfParallelism { get; set; }
+
+    public required string Email { get; set; }
```
Thomas-Avery, line 16:
```suggestion
    [StrictEmailAddress]
    [StringLength(256)]
    public required string Email { get; set; }
```

src/Api/KeyManagement/Models/Requests/RotateAccountKeysAndDataRequestModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿#nullable enable
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+public class RotateUserAccountKeysAndDataRequestModel
+{
+    public required string OldMasterKeyAuthenticationHash { get; set; }
```
Thomas-Avery, line 6:
```suggestion
    [StringLength(300)]
    public required string OldMasterKeyAuthenticationHash { get; set; }
```

Thomas-Avery (changes requested):
Looks good! Sorry, I missed checking these data validation attributes earlier.
Can we double-check these on the request models? I made some suggestions, looking at other request models.

After this I'm good to approve.

quexten:
I believe I addressed all of the validation attributes (had to adjust some tests).

Thomas-Avery (dismissed):
LGTM, thank you!

JaredSnider-Bitwarden (dismissed):

src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs
```diff
@@ -0,0 +1,145 @@
...
+    public async Task<IdentityResult> rotateUserAccountKeysAsync(User user, RotateUserAccountKeysData model)
+    {
+        if (user == null)
+        {
+            throw new ArgumentNullException(nameof(user));
+        }
+
+        if (!await _userService.CheckPasswordAsync(user, model.OldMasterPasswordHash))
+        {
+            return IdentityResult.Failed(_identityErrorDescriber.PasswordMismatch());
```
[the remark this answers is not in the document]
mzieniukbw: Exactly, simplify, by throwing error instead. On the client side, we do not do anything special, just display the error message, regardless of what it is.

src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        if (!await _userService.CheckPasswordAsync(user, model.OldMasterKeyAuthenticationHash))
+        {
+            return IdentityResult.Failed(_identityErrorDescriber.PasswordMismatch());
+        }
+
+        var now = DateTime.UtcNow;
+        user.RevisionDate = user.AccountRevisionDate = now;
+        user.LastKeyRotationDate = now;
```
mzieniukbw, line 76: 🎨 I don't know which one is better, but feels weird right now:
```
        user.RevisionDate = now;
        user.AccountRevisionDate = now;
        user.LastKeyRotationDate = now;
```
or
```
user.RevisionDate = user.AccountRevisionDate = user.LastKeyRotationDate = now;
```

src/Infrastructure.Dapper/Repositories/UserRepository.cs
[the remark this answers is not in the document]
mzieniukbw: Maybe some legacy reasons, probably was null, so we expect null ? Let's leave it.

mzieniukbw (approved):

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -254,6 +254,41 @@ public async Task UpdateUserKeyAndEncryptedDataAsync(
...
+            foreach (var action in updateDataActions)
+            {
+                await action(connection, transaction);
+            }
+            transaction.Commit();
+        }
+        catch
+        {
+            UnprotectData(user);
+            throw;
```
mzieniukbw, line 286: don't we have to explicitly rollback transaction before throwing ?

mzieniukbw (changes requested):

src/Infrastructure.Dapper/Repositories/UserRepository.cs
quexten: Added the rollback! Thank you for catching it.

mzieniukbw (approved):

Thomas-Avery (approved):

PR #5229, opened 2025-01-07
[pm-13985] Update cancel button to avoid post back

JimmyVo16:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-13985

## 📔 Objective

**Problem**: It's hitting `POST /Providers/Edit/{id}` whenever the user clicks `Cancel`. This endpoint requires the `Provider_Edit` permission, causing a permission error for some user roles.

**Fix**: Create a cancel endpoint that doesn't require `Provider_Edit` permission  to properly redirect the user to the intended page 

## 📸 Screenshots

https://github.com/user-attachments/assets/bd9d6713-599e-4322-ab18-b7cae29583a4

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Admin/AdminConsole/Views/Providers/CreateOrganization.cshtml
```diff
@@ -19,9 +19,9 @@
 <div class="d-flex mt-4">
     <button type="submit" class="btn btn-primary" form="edit-form">Save</button>
     <div class="ms-auto d-flex">
-        <form asp-controller="Providers" asp-action="Edit" asp-route-id="@Model.Provider.Id"
```
jrmccannon, line 22: By using `history.back()`, you're removing the this page from the history stack. 

This form inside a form is posting to `Edit(Guid id, ProviderEditModel model)` instead of the intended `Edit(Guid id)`.  Maybe we could just add a `Cancel` post endpoint that takes the id and redirects to `Edit(Guid id)` instead?

jrmccannon (changes requested):

src/Admin/AdminConsole/Views/Providers/CreateOrganization.cshtml
JimmyVo16: I'm happy to implement your recommendation using the Cancel post endpoint.

I do have a question, though: could you elaborate on some use cases where removing a page from the history stack might cause issues? From my understanding, `history.back()` behaves like clicking the browser's back button, so I want to ensure I'm considering all potential pitfalls.

jrmccannon: I don't know if there are explicit rules against it, but normally, when a cancel button is hit I'd expect just a normal redirect.

Also, the user could theoretically route directly to the form from any url. It would be strange if they clicked cancel and were taken to the last url they were at, potentially out the admin portal. A redirect back to the Edit covers and weird edge cases by explicitly saying "you go to here" when the cancel button is hit.

Does that make sense?

I think I need more sleep.

JimmyVo16: Yeah, that sounds reasonable. I was avoiding a postback, but I doubt performance is an issue here, so consistency is more important.

jrmccannon (approved):

PR #5253, opened 2025-01-13
[PM-16982] Add argon2-default flag

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16982

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


mzieniukbw (approved):
I noticed that minimum memory is different in client and server. In client's 15, on server it's 16. Could we make it consistent ? (although i have not considered any migration steps, maybe confirm with DBOPS if we have anyone on 15?) https://github.com/bitwarden/server/blob/5423e5d52fe2aac28db457107579aa91709c9bb7/src/Core/Constants.cs#L58

quexten:
@mzieniukbw I think the range is only used for settings validation, so in this case it might not even matter if users are on lower settings than the range, it just would block new changes that fall outside of the range. The PBKDF2 range starts at 600000 iterations but we have many users below that.

Follow-up PR though.

PR #5261, opened 2025-01-14
[pm-16949] Include revoked users in applicable policies

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16949

## 📔 Objective

**Cause**: Invoked users were not included in the applicable policies, causing the 2FA policy not to be enforced when restoring a revoked user. 

**Fix**: update the calling code to pass in `OrganizationUserStatusType.Revoked)`, which has a value of `-1` for the minimum status, instead of  `OrganizationUserStatusType.Invited`, which has a value of `0`

## 📸 Screenshots

https://github.com/user-attachments/assets/d29b2cf9-51a2-474a-abd6-6b53658ac256

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
+++ b/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
@@ -2247,7 +2247,7 @@ public class OrganizationService : IOrganizationService
         if (!userHasTwoFactorEnabled)
         {
             var invitedTwoFactorPolicies = await _policyService.GetPoliciesApplicableToUserAsync(userId,
-                PolicyType.TwoFactorAuthentication, OrganizationUserStatusType.Invited);
+                PolicyType.TwoFactorAuthentication, OrganizationUserStatusType.Revoked);
             if (invitedTwoFactorPolicies.Any(p => p.OrganizationId == orgUser.OrganizationId))
             {
                 twoFactorCompliant = false;


src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -2247,7 +2247,7 @@ private async Task CheckPoliciesBeforeRestoreAsync(OrganizationUser orgUser, boo
         if (!userHasTwoFactorEnabled)
         {
             var invitedTwoFactorPolicies = await _policyService.GetPoliciesApplicableToUserAsync(userId,
-                PolicyType.TwoFactorAuthentication, OrganizationUserStatusType.Invited);
+                PolicyType.TwoFactorAuthentication, OrganizationUserStatusType.Revoked);
```
BTreston, line 2250: I think this should be an additional check, not replacing the check for Invited user. I am curious to see how this might affect the policy enforcement for invited users. I'm going to add a server dev to the PR for a second opinion though

JimmyVo16: This does account for Invited users as well. 

The code leads to this logic [o.OrganizationUserStatus >= minStatus](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L74). Currently, `minStatus` is `Invited` so it accounts for users with the value of greater than or equal to `0`. By passing in `Revoked` we will account for users with `-1` 


OrganizationUserStatusType values.
image

Let me know if this addresses your concerns.

r-tome: If the user was in `Invited` status before being `Revoked` then there's no way they'd have 2FA configured. We could do one of two things:
- allow the user to be restored (then when they try to accept the invite an error will pop up saying they have to set up 2FA first)
- throw an error when trying to revoke/restore invited users

I'd personally go for the first option but please check with the product team.

r-tome: Sorry, I was thinking about the scenario 1 described in the ticket and that is working as expected

PR #5276, opened 2025-01-15
[PM-16684]

amorask-bitwarden:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Billing/Pricing/PricingClient.cs
```diff
@@ -1,69 +1,111 @@
...
     IFeatureService featureService,
-    GlobalSettings globalSettings) : IPricingClient
+    GlobalSettings globalSettings,
+    HttpClient httpClient,
+    ILogger<PricingClient> logger) : IPricingClient
 {
-    public async Task<Plan?> GetPlan(PlanType planType)
+    public async Task<Plan> GetPlan(PlanType planType)
     {
+        if (globalSettings.SelfHosted)
```
withinfocus, line 24: ❓ Is this possible to even trigger?

src/Core/Billing/Pricing/PricingClient.cs
```diff
@@ -1,69 +1,111 @@
...
-using Grpc.Net.Client;
-using Proto.Billing.Pricing;
+using Microsoft.Extensions.Logging;
+using Plan = Bit.Core.Models.StaticStore.Plan;
 
 #nullable enable
 
 namespace Bit.Core.Billing.Pricing;
 
 public class PricingClient(
```
withinfocus, line 16: 🎨 I see a number of checks for self-hosted out at the edge code, but it ends up calling the pricing client. If possible have those checks all centralized here, for when your return case needs to differ.

src/Core/Billing/Pricing/PricingClient.cs
amorask-bitwarden: I don't see why not unless I'm missing something. It's constructed by the container and injected like it is in the other services using it I believe.

src/Core/Billing/Pricing/PricingClient.cs
amorask-bitwarden: I've renamed the existing `GetPlan` method in the `PricingClient` to be `GetPlanOrThrow` and added a new `GetPlan` that returns a nullable in the case of a Self-Host instance or the plan wasn't found. This eliminates the upstream Self-Host checks.

withinfocus:
The tweaks make sense to me -- on to true team review though.

src/Core/Billing/Pricing/PricingClient.cs
```diff
@@ -1,69 +1,123 @@
...
+        {
+            var plans = await DeserializeAsync<List<PlanDTO>>(response.Content);
+            return plans.Select(Plan (plan) => new PlanAdapter(plan)).ToList();
+        }
+
+        throw new BillingException(
+            message: $"Request to the Pricing Service failed with status {response.StatusCode}");
+    }
+
+    private static async Task<T> DeserializeAsync<T>(HttpContent content)
```
r-tome, line 100: This method could be replaced by `ReadFromJsonAsync` in `System.Net.Http.Json`

src/Core/Billing/Pricing/PricingClient.cs
```diff
@@ -1,69 +1,123 @@
...
-        try
-        {
-            var response =
-                await client.GetPlanByLookupKeyAsync(new GetPlanByLookupKeyRequest { LookupKey = lookupKey });
+        var response = await httpClient.GetAsync($"plans/lookup/{lookupKey}");
 
-            return new PlanAdapter(response);
+        if (response.IsSuccessStatusCode)
+        {
+            var plan = await DeserializeAsync<PlanDTO>(response.Content);
```
r-tome, line 48:
```suggestion
            var plan = await response.Content.ReadFromJsonAsync<PlanDTO>();
```

test/Core.Test/Models/Business/SecretsManagerSubscriptionUpdateTests.cs
```diff
@@ -11,48 +13,57 @@ namespace Bit.Core.Test.Models.Business;
 [SecretsManagerOrganizationCustomize]
 public class SecretsManagerSubscriptionUpdateTests
 {
+    private static TheoryData<Plan> ToPlanTheory(List<PlanType> types)
```
r-tome, line 16: 👌 nice!

src/Api/Billing/Controllers/ProviderBillingController.cs
```diff
@@ -84,13 +86,25 @@ public async Task<IResult> GetSubscriptionAsync([FromRoute] Guid providerId)
 
         var providerPlans = await providerPlanRepository.GetByProviderId(provider.Id);
 
+        var configuredProviderPlans = await Task.WhenAll(providerPlans.Select(async providerPlan =>
```
r-tome, line 89: Could we benefit in performance from a `GetManyPlans` method?

r-tome (commented):
Great work! Just left a suggestion

src/Core/Billing/Pricing/PricingClient.cs
amorask-bitwarden: Great call! Replaced [here](https://github.com/bitwarden/server/pull/5276/commits/8d60635e0bb528d0a75dff8414868ce356a04c97).

src/Core/Billing/Pricing/PricingClient.cs
amorask-bitwarden: Addressed in https://github.com/bitwarden/server/pull/5276/commits/8d60635e0bb528d0a75dff8414868ce356a04c97

src/Api/Billing/Controllers/ProviderBillingController.cs
amorask-bitwarden: Ultimately, we may need to implement a plan "filter" or "search" endpoint in the Pricing Service, but since it didn't exist in `server` to begin with, I'm trying to keep the "logical" changes to a minimum. 

In this case here, the number of provider plans will be, at max, 2, so I'm not concerned about the performance hit for this. The other option currently would be listing all plans into memory from the client and then filtering on them here. I think this is preferable for the time being.

r-tome: Sounds reasonable, thanks!

r-tome (dismissed):
Great work! Thanks for implementing my suggestion

r-tome (approved):

jonashendrickx (dismissed):

cturnbull-bitwarden (dismissed):
Excellent writeup, thank you sir

cturnbull-bitwarden (approved):

r-tome (approved):

PR #5323, opened 2025-01-24
[PM-17562] Initial POC of Distributed Events

brant-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17562

## 📔 Objective

This PR is an initial POC of a new way of handling events. The end goal will be to add an AMQP messaging for all platforms that will broadcast and fan-out events in addition to our normal handling of events. This allows multiple integrations to be built in the future (e.g a Slack message when X event occurs, etc). The Cloud version will ultimately be built on Azure Service Bus, but for this initial POC we wanted to just build something for local  / self-hosted.

This PR adds a new opt-in flow for Events. Specifically:
- A Docker Compose profile that will start a RabbitMQ container
- An `IEventWriteService` implementation that writes events to the RabbitMQ exchange
- A RabbitMQ subscriber that monitors for new events and writes those events to the DB (replicating what we had before, but introducing the RabbitMQ in the middle)
- A RabbitMQ subscriber that will post JSON to a configurable URL
  - This is included as a small proof of concept for the fan-out. It's a way of showing that we can add new subscribers that provide new integrations once this is in place.
- New configurations that allow for specifying all of this.
  - Everything here should be opt-in only. If you don't configure the RabbitMQ service, everything should operate as before.
  - Same with the POST service - even if RabbitMQ is configured, it will only start up if there's a URL configured.

The following is copied from [my PR in the contributing-docs repo](https://github.com/bitwarden/contributing-docs/pull/533). This is added to the Events page to document the new RabbitMQ additions and how to configure and start the service.

## Distributed Events (optional)

There is a new system which will add support for distributing events via an AMQP nessaging system.
In the future this will enable new integrations by allowing for a means to subscribe to
events via messaging stream.

As an initial proof of concept, there is an optional RabbitMQ implementation that refactors the way
events are handled when running locally or self-hosted. Instead of writing directly to the `Events` table
via the `EventsRepository`, it will broadcast each event via a RabbitMQ exchange. A new
`RabbitMqEventRepositoryListener` then subscribes to the RabbitMQ exchange and writes to the
`Events` table via the `EventsRepository`. The end result is the same (events are stored in the
database), but this allows for other integrations to subscribe.

To illustrate this, there is also a `RabbitMqEventHttpPostListener` which subscribes to the RabbitMQ
events exchange and `POST`s each event to a configurable URL. This is meant to be
a simple, concrete example of how multiple integrations are enabled by moving to
distributed events.

``` mermaid
graph TD
	  subgraph With RabbitMQ
        B1[EventService]
        B2[RabbitMQEventWriteService]
        B3[RabbitMQ exchange]
        B4[RabbitMqEventRepositoryListener]
        B5[RabbitMqEventHttpPostListener]
        B6[Events Database Table]
        B7[HTTP Server]

        B1 -->|IEventWriteService| B2 --> B3
        B3--> B4 --> B6
        B3--> B5
        B5 -->|HTTP POST| B7
    end

    subgraph Without RabbitMQ
        A1[EventService]
        A2[RepositoryEventWriteService]
        A3[Events Database Table]

        A1 -->|IEventWriteService| A2 --> A3

end
```

To enable the new RabbitMQ-based event stream, take the following steps:

### Running the RabbitMQ container

1.  Verify that you've set a RabbitMQ username and password in the `.env` file (see `.env.example`
    for an example)

2.  Use Docker Compose to run the container with your current settings:

    ```bash
    docker compose --profile rabbitmq up -d
    ```

3.  This will run the RabbitMQ container with your username and password on localhost with the standard ports

4. To verify this is running, open `http://localhost:15672` in a browser and login with the username and password in your `.env` file.

### Configuring the server to use RabbitMQ for events

1.  Add the following to your `secrets.json` file, changing the defaults to match your `.env` file:

    ``` json
        "eventLogging": {
          "rabbitMq": {
            "hostName": "localhost",
            "username": "bitwarden",
            "password": "SET_A_PASSWORD_HERE_123",
            "exchangeName": "events-exchange",
            "eventRepositoryQueueName": "events-write-queue",
          }
        },
    ```

2. (optional) To use the `RabbitMqEventHttpPostListener`, specify a queue name and destination
    URL that will receive the POST inside of the RabbitMq object:

    ``` json
         "httpPostQueueName": "events-httpPost-queue",
         "httpPostUrl": "<HTTP POST URL>",
    ```
    - Tip: [RequetBin](http://requestbin.com/) provides an easy to setup server that will
    receive these requests and let you inspect them.

3. Re-run the powershell script to add these secrets to each Bitwarden project:

    ```bash
    pwsh setup_secrets.ps1
    ```

3.  Start (or restart) all of your projects to pick up the new settings

With these changes in place, you should see the database events written as before, but you'll also see in the RabbitMQ management interface that the messages are flowing through the configured exchange.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        _httpPostUrl = globalSettings.RabbitMqHttpPostUrl;
+    }
+    protected override async Task HandleMessageAsync(EventMessage eventMessage)
+    {
+        var message = JsonSerializer.Serialize(eventMessage);
+        using var httpClient = new HttpClient();
+        var content = new StringContent(message, Encoding.UTF8, "application/json");
+        var response = await httpClient.PostAsync(_httpPostUrl, content);
+        response.EnsureSuccessStatusCode();
```
justindbaur, line 28: We shouldn't be creating short lived `HttpClient`'s. We should instead created a named `HttpClient` in DI and retrieve it with `IHttpClientFactory` that gets injected here. 

https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use

src/Core/AdminConsole/Services/Implementations/RabbitMqEventListenerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+        consumer.ReceivedAsync += async (model, eventArgs) =>
+        {
+            var body = eventArgs.Body.ToArray();
+            var message = Encoding.UTF8.GetString(body);
+
+            try
+            {
+                var eventMessage = JsonSerializer.Deserialize<EventMessage>(message);
+                await HandleMessageAsync(eventMessage);
+            }
```
justindbaur, line 64: JsonSerializer is going to convert the string back into bytes anyways, no need to allocate an intermediary string just for it to be thrown away.

```suggestion
            try
            {
                var eventMessage = JsonSerializer.Deserialize<EventMessage>(eventArgs.Body.Span);
                await HandleMessageAsync(eventMessage);
            }
```

src/Core/AdminConsole/Services/Implementations/RabbitMqEventWriteService.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    public async Task CreateAsync(IEvent e)
+    {
+        var connection = await _lazyConnection.Value;
+        using var channel = await connection.CreateChannelAsync();
+
+        await channel.ExchangeDeclareAsync(exchange: _exchangeName, type: ExchangeType.Fanout);
+
+        var message = JsonSerializer.Serialize(e);
+        var body = Encoding.UTF8.GetBytes(message);
```
justindbaur, line 35: Same idea here but the other way around.

```suggestion
        var body = JsonSerializer.SerializeToUtf8Bytes(e);
```

src/Core/AdminConsole/Services/Implementations/RabbitMqEventWriteService.cs
```diff
@@ -0,0 +1,59 @@
...
+    public async Task CreateManyAsync(IEnumerable<IEvent> events)
+    {
+        var connection = await _lazyConnection.Value;
+        using var channel = await connection.CreateChannelAsync();
+        await channel.ExchangeDeclareAsync(exchange: _exchangeName, type: ExchangeType.Fanout);
+
+        foreach (var e in events)
+        {
+            var message = JsonSerializer.Serialize(e);
+            var body = Encoding.UTF8.GetBytes(message);
```
justindbaur, line 49: Here too.

```suggestion
            var body = JsonSerializer.SerializeToUtf8Bytes(e);
```

src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
```diff
@@ -0,0 +1,30 @@
...
+        : base(logger, globalSettings)
+    {
+        _httpPostUrl = globalSettings.RabbitMqHttpPostUrl;
+    }
+    protected override async Task HandleMessageAsync(EventMessage eventMessage)
+    {
+        var message = JsonSerializer.Serialize(eventMessage);
+        using var httpClient = new HttpClient();
+        var content = new StringContent(message, Encoding.UTF8, "application/json");
+        var response = await httpClient.PostAsync(_httpPostUrl, content);
```
justindbaur, line 27: There is a special `HttpContent` class for JSON already.

```suggestion
        using var httpClient = new HttpClient();
        var content = JsonContent.Create(eventMessage);
        var response = await httpClient.PostAsync(_httpPostUrl, content);
```

justindbaur (commented):
Happened across this PR so maybe this feedback is a little early, if so sorry about that.

dev/.env.example
```diff
@@ -20,4 +20,8 @@ IDP_SP_ACS_URL=http://localhost:51822/saml2/yourOrgIdHere/Acs
 # Optional reverse proxy configuration
 # Should match server listen ports in reverse-proxy.conf
 API_PROXY_PORT=4100
-IDENTITY_PROXY_PORT=33756
\ No newline at end of file
+IDENTITY_PROXY_PORT=33756
+
+# Optional RabbitMQ configuration
+RABBITMQ_DEFAULT_USER=bitwarden
+RABBITMQ_DEFAULT_PASS=SET_A_PASSWORD_HERE_123
```
withinfocus, line 27: ⛏️ Always end files with newlines.

dev/docker-compose.yml
```diff
@@ -84,6 +84,20 @@ services:
     profiles:
       - idp
 
+  rabbitmq:
```
withinfocus, line 87: 🌱 We aren't going to actually commit this, as our stance for self-host will initially be to "bring your own queuing system". This will end up in documentation vs. being here, as the "out of the box" self-host we ship won't support pub / sub but what we already have, and then again our docs will inform self-hosters about the additional processors offered (as independent repositories with their own deployments).

brant-livefront: @withinfocus OK, should I pull this out from here and just put it in the docs?

Just to be clear, its presence here wouldn't fire up the container unless you explicitly ask to run the `rabbitmq` profile (i.e. like `postgres`). But given this is just an optional addition, it's also easy to be absolutely sure it's opt-in and require people to copy it from the docs.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
brant-livefront: I think this should be fixed now, but let me know if I need to refine it further. 👍

dev/docker-compose.yml
withinfocus: Ah yes, I was just doing a quick scan and thought this was mixed in with the other images. I could see us leaving this as-is.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Bit.Core.Services;
+
+public class RabbitMqEventHttpPostListener : RabbitMqEventListenerBase
+{
+    private readonly IHttpClientFactory _httpClientFactory;
+    private readonly string _httpPostUrl;
+
+    protected override string QueueName => "events-httpPost-queue";
```
withinfocus, line 13: 🎨 Should allow this to be configured.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs
```diff
@@ -0,0 +1,27 @@
...
+    public RabbitMqEventRepositoryListener(
+        IEventRepository eventRepository,
+        ILogger<RabbitMqEventListenerBase> logger,
+        GlobalSettings globalSettings)
+        : base(logger, globalSettings)
+    {
+        _eventWriteService = new RepositoryEventWriteService(eventRepository);
+    }
+
+    protected override string QueueName => "events-write-queue";
```
withinfocus, line 21: 🎨 Also configurable.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
```diff
@@ -0,0 +1,33 @@
...
+    public RabbitMqEventHttpPostListener(
+        IHttpClientFactory httpClientFactory,
+        ILogger<RabbitMqEventListenerBase> logger,
+        GlobalSettings globalSettings)
+        : base(logger, globalSettings)
+    {
+        _httpClientFactory = httpClientFactory;
+        _httpPostUrl = globalSettings.RabbitMqHttpPostUrl;
+    }
+    protected override async Task HandleMessageAsync(EventMessage eventMessage)
```
withinfocus, line 26: ⛏️ Newline between.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventListenerBase.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public override async Task StartAsync(CancellationToken cancellationToken)
+    {
+        _connection = await _factory.CreateConnectionAsync(cancellationToken);
+        _channel = await _connection.CreateChannelAsync(cancellationToken: cancellationToken);
+
+        await _channel.ExchangeDeclareAsync(exchange: _exchangeName, type: ExchangeType.Fanout);
+        await _channel.QueueDeclareAsync(queue: QueueName,
+                                         durable: false,
```
withinfocus, line 42: ❓ Don't we want this to be [durable](https://www.rabbitmq.com/docs/queues#durability)?

src/Core/AdminConsole/Services/Implementations/RabbitMqEventListenerBase.cs
```diff
@@ -0,0 +1,90 @@
...
+                                         exclusive: false,
+                                         autoDelete: false,
+                                         arguments: null,
+                                         cancellationToken: cancellationToken);
+        await _channel.QueueBindAsync(queue: QueueName, exchange: _exchangeName, routingKey: "");
```
withinfocus, line 47: ❓ Are routing and binding keys needed, and should they somehow be part of the implementation? I suppose this is handled by the queue implicitly, but it may need to be configurable.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventWriteService.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public async Task CreateAsync(IEvent e)
+    {
+        var connection = await _lazyConnection.Value;
+        using var channel = await connection.CreateChannelAsync();
+
+        await channel.ExchangeDeclareAsync(exchange: _exchangeName, type: ExchangeType.Fanout);
+
+        var body = JsonSerializer.SerializeToUtf8Bytes(e);
```
withinfocus, line 33: ❓ Is JSON our best option for serialization?

src/Core/Settings/GlobalSettings.cs
```diff
@@ -53,6 +53,8 @@ public virtual string LicenseDirectory
     public virtual SqlSettings PostgreSql { get; set; } = new SqlSettings();
     public virtual SqlSettings MySql { get; set; } = new SqlSettings();
     public virtual SqlSettings Sqlite { get; set; } = new SqlSettings() { ConnectionString = "Data Source=:memory:" };
+    public virtual AmqpSettings RabbitMq { get; set; } = new AmqpSettings();
+    public virtual string RabbitMqHttpPostUrl { get; set; }
```
withinfocus, line 57: 💭 I'm not sure what to call this or where it should go, but it feels too specific here. I think it needs some kind of "event log" prefixing too.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventListenerBase.cs
withinfocus: Also, I like to use `string.Empty`.

withinfocus (changes requested):
Few things and some questions.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventListenerBase.cs
brant-livefront: Good catch, fixed in next commit. Also updated the exchange as well.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventListenerBase.cs
brant-livefront: The routing keys are not needed - as you mentioned, the queues are doing that job for us. But we do have to specify them, so I've changed to `string.Empty` 👍

src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
brant-livefront: I was ambivalent about this when adding it as well. I made the _exchange_ configurable, but wasn't sure on this one (since it maps so directly with this one service). I'll adjust it to being configurable.  👍

src/Core/Settings/GlobalSettings.cs
brant-livefront: This latest commit has what we discussed in Slack: new EvenetLogging top-level object with a RabbitMq object underneath that to contain the RabbitMQ-specific properties.

brant-livefront:
If you're running this to test locally, note that in refactoring all of the global settings, you'll need to adopt the new secrets and replublish them. The sample secrets are included in the description above (I've updated it to match the new format). After adding those, you'll need to run:

    ```bash
    pwsh setup_secrets.ps1 -clear
    ```

withinfocus (dismissed):
This makes sense to me as a first pass and we'll iterate on some additional topics as discussed.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventHttpPostListener.cs
```diff
@@ -0,0 +1,36 @@
...
+        _queueName = globalSettings.EventLogging.RabbitMq.HttpPostQueueName;
+    }
+
+    protected override async Task HandleMessageAsync(EventMessage eventMessage)
+    {
+        using var httpClient = _httpClientFactory.CreateClient(HttpClientName);
+        var content = JsonContent.Create(eventMessage);
+        var response = await httpClient.PostAsync(_httpPostUrl, content);
+        response.EnsureSuccessStatusCode();
+    }
```
justindbaur, line 35: We can just get the HttpClient a single time in the constructor like so:

```suggestion
    private readonly HttpClient _httpClient;
    private readonly string _httpPostUrl;
    private readonly string _queueName;

    protected override string QueueName => _queueName;

    public const string HttpClientName = "EventHttpPostListenerHttpClient";

    public RabbitMqEventHttpPostListener(
        IHttpClientFactory httpClientFactory,
        ILogger<RabbitMqEventListenerBase> logger,
        GlobalSettings globalSettings)
        : base(logger, globalSettings)
    {
        _httpClient = httpClientFactory.CreateClient(HttpClientName);
        _httpPostUrl = globalSettings.EventLogging.RabbitMq.HttpPostUrl;
        _queueName = globalSettings.EventLogging.RabbitMq.HttpPostQueueName;
    }

    protected override async Task HandleMessageAsync(EventMessage eventMessage)
    {
        var content = JsonContent.Create(eventMessage);
        var response = await _httpClient.PostAsync(_httpPostUrl, content);
        response.EnsureSuccessStatusCode();
    }
```

src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    protected override string QueueName => _queueName;
+
+    public RabbitMqEventRepositoryListener(
+        IEventRepository eventRepository,
+        ILogger<RabbitMqEventListenerBase> logger,
+        GlobalSettings globalSettings)
+        : base(logger, globalSettings)
+    {
+        _eventWriteService = new RepositoryEventWriteService(eventRepository);
```
justindbaur, line 21: Are we able to avoid creating a service here and rely on dependency injection instead? Most importantly by doing this if `RepositoryEventWriteService` ever implemented `I{Async}Disposable` in the future since we create it and own it we'd be responsible for disposing it but we might forget to come back here and do it.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventWriteService.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        _factory = new ConnectionFactory
+        {
+            HostName = globalSettings.EventLogging.RabbitMq.HostName,
+            UserName = globalSettings.EventLogging.RabbitMq.Username,
+            Password = globalSettings.EventLogging.RabbitMq.Password
+        };
+        _exchangeName = globalSettings.EventLogging.RabbitMq.ExchangeName;
+
+        _lazyConnection = new Lazy<Task<IConnection>>(CreateConnectionAsync);
```
justindbaur, line 23: `IConnection` implements `IAsyncDisposable` so this class should also implement that and then check:

```c#
if (_lazyConnection.IsValueCreated)
{
    var connection = await _lazyConnection.Value;
    await connection.DisposeAsync();
}

**Review by justindbaur — COMMENTED** [review]
A few more thoughts.

File: src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs (hunk 16 above)
Comment (brant-livefront) [reply src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs:21]: I'd be open to suggestions, but I didn't see a way around this. I'm using the RabbitMQ implementation as the IEventWriteService and that definition is global. So from what I was seeing, I didn't see a way to inject that here but keep the injection to the EventService of the RabbitMQ implementation.

I tried to pattern this off of what we're doing in `AzureQueueHostedService` where it creates its own `RepositoryEventWriteService`.

File: src/Core/AdminConsole/Services/Implementations/RabbitMqEventWriteService.cs (hunk 17 above)
Comment (brant-livefront) [reply src/Core/AdminConsole/Services/Implementations/RabbitMqEventWriteService.cs:23]: Fixed in latest commit. 👍

File: src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs (hunk 16 above)
Comment (justindbaur) [reply src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs:21]: We can take advantage of keyed services to make this happen. We continue to register `RabbitMqEventWriteService` with no key as that is the service 99% of consumers should use. But we can annotate this constructor with `[FromKeyedServices("Persistent"] IEventWriteService` and then have:

```c#
services.AddKeyedSingleton<IEventWriteService, RepositoryEventWriteService>("Persistent");
```

probably right next to where `RabbitMqEventRepositoryListener` is registered. 

My thinking on the name key being `"Persistent"` is that the goal here is to take the event and actually write it somewhere persistent and we could choose a non-repository based implementation and the `Persistent` would still fit vs using the key `Repository` wouldn't. 

We should consider updating `AzureQueueHostedService` from doing this too though. It existed before keyed services I would have fixed this problem more manually like:

```c#
// Register the concrete type in DI
services.AddSingleton<RepositoryEventWriteService>();
// Manually do a factory for the listener
services.AddHostedService<RabbitMqEventRepositoryListener>(sp =>
{
    return new RabbitMqEventRepositoryListener(
        sp.GetRequiredService<RepositoryEventWriteService>(),
        sp.GetRequiredService<ILogger<RabbitMqEventListenerBase>>(),
        sp.GetRequiredService<GlobalSettings>()
    );
});
```

You get the benefit of being able to accept `IEventWriteService` in `RabbitMqEventRepositoryListener` so you can mock it in tests easier and you get the DI owning the lifetime of `RepositoryEventWriteService` but you do have to constantly come back here and make changes when you update the constructor which the keyed services usage gives you automatically.

src/Core/AdminConsole/Services/Implementations/RabbitMqEventRepositoryListener.cs
```diff
@@ -13,12 +13,12 @@ public class RabbitMqEventRepositoryListener : RabbitMqEventListenerBase
     protected override string QueueName => _queueName;
 
     public RabbitMqEventRepositoryListener(
-        IEventRepository eventRepository,
+        [FromKeyedServices("Persistent")] IEventWriteService eventWriteService,
```
withinfocus, line 16: ⛏️ We have this lowercase ("persistent") elsewhere.

jrmccannon (approved):
Looks good from my perspective.

@withinfocus  Is this POC a step in the direction of decoupling services and moving toward an event based architecture? Or is there a different end goal for this?

withinfocus:
Not necessarily, yet -- just allowing our events to be ingested in more places. It's definitely an implicit enabler of that though, for the eventing angle at least. We could use this same backbone for other interprocess communication.

withinfocus (approved):

PR #5336, opened 2025-01-29
[PM-14439] Add PolicyRequirementQuery for enforcement logic

eliykat:
## 🎟️ Tracking

### Background
Enterprise policies are organization-wide settings that limit app behavior for members. For example, if the Disable Send policy is enabled, members of that organization cannot use the Send feature.

Policies are complex because they're cross-domain and cross-team, affecting any area of the application. They can also have different enforcement logic. For example:

- most policies exempt Owners and Admins from their effect, but some don't
- the SSO Policy sometimes exempts Owners, but this depends on an `IGlobalSettings` override
- most policies are only enforced if the user is in an Accepted state, but some need to be checked before then
- different organizations can have conflicting policies which must be reconciled

Today, this is mostly handled by `PolicyService`. However, this has to know about every type of policy, which is not very maintainable or extendible. It also has a poor interface:
- `AnyPoliciesApplicableToUserAsync` returns true if any policy of a given type is Enabled - which is handy for very simple boolean policies, but only covers a subset of policies.
- `GetPoliciesApplicableToUserAsync` returns all policies of a given type that should be enforced against a user, but then leaves the caller to evaluate and reconcile the `Policy` objects (including deserializing the `Data` json blob).

Additionally, many devs don't know about this, and access policies directly using the repository. This bypasses the business logic entirely and leads to repeated bugs.

### Proposed solution

1. We need to stop dealing with raw `Policy` objects in our business logic. This PR introduces the idea of a "policy requirement", which is the result of filtering and combining policies to create a single object. This object represents business requirements rather than data. In other words, it's a bridge between the policy domain and the caller's domain. Callers no longer have to know about the underlying policies.
2. A policy requirement is defined for each policy type. It can be anything - whatever is most convenient for representing a policy type's business requirements.
3. Policy requirements define their own factory function, which takes (at a minimum) all policies that could potentially affect the user. This function defines the policy type's specific rules, e.g. who it applies to.
4. Policy requirements are registered with the `PolicyRequirementQuery`.
5. Callers can call the `PolicyRequirementQuery` to get the requirement they want to check.

This PR implements a draft version of this for feedback.

### TODO

- enable nullable
- review requirements to make sure their business logic is correct
- update all calling code to use the new query
- use feature flags
- SQL migration script
- EF implementation
- (in a later PR) consider adding a response model for each requirement and including it in sync data (so the clients don't have to reimplement this logic)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Sql/dbo/Stored Procedures/PolicyDetails_ReadByUserId.sql
eliykat, line 1: Point of interest: this sproc only returns policies where they're enabled and the organization's plan supports them. This is a fundamental requirement of the policy domain so it's not left up to the policy requirements to decide. Of course if this changed in the future we could lift it up, but this best fits our current logic.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementQuery.cs
eliykat, line 1: There is an intentional split of responsibilities here: all business logic is in the policy requirements, which are written in a functional style. The query is agnostic about what policies are being handled - its only job is to connect policy requirements to dependencies.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendRequirement.cs
eliykat, line 1: Point of interest: we have a separate policy type for "Disable Send" and "Send Options". In practice we always need to check both, so we can encapsulate them in the same requirement.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SingleOrganizationRequirement.cs
eliykat, line 1: Point of interest: an example of a more complex requirement that can encapsulate additional logic (beyond enabled/disabled) that is currently in the caller (in this case, `OrganizationService`).

src/Core/Tools/Services/Implementations/SendService.cs
eliykat, line 1: Simple example usage.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementQuery.cs
```diff
@@ -0,0 +1,58 @@
...
+    private readonly PolicyRequirementRegistry _policyRequirements = new();
+
+    public PolicyRequirementQuery(IGlobalSettings globalSettings, IPolicyRepository policyRepository)
+    {
+        _policyRepository = policyRepository;
+
+        // Register Policy Requirement factory functions below
+        _policyRequirements.Add(SendPolicyRequirement.Create);
+        _policyRequirements.Add(up
+            => SsoPolicyRequirement.Create(up, globalSettings.Sso));
```
eliykat, line 21: Example of how you can just supply the factory function for simple requirements (which will be most of them), or inject additional values if needed (settings, feature flags, etc).

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.AdminConsole.Models.Data.Organizations.Policies;
+using Bit.Core.Models.Data.Organizations.OrganizationUsers;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
+
+public class SendPolicyRequirement : IPolicyRequirement
+{
+    public bool DisableSend { get; init; }
```
r-tome, line 9:
```suggestion
    public bool DisableSend { get; private init; }
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/PolicyRequirementHelpers.cs
r-tome, line 1: This is great!

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/PolicyRequirementHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+public static class PolicyRequirementHelpers
+{
+    public static IEnumerable<OrganizationUserPolicyDetails> GetPolicyType(
+        this IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails,
+        PolicyType type) =>
+            userPolicyDetails.Where(x => x.PolicyType == type);
+
+    public static IEnumerable<OrganizationUserPolicyDetails> ExcludeOwnersAndAdmins(
+        this IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails) =>
+            userPolicyDetails.Where(x => x.OrganizationUserType != OrganizationUserType.Owner);
```
r-tome, line 16:
```suggestion
            userPolicyDetails.Where(x => x.OrganizationUserType is not OrganizationUserType.Owner and not OrganizationUserType.Admin);
```

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementQuery.cs
r-tome: Just an idea but maybe we could use Reflection here to find all classes that inherit `IPolicyRequirement` and automatically register them. Maybe it adds some unnecessary complexity though.

r-tome (commented):
This is looking very good! Please add some xmldoc to the interfaces

jrmccannon (commented):
Looks good overall. Had a few comments.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SingleOrganizationPolicyRequirement.cs
```diff
@@ -0,0 +1,45 @@
...
+    Ok = 1,
+    RequiredByThisOrganization = 2,
+    RequiredByOtherOrganization = 3
+}
+
+public class SingleOrganizationPolicyRequirement : IPolicyRequirement
+{
+    private IEnumerable<OrganizationUserPolicyDetails> PolicyDetails { get; }
+
+    public SingleOrganizationPolicyRequirement(IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails)
```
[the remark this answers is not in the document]
eliykat: Yes we do, however I've moved the PolicyRequirement implementations to a separate branch. I'll address that before putting it up for review.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/TwoFactorAuthenticationPolicyRequirement.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.Models.Data.Organizations.OrganizationUsers;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
+
+public class TwoFactorAuthenticationPolicyRequirement : IPolicyRequirement
+{
+    private IEnumerable<OrganizationUserPolicyDetails> PolicyDetails { get; }
+
+    public TwoFactorAuthenticationPolicyRequirement(IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails)
```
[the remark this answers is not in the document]
eliykat: Same as above

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
eliykat: I've moved the PolicyRequirement implementations to a separate branch but I'll make a note of it 👍

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementQuery.cs
eliykat: I agree this would be too complex, and probably difficult for others to find/understand. My latest changes moved this to use the standard DI system which I think streamlines this a bit.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/PolicyRequirementHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+        this IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails) =>
+            userPolicyDetails.Where(x => !x.IsProvider);
+
+    public static IEnumerable<OrganizationUserPolicyDetails> ExcludeRevokedAndInvitedUsers(
+        this IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails) =>
+            userPolicyDetails.Where(x => x.OrganizationUserStatus >= OrganizationUserStatusType.Accepted);
+
+    public static IEnumerable<OrganizationUserPolicyDetails> ExcludeRevokedUsers(
+        this IEnumerable<OrganizationUserPolicyDetails> userPolicyDetails) =>
+            userPolicyDetails.Where(x => x.OrganizationUserStatus >= OrganizationUserStatusType.Invited);
```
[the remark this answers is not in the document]
eliykat: I don't feel strongly about it - in this case the enums are ordered to represent the progression of statuses (invited -> accepted -> confirmed). _But_ I'm also OK with being more explicit, especially as this is a cross-domain concern and other teams will be looking at this. I've reworked these a little.

eliykat:
This is now ready for review. Main changes since draft:
- I've removed the actual PolicyRequirement implementations. I will add those and wire them up in separate PRs, so that this one can focus on just getting the core query in place.
- I figured out how we can inject the policy requirement factories into the query via the regular DI system, rather than recreating our own mini DI in the query. This is great because (1) much more straightforward, and (2) the query has almost no code to test. The private `AddPolicyRequirement` methods ensure that the delegate is correctly typed to be picked up by the query later.
- refactored the extension helper methods
- implemented the EF query
- added tests for everything
- xmldoc'd the heck outta everything

Follow up PRs will start adding requirements and refactoring calling locations to use them.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
```diff
@@ -12,12 +13,46 @@ public static void AddPolicyServices(this IServiceCollection services)
...
         services.AddScoped<IPolicyValidator, SingleOrgPolicyValidator>();
         services.AddScoped<IPolicyValidator, RequireSsoPolicyValidator>();
         services.AddScoped<IPolicyValidator, ResetPasswordPolicyValidator>();
         services.AddScoped<IPolicyValidator, MaximumVaultTimeoutPolicyValidator>();
         services.AddScoped<IPolicyValidator, FreeFamiliesForEnterprisePolicyValidator>();
     }
+
+    private static void AddPolicyRequirements(this IServiceCollection services)
+    {
+        // Register policy requirement factories here
```
eliykat, line 34: Example usage, given that I removed the actual implementations from this PR:

```cs
// simple case, this will be most of them:
services.AddPolicyRequirement<SendPolicyRequirement>(SendPolicyRequirement.Create);

// occasionally we need to inject something else:
services.AddPolicyRequirement<SsoPolicyRequirement>(sp => policies => SsoPolicyRequirement.Create(policies, sp.GetRequiredService<IGlobalSettings>()));
```

src/Core/AdminConsole/Repositories/IPolicyRepository.cs
```diff
@@ -8,7 +10,15 @@ namespace Bit.Core.AdminConsole.Repositories;
...
+    /// Gets all policies of a given type for an organization.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: do not use this to enforce policies against a user! It returns raw data and does not take into account
+    /// various business rules. Use <see cref="IPolicyRequirementQuery"/> instead.
+    /// </remarks>
     Task<Policy?> GetByOrganizationIdTypeAsync(Guid organizationId, PolicyType type);
     Task<ICollection<Policy>> GetManyByOrganizationIdAsync(Guid organizationId);
     Task<ICollection<Policy>> GetManyByUserIdAsync(Guid userId);
+    Task<IEnumerable<PolicyDetails>> GetPolicyDetailsByUserId(Guid userId);
```
r-tome, line 23: All the xmldocs are great, please add some for this method 😁

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/PolicyRepository/GetPolicyDetailsByUserIdTests.cs
r-tome, line 1: Tests missing for Revoked users

src/Core/AdminConsole/Services/IPolicyService.cs
r-tome, line 1: Perhaps its best to defer marking these methods as obsolete only once everything is implemented/wired up.

src/Sql/dbo/Stored Procedures/PolicyDetails_ReadByUserId.sql
```diff
@@ -0,0 +1,37 @@
...
+WHERE
+    P.Enabled = 1
+    AND O.Enabled = 1
+    AND O.UsePolicies = 1
+    AND (
+            (OU.[Status] != 0 AND OU.[UserId] = @UserId) -- OrgUsers who have accepted their invite and are linked to a UserId
+            OR EXISTS (
+                SELECT 1
+                FROM [dbo].[UserView] U
+                WHERE U.[Id] = @UserId AND OU.[Email] = U.[Email] AND OU.[Status] = 0 -- 'Invited' OrgUsers are not linked to a UserId yet, so we have to look up their email
```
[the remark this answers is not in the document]
eliykat: This is a great point - the reason why we treat invited users differently is because their email is effectively the foreign key rather than the userId. But if an _invited_ user is revoked, they're in the revoked status but still linked by the email only. I'll update this.

src/Core/AdminConsole/Repositories/IPolicyRepository.cs
eliykat: Done!

src/Sql/dbo/Stored Procedures/PolicyDetails_ReadByUserId.sql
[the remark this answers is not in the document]
eliykat: Actually - this works, because we don't enforce policies when restoring a "revoked invited" user. This is because the policies will be checked when they accept the invite after being restored. Ref: https://github.com/bitwarden/server/blob/5d3294c3768aea9312b97691e6216464f496795a/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L2192-L2198

It is a bit of a weird edge case, I think I'm OK with leaving this as handled by the sproc here, because no `PolicyRequirement` should ever be enforced in this case, and it's confusing to try to represent this state in the `PolicyDetails`. In other words: it's too much of an edge case to expose to the other domain. Instead I'll note it in the xmldoc and test coverage.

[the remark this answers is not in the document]
eliykat: Also: this is how the current (very similar) `OrganizationUser_ReadByUserIdWithPolicyDetails` works today and it doesn't seem to have caused any issues 🤞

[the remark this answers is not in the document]
eliykat: This is a bit awkward to explain in comments, so I tried just "fixing" it to include these users. If we don't check policies in this case anyway there's probably no harm, and arguably it makes the query more robust. But here's how the WHERE clause of the query turned out:

```sql
WHERE
    P.Enabled = 1
    AND O.Enabled = 1
    AND O.UsePolicies = 1
    -- Match to UserId - this is done differently depending on the OrganizationUserStatus
    AND CASE
      -- Invited users have a NULL OU.UserId but can be matched on email (including revoked from the invited status)
      WHEN OU.[Status] = 0 OR (OU.[Status] = -1 AND OU.[UserId] IS NULL) THEN EXISTS (
                SELECT 1
                FROM [dbo].[UserView] U
                WHERE U.[Id] = @UserId AND OU.[Email] = U.[Email] AND OU.[Status] = 0
            )
      -- Accepted and confirmed users are matched on UserId (including revoked)
      WHEN OU.[Status] = 1 OR OU.[Status] = 2 OR (OU.[Status] = -1 AND OU.[UserId] IS NOT NULL) THEN OU.[UserId] = @UserId
      ELSE 0
    END = 1
```

That seems more complicated and likely to generate a worse query plan for something we don't actually need, so I continued with my approach above instead. Let me know what you think.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/PolicyRepository/GetPolicyDetailsByUserIdTests.cs
eliykat: Done

eliykat:
Feedback above has been addressed, although the whole "revoked invited users" thing may or may not require further discussion.

DB tests are failing due to an unrelated PR in `main`, I've followed up with KM Team to fix.

r-tome (approved):
Excellent work! Thanks for addressing the feedback

src/Sql/dbo/Stored Procedures/PolicyDetails_ReadByUserId.sql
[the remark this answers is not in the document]
r-tome: I agree with that! Thanks for considering this.

I think it shouldn't even be possible to revoke an invited user. I also agree with you that the Revoked status should be on a separate column. This is a whole separate discussion for a future refactor.

rkac-bw (approved):
LGTM

r-tome (approved):

PR #5188, opened 2024-12-23
[PM-14381] Add POST /tasks/bulk-create endpoint

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-14381

## 📔 Objective
Admin endpoint to bulk create new tasks on the Access Intelligence page. Should accept a list of CipherIdsand the TaskType.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #5344, opened 2025-01-29
[PM-14406] Security Task Notifications

nick-livefront:
## 🎟️ Tracking

[PM-14406](https://bitwarden.atlassian.net/browse/PM-14406)
Client side PR: https://github.com/bitwarden/clients/pull/13135
Stacked on top of: https://github.com/bitwarden/server/pull/5188

## 📔 Objective

Adds notifications for newly created security tasks for all users within an organization that have manage permissions for the applicable ciphers.
- I tried to follow/pick up naming patterns, if I missed any please point them out 😄 
- Adds an email template & a new security task layout.
  - A new layout might not be necessary and could be done later if we wish. I'd be happy to pull this into a single template. I figured that this could be the first of many emails sent out regarding security tasks.
  - The template currently links to a page that hasn't been built yet. That will be handled by [PM-17564](https://bitwarden.atlassian.net/browse/PM-17564).
- Adds stored queries for finding all users that have Manage permissions for the ciphers that are associated with the security tasks.
  - The query returns the UserId, their Email and the applicable cipher id for the security task.
- ~~Add push notification that is sent to the client side for handling~~
  - ~~[Client side PR](https://github.com/bitwarden/clients/pull/13135). This only processes the notification, there is no handling as of yet. This is only to avoid any errors.~~
  - ~~There is work slated to handle that in [PM-10610](https://bitwarden.atlassian.net/browse/PM-10610)~~

Edit: @shane-melton pointed me to some work that the KM team is doing and I have refactored to use their `CreateNotificationCommand` rather than the one that I put together.  

## 📸 Screenshots

Steps in video
1. Bulk create endpoint is hit (work done in https://github.com/bitwarden/server/pull/5188)
2. Email notifications received 
3. ~~Push notifications received~~ No longer applicable

<video src="https://github.com/user-attachments/assets/9b9354f6-890e-421d-9b92-c4ffb072729c" />

## 🦮 Reviewer guidelines

[PM-14406]: https://bitwarden.atlassian.net/browse/PM-14406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-17564]: https://bitwarden.atlassian.net/browse/PM-17564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-10610]: https://bitwarden.atlassian.net/browse/PM-10610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/Vault/Controllers/SecurityTaskController.cs
```diff
@@ -87,6 +104,22 @@ public async Task<ListResponseModel<SecurityTasksResponseModel>> BulkCreateTasks
...
+        var organization = await _organizationRepository.GetByIdAsync(orgId);
+        await _mailService.SendBulkSecurityTaskNotificationsAsync(organization.Name, userTaskCount);
+        foreach (var task in userTaskCount)
+        {
+            var notification = new Notification
+            {
+                UserId = task.UserId,
+                OrganizationId = orgId,
+                Priority = Priority.Informational,
+                ClientType = ClientType.Browser,
```
nick-livefront, line 118: I was unsure of what properties to set here this felt like a good starting point for the task notifications.

shane-melton: Great question, these seem reasonable to me though!

❓ Is this creating multiple notifications per user per task? I haven't dug too deep into the PR yet, but this is what we're going for now and we'll likely need to update this to include the `TaskId` in the `Notification` (this is being done in [PM-14590 [Server] Modify Notification database table](https://bitwarden.atlassian.net/browse/PM-14590?focusedCommentId=84236)

nick-livefront: Currently this sends a single notification (push and email) to an individual. Are you saying that we want to have a notification for each security task that the user can action on? That will be kind of noisy, no? X number of emails for each applicable task.

withinfocus: There will be a notification row in our backend database associated to a task for a variety of reasons to make management of the tasks better, but that is not to imply that communications such as push, email, etc. are 1:1 -- it would be per bulk operation.

nick-livefront: 🚀 I updated with Gbubemi's work that was merged in and then altered my logic to include the `taskId` when creating the notifications

src/Api/Vault/Controllers/SecurityTaskController.cs
```diff
@@ -87,6 +105,37 @@ public async Task<ListResponseModel<SecurityTasksResponseModel>> BulkCreateTasks
         [FromBody] BulkCreateSecurityTasksRequestModel model)
     {
         var securityTasks = await _createManyTasksCommand.CreateAsync(orgId, model.Tasks);
+
+        var securityTaskCiphers = await _getSecurityTasksNotificationDetailsQuery.GetNotificationDetailsByManyIds(orgId, securityTasks);
```
shane-melton, line 109: 🎨 I think it could be nice to put this logic into its own CQRS command to avoid "polluting" the controller with so many other services and repositories.

src/Core/Vault/Models/Data/UserSecurityTaskCipher.cs
shane-melton, line 1: ❓ Should we have the `Taskid` on this model too? It would make creating the notification entities a bit easier in the controller as we wouldn't need to lookup the Task via the CipherId every time.

src/Api/Vault/Controllers/SecurityTaskController.cs
```diff
@@ -87,6 +105,37 @@ public async Task<ListResponseModel<SecurityTasksResponseModel>> BulkCreateTasks
...
+            // Create a notification for the user with the associated task
+            var notification = new Notification
+            {
+                UserId = userSecurityTaskCipher.UserId,
+                OrganizationId = orgId,
+                Priority = Priority.Informational,
+                ClientType = ClientType.Browser,
+                TaskId = securityTask.Id
+            };
+            await _createNotificationCommand.CreateAsync(notification);
```
shane-melton, line 136: 🎨  We'll need to tweak this command to avoid "pushing" every created notification per user out to the clients. We could probably add an optional argument like `bool skipPush = false` to the `CreateAsync()` signature and have it skip the call to the `_pushNotificationService.PushNotificationAsync`.

💭 I think we could get away with only sending the push notification for the last notification created per user. Because, once they get the push notification, the client will be instructed to fetch all the tasks. _**But**_, there may be other drawbacks and maybe we do need to explicitly push a "bulk" notifications created payload.

src/Api/Vault/Controllers/SecurityTaskController.cs
```diff
@@ -87,6 +105,37 @@ public async Task<ListResponseModel<SecurityTasksResponseModel>> BulkCreateTasks
...
+            TaskCount = x.Count()
+        }).ToList();
+
+        var organization = await _organizationRepository.GetByIdAsync(orgId);
+        await _mailService.SendBulkSecurityTaskNotificationsAsync(organization.Name, userTaskCount);
+
+        foreach (var userSecurityTaskCipher in securityTaskCiphers)
+        {
+            // Get the associated security task for the cipher
+            var securityTask = securityTasks.First(x => x.CipherId == userSecurityTaskCipher.CipherId);
```
shane-melton, line 126: ❓ Similar to my other comment, could we add the `TaskId` to the `_getSecurityTasksNotificationDetailsQuery.GetNotificationDetailsByManyIds` response data model so we don't have to look it up repeatedly here?

shane-melton (changes requested):
This is looking really good, nice work! I just have a few suggestions and an change we'll need to make to the `CreateNotificationCommand`.

src/Core/Vault/Models/Data/UserSecurityTaskCipher.cs
nick-livefront: Agreed! I had ended up having to make another Model to account for the DB queries don't know about the security tasks, they only care about the cipherIds.

Now `GetUserSecurityTasksByCipherIdsAsync` returns the `TaskId` as well which I do like a lot better.

[c304667](https://github.com/bitwarden/server/pull/5344/commits/c3046675e2d277916df38ed49abc1c11ac2340d2) & [906fb54](https://github.com/bitwarden/server/pull/5344/commits/906fb5488d3bf8b231e637a953a611dd942c2cdd)

src/Api/Vault/Controllers/SecurityTaskController.cs
nick-livefront: It did get a little out of hand, great suggestion: [cd9a2c5](https://github.com/bitwarden/server/pull/5344/commits/cd9a2c5e79edc4ce55b1d4433c763938e4d4b847)

shane-melton (dismissed):

shane-melton (approved):

withinfocus (commented):
Not a thorough review but a couple things seen.

src/Core/Vault/Commands/CreateManyTaskNotificationsCommand.cs
```diff
@@ -0,0 +1,78 @@
...
+                    OrganizationId = orgId,
+                    Priority = Priority.Informational,
+                    ClientType = ClientType.Browser,
+                    TaskId = userSecurityTaskCipher.TaskId
+                };
+
+                // Only push the last notification for each user
+                bool skipNotificationPush = i != userSecurityTaskCiphers.Count - 1;
+
+                await _createNotificationCommand.CreateAsync(notification, skipNotificationPush);
```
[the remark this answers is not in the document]
withinfocus: The semantics can get confusing here -- I am totally aligned that we should have a database `Notification` row per task per user, no issue there; what I am speaking about is a _push_ for each notification. We can do what we want with dynamic push payloads. I know that's not what in sitting in the codebase right not but that's OK, we can change / amend it for this use case.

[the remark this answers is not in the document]
shane-melton: Yep, agreed we want to send the general "you have tasks" notification here. This was a temporary solution until we ironed out what that "you have tasks" notification looks like. 

You mention dynamic push payloads, but I'm not seeing a means to do that with the current [push notification API](https://github.com/bitwarden/server/blob/9f4aa1ab2bcda4aaef9dcf25350a1f3f78eb3417/src/Core/Platform/Push/Services/AzureQueuePushNotificationService.cs#L180). Specifically, the `IPushNotificationService.PushNotificationAsync` method only accepts the `Notification` entity and primarily supports `Body`/`Title` properties for the payload data. 

We originally planned to add a `NotificationType` to the `Notification` database entity, but perhaps we can do that for the push notification instead and have a `YouHavePendingTasks` push notification type. We'd need to adjust the `IPushNotificationService.PushNotificationAsync` method or perhaps add another method entirely for these unique pus notification types.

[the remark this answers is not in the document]
withinfocus: Yeah, I was expecting a modification to go beyond the database-oriented `Notification` class in the push services here. That works when it's one, but what about when it's many? And don't these task-based notifications not even have a title and body? The pushes don't need to be enumerated but for something other than that domain object to be provided -- the push is just JSON and it can become anything.

rkac-bw (dismissed):
lgtm

src/Core/Vault/Commands/CreateManyTaskNotificationsCommand.cs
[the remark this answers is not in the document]
nick-livefront: @shane-melton @withinfocus 

I chatted with Shane and we arrived a new notification push type for `PendingSecurityTasks`. This results in a notification being stored for each user, for each task but not sent to the user. The `PendingSecurityTasks` type will be pushed to the user.

This doesn't answer your question about title and body, because our scenario doesn't require those. The clients only need to know that they should fetch the associated pending security tasks.  

- Removing all sending of per user - per task [3470e57](https://github.com/bitwarden/server/pull/5344/commits/3470e5784d3b91b39f924264f9e6118d1e39c7d2)
- Add new task type [ddb8bfa](https://github.com/bitwarden/server/pull/5344/commits/ddb8bfa355377acbc89b7a98dc87ed6e588f6521)
-  Push that notification to users [6b02b61](https://github.com/bitwarden/server/pull/5344/commits/6b02b61a8ebab9a21535fb2ed796c1250b01b81d)

shane-melton (approved):

withinfocus (approved):
Looks good after those three commits. I think there's optimizations to be had about now using the explicit new interface method everywhere and we treat the push types more generally with less code (parameterize that perhaps?) but that's a refactor and not explicitly required.

justindbaur (approved):

PR #5345, opened 2025-01-29
[pm-15621] Refactor DeleteManagedOrganizationUserAccountCommand

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15621

## 📔 Objective

1. Refactored DeleteUserAsync to call DeleteManyUsersAsync.
2. Removed the logic for deleting an organization. Since all organizations require at least one owner, EnsureUserIsNotSoleOrganizationOwnerAsync prevents the code from reaching this point.
3. Separated validation from canceling membership.

## 📸 Screenshots

I think test coverage is sufficient here, and it's low risk since it's behind a feature, but I did do a quick smoke test by starting up the service. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,31 +49,12 @@ public DeleteManagedOrganizationUserAccountCommand(
...
-
-        var user = await _userRepository.GetByIdAsync(organizationUser.UserId!.Value);
-        if (user == null)
-        {
-            throw new NotFoundException("Member not found.");
-        }
-
-        await _userService.DeleteAsync(user);
-        await _eventService.LogOrganizationUserEventAsync(organizationUser, EventType.OrganizationUser_Deleted);
+        await DeleteManyUsersAsync(organizationId, new[] { organizationUserId }, deletingUserId);
```
JimmyVo16, line 57: I considered adding a test for this, but it's not straightforward since it calls another method within the same class. I explored using `Substitute.ForPartsOf<T>()`, but it requires making `DeleteManyUsersAsync` virtual. We could also mock everything in `DeleteManyUsersAsync`, but that seems more trouble than it's worth.

eliykat (changes requested):
Nice work @JimmyVo16 , I think this is an improvement. I also have some thoughts about the try/catch blocks here which I shared in our dev channel - I wanted to run it past the team before asking for it here, but let me know what you think.

BTreston (changes requested):
There are some failing tests in `DeleteManagedOrganizationUserAccountCommandTests`

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,140 +50,188 @@ public DeleteManagedOrganizationUserAccountCommand(
...
         if (!orgUser.UserId.HasValue || orgUser.Status == OrganizationUserStatusType.Invited)
         {
             throw new BadRequestException("You cannot delete a member with Invited status.");
         }
-
-        if (deletingUserId.HasValue && orgUser.UserId.Value == deletingUserId.Value)
+    }
+    private static void PreventSelfDeletion(OrganizationUser orgUser, Guid? deletingUserId)
+    {
+        if (!(orgUser.UserId.HasValue && deletingUserId.HasValue))
```
eliykat, line 143: Suggestion: OR would be easier to read here

```suggestion
        if (!orgUser.UserId.HasValue || !deletingUserId.HasValue)
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,140 +50,188 @@ public DeleteManagedOrganizationUserAccountCommand(
...
+            return;
+        }
+        if (orgUser.UserId.Value == deletingUserId.Value)
         {
             throw new BadRequestException("You cannot delete yourself.");
         }
+    }
 
-        if (orgUser.Type == OrganizationUserType.Owner)
+    private async Task EnsureOnlyOwnersCanDeleteOwnersAsync(Guid organizationId, OrganizationUser orgUser, Guid? deletingUserId)
```
eliykat, line 153: Maybe a nit? I think some of these validation methods could be a bit more concise, e.g.

Instead of:

```cs
        if (orgUser.Type != OrganizationUserType.Owner)
        {
            return;
        }

        if (deletingUserId.HasValue && !await _currentContext.OrganizationOwner(organizationId))
        {
            throw new BadRequestException("Only owners can delete other owners.");
        }
```

It could be:

```cs
        if (orgUser.Type == OrganizationUserType.Owner && 
          deletingUserId.HasValue &&
          !await _currentContext.OrganizationOwner(organizationId))
        {
            throw new BadRequestException("Only owners can delete other owners.");
        }
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,140 +50,188 @@ public DeleteManagedOrganizationUserAccountCommand(
...
         if (!managementStatus.TryGetValue(orgUser.Id, out var isManaged) || !isManaged)
         {
             throw new BadRequestException("Member is not managed by the organization.");
         }
     }
 
-    private async Task LogDeletedOrganizationUsersAsync(
-        IEnumerable<OrganizationUser> orgUsers,
-        IEnumerable<(Guid OrgUserId, string? ErrorMessage)> results)
+    private async Task EnsureUserIsNotSoleOrganizationOwnerAsync(User user)
```
eliykat, line 187: This overlaps with `PreventOrganizationSoleOwnerDeletion` (maybe what you meant to refer to in your [comment](https://github.com/bitwarden/server/pull/5345#discussion_r1941831377) above?)

- `PreventOrganizationSoleOwnerDeletion` prevents you from deleting the sole owner of _this_ organization
- `EnsureUserIsNotSoleOrganizationOwnerAsync` prevents you from deleting the sole owner of _any_ organization

I suggest we can just rely on the latter given that it obviously includes the former. Both also require database calls, and we don't want to hit the db unnecessarily.

I also question whether `PreventOrganizationSoleOwnerDeletion` would ever be tripped, given that you can't delete yourself and only owners can delete other owners. If you're the sole owner, who would be able to delete you?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,140 +50,188 @@ public DeleteManagedOrganizationUserAccountCommand(
...
 
-        await LogDeletedOrganizationUsersAsync(orgUsers, results);
+        var users = await _userRepository.GetManyAsync(userIds);
 
-        return results;
+        return users;
     }
 
-    private async Task ValidateDeleteUserAsync(Guid organizationId, OrganizationUser orgUser, Guid? deletingUserId, IDictionary<Guid, bool> managementStatus, bool hasOtherConfirmedOwners)
+    private async Task ValidateDeleteUserAsync(Guid organizationId, OrganizationUser orgUser, User user, Guid? deletingUserId, IDictionary<Guid, bool> managementStatus, bool hasOtherConfirmedOwners)
```
eliykat, line 123: Redundant naming: we wouldn't be validating anything else here.

```suggestion
    private async Task ValidateAsync(Guid organizationId, OrganizationUser orgUser, User user, Guid? deletingUserId, IDictionary<Guid, bool> managementStatus, bool hasOtherConfirmedOwners)
```

eliykat (changes requested):
Just a few minor items.

I think the `userDeletionResults` tuple makes sense (that is, we need to match these objects up and then pass them around together), but is very verbose and awkward. Definitely a case for better/stronger typed ValidationResult classes, which we should experiment with in the future (whoever gets there first, really).

eliykat:
Also - tests are failing. At a glance, maybe just because the order of validation has changed?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,140 +50,188 @@ public DeleteManagedOrganizationUserAccountCommand(
...
+    {
+        EnsureUserStatusIsNotInvited(orgUser);
+        PreventSelfDeletion(orgUser, deletingUserId);
+        EnsureUserIsManagedByOrganization(orgUser, managementStatus);
+        PreventOrganizationSoleOwnerDeletion(orgUser, hasOtherConfirmedOwners);
+
+        await EnsureOnlyOwnersCanDeleteOwnersAsync(organizationId, orgUser, deletingUserId);
+        await EnsureUserIsNotSoleOrganizationOwnerAsync(user);
+        await EnsureUserIsNotSoleProviderOwnerAsync(user);
```
eliykat, line 132: Nice touch calling all the sync validators before the async validators.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
JimmyVo16: After much thinking, yeah i think about right. `EnsureUserIsNotSoleOrganizationOwnerAsync` should handle all cases for use, so we can remove `PreventOrganizationSoleOwnerDeletion`. 

(me thinking out loud) These logics are pretty complex. I wonder if there is a way to make them discoverable than deep in a validation method.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
JimmyVo16:
```c#
        if (orgUser.Type != OrganizationUserType.Owner)
        {
            return;
        }

        if (deletingUserId.HasValue && !await _currentContext.OrganizationOwner(organizationId))
        {
            throw new BadRequestException("Only owners can delete other owners.");
        }
```

I like this option better because it reduces another variable I have to track, so by the time it gets to this point, I can focus on two variables instead of three. 

With that said, it's not something I feel strongly about. I'm fine with either option. Let me know if you want me to update it.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -50,143 +50,172 @@ public DeleteManagedOrganizationUserAccountCommand(
...
+        if (onlyOwnerCount > 0)
         {
-            var orgUser = orgUsers.FirstOrDefault(ou => ou.Id == orgUserId);
-            // If the user was not found or there was an error, we skip logging the event
-            if (orgUser == null || !string.IsNullOrEmpty(errorMessage))
-            {
-                continue;
-            }
+            throw new BadRequestException("Cannot delete this user because it is the sole owner of at least one organization. Please delete these organizations or upgrade another user.");
+        }
```
JimmyVo16, line 175: Actually let me add some tests for these.

JimmyVo16: Done

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
eliykat: I understand that argument - let's leave it how it is.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
eliykat: > (me thinking out loud) These logics are pretty complex. I wonder if there is a way to make them discoverable than deep in a validation method.

I would say validation is the primary source of complexity in our commands (and in old service code that should be commands). It's definitely a strong argument for making this code as simple and as easy to follow as possible.

In terms of discoverability/location, I believe Jared had previously mooted  separate Validator classes which encapsulate the validation logic for a given command. Not sure if that makes it more discoverable, but it would provide a stronger separation than private methods.

eliykat (approved):
Great work @JimmyVo16 !

EDIT: putting a `needs-qa` label on it because it needs QA before merge. However it would be ideal if QA can hit it before we release the larger feature, can you please mention this to them? (a bit awkward refactoring something already cleared for release and behind a feature flag)

JimmyVo16:
Update: QA found an edge case that I missed. 

Changes:
1. For single user deletion, throw an exception if there is an error like the existing behavior.
2. Add test coverage for this.
3. Create `InternalDeleteManyUsersAsync` that would handle all the deletion logic then have `DeleteManyUsersAsync` and `DeleteUserAsync` handle the interfaces their clients expects.

Testing for the edge case

https://github.com/user-attachments/assets/febb2375-bffc-4fad-80a9-9f77b4147275

Regression testing 

https://github.com/user-attachments/assets/4ecd2499-5f4e-48bc-b9e5-5998fdf6585e


https://github.com/user-attachments/assets/6a393dc6-599c-4606-8a50-cd55a016f327

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -55,15 +55,31 @@ IProviderUserRepository providerUserRepository
...
     {
-        await DeleteManyUsersAsync(organizationId, new[] { organizationUserId }, deletingUserId);
+        var result = await InternalDeleteManyUsersAsync(organizationId, new[] { organizationUserId }, deletingUserId);
+
+        var exception = result.Single().exception;
+
+        if (exception != null)
+        {
+            throw exception;
+        }
```
eliykat, line 65: This interface is inconsistent - `DeleteUserAsync` throws if there's a validation error, whereas `DeleteManyUsersAsync` returns validation results up to the caller. I think the latter is a better pattern.

I suggest that `DeleteUserAsync` can still call `DeleteUsersAsync` (how it was before), but it can just return the single result out to the caller; then the caller can handle it. (Which also makes the controller code consistent.)

This method would therefore return a `Task<(Guid OrganizationUserId, string? ErrorMessage)>`. Or in this case, the OrganizationUserId is a bit redundant, so maybe just a `string? ErrorMessage`. If you wanted to take this a step further you could use the `CommandResult` class instead of nullable strings, but I understand if you don't want to do that at this late stage.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -89,21 +105,19 @@ public async Task DeleteUserAsync(Guid organizationId, Guid organizationUserId,
...
                 await CancelPremiumAsync(user);
 
-                userDeletionResults.Add((orgUserId, orgUser, user, string.Empty));
+                userDeletionResults.Add((orgUserId, orgUser, user, null));
             }
-            catch (Exception ex)
+            catch (Exception exception)
             {
-                userDeletionResults.Add((orgUserId, orgUser, user, ex.Message));
+                userDeletionResults.Add((orgUserId, orgUser, user, exception));
```
eliykat, line 112: I understand why you made this change, but given my feedback above I think it's unnecessary. This change also means the try/catch pattern is leaking out to other code (because we need it to give us the `Exception`) which I'd prefer not to do.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
JimmyVo16: Yeah, the interfaces are inconsistent, but they're the existing interfaces. I am trying to avoid breaking them. 

With that said, I do think using `CommandResult` is the right call here and moving away from using exceptions to handle validation errors, but I notice a problem with just returning the error message(s) because, for the single resource endpoint, the HTTP code will be different depending on the error message, unlike the multiple resources endpoint where the HTTP code is the same regardless of the error message. 

I'm going to move this conversation to the AC team Slack channel, so we can iterate quicker.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteManagedOrganizationUserAccountCommand.cs
```diff
@@ -35,158 +36,219 @@ public DeleteManagedOrganizationUserAccountCommand(
...
+    {
+        var userIds = orgUsers
+         .Where(orgUser => orgUser.UserId.HasValue)
+         .Select(orgUser => orgUser.UserId!.Value)
+         .ToList();
+
+        return await _userRepository.GetManyAsync(userIds);
+    }
+
+    private async Task<CommandResult> ValidateAsync(Guid organizationId, OrganizationUser orgUser, User user, Guid? deletingUserId, IDictionary<Guid, bool> managementStatus)
```
JimmyVo16, line 123: @jrmccannon Check out this ValidateAsync

eliykat (commented):
I like the use of `CommandResult` and how the validators are executed. I left some comments in your [related PR](https://github.com/bitwarden/server/pull/5475) about the `CommandResultValidator` but otherwise this is looking good to me.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -578,20 +581,28 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRemo
     [RequireFeature(FeatureFlagKeys.AccountDeprovisioning)]
     [HttpDelete("{id}/delete-account")]
     [HttpPost("{id}/delete-account")]
-    public async Task DeleteAccount(Guid orgId, Guid id)
+    public async Task<IActionResult> DeleteAccount(Guid orgId, Guid id)
```
JimmyVo16, line 584: @eliykat @jrmccannon 

Please review the app code only, not the test code. Start with the controller endpoints for single and multiple removals, then work your way down to the command and validation layers.

The main goal of this first review is to catch any issues early. If everything looks good, I’ll proceed with updating the tests. As you can see, this is a significant refactor.

JimmyVo16:
Currently fixing conflicts right now. It will be ready soon.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedOrganizationUserAccountValidator.cs
```diff
@@ -0,0 +1,173 @@
...
+{
+    public async Task<PartialValidationResult<DeleteUserValidationRequest>> ValidateAsync(List<DeleteUserValidationRequest> requests)
+    {
+        var invalidResults = new List<Invalid<DeleteUserValidationRequest>>();
+        var validResults = new List<Valid<DeleteUserValidationRequest>>();
+
+        foreach (var request in requests)
+        {
+            // The order of the validators matters.
+            // Earlier validators assert nullable properties so that later validators don’t have to.
```
eliykat, line 24: Appreciate the comment 👏

src/Api/Utilities/CommandResultExtensions.cs
```diff
@@ -17,6 +17,18 @@ public static IActionResult MapToActionResult<T>(this CommandResult<T> commandRe
...
+        {
+            NoRecordFoundFailure<T> failure => new ObjectResult(failure.ErrorMessage) { StatusCode = StatusCodes.Status404NotFound },
+            BadRequestFailure<T> failure => new ObjectResult(failure.ErrorMessage) { StatusCode = StatusCodes.Status400BadRequest },
+            Failure<T> failure => new ObjectResult(failure.ErrorMessage) { StatusCode = StatusCodes.Status400BadRequest },
+            Success<T> success => new ObjectResult(success.Value) { StatusCode = StatusCodes.Status200OK },
+            _ => throw new InvalidOperationException($"Unhandled commandResult type: {commandResult.GetType().Name}")
+        };
+    }
+
     public static IActionResult MapToActionResult(this CommandResult commandResult)
```
eliykat, line 32: I think we really need to clean up `ErrorMessages`, `Errors` and `ErrorMessage` in `CommandResult`/`Failure`. I think it's causing this unnecessary boilerplate, needing to distinguish between single vs. multiple error messages.

As far as I can tell, `Failure` will only ever contain 1 error message. If we had multiple, it would correspond to a bulk request which would use `Partial`.

eliykat (changes requested):
Some initial thoughts. Overall I think it works, but I need to spend some more time considering our generic classes for command and validation results - they still feel a bit opaque to me and it's easy to get lost in the generics.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -610,10 +621,16 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkDele
...
-            new OrganizationUserBulkResponseModel(r.OrganizationUserId, r.ErrorMessage)));
+        return MapToOrganizationUserBulkResponseModel(result);
+    }
+
+    private static ListResponseModel<OrganizationUserBulkResponseModel> MapToOrganizationUserBulkResponseModel(Partial<Core.Models.Data.Organizations.DeleteUserResponse> result)
+    {
+        var failures = result.Failures.Select(failure => new OrganizationUserBulkResponseModel(failure.ErroredValue.OrganizationUserId, failure.Message));
+        var successes = result.Successes.Select(success => new OrganizationUserBulkResponseModel(success.OrganizationUserId, string.Empty));
+        return new ListResponseModel<OrganizationUserBulkResponseModel>(failures.Concat(successes));
     }
```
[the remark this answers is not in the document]
JimmyVo16: My main concern with changing this is that it would affect the endpoint interface, which could have downstream consequences. As you can see, this PR is getting pretty big, so I would like to keep the scope limited.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedOrganizationUserAccountCommand.cs
```diff
@@ -1,247 +1,200 @@
...
             {
-                throw new BadRequestException("Organization must have at least one confirmed owner.");
-            }
+                User = user,
+                OrganizationUserId = orgUserId,
+                OrganizationUser = orgUser,
+                IsClaimed = isClaimed,
+                OrganizationId = organizationId,
+                DeletingUserId = deletingUserId,
+            });
```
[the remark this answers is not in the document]
JimmyVo16: I'm curious what the advantages of `yield return` are here. From my understanding, it returns one item at a time. This is great for cases when you're iterating through a large data source and don't want to retrieve all the data at once, or in cases where you might not need to process everything, just enough to satisfy the request.

I don't think our use case fits either of those scenarios, since we essentially get all the data from the database in a single request.

src/Api/Utilities/CommandResultExtensions.cs
JimmyVo16: We discussed this over a call, but I just want to add it here for posterity.

We should remove the array of errors since I don't believe any of our code is currently using it. Another issue is that our client code (at least some) currently only handles a single string for the error message if we pass back an array, it will break the client. I confirmed this with the client that is using this endpoint.

src/Core/AdminConsole/Shared/Validation/ValidationResult.cs
```diff
@@ -16,6 +16,13 @@ public Valid(T Value)
     public T Value { get; init; }
 }
 
+public record PartialValidationResult<T>
+{
+    public List<Invalid<T>> InvalidResults { get; init; }
+
+    public List<Valid<T>> ValidResults { get; init; }
+}
```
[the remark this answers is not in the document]
JimmyVo16: I think I went with this because Jared preferred passing back a single object. I don't have a strong opinion on it, but I will say that being forced to map things back and forth can be confusing if you're not used to it.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IDeleteClaimedOrganizationUserAccountCommand.cs
```diff
@@ -1,19 +1,22 @@
...
+    Task<CommandResult<DeleteUserResponse>> DeleteUserAsync(Guid organizationId, Guid organizationUserId, Guid deletingUserId);
 
     /// <summary>
     /// Removes multiple users from an organization and deletes all of their associated user data.
     /// </summary>
     /// <returns>
     /// An error message for each user that could not be removed, otherwise null.
     /// </returns>
-    Task<IEnumerable<(Guid OrganizationUserId, string? ErrorMessage)>> DeleteManyUsersAsync(Guid organizationId, IEnumerable<Guid> orgUserIds, Guid? deletingUserId);
+    Task<Partial<DeleteUserResponse>> DeleteManyUsersAsync(Guid organizationId, IEnumerable<Guid> orgUserIds, Guid deletingUserId);
```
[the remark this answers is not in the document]
JimmyVo16: Yeah, I agree.

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedOrganizationUserAccountValidatorTests.cs
```diff
@@ -0,0 +1,386 @@
...
+public class DeleteClaimedOrganizationUserAccountValidatorTests
+{
+    [Theory]
+    [BitAutoData]
+    public async Task ValidateAsync_ShouldExecuteSyncValidatorsFirst(
+        Guid organizationId,
+        SutProvider<DeleteClaimedOrganizationUserAccountValidator> sutProvider)
+    {
+        // Arrange
+        var request = new DeleteUserValidationRequest
```
JimmyVo16, line 26: Hey @eliykat , since I'm going on leave, I'm going to try my best to update the tests and address comments that aren’t related to the command result pattern. I can’t address the command result changes until we align as a team on how we want to move forward, so you’ll probably need to update the tests, but hopefully it’ll just be small typing issues.

eliykat: Sounds good to me!

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -578,20 +581,28 @@ public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRemo
...
+    public async Task<IActionResult> DeleteAccount(Guid orgId, Guid id)
     {
         if (!await _currentContext.ManageUsers(orgId))
         {
-            throw new NotFoundException();
+            return NotFound();
         }
 
-        var currentUser = await _userService.GetUserByPrincipalAsync(User);
+        var currentUser = _userService.GetProperUserId(User); ;
```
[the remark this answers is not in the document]
eliykat: Done

JimmyVo16:
Since there are numerous architectural decisions, it will be easier to start a fresh PR.

PR #5374, opened 2025-02-06
[PM-18028] Attempting to enable automatic tax on customer with invalid location

jonashendrickx:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18028

## 📔 Objective

Enabling automatic tax for customers without a country set or with tax rates causes issues.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Billing/Extensions/CustomerExtensions.cs
+++ b/src/Core/Billing/Extensions/CustomerExtensions.cs
@@ -1,0 +1,16 @@
+﻿using Bit.Core.Billing.Constants;
+using Stripe;
+
+namespace Bit.Core.Billing.Extensions;
+
+public static class CustomerExtensions
+{
+
+    /// <summary>
+    /// Determines if a Stripe customer supports automatic tax
+    /// </summary>
+    /// <param name="customer"></param>
+    /// <returns></returns>
+    public static bool HasTaxLocationVerified(this Customer customer) =>
+        customer?.Tax?.AutomaticTax == StripeConstants.AutomaticTaxStatus.Supported;
+}
--- a/src/Core/Billing/Extensions/SubscriptionCreateOptionsExtensions.cs
+++ b/src/Core/Billing/Extensions/SubscriptionCreateOptionsExtensions.cs
@@ -1,0 +1,26 @@
+﻿using Stripe;
+
+namespace Bit.Core.Billing.Extensions;
+
+public static class SubscriptionCreateOptionsExtensions
+{
+    /// <summary>
+    /// Attempts to enable automatic tax for given new subscription options.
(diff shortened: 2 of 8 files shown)


amorask-bitwarden (approved):

src/Core/Billing/Extensions/SubscriptionCreateOptionsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <returns>Returns true when successful, false when conditions are not met.</returns>
+    public static bool EnableAutomaticTax(this SubscriptionCreateOptions options, Customer customer)
+    {
+        // We might only need to check the automatic tax status.
+        if (!customer.HasTaxLocationVerified() && string.IsNullOrWhiteSpace(customer.Address?.Country))
+        {
+            return false;
+        }
+
+        options.DefaultTaxRates = [];
```
cturnbull-bitwarden, line 21: :pick: 
In general, I would opt for immutable operations over mutable ones like this. 

Not to say that there's never a time we should mutate input params. While this encapsulates the logic to ensure that `SubscriptionCreateOptions` always has its `DefaultTaxRates` set to `[]`, which is a good thing, it also introduces a side effect that is hidden from the caller, namely conditionally setting `DefaultTaxRates` and `AutomaticTax` on the input parameter `options` by reference.

Without the engineer knowing ahead of time that this side effect exists, it would be reasonable to assume that they might try to also set these values elsewhere depending on the return value of this method. This also introduces some coupling given that this side effect behavior is tightly governed by the external state, which we could argue makes it a bit harder to decouple later on should we choose.

@amorask-bitwarden has done a good job of setting up a reusable pattern to borrow from, by creating `.From` extensions to handle this kind of immutable mapping logic. Take [OrganizationMetadataResponse](https://github.com/bitwarden/server/blob/main/src/Api/Billing/Models/Responses/OrganizationMetadataResponse.cs#L17) as an example.

Again, this is a nit, just wanted to leave it as a thought

src/Core/Billing/Extensions/CustomerExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public static class CustomerExtensions
+{
+
+    /// <summary>
+    /// Determines if a Stripe customer supports automatic tax
+    /// </summary>
+    /// <param name="customer"></param>
+    /// <returns></returns>
+    public static bool HasTaxLocationVerified(this Customer customer) =>
```
cturnbull-bitwarden, line 14: :tada: nice, I like this

PR #5382, opened 2025-02-06
[PM-17562] Add Azure Service Bus for Distributed Events

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This PR builds on the previous RabbitMQ implementation (PR #5357) of AMQP and adds an implementation using Azure Service Bus as the transport layer for Events. If configured, this replaces Azure Storage Queues - the EventService writes to Azure Service Bus where it is later picked up by a handler that writes the Event to Azure Table Storage.

**Note: The intended behavior is to change nothing in the existing setup unless it's explicitly configured to use the new Azure Service Bus setup. Please review/test with that in mind and double check this PR is, in fact, safely respecting that goal.** 

There is a [companion PR](https://github.com/bitwarden/contributing-docs/pull/540) in the contributing-docs repo that updates the Events page to match this new implementation. 

### Testing Locally
This PR adds a docker compose entry for running the Azure Service Bus Emulator locally and a config file to get it setup correctly. 

1. Make sure you have Azurite setup locally (as [per the normal instructions](https://contributing.bitwarden.com/getting-started/server/guide#azurite) for writing Events to Azure Table Storage). In addition, this assumes you're using the `mssql` default profile and have the `${MSSQL_PASSWORD}` set via `.env`.
2. Run docker compose to add/start the local emulator:

``` bash
docker compose --profile servicebus up -d
```

  - Note: The service bus emulator waits 15 seconds before starting. You can check the console in Docker desktop to make sure it's up before launching the server. 

3. Add the following to your `sercrets.json` in `dev` to configure the service bus:
``` json
	"eventLogging": {  
	  "azureServiceBus": {  
		"connectionString": "\"Endpoint=sb://localhost;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;\"",  
		"topicName": "event-logging",
		"eventRepositorySubscriptionName": "events-write-subscription",  
		"httpPostSubscriptionName": "events-httpPost-subscription"
	  },  
	  "httpPostUrl": "<Optional URL here>"  
	},
```
  - The `httpPostUrl` and `httpPostSubscriptionName` configure the optional HTTP POST handler. As mentioned on[ the original PR](https://github.com/bitwarden/server/pull/5323) this is just a way of showing the fan-out nature of distributed events. 
 
4. Re-run the secrets script to publish the new secrets
   
``` bash
pwsh setup_secrets.ps1 -clear
```

5. Start or re-start all services, including `EventsProcessor`.

### Testing with Azure
When the time comes to test this in a Dev/QA setup that involves the actual Azure configuration, we will need the following steps:
1. Configure Azure Service Bus to add a new Topic ("event-logging" is the one I used above, but this is configurable)
2. Configure that Topic with a subscription for the Azure Table Storage handler (default is "events-write-subscription" but this is configurable).
3. Match the configuration in secrets.json to add these properties and the correct connection string:
``` json
	"eventLogging": {  
	  "azureServiceBus": {  
		"connectionString": "<Connection String>",  
		"topicName": "event-logging",
		"eventRepositorySubscriptionName": "events-write-subscription",  
	  },  
	},
```
  - Note: The HTTP POST handler is not needed and can be omitted.
 4. Verify that `Events.ConnectionString` setting is correctly set.
 5. Deploy with these settings in place.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


dev/docker-compose.yml
```diff
@@ -109,6 +109,21 @@ services:
     profiles:
       - proxy
 
+  servicebusemulator:
```
withinfocus, line 112: 🌱 I am thinking this should be just `service-bus` or `asb`. We don't mention emulation in other profile names.

dev/docker-compose.yml
```diff
@@ -109,6 +109,21 @@ services:
...
+    container_name: "servicebus-emulator"
+    image: mcr.microsoft.com/azure-messaging/servicebus-emulator:latest
+    pull_policy: always
+    volumes:
+      - "./servicebusemulator_config.json:/ServiceBus_Emulator/ConfigFiles/Config.json"
+    ports:
+      - "5672:5672"
+    environment:
+      SQL_SERVER: mssql
```
withinfocus, line 121: 💭 Wondering if we need to call out this coupling somewhere. Probably unnecessary as we expect MSSQL with development.

dev/servicebusemulator_config.json
```diff
@@ -0,0 +1,41 @@
...
+        ],
+        "Topics": [
+          {
+            "Name": "event-logging",
+            "Subscriptions": [
+              {
+                "Name": "events-write-subscription",
+              },
+              {
+                "Name": "events-httpPost-subscription",
```
withinfocus, line 30: ❓ We'll expect at least one listener if this is running, but should we set up two topics or comment this out perhaps? This would just fill up if the post URL wasn't configured right?

src/Core/AdminConsole/Services/Implementations/AzureServiceBusEventListenerService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            var eventMessage = JsonSerializer.Deserialize<EventMessage>(args.Message.Body.ToString());
+
+            await _handler.HandleEventAsync(eventMessage);
+            await args.CompleteMessageAsync(args.Message);
+        };
+
+        _processor.ProcessErrorAsync += args =>
+        {
+            _logger.LogError(args.Exception, "An error occured while processing a message.");
```
withinfocus, line 38: ❓ Can we get anything more out of the message to log more effectively here? Perhaps safely attempt deserialization and get access to the event?

src/Core/AdminConsole/Services/Implementations/AzureServiceBusEventListenerService.cs
```diff
@@ -0,0 +1,62 @@
...
+            return Task.CompletedTask;
+        };
+
+        await _processor.StartProcessingAsync(stoppingToken);
+
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            await Task.Delay(1_000, stoppingToken);
+        }
```
withinfocus, line 47: ❓ What's this for?

src/Core/Settings/GlobalSettings.cs
```diff
@@ -260,8 +260,31 @@ public string JobSchedulerConnectionString
...
+        public virtual string HttpPostUrl { get; set; }
         public RabbitMqSettings RabbitMq { get; set; } = new RabbitMqSettings();
 
+        public class AzureServiceBusSettings
+        {
+            private string _connectionString;
+            private string _topicName;
+
+            public virtual string EventRepositorySubscriptionName { get; set; } = "events-write-subscription";
+            public virtual string HttpPostSubscriptionName { get; set; } = "events-httpPost-subscription";
```
withinfocus, line 273:
```suggestion
            public virtual string HttpPostSubscriptionName { get; set; } = "events-http-post-subscription";
```

⛏️ but can you also update all the docs and wherever else this is used? This seems to be a more traditional pattern for the actual eventing.

src/EventsProcessor/Startup.cs
```diff
@@ -24,9 +27,37 @@ public void ConfigureServices(IServiceCollection services)
...
+            CoreHelpers.SettingHasValue(globalSettings.EventLogging.AzureServiceBus.TopicName))
+        {
+            services.AddSingleton<IEventRepository, TableStorageRepos.EventRepository>();
+            services.AddSingleton<AzureTableStorageEventHandler>();
+            services.AddKeyedSingleton<IEventWriteService, RepositoryEventWriteService>("persistent");
+            services.AddSingleton<IHostedService>(provider =>
+                new AzureServiceBusEventListenerService(
+                    provider.GetRequiredService<AzureTableStorageEventHandler>(),
+                    provider.GetRequiredService<ILogger<AzureServiceBusEventListenerService>>(),
+                    provider.GetRequiredService<GlobalSettings>(),
```
withinfocus, line 45: 🎨 Reuse `globalSettings`?

src/Events/Startup.cs
```diff
@@ -98,7 +98,7 @@ public void ConfigureServices(IServiceCollection services)
                     provider.GetRequiredService<GlobalSettings>(),
                     globalSettings.EventLogging.RabbitMq.EventRepositoryQueueName));
```
withinfocus, line 99: ❓ Somewhat-randomly-placed comment: should logic be here to explicitly allow just one writer? I don't want to have to time configuration changes in a special way for different services.

src/Core/Settings/GlobalSettings.cs
withinfocus, line 273: 🎨 Additionally, after talking with product maybe we call this HTTP POST stuff just "webhook"? We're gonna build some resiliency into that in a later phase e.g. retries and I am thinking it'll be more recognizable too. This could be a follow-up PR too.

withinfocus (changes requested):
Looking awesome! Few small things and some questions.

dev/docker-compose.yml
brant-livefront: I'm good with `service-bus` 👍  I changed the container name to match (in the next commit) as well.

dev/docker-compose.yml
brant-livefront: I do call out in the `contributing-docs` update that this assumes `mssql` is set up. Also, this is **only** for local dev. There are alternate ways in the ASB emulator's compose setups to have it declare its own DB, but I thought it's worth reusing `mssql` since we typically have it running.

dev/servicebusemulator_config.json
brant-livefront: That's a good point about filling up a topic. This is only in local dev (this is for the emulator only - in Azure we'd set up these Topics ahead of time). But it might be worth mentioning in the `contributing-docs` that they should remove this line if they are not planning on using the POST service.

src/Events/Startup.cs
brant-livefront: Thanks for the chat in Slack clarifying this. I think we're good now. 👍

src/Core/Settings/GlobalSettings.cs
brant-livefront: I like this idea and with the suggested queue/subscription rename above, I just went ahead and made the change in this PR. It seemed simpler to do it one time when I had all those configs in view.  👍

src/Core/AdminConsole/Services/Implementations/AzureServiceBusEventListenerService.cs
brant-livefront: Looked into this a bit more. We can be more defensive in the ProcessMessageAsync to catch and log errors more effectively there (which would previously have fallen to ProcessErrorAsync if uncaught). I've added a few additional properties to the error log as well.

src/Core/AdminConsole/Services/Implementations/AzureServiceBusEventListenerService.cs
```diff
@@ -23,28 +23,40 @@ public AzureServiceBusEventListenerService(
...
-            await _handler.HandleEventAsync(eventMessage);
-            await args.CompleteMessageAsync(args.Message);
+                await _handler.HandleEventAsync(eventMessage);
+                await args.CompleteMessageAsync(args.Message);
+            }
+            catch (Exception exception)
+            {
+                _logger.LogError(
+                    exception,
+                    "An error occured while processing message: {MessageId}, with Body: {Body}",
```
withinfocus, line 41: ⚠️ We likely should not log the body since it may have PII. Let's do just the message.

dev/servicebusemulator_config.json
withinfocus: Let's do the opposite, where this isn't here by default but is called out the docs. There's more to come with the webhook listener in time.

withinfocus (changes requested):
Minor couple things!

withinfocus (approved):
Looking good!

jrmccannon (approved):

PR #5398, opened 2025-02-12
[PM-16811] - SCIM Invite Users Optimizations

jrmccannon:
## 🎟️ Tracking
[PM-16811](https://bitwarden.atlassian.net/browse/PM-16811)

## 📔 Objective
The goal of this PR is to improve the Invite Organization User flow from the SCIM POST endpoint.  It will be later used to optimize the InviteOrganizationUser flow.

This was an attempt to surface all validation of inviting a user to the front of the call and to reduce duplicate database calls and validations.

This adds two a few new wrapper classes to allow for better flow control for when a request fails. By adding `ValidationResult` and a typed `CommandResult`, we can "fail" a request for different reasons and allow the caller to decide what to do with the failure.

Please refer to the diagram below of the new call stack.

## 📸 Diagram
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-16811]: https://bitwarden.atlassian.net/browse/PM-16811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/SendOrganizationInvitesCommand.cs
eliykat, line 1: Great work moving this into a separate class. It's definitely its own weird bespoke logic.

eliykat (changes requested):
Here are some initial thoughts, broadly limited to the controller, commands, and various models. TODO: review the validation, tests, and generic Result/Validation classes. For that reason some of my questions/thoughts may be answered by further review, but I wanted to get something checked in.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,239 @@
...
+            OccupiedPmSeats = await organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(request.Organization.OrganizationId),
+            OccupiedSmSeats = await organizationUserRepository.GetOccupiedSmSeatCountByOrganizationIdAsync(request.Organization.OrganizationId)
+        });
+
+        if (validationResult is Invalid<InviteUserOrganizationValidationRequest> invalid)
+        {
+            return new Failure<IEnumerable<OrganizationUser>>(invalid.ErrorMessageString);
+        }
+
+        var valid = validationResult as Valid<InviteUserOrganizationValidationRequest>;
```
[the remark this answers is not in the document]
jrmccannon: I've tried removing the explicit assignment, and it didn't work.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            return new Failure<IEnumerable<OrganizationUser>>(InviteOrganizationUsersErrorMessages.FailedToInviteUsers);
+        }
+
+        return new Success<IEnumerable<OrganizationUser>>(organizationUserCollection.Select(x => x.User));
+    }
+
+    private async Task RevertPasswordManagerChangesAsync(Valid<InviteUserOrganizationValidationRequest> valid, Organization organization)
+    {
+        if (valid.Value.PasswordManagerSubscriptionUpdate.SeatsRequiredToAdd < 0)
```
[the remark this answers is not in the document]
jrmccannon: It should indeed.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,239 @@
...
+            Organization = request.Organization,
+            PerformedBy = request.PerformedBy,
+            PerformedAt = request.PerformedAt,
+            OccupiedPmSeats = await organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(request.Organization.OrganizationId),
+            OccupiedSmSeats = await organizationUserRepository.GetOccupiedSmSeatCountByOrganizationIdAsync(request.Organization.OrganizationId)
+        });
+
+        if (validationResult is Invalid<InviteUserOrganizationValidationRequest> invalid)
+        {
+            return new Failure<IEnumerable<OrganizationUser>>(invalid.ErrorMessageString);
```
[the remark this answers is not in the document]
jrmccannon: Thinking about this what we could do is have a response model instead of just the datamodel. 

That might standardize our commands a bit more. Request models in, response models out.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,239 @@
...
+    IApplicationCacheService applicationCacheService,
+    IMailService mailService,
+    ILogger<InviteOrganizationUsersCommand> logger,
+    IUpdateSecretsManagerSubscriptionCommand updateSecretsManagerSubscriptionCommand,
+    ISendOrganizationInvitesCommand sendOrganizationInvitesCommand
+    ) : IInviteOrganizationUsersCommand
+{
+    public async Task<CommandResult<OrganizationUser>> InviteScimOrganizationUserAsync(InviteScimOrganizationUserRequest request)
+    {
+        var result = await InviteOrganizationUsersAsync(InviteOrganizationUsersRequest.Create(request));
```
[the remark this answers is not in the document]
jrmccannon: I'll rethink this a bit.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/InviteScimOrganizationUserRequest.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        Invite = invite;
+        Organization = organization;
+        PerformedAt = performedAt;
+        ExternalId = externalId;
+    }
+
+    public static InviteScimOrganizationUserRequest Create(OrganizationUserSingleEmailInvite invite,
+        OrganizationDto organization, DateTimeOffset performedAt, string externalId) =>
+        new(invite, organization, performedAt, externalId);
```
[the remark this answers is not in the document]
jrmccannon: I don't know why I did this on this model. I've done this before to put some initialization logic/validation of creating the model into a static factory method.


This one really wasn't needed.  I'll clean some of these up when I'm working through all the DTO cleanup.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
```diff
@@ -57,13 +50,34 @@ public async Task<OrganizationUserUserDetails> PostUserAsync(Guid organizationId
             throw new ConflictException();
         }
 
-        var organization = await _organizationRepository.GetByIdAsync(organizationId);
-        var hasStandaloneSecretsManager = await _paymentService.HasSecretsManagerStandalone(organization);
+        var organization = await organizationRepository.GetByIdAsync(organizationId);
+        var hasStandaloneSecretsManager = await paymentService.HasSecretsManagerStandalone(organization);
```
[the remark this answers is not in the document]
jrmccannon: If a company has secrets manager, every invite will be for both PM and SM?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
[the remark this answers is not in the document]
jrmccannon: The original model from the old service method had multiple email addresses on a single invite object. From the server prospective, I couldn't figure out why that was the case.

During the call, it would just split them all up into individual invites to send so maybe it was just a way to hold multiple configurations? :shrug:

bitwarden_license/test/Scim.IntegrationTest/Controllers/v2/UsersControllerTests.cs
```diff
@@ -276,9 +279,15 @@ public async Task GetList_SearchUserNameWithoutOptionalParameters_Success()
...
-    [Fact]
-    public async Task Post_Success()
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task Post_Success(bool isScimInviteUserOptimizationEnabled)
     {
+        _factory.SubstituteService((IFeatureService featureService)
+            => featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization)
+                .Returns(isScimInviteUserOptimizationEnabled));
```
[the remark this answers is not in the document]
jrmccannon: I found I could do it at the individual test by creating a new factory and reinitializing the db.

bitwarden_license/src/Scim/Models/ScimUserRequestModel.cs
```diff
@@ -25,6 +26,14 @@ public OrganizationUserInvite ToOrganizationUserInvite(ScimProviderType scimProv
         };
     }
 
+    public OrganizationUserSingleEmailInvite ToInvite(ScimProviderType scimProvider, bool hasSecretsManager) =>
+        OrganizationUserSingleEmailInvite.Create(
+            EmailForInvite(scimProvider),
+            [],
+            OrganizationUserType.User,
+            new Permissions(),
+            hasSecretsManager);
```
r-tome, line 35: I prefer to name the parameters, its easier to read
```suggestion
        OrganizationUserSingleEmailInvite.Create(
            email: EmailForInvite(scimProvider),
            accessibleCollections: [],
            type: OrganizationUserType.User,
            permissions: new Permissions(),
            accessSecretsManager: hasSecretsManager);
```

bitwarden_license/src/Scim/Users/PostUserCommand.cs
```diff
@@ -57,13 +50,34 @@ public async Task<OrganizationUserUserDetails> PostUserAsync(Guid organizationId
...
         invite.AccessSecretsManager = hasStandaloneSecretsManager;
 
-        var invitedOrgUser = await _organizationService.InviteUserAsync(organizationId, invitingUserId: null, EventSystemUser.SCIM,
+        if (featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization))
+        {
```
r-tome, line 59: Extract this code block into a private method

src/Core/AdminConsole/Models/Business/OrganizationDto.cs
```diff
@@ -0,0 +1,31 @@
...
+        new(organization.Id,
+            organization.UseCustomPermissions,
+            organization.Seats,
+            organization.MaxAutoscaleSeats,
+            organization.SmSeats,
+            organization.MaxAutoscaleSmSeats,
+            StaticStore.GetPlan(organization.PlanType),
+            organization.GatewayCustomerId,
+            organization.GatewaySubscriptionId,
+            organization.UseSecretsManager);
```
r-tome, line 30: Constructors are the standard C# way for creating objects so I think using a new constructor is more intuitive
```suggestion
    public OrganizationDto(Organization organization) 
        : this(
            OrganizationId: organization.Id,
            UseCustomPermissions: organization.UseCustomPermissions,
            Seats: organization.Seats,
            MaxAutoScaleSeats: organization.MaxAutoscaleSeats,
            SmSeats: organization.SmSeats,
            SmMaxAutoScaleSeats: organization.MaxAutoscaleSmSeats,
            Plan: StaticStore.GetPlan(organization.PlanType),
            GatewayCustomerId: organization.GatewayCustomerId,
            GatewaySubscriptionId: organization.GatewaySubscriptionId,
            UseSecretsManager: organization.UseSecretsManager)
    {
    }
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        if (result is Failure<IEnumerable<OrganizationUser>> failure)
```
r-tome, line 43: How about using a switch statement on the returned type? (quick doodle)
```
switch (result)
{
    case Failure<IEnumerable<OrganizationUser>> failure:
        return new Failure<OrganizationUser>(failure.ErrorMessage);
        
    case Success<IEnumerable<OrganizationUser>> success when success.Value.Any():
        var user = success.Value.First();
        await eventService.LogOrganizationUserEventsAsync([
            (user, EventType.OrganizationUser_Invited, EventSystemUser.SCIM, request.PerformedAt.UtcDateTime)
        ]);
        return new Success<OrganizationUser>(user);
        
    default:
        return new Success<OrganizationUser>(null);
}
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/InviteOrganizationUsersRequest.cs
```diff
@@ -0,0 +1,28 @@
...
+        this.Organization = Organization;
+        this.PerformedBy = PerformedBy;
+        this.PerformedAt = PerformedAt;
+    }
+
+    public static InviteOrganizationUsersRequest Create(InviteScimOrganizationUserRequest request) =>
+        new([OrganizationUserInvite.Create(request.Invite, request.ExternalId)],
+            request.Organization,
+            Guid.Empty,
+            request.PerformedAt);
```
r-tome, line 27: I may not be understanding why these static methods are better than constructors
```suggestion
    public InviteOrganizationUsersRequest(InviteScimOrganizationUserRequest request)
        : this(
            Invites: [OrganizationUserInvite.Create(request.Invite, request.ExternalId)],
            Organization: request.Organization,
            PerformedBy: Guid.Empty,
            PerformedAt: request.PerformedAt)
    {
    }
```

r-tome (commented):
Some first thoughts here, I'll continue reviewing this later

src/Core/AdminConsole/Models/Business/OrganizationDto.cs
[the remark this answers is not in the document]
jrmccannon: Its fine. It can still be a record though :D

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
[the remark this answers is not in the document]
jrmccannon: This is also so that each entry point into the command can have its own contract.  I think SingleEmail probably doesn't need to exist yet (it will for the single invite user flow).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/InviteOrganizationUsersRequest.cs
jrmccannon: I prefer the way the `Create` static method reads in code. Especially when working with `Func`. It also allows you to put some validation logic into the method, which I prefer over putting in the constructor.

I can switch over to constructors since that seems to be others' preference.

r-tome: That is fair, I can see why you don't want logic inside the constructor. But it's not very intuitive that to create an object you have to use the `Create` method over the constructor.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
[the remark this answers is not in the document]
eliykat: I think that makes total sense, particularly because the controller almost never needs the entire data model. At the moment we use data models for everything; here we're trying to be more intentional about it.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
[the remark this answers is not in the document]
eliykat: Not quite - it depends how they've signed up. I've sent you a DM with the background here because it's lengthy and confusing.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/ISendOrganizationInvitesCommand.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Bit.Core.AdminConsole.Entities;
+using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Models;
+using Bit.Core.Entities;
+using Bit.Core.Models.Mail;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers;
+
+public interface ISendOrganizationInvitesCommand
```
r-tome, line 8: Please add xmldoc here

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/IInviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Models;
+using Bit.Core.Models.Commands;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers;
+
+public interface IInviteOrganizationUsersCommand
```
r-tome, line 6: Xmldoc here please

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,250 @@
...
+
+            await RevertPasswordManagerChangesAsync(validatedRequest, organization);
+
+            return new Failure<IEnumerable<OrganizationUser>>(FailedToInviteUsers);
+        }
+
+        return new Success<IEnumerable<OrganizationUser>>(organizationUserCollection.Select(x => x.OrganizationUser));
+    }
+
+    private async Task RevertPasswordManagerChangesAsync(Valid<InviteUserOrganizationValidationRequest> valid, Organization organization)
```
r-tome, line 142: Keeping consistency on the name  helps readability. Other methods also
```suggestion
    private async Task RevertPasswordManagerChangesAsync(Valid<InviteUserOrganizationValidationRequest> validatedRequest, Organization organization)
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/CreateOrganizationUser.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Bit.Core.Entities;
+using Bit.Core.Enums;
+using Bit.Core.Models.Data;
+using Bit.Core.Utilities;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Models;
+
+public class CreateOrganizationUser
```
r-tome, line 8: Add some xmldoc here or maybe rename it to something that would indicate what this is used for.

r-tome (commented):
Overall it looks good to me. I mainly think it needs documentation because sometimes I felt lost at what I was looking at (big PR doesn't help). I'll review this again tomorrow

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/ISendOrganizationInvitesCommand.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    Task SendInvitesAsync(SendInvitesRequest request);
+
+    Task<OrganizationInvitesInfo> BuildOrganizationInvitesInfoAsync(IEnumerable<OrganizationUser> orgUsers, Organization organization, bool initOrganization = false);
```
eliykat, line 12: This is only used in [OrganizationService](https://github.com/bitwarden/server/blob/829e49c3980811d033da96184ed3ba65b748ab38/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1058-L1059), where it's immediately followed by a call to `mailService.SendOrganizationInviteEmailsAsync` - which is exactly what `ISendOrganizationInvitesCommand.SendInvitesAsync` does.

Could orgService call `ISendOrganizationInvitesCommand.SendInvitesAsync` instead, and make this one private?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,250 @@
...
+{
+
+    public const string IssueNotifyingOwnersOfSeatLimitReached = "Error encountered notifying organization owners of seat limit reached.";
+    public const string FailedToInviteUsers = "Failed to invite user(s).";
+    public const string NoUsersToInvite = "No users to invite.";
+    public const string InvalidResultType = "Invalid result type.";
+
+    public async Task<CommandResult<ScimInviteOrganizationUsersResponse>> InviteScimOrganizationUserAsync(OrganizationUserSingleEmailInvite request)
+    {
+        var hasSecretsManager = await paymentService.HasSecretsManagerStandalone(request.InviteOrganization);
```
eliykat, line 44: nit: should be `var hasSecretsManagerStandalone`. Standalone is a term of art here, it's different to just having SM.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,250 @@
...
+    {
+        var existingEmails = new HashSet<string>(await organizationUserRepository.SelectKnownEmailsAsync(
+                request.InviteOrganization.OrganizationId, request.Invites.SelectMany(i => i.Emails), false),
+            StringComparer.InvariantCultureIgnoreCase);
+
+        var invitesToSend = request.Invites
+            .SelectMany(invite => invite.Emails
+                .Where(email => !existingEmails.Contains(email))
+                .Select(email => OrganizationUserInviteDto.Create(email, invite, request.InviteOrganization.OrganizationId))
+            ).ToArray();
```
eliykat, line 82: Further to my previous comment - looks like we flatten out the users here anyway, so we may as well do it early.

Now that I'm on a jag, I want to say that `OrganizationUserInviteDto` is also the same as a single `OrganizationUserInvite` would be. Any thoughts on validating `OrganizationUserInvite` directly? (also solves your naming problem) Or tell me if I'm going overboard

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/OrganizationUserInviteDto.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Bit.Core.Enums;
+using Bit.Core.Models.Data;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Models;
+
+public class OrganizationUserInviteDto
+{
+    public string Email { get; private init; } = string.Empty;
+    public CollectionAccessSelection[] AccessibleCollections { get; private init; } = [];
```
eliykat, line 9: nit: collections can also be accessible via groups, which would not appear in this list (which contains direct assignments only).

`AssignedCollections` would be more accurate.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/Organization/InvitingUserOrganizationValidation.cs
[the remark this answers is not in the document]
eliykat: Also re: naming - `Inviting` is inconsistent with the naming convention elsewhere, I suggest it's `InviteUser<thing>Validation`.

eliykat (changes requested):
Great work, there's lots of feedback here but that's just because it's a big PR. Let me know if you want to discuss anything to agree on an approach before refactoring anything.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
eliykat, line 44: On the topic of this rather expensive call. This _is_ what we discussed in our last meeting, but one of the main goals here was to optimize this, and it still bothers me that this call is so early. I've seen a pattern where an organization can't autoscale seats for another reason (e.g. it has a provider), it'll spam our SCIM endpoints with potentially hundreds of invalid requests, and we're hitting Stripe for every single one.

What if we split validation into PM validation and SM validation, and then only made this call once the PM validation passed? That is breaking the abstraction a little (we're assuming it's not required for PM validation), and we'd want to scatter some comments around, but it is a special case and would solve the problem.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/SecretsManager/SecretsManagerSubscriptionUpdate.cs
```diff
@@ -0,0 +1,60 @@
...
+    public bool UseSecretsManger { get; }
+    public int? Seats { get; }
+    public int? MaxAutoScaleSeats { get; }
+    public int OccupiedSeats { get; }
+    public int AdditionalSeats { get; }
+    public int? PasswordManagerUpdatedSeatTotal { get; }
+    public Plan.SecretsManagerPlanFeatures SecretsManagerPlan { get; }
+    public int? AvailableSeats => Seats - OccupiedSeats;
+    public int SeatsRequiredToAdd => AdditionalSeats - AvailableSeats ?? 0;
+    public int? UpdatedSeatTotal => Seats + SeatsRequiredToAdd;
```
eliykat, line 19: Just in case we do end up using this (contra my comment above), note that this still has a bunch of un-updated PM references (rather than SM).

bitwarden_license/src/Scim/Users/PostUserCommand.cs
```diff
@@ -57,14 +58,63 @@ public async Task<OrganizationUserUserDetails> PostUserAsync(Guid organizationId
...
-        var hasStandaloneSecretsManager = await _paymentService.HasSecretsManagerStandalone(organization);
+        var organization = await organizationRepository.GetByIdAsync(organizationId);
+
+        if (organization == null)
+        {
+            throw new NotFoundException();
+        }
+
+
+        if (featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization))
```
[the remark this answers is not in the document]
jrmccannon: I can move it down into the SCIM specific method. With that change however, there would be no point to this PostUserCommand. Would we want to just remove it when we clean this up?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
jrmccannon: So the pattern I was trying to follow was from the old code flow in OrganizationService. A single invite entry contains a list of emails. We then flatten those down into separate Organization Users and they all get sent separate emails. 

Do you have any idea the use case for this?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/InviteOrganizationUserValidation.cs
```diff
@@ -0,0 +1,88 @@
...
+        return new Valid<InviteUserOrganizationValidationRequest>(new InviteUserOrganizationValidationRequest(
+            request,
+            subscriptionUpdate,
+            smSubscriptionUpdate));
+    }
+
+    public static ValidationResult<IGlobalSettings> ValidateEnvironment(IGlobalSettings globalSettings) =>
+        globalSettings.SelfHosted
+            ? new Invalid<IGlobalSettings>(new CannotAutoScaleOnSelfHostError(globalSettings))
+            : new Valid<IGlobalSettings>(globalSettings);
```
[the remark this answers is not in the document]
jrmccannon: I'll roll this into the PM check then. There we'll be able to determine if they are auto-scaling.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/PasswordManager/PasswordManagerSubscriptionUpdate.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    /// Seats the organization has
+    /// </summary>
+    public int? Seats { get; }
+
+    public int? MaxAutoScaleSeats { get; }
+
+    public int OccupiedSeats { get; }
+
+    public int AdditionalSeats { get; }
```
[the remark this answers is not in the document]
jrmccannon: I updated the names and added some xml docs explaining.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
eliykat: The use case is that when you're inviting users from the web app, you can input 1 email or a comma-delimited list of emails and all those email addresses will receive identical invites (in terms of their role and collection/group assignment). So the "single invite containing a list of emails" is the pattern from the api layer. (Go to the web app to see what I mean.)

My point here is that there's nothing special about that structure, the api layer uses it but the core layer doesn't need to if it doesn't make sense. If we were writing an InviteUserCommand from first principles, ignoring the current api implementation, I think the most straightforward interface would be "here's a list of invites to send" where each invite relates to 1 member. That way you can ask it to send 1 invite, send many invites, whatever. You don't need to distinguish between a single email invite and a multi-email invite, you just have a sequence with 1 or more invites.

Alternatively, we can maintain the current interface where an invite can relate to multiple users - not a big deal if you disagree with my reasoning above. But I think the current `InviteOrganizationUsersRequest` combines both to poor effect - it's a sequence of invites each containing a sequence of emails, which is confusing and I don't think it's required by the app today.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
```diff
@@ -1,39 +1,96 @@
...
+        }
+
+        var plan = await pricingClient.GetPlan(organization.PlanType);
+
+        if (plan == null)
+        {
+            logger.LogError("Plan {planType} not found for organization {organizationId}",
+                organization.PlanType, organization.Id);
+            return null;
+        }
```
eliykat, line 64: Not sure returning `null` is right - we probably want to throw, this is an invalid state.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Errors/ErrorMapper.cs
eliykat, line 1: Invalid xmldoc comments in this file, see the annotations on the Github diff.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -39,11 +40,22 @@ ISendOrganizationInvitesCommand sendOrganizationInvitesCommand
...
     public async Task<CommandResult<ScimInviteOrganizationUsersResponse>> InviteScimOrganizationUserAsync(OrganizationUserSingleEmailInvite request)
     {
-        var hasSecretsManager = await paymentService.HasSecretsManagerStandalone(request.InviteOrganization);
+        var hasSecretsManagerStandalone = await paymentService.HasSecretsManagerStandalone(request.InviteOrganization);
+
+        var orgUsers = await organizationUserRepository.GetManyDetailsByOrganizationAsync(request.InviteOrganization.OrganizationId);
+
+        if (orgUsers.Any(existingUser =>
+                request.Email.Equals(existingUser.Email, StringComparison.InvariantCultureIgnoreCase) ||
+                request.ExternalId.Equals(existingUser.ExternalId, StringComparison.InvariantCultureIgnoreCase)))
```
eliykat, line 52: Do we want to allow for the linguistic adjustments of `InvariantCulture` when comparing emails? I would've expected a byte-for-byte comparison using `OrdinalIgnoreCase`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/PasswordManager/PasswordManagerSubscriptionUpdate.cs
```diff
@@ -7,45 +7,68 @@ namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUse
 public class PasswordManagerSubscriptionUpdate
 {
     /// <summary>
-    /// Seats the organization has
+    /// Seats the organization is currently paying for
     /// </summary>
     public int? Seats { get; }
```
eliykat, line 12: "paying for" is inaccurate, this is also used by free orgs.

eliykat (changes requested):

bitwarden_license/src/Scim/Users/PostUserCommand.cs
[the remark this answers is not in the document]
eliykat: Your latest changes look good to me - much clearer.

I was going to say that we don't need `PostUserCommand` anymore and that you could clean it up as suggested. However, the Scim project is Bitwarden Licensed, and Scim-specific logic needs to stay there as much as possible.  `PostUserCommand` is useful as a glue layer to do any SCIM specific stuff before we pass it into the shared core (AGPL) invite user logic.

While I like your latest changes I'm inclined to say move the existing user check back to the Scim project to be safe, even though it's like 6 lines of code. And it might be useful to keep the Scim command around for this purpose. Open to other ideas though.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,250 @@
...
+
+    private async Task RevertPasswordManagerChangesAsync(Valid<InviteUserOrganizationValidationRequest> validatedResult, Organization organization)
+    {
+        if (validatedResult.Value.PasswordManagerSubscriptionUpdate.SeatsRequiredToAdd > 0)
+        {
+            // When reverting seats, we have to tell payments service that the seats are going back down by what we attempted to add.
+            // However, this might lead to a problem if we don't actually update stripe but throw any ways.
+            // stripe could not be updated, and then we would decrement the number of seats in stripe accidentally.
+            var seatsToRemove = -validatedResult.Value.PasswordManagerSubscriptionUpdate.SeatsRequiredToAdd;
+            await paymentService.AdjustSeatsAsync(organization, validatedResult.Value.InviteOrganization.Plan, -seatsToRemove);
```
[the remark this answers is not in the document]
jrmccannon: Fixed the double negative.

I think to solve this, we would want Payment service call to be idempotent. Or we would need tot track if the update was done at all.

There could be race conditions if multiple requests were sent at once. But, that could also be a use case we don't actually have.

Some more thought should go into this but the two options I see would be

1. track if we actually changed the value and if we did, then revert (this could be done by capturing if the Task to update stripe was successful possibly)

2. Making the call idempotent and simply return it to what it was before the request started.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
jrmccannon: I'm not even sure what exception we would throw?

It's not a bad request. Is there a pattern that we have for alerting the user to something terrible state wise has happened?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
jrmccannon: I'll have to think about this a bit. The issue is that SCIM says if you have SM you add SM to all users.

Meanwhile, you can choose whether you want to add a user to SM through the other flows. 🤷 Even if we bury it farther down, we still need to know if you have SM at the beginning to determine if the user should be added to the SM seat count.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/SecretsManager/SecretsManagerSubscriptionUpdate.cs
jrmccannon: All the Seats stuff relates to the SM seat count separate from Password manager.

amorask-bitwarden (dismissed):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
eliykat: It would be something like:

1. check if PM subscription needs autoscaling -> validate PM subscription changes (doesn't rely on SM seat count)
2. if hasSecretsManagerStandalone -> update the request object to give everyone access to SM
3. check if SM subscription needs autoscaling -> validate SM subscription changes
4. do the invites (with the updated request object)

But this is why I say it's breaking the abstraction, you have to interpose it in the middle of the validation.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
eliykat: Maybe a `NotFoundException`, because the plan wasn't found? Or just an `Exception`. I think it should be surfaced back to the user that something's gone wrong, even if it's not their fault.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/CreateOrganizationUser.cs
eliykat: xmldoc still required

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/GlobalSettings/EnvironmentValidator.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Bit.Core.AdminConsole.Shared.Validation;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Validation.GlobalSettings;
+
+public interface IEnvironmentValidator : IValidator<EnvironmentRequest>;
+
+public class EnvironmentValidator : IEnvironmentValidator
+{
+    public async Task<ValidationResult<EnvironmentRequest>> ValidateAsync(EnvironmentRequest value) =>
```
eliykat, line 9: See the many annotations here in the Github diff:

> This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.

I think that `async` is a convenient way to wrap the return result in a Task, but I guess the compiler disagrees. You may have to do it explicitly.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/InviteOrganizationUserValidator.cs
```diff
@@ -0,0 +1,64 @@
...
+        InviteUserOrganizationValidationRequest request,
+        PasswordManagerSubscriptionUpdate subscriptionUpdate)
+    {
+        try
+        {
+            var smSubscriptionUpdate = new SecretsManagerSubscriptionUpdate(
+                    organization: await organizationRepository.GetByIdAsync(request.InviteOrganization.OrganizationId),
+                    plan: request.InviteOrganization.Plan,
+                    autoscaling: true)
+                .AdjustSeats(request.Invites.Count(x => x.AccessSecretsManager));
```
eliykat, line 49: This doesn't take into account any spare seats that the org may have. e.g. The org may be have 5 SM Seats in their sub, but only 4 users have access to SM, leaving 1 seat available. The invited user would take that seat without autoscaling.

You need to get the number of seats required: see [existing code](https://github.com/bitwarden/server/blob/46d36b1ef8a15b7c27654fe46ffa6aaa8c4b8e86/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L865-L872) for reference. You were doing this with `InviteUserOrganizationValidationRequest.OccupiedSmSeats` but it got dropped somewhere I think.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/InviteOrganizationUserValidator.cs
```diff
@@ -0,0 +1,64 @@
...
+        var subscriptionUpdate = new PasswordManagerSubscriptionUpdate(request);
+
+        var passwordManagerValidationResult = await passwordManagerInviteUserValidator.ValidateAsync(subscriptionUpdate);
+
+        if (passwordManagerValidationResult is Invalid<PasswordManagerSubscriptionUpdate> invalidSubscriptionUpdate)
+        {
+            return invalidSubscriptionUpdate.Map(request);
+        }
+
+        if (request.InviteOrganization.UseSecretsManager && request.Invites.Any(x => x.AccessSecretsManager))
```
eliykat, line 28: This is where you could make the `hasSecretsManagerStandalone` stripe call, per my previous [comment](https://github.com/bitwarden/server/pull/5398#discussion_r2015588675). I don't think it's needed before now, so you could do it here immediately before the SM validation.

I don't like mutating the request in the validator, but it seems like our only option for deferring this call. It's not a perfect solution though, what do you think?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/SecretsManager/SecretsManagerInviteUserValidation.cs
eliykat, line 1: I think this and the other files in this folder can be deleted now as we're relying on the existing SM command validation.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/Organization/InviteUserOrganizationValidator.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Bit.Core.AdminConsole.Models.Business;
+using Bit.Core.AdminConsole.Shared.Validation;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Validation.Organization;
+
+public interface IInviteUserOrganizationValidator : IValidator<InviteOrganization>;
+
+public class InviteUserOrganizationValidator : IInviteUserOrganizationValidator
+{
+    public async Task<ValidationResult<InviteOrganization>> ValidateAsync(InviteOrganization inviteOrganization)
```
eliykat, line 10: Same here re: async vs. `Task.FromResult`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/PasswordManager/PasswordManagerInviteUserValidator.cs
eliykat, line 1: This is looking better now that all PM validation is located here and it's expressly predicated on adjusting the subscription.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/PasswordManager/PasswordManagerSubscriptionUpdate.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    public int AdditionalSeats { get; }
+
+    public int? AvailableSeats => Seats - OccupiedSeats;
+
+    public int SeatsRequiredToAdd => AdditionalSeats - AvailableSeats ?? 0;
+
+    public int? UpdatedSeatTotal => Seats + SeatsRequiredToAdd;
+
+    public bool MaxSeatsReached => UpdatedSeatTotal.HasValue && MaxAutoScaleSeats.HasValue && UpdatedSeatTotal.Value == MaxAutoScaleSeats.Value;
```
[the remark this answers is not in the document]
eliykat: Unaddressed comment: https://github.com/bitwarden/server/pull/5398/files#r2009511713

src/Core/OrganizationFeatures/OrganizationServiceCollectionExtensions.cs
```diff
@@ -171,6 +176,14 @@ private static void AddOrganizationUserCommandsQueries(this IServiceCollection s
...
         services.AddScoped<IAuthorizationHandler, OrganizationUserUserDetailsAuthorizationHandler>();
         services.AddScoped<IHasConfirmedOwnersExceptQuery, HasConfirmedOwnersExceptQuery>();
+
+        services.AddScoped<IInviteOrganizationUsersCommand, InviteOrganizationUsersCommand>();
+        services.AddScoped<ISendOrganizationInvitesCommand, SendOrganizationInvitesCommand>();
+
+        services.AddScoped<IInviteUsersValidator, InviteUsersValidator>();
+        services.AddScoped<IInviteUserOrganizationValidator, InviteUserOrganizationValidator>();
+        services.AddScoped<IPasswordManagerInviteUserValidator, PasswordManagerInviteUserValidator>();
+        services.AddScoped<IEnvironmentValidator, EnvironmentValidator>();
```
eliykat, line 186: Suggested naming now that these are in global DI - it's particularly important that they're clearly namespaced:

- `IInviteUsersValidator`
- `IInviteUsersOrganizationValidator`
- `IInviteUsersPasswordManagerValidator`
- `IInviteUsersEnvironmentValidator`

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/PasswordManager/PasswordManagerInviteUserValidation.cs
```diff
@@ -0,0 +1,42 @@
...
+            return new Valid<PasswordManagerSubscriptionUpdate>(subscriptionUpdate);
+        }
+
+        if (subscriptionUpdate.AdditionalSeats == 0)
+        {
+            return new Valid<PasswordManagerSubscriptionUpdate>(subscriptionUpdate);
+        }
+
+        if (subscriptionUpdate.UpdatedSeatTotal is not null && subscriptionUpdate.MaxAutoScaleSeats is not null &&
+            subscriptionUpdate.UpdatedSeatTotal > subscriptionUpdate.MaxAutoScaleSeats)
```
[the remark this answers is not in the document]
eliykat: Unaddressed comment: https://github.com/bitwarden/server/pull/5398#discussion_r2009512782

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/PasswordManager/PasswordManagerInviteUserValidation.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        if (subscriptionUpdate.PasswordManagerPlan.HasAdditionalSeatsOption is false)
+        {
+            return new Invalid<PasswordManagerSubscriptionUpdate>(
+                new PasswordManagerPlanDoesNotAllowAdditionalSeatsError(subscriptionUpdate));
+        }
+
+        // Apparently MaxAdditionalSeats is never set. Can probably be removed.
+        if (subscriptionUpdate.AdditionalSeats > subscriptionUpdate.PasswordManagerPlan.MaxAdditionalSeats)
```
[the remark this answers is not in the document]
eliykat: Unaddressed comment: https://github.com/bitwarden/server/pull/5398#discussion_r2009518422

eliykat (changes requested):
Looking better and better. I think the various DTO models have been nicely tidied up, and the validation is now clearly grouped into PM & SM and should be working properly.

I think the overall structure is now right, my comments below are relatively low level, or following up on previous comments that haven't been addressed. I've marked all addressed comments as resolved, please double check anything left open.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -168,16 +164,19 @@ private async Task RevertPasswordManagerChangesAsync(Valid<InviteUserOrganizatio
         }
     }
 
-    private async Task RevertSecretsManagerChangesAsync(Valid<InviteUserOrganizationValidationRequest> validatedResult, Organization organization)
+    private async Task RevertSecretsManagerChangesAsync(Valid<InviteUserOrganizationValidationRequest> validatedResult, Organization organization, int? initialSmSeats)
     {
-        if (validatedResult.Value.SecretsManagerSubscriptionUpdate.SeatsRequiredToAdd < 0)
+        if (validatedResult.Value.InviteOrganization.UseSecretsManager && validatedResult.Value.SecretsManagerSubscriptionUpdate.SmSeatsChanged)
```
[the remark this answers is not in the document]
jrmccannon: Made both consistent. Since the validation result does not return a secrets manager update when org does not have secrets manager, there is nothing to do. 

This could be made to be more explicit.

r-tome (changes requested):
This was a lot to review! Its looking pretty good to me but here are some thoughts

src/Core/AdminConsole/Shared/Validation/ValidationResult.cs
```diff
@@ -6,10 +6,39 @@ public abstract record ValidationResult<T>;
...
+
     public T Value { get; init; }
 }
 
 public record Invalid<T> : ValidationResult<T>
 {
-    public IEnumerable<Error<T>> Errors { get; init; }
+    public IEnumerable<Error<T>> Errors { get; init; } = [];
+
+    public string ErrorMessageString => string.Join(" ", Errors.Select(e => e.Message));
```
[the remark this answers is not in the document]
jrmccannon: Currently, no. But, could with larger bulk endpoints.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
```diff
@@ -1,39 +1,100 @@
...
+            throw new ConflictException("User already exists.");
+        }
+
+        var result = await inviteOrganizationUsersCommand.InviteScimOrganizationUserAsync(request);
+
+        var invitedOrganizationUserId = result switch
+        {
+            Success<ScimInviteOrganizationUsersResponse> success => success.Value.InvitedUser.Id,
+            Failure<ScimInviteOrganizationUsersResponse> { ErrorMessage: InviteOrganizationUsersCommand.NoUsersToInvite } => (Guid?)null,
+            Failure<ScimInviteOrganizationUsersResponse> failure when failure.Errors.Length != 0 => throw MapToBitException(failure.Errors),
```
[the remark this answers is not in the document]
eliykat: I agree that I'm not sure about the distinction between `Errors` vs. `ErrorMessages`. I assume that `ErrorMessages` was from the initial implementation before you introduced `Error` and it might be deprecated? But then it would be good to encapsulate that in a getter.

src/Core/Utilities/StrictEmailAddressAttribute.cs
```diff
@@ -12,39 +10,8 @@ public StrictEmailAddressAttribute()
...
-        // * Must end in a letter (including unicode)
-        // See the unit tests for examples of what is allowed.
-        var emailFormat = @"^[\x00-\x7F]+@.+\.\p{L}+$";
-        if (!Regex.IsMatch(emailAddress, emailFormat))
-        {
-            return false;
-        }
-
-        return new EmailAddressAttribute().IsValid(emailAddress);
+        return emailAddress.IsValidEmail() && new EmailAddressAttribute().IsValid(emailAddress);
```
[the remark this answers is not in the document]
eliykat: My main concern here was that the logic should be consistent whether you use the attribute or the extension method. Because I thought that the `EmailAddressAttribute` actually, you know, did something. But I agree it's totally redundant when used in combination with our stricter checks. This is fine then.

I've made a note to remove `EmailAddressAttribute` altogether - but that's a separate PR.

This is resolved.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/InviteOrganizationUserValidator.cs
```diff
@@ -60,5 +79,8 @@ private async Task<ValidationResult<InviteUserOrganizationValidationRequest>> Va
         {
             return new Invalid<InviteUserOrganizationValidationRequest>(new Error<InviteUserOrganizationValidationRequest>(ex.Message, request));
         }
+
+        int GetSecretManagerSeatAdjustment(int occupiedSeats, InviteOrganization organization, int invitesToSend) =>
+            organization.SmSeats - (occupiedSeats + invitesToSend) ?? 0;
```
eliykat, line 84: I still don't think this is right. For example:

- I have a total of 10 SM seats
- subtract 9 occupied (1 seat remaining)
- add 3 new users with SM access

... would be 4 seats to add. But I'm only adding 3 users and I have 1 seat free, it should be 2.

I think it should be: `Math.Abs(organization.SmSeats - occupiedSeats - invitesToSend ?? 0)`.

Please add (or update) a test for this.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
eliykat: The stripe call looks good to me, thanks for iterating on it!

eliykat (changes requested):

r-tome (approved):

rkac-bw (approved):
lgtm

amorask-bitwarden (approved):

PR #5400, opened 2025-02-12
[PM-17562] Refactor to Support Multiple Message Payloads

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

The previous PRs (#5357 #5382) for distributed events handled each event as a separate AMQP message. That is, if you called `CreateManyAsync` it would loop over the events and send each one as a separate message. While this keeps the implementation simple (one event per message, message are always of the same type, etc), it could result in far too many messages in practice and possibly incur larger costs with Azure Service Bus.

This PR refactors the two IEventWriteService AMQP implementations to handle `CreateManyAsync` and send the full array of events as a single message. It also refactors the two listener services to correctly account for if a message was an array of events or a single event. Lastly, it adds a `Many` variant to the IEventMessageHandler` interface and implementations to handle receiving an array instead of a single event.

Since arrays are also easily serializable to JSON, we're still using JSON as the message payload - we just have to account for that when deserializing. All of the documentation/configuration remains the same.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Services/IEventMessageHandler.cs
+++ b/src/Core/AdminConsole/Services/IEventMessageHandler.cs
@@ -5,4 +5,6 @@ namespace Bit.Core.Services;
 public interface IEventMessageHandler
 {
     Task HandleEventAsync(EventMessage eventMessage);
+
+    Task HandleManyEventAsync(IEnumerable<EventMessage> eventMessages);
 }
--- a/src/Core/AdminConsole/Services/IEventMessageHandler.cs
+++ b/src/Core/AdminConsole/Services/IEventMessageHandler.cs
@@ -6,5 +6,5 @@ public interface IEventMessageHandler
 {
     Task HandleEventAsync(EventMessage eventMessage);
 
-    Task HandleManyEventAsync(IEnumerable<EventMessage> eventMessages);
+    Task HandleManyEventsAsync(IEnumerable<EventMessage> eventMessages);
 }
--- a/src/Core/AdminConsole/Services/Implementations/WebhookEventHandler.cs
+++ b/src/Core/AdminConsole/Services/Implementations/WebhookEventHandler.cs
@@ -4,24 +4,26 @@ using Bit.Core.Settings;
 
 namespace Bit.Core.Services;
 
-public class WebhookEventHandler : IEventMessageHandler
+public class WebhookEventHandler(
+    IHttpClientFactory httpClientFactory,
+    GlobalSettings globalSettings)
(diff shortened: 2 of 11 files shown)


r-tome (approved):
Great work!

src/Core/AdminConsole/Services/IEventMessageHandler.cs
```diff
@@ -5,4 +5,6 @@ namespace Bit.Core.Services;
 public interface IEventMessageHandler
 {
     Task HandleEventAsync(EventMessage eventMessage);
+
+    Task HandleManyEventAsync(IEnumerable<EventMessage> eventMessages);
```
withinfocus, line 9:
```suggestion
    Task HandleManyEventsAsync(IEnumerable<EventMessage> eventMessages);
```

🎨 Also check on multiple enumerations here -- sometimes it's more practical to just require a `List<>`.

src/Core/AdminConsole/Services/Implementations/WebhookEventHandler.cs
```diff
@@ -4,24 +4,26 @@
...
 
-public class WebhookEventHandler : IEventMessageHandler
+public class WebhookEventHandler(
+    IHttpClientFactory httpClientFactory,
+    GlobalSettings globalSettings)
+    : IEventMessageHandler
 {
-    private readonly HttpClient _httpClient;
-    private readonly string _webhookUrl;
+    private readonly HttpClient _httpClient = httpClientFactory.CreateClient(HttpClientName);
```
withinfocus, line 12: ❓ I forget the best pattern -- should this just inject a client directly?

brant-livefront: I think this is the right one - we hashed it out on the first PR per [Justin's suggestion](https://github.com/bitwarden/server/pull/5323#discussion_r1929225613). `HttpClient` doesn't have an interface, so instead we inject `IHttpClientFactory` so we can easily swap that out in tests/DI. And then we create a named HttpClient instance in startup.

src/Core/AdminConsole/Services/IEventMessageHandler.cs
brant-livefront: Made this change to the interface. I think we should be good on multiple enumerations - each handler should be able to do its work in a single pass.

withinfocus (approved):

r-tome (approved):

PR #5407, opened 2025-02-14
[PM-18086] Add CanRestore and CanDelete authorization methods.

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18086

## 📔 Objective

1. Add CanRestore and CanDelete authorization methods.
2. Add test coverage for both methods.

Note: This ticket does not include wiring up these methods.

## 📸 Screenshots

No QA since the methods are not being used.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


gbubemismith (approved):
This looks good

eliykat (dismissed):

src/Core/Vault/Authorization/Permissions/NormalCipherPermissions.cs
```diff
@@ -0,0 +1,28 @@
+﻿
+using Bit.Core.Entities;
+using Bit.Core.Models.Data.Organizations;
+using Bit.Core.Vault.Models.Data;
+
+namespace Bit.Core.Vault.Authorization.Permissions;
+
+public class NormalCipherPermissions
+{
+    public static bool CanDelete(User user, CipherDetails cipherDetails, OrganizationAbility organizationAbility)
```
r-tome, line 10: Shouldn't `OrganizationAbility` be nullable? If the cipher belongs solely to the user then there's no `OrganizationAbility` to get.

src/Core/Vault/Authorization/Permissions/NormalCipherPermissions.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        if (user.Id == cipherDetails.UserId)
+        {
+            return true;
+        }
+
+        if (organizationAbility.LimitItemDeletion)
```
r-tome, line 17:
```suggestion
        if (organizationAbility is { LimitItemDeletion: true })
```

r-tome (approved):

r-tome:
@JimmyVo16, I'm not sure we'll use this implementation because I think it might be better to use a service instead. More info here https://github.com/bitwarden/server/pull/5434

r-tome (dismissed):
I'm good with merging this. You can ignore my previous comment because we're not using a service for this logic.

src/Core/Vault/Authorization/Permissions/NormalCipherPermissions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if (user.Id == cipherDetails.UserId)
+        {
+            return true;
+        }
+
+        if (organizationAbility == null)
+        {
+            throw new Exception("Cipher needs to belong to a user or an organization.");
+        }
```
eliykat, line 20: This error message doesn't describe the problem: the problem isn't the cipher, it's that no OrgAbility has been provided by the caller. You could rephrase the error message, e.g. "OrganizationAbility must not be null", or throw an `ArgumentNullException`, or if you want

eliykat (dismissed):

src/Core/Vault/Authorization/Permissions/NormalCipherPermissions.cs
JimmyVo16: Thinking about this more, I believe we're trying to solve two problems here.
- The cipher does not have an owner:
    1. The `cipherDetails` does not have a `userId` or `organizationId`
- The cipher does not belong to the `organizationAbility`:
    1. The client may pass in the wrong organization.
    2. The client may pass `null` for `organizationAbility`.
  
I updated the code and added tests to cover both cases. Let me know what you think.

JimmyVo16: I also noticed that it would probably be confusing to throw an exception for `organizationAbility` being null since it's a nullable parameter.

gbubemismith (approved):

PR #5409, opened 2025-02-17
[PM-18234] Add SendPolicyRequirement

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18234
## 📔 Objective

- Implement a `SendPolicyRequirement`, which summarizes the business requirements of our two Send-related policy types: `DisableSend` and `SendOptions`. This makes it easier to consume for calling code, which doesn't have to concern itself with handling Policy and data objects directly.
- Implement a factory function for the new requirement, which is responsible for **filtering** out policies that won't be enforced, and **combining** multiple policies into a single requirement.
- Connect the factory function to DI so it can be accessed via `IPolicyRequirementQuery` (already implemented).
- Update the calling code to use the new interface (feature flagged).

This is a new pattern that we are just starting to roll out, so if you're reviewing please let me know if you have any questions.

Note that generally there will be 1 `PolicyType` per requirement, however in this case they were closely related enough I thought it was a nice bonus to group them.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
```diff
@@ -0,0 +1,38 @@
...
+    public bool DisableSend { get; init; }
+    public bool DisableHideEmail { get; init; }
+
+    public static SendPolicyRequirement Create(IEnumerable<PolicyDetails> policyDetails)
+    {
+        var filteredPolicies = policyDetails
+            .ExemptRoles([OrganizationUserType.Owner, OrganizationUserType.Admin])
+            .ExemptStatus([OrganizationUserStatusType.Invited, OrganizationUserStatusType.Revoked])
+            .ExemptProviders()
+            .ToList();
```
audreyality, line 21: ❓ How does this affect the filtering already performed by the client? There's a very similar function in the service, IIRC, that I needed to update. Is that being migrated to the server?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
```diff
@@ -0,0 +1,38 @@
...
+                new SendPolicyRequirement
+                {
+                    // Set Disable Send requirement in the initial seed
+                    DisableSend = filteredPolicies.GetPolicyType(PolicyType.DisableSend).Any()
+                },
+                (result, data) => new SendPolicyRequirement
+                {
+                    DisableSend = result.DisableSend,
+                    DisableHideEmail = result.DisableHideEmail || data.DisableHideEmail
+                });
```
audreyality, line 36: 🎨 Consider assigning the result of this LINQ query to an intermediate variable. One-liners in return statements are neat, but a variable you can examine with a watch is neater.

src/Core/Tools/Services/Implementations/SendService.cs
```diff
@@ -308,6 +319,26 @@ private async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
         }
     }
 
+    private async Task ValidateUserCanSaveAsync_vNext(Guid? userId, Send send)
+    {
+        if (!userId.HasValue)
```
audreyality, line 324: 🤔 Should this branch also verify whether the user has any organizations, as happens on line 300? If not, why?

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirementTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
+using Bit.Core.Enums;
+using Bit.Core.Test.AdminConsole.AutoFixture;
+using Xunit;
+
+namespace Bit.Core.Test.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
+
+public class SendPolicyRequirementTests
+{
+    // DisableSend
```
audreyality, line 13: 🎨 Consider using a `#region` here instead of a comment. (This applies to other lines in this file, too.)

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirementTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    [Theory, AutoData]
+    public void DisableSend_IsFalse_IfNoDisableSendPolicies(
+        [PolicyDetails(PolicyType.RequireSso)] PolicyDetails otherPolicy1,
+        [PolicyDetails(PolicyType.SendOptions)] PolicyDetails otherPolicy2)
+    {
+        EnableDisableHideEmail(otherPolicy2);
+        var actual = SendPolicyRequirement.Create([otherPolicy1, otherPolicy2]);
+        Assert.False(actual.DisableSend);
```
audreyality, line 22: 🎨 Consider using whitespace to differentiate the arrange/act/assert patterns in these test cases. (No need to add comments. They're repetitive in a way that whitespace isn't.)

test/Core.Test/Tools/Services/SendServiceTests.cs
```diff
@@ -116,6 +119,90 @@ public async Task SaveSendAsync_DisableHideEmail_DoesntApply_success(SendType se
...
+    // Disable Send policy check - vNext
+    private void SaveSendAsync_Setup_vNext(SendType sendType, SutProvider<SendService> sutProvider, Send send,
+        SendPolicyRequirement sendPolicyRequirement = null)
+    {
+        send.Id = default;
+        send.Type = sendType;
+
+        sutProvider.GetDependency<IPolicyRequirementQuery>().GetAsync<SendPolicyRequirement>(send.UserId!.Value).Returns(
+            sendPolicyRequirement ?? new SendPolicyRequirement());
+        sutProvider.GetDependency<IFeatureService>().IsEnabled(FeatureFlagKeys.PolicyRequirements).Returns(true);
```
audreyality, line 131: 🤔 The use of a null default argument and nullish resolution seems like a code smell to me. You should avoid branching and implicit logic in your test. Consider passing `new SendPolicyRequirement()` at this method's callsite explicitly where necessary.

❓ Why is it necessary to default `send.Id` to `default`? Why isn't that a feature of your fixture instead?

❌ `send.Type` should always be set in the test. The compiler won't mind the extra line, and it'll be more clear to test readers what the test is doing. (Alternatively, you could set this up as part of your test fixture.)

audreyality (commented):
📝 I don't write much server code, and may be critiquing well-established patterns. If that's the case, just point it out and I'll resolve the feedback.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
eliykat: Great question! Today, client and server each have their own separate-but-identical implementations of this logic. This PR refactors the server implementation but doesn't change any functionality - so no corresponding client changes are required at this point in time.

However, in the future I intend to include these PolicyRequirement objects in sync data, so that clients can rely on the server to just give them the answer without having to reimplement this logic. That would make the client-side `PolicyService` basically just a wrapper around state management and nothing else.

src/Core/Tools/Services/Implementations/SendService.cs
eliykat: I removed it because I thought it was an unnecessary leaking of the abstraction. The Send domain shouldn't need to know "if there are no org claims in the JWT then there are no policies and we don't have to check policy requirements". It should just get the policy requirement and let the Policy domain handle it (which it does).

It does make the method slightly less efficient, as checking the JWT could shortcut the db call, but it's not a particularly hot path.

All of that said - I don't feel strongly about it and can restore it if you disagree.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.AdminConsole.Models.Data.Organizations.Policies;
+using Bit.Core.Enums;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
+
+/// <summary>
+/// Policy requirements for the Disable Send and Send Options policies.
+/// </summary>
```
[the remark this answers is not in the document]
eliykat: Done

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
eliykat: Ah see I love my one-liners 😁 you're right about debugging though - done.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirementTests.cs
eliykat: You're right to call me out on this, but for a slightly different reason. Our code style is [not to use region comments](https://contributing.bitwarden.com/contributing/code-style/#code-regions), so here I am... not using a region comment 😁 

I think these comments are probably not very useful anyway, so I've removed them.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirementTests.cs
eliykat: Addressed

test/Core.Test/Tools/Services/SendServiceTests.cs
eliykat: This has all been addressed. I opted to move the `send.Type` assignment back into the tests rather than messing further with fixtures as it's just the simplest and most obvious way to do this.

src/Core/Tools/Services/Implementations/SendService.cs
audreyality: I don't have a strong opinion either way. I think one of the things that lead me to wonder is that policies are based on orgs, but they're never used in the code. Since the DB can back into this information from a user, however, I think we're all good. :)

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendPolicyRequirement.cs
audreyality: Just as far as it goes, I'm not much a fan of one-liners, because they usually span multiple lines. I tolerate them, though, so long as they're not in a return. 😉 

This is an interesting case, though, since most of it is LINQ. I do tend to make an exception when an API is designed with fluency in mind.

test/Core.Test/Tools/Services/SendServiceTests.cs
```diff
@@ -116,6 +119,93 @@ public async Task SaveSendAsync_DisableHideEmail_DoesntApply_success(SendType se
...
+    private void SaveSendAsync_Setup_vNext(SutProvider<SendService> sutProvider, Send send,
+        SendPolicyRequirement sendPolicyRequirement)
+    {
+        sutProvider.GetDependency<IPolicyRequirementQuery>().GetAsync<SendPolicyRequirement>(send.UserId!.Value)
+            .Returns(sendPolicyRequirement);
+        sutProvider.GetDependency<IFeatureService>().IsEnabled(FeatureFlagKeys.PolicyRequirements).Returns(true);
+
+        // Should not be called in these tests
+        sutProvider.GetDependency<IPolicyService>().AnyPoliciesApplicableToUserAsync(
+            Arg.Any<Guid>(), Arg.Any<PolicyType>()).ThrowsAsync<Exception>();
```
audreyality, line 132: 👍🏻 I love when a test causes a method that shouldn't be called to throw. It's such an elegant way to test the invariant.

audreyality (approved):
👍🏻 Looks good to me! Thank you for responding to my comments!

BTreston (approved):

PR #5425, opened 2025-02-19
Ac/pm 17804

BTreston:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
+++ b/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
@@ -46,6 +46,7 @@ public class OrganizationUserRepository : Repository<Core.Entities.OrganizationU
 
     public async Task<ICollection<Guid>> CreateManyAsync(IEnumerable<Core.Entities.OrganizationUser> organizationUsers)
     {
+        organizationUsers = organizationUsers.ToList();
         if (!organizationUsers.Any())
         {
             return new List<Guid>();
--- a/test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepositoryTests.cs
+++ b/test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepositoryTests.cs
@@ -2,6 +2,7 @@
 using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Repositories;
+using Bit.Core.Utilities;
 using Xunit;
 
 namespace Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories;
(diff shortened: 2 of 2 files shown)


src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -46,6 +46,7 @@ from c in dbContext.Collections
 
     public async Task<ICollection<Guid>> CreateManyAsync(IEnumerable<Core.Entities.OrganizationUser> organizationUsers)
     {
+        organizationUsers = organizationUsers.ToList();
```
BTreston, line 49: @eliykat This seems to resolve the issue, not an expert on LINQ but my understanding is that this forces the query to execute and have the results that this method expects. Otherwise changes can happen to the data after we've registered them in the DB causing a potential mismatch.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepositoryTests.cs
```diff
@@ -354,4 +355,75 @@ await organizationUserRepository.CreateAsync(new OrganizationUser
...
+        var orgUsers = users.Select(u => new OrganizationUser
+        {
+            OrganizationId = org.Id,
+            UserId = u.Id,
+            Status = OrganizationUserStatusType.Confirmed,
+            Type = OrganizationUserType.Owner
+        });
+
+        // This fails because the IDs clash, even though the repository should set these to random values
+        var createdOrgUserIds = await organizationUserRepository.CreateManyAsync(orgUsers);
```
BTreston, line 386: I found that the clashing IDs was because they all initialize to some default value like 00000-00000-00000.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepositoryTests.cs
```diff
@@ -354,4 +355,75 @@ await organizationUserRepository.CreateAsync(new OrganizationUser
...
+
+        var orgUsers = users.Select(u => new OrganizationUser
+        {
+            OrganizationId = org.Id,
+            UserId = u.Id,
+            Status = OrganizationUserStatusType.Confirmed,
+            Type = OrganizationUserType.Owner
+        });
+
+        // This fails because the IDs clash, even though the repository should set these to random values
```
eliykat, line 385: Please delete this comment given that the test no longer fails - this was just a signpost for you.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepositoryTests.cs
```diff
@@ -354,4 +355,75 @@ await organizationUserRepository.CreateAsync(new OrganizationUser
...
+            Status = OrganizationUserStatusType.Confirmed,
+            Type = OrganizationUserType.Owner
+        });
+
+        var createdOrgUserIds = await organizationUserRepository.CreateManyAsync(orgUsers);
+
+        var readOrgUsers = await organizationUserRepository.GetManyByOrganizationAsync(org.Id, null);
+        var readOrgUserIds = readOrgUsers.Select(ou => ou.Id);
+
+        // the IDs returned by CreateMany don't match what we read back from the db
```
eliykat, line 426: And delete this comment as well.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
eliykat: OK, now I see. It's specifically because I was passing in a sequence that was not yet enumerated (the `users.Select` operation in the test). The `foreach` block in the repository method forced enumeration of the list (executed the `Select` call), but then threw it away when it went out of scope, so the mutated `OrganizationUser` objects were lost. The sequence would've been enumerated from scratch later on when being written to the db.

This also explains why it was kinda tricky to reproduce.

In any case, forcing enumeration is good practice to avoid issues just like this one, so this is a good fix.

eliykat (changes requested):
Thanks for digging into this. All you need to do is clean up some of my comments.

eliykat (approved):

eliykat:
QA will probably ask you to merge to `main` to test EF, but please check with them.

PR #5426, opened 2025-02-19
Incorrect Read only connection string on development self-hosted environment

mzieniukbw:
## 📔 Objective

Incorrect read only connection string on development self-hosted environment.

This bug is reproduced during selfHostedOverride, which is normally only the case for Bitwarden Developers environments https://github.com/bitwarden/server/blob/ae9bb427a16ff2ff5d1ba3b62101ad7e55bec02e/src/SharedWeb/Utilities/ServiceCollectionExtensions.cs#L582 and when read-only connection string is not explicit set.

Explanation:
The first time the GlobalSettings is read, both `ConnectionString` and then `ReadOnlyConnectionString` setters are called. For the read-only connection string setter, since there is no value, it will be taken from the getter, which is the `ConnectionString`.
The second time the GlobalSettings is read, the `ConnectionString` is overriden to self-hosted connection string. For the `ReadOnlyConnectionString` setter, since the value is not provided, it will be taken from the getter. But since we already store it in `_readOnlyConnectionString`, we return the first `ConnectionString` and not the second one `ConnectionString`.

The fix is to reset the `ReadOnlyConnectionString` stored value, every time the `ConnectionString` changes.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Settings/GlobalSettings.cs
+++ b/src/Core/Settings/GlobalSettings.cs
@@ -241,7 +241,18 @@ public class GlobalSettings : IGlobalSettings
         public string ConnectionString
         {
             get => _connectionString;
-            set => _connectionString = value.Trim('"');
+            set
+            {
+                // On development environment, the self-hosted overrides would not override the read-only connection string, since it is already set from the non-self-hosted connection string.
+                // This causes a bug, where the read-only connection string is pointing to self-hosted database.
+                if (!string.IsNullOrWhiteSpace(_readOnlyConnectionString) &&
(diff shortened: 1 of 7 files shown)


amorask-bitwarden (approved):
Billing changes look good.

Patrick-Pimentel-Bitwarden (approved):
👍  Auth changes looks good.

prograhamming (approved):
Tools changes look good.

PR #5438, opened 2025-02-24
[PM-13128] Breadcrumb Policies

jonashendrickx:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13128

## 📔 Objective

- Add a new permission `UseViewPolicies` to allow reading or opening the `Policies` page.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Admin/AdminConsole/Models/OrganizationEditModel.cs
```diff
@@ -69,6 +69,7 @@ public OrganizationEditModel(
         Seats = org.Seats;
         MaxAutoscaleSeats = org.MaxAutoscaleSeats;
         MaxCollections = org.MaxCollections;
+        UseViewPolicies = org.UseViewPolicies;
```
cturnbull-bitwarden, line 72: There might be a misunderstanding with this PR and the Jira story. All the necessary changes should be achievable on the client side alone. We're just looking to update the display logic to include Teams organizations when viewing policies. The policies feature toggle also exists on the organization table under `UsePolicies`.

jonashendrickx: Based on `https://bitwarden.atlassian.net/browse/PM-13128` I would have to enable `UseViewPolicies` for both `Enterprise` and `Teams` plans. Allowing us to view the `Policies` page. Then the breadcrumbing would be done checking directly for the teams plan to show the Enterprise upgrade badge.

Then the tab would be conditionally shown below to navigate to the `Policies` page. Which might be cleaner than having plan specific code leak all over the place. If we choose to change the naming of plans or introduce new plans, it would be easier to maintain our code base.

image

cturnbull-bitwarden: Roger that. I can see how it would be nice to add additional data to the organization table to split the view and edit permissions. If we do this, then we'll also need to write a script to update every Teams organization to enable this new column. That being the case because, for this to achieve the goal in the ticket of

```
GIVEN:

the user is the owner of a Teams org

the user is viewing the admin console for this org

THEN:

show “policies” in the left hand navigation
```

we'd need this value to be enabled for every teams org.

That said, we can make some assumptions about this authorization/role-based behavior given our criteria that allow us to not make any changes to this repository at all, and instead limit the changes to the clients repo.

If the organization's `UsePolicies` feature flag is not enabled, but the organization's `productTierType` is `Teams`, or `TeamsStarter`, which correspond to `2` and `4` respectively [here](https://github.com/bitwarden/clients/blob/main/libs/common/src/billing/enums/product-tier-type.enum.ts#L4), then we can conditionally display the nav menu item, and allow the user to navigate to the page.

Once the user is on the page, then we can implement the other requirements using similar logic, in addition to any change to make the fields read-only.

```
GIVEN:

the user is viewing the policies page OR

the user is viewing a policy modal

THEN: 

there is an upgrade badge next to the header
```
```
WHEN: 

the user selects the upgrade badge

THEN:

close the policy modal if open

open the current-plan-Teams upgrade modal
```

PR #5440, opened 2025-02-24
[PM-18608] Don't require new device verification on newly created accounts

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18608

## 📔 Objective

When a user first creates their account, they are expected to be onboarding lots of new devices. Limit new device verification during this critical "getting started" time.


Patrick-Pimentel-Bitwarden:
Is this something we are trying to expedite or can we record demo of it working?

ike-kottlowski (approved):

PR #5441, opened 2025-02-24
[PM-18583] Separate Desktop and CLI ClientTypes

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18583

## 📔 Objective

We currently group the desktop and CLI clients together in the `DesktopTypes`.

This causes downstream issues for feature flag targeting, as the `ClientType` is used [here](https://github.com/bitwarden/server/blob/a86741413a4e2268c1e1824d51af8ed193dcfc25/src/Core/Services/Implementations/LaunchDarklyFeatureService.cs#L156-L157) to assign to the LaunchDarkly context for feature flag targeting.

This means that it is impossible to target a different CLI version than desktop version, without specifying individual device types in LaunchDarkly (bypassing the convenience of the `ClientType`).

The reason for this request related to the linked issue is that for New Device Verification, we will have different versions for the CLI and the desktop in the upcoming release (`2025.2.1` vs. `2025.2.0`), and in setting up the feature flags for this release I noticed the grouping.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Enums/ClientType.cs
+++ b/src/Core/Enums/ClientType.cs
@@ -14,5 +14,7 @@ public enum ClientType : byte
     [Display(Name = "Desktop App")]
     Desktop = 3,
     [Display(Name = "Mobile App")]
-    Mobile = 4
+    Mobile = 4,
+    [Display(Name = "CLI")]
+    Cli = 5
 }
--- a/src/Core/Utilities/DeviceTypes.cs
+++ b/src/Core/Utilities/DeviceTypes.cs
@@ -16,7 +16,11 @@ public static class DeviceTypes
         DeviceType.LinuxDesktop,
         DeviceType.MacOsDesktop,
         DeviceType.WindowsDesktop,
-        DeviceType.UWP,
+        DeviceType.UWP
+    ];
+
+    public static IReadOnlyCollection<DeviceType> CliTypes { get; } =
+    [
         DeviceType.WindowsCLI,
         DeviceType.MacOsCLI,
         DeviceType.LinuxCLI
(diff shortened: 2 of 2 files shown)


trmartin4:
@mzieniukbw was there a specific intent behind grouping the desktop and CLI together?  I want to make sure that I don't break any other downstream contracts that might be relying on them being the same `ClientType`.

withinfocus (commented):
Is this safe to do? I am not sure if anything else ... somewhere ... considers the CLI a desktop client type. I understand your intention, but if my question is difficult to answer you could just use device type more explicitly for LD targeting.

trmartin4:
> Is this safe to do? I am not sure if anything else ... somewhere ... considers the CLI a desktop client type. I understand your intention, but if my question is difficult to answer you could just use device type more explicitly for LD targeting.

I am not sure where all the `ClientTypes` are used.  I posited the same question to @mzieniukbw as well.

On the other hand, [this](https://github.com/bitwarden/server/blob/a86741413a4e2268c1e1824d51af8ed193dcfc25/src/Identity/Utilities/LoginApprovingDeviceTypes.cs#L13) is currently a bug, since the CLI does not accept login approval requests.  It's a hidden bug, because the CLI doesn't even register to receive the requests at all, but it shows that we have assumed that `Desktop != CLI` as consumers of the `ClientType`.

Are you suggesting that we would need to evaluate the current context targeting, as well as all other uses of the `ClientType`, before making this change?

withinfocus:
> Are you suggesting that we would need to evaluate the current context targeting, as well as all other uses of the `ClientType`, before making this change?

Yes, just to be safe that the CLI wouldn't be expected to fall into other targets. This seems unlikely, but I just don't know.

trmartin4:
@withinfocus I've reviewed this with the EMs and we do not see any issues with backward compatibility with this, based on targeting rules for current flags.  Are you OK with merging this?

withinfocus:
@trmartin4 if it's proven out to not break something, sure.

PR #5446, opened 2025-02-26
[PM-2199] Implement userkey rotation for all TDE devices

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-2199
Clients PR: https://github.com/bitwarden/clients/pull/13576

## 📔 Objective

Adds trusted devices to the key-rotation v2 endpoint. This requires removing the secretverification request from the `{identifier}/retrieve-keys` endpoint. This is acceptable since it retrieves an *encrypted* copy of the keys, and also is existing behavior for e.g. PRF keys.

Further, this is a net-new, non-optional field (the device array) in the request body. But since userkey-rotation v2 is not rolled out and behind a feature-flag, we can add it as long as we make sure the client version contains the changes when activating the feature flag.

The validator behavior is as follows:
All trusted devices need to be included in the rotation request. No non-trusted devices shall be included in the response model. Trusted devices need to have the encryptedPublicKey and encryptedUserkey present.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Auth/Models/Api/Response/DeviceAuthRequestResponseModel.cs
```diff
@@ -19,7 +18,7 @@ public static DeviceAuthRequestResponseModel From(DeviceAuthDetails deviceAuthDe
             Type = deviceAuthDetails.Type,
             Identifier = deviceAuthDetails.Identifier,
             CreationDate = deviceAuthDetails.CreationDate,
-            IsTrusted = deviceAuthDetails.IsTrusted()
+            IsTrusted = deviceAuthDetails.IsTrusted,
```
quexten, line 21: Note for reviews: This is fixing an existing bug where the device trust status is incorrectly reported in the device list request.

src/Infrastructure.Dapper/Repositories/DeviceRepository.cs
```diff
@@ -109,4 +111,35 @@ await connection.ExecuteAsync(
                 commandType: CommandType.StoredProcedure);
         }
     }
+
+    public UpdateEncryptedDataForKeyRotation UpdateKeysForRotationAsync(Guid userId, IEnumerable<Device> devices)
```
ike-kottlowski, line 115: 🎨 : I'm of the opinion that this should be a Stored Procedure. I understand this is essentially the same thing, but for consistency's sake it might make sense to either create a SP or to use `Devices_Update` via the `Repository.Replace(T obj)`.

ike-kottlowski (commented):
Have a question about the query.

src/Infrastructure.Dapper/Repositories/DeviceRepository.cs
quexten: Hmm, the intent here was to stay consistent with what the other auth repositories do for rotation:

https://github.com/bitwarden/server/blob/10756ca35e9c5ae0da34374e91d24ce4986fb883/src/Infrastructure.Dapper/Auth/Repositories/WebAuthnCredentialRepository.cs#L64-L86

https://github.com/bitwarden/server/blob/10756ca35e9c5ae0da34374e91d24ce4986fb883/src/Infrastructure.Dapper/Auth/Repositories/EmergencyAccessRepository.cs#L106-L146

Are these planned to be changed to SP's too?

ike-kottlowski: Well, now that's interesting 🫠.

Do we think the intent with the rotation queries being done this way is to capture the changes in one query transaction rather than a loop of transactions by using the `Repository.Replace(T obj)`?

quexten: Yeah, one big transaction query is required to ensure consistency during key rotation.

Where possible both on the Client <-> Server API, and on Server <-> DB we've been moving to one big atomic change to prevent any kind of connection loss from making the account move to an inconsistent / undecryptable state.

(For instance, for TDE, if we just use repository.replace, I think it would be possible that a temporary connection loss to the sql db, would lead to only some TDE devices having an updated copy of the userkey, and some devices would be able to log in, some would not).

I think this change to one big transaction was originally made during auth times by @jlf0dev though. Maybe he can confirm the intent.

src/Infrastructure.Dapper/Repositories/DeviceRepository.cs
```diff
@@ -109,4 +111,35 @@ await connection.ExecuteAsync(
...
+                INNER JOIN
+                    OPENJSON(@DeviceCredentials)
+                    WITH (
+                        id UNIQUEIDENTIFIER,
+                        encryptedPublicKey NVARCHAR(MAX),
+                        encryptedUserKey NVARCHAR(MAX)
+                    ) UD
+                    ON UD.id = D.Id
+                WHERE
+                    D.[UserId] = @UserId";
```
Thomas-Avery, line 135:
```suggestion
            const string sql = @"
                UPDATE D
                SET
                    D.[EncryptedPublicKey] = UD.[encryptedPublicKey],
                    D.[EncryptedUserKey] = UD.[encryptedUserKey]
                FROM
                    [dbo].[Device] D
                INNER JOIN
                    OPENJSON(@DeviceCredentials)
                    WITH (
                        id UNIQUEIDENTIFIER,
                        encryptedPublicKey NVARCHAR(MAX),
                        encryptedUserKey NVARCHAR(MAX)
                    ) UD
                    ON UD.[id] = D.[Id]
                WHERE
                    D.[UserId] = @UserId";
```
Might want to double check if these constructed columns with `OPENJSON` are case sensitive.

src/Infrastructure.EntityFramework/Repositories/DeviceRepository.cs
```diff
@@ -91,4 +92,30 @@ public async Task<ICollection<DeviceAuthDetails>> GetManyByUserIdWithDeviceAuth(
...
+                .Where(wc => wc.UserId == userId);
+
+            foreach (var wc in validUserDevices)
+            {
+                var nwc = newDevices.First(eak => eak.Id == wc.Id);
+                wc.EncryptedPublicKey = nwc.EncryptedPublicKey;
+                wc.EncryptedUserKey = nwc.EncryptedUserKey;
+            }
+
+            await dbContext.SaveChangesAsync();
```
Thomas-Avery, line 117:
```suggestion
         var deviceUpdates = devices.ToList();
            using var scope = ServiceScopeFactory.CreateScope();
            var dbContext = GetDatabaseContext(scope);
            var userDevicesQuery = dbContext.Devices.Where(d => d.UserId == userId);
            var validUserDevices = await userDevicesQuery
                .Where(d => deviceUpdates.Any(du => du.Id == d.Id)).ToListAsync();

            foreach (var deviceEntity in validUserDevices)
            {
                var deviceUpdate = deviceUpdates.First(du => du.Id == deviceEntity.Id);
                deviceEntity.EncryptedPublicKey = deviceUpdate.EncryptedPublicKey;
                deviceEntity.EncryptedUserKey = deviceUpdate.EncryptedUserKey;
            }

            await dbContext.SaveChangesAsync();
```

I changed some variable names to make it easier for me to reason about, feel free to change them back. 

Main concern for me here:
```
   var userDevices = await GetDbSet(dbContext)
                .Where(wc => wc.Id == wc.Id)
                .ToListAsync();
```
This will pull the entire Device table into memory. ` .Where(wc => wc.Id == wc.Id)`  is the same as doing `.Where(wc => true)`

Thomas-Avery (changes requested):
Looking good, a few things to check out.

src/Infrastructure.EntityFramework/Repositories/DeviceRepository.cs
quexten: I mostly copied these from the webauthn repository. It seems it has the same `wc.Id == wc.Id`... So that is future tech debt for auth to clean up probably? https://github.com/bitwarden/server/blob/6178bb2db183614edfd300acb0da0c5f8171e40e/src/Infrastructure.EntityFramework/Auth/Repositories/WebAuthnCredentialRepository.cs#L72

cc @JaredSnider-Bitwarden

Thomas-Avery: Agreed for that `WebAuthnCredentialRepository`

Thomas-Avery (approved):
LGTM

src/Infrastructure.EntityFramework/Repositories/DeviceRepository.cs
JaredSnider-Bitwarden: @quexten , ty for the ping. I've got a PR up fixing this in draft here: https://github.com/bitwarden/server/pull/5567. I'll have to figure out how to test the EF key rotation to get it ready for review.

Thomas-Avery: @JaredSnider-Bitwarden if you need any help getting EF setup for a test, feel free to reach out to me.

PR #5452, opened 2025-02-28
[VULN-176] - [Privilege Escalation] - Unauthorised access allows limited access user to change password of Items

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/VULN-176

## 📔 Objective

This PR fixes an issue with "hidden password" users able to update passwords.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Vault/Services/Implementations/CipherService.cs
+++ b/src/Core/Vault/Services/Implementations/CipherService.cs
@@ -13,7 +13,9 @@ using Bit.Core.Tools.Models.Business;
 using Bit.Core.Tools.Services;
 using Bit.Core.Utilities;
 using Bit.Core.Vault.Entities;
+using Bit.Core.Vault.Enums;
 using Bit.Core.Vault.Models.Data;
+using Bit.Core.Vault.Queries;
 using Bit.Core.Vault.Repositories;
 
 namespace Bit.Core.Vault.Services;
@@ -38,6 +40,7 @@ public class CipherService : ICipherService
     private const long _fileSizeLeeway = 1024L * 1024L; // 1MB
     private readonly IReferenceEventService _referenceEventService;
     private readonly ICurrentContext _currentContext;
+    private readonly IGetCipherPermissionsForUserQuery _getCipherPermissionsForUserQuery;
 
     public CipherService(
         ICipherRepository cipherRepository,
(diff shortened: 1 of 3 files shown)


src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -161,6 +166,23 @@ public async Task SaveDetailsAsync(CipherDetails cipher, Guid savingUserId, Date
...
             cipher.RevisionDate = DateTime.UtcNow;
+            if (cipher.Type == CipherType.Login && cipher.Data != null)
+            {
+                var existingCipher = await _cipherRepository.GetByIdAsync(cipher.Id);
+                if (existingCipher != null)
+                {
+                    // Check if user has permission to change passwords
+                    var cipherPermissions = await _getCipherPermissionsForUserQuery.GetByOrganization(cipher.OrganizationId.Value);
+                    if (!cipherPermissions.TryGetValue(cipher.Id, out var permission) || !(permission.ViewPassword && permission.Edit))
+                    {
```
gbubemismith, line 177: ❓ Is there a reason we aren't just terminating the request when the user doesn't have the required permission?

jaasen-livefront: @gbubemismith The user may still have permission to edit non-password fields so we can't just exit at this point. The best we can do is just not update the password field.

gbubemismith: I was thinking of something like this 


```csharp
 var existingCipherPassword = JsonSerializer.Deserialize<CipherLoginData>(existingCipher.Data).Password;
 var newCipherPassword = JsonSerializer.Deserialize<CipherLoginData>(cipher.Data).Password;

// Check if the password is changing and if the user lacks proper permissions
 if (!existingCipherPassword.Equals(newCipherPassword))
 {
       // Check if user has permission to change passwords
       if() {
        throw new BadRequestException("You do not have permissions to change this password.");
        }
  }
```

what do you think?

jaasen-livefront: @gbubemismith Since we have only the encrypted strings there's no way for us to determine if the decrypted passwords are identical as two identical strings always produce different encryption strings.

gbubemismith: Completely forgot about that

jaasen-livefront: @gbubemismith done!

gbubemismith (dismissed):

jaasen-livefront:
@gbubemismith Moved some logic into the new method if you wouldn't mind taking another look. ;)

gbubemismith (approved):

PR #5458, opened 2025-03-03
[PM-17901] Replaced hard-coded Bitwarden Vault URLs

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17901

## 📔 Objective

There were several cases where we had the .com Bitwarden Vault URL hard-coded in our email templates. This PR resolves those according to the requirements in the ticket's "How do we fix it?" section. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Services/Implementations/HandlebarsMailService.cs
+++ b/src/Core/Services/Implementations/HandlebarsMailService.cs
@@ -214,9 +214,9 @@ public class HandlebarsMailService : IMailService
         var message = CreateDefaultMessage($"{organization.DisplayName()} Seat Count Has Increased", ownerEmails);
         var model = new OrganizationSeatsAutoscaledViewModel
         {
-            OrganizationId = organization.Id,
             InitialSeatCount = initialSeatCount,
             CurrentSeatCount = organization.Seats.Value,
+            VaultSubscriptionUrl = VaultSubscriptionUrl(organization.Id)
         };
 
         await AddMessageContentAsync(message, "OrganizationSeatsAutoscaled", model);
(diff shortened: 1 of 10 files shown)


src/Core/Services/Implementations/HandlebarsMailService.cs
```diff
@@ -1223,4 +1223,11 @@ private static string GetUserIdentifier(string email, string userName)
     {
         return string.IsNullOrEmpty(userName) ? email : CoreHelpers.SanitizeForEmail(userName, false);
     }
+
+    public string VaultSubscriptionUrl(Guid organizationId)
```
jrmccannon, line 1227: :art: could we name this `GetVaultSubscriptionUrl(...)`?

Also would it make sense for these to be built off the `WebVaultUrl` variable? Or does that not change based on eu / not eu ?

jrmccannon (dismissed):
Looks good. Just had a small question/suggestion. Non-blocking.

src/Core/Services/Implementations/HandlebarsMailService.cs
amorask-bitwarden: Hey @jrmccannon, thanks for the review! 

I updated the name of the method to be `GetCloudVaultSubscriptionUrl` so it's a bit more clear. The reason we can't use `WebVaultUrl` is that, in Self-Host, that will point to the Self-Host installation's vault. The emails updated as part of this PR need to contain a link to the associated cloud vault.

jrmccannon (approved):

PR #5469, opened 2025-03-05
[PM-18890] Import errors because permissions are reversed

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18890

## 📔 Objective

Given a user, who is not the owner or admin of an organization, but has been giving privileges to Manage a collection, the user must be able to import into that org.
The reason for the failure is this:
 - if the user did not have `ImportCipher` privileges - the program would return `no privilege`
 - if the user did not have `Create` privileges - the program would return `no privilege`
 
This PR changes it to this:
- if the user has `Import Cipher` privileges - then return 'yes`
- Or if the user has `Create` privilege - then return `yes`

## 📸 Screenshots

None at this time

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Tools/Controllers/ImportCiphersController.cs
+++ b/src/Api/Tools/Controllers/ImportCiphersController.cs
@@ -102,13 +102,19 @@ public class ImportCiphersController : Controller
             .Select(c => c.Id)
             .ToHashSet();
 
-        //We need to verify if the user is trying to import into existing collections
+        // when there are no collections, then we can import
+        if (collections.Count == 0)
+        {
+            return true;
+        }
+
+        // are we trying to import into existing collections?
         var existingCollections = collections.Where(tc => orgCollectionIds.Contains(tc.Id));
 
-        // Do we have new collections?
+        // are we trying to create new collections?
         var hasNewCollections = collections.Any(tc => !orgCollectionIds.Contains(tc.Id));
 
-        //suppose we are going to be creating new collections and we have existing collections
+        // suppose we have both new and existing collections
         if (hasNewCollections && existingCollections.Any())
         {
             // since we are creating new collection, user must have import/manage and create collection permission
@@ -120,12 +126,12 @@ public class ImportCiphersController : Controller
             }
             else
             {
-                // user does not have permission to create new collections
+                // user does not have permission to import
                 return false;
             }
         }
 
-        // we are creating new collections and we don't have any existing collections
+        // suppose we have new collections and none of our collections exist
         if (hasNewCollections && !existingCollections.Any())
         {
             // user is trying to create new collections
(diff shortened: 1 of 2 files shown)


djsmith85 (changes requested):
@voommen-livefront I previously, while working/debugging https://github.com/bitwarden/server/pull/5385 I came to the same solution, but it allows a user to create sub-collections within their managed collection. Something which currently we do not want to enable. 

IMHO if we want to support this, then we should also enable members to create sub-collections within the organizational vault UI.

Lets discuss this offline with product again.

src/Api/Tools/Controllers/ImportCiphersController.cs
```diff
@@ -107,17 +107,17 @@ private async Task<bool> CheckOrgImportPermission(List<Collection> collections,
         var existingCollections = collections.Where(tc => orgCollectionIds.Contains(tc.Id));
```
voommen-livefront, line 108: if you are creating a subcollection - than we need to check ImportCiphers and Create

djsmith85 (changes requested):
Discussed this offline on Friday last week. 
@voommen-livefront: Please ask for re-review when ready

src/Api/Tools/Controllers/ImportCiphersController.cs
```diff
@@ -106,18 +105,56 @@ private async Task<bool> CheckOrgImportPermission(List<Collection> collections,
...
+            return true;
         };
 
         //Users allowed to import if they CanCreate Collections
-        if (!(await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.Create)).Succeeded)
+        if ((await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.Create)).Succeeded)
         {
-            return false;
+            return true;
         }
```
djsmith85, line 156: This shouldn't be necessary anymore. In case the user is attempting to create a new collection within an import they would have entered one of the two previous if's starting with `hasNewCollections`

djsmith85 (changes requested):
Changes look good, just some smaller items.
Most of the if/else structures added are a little redundant, but it is always use to add a comment to clarify flow, which helps with understanding the code 👍

src/Api/Tools/Controllers/ImportCiphersController.cs
```diff
@@ -106,18 +105,56 @@ private async Task<bool> CheckOrgImportPermission(List<Collection> collections,
         //We need to verify if the user is trying to import into existing collections
         var existingCollections = collections.Where(tc => orgCollectionIds.Contains(tc.Id));
```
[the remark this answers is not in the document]
voommen-livefront: This is a good point. Done.

src/Api/Tools/Controllers/ImportCiphersController.cs
voommen-livefront: You are right - I removed the code and ran the unit tests and tested the UI and it is not needed.

djsmith85 (approved):

PR #5475, opened 2025-03-06
[PM-15621] Add support for handling multiple CommandResult validators.

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15621

## 📔 Objective

1. Add a method to make handling multiple CommandResult validators easier.
2. Add test coverage.

## Testing

This code is not wired up, but we have good test coverage for it. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


jrmccannon (dismissed):

src/Core/Validators/CommandResultValidator.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Bit.Core.Models.Commands;
+
+namespace Bit.Core.Validators;
+
+public static class CommandResultValidator
+{
+    public static CommandResult ExecuteValidators(Func<CommandResult>[] validators)
```
eliykat, line 7: "Run synchronous validators, then run async validators" is probably going to be a pretty common pattern. I suggest a combined interface, like `ExecuteValidatorsAsync(Func<CommandResult>[], Func<Task<CommandResult>>[])` which can just handle it all for you. That itself would be asynchronous but doesn't matter as long as it's guaranteed to run synchronous validators first.

src/Core/Validators/CommandResultValidator.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Bit.Core.Models.Commands;
+
+namespace Bit.Core.Validators;
+
+public static class CommandResultValidator
```
eliykat, line 5: I know you two have discussed this so I may be out of the loop, but `CommandResultValidator` is an odd name to me. You're validating the inputs of the command, not the result. `CommandRequestValidator`?

eliykat (commented):
I like it, just some minor feedback.

src/Core/Validators/CommandResultValidator.cs
jrmccannon: Jimmy and I talked about this and we are POC'ing some stuff now about this.

Mostly around the difference between Commands and Validators. There are a few more POC PRs we're putting together to try to get some common types out for everyone to look at.

src/Core/Validators/CommandResultValidator.cs
JimmyVo16: I do like having a method that can handle both. Plus, this will ensure that we always handle sync validators first before async validators.

There is a very small trade-off in that if you only need sync validators, you will have to call an async method, which creates a little overhead. However, the value of always ensuring sync validators run first outweighs this minor edge case.

JimmyVo16:
I think I’m going to hold off on an abstraction until we solidify our command/validation pattern.

PR #5508, opened 2025-03-17
[PM-19590] Add k6 load testing script for SyncController's /sync endpoint

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19590

## 📔 Objective

Adding a k6 load testing script at `perf/load/sync.js` to measure the performance of the `/sync` endpoint. This helps us ensure that changes don’t degrade performance.

### Script details

Script runs two load scenarios:

* **Constant Load:**
  * 30 requests per minute for 10 minutes (5 VUs)

* **Ramping Load:**
  * Starts at 30 requests/min, ramps up to 150, then back down (20 VUs)

The script:
* Uses environment variables for authentication
* Sets performance thresholds:
  * Failure rate < 1% (`http_req_failed: rate<0.01`)
  * 95% of requests complete in <400ms (`http_req_duration: p(95)<400`)

### How to run

Example:

```bash
k6 run -e IDENTITY_URL=http://localhost:33656 \
       -e API_URL=http://localhost:4000 \
       -e CLIENT_ID=web \
       -e AUTH_USER_EMAIL=user1@example.com \
       -e AUTH_USER_PASSWORD_HASH=<hash> \
       perf/load/sync.js
```

This gives us a baseline to compare against when making changes to `/sync`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


perf/load/sync.js
```diff
@@ -0,0 +1,95 @@
...
+        { duration: "30s", target: 150 },
+        { duration: "30s", target: 60 },
+        { duration: "30s", target: 0 },
+      ],
+      preAllocatedVUs: 20,
+    },
+  },
+  thresholds: {
+    http_req_failed: ["rate<0.01"],
+    http_req_duration: ["p(95)<400"],
```
withinfocus, line 46: ❓ Have you by chance looked at Datadog for average response time here? Also, it's gonna be dependent on what the test user has in their vault I suppose.

perf/load/sync.js
```diff
@@ -0,0 +1,95 @@
...
+
+export function setup() {
+  return authenticate(IDENTITY_URL, CLIENT_ID, AUTH_USERNAME, AUTH_PASSWORD);
+}
+
+export default function (data) {
+  if (!data || !data.access_token) {
+    console.error("No access token available. Authentication may have failed.");
+    return;
+  }
```
withinfocus, line 58: ❓ Were you seeing this get invoked? The authentication call during setup should fail if this were the case, and stop the test altogether.

withinfocus (commented):
Had a moment to look at this -- love it! Just a couple questions but after that we can merge and try this out.

eliykat (commented):
LGTM and thanks for following this up 👏 .

As discussed it would be good to tune the thresholds so that our faulty code from 2023.2.0 causes a clear failure (given that it caused a clear failure in production). This may not be possible without a static set of test data though, as mentioned by @withinfocus. In particular, that defect requires that there is a certain quantity of organizations with claimed domains on the test instance, which we can't guarantee just running it against whatever data happens to be in QA. 🤔

perf/load/sync.js
r-tome: I forgot to remove that. I put that there for debugging.

perf/load/sync.js
r-tome: I looked at the performance for the past month and 400 seemed to be a good spot: 
image

r-tome:
@eliykat, I couldn't find the average response time for USQA on Datadog, which is the environment used for weekly load testing. Maybe the @bitwarden/team-sre team can help with this?

I ran a local experiment where I set up a database with 1M organizations, each with a claimed domain and user. With the feature flag pm-10308-account-deprovisioning": false, the average response time was around 1 second. However, when the feature flag was enabled, all requests failed (even with the new sproc version). So I think this script is already useful.

eliykat:
Sounds good to me! I'll let SRE do their review as the codeowners.

perf/load/sync.js
watsondm: I just want to add here that if these tests are running exclusively in lower environments, the thresholds for error rate and p95 latency will need to be tuned differently than the production environment.

PR #5521, opened 2025-03-18
Ac/pm 18240 implement policy requirement for reset password policy

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-18240

## 📔 Objective
Implements `ResetPasswordPolicyRequirement` and connects to DI. Enforces policy on callers instead of using policy service or repository calls.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -320,12 +320,13 @@ public class OrganizationUsersController : Controller
             throw new UnauthorizedAccessException();
         }
 
-        var useMasterPasswordPolicy = _featureService.IsEnabled(FeatureFlagKeys.PolicyRequirements) ?
-         await ShouldHandleResetPasswordAsync_vNext(orgId, user.Id) : await ShouldHandleResetPasswordAsync(orgId);
+        var useMasterPasswordPolicy = _featureService.IsEnabled(FeatureFlagKeys.PolicyRequirements)
+        ? (await _policyRequirementQuery.GetAsync<ResetPasswordPolicyRequirement>(user.Id)).AutoEnrollEnabled(orgId)
+        : await ShouldHandleResetPasswordAsync(orgId);
 
         if (useMasterPasswordPolicy && string.IsNullOrWhiteSpace(model.ResetPasswordKey))
         {
-            throw new BadRequestException(string.Empty, "Master Password reset is required, but not provided.");
+            throw new BadRequestException("Master Password reset is required, but not provided.");
         }
 
         await _acceptOrgUserCommand.AcceptOrgUserByEmailTokenAsync(organizationUserId, user, model.Token, _userService);
@@ -351,18 +352,6 @@ public class OrganizationUsersController : Controller
                                           masterPasswordPolicy.GetDataModel<ResetPasswordDataModel>().AutoEnrollEnabled;
 
         return useMasterPasswordPolicy;
-    }
-
-    private async Task<bool> ShouldHandleResetPasswordAsync_vNext(Guid orgId, Guid userId)
-    {
-        var organizationAbility = await _applicationCacheService.GetOrganizationAbilityAsync(orgId);
-
-        if (organizationAbility is not { UsePolicies: true })
-        {
-            return false;
-        }
-
-        return (await _policyRequirementQuery.GetAsync<ResetPasswordPolicyRequirement>(userId)).AutoEnrollEnabled;
     }
 
     [HttpPost("{id}/confirm")]
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -8,6 +8,8 @@ using Bit.Core.AdminConsole.Enums;
 using Bit.Core.AdminConsole.Models.Data.Organizations.Policies;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Authorization;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
 using Bit.Core.AdminConsole.OrganizationFeatures.Shared.Authorization;
 using Bit.Core.AdminConsole.Repositories;
 using Bit.Core.Auth.Enums;
@@ -55,6 +57,7 @@ public class OrganizationUsersController : Controller
     private readonly IRemoveOrganizationUserCommand _removeOrganizationUserCommand;
     private readonly IDeleteManagedOrganizationUserAccountCommand _deleteManagedOrganizationUserAccountCommand;
     private readonly IGetOrganizationUsersManagementStatusQuery _getOrganizationUsersManagementStatusQuery;
+    private readonly IPolicyRequirementQuery _policyRequirementQuery;
     private readonly IFeatureService _featureService;
     private readonly IPricingClient _pricingClient;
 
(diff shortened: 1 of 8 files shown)


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/ResetPasswordPolicyRequirement.cs
eliykat, line 1: This policy is a bit different to others - it's used to trigger a specific action for a specific org. i.e. we need to know whether to enroll this user for _this specific_ organization. As the requirement is currently written, it only tells you if the user is required to be enrolled by _any_ organization.  We can't have a user get enrolled with Org A because Org B has the policy turned on.

(In reality this policy requires the Single Organization policy, so it's pretty unlikely that it would get mixed up, but it's still possible because Single Org does not apply to everyone. Plus this is a sensitive action so we want the logic to be robust.)

I suggest you can handle this by adding a method like `public bool AutoEnroll(Guid organizationId)`, so that callers can check for a specific org. Internally, keep a track of a list of OrgIds that require auto-enrollment. This would replace the current `bool` which is no longer useful.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/ResetPasswordPolicyRequirement.cs
```diff
@@ -0,0 +1,34 @@
...
+/// </summary>
+public class ResetPasswordPolicyRequirement : IPolicyRequirement
+{
+    /// <summary>
+    /// Indicates that new members will automatically be enrolled in account recovery administration when joining an organization.
+    /// </summary>
+    public bool AutoEnrollEnabled { get; init; }
+}
+
+public class ResetPasswordPolicyRequirementFactory : BasePolicyRequirementFactory<ResetPasswordPolicyRequirement>
```
eliykat, line 17: Because we use all direct repository calls today, no-one is exempt due to their role or provider status.

To maintain current behaviour, let's override the `ExemptRoles => []` and `ExemptProviders => false`.

eliykat (changes requested):
Thanks for picking this up!

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1353,14 +1358,26 @@ public async Task UpdateUserResetPasswordEnrollmentAsync(Guid organizationId, Gu
...
-        if (resetPasswordKey == null && resetPasswordPolicy.Data != null)
+        if (_featureService.IsEnabled(FeatureFlagKeys.PolicyRequirements))
         {
-            var data = JsonSerializer.Deserialize<ResetPasswordDataModel>(resetPasswordPolicy.Data, JsonHelpers.IgnoreCase);
-
-            if (data?.AutoEnrollEnabled ?? false)
+            var resetPasswordPolicyRequirement = await _policyRequirementQuery.GetAsync<ResetPasswordPolicyRequirement>(userId);
+            if (resetPasswordKey == null && resetPasswordPolicyRequirement.AutoEnrollEnabled)
             {
                 throw new BadRequestException("Due to an Enterprise Policy, you are not allowed to withdraw from Password Reset.");
```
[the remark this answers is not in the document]
BTreston: In case I don't get an answer, waiting for a reply

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -315,11 +320,13 @@ public async Task Accept(Guid orgId, Guid organizationUserId, [FromBody] Organiz
...
 
-        var useMasterPasswordPolicy = await ShouldHandleResetPasswordAsync(orgId);
+        var useMasterPasswordPolicy = _featureService.IsEnabled(FeatureFlagKeys.PolicyRequirements)
+        ? (await _policyRequirementQuery.GetAsync<ResetPasswordPolicyRequirement>(user.Id)).AutoEnrollEnabled(orgId)
+        : await ShouldHandleResetPasswordAsync(orgId);
 
         if (useMasterPasswordPolicy && string.IsNullOrWhiteSpace(model.ResetPasswordKey))
         {
-            throw new BadRequestException(string.Empty, "Master Password reset is required, but not provided.");
+            throw new BadRequestException("Master Password reset is required, but not provided.");
```
BTreston, line 329: This is the code path being tested that you mentioned. This was causing the message to be the default string.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/ResetPasswordPolicyRequirement.cs
```diff
@@ -9,24 +10,41 @@ namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements
...
                 {
-                    AutoEnrollEnabled = result.AutoEnrollEnabled || data.AutoEnrollEnabled
+                    var dataModel = data.GetDataModel<ResetPasswordDataModel>();
+                    if (dataModel.AutoEnrollEnabled && !result.AutoEnroll.Contains(data.OrganizationId))
+                    {
+                        return new ResetPasswordPolicyRequirement() { AutoEnroll = result.AutoEnroll.Append(data.OrganizationId) };
+                    }
+
+                    return result;
                 });
```
eliykat, line 48: This is basically filtering and then transforming a list. Using more granular linq operations that express each of these steps is easier to understand:

```c#
var orgsWithAutoEnroll = policyDetails
  .Where(p => p.GetDataModel().AutoEnrollEnabled)
  .Select(p => p.OrganizationId)
  .ToHashSet(); // to remove duplicates if you're concerned about them
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/ResetPasswordPolicyRequirement.cs
```diff
@@ -9,24 +10,41 @@ namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements
 public class ResetPasswordPolicyRequirement : IPolicyRequirement
 {
     /// <summary>
-    /// Indicates that new members will automatically be enrolled in account recovery administration when joining an organization.
+    /// List of Organization Ids that require automatic enrollment in password recovery.
     /// </summary>
-    public bool AutoEnrollEnabled { get; init; }
+    public IEnumerable<Guid> AutoEnroll { get; init; }
```
eliykat, line 15: I suggest this is private to avoid people misusing it. `AutoEnrollEnabled` is the public interface.

That said, it is still convenient to arrange directly in tests, so maybe there is still a public setter - i.e. a public init property with a private backing field.

```c#
private IEnumerable<Guid> _autoEnrollOrganizations;
public IEnumerable<Guid> AutoEnrollOrganizations { init => _autoEnrollOrganizations = value; }
```

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/ResetPasswordPolicyRequirementFactoryTests.cs
```diff
@@ -12,23 +13,26 @@ namespace Bit.Core.Test.AdminConsole.OrganizationFeatures.Policies.PolicyRequire
...
     {
         var actual = sutProvider.Sut.Create([]);
 
-        Assert.False(actual.AutoEnrollEnabled);
+        Assert.True(actual.AutoEnroll.IsNullOrEmpty());
     }
 
     [Theory, BitAutoData]
-    public void AutoEnrollAdministration_WithResetPasswordPolices_ReturnsTrue(
+    public void AutoEnrollAdministration_WithAnyResetPasswordPolices_ReturnsEnabledOrganizationIds(
```
eliykat, line 24: I recommend asserting against the public interface - the `AutoEnrollEnabled` method - rather than the internal list.

eliykat (changes requested):

eliykat (approved):

PR #5555, opened 2025-03-26
[PM-19585] Use Authorize attributes for simple role authorization

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19585

## 📔 Objective

Trial the use of using Authorize attributes instead of manually calling AuthorizationService in the body of the method.

- Add an Authorize attribute that uses the requirement specified in the generic type argument. This avoids the use of magic policy strings and then manually having to register each requirement against an authorization policy.
- Add the `OrganizationRequirementHandler`, which gets the orgId from the route and then passes it to `IOrganizationRequirement.Authorize`. Putting this callback logic on the requirement means we don't have a big switch statement in the handler, or many boilerplate handlers, and it's easy to match the requirement to underlying logic. That said, it accommodates simple cases only - if you had more complex authorization logic you could fall back to fully defining your logic in your handler.
- Add some basic integration tests just to prove it works. These would be cumbersome to add for every endpoint so they're really just here for the POC.

Note that this reverts to the more general abstraction of (for example) a "ManageUsers" requirement rather than an endpoint specific "ReadAllUserDetails" requirement. In this way it's at the same level of generality as `ICurrentContext`. However, I think this is fine for simple role-based checks - keep simple things simple, and you can always define more specific requirements if needed.

This also avoids perpetuating `ICurrentContext` as a large all-purpose cross-team class for authorization, while also avoiding the boilerplate of manually defining separate requirements and handlers for every single endpoint.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/Shared/Authorization/IOrganizationRequirement.cs
```diff
@@ -0,0 +1,18 @@
...
+/// <summary>
+/// A requirement that implements this interface will be handled by <see cref="OrganizationRequirementHandler"/>,
+/// which calls AuthorizeAsync with the organization details from the route.
+/// This is used for simple role-based checks.
+/// This may only be used on endpoints with {orgId} in their path.
+/// </summary>
+public interface IOrganizationRequirement : IAuthorizationRequirement
+{
+    // TODO: avoid injecting all of ICurrentContext?
+    public Task<bool> AuthorizeAsync(Guid organizationId, CurrentContextOrganization? organizationClaims, ICurrentContext currentContext);
```
eliykat, line 17: Ideally this would take a restricted set of parameters and not all of `ICurrentContext` (which provides a backdoor to continue its current mis/use).

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -35,6 +35,7 @@ namespace Bit.Api.AdminConsole.Controllers;
 
 [Route("organizations/{orgId}/users")]
 [Authorize("Application")]
+[Authorize<MemberOrProviderRequirement>]
```
eliykat, line 38: We can add this one to the whole controller as an extra layer of protection - ensuring that you need to have some relationship to the org (as a member or provider) to access any endpoint within it. Then individual endpoints can have more specific requirements.

r-tome (commented):
This is really great work! Good idea and execution

src/Api/Startup.cs
```diff
@@ -255,10 +257,9 @@ public void Configure(
 
         // Add authentication and authorization to the request pipeline.
         app.UseAuthentication();
-        app.UseAuthorization();
-
-        // Add current context
+        // Note: ICurrentContext is used in authorization middleware so it must be registered first
         app.UseMiddleware<CurrentContextMiddleware>();
+        app.UseAuthorization();
```
eliykat, line 262: @justindbaur can I get your approval that it's OK to move the Current Context middleware before Authorization middleware? I can't see any harm in it but just double checking.

justindbaur: It is somewhat eyebrow raising I guess is the way I would put it. I cannot think of any problems with our current application but it does make me want to put some warnings on various things. 

`#1` - We should be extra careful to not do anything expensive in `CurrentContext.BuildAsync` especially if those calls could be kicked off with 0 claims. This makes me think `ICurrentContext` needs an owner to help make sure this never happens.
`#2` - We should make sure we are extremely careful to ever create a `ClaimsIdentity` with claims but without an authenticate type and make that be the user on `HttpContext`, as far as I am aware we do not do this, but it is a trick I have employed in the past (personally) with what I considered lower security authentication schemes. 
`#3` I would be extra sure to have your additional requirements after the pre-existing `[Authorize("PolicyName")]` ones so that they execute first.

eliykat: That sounds like a lot of things to be careful of. I'm open to other approaches if it'll avoid setting traps for ourselves in the future.

I'm mostly using `CurrentContext` as a convenient way to access the user's JWT claims (in the form of `CurrentContextOrganization`) for authorization. However, this can also be done by parsing the claims directly from the `ClaimsPrincipal` available via `AuthorizationHandlerContext`, without needing to run middleware or construct the entire `CurrentContext`. For example, it could be an extension method on the claims:

```c#
var organizatonClaims = context.User.Claims.GetCurrentContextOrganization(orgId);
```

The other reason I'm using CurrentContext is to access provider info, but that's really just a wrapper around a database call and could be moved into a different service.

Then the CurrentContextMiddleware could stay where it is, after authentication and authorization.

What do you think?

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -137,31 +138,19 @@ public async Task<OrganizationUserDetailsResponseModel> Get(Guid id, bool includ
         return response;
     }
 
+    [Authorize<MemberOrProviderRequirement>]
```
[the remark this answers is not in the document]
eliykat: It's intentional redundancy in case someone removes an attribute on a particular endpoint. It's probably being overcautious though.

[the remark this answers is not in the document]
eliykat: Removed the duplicate use at the class level.

src/Api/Startup.cs
justindbaur: I personally love grabbing things more directly off the `HttpContext` as an extension method. I'm kind of a `ICurrentContext` hater though. With doing things like that, things are more pay-for-play. The parsing of the claims into a more type safe version are only done when actually needed, vs being done for every request even if nothing in that request ever needs it. 

My only change to that is that if the parsing is slow, you can cache the result into a `HttpContext.Features`. So you would have something like this:

```c#
public static class HttpContextExtensions
{
    public static IOrganizationClaims GetCurrentContextOrganizationClaims(this HttpContext httpContext)
    {
        var orgClaims = httpContext.Features.Get<IOrganizationClaims>();
        
        if (orgClaims != null)
        {
            // Organization claims have already been parsed for this request, return it
            return claims;
        }


        orgClaims = ExpensiveParsing(httpContext.User.Claims);
        httpContext.Features.Set<IOrganizationClaims>(orgClaims);
        return orgClaims;
    }
}
```

eliykat:
@justindbaur I still need to write some more tests, but I've implemented your suggestions to avoid using `ICurrentContext` - can you please take another look? This avoids changing the middleware pipeline order, but also provides a path for deprecating much of the organization code in `CurrentContext`.

JimmyVo16 (dismissed):
I like this pattern a lot. I think it makes things a lot simpler, and we're following established patterns.

This actually inspired me to use an ActionFilter for some of the validation, like `Customer users can't remove Admins.` We can chat more about it on Slack to not distract from this PR.

src/Api/AdminConsole/Authorization/OrganizationRequirementHelpers.cs
```diff
@@ -0,0 +1,20 @@
+﻿#nullable enable
+
+namespace Bit.Api.AdminConsole.Authorization;
+
+public static class OrganizationRequirementHelpers
```
JimmyVo16, line 5: Minor non-blocker: for these helper classes/methods, maybe we should name them after the HTTP context rather than the Organization Requirement? I can see us using these helpers for things beyond the Organization Requirement.

eliykat: This has been rolled into the `HttpContextExtensions` class with the other extension methods.

eliykat:
This is ready for review again. Everything should be fully tested. I reverted changes to the controller so that this is only adding the new functionality, not wiring it up, and no QA should be required at this stage. The only changes to production code are to move the permission claims into the `Claims` class.

src/Api/AdminConsole/Authorization/Requirements/ManageUsersRequirement.cs
```diff
@@ -0,0 +1,17 @@
...
+
+public class ManageUsersRequirement : IOrganizationRequirement
+{
+    public async Task<bool> AuthorizeAsync(
+        CurrentContextOrganization? organizationClaims,
+        Func<Task<bool>> isProviderUserForOrg)
+        => organizationClaims is
+    { Type: OrganizationUserType.Owner or OrganizationUserType.Admin } or
+    { Permissions.ManageUsers: true }
+            || await isProviderUserForOrg();
```
audreyality, line 16: 🎨 This code is difficult to read. Consider using a switch expression.

```suggestion
public async Task<bool> AuthorizeAsync(
    CurrentContextOrganization? organizationClaims,
    Func<Task<bool>> isProviderUserForOrg)
    => organizationClaims switch
    {
        { Type: OrganizationUserType.Owner } => true,
        { Type: OrganizationUserType.Admin } => true,
        { Permissions.ManageUsers: true } => true,
        _ => await isProviderUserForOrg()
    };
```

audreyality (commented):
👍🏻 I quite like how this is organized! The memoization of expensive operations is also a nice touch, and I'm excited that you're augmenting .Net built-ins!

src/Api/AdminConsole/Authorization/HttpContextExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+        if (cachedResult != null)
+        {
+            return cachedResult;
+        }
+
+        var result = await callback();
+        httpContext.Features.Set(result);
+
+        return result;
+    }
```
justindbaur, line 31: Nice!

src/Api/AdminConsole/Authorization/OrganizationClaimsExtensions.cs
```diff
@@ -0,0 +1,123 @@
...
+using Bit.Core.Models.Data;
+
+namespace Bit.Api.AdminConsole.Authorization;
+
+public static class OrganizationClaimsExtensions
+{
+    /// <summary>
+    /// A delegate that returns true if the user has the specified claim type for an organization, false otherwise.
+    /// </summary>
+    private delegate bool HasClaim(string claimType);
```
justindbaur, line 16: I wouldn't do a custom delegate and instead do `Func<string, bool>`. I don't think we technically follow the framework design guidelines here but I try to. 

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/events-and-callbacks

> DO use the new `Func<...>`, `Action<...>`, or `Expression<...>` types instead of custom delegates, when defining APIs with callbacks.

justindbaur (dismissed):
One recommendation but other than that looks good.

src/Api/AdminConsole/Authorization/Requirements/ManageUsersRequirement.cs
eliykat: Nice suggestion 🎨

src/Api/AdminConsole/Authorization/OrganizationClaimsExtensions.cs
eliykat: I've seen this before and don't quite understand it as a general rule. Sometimes it's nice to give things names rather than infer from context or xmldoc. Not a big deal either way though so I've implemented your feedback.

src/Api/AdminConsole/Authorization/AuthorizeAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <typeparam name="T">The IAuthorizationRequirement that will be used to authorize the user.</typeparam>
+public class AuthorizeAttribute<T>
+    : AuthorizeAttribute, IAuthorizationRequirementData
+    where T : IAuthorizationRequirement, new()
+{
+    public IEnumerable<IAuthorizationRequirement> GetRequirements()
+    {
+        var requirement = new T();
+        return [requirement];
+    }
```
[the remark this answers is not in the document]
eliykat: The nice thing about this attribute is that it's requirement agnostic, other than requiring a parameterless constructor (`new()`). It's not actually tied to `IOrganizationRequirement` - any `IAuthorizationRequirement` will do. It just news it up and gives it to the middleware.

If you wanted to implement a more complex requirement, you would be implementing your own Requirement and Requirement Handler, not an attribute. (_Or_ you'd be doing something totally different and writing your own attribute anyway.)

Given this I don't think there's anything to be gained by this extra layer.

src/Api/AdminConsole/Authorization/OrganizationClaimsExtensions.cs
eliykat: Thanks for the explanation!

src/Api/AdminConsole/Authorization/AuthorizeAttribute.cs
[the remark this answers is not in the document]
audreyality: > If you wanted to implement a more complex requirement, you would be implementing your own Requirement and Requirement Handler, not an attribute. (Or you may be writing your own attribute if you wanted to do something tricky, but then that's probably not extending anything I'm doing here.)

If there are multiple requirements, you would inherit from the `CustomAuthorizeAttribute` and emit each of them from `GetRequirements()`.

audreyality (approved):

JimmyVo16 (approved):

eliykat:
Requesting a sign-off from @bitwarden/team-auth-dev because they _should_ own `Claims.cs` and this is fairly critical to authentication.

ike-kottlowski (approved):
Thanks for looping in Auth on this! 

I think this will be a great improvement.

PR #5652, opened 2025-04-14
PM-18939 refactoring send service to 'cqrs'

prograhamming:
## 🎟️ Tracking

(https://bitwarden.atlassian.net/browse/PM-18939)

## 📔 Objective

Refactor current send codebase to align with our version of CQRS

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Tools/SendFeatures/Commands/AnonymousSendCommand.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    // Response: Send, password required, password invalid
+    public async Task<(string, bool, bool)> GetSendFileDownloadUrlAsync(Send send, string fileId, string password)
+    {
+        if (send.Type != SendType.File)
+        {
+            throw new BadRequestException("Can only get a download URL for a file type of Send");
+        }
+
+        var (grantAccess, passwordRequired, passwordInvalid) = _sendAuthorizationService.SendCanBeAccessed(send, password);
```
audreyality, line 39: 🌱 This pattern might need to be adjusted since we're moving to an authorization token model. How might this change in a situation where `send` can't load without a proper claim? (i.e. from a JWT bearer token received in an auth header.)

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
```diff
@@ -0,0 +1,103 @@
...
+        ICurrentContext currentContext)
+    {
+        _sendRepository = sendRepository;
+        _passwordHasher = passwordHasher;
+        _pushNotificationService = pushNotificationService;
+        _referenceEventService = referenceEventService;
+        _currentContext = currentContext;
+    }
+
+    public (bool grant, bool passwordRequiredError, bool passwordInvalidError) SendCanBeAccessed(Send send,
```
audreyality, line 34: 🤔 Is there a higher-level type you could output here? The fields appear to be mutually exclusive, which leads me to believe this should be an `Enum`.

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        var send = await _sendRepository.GetByIdAsync(sendId);
+        var (grantAccess, passwordRequired, passwordInvalid) = SendCanBeAccessed(send, password);
+
+        if (!grantAccess)
+        {
+            return (null, passwordRequired, passwordInvalid);
+        }
+
+        // TODO: maybe move this to a simple ++ sproc?
```
audreyality, line 75: 📝 We're moving away from a database model towards a Json model, so this shouldn't be a sproc.

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        // Update Send data if necessary
+        if (realSize != fileData.Size)
+        {
+            fileData.Size = realSize.Value;
+        }
+        fileData.Validated = true;
+        send.Data = JsonSerializer.Serialize(fileData,
+            JsonHelpers.IgnoreWritingNull);
```
audreyality, line 90: ⛏️ The linter is being a bit silly here (and a few other places).

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -0,0 +1,178 @@
...
+            if (user.Premium)
+            {
+                storageBytesRemaining = user.StorageBytesRemaining();
+            }
+            else
+            {
+                // Users that get access to file storage/premium from their organization get the default
+                // 1 GB max storage.
+                storageBytesRemaining = user.StorageBytesRemaining(
+                    _globalSettings.SelfHosted ? (short)10240 : (short)1);
```
audreyality, line 162: 🎨 I'd extract the ternary into its own variable here.

```suggestion
                short limit = _globalSettings.SelfHosted ? (short)10240 : (short)1;
                storageBytesRemaining = user.StorageBytesRemaining(limit);
```

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
```diff
@@ -0,0 +1,1157 @@
...
+    // {
+    //     send.Id = default;
+    //     send.Type = sendType;
+    //
+    //     sutProvider.GetDependency<IPolicyService>().AnyPoliciesApplicableToUserAsync(
+    //         Arg.Any<Guid>(), PolicyType.DisableSend).Returns(disableSendPolicyAppliesToUser);
+    //
+    //     sutProvider.GetDependency<IUserRepository>().GetByIdAsync(Arg.Any<Guid>()).Returns(user);
+    //     sutProvider.GetDependency<ISendValidationService>().ValidateUserCanSaveAsync(send.UserId, send);
+    // }
```
audreyality, line 83: ⚠️ TODO: REMOVE

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
```diff
@@ -0,0 +1,1157 @@
...
+        {
+            // For existing Sends
+            await _sendRepository.Received(1).UpsertAsync(send);
+            Assert.NotEqual(initialDate, send.RevisionDate);
+            await _pushNotificationService.Received(1).PushSyncSendUpdateAsync(send);
+            await _referenceEventService.DidNotReceive().RaiseEventAsync(Arg.Any<ReferenceEvent>());
+        }
+    }
+
+    // TODO: REMOVE
```
audreyality, line 178: ⛏️ Is this just a comment that should've been removed?

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
```diff
@@ -0,0 +1,1157 @@
...
```
audreyality, line 300: 📝 I'm not going to point out more of these, but you should search for them.

💭 If this is tech debt, use `FIXME` and explain the future action needed. In this case, it would be the conditions under which the code should be removed.

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
```diff
@@ -0,0 +1,1157 @@
...
+
+        Assert.Contains("not of type \"file\"", exception.Message);
+
+        // Verify no further methods were called
+        await _sendValidationService.DidNotReceive().StorageRemainingForSendAsync(Arg.Any<Send>());
+        await _sendRepository.DidNotReceive().CreateAsync(Arg.Any<Send>());
+        await _sendRepository.DidNotReceive().UpsertAsync(Arg.Any<Send>());
+        await _sendFileStorageService.DidNotReceive().GetSendFileUploadUrlAsync(Arg.Any<Send>(), Arg.Any<string>());
+        await _pushNotificationService.DidNotReceive().PushSyncSendCreateAsync(Arg.Any<Send>());
+        await _pushNotificationService.DidNotReceive().PushSyncSendUpdateAsync(Arg.Any<Send>());
```
audreyality, line 573: 🎨 This can be done more elegantly by configuring the object to throw when these methods are called.

audreyality (changes requested):
⚠️ Public and protected properties that you introduced in this PR should have XMLdoc comments.

👍🏻 The bulk of this code is looking really great! Thank you for partitioning the send service. This code's easier to follow than the service.

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Bit.Core.Tools.Entities;
+using Bit.Core.Tools.Models.Data;
+
+namespace Bit.Core.Tools.SendFeatures.Commands.Interfaces;
+
+public interface INonAnonymousSendCommand
```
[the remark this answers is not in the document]
prograhamming: CQRS doesn't strictly dictate that each command should only have one method. The core principles of CQRS are:
- Separate read operations (queries) from write operations (commands)
- Commands modify state but don't return data
- Queries return data but don't modify state

A lot of .NET folks use MediatR with the CQRS design pattern and typically you will see one command under one interface/class. However, it is valid to have command classes with multiple methods as **long as they're all commands (changing state) and logically belong together.**

I separated the commands by how they are used in the controller (Anonymous/non-anonymous). I can change this, but I kept them to together since don't structure our backend code like MediatR. 

Also, In `NonAnonymousSendCommand.SaveFileSendAsync()` I am cheating by returning a string, but didn't want to change the code too much for the 1st refactor with the send stuff you are working on. It feels like we using Controllers to be mediators or orchestrators and strong opinions about throwing only in the controller.

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
prograhamming: Fixed

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
prograhamming: Fixed.

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
prograhamming: Removed all of the TODO: REMOVE.

src/Core/Tools/SendFeatures/Services/Interfaces/ISendAuthorizationService.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Bit.Core.Tools.Entities;
+
+namespace Bit.Core.Tools.Services;
+
+public interface ISendAuthorizationService
+{
+    Task<(Send, bool, bool)> AccessAsync(Guid sendId, string password);
+    public (bool grant, bool passwordRequiredError, bool passwordInvalidError) SendCanBeAccessed(Send send,
```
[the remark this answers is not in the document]
prograhamming: Fixed syntax error.

src/Core/.idea/.idea.Core.dir/.idea/.gitignore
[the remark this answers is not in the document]
prograhamming: Remove these files.

src/Core/Tools/SendFeatures/Services/Interfaces/ISendValidationService.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Bit.Core.Tools.Services;
+
+public interface ISendValidationService
+{
+    long GetMaxFileSize();
+    string GetMaxFileSizeReadable();
+    Task<bool> ValidateSendFile(Send send);
+    Task ValidateUserCanSaveAsync(Guid? userId, Send send);
+    Task ValidateUserCanSaveAsync_vNext(Guid? userId, Send send);
```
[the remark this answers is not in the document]
audreyality: So long as it won't cause other adverse effects, I think I'd like this updated to use the strategy pattern.

src/Core/Tools/SendFeatures/Commands/AnonymousSendCommand.cs
prograhamming: We could use IHttpContextAccessor within this service to be able to check the roles of the user. I think this method could have these pieces refactored:

-  Move the 'validation' pieces to the ISendValidationService to check user input (i.e lines 38-40 and IsNullOrWhiteSpace checks)
- Verification of the password would be moved the `GetSendFileDownloadUrlAsync`
- SendCanBeAccessed would just be concerned with the claims.

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
prograhamming: Fixed

audreyality: I should have been more precise. If your substitutes **throw by default**, then you can override just those methods that you should be calling to provide your desired behaviors. This was easy to set up in Moq; it's not clear how to set it up in NSubstitute. The goal is to eliminate these blocks completely.

As written, this is wrong. The `throw` calls aren't assertions, they're behaviors. As such, in order to be effective they need to be in the "arrange" stage of your test.

If NSubstitute can't provide the "unless I say otherwise, throw" behavior, then your original assertions are preferable.

src/Core/Tools/SendFeatures/Commands/AnonymousSendCommand.cs
audreyality: > We could use IHttpContextAccessor within this service to be able to check the roles of the user.

I'd prefer to use something independent of the execution context (ie. an HTTP request). The controller has access to the `HttpContext`, for example, and could extract the claim.

> Verification of the password would be moved the GetSendFileDownloadUrlAsync

This won't work. The password won't be given to those URLs. Instead, they'll have a bearer token containing a claim authorizing access to the download.

test/Core.Test/Tools/Services/NonAnonymousSendCommandTests.cs
prograhamming: Reverted to the previous code.

prograhamming:
Updated with XMLdoc comments

src/Core/Tools/SendFeatures/Services/Interfaces/ISendStorageService.cs
```diff
@@ -3,14 +3,61 @@
...
 public interface ISendFileStorageService
 {
     FileUploadType FileUploadType { get; }
+    /// <summary>
+    ///  Uploads a new file to the storage.
+    /// </summary>
+    /// <param name="stream"></param>
+    /// <param name="send"></param>
+    /// <param name="fileId"></param>
+    /// <returns></returns>
```
audreyality, line 19: ⛏️ The parameters and return value should be documented in addition to the summary.

src/Core/Tools/SendFeatures/Services/Interfaces/ISendValidationService.cs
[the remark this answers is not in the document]
prograhamming: I think I am going to keep the current code for this due to the feature flag.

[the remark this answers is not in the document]
audreyality: ⚠️ This feedback wasn't properly addressed. The strategy pattern was not used; the code was inlined. This decreases its isolation and makes it harder to test. Either restore the `vNext` item or use the strategy pattern.

audreyality (changes requested):

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
audreyality: ⛏️ This "TODO" comment should be deleted.

src/Core/Tools/SendFeatures/Commands/AnonymousSendCommand.cs
audreyality: I'm going to resolve this and figure out an authentication-compatible API when we move to the auth token flow.

prograhamming:
Reverted to restore `ValidateUserCanSaveAsync_vNext` inside of strategy pattern.

src/Core/Tools/SendFeatures/Services/Interfaces/ISendStorageService.cs
```diff
@@ -13,51 +13,53 @@ public interface ISendFileStorageService
...
     /// </summary>
-    /// <param name="send"></param>
-    /// <param name="fileId"></param>
-    /// <param name="expectedFileSize"></param>
-    /// <param name="leeway"></param>
-    /// <returns></returns>
+    /// <param name="send">Send object help validate file</param>
+    /// <param name="fileId">File id to identify which file to validate</param>
+    /// <param name="expectedFileSize">Expected file size of the file</param>
+    /// <param name="leeway">Leeway will be used as tolerance range when validating the file size</param>
```
audreyality, line 60: ❓ What are its units?

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// Gets the Send file download URL for a Send object.
+    /// </summary>
+    /// <param name="send">Send object to help get file download url and validate file</param>
+    /// <param name="fileId">FileId get file download url</param>
+    /// <param name="password">Password will be validated and used to determine access</param>
+    /// <returns>Async Task object with Tuple containing the string of download url, boolean that identifies if
+    /// passwordRequiredError occurred, and another boolean that identifies if passwordInvalidError occurred.
+    /// </returns>
```
audreyality, line 18: ⛏️ Instead of saying "that identifies if" say "is `true` when". It's more accurate, and the only identifier we're working with is the fileId.

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+/// <summary>
+/// NonAnonymousSendCommand interface provides methods for managing non-anonymous Sends.
+/// </summary>
+public interface INonAnonymousSendCommand
+{
+    /// <summary>
+    /// Saves a Send to the database.
+    /// </summary>
+    /// <param name="send">Send object that is to be saved to database</param>
+    /// <returns>Task object for async operations</returns>
```
audreyality, line 15: ⛏️ This should indicate when the task completes.

```suggestion
    /// <returns>A task that completes when the save is successful or throws if the save is unsuccessful.</returns>
```

❓ Does a throw guarantee no record was written?

📝 All plain tasks (those without a type parameter) should answer these questions.

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    /// <param name="send">Send object that is to be saved to database</param>
+    /// <param name="data">SendFileData to be saved to database</param>
+    /// <param name="fileLength">Length of file help with saving to database</param>
+    /// <returns>Task object for async operations with file upload url</returns>
+    Task<string> SaveFileSendAsync(Send send, SendFileData data, long fileLength);
+    /// <summary>
+    /// Upload a file to an existing Send.
+    /// </summary>
+    /// <param name="stream">Stream of file to be uploaded</param>
```
audreyality, line 29: 🤔 This should probably indicate what happens when a stream is given to the method whose `.Position > 0`.

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
```diff
@@ -0,0 +1,20 @@
...
+/// AnonymousSendCommand interface provides methods for managing anonymous Sends.
+/// </summary>
+public interface IAnonymousSendCommand
+{
+    /// <summary>
+    /// Gets the Send file download URL for a Send object.
+    /// </summary>
+    /// <param name="send">Send object to help get file download url and validate file</param>
+    /// <param name="fileId">FileId get file download url</param>
+    /// <param name="password">Password will be validated and used to determine access</param>
```
audreyality, line 15: ⛏️ You're describing the expected behavior and usage of these parameters. Present tense is more clear. 

🎨 You should also communicate any properties of the parameter that are required but not captured in its type.

```suggestion
    /// <param name="password">A hashed and base64-encoded password. This is compared with the send's password to authorize access.</param>
```

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    /// <param name="send">Send object that is to be saved to database</param>
+    /// <returns>Task object for async operations</returns>
+    Task SaveSendAsync(Send send);
+
+    /// <summary>
+    /// Save File to the database.
+    /// </summary>
+    /// <param name="send">Send object that is to be saved to database</param>
+    /// <param name="data">SendFileData to be saved to database</param>
```
audreyality, line 22: ⚠️ This method says it saves the file in the summary, and the send in the parameter name. The method body looks like it does both. The summary should be updated.

⛏️ When referring to a type, you should use `<c cref="TypeName" />`

```suggestion
    /// <param name="send"><see cref="Send" /> that is to be saved to database</param>
    /// <param name="data"><see cref="SendFileData" /> to be saved to database</param>
```

src/Core/Tools/SendFeatures/Services/Interfaces/ISendStorageService.cs
prograhamming: Leeway is a number, but looks like it is used as a tolerance range when validating the file size. It allows for some flexibility in the comparison of the actual file size (length) against the expectedFileSize. I think it is useful in scenarios where minor discrepancies in file size might occur due to encoding, metadata, or other factors.

Minimum: expectedFileSize - leeway
Maximum: expectedFileSize + leeway

If the file size is outside this range, then the method returns (false, length) to indicate the file size is invalid.

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
prograhamming: fixed

src/Core/Tools/SendFeatures/Services/Interfaces/ISendStorageService.cs
audreyality: So does that mean it's measured in bytes?

prograhamming: Yes, but it is a long data type within the code.

audreyality:
```suggestion
    /// <param name="leeway">
    /// Send file size tolerance in bytes. When an uploaded file's `expectedFileSize`
    /// is outside of the leeway, the storage operation fails.
    /// </param>
    /// <throws>
    /// ❌ Fill this in with an explanation of the error thrown when `leeway` is incorrect
    /// </throws>
```

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
prograhamming: Added syntax in XMLdocs when referring to type.

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
prograhamming: Add verbiage

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
prograhamming: Added this to param summary

audreyality: You noted what it assumes, not how it behaves. This is incorrect because you need to read the code to understand what the effect of that assumption is.

How it behaves is **it saves the file from the current stream position**.

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
audreyality: This hasn't been addressed yet.

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
audreyality: This hasn't been addressed yet.

audreyality (commented):
There's a few more open items. Nothing new.

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
prograhamming: removed todo comment

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
prograhamming: hmmmm interesting. This is what I have locally -> 
``` c#
 /// <summary>
    /// Gets the Send file download URL for a Send object.
    /// </summary>
    /// <param name="send"><see cref="Send" /> used to help get file download url and validate file</param>
    /// <param name="fileId">FileId get file download url</param>
    /// <param name="password">A hashed and base64-encoded password. This is compared with the send's password to authorize access.</param>
    /// <returns>Async Task object with Tuple containing the string of download url and <see cref="SendAccessResult" />
    /// to determine if the user can access send. There are 4 possible results: SendAccessResult.Granted,
    /// SendAccessResult.PasswordRequired,SendAccessResult.PasswordInvalid, and SendAccessResult.Denied
    /// </returns>
    Task<(string, SendAccessResult)> GetSendFileDownloadUrlAsync(Send send, string fileId, string password);
```

src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs
prograhamming: Fixed this

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
audreyality: > There are 4 possible results: SendAccessResult.Granted,
> SendAccessResult.PasswordRequired,SendAccessResult.PasswordInvalid, and SendAccessResult.Denied

This doesn't need to be stated in your comment, since those are the values from the enum.

audreyality (commented):
Looking great! Just a few more comment tweaks and we're good to go!

src/Core/Tools/SendFeatures/Commands/Interfaces/IAnonymousSendCommand.cs
prograhamming: updated

src/Core/Tools/Models/Data/SendAccessResult.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Bit.Core.Tools.Models.Data;
+
+public enum SendAccessResult
```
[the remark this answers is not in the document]
audreyality: Don't forget the members!

src/Core/Tools/Models/Data/SendAccessResult.cs
audreyality, line 1: 📝 There are a bunch of XML comment annotations here that need to be addressed.

src/Core/Tools/Models/Data/SendAccessResult.cs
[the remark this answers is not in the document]
prograhamming: members have been implemented.

src/Core/Tools/Models/Data/SendAccessResult.cs
prograhamming: fixed them

src/Core/Tools/Models/Data/SendAccessResult.cs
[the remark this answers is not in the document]
prograhamming: members have been implemented.

src/Core/Tools/Models/Data/SendAccessResult.cs
prograhamming: fixed them

prograhamming: fixed them

audreyality (dismissed):

Thomas-Avery (approved):
KM changes LGTM

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
```diff
@@ -39,4 +39,11 @@ public interface INonAnonymousSendCommand
     /// <param name="send"><see cref="Send" /> is used to delete from database and file storage</param>
     /// <returns>Task completes once <see cref="Send" /> has been deleted from database and file storage.</returns>
     Task DeleteSendAsync(Send send);
+
+    /// <summary>
+    /// Validates the Send file, but will delete or update send based on validation.
+    /// </summary>
+    /// <param name="send"><see cref="Send" /> needed to validate file and update or delete <see cref="Send" /></param>
+    /// <returns>Boolean whether the file was valid or not</returns>
+    Task<bool> UpdateSendOnValidation(Send send);
```
audreyality, line 48: ⚠️ This method does too much. It shouldn't both delete and update a send. Split it into two separate methods, one does the delete (when validation fails?) and the other an update (when validation succeeds?).

src/Core/Tools/SendFeatures/Services/Interfaces/ISendFileSettingHelper.cs
audreyality, line 1: ❌ This file needs to include units of measurement for the file size. Presumably the size is measured in bytes? (As opposed to KB/KiB or MB/MiB.)

src/Core/Tools/SendFeatures/Services/SendFileSettingHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Bit.Core.Tools.Services;
+
+namespace Bit.Core.Tools.SendFeatures;
+
+public class SendFileSettingHelper : ISendFileSettingHelper
+{
+    private readonly long _fileSizeLeeway = 1024L * 1024L; // 1MB
+    private readonly long MAX_FILE_SIZE = Constants.FileSize501mb;
+    private readonly string MAX_FILE_SIZE_READABLE = "500 MB";
```
audreyality, line 9: ⛏️ Seems like this constant should be referred to in the `SendsController` when the file size is too big.

src/Core/Tools/SendFeatures/Services/SendFileSettingHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public long GetMaxFileSize()
```
audreyality, line 11: ⛏️ `GetMaxFileSize`, `GetMaxFileSizeReadable`, and `GetFileSizeLeeway` should be readonly properties unless there's a compelling reasonm to keep them as functions.

audreyality (changes requested):
👍🏻 Overall this is looking pretty good. 

You mainly got dinged on not keeping up with my prior feedback.

src/Core/Tools/SendFeatures/Services/Interfaces/ISendFileSettingHelper.cs
prograhamming: I added documentation to the files to give better insights to what they are.

src/Core/Tools/SendFeatures/Services/SendFileSettingHelper.cs
prograhamming: Made this happen with changing SendFileSettingHelper to a static class.

src/Core/Tools/SendFeatures/Services/SendFileSettingHelper.cs
prograhamming: I made SendFileSettingHelper into a static class since for now it will just be holding constant fields. In the future it can be a place to hold helper methods as well, but figured a static class was a better fit for the current data and can be used within the SendController without needing to add a constant variable there.

src/Core/Tools/SendFeatures/Commands/Interfaces/INonAnonymousSendCommand.cs
audreyality: This feedback was resolved offline. This was resolved by renaming the method to `ConfirmFileSize`, which more effectively describes the outcome of running the command.

audreyality (dismissed):
👍🏻 Thanks for investing so much time on this @prograhamming!

Thomas-Avery (approved):
KM changes look good.

src/Core/Tools/SendFeatures/Services/SendFileSettingHelper.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Bit.Core.Tools.SendFeatures;
+﻿using Bit.Core.Tools.Entities;
```
audreyality, line 1: ❓ Why wasn't this import needed before?

audreyality (approved):
Just a question. Non-blocking.

src/Core/Tools/SendFeatures/Services/SendFileSettingHelper.cs
prograhamming: JetBrains Rider IDE didn't pick it up beforehand. I'm not sure why it didn't. It is needed for the XMLdocs comment if you want to be able to click the 'Send' type and it take you to the correct type.

PR #5655, opened 2025-04-15
[PM-18237] Add RequireSsoPolicyRequirement

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18237

## 📔 Objective

> Add an IPolicyRequirement for this policy and connect it to dependency injection
> 
> Update the code that currently enforces this policy so that it uses the new IPolicyRequirementQuery instead of the current service or repository call (if the feature flag is enabled).
> 
> Update tests.

The `WebAuthnController` lets users with the `Accepted` status create passkeys, but `BaseRequestValidator` requires users to be `Confirmed` to log in. To handle this difference, I added separate properties in `RequireSsoPolicyRequirement`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/Auth/Controllers/WebAuthnController.cs
```diff
@@ -118,7 +124,9 @@ public async Task Post([FromBody] WebAuthnLoginCredentialCreateRequestModel mode
 
     private async Task ValidateRequireSsoPolicyDisabledOrNotApplicable(Guid userId)
     {
-        var requireSsoLogin = await _policyService.AnyPoliciesApplicableToUserAsync(userId, PolicyType.RequireSso);
+        var requireSsoLogin = _featureService.IsEnabled(FeatureFlagKeys.PolicyRequirements)
+            ? (await _policyRequirementQuery.GetAsync<RequireSsoPolicyRequirement>(userId)).CanUsePasskeyLogin
+            : await _policyService.AnyPoliciesApplicableToUserAsync(userId, PolicyType.RequireSso);
```
eliykat, line 129: I think the boolean is flipped here because of the change in interfaces.

Today the error is thrown if `requireSsoLogin == true`, but after the refactor it would be thrown if `CanUsePassKeyLogin == false`.

It might be clearer to make a `vNext` method.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirement.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    public override PolicyType PolicyType => PolicyType.RequireSso;
+
+    protected override IEnumerable<OrganizationUserType> ExemptRoles =>
+        _globalSettings.Sso.EnforceSsoPolicyForAllUsers
+            ? Array.Empty<OrganizationUserType>()
+            : [OrganizationUserType.Owner, OrganizationUserType.Admin];
+
+    protected override IEnumerable<OrganizationUserStatusType> ExemptStatuses =>
+        Array.Empty<OrganizationUserStatusType>();
```
eliykat, line 52: I think we still want to exempt revoked and invited users. I guess it doesn't matter too much because the methods on the requirement do their own filtering, but imo it's still good practice to filter out everything that we know won't apply at this stage.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirement.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        _policyDetails = policyDetails;
+    }
+
+    /// <summary>
+    /// Indicates whether the user can use passkey login.
+    /// Policy is enforced for users with status >= Accepted.
+    /// </summary>
+    public bool CanUsePasskeyLogin => _policyDetails.Any(p =>
+        p.OrganizationUserStatus >= OrganizationUserStatusType.Accepted);
```
eliykat, line 24: Maybe this is why the tests are passing... this bool is also flipped. You can only use passkey login if RequireSSO _doesn't_ apply to you, but this returns true if it _does_ apply. So it's flipped here and then flipped back in the caller.

I think this should be `!_policyDetails.Any`.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirementFactoryTests.cs
```diff
@@ -0,0 +1,79 @@
...
+[SutProviderCustomize]
+public class RequireSsoPolicyRequirementFactoryTests
+{
+    [Theory, BitAutoData]
+    public void CanUsePasskeyLogin_WithNoPolicies_ReturnsFalse(SutProvider<RequireSsoPolicyRequirementFactory> sutProvider)
+    {
+        var actual = sutProvider.Sut.Create([]);
+
+        Assert.False(actual.CanUsePasskeyLogin);
+    }
```
eliykat, line 20: Further to my comments above - this business logic is wrong. You can always use passkey login, unless the policy forbids it. So no policies should mean this returns true.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirementFactoryTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    [Theory, BitAutoData]
+    public void SsoRequired_WithInvalidPolicies_ReturnsFalse(
+        SutProvider<RequireSsoPolicyRequirementFactory> sutProvider)
+    {
+        var actual = sutProvider.Sut.Create(
+        [
+            new PolicyDetails { PolicyType = PolicyType.RequireSso, OrganizationUserStatus = OrganizationUserStatusType.Revoked },
+            new PolicyDetails { PolicyType = PolicyType.RequireSso, OrganizationUserStatus = OrganizationUserStatusType.Invited },
+            new PolicyDetails { PolicyType = PolicyType.RequireSso, OrganizationUserStatus = OrganizationUserStatusType.Accepted }
```
eliykat, line 64: This would be better as theory data, as any one of these should cause this result. Better to test them one-by-one.

test/Identity.Test/IdentityServer/BaseRequestValidatorTests.cs
```diff
@@ -276,6 +281,99 @@ public async Task ValidateAsync_GrantTypes_OrgSsoRequiredTrue_ShouldSetSsoResult
...
+        // Configure requirement to require SSO and exempt accepted users
+        var requirement = new RequireSsoPolicyRequirement(
+            [
+                new PolicyDetails
+                {
+                    PolicyType = PolicyType.RequireSso,
+                    OrganizationUserStatus = OrganizationUserStatusType.Confirmed
+                }
+            ]);
+        _policyRequirementQuery.GetAsync<RequireSsoPolicyRequirement>(Arg.Any<Guid>()).Returns(requirement);
```
eliykat, line 311: This is more test setup than we want when we're not testing the requirement itself. For example, this validator class doesn't care about the `OrganizationUserStatus`, t just wants to make sure that it throws an error if `RequireSso == true`.

I suggest that the policy filtering logic is moved to the constructor of the `RequireSsoPolicyRequirement`, which sets literal values for `CanUsePasskeyLogin` and `RequireSso` rather than using getter logic.  Then make those properties init-only. That way we still get the same behavior, but in these tests you can just do `var requirement = new RequireSsoPolicyRequirement { RequireSso = true };` which is a much nicer interface.

eliykat (changes requested):
Great work, thanks for picking this one up!

test/Identity.Test/IdentityServer/BaseRequestValidatorTests.cs
r-tome: Exellent suggestion -- Its way simpler!

r-tome:
Closed and reopened because `workflow-linter.yml` is not triggered

r-tome:
@eliykat thanks for the great review! You were absolutely right, I had the logic written as the exact opposite as it should be 😅

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
```diff
@@ -348,9 +352,12 @@ private async Task<bool> RequireSsoLoginAsync(User user, string grantType)
...
         // Check if user belongs to any organization with an active SSO policy
-        var anySsoPoliciesApplicableToUser = await PolicyService.AnyPoliciesApplicableToUserAsync(
-                                                user.Id, PolicyType.RequireSso, OrganizationUserStatusType.Confirmed);
-        if (anySsoPoliciesApplicableToUser)
+        var ssoRequired = FeatureService.IsEnabled(FeatureFlagKeys.PolicyRequirements) ?
+            (await PolicyRequirementQuery.GetAsync<RequireSsoPolicyRequirement>(user.Id))
+                .SsoRequired
+            : await PolicyService.AnyPoliciesApplicableToUserAsync(
+                user.Id, PolicyType.RequireSso, OrganizationUserStatusType.Confirmed);
+        if (ssoRequired)
```
eliykat, line 360: Formatting here needs fixing, per our usual style:
```
var ssoRequired = condition
  ? case1
  : case2;
```

The `?` needs to be brought down a line.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirementFactoryTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        [
+            new PolicyDetails
+            {
+                PolicyType = PolicyType.RequireSso,
+                OrganizationUserStatus = userStatus
+            }
+        ]);
+
+        Assert.True(actual.CanUsePasskeyLogin);
+    }
```
eliykat, line 58: Not sure about the naming of these tests - you are a member of the organization even if you're revoked, for example.

I suggest `WithExemptStatus` / `WithoutExemptStatus`.

Same for the other tests below.

eliykat (changes requested):

eliykat (approved):
Great work!

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
JaredSnider-Bitwarden, line 1: Could you add a test to the BaseRequestValidatorTests for retrieving the policy data by the `PolicyRequirementQuery`?  We have an existing similar test in `ValidateAsync_GrantTypes_OrgSsoRequiredTrue_ShouldSetSsoResult`

JaredSnider-Bitwarden (changes requested):

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
eliykat: There are new tests for:
- `ValidateAsync_GrantTypes_WithPolicyRequirementsEnabled_OrgSsoRequiredTrue_ShouldSetSsoResult`
- `ValidateAsync_GrantTypes_WithPolicyRequirementsEnabled_OrgSsoRequiredFalse_ShouldSucceed`

Do they cover what you're thinking of, or are specific changes needed?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirement.cs
```diff
@@ -0,0 +1,60 @@
...
+        _globalSettings.Sso.EnforceSsoPolicyForAllUsers
+            ? Array.Empty<OrganizationUserType>()
+            : [OrganizationUserType.Owner, OrganizationUserType.Admin];
+
+    public override RequireSsoPolicyRequirement Create(IEnumerable<PolicyDetails> policyDetails)
+    {
+        var result = new RequireSsoPolicyRequirement
+        {
+            CanUsePasskeyLogin = policyDetails.All(p =>
+                p.OrganizationUserStatus < OrganizationUserStatusType.Accepted),
```
jrmccannon, line 53: Could we make these explicit? Using > or < on enums makes me itchy.

jrmccannon (approved):
Thank you. Looks good to me.

eliykat (approved):

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
JaredSnider-Bitwarden: Nope, those are perfect! Tyvm!

JaredSnider-Bitwarden (approved):

PR #5657, opened 2025-04-15
[PM-20140] error on deleting tde user from org

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-20140

## 📔 Objective
Prevent deleting users from an organization if they do not have a master password.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
@@ -30,7 +30,7 @@ public class RemoveOrganizationUserCommand : IRemoveOrganizationUserCommand
     public const string RemoveAdminByCustomUserErrorMessage = "Custom users can not remove admins.";
     public const string RemoveLastConfirmedOwnerErrorMessage = "Organization must have at least one confirmed owner.";
     public const string RemoveClaimedAccountErrorMessage = "Cannot remove member accounts claimed by the organization. To offboard a member, revoke or delete the account.";
-    public const string RemoveTDEUsersErrorMessage = "Cannor remove member accounts with TDE enabled";
+    public const string RemoveTDEUsersErrorMessage = "Cannot remove member accounts with TDE enabled";
 
     public RemoveOrganizationUserCommand(
         IDeviceRepository deviceRepository,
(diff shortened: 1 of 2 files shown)


src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
eliykat, line 1: This method is used by Directory Connector. I don't think we can throw here, or it'll abort the whole sync and the customer won't be able to do much about it. Instead these users should be filtered out from the `removeUsersSet`, like owners are today.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RemoveOrganizationUserCommand.cs
```diff
@@ -197,6 +201,12 @@ await _pushRegistrationService.DeleteUserRegistrationOrganizationAsync(devices,
     {
         var orgUsers = await _organizationUserRepository.GetManyAsync(organizationUsersId);
         var filteredUsers = orgUsers.Where(u => u.OrganizationId == organizationId).ToList();
+        var users = await _userRepository.GetManyAsync(filteredUsers.Select(u => u.Id));
+
+        if (users.Any(u => u.HasMasterPassword() == false))
+        {
+            throw new BadRequestException(RemoveTDEUsersErrorMessage);
+        }
```
eliykat, line 209: This command is a real jumble of code paths. But I don't think this is executed by all public methods - only the 2 "bulk" public methods.

I suggest
- moving to a new private method
- calling that private method from here (used by the 2 bulk methods)
- also calling it from `ValidateRemoveUser` (seems to be used by all other methods?)

But please check and make sure this covers all code paths.

eliykat (changes requested):
> Should we do this same checking for operations that failed? For example, in the InviteOrganizationUsersCommand if there was an exception in InviteOrganizationUsersAsync we call DeleteManyAsync to "revert" the changes that were in progress. Should we do this same check here?

No - the risk we're covering here is that users will be locked out of their account if they're removed from the organization. An invite that is briefly created and then reverted won't cause this situation because the user hasn't accepted it and removed their MP yet. So that's fine.

eliykat:
PS. please add unit tests and at least 1 integration test (see `Api.IntegrationTest`, let me know if you have any questions).

eliykat (commented):
After my initial feedback above, I double checked how it works today and left a comment on the ticket seeking some feedback from Product and Integration. This will probably require some further changes based on what they say, so we can wait for that feedback first.

eliykat:
Please see discussion on the Jira ticket - we've decided that this is not required for now, we'll re-release Directory Connector once we've addressed the actual bug. In the medium term we'll look at moving Directory Connector to using revoke instead of remove, but that can be non-blocking.

PR #5690, opened 2025-04-22
[PM-20064] Add cascade deletion for cipher with tasks

nick-livefront:
## 🎟️ Tracking

[PM-20064](https://bitwarden.atlassian.net/browse/PM-20064)

## 📔 Objective

Permanently deleting a cipher is currently failing when the cipher has a security task due to the foreign key constraint.
- Add cascade deletions for Notifications and Security Tasks

## 📸 Screenshots

Deletions from the UI:
<video src="https://github.com/user-attachments/assets/84642757-b6ef-4fe7-87fd-c4de58bdf188" />

## 🦮 Reviewer guidelines

[PM-20064]: https://bitwarden.atlassian.net/browse/PM-20064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Infrastructure.EntityFramework/Vault/Configurations/SecurityTaskEntityTypeConfiguration.cs
+++ b/src/Infrastructure.EntityFramework/Vault/Configurations/SecurityTaskEntityTypeConfiguration.cs
@@ -25,6 +25,18 @@ public class SecurityTaskEntityTypeConfiguration : IEntityTypeConfiguration<Secu
             .IsClustered(false);
 
         builder
+            .HasOne(p => p.Organization)
+            .WithMany()
+            .HasForeignKey(p => p.OrganizationId)
+            .OnDelete(DeleteBehavior.Cascade);
+
+        builder
+            .HasOne(p => p.Cipher)
+            .WithMany()
+            .HasForeignKey(p => p.CipherId)
+            .OnDelete(DeleteBehavior.Cascade);
+
+        builder
             .ToTable(nameof(SecurityTask));
     }
 }
--- a/src/Sql/NotificationCenter/dbo/Tables/Notification.sql
+++ b/src/Sql/NotificationCenter/dbo/Tables/Notification.sql
@@ -14,7 +14,7 @@ CREATE TABLE [dbo].[Notification]
     CONSTRAINT [PK_Notification] PRIMARY KEY CLUSTERED ([Id] ASC),
     CONSTRAINT [FK_Notification_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]),
     CONSTRAINT [FK_Notification_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id]),
-    CONSTRAINT [FK_Notification_SecurityTask] FOREIGN KEY ([TaskId]) REFERENCES [dbo].[SecurityTask] ([Id])
+    CONSTRAINT [FK_Notification_SecurityTask] FOREIGN KEY ([TaskId]) REFERENCES [dbo].[SecurityTask] ([Id]) ON DELETE CASCADE,
 );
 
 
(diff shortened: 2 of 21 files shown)


src/Infrastructure.EntityFramework/Vault/Configurations/SecurityTaskEntityTypeConfiguration.cs
```diff
@@ -24,6 +24,18 @@ public void Configure(EntityTypeBuilder<SecurityTask> builder)
             .HasIndex(s => s.CipherId)
             .IsClustered(false);
 
+        builder
```
withinfocus, line 27: ℹ️ We're trying to move top-level connections to `dbo.Organization` to the delete proc -- see one example at https://github.com/bitwarden/server/pull/5602/files#diff-23455d0afecf5d129c0181b1545efe42715cdaf3bff9a7f81a34bca7979e4f03. This is a mixed bag here so it's not strictly required, but we feel it's the "new way".

util/Migrator/DbScripts/2025-03-21_00_NotificationCascadeDelete.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN
+    IF EXISTS (SELECT 1 FROM sys.foreign_keys WHERE name = 'FK_Notification_SecurityTask')
```
withinfocus, line 2: ❓ My approach recently to do this is at https://github.com/bitwarden/server/pull/5602/files#diff-a83ace09075d386bef412538689ffa356b1400a82f05767e10604bf18c955219R2 -- @rkac-bw which check is best? No difference?

src/Sql/NotificationCenter/dbo/Tables/Notification.sql
```diff
@@ -14,7 +14,7 @@ CREATE TABLE [dbo].[Notification]
     CONSTRAINT [PK_Notification] PRIMARY KEY CLUSTERED ([Id] ASC),
     CONSTRAINT [FK_Notification_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]),
     CONSTRAINT [FK_Notification_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id]),
-    CONSTRAINT [FK_Notification_SecurityTask] FOREIGN KEY ([TaskId]) REFERENCES [dbo].[SecurityTask] ([Id])
+    CONSTRAINT [FK_Notification_SecurityTask] FOREIGN KEY ([TaskId]) REFERENCES [dbo].[SecurityTask] ([Id]) ON DELETE CASCADE,
```
withinfocus, line 17: ⚠️ I thought this would cause an error due to the trailing comma -- is that not the case?

nick-livefront: It did not cause issues for me locally but that's not going to say it won't elsewhere. Removed [4a1eb4b](https://github.com/bitwarden/server/pull/5690/commits/4a1eb4b3b02f41147e77107fc6b199c7a4ce3ba6)

util/Migrator/DbScripts/2025-03-21_00_NotificationCascadeDelete.sql
rkac-bw: Sort of the same, both are good. This version that uses sys.foreign_keys instead of information_schema.table_constraints, is more specific to SQL Server is the ideal place to check for constraint information, and the optimization of top 1 minimizes bad plan checks

rkac-bw (approved):
LGTM

nick-livefront:
@shane-melton @rkac-bw There was a relatively similar issue regarding the `PasswordHealthReportApplication` that is called out in the same ticket. That I added a fix for here. [a8c10a9](https://github.com/bitwarden/server/pull/5690/commits/a8c10a902f91027975a3ff5c635e771ef8df13f0)

I noticed that the [DatabaseContextModelSnapshot](https://github.com/bitwarden/server/blob/4a1eb4b3b02f41147e77107fc6b199c7a4ce3ba6/util/PostgresMigrations/Migrations/DatabaseContextModelSnapshot.cs#L2823-L2827) already contained the cascade for EF which is why there are not migrations in that commit.

shane-melton (approved):

rkac-bw (approved):
lgtm

PR #5694, opened 2025-04-22
[PM-17777] sponsorships consume seats

cturnbull-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17777

## 📔 Objective

This PR aims to introduce the behavior where sponsorships sent by an organization admin on behalf of another user will consume a seat. All of the expected seat count and autoscaling/not-autoscaling behavior and validation also applies now to these admin initiated sponsorships.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/Billing/Controllers/OrganizationSponsorshipsController.cs
```diff
@@ -86,9 +86,9 @@ public async Task CreateSponsorship(Guid sponsoringOrgId, [FromBody] Organizatio
 
         if (!_featureService.IsEnabled(Bit.Core.FeatureFlagKeys.PM17772_AdminInitiatedSponsorships))
         {
-            if (model.SponsoringUserId.HasValue)
+            if (model.IsAdminInitiated.GetValueOrDefault())
             {
-                throw new NotFoundException();
+                throw new BadRequestException();
```
cturnbull-bitwarden, line 91: :memo: This is a tweak on the feature flag "off" behavior. If an admin-initiated sponsorship request comes through, I changed this to bad request instead of not found. This endpoint existed prior to this feature being implemented, so it's more of a bad request in that admins shouldn't be sending sponsorships when the flag is off.

src/Api/Models/Request/Organizations/OrganizationSponsorshipCreateRequestModel.cs
```diff
@@ -17,11 +17,7 @@ public class OrganizationSponsorshipCreateRequestModel
     [StringLength(256)]
     public string FriendlyName { get; set; }
 
-    /// <summary>
-    /// (optional) The user to target for the sponsorship.
-    /// </summary>
-    /// <remarks>Left empty when creating a sponsorship for the authenticated user.</remarks>
-    public Guid? SponsoringUserId { get; set; }
+    public bool? IsAdminInitiated { get; set; }
```
cturnbull-bitwarden, line 20: :memo: I updated the request model to not require a userId to be passed in since `_currentContext.UserId` is always the correct sponsoring user. Instead, I replaced it with this `IsAdminInitiated` property to tell the system which kind of sponsorship this is

src/Core/OrganizationFeatures/OrganizationSponsorships/FamiliesForEnterprise/CreateSponsorshipCommand.cs
```diff
@@ -14,11 +14,17 @@ namespace Bit.Core.OrganizationFeatures.OrganizationSponsorships.FamiliesForEnte
...
-    public async Task<OrganizationSponsorship> CreateSponsorshipAsync(Organization sponsoringOrganization,
-        OrganizationUser sponsoringMember, PlanSponsorshipType sponsorshipType, string sponsoredEmail,
-        string friendlyName, string notes)
+    public async Task<OrganizationSponsorship> CreateSponsorshipAsync(
+        Organization sponsoringOrganization,
+        OrganizationUser sponsoringMember,
+        PlanSponsorshipType sponsorshipType,
+        string sponsoredEmail,
+        string friendlyName,
+        bool isAdminInitiated,
```
cturnbull-bitwarden, line 26: :memo: Whitespace formatting - sorry about that. New param is `bool isAdminInitiated` which is passed down from the request model

src/Core/OrganizationFeatures/OrganizationSponsorships/FamiliesForEnterprise/CreateSponsorshipCommand.cs
```diff
@@ -48,48 +54,44 @@ public async Task<OrganizationSponsorship> CreateSponsorshipAsync(Organization s
             throw new BadRequestException("Can only sponsor one organization per Organization User.");
         }
 
-        var sponsorship = new OrganizationSponsorship();
-        sponsorship.SponsoringOrganizationId = sponsoringOrganization.Id;
-        sponsorship.SponsoringOrganizationUserId = sponsoringMember.Id;
-        sponsorship.FriendlyName = friendlyName;
-        sponsorship.OfferedToEmail = sponsoredEmail;
-        sponsorship.PlanSponsorshipType = sponsorshipType;
+        if (isAdminInitiated)
```
cturnbull-bitwarden, line 57: :memo: I moved the validation logic to a private method for this feature, and then there's a Rider refactor to use object initialization.

src/Core/OrganizationFeatures/OrganizationSponsorships/FamiliesForEnterprise/CreateSponsorshipCommand.cs
```diff
@@ -48,48 +54,44 @@ public async Task<OrganizationSponsorship> CreateSponsorshipAsync(Organization s
...
 
         if (existingOrgSponsorship != null)
         {
             // Replace existing invalid offer with our new sponsorship offer
             sponsorship.Id = existingOrgSponsorship.Id;
         }
 
-        var isAdminInitiated = false;
-        if (currentContext.UserId != sponsoringMember.UserId)
+        if (isAdminInitiated && sponsoringOrganization.Seats.HasValue)
```
cturnbull-bitwarden, line 79: :memo: This is the first actual logic change to this command. If this sponsorship is admin initiated, then we'll auto add seats to accommodate this sponsorship.

src/Core/OrganizationFeatures/OrganizationSponsorships/FamiliesForEnterprise/CreateSponsorshipCommand.cs
```diff
@@ -48,48 +54,44 @@ public async Task<OrganizationSponsorship> CreateSponsorshipAsync(Organization s
...
         {
-            var organization = currentContext.Organizations.First(x => x.Id == sponsoringOrganization.Id);
-            OrganizationUserType[] allowedUserTypes =
-            [
-                OrganizationUserType.Admin,
-                OrganizationUserType.Owner
-            ];
-
-            if (!organization.Permissions.ManageUsers && allowedUserTypes.All(x => x != organization.Type))
+            if (isAdminInitiated)
```
cturnbull-bitwarden, line 86: :memo: Since org admins can send more than one sponsorship, we'll create a new one every time `isAdminInitiated` is `true` instead of replacing the previous one

src/Infrastructure.EntityFramework/AdminConsole/Repositories/Queries/OrganizationUserReadOccupiedSeatCountByOrganizationIdQuery.cs
```diff
@@ -14,9 +14,23 @@ public OrganizationUserReadOccupiedSeatCountByOrganizationIdQuery(Guid organizat
...
-                    where ou.OrganizationId == _organizationId && ou.Status >= OrganizationUserStatusType.Invited
-                    select ou;
-        return query;
+        var orgUsersQuery = from ou in dbContext.OrganizationUsers
+                            where ou.OrganizationId == _organizationId && ou.Status >= OrganizationUserStatusType.Invited
+                            select ou;
+
+        // As of https://bitwarden.atlassian.net/browse/PM-17772, a seat is also occupied by a Families for Enterprise sponsorship sent by an
+        // organization admin, even if the user sent the invitation doesn't have a corresponding OrganizationUser in the Enterprise organization.
+        var sponsorshipsQuery = from os in dbContext.OrganizationSponsorships
```
cturnbull-bitwarden, line 23: :memo: This file and the next are core to this story's requirements. Since admin-initiated sponsorships now consume a seat, we need to reflect that in the "consumed seat count" logic returned here. This EF file and the stored proc changes in the next files aim to accomplish that.

I'd appreciate your input if you think there's a better way to do this! While I explained my thoughts, don't take them as gospel. You may have a better approach in mind

cturnbull-bitwarden (commented):
Left some :memo: thoughts explaining some of what I did here

amorask-bitwarden (approved):
Looks great - awesome work and thank you for the detailed documentation in the changes. That was very helpful. You've got some failing tests so just reach out when you need another approval after resolution.

BTreston (commented):
The code looks good, I just have concerns about making this change without a feature flag. It would be definitely be the safer route. Also not sure about directly editing the sprocs vs creating vnext implementations that live behind the feature flag (same reasoning here). Since its read operation I am wiling to let it ride, but I'm gonna rope in a server dev for second thoughts.

r-tome:
> The code looks good, I just have concerns about making this change without a feature flag. It would be definitely be the safer route. Also not sure about directly editing the sprocs vs creating vnext implementations that live behind the feature flag (same reasoning here). Since its read operation I am wiling to let it ride, but I'm gonna rope in a server dev for second thoughts.

I agree with your reasoning @BTreston. I'm also not a huge advocate of having a `v2 `version of a stored procedure but in this case it really changes how we calculate the seat count.

cturnbull-bitwarden:
@BTreston and @r-tome , good thoughts to bring up! I ended up bringing this up with Micah, and here's what we ended up with.

Right now, when we turn off the feature flag for this one, it'll prevent org admins from being able to call this endpoint. If we turn on the LD flag in prod, org admins then go in and send sponsorships, then if we should go back into LD and turn the flag off for some reason, the behavior should remain where these sponsorships continue to count towards the seat count since those sponsorships still exist.

There's already existing logic from a prior PR that gates this feature behind the feature flag at the controller level. If the flag is off, then org admins won't be able to send sponsorships on behalf of someone else since we'll be returning a `400` to them.

amorask-bitwarden (approved):

BTreston (approved):

r-tome (approved):

rkac-bw (approved):
lgtm

PR #5748, opened 2025-04-28
[PM-20416]Downgraded paid org cannot upgrade from Free

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20416
## 📔 Objective

Replication steps

Create a paid org with a subscription

cancel the subscription and downgrade the organization to a Free org (do not remove the customer or subscription info)

navigate to billing > Subscription

select “upgrade plan”

Expected: the upgrade modal allows the free org to select between paid org types

Actual: the upgrade modal only allows the org to select free org type

Open image-20250418-140903.png
image-20250418-140903.png

## 📸 Screenshots

https://github.com/user-attachments/assets/6985ecf6-e807-4aed-86ce-36e83b5a2b97

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Admin/AdminConsole/Controllers/OrganizationsController.cs
+++ b/src/Admin/AdminConsole/Controllers/OrganizationsController.cs
@@ -255,6 +255,12 @@ public class OrganizationsController : Controller
             Seats = organization.Seats
         };
 
+        if (model.PlanType == PlanType.Free && model.Seats > 2)
+        {
+            TempData["Error"] = "Organizations with more than 2 seats cannot be downgraded to the Free plan";
+            return RedirectToAction("Edit", new { id });
+        }
+
         UpdateOrganization(organization, model);
 
         var plan = await _pricingClient.GetPlanOrThrow(organization.PlanType);
@@ -483,6 +489,11 @@ public class OrganizationsController : Controller
             organization.Gateway = model.Gateway;
             organization.GatewayCustomerId = model.GatewayCustomerId;
             organization.GatewaySubscriptionId = model.GatewaySubscriptionId;
+        }
+
+        if (model.PlanType == PlanType.Free && organization.GatewaySubscriptionId is not null)
+        {
+            organization.GatewaySubscriptionId = null;
         }
     }
 


amorask-bitwarden (dismissed):

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -484,6 +490,11 @@ private void UpdateOrganization(Organization organization, OrganizationEditModel
             organization.GatewayCustomerId = model.GatewayCustomerId;
             organization.GatewaySubscriptionId = model.GatewaySubscriptionId;
         }
+
+        if (model.PlanType == PlanType.Free && organization.GatewaySubscriptionId is not null)
+        {
+            organization.GatewaySubscriptionId = null;
+        }
```
eliykat, line 497: Are we sure we want to do this? Clay's comment in Slack was:

> I would rather the code be resilient enough for the canceled subscriptionId to remain present just like it is for Premium memberships, which will also mitigate the need for us to go back and manually remove this from all of the existing downgraded Free accounts.

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -255,6 +255,12 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
             Seats = organization.Seats
         };
 
+        if (model.PlanType == PlanType.Free && model.Seats > 2)
+        {
+            TempData["Error"] = "Organizations with more than 2 seats cannot be downgraded to the Free plan";
+            return RedirectToAction("Edit", new { id });
+        }
```
eliykat, line 262: We've hit issues here before where existing orgs cannot be edited (for an unrelated reason) because it trips new validation code. For example, if we have any orgs today on the Free plan with more than 2 seats, CS will no longer be able to edit anything about them because they'll hit this error. (Obviously we _shouldn't_ have that but I can't guarantee it.)

I recommend only enforcing this if the plan has changed, e.g. `if (organization.PlanType != PlanType.Free && model.PlanType == PlanType.Free && model.Seats > 2)`.

eliykat (changes requested):

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
cyprain-okeke: The customerId is retain which contains details of all existing subscription in stripe. Retaining This will make the subscription details to have two subscription on the admin console, when the account is cancelled and downgraded to free. (The free and the cancelled plan)

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -255,6 +255,15 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
         };
 
+        if (organization.PlanType != PlanType.Free && model.PlanType == PlanType.Free && model.Seats > 2)
+        {
+            model.Seats = 2;
+            model.MaxCollections = 2;
+            model.MaxStorageGb = null;
+            model.ExpirationDate = null;
+            model.Enabled = true;
+        }
```
eliykat, line 265: I'd rather avoid silently overwriting data after form submission - this is likely to be unexpected behaviour to the our internal teams. If we want to do this, we should do it in the UI so that the user sees the change before submitting the form.

This is also going to reduce the number of seats, but without reducing the number of organization members. So the organization will still have (for example) 10 members but they'll only be charged for 2, which is not what we want.

I think it's OK to throw an error here now that the `if` logic has been updated. That is:

```c#
        if (organization.PlanType != PlanType.Free && model.PlanType == PlanType.Free && model.Seats > 2)
        {
            TempData["Error"] = "Organizations with more than 2 seats cannot be downgraded to the Free plan";
            return RedirectToAction("Edit", new { id });
        }
```

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
eliykat: No problem, I'm less familiar with how these Ids are used, so I'll leave it with you to discuss with Clay 👍

eliykat (changes requested):

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
cyprain-okeke: revert

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
cyprain-okeke: I have discussed with Clay and He insist of retaining the old subscription Id. I have remove the code

eliykat (approved):
AC changes are approved, thanks @cyprain-okeke

amorask-bitwarden (dismissed):

amorask-bitwarden (approved):

eliykat (changes requested):
A couple of small things I just noticed. Hopefully shouldn't need QA to be re-done, but let me know if it causes any problems.

amorask-bitwarden (dismissed):

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -255,29 +255,35 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
-            model.MaxStorageGb = null;
-            model.ExpirationDate = null;
-            model.Enabled = true;
+            if (organization.PlanType != PlanType.Free && model.PlanType.Value == PlanType.Free && model.MaxCollections > freePlan.PasswordManager.MaxCollections)
+            {
+                TempData["Error"] = $"Organizations with more than {freePlan.PasswordManager.MaxCollections} collections cannot be downgraded to the Free plan. Your organization currently has {organization.MaxCollections} collections.";
+                return RedirectToAction("Edit", new { id });
+            }
+
+            if (organization.PlanType != PlanType.Free && model.PlanType.Value == PlanType.Free)
```
eliykat, line 277: `organization.PlanType != PlanType.Free && model.PlanType.Value == PlanType.Free` is repeated in each nested `if` statement. I suggest it's moved up to the outer `if` statement, e.g.

```c#
var isDowngradingToFree = model.PlanType.HasValue && organization.PlanType != PlanType.Free && model.PlanType.Value == PlanType.Free;
if (isDowngradingToFree)
{
  // ... your code here
}
```

eliykat (changes requested):

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
cyprain-okeke: resolved

eliykat (approved):
Thanks for working through the changes!

In the future we probably want to avoid adding more to this controller, if we want to start validating plan changes here it would be good to reuse some of the validation we have on self-serve changes.

amorask-bitwarden (approved):

eliykat (dismissed):

amorask-bitwarden (approved):

eliykat (approved):

PR #5754, opened 2025-04-29
[PM-19029][PM-19203] Addressing `UserService` tech debt around `ITwoFactorIsEnabledQuery`

ike-kottlowski:
## 🎟️ Tracking

[PM-19029](https://bitwarden.atlassian.net/browse/PM-19029)
[PM-19203](https://bitwarden.atlassian.net/browse/PM-19073)

## 📔 Objective

Objective is to fix the `ITwofactorIsEnabledQuery` to properly show whether a user has two factor authentication enabled in the admin portal.

The error had to do with the query prioritizing premium two factor providers to show whether or not a user had access to two factor authentication. Admin portal now accurately represents a user's two factor authentication configuration in combination with their premium status.

I opted to use this bug as an opportunity to address tech debt around the user service by replacing the `TwoFactorIsEnabledAsync()` in the `IUserService` with the `ITwoFactorIsEnabledQuery` implementation.

Modifications were also made to the two factor token providers to reduce complexity and reliance on the IUserService. The user object has a sense of its two factor already, thus not requiring the IUserService. We are able to just ask the user object directly.

## 📸 Screenshots

https://github.com/user-attachments/assets/07cb1c8b-d9c2-4030-a2ac-e3128fcdf1bd

User tries to log in with a premium two factor; User is prompted for YubiKey; In Admin Portal Admin turns off premium, which disables two factor auth for the user; User tries to login and is required to verify the device, This means the two factor authentication is turned off; User verifies device; User navigates to two step login and there are no two step logins are available due to lack of subscription; YubiKey is still configured, but not manageable; In Admin Portal Two Factor Authentication is still disabled, which is correct; User enables non-premium Email two factor; In Admin portal two factor is now enabled; Two factor enabled is `true` even with premium off because email is not a premium two factor; Admin enables premium; User is again able to modify their YubiKey configuration;

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-19029]: https://bitwarden.atlassian.net/browse/PM-19029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-19203]: https://bitwarden.atlassian.net/browse/PM-19203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
eliykat, line 1: One suggestion, then I am vanishing in a puff of smoke from this PR review: today, there are 2 main public methods: one for a single UserId, and one for multiple UserIds. Each has a separate implementation of the logic which has to be maintained and tested. While you're here tackling tech debt, it would be nice to just have the single method call the bulk method.

e.g.

```C#
var results = await TwoFactorIsEnabledAsync([userId]);
return results.Single();
```

I mention it because AC Team wrote the intial implementation and I think it was a mistake to repeat so much logic; it's a pattern I'm trying to remove in our own code.

ike-kottlowski: An excellent suggestion.

I did try to combine the logic but the issue I ran into was timing the database call vs early return.

In the single user implementation, we already have access to the some of the data locally. So we can have early returns for states we know to be `false`.

In the bulk user implementation, the only early return logic we can try is whether the `Guid[]` is valid. Then we immediately make a database look up.

I decided for the vast majority of calls being made to the `ITwoFactorIsEnabledQuery` that it would be for single users, and a sub-set of those users won't have MFA enabled. We would be able to detect users without MFA much more quickly with the early returns vs using the `bulk.single()`.

That being said, I wonder if we could invert the idea. Use the single case to store the logic and the bulk case to call the single case.

eliykat: Good points.

I was going to suggest this: change the bulk interface to take `IEnumerable<ITwoFactorProvidersUser>` rather than `IEnumerable<Guid>`. (There is already a method for that, but it throws away the object and just calls the bulk method with the Guid.) The main logic would live in this method, and you'd force the caller to fetch and pass in the whole objects, like we do with the single method. That allows you to shortcut based on the information on the `ITwoFactorProvidersUser` object (whether single or bulk), and as a bonus it makes the interface more consistent.

However, that's just pushing the db call out to the caller, so maybe it's not really shortcutting anything. It makes me wonder if `ITwoFactorProvidersUser` should include the "calculated premium" property, given that this is required to properly evaluate their 2FA providers. Why even pass in an `ITwoFactorProvidersUser` when it's not the model you need? I'm not sure what ripple effect that would have on the code though.

> I wonder if we could invert the idea. Use the single case to store the logic and the bulk case to call the single case.

I think the challenge there is that you still need to move your database calls upfront so that you can then loop over your user/s without calling the db each time.

ike-kottlowski: I utilized a delegate in a private method to enable the calling method to decide how premium is fetched.

```C#
private async static Task<bool> TwoFactorEnabledAsync(
        Dictionary<TwoFactorProviderType, TwoFactorProvider> providers,
        Func<Task<bool>> hasPremiumAccessDelegate)
```

In the single user case we call the database, in the bulk case we reference the object that was already fetched from the database.

> It makes me wonder if ITwoFactorProvidersUser should include the "calculated premium" property, given that this is required to properly evaluate their 2FA providers. 

I have an item on my backlog to chat with Billing/Admin Console (not sure who owns premium _and_ organization premium) about if we can modify the `GetUsers` stored procedure to have the `[dbo].[user].[premium]` to always be the calculated premium.

eliykat: I think the delegate is a nice solution. That backlog item probably falls to AC Team because it's primarily about querying organization users, so I'll shoot you a message about it. Thanks the discussion here!

src/Core/Entities/User.cs
```diff
@@ -137,19 +141,17 @@ public string SubscriberType()
...
-            if (_twoFactorProviders == null)
-            {
-                _twoFactorProviders =
-                    JsonHelpers.LegacyDeserialize<Dictionary<TwoFactorProviderType, TwoFactorProvider>>(
-                        TwoFactorProviders);
-            }
-
-            // U2F is no longer supported, and all users keys should have been migrated to WebAuthn.
-            // To prevent issues with accounts being prompted for unsupported U2F we remove them
-            if (_twoFactorProviders.ContainsKey(TwoFactorProviderType.U2f))
```
JaredSnider-Bitwarden, line 149: ❓ : why remove this check?

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
JaredSnider-Bitwarden, line 1: ❓ : could we please target a bit higher test coverage on this query?

JaredSnider-Bitwarden (changes requested):
Nice refactor! Just 2 questions.

src/Core/Entities/User.cs
ike-kottlowski: The check is combined into ln. 154. in the `Remove()` call.

`Remove()` doesn't complain when something doesn't exist. It accomplishes the same goal with only a single lookup rather than: checking for existence (`ContainsKey()`) and then checking again to remove it. 

[Docs for Remove method](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.remove?view=net-8.0)

I think this makes the implementation clearer, along with the `[Obsolete]` decorator added to the `TwoFactorProviderType.U2f`

JaredSnider-Bitwarden: Ah, ty for the extra context. That makes sense!

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
ike-kottlowski: Coverage added in [1528334](https://github.com/bitwarden/server/pull/5754/commits/1528334f680653a168c055cb37cda10336da93ca)

JaredSnider-Bitwarden (approved):
Ty!

cturnbull-bitwarden (approved):

gbubemismith (approved):
Changes look good

BTreston (approved):

PR #5756, opened 2025-04-30
[PM-20092] Refactor OrganizationUsersController Get to return account recovery users

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-20092

## 📔 Objective
This PR refactors the existing OrganizationUsersController Get endpoint and adds support for returning users enrolled in account recovery if the accessing user has `ManageResetPassword` custom permissions.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserUserDetailsQuery.cs
```diff
@@ -37,4 +50,45 @@ public async Task<IEnumerable<OrganizationUserUserDetails>> GetOrganizationUserU
...
+    /// Get the organization user user details, two factor enabled status, and
+    /// claimed status for the provided request.
+    /// </summary>
+    /// <param name="request">Request details for the query</param>
+    /// <returns>List of OrganizationUserUserDetails</returns>
+    public async Task<IEnumerable<(OrganizationUserUserDetails, bool, bool)>> Get(OrganizationUserUserDetailsQueryRequest request)
+    {
+        var organizationUsers = await GetOrganizationUserUserDetails(request);
+
+        var organizationUsersTwoFactorEnabled = await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(organizationUsers);
```
eliykat, line 64: Please transform this result into a HashSet or similar so that we don't have nested loops below. (i.e. looping over this result inside the loop for each user.)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserUserDetailsQuery.cs
```diff
@@ -37,4 +50,45 @@ public async Task<IEnumerable<OrganizationUserUserDetails>> GetOrganizationUserU
...
+    /// <param name="request">Request details for the query</param>
+    /// <returns>List of OrganizationUserUserDetails</returns>
+    public async Task<IEnumerable<(OrganizationUserUserDetails, bool, bool)>> GetConfirmed(OrganizationUserUserDetailsQueryRequest request)
+    {
+
+        var result = await Get(request);
+        var responses = result.Where(r => r.Item1.Status.Equals(OrganizationUserStatusType.Confirmed));
+
+        return responses;
+    }
```
eliykat, line 92: I like how this is just a wrapper around another public method, that is generally the approach I prefer. However, it means we're getting a lot of data (2FA and claimed status) for users that we don't need. So in this case I do actually recommend some light code duplication so that we get the orgUsers, filter for the ones we want, and _then_ get the 2FA and confirmed status just for those users.

You could use a private method, but the methods in this class are so short and the heavy lifting is already encapsulated in the other queries, so I think a copy/paste & tweak of `Get` is fine.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IOrganizationUserUserDetailsQuery.cs
```diff
@@ -6,4 +6,8 @@ namespace Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
 public interface IOrganizationUserUserDetailsQuery
 {
     Task<IEnumerable<OrganizationUserUserDetails>> GetOrganizationUserUserDetails(OrganizationUserUserDetailsQueryRequest request);
+
+    Task<IEnumerable<(OrganizationUserUserDetails, bool, bool)>> Get(OrganizationUserUserDetailsQueryRequest request);
+
+    Task<IEnumerable<(OrganizationUserUserDetails, bool, bool)>> GetConfirmed(OrganizationUserUserDetailsQueryRequest request);
```
eliykat, line 12: Given that we have no other uses for this right now, I'm OK with this being specific - I suggest `GetAccountRecoveryEnrolledUsers`. This also allows you to do more filtering earlier, and reduces the logic in the controller.

(Account Recovery is the updated name for this feature)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Authorization/OrganizationUserUserDetailsAuthorizationHandler.cs
```diff
@@ -49,4 +53,18 @@ private async Task<bool> CanReadAllAsync(Guid organizationId)
         // Allow provider users to read all organization users if they are a provider for the target organization
         return await _currentContext.ProviderUserForOrgAsync(organizationId);
     }
+
+    private bool ReadWithAccountRecovery(Guid organizationId)
+    {
+        var organization = _currentContext.GetOrganization(organizationId);
+        if (organization is
+        { Permissions.ManageUsers: false } and
+        { Permissions.ManageResetPassword: true }
```
eliykat, line 62: Not strictly true I think. If you have `ManageUsers && ManageResetPassword`, you have _permission_ to access this data, even if in practice you're going to receive the full data set. But that's for the controller to manage - the authorization code itself shouldn't have any knowledge of that.

This should allow:
- owners
- admins
- ManageResetPassword custom permission
- providers

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Authorization/OrganizationUserUserDetailsAuthorizationHandler.cs
eliykat, line 1: We have some new, more reusable authorization requirements that you can use here.

See the [ManageUsersRequirement](https://github.com/bitwarden/server/blob/92701d8cd0dd096d7abffbf38d591b930285a14b/src/Api/AdminConsole/Authorization/Requirements/ManageUsersRequirement.cs#L8-L9). It implements `IOrganizationRequirement`, which defines what roles meets the requirement. Its handler, the `OrganizationRequirementHandler`, will automatically get the orgId from the route params and executes the callback on the requirement.

This means that your authorize calls can look like this:

```cs
var manageUsers = (await _authorizationService.AuthorizeAsync(User, new ManageUsersRequirement())).Succeeded;
```

Without defining your own operation or handler, and without having to pass in the `OrganizationScope` object.

We don't have a `ManageResetPasswordRequirement` yet but you could easily create one by using the `ManageUsersRequirement` as a guide.

eliykat (changes requested):
I think this is looking really good 👏

src/Api/AdminConsole/Authorization/Requirements/ManageAccountRecoveryRequirement.cs
```diff
@@ -0,0 +1,21 @@
...
+public class ManageAccountRecoveryRequirement : IOrganizationRequirement
+{
+    public async Task<bool> AuthorizeAsync(
+        CurrentContextOrganization? organizationClaims,
+        Func<Task<bool>> isProviderUserForOrg)
+        => organizationClaims switch
+        {
+            { Type: OrganizationUserType.Owner } => true,
+            { Type: OrganizationUserType.Admin } => true,
+            { Permissions.ManageUsers: true } => true,
```
eliykat, line 17: After this change, these are fully separate permissions. So `ManageUsers` cannot manage account recovery.

```suggestion
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserUserDetailsQuery.cs
```diff
@@ -82,11 +82,22 @@ public async Task<IEnumerable<OrganizationUserUserDetails>> GetOrganizationUserU
...
     /// <returns>List of OrganizationUserUserDetails</returns>
-    public async Task<IEnumerable<(OrganizationUserUserDetails, bool, bool)>> GetConfirmed(OrganizationUserUserDetailsQueryRequest request)
+    public async Task<IEnumerable<(OrganizationUserUserDetails OrgUser, bool TwoFactorEnabled, bool ClaimedByOrganization)>> GetAccountRecoveryEnrolledUsers(OrganizationUserUserDetailsQueryRequest request)
     {
+        var organizationUsers = await GetOrganizationUserUserDetails(request);
+
+        var organizationUsersTwoFactorEnabled = (await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(organizationUsers)).ToDictionary(u => u.user.Id);
+        var organizationUsersClaimedStatus = await _getOrganizationUsersClaimedStatusQuery.GetUsersOrganizationClaimedStatusAsync(request.OrganizationId, organizationUsers.Select(o => o.Id));
+        var responses = organizationUsers
+            .Where(o => o.Status.Equals(OrganizationUserStatusType.Confirmed) && o.UsesKeyConnector == false && !String.IsNullOrEmpty(o.ResetPasswordKey))
```
eliykat, line 85: This is still getting all data for all users, then filtering as the last step. My recommendation here is to do it earlier:

```c#
var organizationUsers = await GetOrganizationUserUserDetails(request);

// Filter here
var enrolledOrganizationUsers = organizationUsers
            .Where(o => o.Status.Equals(OrganizationUserStatusType.Confirmed) && o.UsesKeyConnector == false && !String.IsNullOrEmpty(o.ResetPasswordKey));

// Then we only get additional data for the users we care about
var organizationUsersTwoFactorEnabled = (await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(enrolledOrganizationUsers)).ToDictionary(u => u.user.Id);
var organizationUsersClaimedStatus = await _getOrganizationUsersClaimedStatusQuery.GetUsersOrganizationClaimedStatusAsync(request.OrganizationId, enrolledOrganizationUsers.Select(o => o.Id));
```

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -209,32 +210,30 @@ private async Task<ListResponseModel<OrganizationUserUserDetailsResponseModel>>
...
             User, new OrganizationScope(orgId), OrganizationUserUserDetailsOperations.ReadAll)).Succeeded;
 
-        if (!readAllAuthorized)
+        if (readAllAuthorized)
         {
-            var readWithAccountRecovery = (await _authorizationService.AuthorizeAsync(
-                User, new OrganizationScope(orgId), OrganizationUserUserDetailsOperations.ReadAccountRecovery)).Succeeded;
+            var results = await _organizationUserUserDetailsQuery.Get(request);
+            var responseList = results
+            .Select(result => new OrganizationUserUserDetailsResponseModel(result.OrgUser, result.TwoFactorEnabled, result.ClaimedByOrganization))
```
eliykat: I'd rather avoid having multiple bools as positional arguments. Please add a new ctor for this model that takes the tuple directly.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -209,32 +210,30 @@ private async Task<ListResponseModel<OrganizationUserUserDetailsResponseModel>>
...
+            .ToList();
 
-            if (!readWithAccountRecovery)
-            {
-                throw new NotFoundException();
-            }
+            return (new ListResponseModel<OrganizationUserUserDetailsResponseModel>(responseList));
```
eliykat: Unresolved earlier comment: https://github.com/bitwarden/server/pull/5756#discussion_r2069790516
(unless dotnet format is doing this in which case it's fine)

eliykat (changes requested):

eliykat (approved):
Nice work.

r-tome (approved):

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -162,6 +163,12 @@ public async Task<ListResponseModel<OrganizationUserUserMiniDetailsResponseModel
     [HttpGet("")]
     public async Task<ListResponseModel<OrganizationUserUserDetailsResponseModel>> Get(Guid orgId, bool includeGroups = false, bool includeCollections = false)
     {
+
+        if (!_featureService.IsEnabled(FeatureFlagKeys.AccountDeprovisioning))
```
r-tome, line 167: I think this is wrong
```suggestion
        if (_featureService.IsEnabled(FeatureFlagKeys.SeparateCustomRolePermissions))
```

r-tome (changes requested):

eliykat (approved):
Good spot @r-tome 😁

r-tome (approved):

PR #5757, opened 2025-05-01
[PM-18555] Main part of notifications refactor

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18555

## 📔 Objective

I promise this is me trying to keep the changeset small 😭.

2e945df6700fa3a88a7eff8e7b2eb65a92dabf4c, 1fade79fa3e60ee7320e51ba71e2ca768da52c0a, and cbc929829c06d578e129e371cf06f1204cf6bc49 add various tests to help keep similar behavior as to before any changes get added. Some of these tests disappear in later commits because I am proposing changing some behavior.

dcd1fbbd59dcf789e59cc23fd90b1af89f8926f5 adds the new proposed services and associated types but does not implement them yet. 

25b495b2adb0dc172702f115eae5b29e83a739f5 and bc5b44cdfadae5ef1f26af78fd06afcbdcb64ac8 start implementing the services and deletes a lot of the now unnecessary code from the various `IPushNotificationService` implementation. This changes `IPushNotificationService` to using default implementations for all methods except `PushAsync` and `PushCipherAsync`. The ideal state for this is that these are instead extension methods that can live in various other team owned code. Switching to extension methods changes the mechanics of mocking and expecting this code though and that causes a fair amount of down stream changes. Doing this switch will come later on a team by team basis. These commits also most fix the existing tests with some accepted differences that I will point out in comments in the code. 

fe08961744d01e089aee0197299dcf46cb170c5b further fixes some tests that were failing after the unification. 

ee4a95c3950bf6cc70f350b00b09af186704fb2f is a commit fixing formatting.

edb419f89c24b135fd442a3e61515fae3e847633 switches the `PushSendRequestModel` to being nullable guids instead of a nullable string. The only values we ourselves would put into these models would be `Guid`'s and as far as I can tell passing in values of non-guids WOULD allow the notification into our infrastructure but these notifications would be not be delivered to anyone because they wouldn't correspond to real devices/orgs/users/installations to Bitwarden. I spoke with @dani-garcia to check if we expect this would break Vaultwarden users who are using our cloud to relay notifications and as far as we can tell they are generating valid v4 UUIDs that are sent via a string in JSON and that is the expected way these GUIDs would make it to us. 

0bb93204e0e60e149b979b23fb60cc518315ad7b Removes TODO comments of things I did in the previous commit.

The things left over from this PR are moving to extension methods and updating tests to not break because of that, adding some documentation, making a pure `AddPushServices` extension for `IServiceCollection`, and cleaning up [this method](https://github.com/bitwarden/server/blob/e77acbc5ad419b3641eab6ba381f4b3d55105e10/src/Notifications/HubHelpers.cs#L15) in a way that doesn't require addition every time new notifications are added. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


test/Core.Test/Platform/Push/Services/NotificationsApiPushNotificationServiceTests.cs
```diff
@@ -221,7 +220,7 @@ protected override JsonNode GetPushSendCreatePayload(Send send)
                 ["UserId"] = send.UserId,
                 ["RevisionDate"] = send.RevisionDate,
             },
-            ["ContextId"] = null,
+            ["ContextId"] = DeviceIdentifier,
```
justindbaur, line 223: I spoke with @addisonbeck about these changes from previously NOT sending the DeviceIdentifier as the `ContextId` but now doing it. `NotificationsApiPushNotificationService` is the only service that wasn't sending this but the other 3 implementations were. 

`RelayPushNotificationsService` sending `true`
https://github.com/bitwarden/server/blob/e77acbc5ad419b3641eab6ba381f4b3d55105e10/src/Core/Platform/Push/Services/RelayPushNotificationService.cs#L173-L186

`AzureQueuePushNotificationService` sending `true`
https://github.com/bitwarden/server/blob/e77acbc5ad419b3641eab6ba381f4b3d55105e10/src/Core/Platform/Push/Services/AzureQueuePushNotificationService.cs#L231-L244

`NotificationHubPushNotificationService` sending `true`
https://github.com/bitwarden/server/blob/e77acbc5ad419b3641eab6ba381f4b3d55105e10/src/Core/NotificationHub/NotificationHubPushNotificationService.cs#L177-L190

BUT `NotificationsApiPushNotificationService` sending in `false`
https://github.com/bitwarden/server/blob/e77acbc5ad419b3641eab6ba381f4b3d55105e10/src/Core/Platform/Push/Services/NotificationsApiPushNotificationService.cs#L245-L258

We both don't think this was on purpose so it is being fixed in this PR and these test changes are the proof of that.

tangowithfoxtrot (approved):
Thanks for going over this with me! LGTM

addisonbeck (approved):
Approving for merge, as I see this has already been tested and reviewed. I skimmed the changes but didn't spent much time trying to find anything to critique.

PR #5795, opened 2025-05-09
[PM-17562] Add support for retires on event integrations

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This PR adds a new layer of AMQP messaging (for RabbitMQ now, but ASB will be added in another PR). This allows us to queue each individual integration message to allow for retries and dead-lettering.

The flow of events with this new change (assuming the feature flag is On and Rabbit is configured):
1. Event comes into the system
2. Event gets published to the main exchange, which is a fanout exchange. This means it is immediately published to the event-integration queues (1 per integration, currently webhook and slack), and the event repository queue
3. Listener/Handler pairs on each queue receive the message and run
   - For the event repository handler, it writes the event to the `EventRepository`.
4. For the event-integration queues, the EventIntegrationHandler pulls the current configurations for this Organization, Event Type, and `IntegrationType`. For each configuration: 
   - It parses the configuration into a typed configuration object
   - It processes the Template string and fills in tokens with the contents of EventMessage or any other fetched properties
   - It bundles all of this up into an `IntegrationMessage`
   - And it publishes that message to another exchange - the Integration Exchange - with the routing key of the `IntegrationType`.
5. The Integration Exchange is not fanout. It publishes direct to exactly one queue depending on the routing key. The queues are one per integration (i.e. slack, webhook).
6. The specific queue receives the message and another Listener/Handler pair runs. These are directly tied to the integration and take the IntegrationMessage (which has the necessary configuration and message) and publish it to the integration.
7. The handlers return a result that indicates a success/failure and (if failure) if the message can be retried. The listener checks the result. If it's a success, it's simply acknowledged and the process ends. If it's retryable, it is published to the retry queue for this integration. If it is not retryable, or it has exceeded the configurable number of retries, it is sent to the dead letter queue.

In the RabbitMQ implementation, we are using a very simple retry queue. It has a TTL of 30 seconds with a dead-letter pointing back to its original queue. This means that messages are put in the retry queue and no handlers ever pick them up. When the TTL expires, Rabbit dead-letters them back to the main integration queue, where they are picked up and reprocessed. Integration messages contain the number of retries, so we don't do this endlessly. 

Although this implementation adds things like `NotBeforeUtc` to support scheduling messages and obeying `Retry-After` headers, we ignore that for the RabbitMQ implementation. Azure Service Bus will use these directly since it supports scheduling out of the box. RabbitMQ needs a special package to add delay scheduling which adds complexity that self-hosted instances probably do not need.

We are also using one dead-letter-queue for all integrations. We could make this per-integration if we wanted to, it just adds more complexity in the configuration / global settings. Since messages in the DLQ are really just for information purposes (i.e. we're not planning on programmatically re-queueing them), it seemed cleaner to put them all in one place where they could be viewed.

Lastly, this PR also cleans up some of the Startup.cs files with some methods in our ServiceCollectionExtensions.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Startup.cs
+++ b/src/Api/Startup.cs
@@ -28,10 +28,8 @@ using Bit.Core.Tools.Entities;
 using Bit.Core.Vault.Entities;
 using Bit.Api.Auth.Models.Request.WebAuthn;
 using Bit.Api.Billing;
-using Bit.Core.AdminConsole.Services.NoopImplementations;
 using Bit.Core.Auth.Models.Data;
 using Bit.Core.Auth.Identity.TokenProviders;
-using Bit.Core.Services;
 using Bit.Core.Tools.ImportFeatures;
 using Bit.Core.Tools.ReportFeatures;
 using Bit.Core.Auth.Models.Api.Request;
@@ -222,18 +220,7 @@ public class Startup
             services.AddHostedService<Core.HostedServices.ApplicationCacheHostedService>();
         }
 
-        // Slack
-        if (CoreHelpers.SettingHasValue(globalSettings.Slack.ClientId) &&
-            CoreHelpers.SettingHasValue(globalSettings.Slack.ClientSecret) &&
-            CoreHelpers.SettingHasValue(globalSettings.Slack.Scopes))
-        {
-            services.AddHttpClient(SlackService.HttpClientName);
-            services.AddSingleton<ISlackService, SlackService>();
-        }
-        else
-        {
-            services.AddSingleton<ISlackService, NoopSlackService>();
-        }
+        services.AddSlackService(globalSettings);
     }
 
     public void Configure(
--- a/src/Core/AdminConsole/Models/Data/Integrations/IntegrationHandlerResult.cs
+++ b/src/Core/AdminConsole/Models/Data/Integrations/IntegrationHandlerResult.cs
@@ -1,0 +1,16 @@
+﻿namespace Bit.Core.Models.Data.Integrations;
+
+public class IntegrationHandlerResult
+{
+    public IntegrationHandlerResult(bool success, IIntegrationMessage message)
+    {
+        Success = success;
+        Message = message;
+    }
+
+    public bool Success { get; set; } = false;
+    public bool Retryable { get; set; } = false;
+    public IIntegrationMessage Message { get; set; }
+    public DateTime? NotBeforeUtc { get; set; }
+    public string FailureReason { get; set; } = string.Empty;
+}
--- a/src/Core/AdminConsole/Models/Data/Integrations/IntegrationMessage.cs
+++ b/src/Core/AdminConsole/Models/Data/Integrations/IntegrationMessage.cs
@@ -1,0 +1,44 @@
+﻿using System.Text.Json;
+using Bit.Core.Enums;
+
+namespace Bit.Core.Models.Data.Integrations;
+
+public interface IIntegrationMessage
+{
+    IntegrationType IntegrationType { get; }
+    int RetryCount { get; set; }
+    DateTime? NotBeforeUtc { get; set; }
+    void ApplyRetry(DateTime? handlerNotBeforeUtc);
+    string ToJson();
+}
+
+public class IntegrationMessage<T> : IIntegrationMessage
+{
+    public IntegrationType IntegrationType { get; set; }
+    public T Configuration { get; set; }
+    public string RenderedTemplate { get; set; }
+    public int RetryCount { get; set; } = 0;
+    public DateTime? NotBeforeUtc { get; set; }
+
+    public void ApplyRetry(DateTime? handlerNotBeforeUtc)
+    {
+        RetryCount++;
+
+        var baseTime = handlerNotBeforeUtc ?? DateTime.UtcNow;
+        var backoffSeconds = Math.Pow(2, RetryCount);
+        var jitterSeconds = Random.Shared.Next(0, 3);
+
+        NotBeforeUtc = baseTime.AddSeconds(backoffSeconds + jitterSeconds);
+    }
+
+    public string ToJson()
+    {
+        return JsonSerializer.Serialize(this);
+    }
+
+    public static IntegrationMessage<T> FromJson(string json)
+    {
+        return JsonSerializer.Deserialize<IntegrationMessage<T>>(json)
+               ?? throw new InvalidOperationException("Invalid JSON payload.");
+    }
+}
--- a/src/Core/AdminConsole/Services/Implementations/EventIntegrationHandler.cs
+++ b/src/Core/AdminConsole/Services/Implementations/EventIntegrationHandler.cs
@@ -1,0 +1,76 @@
+﻿using System.Text.Json;
+using Bit.Core.AdminConsole.Utilities;
+using Bit.Core.Enums;
+using Bit.Core.Models.Data;
+using Bit.Core.Models.Data.Integrations;
+using Bit.Core.Repositories;
+
+namespace Bit.Core.Services;
+
+public class EventIntegrationHandler<T>(
+    IntegrationType integrationType,
+    IIntegrationPublisher integrationPublisher,
+    IOrganizationIntegrationConfigurationRepository configurationRepository,
+    IUserRepository userRepository,
+    IOrganizationRepository organizationRepository)
+    : IEventMessageHandler
+{
+    public async Task HandleEventAsync(EventMessage eventMessage)
+    {
+        var organizationId = eventMessage.OrganizationId ?? Guid.Parse("f431e04c-f2c3-473c-8cd1-b291014b0236");
+        var configurations = await configurationRepository.GetConfigurationDetailsAsync(
+            organizationId,
+            integrationType,
+            eventMessage.Type);
+
+        foreach (var configuration in configurations)
+        {
+            var context = await BuildContextAsync(eventMessage, configuration.Template);
+            var renderedTemplate = IntegrationTemplateProcessor.ReplaceTokens(configuration.Template, context);
+
+            var config = configuration.MergedConfiguration.Deserialize<T>()
+                         ?? throw new InvalidOperationException($"Failed to deserialize to {typeof(T).Name}");
+
+            var message = new IntegrationMessage<T>
+            {
+                IntegrationType = integrationType,
+                Configuration = config,
+                RenderedTemplate = renderedTemplate,
+                RetryCount = 0,
+                NotBeforeUtc = null
+            };
+
+            await integrationPublisher.PublishAsync(message);
+        }
+    }
+
+    public async Task HandleManyEventsAsync(IEnumerable<EventMessage> eventMessages)
+    {
+        foreach (var eventMessage in eventMessages)
+        {
+            await HandleEventAsync(eventMessage);
+        }
+    }
+
+    private async Task<IntegrationTemplateContext> BuildContextAsync(EventMessage eventMessage, string template)
+    {
+        var context = new IntegrationTemplateContext(eventMessage);
+
+        if (IntegrationTemplateProcessor.TemplateRequiresUser(template) && eventMessage.UserId.HasValue)
+        {
+            context.User = await userRepository.GetByIdAsync(eventMessage.UserId.Value);
+        }
+
+        if (IntegrationTemplateProcessor.TemplateRequiresActingUser(template) && eventMessage.ActingUserId.HasValue)
+        {
+            context.ActingUser = await userRepository.GetByIdAsync(eventMessage.ActingUserId.Value);
+        }
+
+        if (IntegrationTemplateProcessor.TemplateRequiresOrganization(template) && eventMessage.OrganizationId.HasValue)
+        {
+            context.Organization = await organizationRepository.GetByIdAsync(eventMessage.OrganizationId.Value);
+        }
+
+        return context;
+    }
+}
--- a/src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
+++ b/src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
@@ -1,0 +1,180 @@
+﻿using System.Text;
+using Bit.Core.Settings;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using RabbitMQ.Client;
+using RabbitMQ.Client.Events;
+
+namespace Bit.Core.Services;
+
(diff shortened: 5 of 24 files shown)


src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
```diff
@@ -0,0 +1,180 @@
...
+        // Declare retry queue (30s TTL, dead-letters back to main queue)
+        await _channel.QueueDeclareAsync(queue: _retryQueueName,
+                                         durable: true,
+                                         exclusive: false,
+                                         autoDelete: false,
+                                         arguments: new Dictionary<string, object>
+                                         {
+                                             { "x-dead-letter-exchange", _exchangeName },
+                                             { "x-dead-letter-routing-key", _routingKey },
+                                             { "x-message-ttl", 30000 } // 30s
```
withinfocus, line 83: ❓ Should this be configurable?

src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
```diff
@@ -0,0 +1,180 @@
...
+    protected override async Task ExecuteAsync(CancellationToken cancellationToken)
+    {
+        var consumer = new AsyncEventingBasicConsumer(_channel);
+        consumer.ReceivedAsync += async (_, ea) =>
+        {
+            var json = Encoding.UTF8.GetString(ea.Body.Span);
+
+            try
+            {
+                var result = await _handler.HandleAsync(json);
```
withinfocus, line 115: 🎨 Comments throughout this flow will help point out what it's doing and why.

src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    await PublishToDeadLetterAsync(message.ToJson());
+                    _logger.LogWarning("Non-retryable failure. Sent to DLQ.");
+                }
+
+                await _channel.BasicAckAsync(ea.DeliveryTag, false, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Unhandled error processing integration message");
```
withinfocus, line 152:
```suggestion
                _logger.LogError(ex, "Unhandled error processing integration message.");
```

⛏️ Maybe? Let's just be consistent.

src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
```diff
@@ -0,0 +1,180 @@
...
+                await _channel.BasicAckAsync(ea.DeliveryTag, false, cancellationToken);
+            }
+        };
+
+        await _channel.BasicConsumeAsync(queue: _queueName, autoAck: false, consumer: consumer, cancellationToken: cancellationToken);
+    }
+    private async Task PublishToDeadLetterAsync(string json)
```
withinfocus, line 159: ⛏️ Newline between methods.

src/Core/AdminConsole/Services/Implementations/WebhookIntegrationHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+            case HttpStatusCode.TooManyRequests:
+            case HttpStatusCode.RequestTimeout:
+            case HttpStatusCode.InternalServerError:
+            case HttpStatusCode.BadGateway:
+            case HttpStatusCode.ServiceUnavailable:
+            case HttpStatusCode.GatewayTimeout:
+                result.Retryable = true;
+                result.FailureReason = response.ReasonPhrase;
+
+                if (response.Headers.TryGetValues("Retry-After", out var values) &&
```
withinfocus, line 33: 🎨 I had forgotten about this but it's handy. What if it's a [date](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After#http-date) though?

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -582,6 +548,193 @@ public static GlobalSettings AddGlobalSettingsServices(this IServiceCollection s
...
+        }
+        else
+        {
+            services.AddSingleton<ISlackService, NoopSlackService>();
+        }
+
+        return services;
+    }
+
+    public static IServiceCollection AddRabbitMqEventRepositoryListener(this IServiceCollection services, GlobalSettings globalSettings)
```
withinfocus, line 673: 🎨 Looks like some of these can be `private` now.

withinfocus (changes requested):
Little things and questions.

The one thing I have been sorta staring at:

> In the RabbitMQ implementation, we are using a very simple retry queue. It has a TTL of 30 seconds with a dead-letter pointing back to its original queue. This means that messages are put in the retry queue and no handlers ever pick them up. When the TTL expires, Rabbit dead-letters them back to the main integration queue, where they are picked up and reprocessed.

Is this standard / normal setup with Rabbit? And per your other notes, we can't use the `NotBeforeUtc` instead of the 30s that's hardcoded? I am trying to think about how we can compensate for some of these gaps if this is all Rabbit can do.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -582,6 +548,193 @@ public static GlobalSettings AddGlobalSettingsServices(this IServiceCollection s
...
+                globalSettings.EventLogging.RabbitMq.WebhookIntegrationRetryQueueName,
+                globalSettings.EventLogging.RabbitMq.IntegrationDeadLetterQueueName,
+                IntegrationType.Webhook,
+                globalSettings);
+        }
+
+        return services;
+    }
+
+    public static IServiceCollection AddSlackService(this IServiceCollection services, GlobalSettings globalSettings)
```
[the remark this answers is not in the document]
brant-livefront: This one actually is _just_ the SlackService - i.e. not necessarily related to any of the other integration pieces. This is what powers the Slack integration, but also what powers the OAuth and channel/email lookups and such.

src/Api/Startup.cs
```diff
@@ -222,18 +220,7 @@ public void ConfigureServices(IServiceCollection services)
...
-            CoreHelpers.SettingHasValue(globalSettings.Slack.Scopes))
-        {
-            services.AddHttpClient(SlackService.HttpClientName);
-            services.AddSingleton<ISlackService, SlackService>();
-        }
-        else
-        {
-            services.AddSingleton<ISlackService, NoopSlackService>();
-        }
+        services.AddSlackService(globalSettings);
```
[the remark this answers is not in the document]
brant-livefront: 👍 Good note, I'll add a comment that we need this to power the OAuth API.

src/Core/AdminConsole/Models/Data/Integrations/IntegrationMessage.cs
```diff
@@ -0,0 +1,44 @@
...
+
+public class IntegrationMessage<T> : IIntegrationMessage
+{
+    public IntegrationType IntegrationType { get; set; }
+    public T Configuration { get; set; }
+    public string RenderedTemplate { get; set; }
+    public int RetryCount { get; set; } = 0;
+    public DateTime? NotBeforeUtc { get; set; }
+
+    public void ApplyRetry(DateTime? handlerNotBeforeUtc)
```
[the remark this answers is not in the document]
brant-livefront: While it's always supplied it can be null - because the `NotBeforeUtc` in the result can be null. My idea of the contract is that the handler only sets a `NotBeforeUtc` if they have a concrete reason to (e.g. `retry-after` header). If they get something like a 500 error where we can retry but we don't have a specific delay we know of, the handler just leaves it null.

We could remove nullability here if we just always default to `UtcNow`. The handler could return `UtcNow` explicitly unless it has a known delay. That's functionally equivalent to what we're doing now since we default to `UtcNow` if it's null in ApplyRetry. Let me know if you want to go that route and just remove all of the nulls.

src/Core/AdminConsole/Models/Data/Integrations/IntegrationMessage.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public string ToJson()
+    {
+        return JsonSerializer.Serialize(this);
+    }
+
+    public static IntegrationMessage<T> FromJson(string json)
+    {
+        return JsonSerializer.Deserialize<IntegrationMessage<T>>(json)
+               ?? throw new InvalidOperationException("Invalid JSON payload.");
```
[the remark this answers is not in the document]
brant-livefront: Good catch - actually this Deserialize will not return `null`, it'll throw. In fact, I have a unit test that verifies it throws the JsonException, so this default can be removed. 👍

src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
brant-livefront: I'll add a configurable amount for this with 30 as the default.  👍 

Technically, you could create the queue on your own as an admin and this would _not_ override your settings (you still specify this in code for Rabbit, but if it already exists, it just uses what's already there). But it seems like that's asking a lot of people setting this up if all they wanted was to adjust this one thing.

src/Core/AdminConsole/Services/Implementations/WebhookIntegrationHandler.cs
brant-livefront: Good catch. 👍 I will add handling for that as well.

src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
```diff
@@ -117,16 +119,19 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
 
                 if (result.Success)
                 {
+                    // Successful integration send. Acknowledge message delivery and return.
```
withinfocus, line 122:
```suggestion
                    // Successful integration send. Acknowledge message delivery and return
```

Really ⛏️y.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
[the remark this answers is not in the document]
withinfocus: Yeah, this comment is really just a "what if" when we add Teams, etc.

src/Core/AdminConsole/Models/Data/Integrations/IntegrationHandlerResult.cs
```diff
@@ -0,0 +1,16 @@
...
+    public IntegrationHandlerResult(bool success, IIntegrationMessage message)
+    {
+        Success = success;
+        Message = message;
+    }
+
+    public bool Success { get; set; } = false;
+    public bool Retryable { get; set; } = false;
+    public IIntegrationMessage Message { get; set; }
+    public DateTime? NotBeforeUtc { get; set; }
```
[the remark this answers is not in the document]
withinfocus: `DelayUntilDate` sounds good to me.

src/Core/AdminConsole/Models/Data/Integrations/IntegrationMessage.cs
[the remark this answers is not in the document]
withinfocus: I missed that return case -- this works as-is then, and making the decision to use "now" here vs. setting it in the caller.

withinfocus (changes requested):
Almost there.

withinfocus (dismissed):

src/Core/AdminConsole/Services/Implementations/EventIntegrationHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+    IOrganizationRepository organizationRepository)
+    : IEventMessageHandler
+{
+    public async Task HandleEventAsync(EventMessage eventMessage)
+    {
+        var organizationId = eventMessage.OrganizationId ?? Guid.Empty;
+        var configurations = await configurationRepository.GetConfigurationDetailsAsync(
+            organizationId,
+            integrationType,
+            eventMessage.Type);
```
eliykat, line 24: I assume if the Guid is null/empty, the repository call won't return anything; should we just return early instead?

src/Core/AdminConsole/Models/Data/Integrations/IntegrationHandlerResult.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Bit.Core.Models.Data.Integrations;
```
eliykat, line 1: Namespace does not match the folder structure - looks like this is the case for most files here.

src/Core/AdminConsole/Services/Implementations/WebhookIntegrationHandler.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System.Globalization;
+using System.Net;
+using System.Text;
+using Bit.Core.Models.Data.Integrations;
+
+#nullable enable
```
eliykat, line 6: Good to see the nullable context here. I actually thought we had some documentation requiring it or at least encouraging it but we don't. Can it be enabled on more files here? (my view is it should be enabled on all new files even though it's tedious)

eliykat (changes requested):
I appreciate the detailed PR description, it provided much needed context. I'm mostly just following along here and relying on @withinfocus' review for the detail. The issues below are pretty minor so I'm happy to discuss.

src/Core/AdminConsole/Services/Implementations/WebhookIntegrationHandler.cs
brant-livefront: I added it to a few. I wasn't sure what the convention was, but I'm happy to enable it in more places. 👍 

I have 2 other branches that are hooked off this PR, one of which is a bigger refactor. If you don't mind I'd like to add it to more places, but in that other branch where I'm a little bit more confident in not causing big merge conflicts.

eliykat: That's fine with me, thanks. I followed up on it internally and apparently there's no firm requirement for it at the moment, although fwiw I have a strong preference towards using it where possible. This is good enough to get on with it though :)

eliykat (approved):

PR #5800, opened 2025-05-09
Ac/pm 19145 refactor organization service.import async

BTreston:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
```diff
@@ -77,6 +76,42 @@ await eventService.LogOrganizationUserEventAsync<IOrganizationUser>(
...
+                        new Error<InviteOrganizationUsersResponse>(
+                            failure.Error.Message,
+                            new InviteOrganizationUsersResponse(failure.Error.ErroredValue.InvitedUsers, organizationId)
+                            )
+                        );
+
+            case Success<InviteOrganizationUsersResponse> success when success.Value.InvitedUsers.Any():
+
+                // add a bulk method?
+                foreach (var user in success.Value.InvitedUsers)
```
jrmccannon, line 96: There's already a bulk method.

```
IEventService
...
    Task LogOrganizationUserEventsAsync<T>(IEnumerable<(T, EventType, EventSystemUser, DateTime?)> events) where T : IOrganizationUser;```

File: src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs [hunk 2]
```diff
@@ -0,0 +1,351 @@
...
+            default:
+                throw new InvalidOperationException($"Unhandled commandResult type: {commandResult.GetType().Name}");
+        }
+    }
+
+    private async Task<CommandResult<InviteOrganizationUsersResponse>> InviteUsersAsync(List<OrganizationUserInviteCommandModel> invites, Organization organization)
+    {
+        var plan = await _pricingClient.GetPlanOrThrow(organization.PlanType);
+        var inviteOrganization = new InviteOrganization(organization, plan);
+        var request = new InviteOrganizationUsersRequest(invites.ToArray(), inviteOrganization, organization.Id, DateTimeOffset.UtcNow);
```
BTreston, line 216: Should this 3rd param (`PerformedBy`) be `organization.Id` or `Guid.Empty`?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,337 @@
...
+        }
+        await _organizationUserRepository.UpsertManyAsync(usersToUpsert);
+    }
+
+    private async Task AddNewUsers(Organization organization,
+            IEnumerable<ImportedOrganizationUser> newUsers,
+            OrganizationUserImportData importUserData)
+    {
+
+        var hasStandaloneSecretsManager = await _paymentService.HasSecretsManagerStandalone(organization);
```
jrmccannon: You don't need to do this. It is handled inside of InviteOrgUsers

https://github.com/bitwarden/server/blob/6411cc63ca0af7a21386214eb9a8e31e8e219f44/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Validation/InviteOrganizationUserValidator.cs#L36

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        var existingUsers = await _organizationUserRepository.GetManyDetailsByOrganizationAsync(organizationId);
+
+        var importUserData = new OrganizationUserImportData
+        {
+            NewUsersSet = new HashSet<string>(newUsers?.Select(u => u.ExternalId) ?? new List<string>()),
+            ExistingUsers = existingUsers,
+            ExistingExternalUsers = GetExistingExternalUsers(existingUsers),
+            ExistingExternalUsersIdDict = GetExistingExternalUsers(existingUsers).ToDictionary(u => u.ExternalId, u => u.Id)
```
jrmccannon: You could maybe move this to a constructor and have it pass the `newUsers` and `existingUsers` as parameters. That way you don't have to call `GetExistingExternalUsers` twice.

```
ExistingUsers = existingUsers,
ExistingExternalUsers = GetExistingExternalUsers(existingUsers),
ExistingExternalUsersIdDict = ExistingExternalUsers.ToDictionary(u => u.ExternalId, u => u.Id)
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,337 @@
...
+        var removeUsersSet = new HashSet<string>(removeUserExternalIds)
+            .Except(importUserData.NewUsersSet)
+            .Where(u => existingUsersDict.ContainsKey(u) && existingUsersDict[u].Type != OrganizationUserType.Owner)
+            .Select(u => existingUsersDict[u]);
+
+        await _organizationUserRepository.DeleteManyAsync(removeUsersSet.Select(u => u.Id));
+        events.AddRange(removeUsersSet.Select(u => (
+          u,
+          EventType.OrganizationUser_Removed,
+          (DateTime?)DateTime.UtcNow
```
jrmccannon: Instead of `DateTime.UtcNow` you can use the TimeProvider.  It allows you to override it in tests.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,337 @@
...
+            .Where(u => string.IsNullOrWhiteSpace(u.ExternalId))
+            .ToDictionary(u => u.Email);
+        var newUsersEmailsDict = newUsers.ToDictionary(u => u.Email);
+        var newAndExistingUsersIntersection = existingUsersEmailsDict.Keys.Intersect(newUsersEmailsDict.Keys).ToList();
+        var organizationUsers = (await _organizationUserRepository.GetManyAsync(importUserData.ExistingUsers.Select(u => u.Id).ToList())).ToDictionary(u => u.Id);
+        var usersToUpsert = new List<OrganizationUser>();
+
+        foreach (var user in newAndExistingUsersIntersection)
+        {
+            var organizationUser = organizationUsers[existingUsersEmailsDict[user].Id];
```
jrmccannon: If the user doesn't exist, I think this will throw an exception. You want to use `TryGetValue()` instead

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,337 @@
...
+        await _organizationUserRepository.DeleteManyAsync(removeUsersSet.Select(u => u.Id));
+        events.AddRange(removeUsersSet.Select(u => (
+          u,
+          EventType.OrganizationUser_Removed,
+          (DateTime?)DateTime.UtcNow
+          ))
+        );
+    }
+
+    private async Task UpsertExistingUsers(Organization organization,
```
jrmccannon: organization parameter is unused

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,337 @@
...
+        var events = new List<(OrganizationUserUserDetails ou, EventType e, DateTime? d)>();
+
+        await RemoveExistingExternalUsers(removeUserExternalIds, events, importUserData);
+
+        if (overwriteExisting)
+        {
+            await OverwriteExisting(events, importUserData);
+        }
+
+        await UpsertExistingUsers(organization, newUsers, importUserData);
```
jrmccannon: So the point of this step is to find all existing users that match the email provided and create new org users for them if they don't exist for this organization?

jrmccannon (changes requested):
Just a first pass on this. I'll continue digging in tomorrow.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
BTreston: Essentially we are just updating the existing org users (if any) to have the updated `externalId` that the corresponding user in the `importData` has.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
BTreston: There are some spots where we have to continue using DateTime.UtcNow until other parts of the system are refactored (calling out the  `Groups` object specifically). This is probably best left for a followup PR.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+            return;
+        }
+
+        // Marry existing users
+        var existingUsersEmailsDict = importUserData.ExistingUsers
+            .Where(u => string.IsNullOrWhiteSpace(u.ExternalId))
+            .ToDictionary(u => u.Email);
+        var newUsersEmailsDict = newUsers.ToDictionary(u => u.Email);
+        var newAndExistingUsersIntersection = existingUsersEmailsDict.Keys.Intersect(newUsersEmailsDict.Keys).ToList();
+        var organizationUsers = (await _organizationUserRepository.GetManyAsync(importUserData.ExistingUsers.Select(u => u.Id).ToList())).ToDictionary(u => u.Id);
```
eliykat: Good call using the bulk repository method. Comparing to the current logic, should this use the `newAndExistingUsersIntersection` rather than `importUserData.ExistingUsers`? This would also match the users you're looping over below.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+        _eventService = eventService;
+        _currentContext = currentContext;
+        _organizationService = organizationService;
+        _inviteOrganizationUsersCommand = inviteOrganizationUsersCommand;
+        _pricingClient = pricingClient;
+    }
+
+    public async Task ImportAsync(Guid organizationId,
+        IEnumerable<ImportedGroup> groups,
+        IEnumerable<ImportedOrganizationUser> newUsers,
```
eliykat: `newUsers` threw me for a bit. They're not necessarily new users (they may already exist in the database), they're just the users we've been asked to import. I suggest `importedUsers` for clarity (and therefore `importedGroups` for consistency).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+        await _organizationUserRepository.DeleteManyAsync(removeUsersSet.Select(u => u.Id));
+        events.AddRange(removeUsersSet.Select(u => (
+          u,
+          EventType.OrganizationUser_Removed,
+          (DateTime?)DateTime.UtcNow
+          ))
+        );
+    }
+
+    private async Task UpsertExistingUsers(IEnumerable<ImportedOrganizationUser> newUsers, OrganizationUserImportData importUserData)
```
eliykat: I think `Update` is more accurate - if we're only upserting existing users then we're not inserting anything.

eliykat: Also, to carry the renaming down, this would be `importedUsers` rather than `newUsers`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+        {
+            return;
+        }
+
+        // Marry existing users
+        var existingUsersEmailsDict = importUserData.ExistingUsers
+            .Where(u => string.IsNullOrWhiteSpace(u.ExternalId))
+            .ToDictionary(u => u.Email);
+        var newUsersEmailsDict = newUsers.ToDictionary(u => u.Email);
+        var newAndExistingUsersIntersection = existingUsersEmailsDict.Keys.Intersect(newUsersEmailsDict.Keys).ToList();
```
eliykat: Continuing the renaming: instead of describing this as `newAndExistingUsersIntersection` (which was the source of my initial confusion), this can simply be called `userDetailsToUpdate` (or similar).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    private async Task UpsertExistingUsers(IEnumerable<ImportedOrganizationUser> newUsers, OrganizationUserImportData importUserData)
+    {
+        if (!newUsers.Any())
+        {
+            return;
+        }
+
+        // Marry existing users
```
eliykat: Please move this to an xmldoc and explain a bit more, e.g.

> Find existing members in the imported users set by matching on their email, and update their externalId.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
eliykat, line 1: You've done a good job breaking up the existing code into separate methods. That said, this logic is pretty dense and not always easy to follow. Can you please add comments/xmldoc throughout to try to step the reader through the logic?

src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserImportData.cs
```diff
@@ -0,0 +1,24 @@
...
+    public ICollection<OrganizationUserUserDetails> ExistingUsers { get; init; }
+    public IEnumerable<OrganizationUserUserDetails> ExistingExternalUsers { get; init; }
+    public Dictionary<string, Guid> ExistingExternalUsersIdDict { get; init; }
+
+    public OrganizationUserImportData(ICollection<OrganizationUserUserDetails> existingUsers, HashSet<string> newUsersSet)
+    {
+        NewUsersSet = newUsersSet;
+        ExistingUsers = existingUsers;
+        ExistingExternalUsers = GetExistingExternalUsers(existingUsers);
+        ExistingExternalUsersIdDict = GetExistingExternalUsers(existingUsers).ToDictionary(u => u.ExternalId, u => u.Id);
```
eliykat: This is calling `GetExistingExternalUsers` twice. Instead:

```c#
ExistingExternalUsers = GetExistingExternalUsers(existingUsers);
ExistingExternalUsersIdDict = ExistingExternalUsers.ToDictionary(u => u.ExternalId, u => u.Id);
```

src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserImportData.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace Bit.Core.Models.Data.Organizations.OrganizationUsers;
+
+public class OrganizationUserImportData
+{
+    public HashSet<string> NewUsersSet { get; init; }
+    public ICollection<OrganizationUserUserDetails> ExistingUsers { get; init; }
+    public IEnumerable<OrganizationUserUserDetails> ExistingExternalUsers { get; init; }
+    public Dictionary<string, Guid> ExistingExternalUsersIdDict { get; init; }
```
eliykat: I think these all can/should be `readonly`. xmldoc would also be really useful (e.g. `HashSet<string>` is pretty vague)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+            throw new NotFoundException();
+        }
+
+        if (!organization.UseDirectory)
+        {
+            throw new BadRequestException("Organization cannot use directory syncing.");
+        }
+
+        var existingUsers = await _organizationUserRepository.GetManyDetailsByOrganizationAsync(organizationId);
+        var importUserData = new OrganizationUserImportData(existingUsers, new HashSet<string>(newUsers?.Select(u => u.ExternalId) ?? new List<string>()));
```
eliykat: The `HashSet` construction is better done in the constructor of `OrganizationUserImportData`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+    private async Task AddNewUsers(Organization organization,
+            IEnumerable<ImportedOrganizationUser> newUsers,
+            OrganizationUserImportData importUserData)
+    {
+        var userInvites = new List<OrganizationUserInviteCommandModel>();
+
+        foreach (var user in newUsers)
+        {
+            userInvites.Add(new OrganizationUserInviteCommandModel(user.Email, user.ExternalId));
+        }
```
eliykat: I don't think this matches existing logic - this is inviting all imported users, whereas the existing code excludes existing users: https://github.com/bitwarden/server/blob/66301bb4dc1ec7ea99d55301b14cfaeac216f1a4/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1172-L1175

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    private async Task<CommandResult<InviteOrganizationUsersResponse>> InviteUsersAsync(List<OrganizationUserInviteCommandModel> invites, Organization organization)
+    {
+        var plan = await _pricingClient.GetPlanOrThrow(organization.PlanType);
+        var inviteOrganization = new InviteOrganization(organization, plan);
+        var request = new InviteOrganizationUsersRequest(invites.ToArray(), inviteOrganization, Guid.Empty, DateTimeOffset.UtcNow);
+
+        return await _inviteOrganizationUsersCommand.InviteImportedOrganizationUsersAsync(request);
```
eliykat: @jrmccannon do we want to use this yet? The invite users command is itself behind a feature flag that hasn't been turned on yet, and I think you've still got a bit of work to do on it. I'm leaning towards using the existing invite user flow just so that we don't have stacked feature flags or PRs blocking eachother.

eliykat (changes requested):
This looks great so far. This is a partial review only but I should be able to finish it tomorrow.

General comments:

1. Please use `#nullable enable` on all new files: https://contributing.bitwarden.com/architecture/adr/csharp-nullable-reference-types
2. While I can see you've largely kept the existing implementation, there's enough refactoring here that it's worth putting this behind a feature flag.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
jrmccannon: Now that you point it out, yeah should probably hold off on it.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,461 @@
...
+        foreach (var invitedUser in invitedUsers)
+        {
+            importUserData.ExistingExternalUsersIdDict.TryAdd(invitedUser.ExternalId!, invitedUser.Id);
+        }
+    }
+
+    /// <summary>
+    /// Creates an InviteOrganizationUsersRequest for the provided invites and sends the request via the InviteOrganizationUsersCommand.
+    /// </summary>
+    private async Task vNextInviteUsersAsync(List<OrganizationUserInviteCommandModel> userInvites,
```
BTreston, line 276: I elected to leave the command call in as a `vNext` method. The implementation can be a little less disorganized, but figured I'd get some feedback first.

src/Api/AdminConsole/Public/Controllers/OrganizationController.cs
```diff
@@ -46,12 +55,26 @@ public async Task<IActionResult> Import([FromBody] OrganizationImportRequestMode
             throw new BadRequestException("You cannot import this much data at once.");
         }
 
-        await _importOrganizationUserCommand.ImportAsync(
-            _currentContext.OrganizationId.Value,
-            model.Groups.Select(g => g.ToImportedGroup(_currentContext.OrganizationId.Value)),
-            model.Members.Where(u => !u.Deleted).Select(u => u.ToImportedOrganizationUser()),
-            model.Members.Where(u => u.Deleted).Select(u => u.ExternalId),
-            model.OverwriteExisting.GetValueOrDefault());
+        if (_featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization))
```
eliykat: This work is separate to @jrmccannon 's invite user work. We don't want to put too much behind 1 feature flag. You can define your own feature flag and let Vince know so he can create it in LaunchDarkly. Really we can create feature flags whenever we feel like we need them, the important thing is that they're small and short-lived.

src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserImportData.cs
```diff
@@ -1,24 +1,30 @@
...
-    public HashSet<string> NewUsersSet { get; init; }
-    public ICollection<OrganizationUserUserDetails> ExistingUsers { get; init; }
-    public IEnumerable<OrganizationUserUserDetails> ExistingExternalUsers { get; init; }
-    public Dictionary<string, Guid> ExistingExternalUsersIdDict { get; init; }
+    /// <summary>
+    /// Set of user ExternalIds that are being imported
+    /// </summary>
+    public readonly HashSet<string> ImportedExternalIds;
+    /// <summary>
+    /// Exising organization users details
```
eliykat: typo, and while I'm at it:

```suggestion
    /// All existing OrganizationUsers for the organization
```

src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserImportData.cs
```diff
@@ -1,24 +1,30 @@
...
+    /// </summary>
+    public readonly ICollection<OrganizationUserUserDetails> ExistingUsers;
+    /// <summary>
+    /// List of ExternalIds belonging to existing organization Users
+    /// </summary>
+    public readonly IEnumerable<OrganizationUserUserDetails> ExistingExternalUsers;
```
eliykat: Inaccurate xmldoc - it's not a list of ExternalIds.

```suggestion
    /// <summary>
    /// Existing OrganizationUsers with ExternalIds set.
    /// </summary>
    public readonly IEnumerable<OrganizationUserUserDetails> ExistingExternalUsers;
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -128,44 +137,71 @@ private async Task RemoveExistingExternalUsers(IEnumerable<string> removeUserExt
...
             .ToDictionary(u => u.Email);
-        var newUsersEmailsDict = newUsers.ToDictionary(u => u.Email);
-        var newAndExistingUsersIntersection = existingUsersEmailsDict.Keys.Intersect(newUsersEmailsDict.Keys).ToList();
-        var organizationUsers = (await _organizationUserRepository.GetManyAsync(importUserData.ExistingUsers.Select(u => u.Id).ToList())).ToDictionary(u => u.Id);
-        var usersToUpsert = new List<OrganizationUser>();
+        var importedUsersEmailsDict = importedUsers.ToDictionary(u => u.Email);
 
-        foreach (var user in newAndExistingUsersIntersection)
+        // Determine which users to update.
+        var userDetailsToUpdate = existingUsersEmailsDict.Keys.Intersect(importedUsersEmailsDict.Keys).ToList();
```
eliykat: Bad naming recommendation by me: really this is `userEmailsToUpdate`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -128,44 +137,71 @@ private async Task RemoveExistingExternalUsers(IEnumerable<string> removeUserExt
...
+        var userIdsToUpdate = importUserData.ExistingUsers
+            .Where(u => userDetailsToUpdate.Contains(u.Email))
+            .Select(u => u.Id)
+            .ToList();
```
eliykat: You could just pull these out of the dictionary, which is faster and closer to how the existing code works.

```c#
var userIdsToUpdate = userEmailsToUpdate.Select(e => existingUsersEmailsDict[e].Id).ToList();
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -128,44 +137,71 @@ private async Task RemoveExistingExternalUsers(IEnumerable<string> removeUserExt
...
+        var organizationUsers = (await _organizationUserRepository.GetManyAsync(userIdsToUpdate)).ToDictionary(u => u.Id);
+
+        foreach (var userEmail in userDetailsToUpdate)
         {
-            existingUsersEmailsDict.TryGetValue(user, out var existingUser);
-            organizationUsers.TryGetValue(existingUser.Id, out var organizationUser);
+            // verify userEmail has an associated OrganizationUser
+            existingUsersEmailsDict.TryGetValue(userEmail, out var existingUser);
+            organizationUsers.TryGetValue(existingUser!.Id, out var organizationUser);
+            importedUsersEmailsDict.TryGetValue(userEmail, out var user);
```
eliykat: Lots of users floating around here, better to be explicit:

```suggestion
            importedUsersEmailsDict.TryGetValue(userEmail, out var importedUser);
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -16,6 +16,8 @@
 using Bit.Core.Repositories;
 using Bit.Core.Services;
 
+#nullable enable
```
eliykat: Should be the first line in the file.
The other new files here need this as well, even models.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -82,27 +95,21 @@ public async Task ImportAsync(Guid organizationId,
             await OverwriteExisting(events, importUserData);
         }
 
-        await UpsertExistingUsers(newUsers, importUserData);
+        await Update(importedUsers, importUserData);
```
eliykat: ⛏️  I should've been more explicit: `UpdateExistingUsers`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -218,33 +296,45 @@ private async Task OverwriteExisting(
...
-        if (!groups.Any())
+        if (!importedGroups.Any())
         {
             return;
         }
 
         if (!organization.UseGroups)
         {
-            throw new BadRequestException("Organization cannot use groups.");
+            throw new BadRequestException("Organization cannot use importedGroups.");
```
eliykat: Error message should be unchanged.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,428 @@
...
+    /// Imports group data into the organization by saving new groups and updating existing ones.
+    /// </summary>
+    /// <param name="organization">The organization into which groups are being imported.</param>
+    /// <param name="importedGroups">A collection of groups to be imported.</param>
+    /// <param name="importUserData">Data containing information about existing and imported users.</param>
+    private async Task ImportGroups(Organization organization,
+            IEnumerable<ImportedGroup> importedGroups,
+            OrganizationUserImportData importUserData)
+    {
+
```
eliykat: Errant whitespace

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -0,0 +1,428 @@
...
+    /// <summary>
+    /// Saves newly imported groups that do not already exist in the organization.
+    /// Sets their creation and revision dates, associates users with each group.
+    /// </summary>
+    /// <param name="importGroupData">Data containing both imported and existing groups.</param>
+    /// <param name="importUserData">Data containing information about existing and imported users.</param>
+    private async Task SaveNewGroups(OrganizationGroupImportData importGroupData, OrganizationUserImportData importUserData)
+    {
+        var newGroups = importGroupData.Groups
+            .Where(g => !importGroupData.ExistingExternalGroups.ToDictionary(g => g.ExternalId!).ContainsKey(g.Group.ExternalId!))
```
eliykat: I assume that this will call `ToDictionary` every time it executes this lambda for each item in `importGroupData.Groups`. Maybe linq's smart enough to avoid this? But I assume not. Best to construct the dictionary outside of the callback.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
```diff
@@ -196,14 +237,51 @@ private async Task<CommandResult<InviteOrganizationUsersResponse>> InviteUsersAs
         return await _inviteOrganizationUsersCommand.InviteImportedOrganizationUsersAsync(request);
```
eliykat: This isn't the vnext method, so it should be using `OrganizationService`.

I don't think you need to worry about the new command here at all - we'll roll it out across all callers when it's ready, for now this code can just rely on the existing `OrganizationService` method. The command is only being used for scim at the moment.

src/Core/AdminConsole/Models/Data/Organizations/OrganizationGroupImportData.cs
eliykat, line 1: We've moved towards grouping all related files together by feature. Please move your new files (models, command, interface, etc) into `OrganizationFeatures/Import`. (Maybe it could be under `OrganizationFeatures/OrganizationUsers/Import`, but it imports both users and groups, so I think its own folder at this level is better.)

You can still create separate folders under `OrganizationFeatures/Import` for models etc. if useful. But then it's all in the same parent dir. Have a look at the other folders in `OrganizationFeatures` as examples.

src/Core/AdminConsole/Models/Data/Organizations/OrganizationGroupImportData.cs
```diff
@@ -0,0 +1,20 @@
...
+using Bit.Core.AdminConsole.Models.Business;
+
+namespace Bit.Core.Models.Data.Organizations;
+
+public class OrganizationGroupImportData
+{
+    public readonly IEnumerable<ImportedGroup> Groups;
+    public readonly ICollection<Group> ExistingGroups;
+    public readonly IEnumerable<Group> ExistingExternalGroups;
+    public readonly IDictionary<string, ImportedGroup> GroupsDict;
```
eliykat: xmldoc here would be great as well.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
eliykat, line 1: Similarly, I'd like to stash all of these changes for now, just to limit the scope of this PR. The goal here is to move the import code to its own command, not switch over to the new invite user flow. Hold on to them for later though. That includes changes to the new invite user models etc.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
eliykat: I think this is still an issue. ([context](https://github.com/bitwarden/server/pull/5800#discussion_r2128577928))

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommand.cs
eliykat, line 1: This imports groups as well - I suggest `ImportOrganizationUsersAndGroupsCommand`. Verbose but accurate.

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/ImportOrganizationUserCommandTests.cs
eliykat, line 1: These existing tests are pretty high level and I don't trust them a whole lot. Can you please add some Api integration tests - see the `test/Api.IntegrationTest` project (already contains some examples). Aim to have a mix of some new and existing users and groups, then assert (in detail) that they were updated as expected.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/InviteOrganizationUsersCommand.cs
BTreston: @eliykat If we want to reduce the scope here to just a lift and shift (which is how I interpret this) then I would suggest we create a new ticket for the lift and shift, put this ticket/PR on hold pending Jared's updates, and then revisit at another point when its more appropriate.

eliykat: No, your refactoring in the `ImportOrganizationUsersAndGroupsCommand` is great and I want to keep it. I mean that when you get to the point of inviting the new users you're adding, call the existing `OrganizationService.InviteUsersAsync` method (like the [current implementation does](https://github.com/bitwarden/server/blob/463dc1232d8097ac17b328259697c85c4a9fa038/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1222)) instead of using the new `InviteOrganizationUsersCommand`. Then, you no longer have to make any changes to the `InviteOrganizationUsersCommand` in this PR.

BTreston: Just so I understand clearly: 

- feature flag pivots between the existing `importAsync` call in `OrganizationService`and the refactored command
- all code related to the `InviteUserCommand` should be pulled and saved for later
- keep the refactored logic in the `ImportOrganizationUsersAndGroupsCommand`

eliykat: Yessir I think that's right!

test/Api.IntegrationTest/Helpers/OrganizationTestHelpers.cs
```diff
@@ -130,4 +132,20 @@ public static async Task CreateVerifiedDomainAsync(ApiApplicationFactory factory
...
+
+    public static async Task<Group> CreateGroup(ApiApplicationFactory factory, Guid organizationId)
+    {
+
+        var groupRepository = factory.GetService<IGroupRepository>();
+        var group = new Group
+        {
+            OrganizationId = organizationId,
+            Id = new Guid(),
+            ExternalId = new Guid().ToString(),
```
eliykat: Confusingly, `new Guid()` will give you a Guid which is all zeroes, which is probably not what you expect. `Guid.NewGuid()` will give you a random Guid, which is what you want here.

src/Core/AdminConsole/OrganizationFeatures/Import/ImportOrganizationUsersAndGroupsCommand.cs
```diff
@@ -199,73 +195,52 @@ private async Task AddNewUsers(Organization organization,
             IEnumerable<ImportedOrganizationUser> importedUsers,
             OrganizationUserImportData importUserData)
     {
-        var userInvites = new List<OrganizationUserInviteCommandModel>();
+        // Determine which users are already in the organization
+        var existingUsersSet = new HashSet<string>(importUserData.ExistingExternalUsersIdDict.Keys).ToList();
```
eliykat: Unnecessary list. (HashSet also forces enumeration if that's what you're worried about.)

```suggestion
        var existingUsersSet = new HashSet<string>(importUserData.ExistingExternalUsersIdDict.Keys);
```

src/Core/AdminConsole/OrganizationFeatures/Import/ImportOrganizationUsersAndGroupsCommand.cs
```diff
@@ -199,73 +195,52 @@ private async Task AddNewUsers(Organization organization,
...
+        var seatsAvailable = int.MaxValue;
+        var enoughSeatsAvailable = true;
+
+        if (organization.Seats.HasValue)
         {
-            userInvites.Add(new OrganizationUserInviteCommandModel(user.Email, user.ExternalId));
+            var occupiedSeats = await _organizationRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
+            seatsAvailable = organization.Seats.Value - occupiedSeats.Total;
+            enoughSeatsAvailable = seatsAvailable >= usersToAdd.Count;
         }
```
eliykat: This _does_ follow the current implementation, but if you look at it (specifically in Rider with intellisense), I don't think this block of code does anything. `enoughSeatsAvailable` is never actually used. And I think that's OK, because the OrganizationService invite code should handle autoscaling and subscription limits.

So I think you can delete this block that purports to check the seat limit.

test/Api.IntegrationTest/AdminConsole/Import/ImportOrganizationUsersAndGroupsCommandTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        var group = await OrganizationTestHelpers.CreateGroup(_factory, _organization.Id);
+
+        var request = new OrganizationImportRequestModel();
+        request.LargeImport = false;
+        request.OverwriteExisting = false;
+        request.Groups = [
+            new OrganizationImportRequestModel.OrganizationImportGroupRequestModel
+            {
+                Name = group.Name,
+                ExternalId = group.ExternalId,
```
eliykat: I think this needs to use a new value (e.g. a new Guid) like your other tests, not the group's current externalId, otherwise we're not updating anything.

test/Api.IntegrationTest/AdminConsole/Import/ImportOrganizationUsersAndGroupsCommandTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        };
+
+        var request = new OrganizationImportRequestModel();
+        request.LargeImport = false;
+        request.OverwriteExisting = false;
+        request.Groups = [
+            new OrganizationImportRequestModel.OrganizationImportGroupRequestModel
+            {
+                Name = existingGroup.Name,
+                ExternalId = existingGroup.ExternalId,
```
eliykat, line 282: Same here, should update the value, not use the current value.

test/Api.IntegrationTest/Helpers/OrganizationTestHelpers.cs
```diff
@@ -141,7 +141,7 @@ public static async Task<Group> CreateGroup(ApiApplicationFactory factory, Guid
         {
             OrganizationId = organizationId,
             Id = new Guid(),
-            ExternalId = new Guid().ToString(),
+            ExternalId = Guid.NewGuid().ToString(),
```
eliykat: Still using `new Guid()` for the `Id`.

test/Api.IntegrationTest/AdminConsole/Import/ImportOrganizationUsersAndGroupsCommandTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+        // Assert against the database values
+        var groupRepository = _factory.GetService<IGroupRepository>();
+        var existingGroup = await groupRepository.GetByIdAsync(group.Id);
+
+        Assert.NotNull(existingGroup);
+        Assert.Equal(existingGroup.Id, group.Id);
+        Assert.Equal(existingGroup.Name, group.Name);
+        Assert.Equal(existingGroup.ExternalId, group.ExternalId);
```
eliykat: This is unexpected to me.

1. you create the group with a guid as the externalId (in the test helper)
2. your request _should_ change the externalId to a different random guid
3. but this assertion effectively says that the externalId hasn't changed (it compares it to the original group object, not to the updated externalId value)

I think the test should be failing as it's currently written 🤔

test/Api.IntegrationTest/AdminConsole/Import/ImportOrganizationUsersAndGroupsCommandTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        var newGroupInDb = groups.Where(g => g.ExternalId == newGroup.ExternalId).FirstOrDefault();
+        Assert.NotNull(newGroupInDb);
+        Assert.Equal(newGroupInDb.Name, newGroup.Name);
+        Assert.Equal(newGroupInDb.ExternalId, newGroup.ExternalId);
+
+        var existingGroupInDb = groups.Where(g => g.ExternalId == existingGroup.ExternalId).FirstOrDefault();
+        Assert.NotNull(existingGroupInDb);
+        Assert.Equal(existingGroupInDb.Id, existingGroup.Id);
+        Assert.Equal(existingGroupInDb.Name, existingGroup.Name);
+        Assert.Equal(existingGroupInDb.ExternalId, existingGroup.ExternalId);
```
eliykat: Same here, isn't this asserting that the externalId hasn't changed?

test/Api.IntegrationTest/AdminConsole/Import/ImportOrganizationUsersAndGroupsCommandTests.cs
BTreston: We don't have a way to update the groups `ExternalId` with this implementation since `ExternalId` is the unique identifier in the `OrganizationGroupImportData.GroupsDict`. The test passes because the group in the request is being added to the repo, so there are actually 2 groups present.
`OrganizationImportGroupRequestModel` does not have an `Id` property either, it seems to always assume a new group is made.

eliykat: Oh I see - I was thinking of the user logic, where it matches existing users by email and then updates their externalId. But for groups, the externalId is the only unique reference, so there's nothing else to match on.

Updated suggestion: for the "existing group" tests, make the `externalId`s match, but have it update the name (and maybe members) instead. Just so we are testing update behavior.

eliykat (commented):
Commented in thread above.

eliykat (approved):

jrmccannon (approved):

PR #5825, opened 2025-05-15
Auth/pm 17111/add browser to list of approving clients

Patrick-Pimentel-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17111

## 📔 Objective

Added a feature flag around the list of approving clients.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Identity/Utilities/LoginApprovingClientTypes.cs
```diff
@@ -5,18 +5,28 @@ namespace Bit.Identity.Utilities;
 public static class LoginApprovingClientTypes
 {
     private static readonly IReadOnlyCollection<ClientType> _clientTypesThatCanApprove;
+    private static readonly IReadOnlyCollection<ClientType> _featureFlaggedClientTypesThatCanApprove;
 
     static LoginApprovingClientTypes()
     {
-        var clientTypes = new List<ClientType>
+        var featureFlaggedClientTypes = new List<ClientType>
```
trmartin4, line 12: Is the reason why we can't control the list of types here (and expose 2 different lists) because we don't have access to `IFeatureService`?

trmartin4 (changes requested):
Just requesting clarification on the question I had.

src/Identity/Utilities/LoginApprovingClientTypes.cs
Patrick-Pimentel-Bitwarden: Yeah it's a static class and we wouldn't be able to reference non-static services like the feature service.

Patrick-Pimentel-Bitwarden: That was my first thought and I tried to do the feature flagging there but didn't see a proper way how.

Patrick-Pimentel-Bitwarden: Help me understand what you don't like about this approach? I thought this approach made it very clear which list is used when. I can understand it's not incredibly DRY but it is a one off that will be cleaned up and doesn't have complicated branching logic.

Patrick-Pimentel-Bitwarden: Your thought is to make this static class not static and to do the feature flag evaluation inside the src/Identity/Utilities/LoginApprovingClientTypes.cs right?

trmartin4: Sure!  I think ideally if we could isolate the cleanup of the feature flag into one file, that would make less work later, and more straightforward cleanup.  An idea could be to leave the browser in the list (so that file doesn't have to change when we clean up the flag, or in this PR), and instead check the feature flag, and, if `false`, _remove_ the browser from the list of valid types in the `UserDecryptionOptionsBuilder`, after making a deep copy of the list locally.  Then, that local copy could be used in the LINQ query.

If you feel like that isn't a good approach, and would prefer to keep them separate lists, my only concern there would be around naming.  Calling it `FeatureFlaggedClientTypes` doesn't express to those who are looking at the code in the `LoginApprovingClientTypes` what the intent or usage of that list should be.  Another team could see that and want to use that list for something and not be clear about what "feature flag" we were referring to. Perhaps either naming it something like `ClientTypesWithoutBrowserThatCanApprove` with a comment explaining why we're making two lists would be clearer for others looking at the code.

Patrick-Pimentel-Bitwarden: We can absolutely go with the first solution!

I totally understand your concerns with the naming. I was a little quick with my solution because of the timebox I thought we were working in. I'll push up the changes for the first solution tonight!

Patrick-Pimentel-Bitwarden: I'm sorry though, on second thought if I do the deep copy and do all this logic right next to the utilization of the static list could it be a problem if another dev used the list then it would also be confusing if they needed to follow the same procedure for doing the clone next to the list and feature flagging it.

This static list of stuff is only used once for this purpose so I'm thinking I'll just make a static function that takes in a boolean for the flag value and that way we consolidate doing the ff logic for which list to send up in the LoginApprovingClientTypes.cs

I'll get this pushed up in a bit, walking to dinner.

trmartin4 (approved):
Excellent solution!  Thank you for iterating on it with me.

PR #5836, opened 2025-05-19
[PM 20621]Update error message when lowering seat count

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20621
## 📔 Objective

When attempting to lower an organization’s seat count below its “consumed” seat count, i.e. ,the number of members and the number of admin-iniated sponsorships, then the shown error message should be updated to reflect the number of sponsorships and members separately from each other.

“Your organization has [x] members and [y] sponsored families. To decrease the seat count below [x + y], you must remove members or sponsorships.”

## 📸 Screenshots

Screenshot 2025-05-19 at 16 06 26

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -342,11 +342,15 @@ private async Task<string> AdjustSeatsAsync(Organization organization, int seatA
 
         if (!organization.Seats.HasValue || organization.Seats.Value > newSeatTotal)
         {
-            var occupiedSeats = await _organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
-            if (occupiedSeats > newSeatTotal)
+            var totalConsumedSeats = await _organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
+            var organizationUsers = await _organizationUserRepository.GetManyByOrganizationAsync(organization.Id, null);
+            var organizationUserOccupiedSeats = organizationUsers.Where(user => user.Status >= 0).Count();
+            var sponsoredFamiliesOccupiedSeats = totalConsumedSeats - organizationUserOccupiedSeats;
```
eliykat, line 348: This is breaking the abstraction of `organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync`. That query is intended to abstract away who/what occupies a seat, but now we're repeating that logic here to figure out how many family sponsorships the org has.

If we need these figures separately, I suggest that the `GetOccupiedSeatCountByOrganizationIdAsync` call could return an object containing: user occupied seat count, sponsored families occupied seat count, and total occupied seat count. This could still be returned by the 1 sproc.

For example, this code would then look like this:

```cs
var consumedSeats = organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
if (consumedSeats.Total > newSeatTotal)
{
  throw new BadRequestException($"Your organization has {consumedSeats.Users} members and {consumedSeats.Sponsored} sponsored families. " +
  $"To decrease the seat count below {consumedSeats.Total}, you must remove members or sponsorships.");
}
```

I also think we should keep consistent language with occupied vs. consumed - unless this is a conscious decision by Billing Team to start using consumed instead, which is fine.

eliykat (changes requested):
There are similar error messages that also need to be updated in UpgradeOrganizationPlanCommand.

@jrmccannon it seems like there should be validation for this in your invite user rewrite that also needs to be updated, but I can't find it. Can you please help?

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadOccupiedSeatCountByOrganizationId.sql
```diff
@@ -33,5 +33,33 @@ BEGIN
...
+                (ToDelete = 1 AND ValidUntil IS NOT NULL AND ValidUntil > GETUTCDATE())
+            )
+            AND (
+                -- SENT status: When SponsoredOrganizationId is null
+                SponsoredOrganizationId IS NULL
+                OR
+                -- ACCEPTED status: When SponsoredOrganizationId is not null and ValidUntil is null or in the future
+                (SponsoredOrganizationId IS NOT NULL AND (ValidUntil IS NULL OR ValidUntil > GETUTCDATE()))
+            )
+        ) as Total
```
eliykat, line 64: Do we still need this given that `OrganizationSeatCounts` has its own getter which calculates the total? I think we want that getter OR this calculation in the sproc, not both.

I suggest the getter on the model is the simpler solution.

---

However, if you did want to calculate it in the sproc, this is how to do it without repeating code:

e.g.

```sql
SELECT Subquery.Users, Subquery.Sponsorships, Subquery.Users + Subquery.Sponsorships as Total
FROM (
  SELECT ... as Users,
  ... as Sponsored
) Subquery
```

By putting it in a sub-query you can reference the results in the outer query. [SO reference](https://stackoverflow.com/a/62825452)

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadOccupiedSeatCountByOrganizationId.sql
eliykat, line 1: Unfortunately because you're changing the interface of the sproc (it now returns 3 results instead of 1), you have to make a new sproc instead. This is because if we have to rollback the release, the old code will be expecting 1 result but it'll be getting 3 results from the updated sproc (which isn't rolled back). So we have to preserve backwards compatibility for 1 release.

This is easily fixed by making a new sproc with the new logic and updating the repository code to use it. Then in the following release, you can drop the old sproc. 

This is part of [EDD](https://contributing.bitwarden.com/contributing/database-migrations/edd).

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -88,16 +88,16 @@ public async Task<int> GetCountByOrganizationAsync(Guid organizationId, string e
         }
     }
 
-    public async Task<int> GetOccupiedSeatCountByOrganizationIdAsync(Guid organizationId)
+    public async Task<OrganizationSeatCounts> GetOccupiedSeatCountByOrganizationIdAsync(Guid organizationId)
     {
         using (var connection = new SqlConnection(ConnectionString))
         {
-            var result = await connection.ExecuteScalarAsync<int>(
+            var result = await connection.QueryFirstOrDefaultAsync<OrganizationSeatCounts>(
```
eliykat, line 95: This should always return a single object, I think `QueryAsync` is what you want here.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -227,10 +227,28 @@ public async Task<int> GetCountByOrganizationAsync(Guid organizationId, string e
...
+            var users = await dbContext.OrganizationUsers
+                .Where(ou => ou.OrganizationId == organizationId && ou.Status >= 0)
+                .CountAsync();
+
+            var sponsored = await dbContext.OrganizationSponsorships
+                .Where(os => os.SponsoringOrganizationId == organizationId &&
+                    os.IsAdminInitiated &&
+                    (os.ToDelete == false || (os.ToDelete == true && os.ValidUntil != null && os.ValidUntil > DateTime.UtcNow)) &&
+                    (os.SponsoredOrganizationId == null || (os.SponsoredOrganizationId != null && (os.ValidUntil == null || os.ValidUntil > DateTime.UtcNow))))
+                .CountAsync();
```
eliykat, line 244: I think this is a much better approach than the current `OrganizationUserReadOccupiedSeatCountByOrganizationIdQuery`. But does this mean we can delete that query?

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadOccupiedSeatCountByOrganizationId.sql
eliykat: In fact, you could probably take this opportunity to create a new repository method in `OrganizationRepository` given that this is no longer about OrganizationUsers.

eliykat (changes requested):
I can see that this was a bit of refactoring, but I think overall it's a better approach now that sponsorships take up seats as well. Nice work!

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
cyprain-okeke: removed

src/Core/AdminConsole/Repositories/IOrganizationRepository.cs
```diff
@@ -25,4 +26,14 @@ public interface IOrganizationRepository : IRepository<Organization, Guid>
...
+
+    /// <summary>
+    /// Returns the number of occupied seats for an organization.
+    /// Occupied seats are OrganizationUsers that have at least been invited.
+    /// As of https://bitwarden.atlassian.net/browse/PM-17772, a seat is also occupied by a Families for Enterprise sponsorship sent by an
+    /// organization admin, even if the user sent the invitation doesn't have a corresponding OrganizationUser in the Enterprise organization.
+    /// </summary>
+    /// <param name="organizationId">The ID of the organization to get the occupied seat count for.</param>
+    /// <returns>The number of occupied seats for the organization.</returns>
+    Task<OrganizationSeatCounts> GetOccupiedSeatCountByOrganizationIdAsync(Guid organizationId);
```
eliykat, line 38: Great xmldoc!

nit: "OrganizationUsers that have at least been invited." could be clearer. It relies on the reader knowing how the `OrganizationUserStatusType` enum is ordered (e.g. that revoked is less than invited).

I suggest: "OrganizationUsers occupy a seat, unless they are revoked."

src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
```diff
@@ -27,7 +27,7 @@ public interface IOrganizationUserRepository : IRepository<OrganizationUser, Gui
     /// </summary>
     /// <param name="organizationId">The ID of the organization to get the occupied seat count for.</param>
     /// <returns>The number of occupied seats for the organization.</returns>
-    Task<OrganizationSeatCounts> GetOccupiedSeatCountByOrganizationIdAsync(Guid organizationId);
+    Task<int> GetOccupiedSeatCountByOrganizationIdAsync(Guid organizationId);
```
eliykat, line 30: You can delete this method from the code now, it's only the MSSQL sproc that needs to be kept in place for an extra release.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -199,12 +200,27 @@ public async Task<ICollection<Organization>> GetAddableToProviderByUserIdAsync(
...
+                commandType: CommandType.StoredProcedure);
 
-        return (await connection.QueryAsync<Organization>(
-            $"[{Schema}].[{Table}_ReadManyByIds]",
-            new { OrganizationIds = ids.ToGuidIdArrayTVP() },
-            commandType: CommandType.StoredProcedure))
-            .ToList();
+            return results.ToList();
+        }
+    }
```
eliykat, line 212: Unrelated changes to this method - please revert.

src/Core/AdminConsole/Repositories/IOrganizationRepository.cs
eliykat, line 1: Please add integration tests for this new method in `Infrastructure.IntegrationTest/AdminConsole/Repositories`. There are some examples in there to follow, but the basic idea is to set up a test organization with some users (in different states) and some sponsorships, and then check that the sproc calculates the occupied seats correctly.

eliykat (changes requested):

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
cyprain-okeke: revert

src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
cyprain-okeke: removed

cturnbull-bitwarden (commented):
Looks good, though it looks like you've got a merge conflict

eliykat (dismissed):
Thanks @cyprain-okeke !

util/Migrator/DbScripts/2025-05-20_00_UpdateOrgReadOccupiedSeatCountForSponsorships.sql
```diff
@@ -0,0 +1,44 @@
+IF OBJECT_ID('[dbo].[Organization_ReadOccupiedSeatCountByOrganizationId]') IS NOT NULL
+BEGIN
+    DROP PROCEDURE [dbo].[Organization_ReadOccupiedSeatCountByOrganizationId]
```
rkac-bw: no need to do object look up, can just use 
`CREATE OR ALTER PROCEDURE` syntax

rkac-bw (dismissed):
lgtm

JaredSnider-Bitwarden (dismissed):

eliykat (dismissed):

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -294,11 +294,12 @@ private async Task<string> AdjustSeatsAsync(Organization organization, int seatA
...
         {
-            var occupiedSeats = await _organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
-            if (occupiedSeats > newSeatTotal)
+            var seatCounts = await _organizationRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
+
+            if (seatCounts.Total > newSeatTotal)
             {
-                throw new BadRequestException($"Your organization currently has {occupiedSeats} seats filled. " +
-                    $"Your new plan only has ({newSeatTotal}) seats. Remove some users.");
+                throw new BadRequestException($"Your organization has {seatCounts.Users} members and {seatCounts.Sponsored} sponsored families. " +
```
connerbw: This new message shouldn't replace the old in every circumstance. It should look something like this:

```
if (organization.UseAdminSponsoredFamilies || seatCounts.Sponsored > 0)
{
    show new error message
}
else
{
    show old error message
}
```

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpgradeOrganizationPlanCommand.cs
```diff
@@ -107,12 +107,12 @@ public async Task<Tuple<bool, string>> UpgradePlanAsync(Guid organizationId, Org
...
-            var occupiedSeats =
-                await _organizationUserRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
-            if (occupiedSeats > updatedPasswordManagerSeats)
+            var seatCounts =
+                await _organizationRepository.GetOccupiedSeatCountByOrganizationIdAsync(organization.Id);
+            if (seatCounts.Total > updatedPasswordManagerSeats)
             {
-                throw new BadRequestException($"Your organization currently has {occupiedSeats} seats filled. " +
-                                              $"Your new plan only has ({updatedPasswordManagerSeats}) seats. Remove some users.");
+                throw new BadRequestException($"Your organization has {seatCounts.Users} members and {seatCounts.Sponsored} sponsored families. " +
```
connerbw: Same here, this new message shouldn't replace the old in every circumstance

rkac-bw (dismissed):
lgtm

cturnbull-bitwarden (dismissed):

eliykat (dismissed):

JaredSnider-Bitwarden (approved):

JaredSnider-Bitwarden (approved):

cturnbull-bitwarden (approved):

rkac-bw (approved):
lgtm

PR #5839, opened 2025-05-20
PM-20574 & PM-20575 Adding Risk Insight Report tables, repositories, and migrations

prograhamming:
## 🎟️ Tracking

(https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?assignee=712020%3Aa479bb6c-74c0-4e0d-bd65-bc64b95c604a&selectedIssue=PM-20574)

https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?assignee=712020%3Aa479bb6c-74c0-4e0d-bd65-bc64b95c604a&selectedIssue=PM-20575

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Sql/Dirt/dbo/Tables/RiskInsightReport.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE [dbo].[RiskInsightReport] (
```
withinfocus, line 1: ⚠️ I said this in the beginning of this very large project so it may have been lost at this point, but "Risk Insights" is a product name and really shouldn't be in structured code. The original database design documents and suggestions explicitly kept it out too (and it changed names a couple times). These are generally-speaking "security task reports", "task reports", or just "reports".

src/Sql/Dirt/dbo/Tables/RiskInsightCriticalApplication.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE [dbo].[RiskInsightCriticalApplication] (
```
withinfocus, line 1: ⚠️ Same thread as report naming.

withinfocus (changes requested):
Bit of a 💣  I realize but hopefully easy to refactor.

src/Core/Dirt/Reports/Entities/RiskInsightCriticalApplication.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Bit.Core.Dirt.Reports.Entities;
+
+public class RiskInsightCriticalApplication : ITableObject<Guid>, IRevisable
+{
+    public Guid Id { get; set; }
+    public Guid OrganizationId { get; set; }
+    public string Applications { get; set; } = string.Empty;
+    public DateTime CreationDate { get; set; } = DateTime.UtcNow;
+    public DateTime RevisionDate { get; set; } = DateTime.UtcNow;
+
```
voommen-livefront, line 15: @prograhamming - I think we need a Date here. The reason is this: Every time a report is changed, we add a row to this table. However, when we retrieve data, we only need the latest record - which should be determined by a `Date` column (just like you have in the `RiskInsightReport` table.

src/Core/Dirt/Reports/Repositories/IRiskInsightCriticalApplicationRepository.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Bit.Core.Dirt.Reports.Entities;
+using Bit.Core.Repositories;
+
+namespace Bit.Core.Dirt.Reports.Repositories;
+
+public interface IRiskInsightCriticalApplicationRepository : IRepository<RiskInsightCriticalApplication, Guid>
+{
+    Task<ICollection<RiskInsightCriticalApplication>> GetByOrganizationIdAsync(Guid organizationId);
+}
```
voommen-livefront, line 9: In my opinion (need to double check with Tom) - when we query these tables, we need the latest by organization. So preferably, we need a method called `GetLatestByOrganizationIdAsync` - and - it should return a single record and not a collection.

src/Core/Dirt/Reports/Repositories/IRiskInsightReportRepository.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Bit.Core.Dirt.Reports.Entities;
+using Bit.Core.Repositories;
+
+namespace Bit.Core.Dirt.Reports.Repositories;
+
+public interface IRiskInsightReportRepository : IRepository<RiskInsightReport, Guid>
+{
+    Task<ICollection<RiskInsightReport>> GetByOrganizationIdAsync(Guid organizationId);
```
voommen-livefront, line 8: Here too, we need a method that `GetLatestByOrganizationIdAsync` and it returns the record with the latest `Date`.

src/Sql/Dirt/dbo/Tables/RiskInsightCriticalApplication.sql
```diff
@@ -0,0 +1,12 @@
...
+    [Id]                       UNIQUEIDENTIFIER NOT NULL,
+    [OrganizationId]           UNIQUEIDENTIFIER NOT NULL,
+    [Applications]             NVARCHAR(MAX)    NOT NULL,
+    [CreationDate]             DATETIME2 (7)    NOT NULL,
+    [RevisionDate]             DATETIME2 (7)    NOT NULL,
+    CONSTRAINT [PK_RiskInsightCriticalApplication] PRIMARY KEY CLUSTERED ([Id] ASC),
+    CONSTRAINT [FK_RiskInsightCriticalApplication_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id])
```
voommen-livefront, line 8:
```suggestion
    CONSTRAINT [FK_RiskInsightCriticalApplication_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]) ON DELETE CASCADE
```
This is necessary because, there is an option to delete an `Organization` and, when an org is deleted, it should delete these records as well.

src/Sql/Dirt/dbo/Tables/RiskInsightReport.sql
```diff
@@ -0,0 +1,13 @@
...
+    [Id]                       UNIQUEIDENTIFIER NOT NULL,
+    [OrganizationId]           UNIQUEIDENTIFIER NOT NULL,
+    [Date]                     DATETIME2 (7)    NOT NULL,
+    [ReportData]               NVARCHAR(MAX)    NOT NULL,
+    [CreationDate]             DATETIME2 (7)    NOT NULL,
+    [RevisionDate]             DATETIME2 (7)    NOT NULL,
+    CONSTRAINT [PK_RiskInsightReport] PRIMARY KEY CLUSTERED ([Id] ASC),
+    CONSTRAINT [FK_RiskInsightReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id])
```
voommen-livefront, line 9:
```suggestion
    CONSTRAINT [FK_RiskInsightReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]) ON DELETE CASCADE 
```
This is needed because when an organization is deleted, it must delete records in this table as well.

util/Migrator/DbScripts/2025-05-16-00_RiskInsightReport.sql
```diff
@@ -0,0 +1,111 @@
...
+    CREATE TABLE [dbo].[RiskInsightReport]
+    (
+    [Id]                       UNIQUEIDENTIFIER NOT NULL,
+    [OrganizationId]           UNIQUEIDENTIFIER NOT NULL,
+    [Date]                     DATETIME2 (7)    NOT NULL,
+    [ReportData]               NVARCHAR(MAX)    NOT NULL,
+    [CreationDate]             DATETIME2 (7)    NOT NULL,
+    [RevisionDate]             DATETIME2 (7)    NOT NULL,
+    CONSTRAINT [PK_RiskInsightReport] PRIMARY KEY CLUSTERED ([Id] ASC),
+    CONSTRAINT [FK_RiskInsightReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id])
```
voommen-livefront, line 12:
```suggestion
    CONSTRAINT [FK_RiskInsightReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]) ON DELETE CASCADE
```
just repeating what was mentioned earlier

util/MySqlMigrations/Migrations/20250520034740_2025-05-16-00_RiskInsightReport.sql.cs
voommen-livefront, line 1: We may not need this file

util/MySqlMigrations/Migrations/20250520034928_2025-05-16-01_RiskInsightCriticalApplication.sql.cs
voommen-livefront, line 1: We may not need this migration as it is empty.

util/PostgresMigrations/Migrations/20250520034935_2025-05-16-01_RiskInsightCriticalApplication.sql.cs
voommen-livefront, line 1: Empty migration, we can safely remove this file

util/SqliteMigrations/Migrations/20250520034931_2025-05-16-01_RiskInsightCriticalApplication.sql.cs
voommen-livefront, line 1: Empty migration, can be safely removed.

voommen-livefront (changes requested):
Just minor changes - if you don't think we need an index with `OrganizationId` + `Date` - no worries, we can request them when we need it.
We are missing MySql Migrations - at this time, I only spotted empty migration files.

src/Sql/Dirt/dbo/Tables/RiskInsightCriticalApplication.sql
withinfocus: Guidance here is for direct children of `Organization` to use our deletion proc for SQL Server e.g. https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Stored%20Procedures/Organization_DeleteById.sql#L87 -- this results in less computational pressure on the DB when an organization is actually deleted. Change that proc and make a new one that's called there that deletes by organization ID.

For EF however, ensure that the entity type configuration does indicate this; we let self-hosters utilize the cascaded delete as their environments will not be under the pressures ours has.

djsmith85 (approved):
Approving for Tools as we are only tagged for the file moves regarding the PasswordHealthReportApp

src/Sql/Dirt/dbo/Tables/RiskInsightReport.sql
prograhamming: Updated name

src/Sql/Dirt/dbo/Tables/RiskInsightCriticalApplication.sql
prograhamming: Updated name

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_ReadyById.sql
```diff
@@ -0,0 +1,16 @@
+CREATE PROCEDURE [dbo].[OrganizationApplication_ReadById]
+    @Id UNIQUEIDENTIFIER
```
rkac-bw: SQL File name does not match stored procedure name

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadById.Sql
```diff
@@ -0,0 +1,17 @@
+CREATE PROCEDURE [dbo].[OrganizationReport_ReadById]
+    @Id UNIQUEIDENTIFIER
```
rkac-bw: :pick: lower case sql extension for conistency

util/Migrator/DbScripts/2025-06-10-00_OrganizationReport.sql
```diff
@@ -0,0 +1,111 @@
...
+GO
+
+IF OBJECT_ID('dbo.OrganizationReportView') IS NOT NULL
+BEGIN
+    DROP VIEW [dbo].[OrganizationReportView]
+END
+GO
+
+CREATE VIEW [dbo].[OrganizationReportView] AS
+    SELECT * FROM [dbo].[OrganizationReport];
```
rkac-bw: Yu can use "CREATE OR ALTER VIEW" instead of looking up object, supported since SQL 2016

prograhamming: updated sql

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadById.Sql
prograhamming: fixed file extension.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_ReadyById.sql
prograhamming: fixed file name

src/Sql/Dirt/dbo/Tables/RiskInsightCriticalApplication.sql
prograhamming: Changed the proc to add these tables and changed the entity type configuration to indicate the cascaded delete.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_DeleteById.sql
```diff
@@ -0,0 +1,10 @@
+CREATE PROCEDURE [dbo].[OrganizationApplication_DeleteById]
+    @Id UNIQUEIDENTIFIER
+AS
+    SET NOCOUNT ON;
+
+    IF @Id IS NULL
+       THROW 50000, 'Id cannot be null', 1;
```
withinfocus: ⚠️ I haven't seen this elsewhere and it is effectively impossible, plus we don't handle these error codes do we? Can just get rid of it -- we have issues if null GUIDs come into this.

withinfocus: Get rid of all these.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_ReadById.sql
```diff
@@ -0,0 +1,16 @@
...
+    IF @Id IS NULL
+       THROW 50000, 'Id cannot be null', 1;
+
+    SELECT
+        [Id],
+        [OrganizationId],
+        [Applications],
+        [CreationDate],
+        [RevisionDate]
+    FROM [dbo].[OrganizationApplication]
```
withinfocus: ❌ Create a view and select from that exclusively for any read operations.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_Create.sql
```diff
@@ -0,0 +1,12 @@
...
+    @Id UNIQUEIDENTIFIER OUTPUT,
+    @OrganizationId UNIQUEIDENTIFIER,
+    @Date DATETIME2(7),
+    @ReportData NVARCHAR(MAX),
+    @CreationDate DATETIME2(7),
+    @RevisionDate DATETIME2(7)
+AS
+    SET NOCOUNT ON;
+
+    INSERT INTO [dbo].[OrganizationReport]( [Id],[OrganizationId],[Date],[ReportData],[CreationDate],[RevisionDate] )
```
withinfocus: ⚠️ This and a few other files have compressed formatting -- did VS Code or similar do this? I Personally run `prettier` on save but that's me, but in any case we want to match the style of everything else (which it will do).

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
```diff
@@ -0,0 +1,19 @@
...
+    );
+GO
+
+CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId]
+    ON [dbo].[OrganizationReport]([OrganizationId] ASC);
+GO
+
+CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_Date]
+    ON [dbo].[OrganizationReport]([OrganizationId] ASC, [Date] DESC);
+GO
```
withinfocus: ⛏️ All files should end in newlines.

withinfocus: @rkac-bw is it advisable to have an composite index on date like this? Not sure about the use case.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE [dbo].[OrganizationReport] (
+    [Id]                       UNIQUEIDENTIFIER NOT NULL,
+    [OrganizationId]           UNIQUEIDENTIFIER NOT NULL,
+    [Date]                     DATETIME2 (7)    NOT NULL,
+    [ReportData]               NVARCHAR(MAX)    NOT NULL,
+    [CreationDate]             DATETIME2 (7)    NOT NULL,
+    [RevisionDate]             DATETIME2 (7)    NOT NULL,
```
withinfocus: ❓ Are these ever revised? I see the update proc but is that gonna happen?

src/Sql/dbo/Dirt/Views/OrganizationApplicationView.sql
```diff
@@ -0,0 +1,5 @@
+CREATE VIEW [dbo].[OrganizationApplicationView]
```
withinfocus, line 1: ❌ These aren't used, per my other comment.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
```diff
@@ -137,6 +137,20 @@ BEGIN
...
+    WHERE
+        [Id] = @Id
+
+    -- Delete Organization Report
+    DELETE
+    FROM
+        [dbo].[OrganizationReport]
+    WHERE
+        [Id] = @Id
+
```
withinfocus: ❌ This isn't correct and you have to delete by organization ID.

util/Migrator/DbScripts/2025-06-13-00_OrganizationReport.sql
```diff
@@ -0,0 +1,110 @@
+IF OBJECT_ID('dbo.OrganizationReport') IS NULL
```
withinfocus: ⚠️ I suggest having one migration for this but as long as you're sure this will execute in the correct order then it's alright.

withinfocus (changes requested):

rkac-bw (approved):
LGTM

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
rkac-bw: This would be advisable if there was a query that searched reports by organizationid and wanted to get the most recent one, not seeing any query like that here
If there was and there was a cardinality of multiple reports for an org for a date value then it would be beneficial

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_DeleteById.sql
zeal-vijayoommen: I think I know the origins of this. I will create a task to fix this. It occurs in one other place only.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_ReadById.sql
zeal-vijayoommen: We had a view, but we didn't use it. Updated this to use the view in both places (`OrganizationApplication` and `OrganizationReport`)

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_DeleteById.sql
voommen-livefront: I have removed this from stored procedures and from the .sql file. 
You are right - I did look up other procedures and they don't have that error condition. 
I had done something similar in the procedures for `PasswordHealthReportApplication` tables - and I'll create a task to remove it there as well.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_Create.sql
voommen-livefront: Thanks for catching this - I just got Sql Formatter installed and formatted all .sql files in this PR.

src/Sql/dbo/Dirt/Views/OrganizationApplicationView.sql
voommen-livefront: I changed the `_ReadyById` and `_ReadByOrganizationId` to use this view.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
voommen-livefront: WOW! Thank you! I can't believe I missed this all this time!

util/Migrator/DbScripts/2025-06-13-00_OrganizationReport.sql
voommen-livefront: If it's ok - can I keep it as two migrations please? Its just that the procedures and tables get mixed up when we read it in one file.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
voommen-livefront: You are right - these records on the server are encrypted and never edited. I kept the two date columns to comply with the standards in other tables.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
voommen-livefront: While querying the `OrganizationReports` table, almost always, we need the latest record - as that's the data that will be rendered on the UI.
I created the index here, rather than another PR.

voommen-livefront: I updated the files with the formatter and then checked the queries. I should be good I think 🤞

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadById.sql
```diff
@@ -3,15 +3,12 @@ CREATE PROCEDURE [dbo].[OrganizationReport_ReadById]
...
-FROM [dbo].[OrganizationReport]
-WHERE [Id] = @Id;
+    SELECT
+        [Id],
+        [OrganizationId],
+        [Date],
+        [ReportData],
+        [CreationDate],
+        [RevisionDate]
+    FROM [dbo].[OrganizationReport]
```
withinfocus: ❌ Need to use the view, and you just `SELECT *` from it.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
withinfocus: What I meant by this comment is that you then don't need a revision date column nor an update proc at all -- it will never be used.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadById.sql
voommen-livefront: You are right! I should've done this earlier. Sorry about that. Thank you for catching this.

src/Core/Dirt/Entities/OrganizationApplication.cs
```diff
@@ -0,0 +1,20 @@
+﻿#nullable enable
+
+using Bit.Core.Entities;
+using Bit.Core.Utilities;
+
+namespace Bit.Core.Dirt.Entities;
+
+public class OrganizationApplication : ITableObject<Guid>, IRevisable
```
withinfocus, line 8: I had to look up what `IRevisable` even was -- it can stay I guess but it really never caught on here.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_Update.sql
```diff
@@ -0,0 +1,14 @@
...
+    @Applications NVARCHAR(MAX),
+    @RevisionDate DATETIME2(7)
+AS
+    SET NOCOUNT ON;
+    
+    UPDATE [dbo].[OrganizationApplication]
+    SET
+        [OrganizationId] = @OrganizationId,
+        [Applications] = @Applications,
+        [RevisionDate] = @RevisionDate
```
withinfocus, line 15: ❌ While the value shouldn't change, we still support all columns being updateable and therefore `CreationDate` should still be here. The idea is that the proc can be used regardless of the higher-level logic. This isn't necessarily practiced or is my requested change practical, but it is the norm.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
```diff
@@ -0,0 +1,18 @@
...
+    );
+GO
+
+CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId]
+    ON [dbo].[OrganizationReport]([OrganizationId] ASC);
+GO
+
+CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_Date]
+    ON [dbo].[OrganizationReport]([OrganizationId] ASC, [Date] DESC);
+GO
```
withinfocus: ⛏️ End this file with a newline.

util/Migrator/DbScripts/2025-06-13-01_OrganizationApplication.sql
```diff
@@ -0,0 +1,105 @@
...
+
+
+
+
+
+
+
+
+
+
```
withinfocus: ⛏️ Clean this up.

withinfocus (changes requested):
Just a few minor things. Almost there.

src/Core/Dirt/Entities/OrganizationApplication.cs
voommen-livefront: I don't fully understand IRevisable - I put it there since most other entities have it.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationApplication_Update.sql
voommen-livefront: This makes sense. I have added the `CreationDate` back to the update proc

src/Core/Dirt/Entities/OrganizationApplication.cs
withinfocus: There's only [a few](https://github.com/search?q=repo%3Abitwarden/server%20IRevisable&type=code) actually.

util/Migrator/DbScripts/2025-06-13-01_OrganizationApplication.sql
voommen-livefront: Sometimes we miss the obvious! I can't believe I had this many spaces!!

util/Migrator/DbScripts/2025-06-13-01_OrganizationApplication.sql
```diff
@@ -0,0 +1,94 @@
...
+        *
+    FROM [dbo].[OrganizationApplicationView]
+    WHERE [Id] = @Id;
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationApplication_Update]
+    @Id UNIQUEIDENTIFIER OUTPUT,
+    @OrganizationId UNIQUEIDENTIFIER,
+    @Applications NVARCHAR(MAX),
+    @RevisionDate DATETIME2(7)
```
withinfocus, line 76: ❌ You had some background updates while I was reviewing -- creation date needs to be here too.

withinfocus (changes requested):

util/Migrator/DbScripts/2025-06-13-01_OrganizationApplication.sql
withinfocus: Woops, happened again and here it is.

withinfocus (dismissed):

withinfocus (approved):

voommen-livefront (approved):

PR #5840, opened 2025-05-20
[PM-18238] Add RequireTwoFactorPolicyRequirement

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18238

## 📔 Objective

Add a new `RequireTwoFactorPolicyRequirement` to be used for checking if a user complies with the 2FA policy on various services (`UserService`, `RestoreOrganizationUserCommand`, `ConfirmOrganizationUserCommand`, and `AcceptOrgUserCommand`)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -1782,72 +1775,6 @@ public async Task<List<Tuple<OrganizationUser, string>>> RevokeUsersAsync(Guid o
         return result;
     }
 
-    private async Task CheckPoliciesBeforeRestoreAsync(OrganizationUser orgUser, bool userHasTwoFactorEnabled)
```
r-tome, line 1785: Unused method, leftover when lifting and shifting into command

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/RestoreUser/RestoreOrganizationUserCommandTests.cs
r-tome, line 1: vNext tests were redundant

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireTwoFactorPolicyRequirement.cs
```diff
@@ -0,0 +1,81 @@
...
+        _policyDetails = policyDetails;
+    }
+
+    /// <summary>
+    /// Determines if the user can accept an invitation to an organization.
+    /// </summary>
+    /// <param name="twoFactorEnabled">Whether the user has two-step login enabled.</param>
+    /// <param name="organizationId">The ID of the organization.</param>
+    /// <returns>True if the user can accept the invitation, false otherwise.</returns>
+    public bool CanAcceptInvitation(bool twoFactorEnabled, Guid organizationId) =>
```
BTreston, line 24: 💭 These seem like they would fit better in the factory method below that initialize instance fields in the requirement object here. However, the current interface doesn't allow us to pass in additional data to generate the policy requirement apart from just `PolicyDetails` so I see why it is done this way. I wonder if its worth exploring expanding the factory interface to accommodate this kind of use case. The way it is now, the factory really just ends up being a wrapper for this class which seems counter intuitive imo.

BTreston (approved):
Looks good, just some non-blocking thoughts that might be worth investigating

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireTwoFactorPolicyRequirement.cs
eliykat: This is a good thought. I also don't love the `bool twoFactorEnabled` interface.

I can think of 2 alternatives:
1. you can inject whatever you want into the factory @BTreston, you're not limited to only the `PolicyDetails` that are passed in. (e.g. see the [SSO requirement](https://github.com/bitwarden/server/blob/198d96e1555d9c899db36728b37cbbcfc75f99f3/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireSsoPolicyRequirement.cs#L36-L40).) So you could inject the 2FA query and handle that inside the factory, rather than relying on the caller to tell you. However, the factory code is not currently async (no `Task`s) so that would require some refactoring to move to an async interface. I'm also not sure whether it's making the requirement responsible for too much, which leads to the second option.

2. limit how much this interface is doing: instead of `CanAcceptInvitation` make it more like `TwoFactorRequiredToAcceptInvitation`, then the caller decides whether to check 2FA based on the result. That wording is pretty bad. But I think this can be simplified further: `TwoFactorRequired(Guid organizationId) => _policyDetails.Any(p => p.OrganizationId == organizationId)` would work for all cases here. If you're confirming a user, restoring a user, or accepting an invite, what matters is the organization you're doing it for, not your status (which presumably the caller will check as part of the larger confirm/revoke flow etc). `TwoFactorPoliciesForActiveMemberships` is really the only one that cares about the user status.

This is the last thing on a Friday so please prove me wrong 😁

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireTwoFactorPolicyRequirement.cs
```diff
@@ -0,0 +1,81 @@
...
+                OrganizationUserStatusType.Confirmed));
+
+    /// <summary>
+    /// Gets the two-factor policies for active memberships.
+    /// </summary>
+    /// <returns>The two-factor policies for active memberships.</returns>
+    public IEnumerable<PolicyDetails> TwoFactorPoliciesForActiveMemberships =>
+        _policyDetails.Where(p => p.OrganizationUserStatus is
+                OrganizationUserStatusType.Accepted or
+                OrganizationUserStatusType.Confirmed);
```
eliykat, line 69: I think this could be more explicit. What does the policy _require_ - what is this interface used for?

For example:

```c#
    /// <summary>
    /// A list of organization IDs that require that the user have 2FA enabled. The user should be revoked from these organizations if they disable all 2FA methods.
    /// </summary>
    public IEnumerable<Guid> OrganizationsRequiringTwoFactor => 
```

r-tome: I had to also return `OrganizationUserId` because that is used in `UserService` to revoke the user so I changed the return type to a Tuple to simplify it.
I also updated the xmldoc like you suggested.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireTwoFactorPolicyRequirement.cs
r-tome: I decided to go with the second option. My first approach was to offload any policy logic from the calling places but that made it so that `RequireTwoFactorPolicyRequirement` was doing too much. 

Thank you both for the great thoughts!

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptOrgUserCommandTests.cs
```diff
@@ -167,7 +169,7 @@ public async Task AcceptOrgUserAsync_UserWithout2FAJoining2FARequiredOrg_ThrowsB
         SetupCommonAcceptOrgUserMocks(sutProvider, user, org, orgUser, adminUserDetails);
 
         // User doesn't have 2FA enabled
-        _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(user).Returns(false);
```
eliykat, line 170: Is it possible to delete this class property altogether? I can't see that it does anything because it's not actually injected into the sut.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireTwoFactorPolicyRequirement.cs
eliykat: lgtm!

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
```diff
@@ -168,6 +183,32 @@ private async Task CheckPoliciesAsync(Guid organizationId, User user,
...
+    /// <param name="userTwoFactorEnabled">Whether the user has two-step login enabled.</param>
+    /// <exception cref="BadRequestException">Thrown if the policy applies to the organization and 
+    /// the user does not have two-step login enabled.</exception>
+    private async Task ValidateTwoFactorAuthenticationPolicyAsync(User user, Guid organizationId, bool userTwoFactorEnabled)
+    {
+        if (userTwoFactorEnabled)
+        {
+            // If the user has two-step login enabled, we skip checking the 2FA policy
+            return;
+        }
```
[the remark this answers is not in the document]
r-tome: I agree its better to be consistent, therefore I also changed `AcceptOrgUserCommand` to check 2FA first.

> there's nothing stopping you from moving the 2FA query call down into these methods

The 2FA check is done in bulk that's why I kept it at the top.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/RequireTwoFactorPolicyRequirement.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /// <summary>
+    /// Checks if two-factor authentication is required for the organization due to an active policy.
+    /// </summary>
+    /// <param name="organizationId">The ID of the organization to check.</param>
+    /// <returns>True if two-factor authentication is required for the organization, false otherwise.</returns>
+    /// <remarks>
+    /// This does not check the user's membership status.
+    /// </remarks>
```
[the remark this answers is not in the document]
r-tome: I went with your first suggestion. This property is about if the policy applies to the organization, their user's membership status doesn't matter at this point.

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptOrgUserCommandTests.cs
r-tome: It is used to check the user's 2FA. I removed redundant setups as this is already done in `SetupCommonAcceptOrgUserMocks`

eliykat (commented):
Other than @BTreston 's suggested change, looks good to me!

src/Core/Services/Implementations/UserService.cs
```diff
@@ -1394,9 +1399,29 @@ public void SetTwoFactorProvider(User user, TwoFactorProviderType type, bool set
 
     private async Task CheckPoliciesOnTwoFactorRemovalAsync(User user)
     {
+        if (_featureService.IsEnabled(FeatureFlagKeys.PolicyRequirements))
+        {
+            var requirement = await _policyRequirementQuery.GetAsync<RequireTwoFactorPolicyRequirement>(user.Id);
+
+            var revokeOrgUserTasks = requirement.OrganizationsRequiringTwoFactor.Select(async o =>
+            {
+                var organization = await _organizationRepository.GetByIdAsync(o.OrganizationId);
```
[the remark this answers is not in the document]
r-tome: Great idea!

BTreston (approved):
Looks good 👍

eliykat (approved):

PR #5855, opened 2025-05-22
Document database projects and complete EDD support

withinfocus:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C024Z7LQNLB/p1747902781176209

## 📔 Objective

Adds some light documentation within the various SQL projects to explain their purpose and how they inter-relate, with links out to usage.

Completes what's needed for [online environments to use EDD](https://contributing.bitwarden.com/contributing/database-migrations/edd#online-environments), specifically non-backwards-compatible migrations.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/_move_edd_db_scripts.yml
+++ b/.github/workflows/_move_edd_db_scripts.yml
@@ -1,5 +1,5 @@
-name: _move_finalization_db_scripts
-run-name: Move finalization database scripts
+name: _move_edd_db_scripts
+run-name: Move EDD database scripts
 
 on:
   workflow_call:
@@ -14,7 +14,8 @@ jobs:
     runs-on: ubuntu-22.04
     outputs:
       migration_filename_prefix: ${{ steps.prefix.outputs.prefix }}
-      copy_finalization_scripts: ${{ steps.check-finalization-scripts-existence.outputs.copy_finalization_scripts }}
+      copy_edd_scripts: ${{ steps.check-script-existence.outputs.copy_edd_scripts }}
+
     steps:
       - name: Log in to Azure
         uses: Azure/login@e15b166166a8746d1a47596803bd8c1b595455cf # v1.6.0
@@ -37,22 +38,22 @@ jobs:
         id: prefix
         run: echo "prefix=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
 
-      - name: Check if any files in DB finalization directory
-        id: check-finalization-scripts-existence
+      - name: Check if any files in DB transition or finalization directories
+        id: check-script-existence
         run: |
-          if [ -f util/Migrator/DbScripts_finalization/* ]; then
-            echo "copy_finalization_scripts=true" >> $GITHUB_OUTPUT
+          if [ -f util/Migrator/DbScripts_transition/* -o -f util/Migrator/DbScripts_finalization/* ]; then
+            echo "copy_edd_scripts=true" >> $GITHUB_OUTPUT
           else
-            echo "copy_finalization_scripts=false" >> $GITHUB_OUTPUT
+            echo "copy_edd_scripts=false" >> $GITHUB_OUTPUT
           fi
 
-  move-finalization-db-scripts:
-    name: Move finalization database scripts
+  move-scripts:
+    name: Move scripts
     runs-on: ubuntu-22.04
     needs: setup
-    if: ${{ needs.setup.outputs.copy_finalization_scripts == 'true' }}
+    if: ${{ needs.setup.outputs.copy_edd_scripts == 'true' }}
     steps:
-      - name: Checkout
+      - name: Check out repo
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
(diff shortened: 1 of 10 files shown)


.github/workflows/_move_edd_db_scripts.yml
```diff
@@ -37,22 +38,22 @@ jobs:
...
         run: echo "prefix=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
 
-      - name: Check if any files in DB finalization directory
-        id: check-finalization-scripts-existence
+      - name: Check if any files in DB transition or finalization directories
+        id: check-script-existence
         run: |
-          if [ -f util/Migrator/DbScripts_finalization/* ]; then
-            echo "copy_finalization_scripts=true" >> $GITHUB_OUTPUT
+          if [ -f util/Migrator/DbScripts_transition/* -o -f util/Migrator/DbScripts_finalization/* ]; then
```
withinfocus, line 44: ℹ️ Check both directories for scripts now.

.github/workflows/_move_edd_db_scripts.yml
```diff
@@ -61,35 +62,58 @@ jobs:
...
+      - name: Move scripts and finalization database schema
         id: move-files
         env:
           PREFIX: ${{ needs.setup.outputs.migration_filename_prefix }}
         run: |
-          src_dir="util/Migrator/DbScripts_finalization"
+          moved_files=""
+
+          # scripts
+          src_dirs="util/Migrator/DbScripts_transition,util/Migrator/DbScripts_finalization"
```
withinfocus, line 80: ℹ️ Gathers and moves all the scripts with the existing renaming. Transition scripts happen before finalization.

.github/workflows/_move_edd_db_scripts.yml
```diff
@@ -61,35 +62,58 @@ jobs:
...
+
+              filename=$(basename "$file")
+              new_filename="${PREFIX}_${filenumber}_${filename}"
+              dest_file="$dest_dir/$new_filename"
+
+              mv "$file" "$dest_file"
+              moved_files="$moved_files \n $filename -> $new_filename"
 
+              # Replace any finalization references due to the move
+              sed -i -e 's/dbo_finalization/dbo/g' "$dest_file"
```
withinfocus, line 96: ℹ️ Both the scripts and schema below get a replacement of the `dbo_finalization` schema reference, if present (should be).

.github/workflows/_move_edd_db_scripts.yml
```diff
@@ -61,35 +62,58 @@ jobs:
...
+
+              i=$((i+1))
+            done
+          done
+
+          # schema
+          src_dir="src/Sql/dbo_finalization"
+          dest_dir="src/Sql/dbo"
+
+          for file in "$src_dir"/*; do
```
withinfocus, line 106: ℹ️ Moves the schema changes.

.github/workflows/_move_edd_db_scripts.yml
```diff
@@ -137,16 +161,16 @@ jobs:
...
             --head "$BRANCH" \
             --label "automated pr" \
             --body "
-              ## Automated movement of DbScripts_finalization to DbScripts
+              Automated movement of EDD database scripts.
 
-              ## Files moved:
+              Files moved:
               $(echo -e "$MOVED_FILES")
               ")
```
withinfocus, line 175: ℹ️ This could be prettier but it works. https://github.com/bitwarden/server/pull/3675 is the only PR I have seen that used this.

util/Migrator/Migrator.csproj
```diff
@@ -3,6 +3,7 @@
   <ItemGroup>
     <EmbeddedResource Include="DbScripts\**\*.sql" />
     <EmbeddedResource Include="DbScripts_transition\**\*.sql" />
+    <EmbeddedResource Include="DbScripts_finalization\**\*.sql" />
```
withinfocus, line 6: ℹ️ Caught this.

util/Migrator/MigratorConstants.cs
```diff
@@ -4,5 +4,5 @@ public static class MigratorConstants
 {
     public const string SqlTableJournalName = "Migration";
     public const string DefaultMigrationsFolderName = "DbScripts";
-    public const string TransitionMigrationsFolderName = "DbScripts_data_migration";
+    public const string TransitionMigrationsFolderName = "DbScripts_transition";
```
withinfocus, line 7: ℹ️ Also caught this. We haven't had any, but the self-host method that uses this would have not worked.

What about finalization migrations in self-host? The [docs](https://contributing.bitwarden.com/contributing/database-migrations/edd#offline-environments) say those don't run offline, but this is an issue right?

.github/workflows/_move_edd_db_scripts.yml
```diff
@@ -61,35 +62,58 @@ jobs:
...
+              # Replace any finalization references due to the move
+              sed -i -e 's/dbo_finalization/dbo/g' "$dest_file"
+
+              i=$((i+1))
+            done
+          done
+
+          # schema
+          src_dir="src/Sql/dbo_finalization"
+          dest_dir="src/Sql/dbo"
```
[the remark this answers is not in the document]
withinfocus: For this process can we not just expect / require one folder? This needs DbOps review anyway in my opinion. Allowing this to get sucked into the rabbit hole of `CODEOWNERS` is a bad idea to me.

.github/workflows/_move_edd_db_scripts.yml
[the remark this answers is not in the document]
withinfocus: This is what was here already -- what do we want?

.github/workflows/_move_edd_db_scripts.yml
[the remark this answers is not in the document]
Hinton: The issue is the move literally won't move everything or move them to the correct spot.

If AC has a sproc in `src/Sql/AC/Sprocs/abc.sql` and they want to make a finalization they could put it in `src/Sql/dbo_finalization/Sprocs/abc.sql` but that won't move it to the correct place.

We have to lift everything into `dbo` so that the AC file is under `src/Sql/dbo/AC/Sprocs/abc.sql`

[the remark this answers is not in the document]
withinfocus: So you're suggested we get rid of the team folders in `src/Sql/dbo`? I don't see an alternative.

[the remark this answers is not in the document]
withinfocus: See #5913.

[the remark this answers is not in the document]
withinfocus: Ended up landing #5923 and now using `rsync` here for the move. Still need to test this whole process.

src/Sql/Sql.sqlproj
```diff
@@ -11,8 +11,6 @@
     <TargetFrameworkProfile />
   </PropertyGroup>
   <ItemGroup>
-    <Build Remove="dbo_future/**/*" />
-
```
Hinton, line 15: Do we know if this is fine? I'm slightly worried about this confusing DACPAC and future files being included.

(DACPAC can be used if you run compare schema)

util/Migrator/MigratorConstants.cs
Hinton: They would be run on the next release when they are prompted from finalization to regular migrations?

.github/workflows/_move_edd_db_scripts.yml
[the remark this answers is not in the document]
Hinton: It was but we need to ensure migrations are run in the correct order. If we have a migration named `YYYY-0M-FinalizationMigration.sql` as per the docs, this will probably run out of order as, and it *should* be prefixed with the full date of the PR. I.e. if we open the EDD PR on 2025-07-05, the date prefix needs to change to `2025-07-05_00_FinalizationMigration`.

Hinton (commented):
We have a couple of more files that needs to be moved. https://github.com/bitwarden/server/tree/main/src/Sql

Might be useful to add a linter step which errors if teams adds files outside any of the dbo folders?  Beyond that I think the date rename is going to cause issues and be something we have to address before this can be run automatically.

src/Sql/Sql.sqlproj
withinfocus: This is a remnant of an older process and was just missed with cleanup. I don't know of anything using "future".

withinfocus:
> We have a couple of more files that needs to be moved. https://github.com/bitwarden/server/tree/main/src/Sql
> 
> Might be useful to add a linter step which errors if teams adds files outside any of the dbo folders? Beyond that I think the date rename is going to cause issues and be something we have to address before this can be run automatically.

I fixed this in #6094, not sure why that was done when it's visible that nothing else is set up that way. Linting this feels like overkill -- you need to put certain files in certain places ...

> They would be run on the next release when they are prompted from finalization to regular migrations?

This was another mistake that just hadn't been noticed since we haven't had a transition migration since. There's no process change and you can trace the constant usage but the transition ones run automatically upon deployment after the regular migrations.

> It was but we need to ensure migrations are run in the correct order. If we have a migration named YYYY-0M-FinalizationMigration.sql as per the docs, this will probably run out of order as, and it should be prefixed with the full date of the PR. I.e. if we open the EDD PR on 2025-07-05, the date prefix needs to change to 2025-07-05_00_FinalizationMigration.

I can add this.

.github/workflows/_move_edd_db_scripts.yml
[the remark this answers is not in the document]
withinfocus: I took a look at this and it's already what it's doing -- the transition and finalization scripts will be prefixed by the date this runs.

withinfocus:
@Hinton this is ready again -- I don't think there's anything else to do.

Hinton (approved):

vgrassia (approved):

PR #5879, opened 2025-05-26
Chore: document SutProvider and clean up UserServiceTests

eliykat:
## 🎟️ Tracking

N/A
## 📔 Objective

Clean up `UserServiceTests`, which had some hacks and workarounds to compensate for improper use of `SutProvider`. These were largely solved by following the `SutProvider` builder pattern - setting dependencies before calling `SutProvider.Create`. I did come across 1 issue which I think is a bug in our `SutProvider` implementation, I wasn't able to fix it but I have at least documented the workaround. See the comment below.

A practical benefit here is that we don't have to maintain the list of dependencies in the tests anymore. Therefore, this was also a prompt for deleting a handful of unused dependencies in `UserService`.

Lastly, I took the opportunity to document the `SutProvider` interface and some of its underlying magic, which is fairly opaque otherwise (if ubiquitous).

@ike-kottlowski Requesting your review on `UserServiceTests` please, as it looks like you added these tests. I thought you may be interested in these changes.

@MGibson1 Requesting your review for accuracy on the xmldoc.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


test/Common/AutoFixture/SutProvider.cs
```diff
@@ -39,46 +62,69 @@ public SutProvider<TSut> SetDependency(Type dependencyType, object dependency, s
...
-            else
-            {
-                throw new ArgumentException(string.Concat($"Dependency of type {dependencyType.Name} and name ",
-                    $"{parameterName} does not exist. Available dependency names are: ",
-                    string.Join(", ", knownDependencies.Keys)));
-            }
-        }
-        else
-        {
-            throw new ArgumentException($"Dependency of type {dependencyType.Name} and name {parameterName} has not been set.");
```
eliykat, line 65: Changes to code here are just removing redundant `else`s. Hide whitespace changes for a better indication.

test/Core.Test/Services/UserServiceTests.cs
```diff
@@ -753,54 +705,54 @@ private static IUserTwoFactorTokenProvider<User> SetupFakeTokenProvider(SutProvi
...
+    /// PasswordHasher property with a new substitute, so it loses the configured sutProvider mock.
+    /// This doesn't usually happen because our dependencies are not usually public.
+    /// Call this AFTER SutProvider.Create().
+    /// </summary>
+    private static SutProvider<UserService> FixPasswordHasherBug(this SutProvider<UserService> sutProvider)
+    {
+        // Get the configured sutProvider mock and assign it back to the public property in the base class
+        sutProvider.Sut.PasswordHasher = sutProvider.GetDependency<IPasswordHasher<User>>();
+        return sutProvider;
+    }
```
eliykat, line 745: If a dependency is exposed as a public property, it is injected into the constructor as expected by SutProvider, but is then immediately overwritten (with a new mock) when Autofixture initializes all public properties. This just overwrites the latter mock with the one in SutProvider. This is a workaround rather than a proper fix, but it's a pretty rare circumstance (our dependencies are usually private) and I didn't have the time to dig into it further.

r-tome (dismissed):

test/Common/AutoFixture/SutProvider.cs
```diff
@@ -39,45 +62,68 @@ public SutProvider<TSut> SetDependency(Type dependencyType, object dependency, s
         return this;
     }
 
+    /// <summary>
+    /// Gets a dependency of the sut. Can only be called after the SutProvider has been initialized with <see cref="Create"/>.
```
MGibson1: I don't think this is quite true. Dependencies are only automatically set when the `Create` method is called, but if they are manually set, `GetDependency` should still work.

test/Common/AutoFixture/SutProvider.cs
```diff
@@ -116,16 +176,20 @@ public object Create(object request, ISpecimenContext context)
...
-            else if (_sutProvider.DependencyIsSet(parameterInfo.ParameterType, ""))
+
+            // Use the default dependency set for this type, if any (i.e. no parameter name has been specified)
+            if (_sutProvider.DependencyIsSet(parameterInfo.ParameterType, ""))
             {
                 return _sutProvider.GetDependency(parameterInfo.ParameterType, "");
             }
 
+            // Fallback: create an NSubstitute mock (assuming the WithAutoNSubstitutions customization has been used)
+            // and register it using SetDependency so it can be retrieved later.
```
MGibson1: This isn't specific to NSubstitute, it just uses the rest of the fixture chain to populate this value rather than a previously stored value. 

The comment below is saying that this is identical to a `fixture.Create` call for this parameter value, but the required method doesn't exist so we need to do this insanity.

MGibson1 (commented):
Two nits with sutprovider docs, but overall very helpful.

I also DM'ed you a diff that _might_ help with your empty property problem. It sounds like an AutoFixture customization resolution issue ([blog about customization ordering](https://blog.ploeh.dk/2012/07/31/TheorderofAutoFixtureCustomizationsmatter/)), so I just reordered to ensure the parameter relay was first.

test/Common/AutoFixture/SutProvider.cs
eliykat: Wording has been adjusted

test/Common/AutoFixture/SutProvider.cs
eliykat: I was trying to communicate the practical effect of this (i.e. generally an NSubstitute mock) - saying it's equivalent to `fixture.Create<T>` isn't necessary clear to all readers. I've updated the wording to be more accurate though.

MGibson1: fair, I mentioned because a function of the sut provider is that you can pass in an existing AutoFixture instance, which may be pre-populated with customizations or frozen types

r-tome (approved):

PR #5883, opened 2025-05-27
[PM-21975] Add Security Stamp claim to persisted grant

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21975

## 📔 Objective

When a user deauthorizes all sessions (or performs other account actions that invalidate the authentication guarantees of their account), we update the user's SecurityStamp.

In order to enforce this, we rely on the code [here](https://github.com/bitwarden/server/blob/main/src/Identity/IdentityServer/ProfileService.cs#L75-L76) that is intended to ensure that a refresh token is not valid if the security stamp does not match.  IdentityServer will call `IsActiveAsync()` to determine whether an access token should be issued.

However, this has never worked, because we were not adding the `Claims.SecurityStamp` claim to the refresh token.  This meant that it was never available for comparison and therefore never enforced.

In order to add the claim to the refresh token, we should be adding it to the `GrantValidationResult` that we're building in the `BaseRequestValidator`.  This ensures that when IdentityServer builds the response to the `/connect/token` request (including the `refresh_token`), the security stamp claim will be included.

**This will be enforced whenever the client requests a new access token with their refresh token.**. This occurs:
- When the client determines that the access token is within 5 minutes of expiration (~every 60 minutes)
- When we explicitly call `refreshIdentityToken()` on the client.  Most likely, for most users, this will occur on sync.

## 📸 Screenshots

https://github.com/user-attachments/assets/ed073563-552e-4c30-86cb-b2e017cf6126

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Utilities/CoreHelpers.cs
+++ b/src/Core/Utilities/CoreHelpers.cs
@@ -712,6 +712,7 @@ public static class CoreHelpers
             new(Claims.Premium, isPremium ? "true" : "false"),
             new(JwtClaimTypes.Email, user.Email),
             new(JwtClaimTypes.EmailVerified, user.EmailVerified ? "true" : "false"),
+            // TODO: [https://bitwarden.atlassian.net/browse/PM-22171] Remove this since it is already added from the persisted grant
             new(Claims.SecurityStamp, user.SecurityStamp),
         };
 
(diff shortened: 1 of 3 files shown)


JaredSnider-Bitwarden (approved):
Excellent changes + stellar docs. Tyvm!

PR #5895, opened 2025-05-28
[PM-21918] update send api models to support new `email` field

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21918

## 📔 Objective

Update send API models to support email/OTP administration. 

>[!IMPORTANT]
> This PR includes documentation for the related models. I mostly inferred what each of these are. Please focus on them during review to confirm that the comments are correct.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Tools/Models/Request/SendRequestModel.cs
+++ b/src/Api/Tools/Models/Request/SendRequestModel.cs
@@ -7,33 +7,106 @@ using Bit.Core.Tools.Models.Data;
 using Bit.Core.Tools.Services;
 using Bit.Core.Utilities;
 
+using static System.StringSplitOptions;
+
 namespace Bit.Api.Tools.Models.Request;
 
+/// <summary>
+/// A send request issued by a Bitwarden client
+/// </summary>
 public class SendRequestModel
 {
+    /// <summary>
+    /// Indicates whether the send contains text or file data.
+    /// </summary>
     public SendType Type { get; set; }
+
+    /// <summary>
+    /// Estimated length of the file accompanying the send. <see langword="null"/> when
+    /// <see cref="Type"/> is <see cref="SendType.Text"/>.
+    /// </summary>
     public long? FileLength { get; set; } = null;
+
+    /// <summary>
+    /// Label for the send.
+    /// </summary>
     [EncryptedString]
     [EncryptedStringLength(1000)]
     public string Name { get; set; }
+
+    /// <summary>
+    /// Notes for the send. This is only visible to the owner of the send.
+    /// </summary>
     [EncryptedString]
     [EncryptedStringLength(1000)]
     public string Notes { get; set; }
+
+    /// <summary>
+    /// A base64-encoded byte array containing the Send's encryption key. This key is
+    /// also provided to send recipients in the Send's URL.
+    /// </summary>
     [Required]
     [EncryptedString]
     [EncryptedStringLength(1000)]
     public string Key { get; set; }
+
+    /// <summary>
+    /// The maximum number of times a send can be accessed before it expires.
+    /// When this value is <see langword="null" />, there is no limit.
+    /// </summary>
     [Range(1, int.MaxValue)]
     public int? MaxAccessCount { get; set; }
+
+    /// <summary>
+    /// The date after which a send cannot be accessed. When this value is
+    /// <see langword="null"/>, there is no expiration date.
+    /// </summary>
     public DateTime? ExpirationDate { get; set; }
+
+    /// <summary>
+    /// The date after which a send may be automatically deleted from the server.
+    /// When this is <see langword="null" />, the send may be deleted after it has
+    /// exceeded the global send timeout limit.
+    /// </summary>
     [Required]
     public DateTime? DeletionDate { get; set; }
+
+    /// <summary>
+    /// Contains file metadata uploaded with the send.
+    /// The file content is uploaded separately.
+    /// </summary>
     public SendFileModel File { get; set; }
+
+    /// <summary>
+    /// Contains text data uploaded with the send.
+    /// </summary>
     public SendTextModel Text { get; set; }
+
+    /// <summary>
+    /// Base64-encoded byte array of a password hash that grants access to the send.
+    /// Mutually exclusive with <see cref="Emails"/>.
+    /// </summary>
     [StringLength(1000)]
     public string Password { get; set; }
+
+    /// <summary>
+    /// Comma-separated list of emails that may access the send using OTP
+    /// authentication. Mutually exclusive with <see cref="Password"/>.
+    /// </summary>
+    [StringLength(1024)]
+    public string Emails { get; set; }
+
+    /// <summary>
+    /// When <see langword="true"/>, send access is disabled.
+    /// Defaults to <see langword="false"/>.
+    /// </summary>
     [Required]
     public bool? Disabled { get; set; }
+
+    /// <summary>
+    /// When <see langword="true"/> send access hides the user's email address
+    /// and displays a confirmation message instead. Defaults to <see langword="false"/>.
+    /// </summary>
     public bool? HideEmail { get; set; }
 
     public Send ToSend(Guid userId, ISendAuthorizationService sendAuthorizationService)
@@ -78,6 +151,12 @@ public class SendRequestModel
         return existingSend;
     }
 
+    /// <summary>
+    /// Validates that the request is internally consistent for send creation.
+    /// </summary>
+    /// <exception cref="BadRequestException">
+    /// Thrown when the send's expiration date has already expired.
+    /// </exception>
     public void ValidateCreation()
     {
         var now = DateTime.UtcNow;
@@ -91,6 +170,13 @@ public class SendRequestModel
         ValidateEdit();
     }
 
+    /// <summary>
+    /// Validates that the request is internally consistent for send administration.
+    /// </summary>
+    /// <exception cref="BadRequestException">
+    /// Thrown when the send's deletion date has already expired or when its
+    /// expiration occurs after its deletion.
+    /// </exception>
     public void ValidateEdit()
     {
         var now = DateTime.UtcNow;
@@ -135,6 +221,13 @@ public class SendRequestModel
         {
             existingSend.Password = authorizationService.HashPassword(Password);
         }
+        else if (!string.IsNullOrWhiteSpace(Emails))
+        {
+            // normalize encoding
+            var emails = Emails.Split(',', RemoveEmptyEntries | TrimEntries);
+            existingSend.Emails = string.Join(", ", emails);
+        }
+
         existingSend.Disabled = Disabled.GetValueOrDefault();
         existingSend.HideEmail = HideEmail.GetValueOrDefault();
         return existingSend;
@@ -146,8 +239,15 @@ public class SendRequestModel
     }
 }
 
+/// <summary>
+/// A send request issued by a Bitwarden client
+/// </summary>
 public class SendWithIdRequestModel : SendRequestModel
 {
+    /// <summary>
+    /// Identifies the send. When this is <see langword="null" />, the client is requesting
+    /// a new send.
+    /// </summary>
     [Required]
     public Guid? Id { get; set; }
 }
--- a/src/Api/Tools/Models/Response/SendAccessResponseModel.cs
+++ b/src/Api/Tools/Models/Response/SendAccessResponseModel.cs
@@ -8,8 +8,24 @@ using Bit.Core.Utilities;
 
 namespace Bit.Api.Tools.Models.Response;
 
+/// <summary>
+/// A response issued to a Bitwarden client in response to access operations.
+/// </summary>
 public class SendAccessResponseModel : ResponseModel
 {
+    /// <summary>
+    /// Instantiates a send access response model
+    /// </summary>
+    /// <param name="send">Content to transmit to the client.</param>
+    /// <param name="globalSettings">
+    /// Settings that control response generation.
+    /// </param>
+    /// <exception cref="ArgumentNullException">
+    /// Thrown when <paramref name="send"/> is <see langword="null" />
+    /// </exception>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="send" /> has an invalid <see cref="Send.Type"/>.
+    /// </exception>
     public SendAccessResponseModel(Send send, GlobalSettings globalSettings)
         : base("send-access")
     {
@@ -42,11 +58,47 @@ public class SendAccessResponseModel : ResponseModel
         ExpirationDate = send.ExpirationDate;
     }
 
+    /// <summary>
+    /// Identifies the send in a send URL
+    /// </summary>
     public string Id { get; set; }
+
+    /// <summary>
+    /// Indicates whether the send contains text or file data.
+    /// </summary>
     public SendType Type { get; set; }
+
+    /// <summary>
+    /// Label for the send. This is only visible to the owner of the send.
+    /// </summary>
+    /// <remarks>
+    /// This field contains a base64-encoded byte array. The array contains
+    /// the E2E-encrypted encrypted content.
+    /// </remarks>
     public string Name { get; set; }
+
+    /// <summary>
+    /// Describes the file attached to the send.
+    /// </summary>
+    /// <remarks>
+    /// File content is downloaded separately using
+    /// <see cref="Bit.Api.Tools.Controllers.SendsController.GetSendFileDownloadData" />
+    /// </remarks>
     public SendFileModel File { get; set; }
+
+    /// <summary>
+    /// Contains text data uploaded with the send.
+    /// </summary>
     public SendTextModel Text { get; set; }
+
+    /// <summary>
+    /// The date after which a send cannot be accessed. When this value is
+    /// <see langword="null"/>, there is no expiration date.
+    /// </summary>
     public DateTime? ExpirationDate { get; set; }
+
+    /// <summary>
+    /// Indicates the person that created the send to the accessor.
+    /// </summary>
     public string CreatorIdentifier { get; set; }
 }
--- a/src/Api/Tools/Models/Response/SendResponseModel.cs
+++ b/src/Api/Tools/Models/Response/SendResponseModel.cs
@@ -8,8 +8,25 @@ using Bit.Core.Utilities;
 
 namespace Bit.Api.Tools.Models.Response;
 
+/// <summary>
+/// A response issued to a Bitwarden client in response to ownership operations.
+/// </summary>
+/// <seealso cref="SendAccessResponseModel" />
 public class SendResponseModel : ResponseModel
 {
+    /// <summary>
+    /// Instantiates a send response model
+    /// </summary>
+    /// <param name="send">Content to transmit to the client.</param>
+    /// <param name="globalSettings">
+    /// Settings that control response generation.
+    /// </param>
+    /// <exception cref="ArgumentNullException">
+    /// Thrown when <paramref name="send"/> is <see langword="null" />
+    /// </exception>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="send" /> has an invalid <see cref="Send.Type"/>.
+    /// </exception>
     public SendResponseModel(Send send, GlobalSettings globalSettings)
         : base("send")
     {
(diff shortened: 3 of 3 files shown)


audreyality:
📝 This PR may wait to merge until the CLI logic is stable.

src/Api/Tools/Models/Response/SendResponseModel.cs
```diff
@@ -8,8 +8,25 @@
...
+/// A response issued to a Bitwarden client in response to ownership operations.
+/// </summary>
+/// <seealso cref="SendAccessResponseModel" />
 public class SendResponseModel : ResponseModel
 {
+    /// <summary>
+    /// Instantiates a send response model
+    /// </summary>
+    /// <param name="send">Content to transmit to the client.</param>
+    /// <param name="globalSettings">
```
djsmith85: ⚠️  The globalSettings param isn't used anywhere and can be removed

src/Api/Tools/Models/Response/SendAccessResponseModel.cs
```diff
@@ -8,8 +8,24 @@
...
+/// <summary>
+/// A response issued to a Bitwarden client in response to access operations.
+/// </summary>
 public class SendAccessResponseModel : ResponseModel
 {
+    /// <summary>
+    /// Instantiates a send access response model
+    /// </summary>
+    /// <param name="send">Content to transmit to the client.</param>
+    /// <param name="globalSettings">
```
djsmith85: ⚠️  The globalSettings param isn't used anywhere and can be removed

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -7,33 +7,106 @@
...
     [Required]
     public bool? Disabled { get; set; }
+
+    /// <summary>
+    /// When <see langword="true"/> send access hides the user's email address
+    /// and displays a confirmation message instead. Defaults to <see langword="false"/>.
+    /// </summary>
     public bool? HideEmail { get; set; }
 
     public Send ToSend(Guid userId, ISendAuthorizationService sendAuthorizationService)
```
djsmith85, line 127: ⚠️ Missing docs on the public method ToSend and overload below

src/Api/Tools/Models/Response/SendResponseModel.cs
audreyality: While I agree, that would drastically increase the scope of this PR. I think this is better done as a followup. Added FIXME in 360a3dacf16850a03bb3. #6016 resolves the FIXME.

src/Api/Tools/Models/Request/SendRequestModel.cs
audreyality: Addressed by 74da4abb6224ddebd8b5915d1926f0f. 

I also added fixmes for methods that don't present a uniform level of abstraction.

src/Api/Tools/Models/Response/SendAccessResponseModel.cs
audreyality: While I agree, that would drastically increase the scope of this PR. I think this is better done as a followup. Added FIXME in 7b0b94950eae259d3452603c9fe2. #6016 resolves the FIXME.

audreyality:
Moving this back to draft because the server requirements have changed.

src/Api/Vault/Models/Response/SyncResponseModel.cs
```diff
@@ -54,7 +54,7 @@ public SyncResponseModel(
             c => new CollectionDetailsResponseModel(c)) ?? new List<CollectionDetailsResponseModel>();
         Domains = excludeDomains ? null : new DomainsResponseModel(user, false);
         Policies = policies?.Select(p => new PolicyResponseModel(p)) ?? new List<PolicyResponseModel>();
-        Sends = sends.Select(s => new SendResponseModel(s, globalSettings));
```
audreyality, line 59: 📝 @jaasen-livefront - this is the only vault change. We're removing this parameter because it's unused.

jaasen-livefront: Love these comments!

jaasen-livefront (dismissed):
Approve vault change

djsmith85 (dismissed):

djsmith85 (approved):

src/Api/Tools/Controllers/SendsController.cs
```diff
@@ -293,8 +295,8 @@ public async Task<SendResponseModel> Put(string id, [FromBody] SendRequestModel
             throw new NotFoundException();
         }
 
-        await _nonAnonymousSendCommand.SaveSendAsync(model.ToSend(send, _sendAuthorizationService));
-        return new SendResponseModel(send, _globalSettings);
+        await _nonAnonymousSendCommand.SaveSendAsync(model.UpdateSend(send, _sendAuthorizationService));
+        return new SendResponseModel(send);
     }
 
     [HttpPut("{id}/remove-password")]
```
itsadrago, line 307: Do we need a remove-emails endpoint?

harr1424: I am actually not sure, did we solution that? I foresee a situation where only a subset of emails might need to be removed from any given send.

itsadrago: Yes, we should be able to add emails and remove some or all as desired by the user. The current process of updating a Send should be able to handle that. I'm not sure why we have a remove-password endpoint versus just letting the owner of the Send remove it through editing the Send

util/SqliteMigrations/_________
itsadrago, line 1: What is this file for? I'm assuming it shouldn't be checked in

harr1424: I am still not clear on which migrations need to be checked in, I removed all the EF specific migration directories (SQLite, MariaDB, PostgreSQL,MySQL) created during `pwsh migrate.psl -all` and then I saw many deleted files appear in the diff against `main` (see [7df56e3](https://github.com/bitwarden/server/pull/5895/commits/7df56e346ab3402387bebffc1d112d73214632fa)). I was going to work on getting the failing test fixed and then review our documentation once more.

harr1424: But yes I agree that specific file should be removed and will update you if I find any documentation regarding committing specific DB vendor migrations

src/Api/Tools/Models/Response/SendResponseModel.cs
```diff
@@ -24,13 +37,17 @@ public SendResponseModel(Send send, GlobalSettings globalSettings)
         Id = send.Id;
         AccessId = CoreHelpers.Base64UrlEncode(send.Id.ToByteArray());
         Type = send.Type;
+        AuthType = send.AuthType ?? (!string.IsNullOrWhiteSpace(send.Password)
```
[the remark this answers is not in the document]
harr1424: @theMickster this is a helpful comment, but shouldn't Claude take into context our [contribution guidelines](https://contributing.bitwarden.com/contributing/code-style/csharp#conditionals)? 

Here Claude has violated guidelines regarding conditionals and brackets.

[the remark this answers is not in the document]
theMickster: You need to remember that Claude Code cannot possible read every single contributing document on every single PR. The PR code review would never complete because it would take forever. The tool is just another helper in our toolbox and we take it with a grain of salt. I think this is a nit pick that doesn't require deep attention.  It our jobs as engineers to interpret the suggestion just like we would from a person. 

Please do call out using the @ + claude tag when you see genuine, invalid or harmful code being suggested. however, style-cop related things aren't really relevant.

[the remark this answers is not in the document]
harr1424: @theMickster my apologies if my comment seemed nit-picky, it was intended to be an idea for a possible improvement (using our contributing docs as context). Thanks for all of your work on this integration, it has improved significantly since the very first reviews!

util/Migrator/DbScripts/2025-12-18_00_SendAuthType.sql
```diff
@@ -0,0 +1,140 @@
+IF COL_LENGTH('[dbo].[Send]', 'AuthType') IS NULL
+BEGIN
+ALTER TABLE [dbo].[Send]
+    ADD [AuthType] TINYINT NULL;
+END
+GO
+
+-- Update Send_Create to include AuthType parameter
+CREATE OR ALTER PROCEDURE [dbo].[Send_Create]
```
mzieniukbw, line 9: ❓ Why do we need two database migration scripts, that both update the same stored procedures ?

harr1424: @mzieniukbw `2025-12-18_00_SendAuthType.sql` was written first in order to add the new column to the `Send` table. Afterwards, existing unit tests were checked and new unit tests added to cover logic related to the new column. During this time a decision was pending involving how to store the emails to associated with a `Send` using the email OTP authentication scheme. It was decided that a comma separated list of emails would be stored in the `Emails` column, and that `VARCHAR(4000)` would be a suitable datatype for this column.  Once this was decided, I added the migration script `2025-12-18_01_SendEmailsLength.sql`. 

If you prefer the updated logic is consolidated into a single migration script, please let me know.

mzieniukbw: I don't mind.

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -10,48 +10,141 @@
...
+    /// Mutually exclusive with <see cref="Emails"/>.
+    /// </summary>
     [StringLength(1000)]
     public string Password { get; set; }
+
+    /// <summary>
+    /// Comma-separated list of emails that may access the send using OTP
+    /// authentication. Mutually exclusive with <see cref="Password"/>.
+    /// </summary>
+    [StringLength(1024)]
```
itsadrago: Needs to be updated to a length of 4000

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -134,12 +246,41 @@ private Send ToSendBase(Send existingSend, ISendAuthorizationService authorizati
         existingSend.ExpirationDate = ExpirationDate;
         existingSend.DeletionDate = DeletionDate.Value;
         existingSend.MaxAccessCount = MaxAccessCount;
-        if (!string.IsNullOrWhiteSpace(Password))
+
+        if (!string.IsNullOrWhiteSpace(Emails))
+        {
+            // normalize encoding
+            var emails = Emails.Split(',', RemoveEmptyEntries | TrimEntries);
+            existingSend.Emails = string.Join(", ", emails);
```
itsadrago: Do we want to add a space between email addresses? How you want to display the comma-separated string should be up to the presentation tier and not a part of the storage tier

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -134,12 +246,41 @@ private Send ToSendBase(Send existingSend, ISendAuthorizationService authorizati
...
+                existingSend.AuthType = Core.Tools.Enums.AuthType.Password;
+            }
+            else if (!string.IsNullOrWhiteSpace(existingSend.Emails))
+            {
+                existingSend.AuthType = Core.Tools.Enums.AuthType.Email;
+            }
+            else
+            {
+                existingSend.AuthType = Core.Tools.Enums.AuthType.None;
+            }
```
itsadrago: Should you use `SendUtilities.InferAuthType(send)` here?

util/Migrator/DbScripts/2025-12-18_00_SendAuthType.sql
```diff
@@ -0,0 +1,140 @@
...
+    WHERE
+        [Id] = @Id
+
+    IF @UserId IS NOT NULL
+    BEGIN
+        EXEC [dbo].[User_BumpAccountRevisionDate] @UserId
+    END
+    -- TODO: OrganizationId bump?
+END
+GO
```
rkac-bw, line 140: Should refresh view after schema change 
```suggestion
GO
EXECUTE sp_refreshview N'[dbo].[SendView]'
GO
```

src/Sql/dbo/Tools/Stored Procedures/Send_Create.sql
```diff
@@ -48,6 +50,7 @@ BEGIN
         @UserId,
         @OrganizationId,
         @Type,
+     @AuthType,
```
rkac-bw: :pick: indentation off , has 5 spaces instead of 8

src/Api/Tools/Controllers/SendsController.cs
```diff
@@ -286,36 +293,39 @@ await Request.GetFileAsync(async (stream) =>
...
     public async Task<SendResponseModel> PutRemovePassword(string id)
     {
-        var userId = _userService.GetProperUserId(User).Value;
+        var userId = _userService.GetProperUserId(User) ?? throw new InvalidOperationException("User ID not found");
         var send = await _sendRepository.GetByIdAsync(new Guid(id));
         if (send == null || send.UserId != userId)
         {
             throw new NotFoundException();
```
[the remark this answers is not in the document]
itsadrago: @harr1424 what do you think about this?

src/Api/Tools/Controllers/SendsController.cs
itsadrago: I found out that the UI offers the option to delete a password without editing a Send on almost all the clients. I have a question out to product and design about whether emails should be treated the same way

src/Api/Tools/Controllers/SendsController.cs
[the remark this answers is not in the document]
harr1424: @mathewpriya @sukhleenb do either of you have any input why we may have added a dedicated endpoint for removing password based authentication? As opposed to using the existing endpoint to edit the send and remove the password instead? 

@itsadrago I am definitely not opposed to adding the endpoint if it will be used, but sometimes Claude's comments are a bit silly once any investigation takes place.

src/Api/Tools/Models/Request/SendRequestModel.cs
harr1424: Good catch!

src/Sql/dbo/Tools/Stored Procedures/Send_Create.sql
```diff
@@ -3,6 +3,7 @@
     @UserId UNIQUEIDENTIFIER,
     @OrganizationId UNIQUEIDENTIFIER,
     @Type TINYINT,
+    @AuthType TINYINT,
```
mkincaid-bw: Stored procs need to support both the new version and previous version of the server code. Adding a new required parameter does not follow [EDD](https://contributing.bitwarden.com/contributing/database-migrations/edd/). When adding a new parameter to an existing proc, it should have a default value and ideally should be added to the end of the list of parameters (that part is not as critical though since we use named parameters when calling stored procs from code).

src/Sql/dbo/Tools/Stored Procedures/Send_Update.sql
```diff
@@ -3,6 +3,7 @@
     @UserId UNIQUEIDENTIFIER,
     @OrganizationId UNIQUEIDENTIFIER,
     @Type TINYINT,
+    @AuthType TINYINT,
```
mkincaid-bw: Same comment as above.

src/Api/Tools/Models/Request/SendRequestModel.cs
harr1424: I don't think the spaces are necessary here, we don't expect many sends that use email OTP to push the 4000 character limit, but why waste the space? I don't think the server needs to be concerned with displaying these in a format optimized for human readability, but rather optimal with respect to space. On the client side we can do an inexpensive string.replace(',', ', ') or whatever we need to optimize for UI/UX on the clients.

src/Sql/dbo/Tools/Tables/Send.sql
```diff
@@ -3,10 +3,11 @@
     [UserId]         UNIQUEIDENTIFIER NULL,
     [OrganizationId] UNIQUEIDENTIFIER NULL,
     [Type]           TINYINT          NOT NULL,
+    [AuthType]       TINYINT          NULL,
```
mkincaid-bw: When he migration script runs, the `AuthType` column will be added as the last column in the table. This should be reflected in the object file so it matches what the schema will actually look like, ie: 

```sql
...
[CipherId]       UNIQUEIDENTIFIER NULL,
[AuthType]       TINYINT          NULL,
```

src/Api/Tools/Models/Request/SendRequestModel.cs
harr1424: Another great catch, I will update this.

src/Sql/dbo/Tools/Stored Procedures/Send_Create.sql
harr1424: My apologies, the property mapped to this column is nullable, and I believed `NULL` would be assigned here. I will update this to explicitly assign the appropriate enum variant and move the line as well.

src/Sql/dbo/Tools/Stored Procedures/Send_Update.sql
```diff
@@ -15,7 +15,8 @@
     @Disabled BIT,
     @HideEmail BIT,
     @CipherId UNIQUEIDENTIFIER = NULL,
-    @Emails NVARCHAR(1024) = NULL
+    @Emails NVARCHAR(4000) = NULL,
+    @Type TINYINT = 2
```
mkincaid-bw: Should this be `AuthType` instead of `Type`?

util/Migrator/DbScripts/2025-12-18_00_SendAuthType.sql
```diff
@@ -0,0 +1,142 @@
...
+
+-- Update Send_Create to include AuthType parameter
+CREATE OR ALTER PROCEDURE [dbo].[Send_Create]
+    @Id UNIQUEIDENTIFIER OUTPUT,
+    @UserId UNIQUEIDENTIFIER,
+    @OrganizationId UNIQUEIDENTIFIER,
+    @Type TINYINT,
+    @AuthType TINYINT,
```
mkincaid-bw: The changes in the object files should also be reflected here.

util/Migrator/DbScripts/2025-12-18_01_SendEmailsLength.sql
```diff
@@ -0,0 +1,146 @@
...
+END
+GO
+
+-- Update Send_Create to use NVARCHAR(4000) for Emails
+CREATE OR ALTER PROCEDURE [dbo].[Send_Create]
+    @Id UNIQUEIDENTIFIER OUTPUT,
+    @UserId UNIQUEIDENTIFIER,
+    @OrganizationId UNIQUEIDENTIFIER,
+    @Type TINYINT,
+    @AuthType TINYINT,
```
mkincaid-bw: The changes in the object files should also be reflected here.

src/Api/Tools/Models/Request/SendRequestModel.cs
itsadrago: 100%

itsadrago (dismissed):

itsadrago (approved):

src/Sql/dbo/Tools/Stored Procedures/Send_Create.sql
```diff
@@ -17,7 +17,8 @@
     @CipherId UNIQUEIDENTIFIER = NULL,
 --  FIXME: remove null default value once this argument has been
 --         in 2 server releases
-    @Emails NVARCHAR(1024) = NULL
+    @Emails NVARCHAR(4000) = NULL,
+    @AuthType TINYINT = 2
```
mkincaid-bw: Default value should be `NULL`

```suggestion
    @AuthType TINYINT = NULL
```

src/Sql/dbo/Tools/Stored Procedures/Send_Update.sql
```diff
@@ -15,7 +15,8 @@
     @Disabled BIT,
     @HideEmail BIT,
     @CipherId UNIQUEIDENTIFIER = NULL,
-    @Emails NVARCHAR(1024) = NULL
+    @Emails NVARCHAR(4000) = NULL,
+    @AuthType TINYINT = 2
```
mkincaid-bw: Default value should be `NULL`

```suggestion
    @AuthType TINYINT = NULL
```

itsadrago (approved):

shane-melton (approved):

PR #5899, opened 2025-05-30
[PM-17562] Add strict delay support for RabbitMQ; Refactor implementation

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This PR extends the previous two PRs (#5795 #5880) to cleanup and refine a few aspects:
- Strict RabbitMQ delay support (more on this below)
- Add #nullable support for far more classes per @eliykat 's request on the previous PRs
- Pull out some pieces that were tied to Rabbit or ASB but could be reused across both
- Added `RabbitMqService` and `AzureServiceBusService` (with interfaces for testing) to consolidate the code around setting up both - things like configuring exchanges/topics and queues/subscription are now handled in one place. 
- Add far more unit testing to pieces that we previously couldn't test due to being tied tightly to Rabbit/ASB

### Strict delay support for RabbitMQ
This PR adds two different ways of configuring RabbitMQ to honor the `DelayUntilDate` for retries:
1. Delay plugin
   - [Delay plugin GitHub repo](https://github.com/rabbitmq/rabbitmq-delayed-message-exchange)
   - This plugin enables a delayed message exchange in RabbitMQ that support delaying a message for an amount of time specified in a special header.
   - This allows us to forego using any retry queues and rely instead on the delay exchange. When a message is marked with the header it gets published to the exchange and the exchange handles all the functionality of holding it until the appropriate time.
   - There is a new `useDelayPlugin` `bool` in `RabbitMqSettings` in `GlobalSettings` that defaults to false. It must be set the `true` for the system to create the delayed exchange and start using it.

#### Configuration to enable the delay plugin
- Turn on the `useDelayPlugin` flag. Push out secrets with the PowerShell script.
- Add the plugin to your RabbitMQ instance. The easiest way to do this is to add the following line to Docker compose, under the `volumes:`. This puts the binary into the container that Docker spins up

``` 
  -  ./rabbitmq_delayed_message_exchange-4.1.0.ez:/opt/rabbitmq/plugins/rabbitmq_delayed_message_exchange-4.1.0.ez 
```

- Build/rebuild your RabbitMQ instance with this in place to enable the ability to use the delay plugin.
- If you have previously created the Integration exchange, you have to delete it and let the server code recreate it (Rabbit will use the existing exchange without delay enabled if it already exists)
- Restart the servers and verify that the exchange was created successfully.

2. Retry queues + timing check
   - If the delay plugin setting is off, we revert to a similar setup introduced in #5795 where we push the retry off to a retry queue for a fixed amount of time before it gets re-published back to the main queue.
   - However, this PR adds and additional check when the message comes off the queue to see if the `DelayUntilDate` is already passed. 
     - If it has passed, we then handle the integration normally and retry the request.
     - If it is still in the future, we put the message back on the retry queue for an additional wait.
   - While this does use extra processing, it gives us better support for honoring the delays even if the delay plugin isn't enabled. Since this solution is only intended for self-host, it should be a pretty minimal impact with short delays and a small number of retries.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/WebhookIntegrationHandler.cs
```diff
@@ -29,7 +31,7 @@ public override async Task<IntegrationHandlerResult> HandleAsync(IntegrationMess
             case HttpStatusCode.ServiceUnavailable:
             case HttpStatusCode.GatewayTimeout:
                 result.Retryable = true;
-                result.FailureReason = response.ReasonPhrase;
+                result.FailureReason = response.ReasonPhrase ?? string.Empty;
```
withinfocus, line 34: ❓ Should this and the same thing below return a more "default" failure reason instead of an empty string?

withinfocus (approved):
Note that per https://contributing.bitwarden.com/architecture/adr/csharp-nullable-reference-types which we literally just proposed this week, we will flip some of this eventually. What's here is what we do today though.

I have some questions and maybe there's an improvement or two but this works. Huge amount of changes that are difficult to fully process via reading, but the tests will need to cover it.

src/Core/AdminConsole/Services/Implementations/RabbitMqIntegrationListenerService.cs
```diff
@@ -9,183 +12,137 @@ namespace Bit.Core.Services;
...
+    internal async Task ProcessReceivedMessageAsync(BasicDeliverEventArgs ea, CancellationToken cancellationToken)
+    {
+        var channel = await _lazyChannel.Value;
+        try
         {
             var json = Encoding.UTF8.GetString(ea.Body.Span);
 
-            try
+            // Determine if the message came off of the retry queue too soon
+            // If so, place it back on the retry queue
```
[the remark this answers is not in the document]
brant-livefront: This is the case when we are _not_ using the Rabbit delay plugin. If we are just using a retry queue with a fixed amount of time, that amount of time may not be longer than `DelayUntilDate`. So we have to check and see if the delay has been sufficient and then re-sleep the message until it is.

For instance if we have a requested delay of 1 min and a retry queue with a delay of 30s, the first time out of the queue it won't have delayed long enough. So we put it back on and the next publish would be ready to go.

This is not an issue when using the delay plugin. It doesn't publish the message until the delay is complete.

src/Core/AdminConsole/Services/Implementations/WebhookIntegrationHandler.cs
brant-livefront: Good point. I changed it in the next commit to at least provide us with the status code:

``` csharp
    result.FailureReason = response.ReasonPhrase ?? $"Failure with status code {(int)response.StatusCode}";
```

src/Core/AdminConsole/Services/EventLoggingListenerService.cs
```diff
@@ -37,7 +37,7 @@ internal async Task ProcessReceivedMessageAsync(string body, string messageId)
             }
             else
             {
-                _logger.LogError($"An error occurred while processing message {messageId} - Invalid JSON");
+                _logger.LogError("An error occured while processing message: {MessageId} - Invalid JSON", messageId);
```
withinfocus:
```suggestion
                _logger.LogError("An error occurred while processing message: {MessageId} - Invalid JSON", messageId);
```

Did Sonar suggest this? Two Rs!

withinfocus (changes requested):
Typo. I know you're working on the message ID topic too.

r-tome (approved):

brant-livefront:
@r-tome @withinfocus My apologies on this - I realized this morning that I had done all the work in guarding for the MessageId being null, but I hadn't actually made sure we _set_ the MessageId when publishing. This last commit fixes that (and the tests).

PR #5900, opened 2025-05-30
[PM-22204] - update cipher/share endpoint to return revision date

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22204

## 📔 Objective

This PR updates the cipher/share endpoint to return revision date. Some cleanup to the Task was also done as well as specs added.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Vault/Controllers/CiphersController.cs
+++ b/src/Api/Vault/Controllers/CiphersController.cs
@@ -1073,7 +1073,7 @@ public class CiphersController : Controller
         var userId = _userService.GetProperUserId(User).Value;
 
         var ciphers = await _cipherRepository.GetManyByUserIdAsync(userId, withOrganizations: false);
-        var ciphersDict = ciphers.ToDictionary(d => d.Id);
+        var ciphersDict = ciphers.ToDictionary(c => c.Id);
 
         // Validate the model was encrypted for the posting user
         foreach (var cipher in model.Ciphers)
@@ -1082,7 +1082,7 @@ public class CiphersController : Controller
                 throw new BadRequestException("Cipher was not encrypted for the current user. Please try again.");
         }
 
-        var shareInfos = new List<(Cipher cipher, DateTime? lastKnownRevisionDate)>();
+        var shareCiphers = new List<(Cipher, DateTime?)>();
         foreach (var cipher in model.Ciphers)
         {
             if (!ciphersDict.TryGetValue(cipher.Id.Value, out var existingCipher))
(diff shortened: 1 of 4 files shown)


src/Api/Vault/Controllers/CiphersController.cs
```diff
@@ -1064,7 +1064,7 @@ public async Task MoveMany([FromBody] CipherBulkMoveRequestModel model)
 
     [HttpPut("share")]
     [HttpPost("share")]
-    public async Task PutShareMany([FromBody] CipherBulkShareRequestModel model)
+    public async Task<Dictionary<Guid, DateTime>> PutShareMany([FromBody] CipherBulkShareRequestModel model)
```
cd-bitwarden, line 1067: Hey! I was wondering if you looked into returning a list of CipherResponseModel similar to how PutShare works. Is there a reason why we wouldn't use the response model? 

`public async Task<CipherResponseModel> PutShare(Guid id, [FromBody] CipherShareRequestModel model)
`

cd-bitwarden (commented):
Left a comment about something I'm curious about :)

src/Api/Vault/Controllers/CiphersController.cs
jaasen-livefront: @cd-bitwarden Returning the entire model seems a bit heavy-handed since we're only concerned with the `revisionDate`. Since this could involve many ciphers and collections I was hoping to keep it as lean as possible to avoid unnecessary lookups.

cd-bitwarden: My preference would be to return the full object as it exists in the database. From what I understand, we use this data to update the client-side cache, and any mismatch can trigger errors like "cipher is out of date." I get your point about the amount of data being returned, but having the exact post-update cipher details from the database could help prevent these issues in the future. Might be worth getting a third opinion— @shane-melton , do you have any thoughts?

shane-melton: I think returning the `CipherMiniResponseModel` would be a happy medium between both and is an existing pattern we follow for other bulk/many endpoints. (See [RestoreMany](https://github.com/bitwarden/server/blob/97fbf219774c271feb7752b4f42dc1f1574a3aa3/src/Api/Vault/Controllers/CiphersController.cs#L952) for example)

jaasen-livefront: Ready for re-review. ;)

PR #5906, opened 2025-06-02
Km/db signing keys

quexten:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Entities/UserSignatureKeyPair.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.ComponentModel.DataAnnotations;
```
withinfocus: ❓ Should more of these go under code ownership?

src/Core/KeyManagement/Models/Data/SignatureKeyPairData.cs
```diff
@@ -0,0 +1,11 @@
+﻿#nullable enable
+using Bit.Core.Enums;
+
+namespace Bit.Core.KeyManagement.Models.Data;
+
+public class SignatureKeyPairData
+{
+    required public SignatureAlgorithm SignatureAlgorithm { get; set; }
```
withinfocus: ⛏️ I think we usually swap `required` and `public`.

src/Core/KeyManagement/Models/Data/SignatureKeyPairData.cs
```diff
@@ -0,0 +1,11 @@
+﻿#nullable enable
```
withinfocus, line 1: ⛏️ We often have one empty line after the nullable enablement.

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
```diff
@@ -108,6 +109,7 @@ protected override void OnModelCreating(ModelBuilder builder)
         var eSsoConfig = builder.Entity<SsoConfig>();
         var eTaxRate = builder.Entity<TaxRate>();
         var eUser = builder.Entity<User>();
+        var eUserSignatureKeyPair = builder.Entity<UserSignatureKeyPair>();
```
withinfocus: 🎨 We have been trying lately to move these customizations and defaults out to specific entity type config classes.

src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_ReadByUserId.sql
```diff
@@ -0,0 +1,8 @@
+CREATE PROCEDURE [dbo].[UserSignatureKeyPair_ReadByUserId]
+    @UserId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SELECT *
+    FROM [dbo].[UserSignatureKeyPair]
```
withinfocus: 🎨 You'll want to create a view for this new table and select from that.

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
```diff
@@ -0,0 +1,11 @@
+﻿CREATE TABLE [dbo].[UserSignatureKeyPair] (
+    [Id]                        UNIQUEIDENTIFIER NOT NULL,
```
withinfocus: ❓ Does this table really even need an ID as its primary key? Via the logic I see here it's unique per user ID.

withinfocus (changes requested):
Really just the usage of a new view is what sticks out to me. @rkac-bw should also add his thoughts.

I'm assuming the usage of the new column on users will come at a later date.

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
quexten: I'll try to explain the reasoning behind this, and present some alternatives:

Currently we only have one signature keypair per user. However, eventually (either as a manual operation, or possibly for migration to post-quantum signature schemes), we would have a transition phase, where both keys, and a signed message attesting the upgrade between the two would need to be present. 

So, we could defer this problem to that time, and drop the id column for now. In that case, we could do one of:
1. Create an id column, and write a migration backfilling it for all old keys
2. Save / move the old signature keypair (or actually just the verifying key) to the "signature keypair rotation message" (either into it, or "next to" it). 

The second option might be the best, allowing us not to represent the complexity of managing multiple key versions in this table at all.

Thoughts?

src/Core/KeyManagement/Models/Data/SignatureKeyPairData.cs
quexten: Fixed in https://github.com/bitwarden/server/pull/5906/commits/9e0cdf3383ce93dc86bc93b6a3925ce4b04a5097

src/Core/KeyManagement/Models/Data/SignatureKeyPairData.cs
quexten: Fixed in https://github.com/bitwarden/server/pull/5906/commits/9e0cdf3383ce93dc86bc93b6a3925ce4b04a5097

quexten:
> I'm assuming the usage of the new column on users will come at a later date.

Yes, the follow-up PR (not yet opened) addresses that and make use of it.

src/Core/Entities/UserSignatureKeyPair.cs
quexten: Moved this to KM ownership

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
quexten: Moved this out.

src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_ReadByUserId.sql
quexten: Done.

quexten:
> Really just the usage of a new view is what sticks out to me. @rkac-bw should also add his thoughts.

I'll request review in that case so we make sure we don't forget

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
rkac-bw: I think it would be better to have 1:1 relation

Cleaner data model for now
Better represents meaning of key rotation with addition of rotation history to separate table when needed
Easier to query rotation history
Allows storing rotation-specific metadata


```
CREATE TABLE [dbo].[UserSignatureKeyPair] (
    [UserId] UNIQUEIDENTIFIER NOT NULL,
    [SignatureAlgorithm] TINYINT NOT NULL,
    [SigningKey] VARCHAR(MAX) NOT NULL,
    [VerifyingKey] VARCHAR(MAX) NOT NULL,
    [CreationDate] DATETIME2(7) NOT NULL,
    [RevisionDate] DATETIME2(7) NOT NULL,
    CONSTRAINT [PK_UserSignatureKeyPair] PRIMARY KEY ([UserId]),
    CONSTRAINT [FK_UserSignatureKeyPair_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id])
);
```
Then add a rotation table when rotation is added

```
CREATE TABLE [dbo].[UserSignatureKeyPairHistory] (
    [Id] UNIQUEIDENTIFIER NOT NULL,
    [UserId] UNIQUEIDENTIFIER NOT NULL,
    [OldSignatureAlgorithm] TINYINT NOT NULL,
    [NewSignatureAlgorithm] TINYINT NOT NULL,
    [OldVerifyingKey] VARCHAR(MAX) NOT NULL,
    [NewVerifyingKey] VARCHAR(MAX) NOT NULL,
    [RotationSignature] VARCHAR(MAX) NOT NULL,
    [RotationDate] DATETIME2(7) NOT NULL,
    CONSTRAINT [PK_UserSignatureKeyPairRotation] PRIMARY KEY ([Id]),
    CONSTRAINT [FK_UserSignatureKeyPairRotation_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id])
);
```

rkac-bw: In regards to having a transition where two keys exist it can get messy in one table so may need a transition table...
Thinking about that

rkac-bw: For future extensibility the composite key looks better but managing state in one table can be non-trivial over time.
Using a history approach seems to reduce  complexity. The history table becomes your single source of truth for "what key was valid when" and the main table is just a performance optimization for current state.

You could manage transitions with metadata regarding current algorithm and transition state and handle logic for handling multiple states through history possibly

```
CREATE TABLE [dbo].[UserSignatureKeyPairHistory] (
    [Id]                        UNIQUEIDENTIFIER NOT NULL,
    [UserId]                    UNIQUEIDENTIFIER NOT NULL,
    [SignatureKeyPairAlgorithm] TINYINT NOT NULL,
    [VerifyingKey]              VARCHAR(MAX) NOT NULL, -- Keep for signature verification
    [ValidFrom]                 DATETIME2 (7) NOT NULL,
    [ValidTo]                   DATETIME2 (7) NOT NULL,
    [RotationReason]            NVARCHAR(100) NULL,
    [TransitionMessage]         VARCHAR(MAX) NULL, -- Signed attestation if needed
    [CreationDate]              DATETIME2 (7) NOT NULL,
    CONSTRAINT [PK_UserSignatureKeyPairHistory] PRIMARY KEY CLUSTERED ([Id] ASC),
    CONSTRAINT [FK_UserSignatureKeyPairHistory_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id])
);
```
@withinfocus thoughts?

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
```diff
@@ -11,12 +11,20 @@ CREATE TABLE [dbo].[UserSignatureKeyPair] (
 );
 GO
 
+CREATE VIEW [dbo].[UserSignatureKeyPairView]
```
withinfocus: 🎨 To be backwards-compatible drop the view if it exists and then add it.

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
withinfocus: This conversation is actually selling me on keeping the `Id` column as-is, with the primary reason being that it's "normal" or "standard" in our relational design. If additional keys were needing to be stored, this could be as simple as a foreign key reference to another row in this table e.g. `PrecedingId` or somesuch, and queries could be adjusted to ask for the one where that's null (meaning it's the most recent). While @rkac-bw is bringing up some valuable datapoints with his history table, I don't think we need any of that in this use case -- it must be one of a line of rotations.

withinfocus (changes requested):
Migration script needing adjustment is all I am looking for now.

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
quexten: Ok, I'll keep the key as is for now, and re-consider the proposed methods here for rotating keys when we actually get to implementing that.

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
quexten: https://github.com/bitwarden/server/pull/5906/commits/fa2e314c4fe98126996edf6fdae56c83d8f270b3

rkac-bw: Our pattern for this has been to use CREATE OR ALTER VIEW , avoids an extra query

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
rkac-bw: if we use the same table eventually then we would require a preceding id column:
```
CREATE TABLE [dbo].[UserSignatureKeyPair] (
    [Id] UNIQUEIDENTIFIER NOT NULL PRIMARY KEY,
    [UserId] UNIQUEIDENTIFIER NOT NULL,
    [SignatureAlgorithm] TINYINT NOT NULL,
    [SigningKey] VARCHAR(MAX) NOT NULL,
    [VerifyingKey] VARCHAR(MAX) NOT NULL,
    [PrecedingId] UNIQUEIDENTIFIER NULL, -- Points to the previous key in the chain
    [CreationDate] DATETIME2(7) NOT NULL,
    CONSTRAINT [FK_UserSignatureKeyPair_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id]),
    CONSTRAINT [FK_UserSignatureKeyPair_Preceding] FOREIGN KEY ([PrecedingId]) REFERENCES [dbo].[UserSignatureKeyPair] ([Id])
);
```
Getting the current record for user would simply be:
`SELECT * FROM UserSignatureKeyPair WHERE UserId = @userId AND PrecedingId IS NULL;`

And would perform well with a filtered index:

```
CREATE NONCLUSTERED INDEX IX_UserSignatureKeyPair_LatestOnly
ON UserSignatureKeyPair (UserId)
WHERE PrecedingId IS NULL;
```

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
quexten: Fixed

src/Core/KeyManagement/Entities/UserSignatureKeyPair.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Bit.Core.KeyManagement.Entities;
+
+public class UserSignatureKeyPair : ITableObject<Guid>, IRevisable
+{
+    public Guid Id { get; set; }
+    public Guid UserId { get; set; }
+    public SignatureAlgorithm SignatureAlgorithm { get; set; }
+
+    [MaxLength(500)]
+    required public string VerifyingKey { get; set; }
```
Thomas-Avery, line 15:
```suggestion
    public required string VerifyingKey { get; set; }
```

:pick: normally public goes first.

src/Core/KeyManagement/Entities/UserSignatureKeyPair.cs
```diff
@@ -0,0 +1,28 @@
...
+    [MaxLength(500)]
+    required public string SigningKey { get; set; }
```
Thomas-Avery, line 16:
```suggestion
    public required string SigningKey { get; set; }
```

src/Core/KeyManagement/Models/Data/SignatureKeyPairData.cs
```diff
@@ -0,0 +1,22 @@
...
+
+public class SignatureKeyPairData
+{
+    public required SignatureAlgorithm SignatureAlgorithm { get; set; }
+    public required string WrappedSigningKey { get; set; }
+    public required string VerifyingKey { get; set; }
+
+    [JsonConstructor]
+    [System.Diagnostics.CodeAnalysis.SetsRequiredMembersAttribute]
+    public SignatureKeyPairData(SignatureAlgorithm signatureAlgorithm, string wrappedSigningKey, string verifyingKey)
```
Thomas-Avery, line 16: I didn't know `System.Text.Json` didn't respect nullable annotations by [default](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/nullable-annotations). Great catch, it might be worth adding some unit tests here so we don't regress in behavior.

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
withinfocus: Aren't there compatibility issues with that?

withinfocus (dismissed):
Pending my one comment on compatibility, but if this runs for us then so be it.

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
rkac-bw: The CREATE OR ALTER VIEW syntax was added in SQL server 2016, we should be good

src/Infrastructure.EntityFramework/KeyManagement/Repositories/UserSigningKeysRepository.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public UserSignatureKeyPairRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper) : base(serviceScopeFactory, mapper, context => context.UserSignatureKeyPair)
+    {
+    }
+
+    public async Task<SignatureKeyPairData?> GetByUserIdAsync(Guid userId)
+    {
+        await using var scope = ServiceScopeFactory.CreateAsyncScope();
+        var dbContext = GetDatabaseContext(scope);
+        var signingKeys = await dbContext.UserSignatureKeyPair.FindAsync(userId);
```
Thomas-Avery:
```suggestion
       var signingKeys = await dbContext.UserSignatureKeyPair.Where(kp => kp.UserId == userId).FirstOrDefaultAsync();
```

`FindAsync` searches the primary key for the ID provided.

src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_SetForRotation.sql
```diff
@@ -0,0 +1,15 @@
+CREATE PROCEDURE [dbo].[UserSignatureKeyPair_UpdateForRotation]
```
Thomas-Avery: The name of this file and the contents of this file don't match. 

`[UserSignatureKeyPair_UpdateForRotation]` code
`src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_SetForRotation.sql` file name.

src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_UpdateForRotation.sql
```diff
@@ -0,0 +1,13 @@
+CREATE PROCEDURE [dbo].[UserSignatureKeyPair_SetForRotation]
```
Thomas-Avery: The name of this file and the contents of this file don't match. 

`[dbo].[UserSignatureKeyPair_SetForRotation]` code
`src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_UpdateForRotation.sql` file name.

src/Infrastructure.EntityFramework/KeyManagement/Repositories/UserSigningKeysRepository.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    public UpdateEncryptedDataForKeyRotation SetUserSignatureKeyPair(Guid userId, SignatureKeyPairData signingKeys)
+    {
+        return async (_, _) =>
+        {
+            await using var scope = ServiceScopeFactory.CreateAsyncScope();
+            var dbContext = GetDatabaseContext(scope);
+            var entity = new Models.UserSignatureKeyPair
+            {
+                Id = Guid.NewGuid(),
```
Thomas-Avery: ⚠️  We want to call `UserSignatureKeyPair.SetNewId();` when generating GUIDs for the database.

This makes the GUIDs sequential more context on the `GenerateComb`.

src/Infrastructure.Dapper/KeyManagement/Repositories/UserSignatureKeyPairRepository.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    public UpdateEncryptedDataForKeyRotation SetUserSignatureKeyPair(Guid userId, SignatureKeyPairData signingKeys)
+    {
+        return async (SqlConnection connection, SqlTransaction transaction) =>
+        {
+            await connection.QueryAsync(
+                "[dbo].[UserSignatureKeyPair_SetForRotation]",
+                new
+                {
+                    Id = Guid.NewGuid(),
```
Thomas-Avery: ⚠️  We want to call `UserSignatureKeyPair.SetNewId();` when generating GUIDs for the database.

This makes the GUIDs sequential more context on the `GenerateComb`.

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
```diff
@@ -0,0 +1,320 @@
...
+    [UserId]                    UNIQUEIDENTIFIER NOT NULL,
+    [SignatureKeyPairAlgorithm] TINYINT NOT NULL,
+    [SigningKey]                VARCHAR(MAX) NOT NULL,
+    [VerifyingKey]              VARCHAR(MAX) NOT NULL,
+    [CreationDate]              DATETIME2 (7) NOT NULL,
+    [RevisionDate]              DATETIME2 (7) NOT NULL,
+    CONSTRAINT [PK_UserSignatureKeyPair] PRIMARY KEY CLUSTERED ([Id] ASC),
+    CONSTRAINT [FK_UserSignatureKeyPair_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id])
+);
+GO
```
Thomas-Avery:
```suggestion
IF OBJECT_ID('[dbo].[UserSignatureKeyPair]') IS NULL
BEGIN
    CREATE TABLE [dbo].[UserSignatureKeyPair]
    (
        [Id] UNIQUEIDENTIFIER NOT NULL,
        [UserId] UNIQUEIDENTIFIER NOT NULL,
        [SignatureKeyPairAlgorithm] TINYINT NOT NULL,
        [SigningKey] VARCHAR(MAX) NOT NULL,
        [VerifyingKey] VARCHAR(MAX) NOT NULL,
        [CreationDate] DATETIME2 (7) NOT NULL,
        [RevisionDate] DATETIME2 (7) NOT NULL,
        CONSTRAINT [PK_UserSignatureKeyPair] PRIMARY KEY CLUSTERED ([Id] ASC),
        CONSTRAINT [FK_UserSignatureKeyPair_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id])
    );
END
GO
```

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
```diff
@@ -0,0 +1,320 @@
...
+
+CREATE OR ALTER VIEW [dbo].[UserSignatureKeyPairView]
+AS
+SELECT
+    *
+FROM
+    [dbo].[UserSignatureKeyPair]
+GO
+
+CREATE PROCEDURE [dbo].[UserSignatureKeyPair_ReadByUserId]
```
Thomas-Avery:
```suggestion
CREATE OR ALTER PROCEDURE [dbo].[UserSignatureKeyPair_ReadByUserId]
```

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
```diff
@@ -0,0 +1,320 @@
...
+    @UserId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SELECT *
+    FROM [dbo].[UserSignatureKeyPairView]
+    WHERE [UserId] = @UserId;
+END
+GO
+
+CREATE PROCEDURE [dbo].[UserSignatureKeyPair_UpdateForRotation]
```
Thomas-Avery:
```suggestion
CREATE OR ALTER PROCEDURE [dbo].[UserSignatureKeyPair_UpdateForRotation]
```

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
```diff
@@ -0,0 +1,320 @@
...
+    UPDATE [dbo].[UserSignatureKeyPair]
+    SET [SignatureKeyPairAlgorithm] = @SignatureKeyPairAlgorithm,
+        [SigningKey] = @SigningKey,
+        [VerifyingKey] = @VerifyingKey,
+        [RevisionDate] = @RevisionDate
+    WHERE [UserId] = @UserId;
+END
+GO
+
+CREATE PROCEDURE [dbo].[UserSignatureKeyPair_SetForRotation]
```
Thomas-Avery:
```suggestion
CREATE OR ALTER PROCEDURE [dbo].[UserSignatureKeyPair_SetForRotation]
```

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
Thomas-Avery, line 1: Now that https://github.com/bitwarden/server/pull/5923/files is merged in we should move all these new SQL files under our team ownership.

`src/Sql/dbo/KeyManagement/*`

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
```diff
@@ -0,0 +1,11 @@
...
+    [UserId]                    UNIQUEIDENTIFIER NOT NULL,
+    [SignatureKeyPairAlgorithm] TINYINT NOT NULL,
+    [SigningKey]                VARCHAR(MAX) NOT NULL,
+    [VerifyingKey]              VARCHAR(MAX) NOT NULL,
+    [CreationDate]              DATETIME2 (7) NOT NULL,
+    [RevisionDate]              DATETIME2 (7) NOT NULL,
+    CONSTRAINT [PK_UserSignatureKeyPair] PRIMARY KEY CLUSTERED ([Id] ASC),
+    CONSTRAINT [FK_UserSignatureKeyPair_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id]),
+);
```
Thomas-Avery:
```
GO
CREATE NONCLUSTERED INDEX [IX_UserSignatureKeyPair_UserId]
    ON [dbo].[UserSignatureKeyPair]([UserId] ASC);

```
💭 

If we will be searching by UserID it might make sense to have an index for that column. 

Entity Framework Core will be automatically generating these indexes for us on the other DBs.

I'll leave the final decision to the DB team on this.

src/Infrastructure.EntityFramework/Configurations/UserSignatureKeyPairEntityTypeConfiguration.cs
Thomas-Avery, line 1: Can we move this configuration to our ownership? 

`src/Infrastructure.EntityFramework/KeyManagement/Configurations`

src/Infrastructure.EntityFramework/Models/UserSigningKeys.cs
Thomas-Avery, line 1: Can we move this model to our ownership? 

`src/Infrastructure.EntityFramework/KeyManagement/Models/`

src/Core/Enums/SignatureAlgorithm.cs
Thomas-Avery, line 1: Can we move this to KM ownership? `src/Core/KeyManagement/Enums`

Thomas-Avery (changes requested):
Thank you for splitting this up for easier reviewing. A few things to check out from my perspective.

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
rkac-bw: If it stays composite key as pk, then yes index on userid would be good

src/Infrastructure.EntityFramework/KeyManagement/Repositories/UserSigningKeysRepository.cs
quexten: Fixed

src/Infrastructure.Dapper/KeyManagement/Repositories/UserSignatureKeyPairRepository.cs
quexten: Fixed

quexten:
Apologies for the force-push! I somehow accidentally merged the follow-up PR and had to undo it.

src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_SetForRotation.sql
quexten: Seems these files were swapped; thank you for catching.

quexten: Fixed

src/Sql/dbo/Stored Procedures/UserSignatureKeyPair_UpdateForRotation.sql
quexten: Fixed

src/Core/KeyManagement/Models/Data/SignatureKeyPairData.cs
quexten: This is actually already covered as part of other unit tests/integration tests, which is how I found out about it, so regression should be covered.

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
quexten: Done

src/Sql/dbo/Tables/UserSignatureKeyPair.sql
quexten: Added the index, great callout!

src/Infrastructure.EntityFramework/KeyManagement/Repositories/UserSigningKeysRepository.cs
```diff
@@ -0,0 +1,74 @@
...
+using Bit.Core.Utilities;
+using Bit.Infrastructure.EntityFramework.Repositories;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Bit.Infrastructure.EntityFramework.KeyManagement.Repositories;
+
+public class UserSignatureKeyPairRepository : Repository<Core.KeyManagement.Entities.UserSignatureKeyPair, Models.UserSignatureKeyPair, Guid>, IUserSignatureKeyPairRepository
+{
+    public UserSignatureKeyPairRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper) : base(serviceScopeFactory, mapper, context => context.UserSignatureKeyPair)
```
rkac-bw: Should file name match repository name?
UserSigningKeysRepository.cs
repository: UserSignatureKeyPairRepository

src/Core/KeyManagement/Entities/UserSignatureKeyPair.cs
```diff
@@ -0,0 +1,28 @@
...
+
+public class UserSignatureKeyPair : ITableObject<Guid>, IRevisable
+{
+    public Guid Id { get; set; }
+    public Guid UserId { get; set; }
+    public SignatureAlgorithm SignatureAlgorithm { get; set; }
+
+    [MaxLength(500)]
+    required public string VerifyingKey { get; set; }
+    [MaxLength(500)]
```
rkac-bw: Maxlength 500 here but table defined as varchar(max) in tsql script

quexten: ~~Hmm, is that actually an issue? I think this should stay (max) in the DB. Whenever we take on post-quantum keys, these will be significantly larger (1-2 KiB for ML-DSA). For now, I thought it was reasonable to limit users from setting larger strings for now, since we only have tiny keys, but not limit the database so that we don't have to write a migration when that forseeable time comes.~~

Oh, i realized this is not the request model, apologies, i'll remove the limit here.

rkac-bw (approved):
lgtm

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
```diff
@@ -71,6 +71,7 @@ public DatabaseContext(DbContextOptions<DatabaseContext> options)
     public DbSet<TaxRate> TaxRates { get; set; }
     public DbSet<Transaction> Transactions { get; set; }
     public DbSet<User> Users { get; set; }
+    public DbSet<UserSignatureKeyPair> UserSignatureKeyPair { get; set; }
```
justindbaur:
```suggestion
    public DbSet<UserSignatureKeyPair> UserSignatureKeyPairs { get; set; }
```

nit: The property name should be plural but this shouldn't require an update to the migration since you do `builder.ToTable(nameof(UserSignatureKeyPair));` elsewhere so you can do that later if you'd like.

util/Migrator/DbScripts/2025-06-01_00_AddSignatureKeyPairTable.sql
Thomas-Avery: @quexten Looks like most of these comments on the migration script aren't currently applied. 

Maybe got wiped out in the force push?

quexten: I'm not entirely sure what happened here, but they have been re-applied.

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
quexten: Addressed this in [503e2b3](https://github.com/bitwarden/server/pull/5906/commits/503e2b36aced25559a8af185b3f62ab1305350a1)

src/Infrastructure.EntityFramework/KeyManagement/Repositories/UserSignatureKeyPairRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Bit.Infrastructure.EntityFramework.KeyManagement.Repositories;
+
+public class UserSignatureKeyPairRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper) : Repository<Core.KeyManagement.Entities.UserSignatureKeyPair, Models.UserSignatureKeyPair, Guid>(serviceScopeFactory, mapper, context => context.UserSignatureKeyPairs), IUserSignatureKeyPairRepository
+{
+    public async Task<SignatureKeyPairData?> GetByUserIdAsync(Guid userId)
+    {
+        await using var scope = ServiceScopeFactory.CreateAsyncScope();
+        var dbContext = GetDatabaseContext(scope);
+        var signingKeys = await dbContext.UserSignatureKeyPairs.FindAsync(userId);
```
Thomas-Avery:
```suggestion
        var signingKeys = await dbContext.UserSignatureKeyPairs.FirstOrDefaultAsync(x => x.UserId == userId);
```

quexten: Note to myself: When breaking down PR's into a chain make sure fixes required for the first PR are carried over from follow-up PR's. I had fixed this in one of the follow-ups but had not upstreamed it to this PR.. Thank you for catching.

Thomas-Avery (approved):
LGTM

withinfocus (approved):

PR #5907, opened 2025-06-02
[PM-21031] Optimize GET Members endpoint performance

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21031

## 📔 Objective

This PR improves the performance of the GET `/organizations/{orgId}/users` endpoint behind the `MembersGetEndpointOptimization` feature flag. When enabled, it significantly reduces DB round trips and improves query times.

This is the benchmark result from running `OrganizationUsersControllerPerformanceTest`, comparing the current (unoptimized) version with the new (optimized) version using the feature flag:
- Seed: 60000 seats
- Unoptimized: 90474 ms
- Optimized:   19745 ms
- Improvement: 78.2%

## Summary of Changes
- Optimized Endpoint Methods
    - `Get_vNext()` and `GetAccountRecoveryEnrolledUsers_vNext()`  
        - Both use `Task.WhenAll()` to run multiple DB queries in parallel  
        - Speeds up retrieval of 2FA and claimed domain status

- Repository Enhancements
    - `GetManyDetailsByOrganizationAsync_vNext()`  
        - Uses `OrganizationUserUserDetails_ReadByOrganizationId_V2`  
        - Returns users, groups, and collections in a single DB call
    - `GetManyByOrganizationWithClaimedDomainsAsync_vNext()`  
        - Uses `OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2`  
        - Leverages CTE for more efficient domain matching

- Performance Gains  
    - Parallel query execution reduces latency  
    - Consolidated DB calls improve scalability  
    - Noticeable speed-up for orgs with large user counts

- Integration Tests  
  - Added tests in `OrganizationUserRepositoryTests.cs`  
  - Ensure vNext methods preserve existing behavior while improving performance

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/bitwarden_license/src/Scim/Models/ScimUserRequestModel.cs
+++ b/bitwarden_license/src/Scim/Models/ScimUserRequestModel.cs
@@ -6,9 +6,9 @@ using Bit.Core.AdminConsole.Models.Business;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Models;
 using Bit.Core.Enums;
 using Bit.Core.Exceptions;
+using Bit.Core.Models.Business;
 using Bit.Core.Models.Data;
 using Bit.Core.Utilities;
-using OrganizationUserInvite = Bit.Core.Models.Business.OrganizationUserInvite;
 
 namespace Bit.Scim.Models;
 
@@ -47,7 +47,7 @@ public class ScimUserRequestModel : BaseScimUserModel
         return new InviteOrganizationUsersRequest(
             invites:
             [
-                new Core.AdminConsole.OrganizationFeatures.OrganizationUsers.InviteUsers.Models.OrganizationUserInvite(
+                new OrganizationUserInviteCommandModel(
                         email: email,
                         externalId: ExternalIdForInvite()
                     )
--- a/src/Api/AdminConsole/Models/Request/Organizations/OrganizationIntegrationConfigurationRequestModel.cs
+++ b/src/Api/AdminConsole/Models/Request/Organizations/OrganizationIntegrationConfigurationRequestModel.cs
@@ -1,10 +1,10 @@
-﻿using System.ComponentModel.DataAnnotations;
+﻿#nullable enable
+
 using System.Text.Json;
 using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 using Bit.Core.Enums;
 
-#nullable enable
 
 namespace Bit.Api.AdminConsole.Models.Request.Organizations;
 
@@ -12,8 +12,7 @@ public class OrganizationIntegrationConfigurationRequestModel
 {
     public string? Configuration { get; set; }
 
-    [Required]
-    public EventType EventType { get; set; }
+    public EventType? EventType { get; set; }
 
     public string? Filters { get; set; }
 
--- a/src/Api/AdminConsole/Models/Response/Organizations/OrganizationIntegrationConfigurationResponseModel.cs
+++ b/src/Api/AdminConsole/Models/Response/Organizations/OrganizationIntegrationConfigurationResponseModel.cs
@@ -25,6 +25,6 @@ public class OrganizationIntegrationConfigurationResponseModel : ResponseModel
     public string? Configuration { get; set; }
     public string? Filters { get; set; }
     public DateTime CreationDate { get; set; }
-    public EventType EventType { get; set; }
+    public EventType? EventType { get; set; }
     public string? Template { get; set; }
 }
--- a/src/Api/AdminConsole/Public/Controllers/OrganizationController.cs
+++ b/src/Api/AdminConsole/Public/Controllers/OrganizationController.cs
@@ -4,8 +4,9 @@
 using System.Net;
 using Bit.Api.AdminConsole.Public.Models.Request;
 using Bit.Api.Models.Public.Response;
+using Bit.Core;
+using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
 using Bit.Core.Context;
-using Bit.Core.Enums;
 using Bit.Core.Exceptions;
 using Bit.Core.Services;
 using Bit.Core.Settings;
@@ -21,15 +22,21 @@ public class OrganizationController : Controller
     private readonly IOrganizationService _organizationService;
     private readonly ICurrentContext _currentContext;
     private readonly GlobalSettings _globalSettings;
+    private readonly IImportOrganizationUsersAndGroupsCommand _importOrganizationUsersAndGroupsCommand;
+    private readonly IFeatureService _featureService;
 
     public OrganizationController(
         IOrganizationService organizationService,
         ICurrentContext currentContext,
-        GlobalSettings globalSettings)
+        GlobalSettings globalSettings,
+        IImportOrganizationUsersAndGroupsCommand importOrganizationUsersAndGroupsCommand,
+        IFeatureService featureService)
     {
         _organizationService = organizationService;
         _currentContext = currentContext;
         _globalSettings = globalSettings;
+        _importOrganizationUsersAndGroupsCommand = importOrganizationUsersAndGroupsCommand;
+        _featureService = featureService;
     }
 
     /// <summary>
@@ -50,13 +57,26 @@ public class OrganizationController : Controller
             throw new BadRequestException("You cannot import this much data at once.");
         }
 
-        await _organizationService.ImportAsync(
-            _currentContext.OrganizationId.Value,
-            model.Groups.Select(g => g.ToImportedGroup(_currentContext.OrganizationId.Value)),
-            model.Members.Where(u => !u.Deleted).Select(u => u.ToImportedOrganizationUser()),
-            model.Members.Where(u => u.Deleted).Select(u => u.ExternalId),
-            model.OverwriteExisting.GetValueOrDefault(),
-            EventSystemUser.PublicApi);
+        if (_featureService.IsEnabled(FeatureFlagKeys.ImportAsyncRefactor))
+        {
+            await _importOrganizationUsersAndGroupsCommand.ImportAsync(
+                _currentContext.OrganizationId.Value,
+                model.Groups.Select(g => g.ToImportedGroup(_currentContext.OrganizationId.Value)),
+                model.Members.Where(u => !u.Deleted).Select(u => u.ToImportedOrganizationUser()),
+                model.Members.Where(u => u.Deleted).Select(u => u.ExternalId),
+                model.OverwriteExisting.GetValueOrDefault());
+        }
+        else
+        {
+            await _organizationService.ImportAsync(
+                _currentContext.OrganizationId.Value,
+                model.Groups.Select(g => g.ToImportedGroup(_currentContext.OrganizationId.Value)),
+                model.Members.Where(u => !u.Deleted).Select(u => u.ToImportedOrganizationUser()),
+                model.Members.Where(u => u.Deleted).Select(u => u.ExternalId),
+                model.OverwriteExisting.GetValueOrDefault(),
+                Core.Enums.EventSystemUser.PublicApi);
+        }
+
         return new OkResult();
     }
 }
--- a/src/Api/Billing/Controllers/OrganizationBillingController.cs
+++ b/src/Api/Billing/Controllers/OrganizationBillingController.cs
@@ -382,4 +382,35 @@ public class OrganizationBillingController(
 
         return TypedResults.Ok(response);
     }
+
+
+    [HttpPost("change-frequency")]
+    [SelfHosted(NotSelfHostedOnly = true)]
+    public async Task<IResult> ChangePlanSubscriptionFrequencyAsync(
+        [FromRoute] Guid organizationId,
+        [FromBody] ChangePlanFrequencyRequest request)
+    {
+        if (!await currentContext.EditSubscription(organizationId))
+        {
+            return Error.Unauthorized();
+        }
+
+        var organization = await organizationRepository.GetByIdAsync(organizationId);
+
+        if (organization == null)
+        {
+            return Error.NotFound();
+        }
+
+        if (organization.PlanType == request.NewPlanType)
+        {
+            return Error.BadRequest("Organization is already on the requested plan frequency.");
+        }
+
+        await organizationBillingService.UpdateSubscriptionPlanFrequency(
+            organization,
+            request.NewPlanType);
+
+        return TypedResults.Ok();
+    }
 }
--- a/src/Api/Billing/Models/Requests/ChangePlanFrequencyRequest.cs
+++ b/src/Api/Billing/Models/Requests/ChangePlanFrequencyRequest.cs
@@ -1,0 +1,10 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Core.Billing.Enums;
+
+namespace Bit.Api.Billing.Models.Requests;
+
+public class ChangePlanFrequencyRequest
+{
+    [Required]
+    public PlanType NewPlanType { get; set; }
+}
--- a/src/Api/Billing/Queries/Organizations/OrganizationWarningsQuery.cs
+++ b/src/Api/Billing/Queries/Organizations/OrganizationWarningsQuery.cs
@@ -12,6 +12,7 @@ using Bit.Core.Billing.Services;
 using Bit.Core.Context;
 using Bit.Core.Services;
 using Stripe;
+using static Bit.Core.Billing.Utilities;
 using FreeTrialWarning = Bit.Api.Billing.Models.Responses.Organizations.OrganizationWarningsResponse.FreeTrialWarning;
 using InactiveSubscriptionWarning =
     Bit.Api.Billing.Models.Responses.Organizations.OrganizationWarningsResponse.InactiveSubscriptionWarning;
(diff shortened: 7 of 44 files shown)


src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2.sql
```diff
@@ -0,0 +1,20 @@
+﻿CREATE PROCEDURE [dbo].[OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    WITH CTE_UserWithDomain AS (
+        SELECT
+            OU.*,
+            SUBSTRING(U.Email, CHARINDEX('@', U.Email) + 1, LEN(U.Email)) AS EmailDomain
```
jrmccannon: An additional idea might to be to make this a computed column so we could index against it. Check with @rkac-bw to see if that's necessary at all. We've been making improvements to sprocs for getting the domain out of the email, so maybe it's time to do it. 🤷

rkac-bw: So creating a computed column on 12 million rows might have an extended down time during migration, can we not just create a column and have a migration that populates it in batches to reduce locking and then parse in the code to insert domain name?

r-tome: That sounds like a great idea to me. I'll work on the batch script

rkac-bw: A second thought on this, @withinfocus suggested an indexed view and I tested it out and it is non blocking on creation:
```
CREATE VIEW dbo.UserEmailDomain
WITH SCHEMABINDING
AS
SELECT 
    Id,
    Email,
    SUBSTRING(Email, CHARINDEX('@', Email) + 1, LEN(Email)) AS EmailDomain
FROM dbo.[User]
WHERE Email IS NOT NULL 
    AND CHARINDEX('@', Email) > 0


CREATE UNIQUE CLUSTERED INDEX IX_UserEmailDomain_Id 
ON dbo.UserEmailDomain (Id);
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/GetOrganizationUsersClaimedStatusQuery.cs
```diff
@@ -27,7 +30,9 @@ public async Task<IDictionary<Guid, bool>> GetUsersOrganizationClaimedStatusAsyn
             if (organizationAbility is { Enabled: true, UseOrganizationDomains: true })
             {
                 // Get all organization users with claimed domains by the organization
-                var organizationUsersWithClaimedDomain = await _organizationUserRepository.GetManyByOrganizationWithClaimedDomainsAsync(organizationId);
+                var organizationUsersWithClaimedDomain = _featureService.IsEnabled(FeatureFlagKeys.MembersGetEndpointOptimization)
+                    ? await _organizationUserRepository.GetManyByOrganizationWithClaimedDomainsAsync_vNext(organizationId)
+                    : await _organizationUserRepository.GetManyByOrganizationWithClaimedDomainsAsync(organizationId);
```
withinfocus, line 35: 🌱 Run this in parallel and wait for them before L38?

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2.sql
withinfocus: We call these "details views" generally and you'll see some around. What's new for us here is being schema-bound, so we should think through how these are applied because the view will need to be dropped before column changes happen.

rkac-bw: One possibility is to switch to non indexed view for migration period so that functionality remains, but that can time out due to lack of index

-- Step 1: Drop the indexes
`DROP INDEX IX_UserEmailDomain_Id ON dbo.UserEmailDomain;`

-- Step 2: Drop and recreate view without SCHEMABINDING
```
DROP VIEW dbo.UserEmailDomain;

CREATE VIEW dbo.UserEmailDomain
AS  -- No WITH SCHEMABINDING
SELECT 
    Id,
    Email,
    SUBSTRING(Email, CHARINDEX('@', Email) + 1, LEN(Email)) AS EmailDomain
FROM dbo.[User]
WHERE Email IS NOT NULL 
    AND CHARINDEX('@', Email) > 0;
```

-- Step 3: Do your migration work on User table
`ALTER TABLE dbo.[User] ADD/DROP/MODIFY columns..`.

-- Step 4: Recreate with SCHEMABINDING and indexes
```
DROP VIEW dbo.UserEmailDomain;
CREATE VIEW dbo.UserEmailDomain WITH SCHEMABINDING AS...
CREATE UNIQUE CLUSTERED INDEX...
```

other wise we need to handle the lack of indexed view during migration in code

r-tome: That seems like a great idea to me. No downtime, right?
How do we make sure people follow these steps when making changes to the user table?

rkac-bw: So this seems like a table, with schemabinding, will be cumbersome going forward and have potential issues during migrations

We could just do this in sql 
add column
```
ALTER TABLE OrganizationUser
ADD Domain NVARCHAR(255) NULL;
```
add domain data for insert

```
CREATE PROCEDURE [dbo].[OrganizationUser_Insert]
    @Email NVARCHAR(255),
    <OtherColumns> NVARCHAR(MAX) -
AS
BEGIN
    SET NOCOUNT ON;

    INSERT INTO OrganizationUser (Email, Domain, <OtherColumns>)
    VALUES (
        @Email,
        CASE 
            WHEN @Email IS NOT NULL AND CHARINDEX('@', @Email) > 0 
            THEN SUBSTRING(@Email, CHARINDEX('@', @Email) + 1, LEN(@Email))
            ELSE NULL
        END, -- Extract domain or leave NULL
        @OtherColumn
    );
END;
```
add domain data for update
```
CREATE PROCEDURE [dbo].[OrganizationUser_Update]
    @Id UNIQUEIDENTIFIER,
    @Email NVARCHAR(255),
    @OtherColumn NVARCHAR(MAX) -- Add other columns as needed
AS
BEGIN
    SET NOCOUNT ON;

    UPDATE OrganizationUser
    SET 
        Email = @Email,
        Domain = CASE 
                    WHEN @Email IS NOT NULL AND CHARINDEX('@', @Email) > 0 
                    THEN SUBSTRING(@Email, CHARINDEX('@', @Email) + 1, LEN(@Email))
                    ELSE NULL
                 END, -- Extract domain or leave NULL
        OtherColumn = @OtherColumn
    WHERE Id = @Id;
END;
```
And during migration we can batch for back fill
```

DECLARE @BatchSize INT = 1000; -- Number of rows to process in each batch
DECLARE @RowsAffected INT = 1; -- Tracks how many rows were updated in each iteration

WHILE @RowsAffected > 0
BEGIN
    -- Update a batch of rows
    UPDATE TOP (@BatchSize) OrganizationUser
    SET Domain = SUBSTRING(Email, CHARINDEX('@', Email) + 1, LEN(Email))
    WHERE Email IS NOT NULL
        AND CHARINDEX('@', Email) > 0
        AND Domain IS NULL;

    -- Check how many rows were updated
    SET @RowsAffected = @@ROWCOUNT;

    -- Optionally, add a delay between batches to reduce load on the server
    -- WAITFOR DELAY '00:00:01'; -- 1-second delay
END;
```
@withinfocus thoughts?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/GetOrganizationUsersClaimedStatusQuery.cs
r-tome: I thought about it but the `organizationAbility` value comes from the cache, so it's quick to access. If `if (organizationAbility is { Enabled: true, UseOrganizationDomains: true })` returns `false` there's no need to query the database.

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2.sql
eliykat: A few thoughts re-reading the above thread:

1. using the CTE (the current approach in this PR) seems like the simplest option in terms of ongoing maintenance - are the alternatives guaranteed to be significantly faster? Or should we go with the simpler solution and monitor?

2. regarding the [view solution](https://github.com/bitwarden/server/pull/5907#discussion_r2126316413): dropping and recreating the view when the table is changed doesn't seem like a huge deal to me - we often have to refresh or update views when we update tables today, and we don't even use views very much (arguably not as much as we should). And the upside is that we don't have to duplicate data. That said, I take the point that it does make db changes harder.

3. regarding the proposal to [store the data on the OrganizationUser table](https://github.com/bitwarden/server/pull/5907#discussion_r2129632944): if we go this route I think it could be simpler - specifically a `Claimed` boolean column, which indicates whether that orgUser is claimed (managed) by their organization. This is all we are really trying to find out when comparing email domains. That would also simplify the sync logic, which today has a [separate call](https://github.com/bitwarden/server/blob/ca09a4e9f55ceafc897e1c9c75529ec27f12122a/src/Api/Vault/Controllers/SyncController.cs#L111-L113) to find claimed orgs, but with this solution it would get this info for free when reading the OrgUser details.

One drawback is maintaining the data in a correct state - the column would need to be updated:
- when an OrgUser is created
- when an OrgDomain is verified (not created)
- when an OrgDomain is deleted
- theoretically when a User changes their email address, but in reality you're restricted from doing this in a way that would "un-claim" your account, so this is fine

So we would have to ensure all code paths correctly update the corresponding OrgUsers, in both MSSQL and EF implementations. Our db entity models are also passed around pretty freely at the moment (i.e. callers can update entire rows) so anyone could overwrite this for a specific orgUser - although this is a risk with all our models. We can still do it if it makes sense, just be aware that it's also not free of technical debt or maintenance.

eliykat: @rkac-bw and @withinfocus - any further thoughts on our direction here?

withinfocus: After being out and re-reading the thread, the schema-bound view stands out to me as the cleanest with respect to data presence (and lack of duplication) as well as maintenance. Yes, there's a little more to do when working with one but we'll see impacts and bugs immediately with what we have in CI. Is there something I am missing besides what Thomas stated around its recreation?

I like where the third solution could be headed but that's a significant refactor the team would need to take on -- is that something you want to do?

r-tome: The consensus seems to be that the schema-bound view is the simplest and best approach for now.
@eliykat, would you be okay with me going ahead with this for now? We can revisit the idea of the third option as part of a future refactor if it makes sense then.

src/Sql/dbo/Views/UserEmailDomainView.sql
```diff
@@ -0,0 +1,14 @@
+CREATE VIEW [dbo].[UserEmailDomainView]
```
withinfocus, line 1: 💭 Should this use the same `...Details` naming we have for hybrid objects? I think it should, since it doesn't represent a single table. It's only used in a join though, at least for now, so it feels open for discussion.

util/Migrator/DbScripts/2025-06-17_00_OrgUsersQueryOptimizations.sql
```diff
@@ -0,0 +1,64 @@
...
+    Email,
+    SUBSTRING(Email, CHARINDEX('@', Email) + 1, LEN(Email)) AS EmailDomain
+FROM dbo.[User]
+WHERE Email IS NOT NULL 
+    AND CHARINDEX('@', Email) > 0
+GO
+
+CREATE UNIQUE CLUSTERED INDEX IX_UserEmailDomainView_Id 
+ON dbo.UserEmailDomainView (Id);
+GO
```
withinfocus: ❌ This will fail if the index already exists.

src/Sql/dbo/Views/UserEmailDomainView.sql
r-tome: This represents columns from just the `User` table (no JOINs), whereas other `Details` views always JOIN multiple tables together

withinfocus (commented):
@rkac-bw please give this a final review.

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2.sql
rkac-bw: I did some performance testing and found that rewriting query as CTE, using with recompile (due to  small number of large orgs skew) and three non clustered indexes that it out performs the indexed view

[Jira Task: PR-5907 Performance Analysis: OrganizationUser Domain Filtering](https://bitwarden.atlassian.net/browse/PM-23344 )

This will avoid schemabinding and deployment issues

rkac-bw: Unless @withinfocus or any one else has comments I think we should go ahead with rewriting query as CTE, using with recompile (due to small number of large orgs skew) and three non clustered indexes and the view without schema binding, let me now if you have questions

r-tome:
I have implemented the suggestions by @rkac-bw above in https://github.com/bitwarden/server/pull/5907#discussion_r2185690021

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/GetOrganizationUsersClaimedStatusQuery.cs
rkac-bw: Cache may have issues for that lookup https://bitwarden.atlassian.net/wiki/x/iYDibg

r-tome: Thats an issue on the cache mechanism, right? If so, that seems to be out of scope here.

jrmccannon (dismissed):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/GetOrganizationUsersClaimedStatusQuery.cs
withinfocus: https://bitwarden.atlassian.net/browse/PM-23845

util/Migrator/DbScripts/2025-06-17_00_OrgUsersQueryOptimizations.sql
```diff
@@ -0,0 +1,98 @@
...
+
+IF NOT EXISTS(SELECT name FROM sys.indexes WHERE name = 'IX_User_Id_EmailDomain')
+    BEGIN
+        CREATE NONCLUSTERED INDEX [IX_User_Id_EmailDomain]
+            ON [dbo].[User] ([Id], [Email])
+    END
+GO
+
+IF NOT EXISTS(SELECT name FROM sys.indexes WHERE name = 'IX_OrganizationDomain_Org_VerifiedDomain')
```
rkac-bw: Check uses wrong name (IX_OrganizationDomain_Org_VerifiedDomain) but creates
  IX_OrganizationDomain_OrganizationId_VerifiedDate
  - This will create duplicate indexes on every migration run

util/Migrator/DbScripts/2025-06-17_00_OrgUsersQueryOptimizations.sql
```diff
@@ -0,0 +1,98 @@
...
+    BEGIN
+        CREATE NONCLUSTERED INDEX [IX_OrganizationDomain_OrganizationId_VerifiedDate]
+            ON [dbo].[OrganizationDomain] ([OrganizationId], [VerifiedDate])
+            INCLUDE ([DomainName])
+            WHERE [VerifiedDate] IS NOT NULL
```
rkac-bw, line 34: No EF migration files?

util/Migrator/DbScripts/2025-06-17_00_OrgUsersQueryOptimizations.sql
r-tome: Nicely spotted! Its fixed

withinfocus (approved):
Just clearing my approval here.

PR #5908, opened 2025-06-02
[PM-22097] Add Columns to Collections for Org User Default Collection

jrmccannon:
## 🎟️ Tracking
[PM-22097](https://bitwarden.atlassian.net/browse/PM-22097)

## 📔 Objective
This adds two columns to the Collections table. `Type` and `UserDefaultCollectionEmail`.

```sql
    [UserDefaultCollectionEmail]    NVARCHAR(256)   NULL,
    [Type]    TINYINT   NOT NULL DEFAULT(0),
```

`Type` refers to `CollectionType`. This will be used to identify which collections are `Shared` (which is the current behavior for all collections) and `DefaultUserCollection` which will be true when an organization enables the Personal Vault policy. (`PolicyType.PersonalOwnership`).
```csharp
public enum CollectionType
{
    Shared,
    DefaultUserCollection,
}
```

Updates were made to allow for default values passed to stored procedures where they would be used.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-22097]: https://bitwarden.atlassian.net/browse/PM-22097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JimmyVo16 (approved):

src/Core/Enums/CollectionType.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Bit.Core.Enums;
+
+public enum CollectionType
+{
+    Shared,
+    DefaultUserCollection,
```
eliykat: We tend to expressly number our enums, which I also prefer. So that a simple change in order doesn't change values.

For consistency, I also suggest:
- `SharedCollection` rather than just `Shared`
- either `DefaultUserCollection` or `UserDefaultCollection` but not both - it is currently inconsistent between this and the db column. I think I was inconsistent when writing the tickets. `DefaultUserCollection` reads a little better to me but I don't feel strongly about it.

src/Sql/dbo/Stored Procedures/Collection_ReadByUserId.sql
```diff
@@ -13,7 +13,9 @@ BEGIN
         ExternalId,
         MIN([ReadOnly]) AS [ReadOnly],
         MIN([HidePasswords]) AS [HidePasswords],
-        MAX([Manage]) AS [Manage]
+        MAX([Manage]) AS [Manage],
+        UserDefaultCollectionEmail,
+        Type as [Type]
```
eliykat: Is the alias needed when it's the same?

I also recommend using the square brackets on all column names, e.g. `[Type]`, just to be safe. It doesn't appear to be a reserved word, but it is being syntax highlighted.

eliykat (changes requested):

eliykat (dismissed):

PR #5911, opened 2025-06-03
[PM-22344] - fix Error: Cannot Decrypt when moving a vault item to a collection

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22344

## 📔 Objective

This PR fixes an issue where the `/cipher/share` endpoint was causing the shared cipher's encryption to break. This was due to a change from [this PR](https://github.com/bitwarden/server/pull/5900/files?diff=split&w=0) that changed using `ToCipher` to casting with `Cipher` which caused a loss of the associated data and thus a failure of decryption in the client. It also returns as a `ListResponseModel` and updates the tests.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Vault/Controllers/CiphersController.cs
+++ b/src/Api/Vault/Controllers/CiphersController.cs
@@ -1086,7 +1086,7 @@ public class CiphersController : Controller
             }
         }
 
-        var shareCiphers = new List<(Cipher, DateTime?)>();
+        var shareCiphers = new List<(CipherDetails, DateTime?)>();
         foreach (var cipher in model.Ciphers)
         {
             if (!ciphersDict.TryGetValue(cipher.Id.Value, out var existingCipher))
@@ -1096,7 +1096,7 @@ public class CiphersController : Controller
 
             ValidateClientVersionForFido2CredentialSupport(existingCipher);
 
-            shareCiphers.Add((cipher.ToCipher(existingCipher), cipher.LastKnownRevisionDate));
+            shareCiphers.Add((cipher.ToCipherDetails(existingCipher), cipher.LastKnownRevisionDate));
         }
 
         var updated = await _cipherService.ShareManyAsync(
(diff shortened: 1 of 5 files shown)


gbubemismith (approved):

shane-melton:
@jaasen-livefront Switching to a `ListResponseModel` (which I'm in favor of) will also require changing the client service(s) to match the new response type.

shane-melton (changes requested):
We've got some failing tests we need to fix before merge

gbubemismith (approved):

src/Api/Vault/Controllers/CiphersController.cs
```diff
@@ -1106,7 +1106,8 @@ public async Task<CipherMiniResponseModel[]> PutShareMany([FromBody] CipherBulkS
             userId
         );
 
-        return updated.Select(c => new CipherMiniResponseModel(c, _globalSettings, false)).ToArray();
+        var response = updated.Select(c => new CipherMiniResponseModel(c, _globalSettings, false)).ToArray();
```
gbubemismith: We don't need the toArray anymore

jaasen-livefront: Fixed!

gbubemismith (dismissed):

src/Api/Vault/Controllers/CiphersController.cs
```diff
@@ -1106,7 +1106,8 @@ public async Task<CipherMiniResponseModel[]> PutShareMany([FromBody] CipherBulkS
             userId
         );
 
-        return updated.Select(c => new CipherMiniResponseModel(c, _globalSettings, false)).ToArray();
+        var response = updated.Select(c => new CipherMiniResponseModel(c, _globalSettings, false));
```
shane-melton: 💭  Using `false` here for `orgUseTotp` seems like it would break totp generation for ciphers after they're shared with an organization (at least until the vault re-syncs).

jaasen-livefront: @shane-melton Good point. I've pushed a fix. Looks like this is the only place we're calling `ShareManyAsync` so the change shouldn't have any side-effects elsewhere.

shane-melton: Sounds good, looks like we have some tests failing again though 😢

shane-melton (approved):

PR #5912, opened 2025-06-03
[PM-22380] Enable NRT for some Core project files

benbryant0:
## 🎟️ Tracking

Related to [this ADR](https://contributing.bitwarden.com/architecture/adr/csharp-nullable-reference-types/), but until that's accepted I'm working in the other direction and enabling per-file instead (as discussed [here](https://github.com/orgs/bitwarden/discussions/15033)).

## 📔 Objective

This PR enables NRT (nullable reference types) for 25 files in the Core project and does some minimal work to fix the resulting errors.

I've intentionally targeted files that don't have any ownership according to the [CODEOWNERS file](https://github.com/bitwarden/server/blob/6411cc63ca0af7a21386214eb9a8e31e8e219f44/.github/CODEOWNERS), but a file in a test project that had to be changed does belong to the Key Management team.

Since the scope of these changes is pretty limited, it's not guaranteed that every field/variable/whatever that can be null has been marked as such. I've only focused on fixing errors, and been guided by callers of functions & constructors when those were also in nullable contexts and started erroring.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/NotificationHub/NotificationHubConnection.cs
```diff
@@ -155,9 +158,10 @@ public static NotificationHubConnection From(GlobalSettings.NotificationHubSetti
         };
     }
 
+    [MemberNotNull(nameof(_hubClient))]
     private NotificationHubConnection Init()
     {
-        HubClient = NotificationHubClient.CreateClientFromConnectionString(ConnectionString, HubName, EnableSendTracing);
+        _hubClient = NotificationHubClient.CreateClientFromConnectionString(ConnectionString, HubName, EnableSendTracing);
```
benbryant0, line 164: This avoids needing to use a null-forgiving operator in the `HubClient` getter. The null analysis doesn't seem smart enough to work out that assigning to `HubClient` would just directly assign to `_hubClient`, so it has to be assigned to directly here. This makes the `HubClient` setter unused now, so that could be removed.

src/Core/Jobs/JobFactory.cs
```diff
@@ -16,7 +18,7 @@ public JobFactory(IServiceProvider container)
     public IJob NewJob(TriggerFiredBundle bundle, IScheduler scheduler)
     {
         var scope = _container.CreateScope();
-        return scope.ServiceProvider.GetService(bundle.JobDetail.JobType) as IJob;
+        return (scope.ServiceProvider.GetService(bundle.JobDetail.JobType) as IJob)!;
```
benbryant0, line 21: As I understand it this should only be null if a job type hasn't been registered in DI before being executed by Quartz.NET, which shouldn't happen and that'll just lead to Quartz throwing a bit after calling this anyway.

I haven't used Quartz before, but from looking into it a bit I think as of ~4 years ago this isn't necessary, as Quartz can handle all of this itself. It looks like there could be a pretty big cleanup of most of the use of Quartz, which would get rid of this, but that's a whole other thing.

src/Core/Jobs/BaseJobsHostedService.cs
```diff
@@ -142,7 +145,10 @@ public virtual async Task StartAsync(CancellationToken cancellationToken)
 
     public virtual async Task StopAsync(CancellationToken cancellationToken)
     {
-        await _scheduler?.Shutdown(cancellationToken);
```
benbryant0, line 145: If `_scheduler` is null, this just throws a nullref exception as you can't await null.

src/Core/Jobs/BaseJobsHostedService.cs
benbryant0, line 1: The changes here are mostly around `Jobs` potentially being null. It's a bit odd that there's some logic in `StartAsync` that only executes if `Jobs` is **not** null, but then right after that `Jobs` would be used assuming that it wasn't null. I just shuffled stuff around to make that not an issue.

quexten:
@benbryant0 I synced up the branch to main, to run the CI pipeline, please watch out to pull the changes if making further changes.

test/Api.Test/KeyManagement/Controllers/AccountsKeyManagementControllerTests.cs
```diff
@@ -175,7 +175,7 @@ public async Task RotateUserKeyWrongData_Throws(SutProvider<AccountsKeyManagemen
         }
         catch (BadRequestException ex)
         {
-            Assert.NotEmpty(ex.ModelState.Values);
+            Assert.NotEmpty(ex.ModelState!.Values);
```
justindbaur, line 178: In my opinion I would use an assertion that this thing is non null instead of the forgiving operator in tests.

```suggestion
            Assert.NotNull(ex.ModelState);
            Assert.NotEmpty(ex.ModelState.Values);
```

test/Api.Test/KeyManagement/Controllers/AccountsKeyManagementControllerTests.cs
```diff
@@ -210,7 +210,7 @@ public async Task PostSetKeyConnectorKeyAsync_SetKeyConnectorKeyFails_ThrowsBadR
         var badRequestException =
             await Assert.ThrowsAsync<BadRequestException>(() => sutProvider.Sut.PostSetKeyConnectorKeyAsync(data));
 
-        Assert.Equal(1, badRequestException.ModelState.ErrorCount);
+        Assert.Equal(1, badRequestException.ModelState!.ErrorCount);
```
justindbaur, line 213: Same here

test/Api.Test/KeyManagement/Controllers/AccountsKeyManagementControllerTests.cs
```diff
@@ -284,7 +284,7 @@ public async Task PostConvertToKeyConnectorAsync_ConvertToKeyConnectorFails_Thro
         var badRequestException =
             await Assert.ThrowsAsync<BadRequestException>(() => sutProvider.Sut.PostConvertToKeyConnectorAsync());
 
-        Assert.Equal(1, badRequestException.ModelState.ErrorCount);
+        Assert.Equal(1, badRequestException.ModelState!.ErrorCount);
```
justindbaur, line 287: And here

src/Core/Jobs/JobFactory.cs
justindbaur: This could definitely use some cleanup it just never gets prioritized since it keeps working 🤷.

benbryant0:
Thanks!

@quexten I'll leave the nits for now since you're fine with them and they're in test files with no change to existing behaviour. I do agree with them though, so I'll make a note to fix them when I next have a PR that touches KM ownership.

I just rebased on main, so should be good to go.

quexten:
Merged! Thank you for the contribution.

PR #5929, opened 2025-06-06
feat(change-password-component): Change Password Update [18720] - Now…

Patrick-Pimentel-Bitwarden:
… sending back accepted mp policies on base validator

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Vault/Controllers/SyncController.cs
+++ b/src/Api/Vault/Controllers/SyncController.cs
@@ -81,12 +81,15 @@ public class SyncController : Controller
             throw new BadRequestException("User not found.");
         }
 
-        var organizationUserDetails = await _organizationUserRepository.GetManyDetailsByUserAsync(user.Id,
+        var organizationUserDetails = await _organizationUserRepository.GetManyDetailsByUserAsync(
+            user.Id,
             OrganizationUserStatusType.Confirmed);
-        var providerUserDetails = await _providerUserRepository.GetManyDetailsByUserAsync(user.Id,
+        var providerUserDetails = await _providerUserRepository.GetManyDetailsByUserAsync(
+            user.Id,
             ProviderUserStatusType.Confirmed);
         var providerUserOrganizationDetails =
-            await _providerUserRepository.GetManyOrganizationDetailsByUserAsync(user.Id,
+            await _providerUserRepository.GetManyOrganizationDetailsByUserAsync(
+                user.Id,
                 ProviderUserStatusType.Confirmed);
         var hasEnabledOrgs = organizationUserDetails.Any(o => o.Enabled);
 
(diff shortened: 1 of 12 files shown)


src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -192,6 +198,15 @@ public async Task<PolicyResponseModel> GetMasterPasswordPolicy(Guid orgId)
         return new PolicyResponseModel(policy);
     }
 
+    [HttpGet("~/policies/master-password")]
+    [RequireFeature(FeatureFlagKeys.ChangeExistingPasswordRefactor)]
+    public async Task<MasterPasswordPolicyResponseModel> GetMasterPasswordPolicy()
+    {
+        var userId = _userService.GetProperUserId(User).Value;
+
+        return new MasterPasswordPolicyResponseModel(await _policyService.GetMasterPasswordPolicyForUserAsync(new Guid(userId.ToString()), true));
```
eliykat, line 207: Why cast to a string and then back to a Guid? `userId` should already be a Guid.

eliykat (commented):
I actually think we can solve this for you in https://github.com/bitwarden/server/pull/5936. We would keep your endpoint, but it would call the new `PolicyRequirementQuery` which we can configure to include accepted and confirmed users, rather than implementing a bespoke repository query. I'll look at it all together next week and get back to you.

BTreston:
> I actually think we can solve this for you in #5936. We would keep your endpoint, but it would call the new `PolicyRequirementQuery` which we can configure to include accepted and confirmed users, rather than implementing a bespoke repository query. (No documentation around this query yet but I'll be sharing some soon.) I'll look at it all together next week and get back to you, I've also asked @BTreston to take a look.

@eliykat I think its possible to do it with a new `PolicyRequirement`, but it would mean we have 2 `PolicyRequirements` associated with 1 policy. If we're cool with that then 👍. Either that or we extend the capabilities of `PolicyRequirementQuery` to be able to specify somehow which factory we want to use and extend the `MasterPasswordPolicyRequirementFactory` with an override for `ExemptStatuses` to exempt the ones we don't care about (`Invited` and `Revoked`)? I am not sure if we are currently able to satisfy 2 different `ExemptSatuses` requirements in a single `PolicyRequirementFactory` (unless I am missing something) as that seems to be the requirement this functionality would support. Might require some thought on the approach and implementation, but regardless I think its possible.

src/Api/AdminConsole/Controllers/PoliciesController.cs
Patrick-Pimentel-Bitwarden: Yeah sorry, that was leftover from previous changes. You are correct.

Patrick-Pimentel-Bitwarden:
Just a heads up regarding this work. We have moved away from the server changes in favor of fixing the login flow in the clients. This pr will be closed by next week once we have all the other work in and confirmed to be working.

eliykat:
OK - we'll hold off on doing anything here for now. Thanks for taking a look @BTreston .

PR #5937, opened 2025-06-07
Nrt/core part2

benbryant0:
## 🎟️ Tracking

Related to [this ADR](https://contributing.bitwarden.com/architecture/adr/csharp-nullable-reference-types/), but until that's accepted I'm working in the other direction and enabling per-file instead (as discussed [here](https://github.com/orgs/bitwarden/discussions/15033)).

## 📔 Objective

nullabubblelubble

This PR enables NRT (nullable reference types) for almost all remaining files in Core/Utilities and Core/SecretsManager (all the files that didn't result in changes matching anyone in [CODEOWNERS](https://github.com/bitwarden/server/blob/6411cc63ca0af7a21386214eb9a8e31e8e219f44/.github/CODEOWNERS))

I've split the changes into commits based on the folders. The only changes that required more than just annotations were a few changes in Core/Utilities, so I broke those out into another commit and I'll dump some notes/explanations as another comment.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Utilities/BitPayClient.cs
+++ b/src/Core/Utilities/BitPayClient.cs
@@ -2,9 +2,11 @@
 
 namespace Bit.Core.Utilities;
 
+#nullable enable
+
 public class BitPayClient
 {
-    private readonly BitPayLight.BitPay _bpClient;
+    private readonly BitPayLight.BitPay? _bpClient;
 
     public BitPayClient(GlobalSettings globalSettings)
     {
@@ -15,13 +17,21 @@ public class BitPayClient
         }
     }
 
-    public Task<BitPayLight.Models.Invoice.Invoice> GetInvoiceAsync(string id)
+    public async Task<BitPayLight.Models.Invoice.Invoice?> GetInvoiceAsync(string id)
     {
-        return _bpClient.GetInvoice(id);
+        if (_bpClient is null)
+        {
+            return null;
+        }
+        return await _bpClient.GetInvoice(id);
     }
 
-    public Task<BitPayLight.Models.Invoice.Invoice> CreateInvoiceAsync(BitPayLight.Models.Invoice.Invoice invoice)
+    public async Task<BitPayLight.Models.Invoice.Invoice?> CreateInvoiceAsync(BitPayLight.Models.Invoice.Invoice invoice)
     {
-        return _bpClient.CreateInvoice(invoice);
+        if (_bpClient is null)
+        {
+            return null;
+        }
+        return await _bpClient.CreateInvoice(invoice);
     }
 }
--- a/src/Core/Utilities/CustomRedisProcessingStrategy.cs
+++ b/src/Core/Utilities/CustomRedisProcessingStrategy.cs
@@ -8,6 +8,8 @@ using StackExchange.Redis;
 
 namespace Bit.Core.Utilities;
 
+#nullable enable
+
 /// <summary>
 /// A modified version of <see cref="AspNetCoreRateLimit.Redis.RedisProcessingStrategy"/> that gracefully
 /// handles a disrupted Redis connection. If the connection is down or the number of failed requests within
@@ -56,7 +58,7 @@ public class CustomRedisProcessingStrategy : RedisProcessingStrategy
         if (_memoryCache.TryGetValue<TimeoutCounter>(_redisTimeoutCacheKey, out var timeoutCounter))
         {
             // We've exceeded threshold, backoff Redis and skip rate limiting for now
-            if (timeoutCounter.Count >= _distributedSettings.MaxRedisTimeoutsThreshold)
+            if (timeoutCounter?.Count >= _distributedSettings.MaxRedisTimeoutsThreshold)
             {
                 _logger.LogDebug(
                     "Redis timeout threshold has been exceeded, backing off and skipping IP rate limiting");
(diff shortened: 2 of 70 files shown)


benbryant0:
<details>
<summary>BitPayClient</summary>

Returns null instead of throwing when _bpClient isn't initialized.

The one caller of `GetInvoicesAsync` handles null and returns immediately. The caller of `CreateInvoicesAsync` will now throw a nullref instead of the method itself.

</details>

<details>
<summary>StrictEmailAddressListAttribute</summary>

No behaviour change, just using pattern matching and cleaning up a bit.
</details>

<details>
<summary>LoggerFactoryExtensions</summary>

Shouldn't be any change, just split it off because it's weird.
`globalSettings` is initialised at the start of the method, but then has its members inconsistently accessed with null-conditional operators. Seems unnecessary and causes errors with NRT turned on.

</details>

<details>
<summary>CustomRedisProcessingStrategy</summary>

Shouldn't be any change. Looks like the cache only has this key set with an initialized value inside the same class.
</details>

<details>
<summary>JsonHelpers</summary>

`CreateDictionaryContract`: this seems like it really should be fine; just won't throw in a case that doesn't seem valid for this custom resolver anyway.
`PermissiveStringConverter.Read`: `reader.GetString()` says it only returns null when the token type is null, and that's not the case here.
`MsEpochConverter.Write`: seems like it just has to be a bug. But that class isn't referenced anywhere anyway.

<details>
<summary>AssemblyHelpers</summary>

`Assembly.GetEntryAssembly()` only returns null if the assembly was loaded by unmanaged code.
`AssemblyInformationalVersionAttribute` seems to always be automatically generated (it's added by the build process even for an empty project).
`GetGitHash` doesn't seem like it should throw or return null since the relevant property is added by `Directory.Build.props`, but one caller null-coalesces the result of `GetGitHash` so better to be safe I guess.
</details>

djsmith85:
@bitwarden/team-secrets-manager-dev, @bitwarden/team-billing-dev: For the files changed, no CODEOWNER has been defined, yet. Please create a separate task to look for files that would fall under your ownership and update the CODEOWNERs file.
Pinging @justindbaur as he's been championing the NRT work ❤️

src/Core/Utilities/BitPayClient.cs
```diff
@@ -15,13 +17,21 @@ public BitPayClient(GlobalSettings globalSettings)
...
 
-    public Task<BitPayLight.Models.Invoice.Invoice> GetInvoiceAsync(string id)
+    public async Task<BitPayLight.Models.Invoice.Invoice?> GetInvoiceAsync(string id)
     {
-        return _bpClient.GetInvoice(id);
+        if (_bpClient is null)
+        {
+            return null;
+        }
+        return await _bpClient.GetInvoice(id);
```
justindbaur, line 26: @bitwarden/team-billing-dev can override me here but I think I would rather have this do:
```suggestion
        Debug.Assert(_bpClient is not null);
        return await _bpClient.GetInvoice(id);
```

And let this, theoretically, throw a NRE in the runtime if this method can possibly be called when we haven't set up the BitPay token.

src/Core/Utilities/BitPayClient.cs
```diff
@@ -15,13 +17,21 @@ public BitPayClient(GlobalSettings globalSettings)
...
 
-    public Task<BitPayLight.Models.Invoice.Invoice> CreateInvoiceAsync(BitPayLight.Models.Invoice.Invoice invoice)
+    public async Task<BitPayLight.Models.Invoice.Invoice?> CreateInvoiceAsync(BitPayLight.Models.Invoice.Invoice invoice)
     {
-        return _bpClient.CreateInvoice(invoice);
+        if (_bpClient is null)
+        {
+            return null;
+        }
+        return await _bpClient.CreateInvoice(invoice);
```
justindbaur, line 35: Same here.

```suggestion
        Debug.Assert(_bpClient is not null);
        return await _bpClient.CreateInvoice(invoice);
```

src/Core/Utilities/CustomRedisProcessingStrategy.cs
```diff
@@ -56,7 +58,7 @@ public override async Task<RateLimitCounter> ProcessRequestAsync(ClientRequestId
         if (_memoryCache.TryGetValue<TimeoutCounter>(_redisTimeoutCacheKey, out var timeoutCounter))
         {
             // We've exceeded threshold, backoff Redis and skip rate limiting for now
-            if (timeoutCounter.Count >= _distributedSettings.MaxRedisTimeoutsThreshold)
+            if (timeoutCounter?.Count >= _distributedSettings.MaxRedisTimeoutsThreshold)
```
justindbaur, line 61: We shouldn't ever be caching a `null` value.

```suggestion
            Debug.Assert(timeoutCounter is not null);
            if (timeoutCounter.Count >= _distributedSettings.MaxRedisTimeoutsThreshold)
```

src/Core/Utilities/StrictEmailAddressListAttribute.cs
justindbaur, line 1: WAAAAY better, thank you!

justindbaur:
@benbryant0 We've now enabled nullable repo wide and this PR would need some updating. Do you still want to pursue this?

PR #5940, opened 2025-06-09
[PM-18699] Add trial path to Stripe metadata

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18699
## 📔 Objective

**Ticket Description**
GIVEN: a user creates a trial following the marketing initiated flow

that is, the trial/send-verification-email endpoint gets their email and sends them a verification email

they click through the email

they either create their account OR log in

they complete trial signup in the stepper flow (new users) or in the product flow (existing users)

THEN: stripe metadata should indicate that this trial was ‘marketing-initiated’

GIVEN: a user creates a trial from within the product without coming from trial/send-verification-email
THEN: stripe metadata should indicate that this trial was ‘product-initiated’

**Implementation Description :** 

PR is implementing the requirement to track whether a trial was initiated through marketing or product flows by adding metadata to Stripe subscriptions. Here's how it works:

1. The InitiationPath property is added to SubscriptionSetup class to track where the trial was initiated from.

2. When creating a subscription in OrganizationBillingService.CreateSubscriptionAsync(), the metadata is set based on the InitiationPath:
```

Metadata = new Dictionary<string, string>
{
    ["organizationId"] = organizationId.ToString(),
    ["trialInitiationPath"] = !string.IsNullOrEmpty(subscriptionSetup.InitiationPath) &&
        subscriptionSetup.InitiationPath.Contains("trial from marketing website")
        ? "marketing-initiated"
        : "product-initiated"
}
```

3. The InitiationPath is propagated from OrganizationSignup to SubscriptionSetup in OrganizationSale.From() which is why I change that upgrade  is OrganizationSignup, because this value is only passed during signup:
```
if (upgrade is OrganizationSignup signup)
{
    subscriptionSetup.InitiationPath = signup.InitiationPath;
}
```
4. The InitiationPath is set in the request models (OrganizationCreateRequestModel and OrganizationNoPaymentCreateRequest) and passed through to OrganizationSignup.

**The implementation looks correct and follows the requirements:**
When a user creates a trial following the marketing initiated flow (through trial/send-verification-email), the InitiationPath will contain "trial from marketing website" and the Stripe metadata will be set to "marketing-initiated"
When a user creates a trial from within the product, the InitiationPath will not contain that string and the metadata will be set to "product-initiated"

## 📸 Screenshots

Screenshot 2025-06-09 at 09 59 49
Screenshot 2025-06-09 at 10 01 34

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Billing/Models/Sales/OrganizationSale.cs
+++ b/src/Core/Billing/Models/Sales/OrganizationSale.cs
@@ -34,7 +34,7 @@ public class OrganizationSale
         var subscriptionSetup = GetSubscriptionSetup(signup);
 
         subscriptionSetup.SkipTrial = signup.SkipTrial;
-        subscriptionSetup.InitiationPath = organization.GatewayCustomerId;
+        subscriptionSetup.InitiationPath = signup.InitiationPath;
 
         return new OrganizationSale
         {


src/Core/Billing/Models/Sales/OrganizationSale.cs
```diff
@@ -45,11 +45,24 @@ public static OrganizationSale From(
 
     public static OrganizationSale From(
         Organization organization,
-        OrganizationUpgrade upgrade) => new()
+        OrganizationUpgrade upgrade)
+    {
+        var signup = new OrganizationSignup
```
amorask-bitwarden: This seems like a lot of reshuffling for the addition and population of a single property. I'd suggest reverting the changes to this file and looking at how [`SkipTrial`](https://github.com/bitwarden/server/blob/main/src/Core/Billing/Models/Sales/OrganizationSale.cs#L36) gets set. I think that would be a less intensive change.

amorask-bitwarden (changes requested):

amorask-bitwarden:
Hi Cy, without a description of your changes, I'm struggling to determine exactly what we're attempting here, but I still feel like we shouldn't require this many changes for the addition of one property. 

Just to work through this a bit, we are just trying to add the `TrialSource` to the `SubscriptionSetup` and then propagate that to the subscription's metadata when the subscription is created, right?

If that's the case, are the trials created using the `OrganizationUpgrade` or the `OrganizationSignup`?

amorask-bitwarden (approved):

PR #5943, opened 2025-06-09
[PM-20112] Member access stored proc and splitting the query

ttalty:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20112

## 📔 Objective

The member access query was grabbing information from multiple data sources and iterating over the data multiple times to properly join them together. 

Worked with @rkac-bw to create a stored procedure `dbo.MemberAccessReport_GetMemberAccessCipherDetailsByOrganizationId` that will provide the base data to avoid joining multiple sources in the server. Instead of selecting the entire cipher information the query only returns the cipher id. If the entire cipher object is returned it would be potentially returning unnecessary duplicates. A cipher can be associated with more than one user. If both user records return the same cipher it would be duplicated data. The client will use this data and lookup the cipher by id when looking at organization ciphers.  

The previous members access report created a top level aggregation with a list of access details as shown in the deleted `MemberAccessReportModel` then the client had to transform that data into the details for the export in the report. In the new `MemberAccessReportDetail` data is represented as the export detail. The client would simply aggregate for the user instead of the service. 

`OrganizationMemberBaseDetailRepository` was added to consume the new stored procedure. 

Both the member access report and risk insights will use the same stored procedure/repo. The data has the org user to cipher association needed for both reports. The members access report needs more information such as account recovery and two factor status. The risk insights report does not require this information. Two new queries were created to replace the old `MemberAccessCipherDetailsQuery` 

`RiskInsightsReportQuery` handling the aggregation of the user to ciphers data. 
`MemberAccessReportQuery` performing the aggregation with the additional permission data, two factor status lookup, and account recovery.

Removed files associated with the old query and modified the `ReportsController` to use the 2 new queries. 

📝 The client would need to be modified to accept the new member access data model. The risk insights report data model would not change and the client does not need any modification. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


voommen-livefront (approved):

prograhamming (approved):

rkac-bw (approved):
lgtm

PR #5955, opened 2025-06-11
[PM-22405] Add debugging instrument for finding invalid OrganizationUser state.

JimmyVo16:
Add debugging instrument for finding invalid OrganizationUser state.

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22405

## 📔 Objective

This PR adds conditional logging at the repository layer to help surface when an OrganizationUser is in an invalid state. The goal is to make it easier to identify which flow is causing the issue.

We will use Datadog to pick up the log message and trace the full flow through Datadog’s tracing tools.

## 📸 Screenshots

Simple sanity test

https://github.com/user-attachments/assets/d9007081-de3a-49f5-9ff4-41d8ae196c6c

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                logger.LogWarning("Warning invalid invited state.");
+            }
+
+            var invalidNoneInviteState = allOrgUsers.Any(user => user.Status != OrganizationUserStatusType.Invited && !user.Email.IsNullOrWhiteSpace());
+
+            if (invalidNoneInviteState)
+            {
+                logger.LogWarning("Warning invalid non invited state.");
+            }
```
eliykat: This will throw a false positive for users who have been invited and then revoked; their status will be revoked (`!= Invited`) but their email will not be null. I suggest being explicit and limiting it to `Accepted` and `Confirmed` states. I think we can probably ignore revoked here.

(For this reason, I think that `Revoked` should be a separate flag rather than an `OrganizationUserStatusType`, because today it hides their "real" or underlying status, and we have kinda [hacky logic](https://github.com/bitwarden/server/blob/4a1212095053bb93e936038a71f44b01ebbb3c36/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1691-L1708) to figure out what to restore them to.)

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
eliykat, line 1: What information would help us while debugging? OrgId? OrgUserId? As currently written I'm not sure this helps us very much. (no PII but guids are fine)

eliykat (changes requested):
We should also log if the "resend invite" method receives a request for a user who is not in the invited status. It's possible that all OrgUsers are in a valid state, but the command is being passed an invalid request.

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
JimmyVo16: We can obtain that information from Datadog, but let's include it here as well for clarity. 

Here is what I'm thinking. Let me know if there is anything else you want me to add. 

1. OrgId
2. OrgUserId
3. statues
4. HasEmail Boolean
5. userId (nullable)

JimmyVo16: We can obtain that information from Datadog, but not for bulk, so let's include it here as well for clarity.

Here is what I'm thinking. Let me know if there is anything else you want me to add.

OrgId
OrgUserId
statues
HasEmail Boolean
userId (nullable)

JimmyVo16: We can obtain that information from Datadog, but not for bulk, so let's include it here as well for clarity.

Here is what I'm thinking. Let me know if there is anything else you want me to add.

1. OrgId
2. OrgUserId
3. statues
4. HasEmail Boolean
5. userId (nullable)

src/Core/AdminConsole/Utilities/DebuggingInstruments/NoopUserInviteDebuggingLogger.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+
+using Bit.Core.AdminConsole.Utilities.DebuggingInstruments;
+using Bit.Core.Entities;
+
+public class NoopUserInviteDebuggingLogger() : IUserInviteDebuggingLogger
```
JimmyVo16: I don’t think the OSS version needs it, since this is only for debugging. Please let me know if I should do anything differently.

src/Core/AdminConsole/Services/Implementations/OrganizationService.cs
```diff
@@ -908,6 +912,8 @@ public async Task<IEnumerable<Tuple<OrganizationUser, string>>> ResendInvitesAsy
         IEnumerable<Guid> organizationUsersId)
     {
         var orgUsers = await _organizationUserRepository.GetManyAsync(organizationUsersId);
+        _userInviteDebuggingLogger.Log(orgUsers);
```
JimmyVo16: The logs in this file should account for all reinvite flows except for provider users and emergency access.

eliykat: I think we're only talking about reinviting orgUsers here, so that's fine.

src/Core/AdminConsole/Utilities/DebuggingInstruments/NoopUserInviteDebuggingLogger.cs
eliykat: Everything in `src` is OSS. Bitwarden licensed code is only in the `bitwarden_license` folder. See our license: https://github.com/bitwarden/server/blob/main/LICENSE.txt

All your changes here are to OSS code, so the noop isn't required.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
eliykat, line 1: Changes to this file can be reverted. Generally you'd register the implementation here, but in this case `OrganizationService` has its own extension class: `OrganizationServiceCollectionExtensions`. That contains everything required for the service to run, so best to add it there so it's as close to the service that requires it.

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
eliykat: Looks comprehensive to me.

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
```diff
@@ -0,0 +1,63 @@
...
+using Bit.Core.Entities;
+using Bit.Core.Enums;
+using Microsoft.Extensions.Logging;
+using Quartz.Util;
+
+namespace Bit.Core.AdminConsole.Utilities.DebuggingInstruments;
+
+public class UserInviteDebuggingLogger(ILogger<UserInviteDebuggingLogger> logger) : IUserInviteDebuggingLogger
+{
+    public void Log(OrganizationUser allOrgUser)
```
eliykat: nit: `allOrgUser` is not a very clear name. What about just `orgUser`?

eliykat (dismissed):

src/Infrastructure.Dapper/DapperServiceCollectionExtensions.cs
eliykat, line 1: I think it makes sense to register it in multiple places given that both service collections (dapper and organizationService) need it to run. However, please use TryAdd per the new ADR here: https://contributing.bitwarden.com/architecture/adr/dotnet-dependency-injection-enhancements to account for the duplication.

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System.Text.Json;
+using Bit.Core.Entities;
+using Bit.Core.Enums;
+using Microsoft.Extensions.Logging;
+using Quartz.Util;
+
+namespace Bit.Core.AdminConsole.Utilities.DebuggingInstruments;
+
+public class UserInviteDebuggingLogger(ILogger<UserInviteDebuggingLogger> logger) : IUserInviteDebuggingLogger
```
jrmccannon: Would it make sense to just have these be extension methods to the ILogger<T>?

Something like this 
```
public static class LoggerClass
{
    public static void LogUserStuff(this ILogger<OrganizationService> logger, IEnumerable<OrganizationUser> users)
    {
        // All your user log logic
    }
}
```
OrgService
```
    public async Task<IEnumerable<Tuple<OrganizationUser, string>>> ResendInvitesAsync(Guid organizationId, Guid? invitingUserId,
        IEnumerable<Guid> organizationUsersId)
    {
        var orgUsers = await _organizationUserRepository.GetManyAsync(organizationUsersId);
        _logger.LogUserStuff(orgUsers);

        var org = await GetOrgById(organizationId);

jrmccannon (changes requested):
Had a suggestion.  Wasn't sure if it would ruin anything. Might just be a bit cleaner and you wouldn't have to worry about any DI problems.

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
eliykat: I like this solution - it's lighter weight than defining a whole new service for it.

eliykat (commented):
My feedback has been addressed, however I agree with @jrmccannon that an extension method is a nicer solution here.

It also occurred to me that traces in Datadog are **highly** sampled and favour "interesting" traces, such as those that throw exceptions. Because the data is entering an invalid state without any exception being thrown, I'm not sure we'll always be able to correlate a log with a trace to identify the flow. Do we need to add some kind of identifier so we know what action triggered the log? (I might be missing something here, genuine question.)

src/Core/AdminConsole/Utilities/DebuggingInstruments/UserInviteDebuggingLogger.cs
JimmyVo16: Yeah, I can change it to this pattern since it has more pros and I don’t see any cons.

JimmyVo16:
> It also occurred to me that traces in Datadog are highly sampled and favour "interesting" traces, such as those that throw exceptions. Because the data is entering an invalid state without any exception being thrown, I'm not sure we'll always be able to correlate a log with a trace to identify the flow. Do we need to add some kind of identifier so we know what action/endpoint triggered the log? (I might be missing something here, genuine question.)

Good question. Let me find the answer to this.

eliykat:
Happy for @jrmccannon to give the final review on this either way!

jrmccannon (approved):
Since we don't want to potentially blow something up, this is a good alternative to just throwing an Exception. Looks good to me.

PR #5957, opened 2025-06-12
[PM-20348] Add pending auth request endpoint

ike-kottlowski:
## 🎟️ Tracking

[PM-20348](https://bitwarden.atlassian.net/browse/PM-20348)

[Client PR](https://github.com/bitwarden/clients/pull/15161)

## 📔 Objective

Create an endpoint that returns only pending standard auth requests for a user. This will enable browser extension approval feature.

## 📸 Screenshots

[See client PR](https://github.com/bitwarden/clients/pull/15161)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-20348]: https://bitwarden.atlassian.net/browse/PM-20348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Auth/Services/IAuthRequestService.cs
```diff
@@ -1,14 +1,51 @@
...
 #nullable enable
 
 namespace Bit.Core.Auth.Services;
 
 public interface IAuthRequestService
 {
-    Task<AuthRequest?> GetAuthRequestAsync(Guid id, Guid userId);
-    Task<AuthRequest?> GetValidatedAuthRequestAsync(Guid id, string code);
+    /// <summary>
+    /// fetches an authRequest by Id. Returns AuthRequest if AuthRequest.UserId mateches
```
Patrick-Pimentel-Bitwarden: ⛏️ 

```suggestion
    /// Fetches an authRequest by Id. Returns AuthRequest if AuthRequest.UserId mateches
```

src/Core/Auth/Services/IAuthRequestService.cs
```diff
@@ -1,14 +1,51 @@
 ﻿using Bit.Core.Auth.Entities;
```
Patrick-Pimentel-Bitwarden, line 1: ❤️ Thank you for documenting the auth request service api.

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
```diff
@@ -69,7 +69,7 @@ public async Task DeleteExpiredAsync_Works(
         // Ensure the repository responds with the amount of items it deleted and it deleted the right amount.
         // NOTE: On local development this might fail on it's first run because the developer could have expired AuthRequests
         // on their machine but aren't running the job that would delete them. The second run of this test should succeed.
```
Patrick-Pimentel-Bitwarden: ⛏️ 

```suggestion
        // Ensure the repository responds with the amount of items it deleted and it deleted the right amount, which could include other auth requests from other tests so we take the minimum known acceptable amount.
```

Patrick-Pimentel-Bitwarden (dismissed):
👍 Great work! Couple of comment updates but looks good.

Patrick-Pimentel-Bitwarden (dismissed):

rkac-bw (dismissed):
lgtm

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
```diff
@@ -1,4 +1,4 @@
-CREATE PROCEDURE [dbo].[AuthRequest_ReadPendingByUserId]
+CREATE   PROCEDURE [dbo].[AuthRequest_ReadPendingByUserId]
```
Patrick-Pimentel-Bitwarden: ⛏️ Odd spacing here.

src/Core/Auth/Models/Data/PendingAuthRequestDetails.cs
```diff
@@ -0,0 +1,89 @@
...
+        AccessCode = accessCode;
+        PublicKey = publicKey;
+        Key = key;
+        MasterPasswordHash = masterPasswordHash;
+        Approved = approved;
+        CreationDate = creationDate;
+        ResponseDate = responseDate;
+        AuthenticationDate = authenticationDate;
+        DeviceId = deviceId;
+    }
```
Patrick-Pimentel-Bitwarden: 🎨 Thoughts on ordering these alphabetically or at least making the parameters in the same order as they are assigned? The assignment seems to match the above function's assignment, would be great if they all had similar ordering.

Patrick-Pimentel-Bitwarden (approved):
👍 Thanks Ike, sorry to have you double back for it to not work because the parameters require an ordering. I forgot that there was strict ordering when coming out of the db.

src/Core/Auth/Models/Data/PendingAuthRequestDetails.cs
ike-kottlowski: The constructor needs to be in the same order the data is read from the query. In this case since we are using the wildcard it'll have to be the order the columns were created, plus the additions: `rn`, `deviceId`. 

I changed the assignment order to match the ctor parameter order.

I would like to omit `rn` but we gain flexibility (able to add and remove columns) with the wildcard. So, I think the code smell might be worth it.

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
ike-kottlowski: fixed

src/Core/Auth/Models/Data/PendingAuthRequestDetails.cs
Patrick-Pimentel-Bitwarden: Agreed!

Patrick-Pimentel-Bitwarden (dismissed):

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
```diff
@@ -0,0 +1,24 @@
...
+        SELECT
+            AR.*,
+            D.Id AS DeviceId,
+            ROW_NUMBER() OVER (PARTITION BY AR.RequestDeviceIdentifier ORDER BY AR.CreationDate DESC) AS rn
+        FROM dbo.AuthRequestView AR
+        LEFT JOIN
+        	Device D ON AR.RequestDeviceIdentifier = D.Identifier
+        WHERE AR.Type IN (0, 1) -- 0 = AuthenticateAndUnlock, 1 = Unlock
+            AND AR.CreationDate >= DATEADD(MINUTE, -@ExpirationMinutes, GETUTCDATE())
+            AND AR.UserId = @UserId
```
rkac-bw: would be better to filter on null here
`AND AR.Approved IS NULL`

instead of after collecting it

src/Core/Auth/Models/Data/PendingAuthRequestDetails.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Constructor for dapper response.
+     * Note: if the DeviceId is null it comes back as an empty guid That could change if the stored
+     * procedure runs on a different database provider.
+     * In order to maintain the flexibility of the wildcard (*) in SQL, the constructor accepts a"row number" rn of type long
+     * parameter. 'rn' was used to order the results in the SQL query. Also, SQL complains about the constructor not
+     * having the same parameters as the SELECT statement and since the SELECT uses the wildcard we need to include everything.
+     * Order matters when mapping from the Stored Procedure, so the columns are in the order they come back from the query.
+     */
```
rkac-bw, line 43: SQL Column Order (AR.*):
  - ...MasterPasswordHash, Approved, CreationDate, ResponseDate, AuthenticationDate, OrganizationId,
  RequestCountryName, DeviceId, rn

  Constructor Parameter Order:
  - ...masterPasswordHash, creationDate, responseDate, authenticationDate, approved, organizationId,
  requestCountryName, deviceId, rn

  The issue is that Approved comes before CreationDate in the SQL output, but the constructor expects
  creationDate before approved.

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
```diff
@@ -0,0 +1,24 @@
...
+    SET NOCOUNT ON;
+
+    ;WITH PendingRequests AS (
+        SELECT
+            AR.*,
+            D.Id AS DeviceId,
+            ROW_NUMBER() OVER (PARTITION BY AR.RequestDeviceIdentifier ORDER BY AR.CreationDate DESC) AS rn
+        FROM dbo.AuthRequestView AR
+        LEFT JOIN
+        	Device D ON AR.RequestDeviceIdentifier = D.Identifier
```
rkac-bw: Not sure if this logic will work for all our data, identifier is not unique
A user has the same device identifier (like a fingerprint) for multiple devices
•When the system tries to find "which device made this request", it sees multiple devices with the same fingerprint
•The system randomly picks one - which might be an old device the user doesn't use anymore


May need to do this?
 ```
AND D.Id = (
        SELECT TOP 1 Id 
        FROM Device D2 
        WHERE D2.Identifier = AR.RequestDeviceIdentifier 
            AND D2.UserId = AR.UserId
            AND D2.RevisionDate >= DATEADD(DAY, -90, GETUTCDATE())
        ORDER BY D2.RevisionDate DESC
    )
```
Actually this would be faster

```
;WITH PendingRequests AS (
    SELECT
        AR.*,
        D.Id AS DeviceId,
        ROW_NUMBER() OVER (PARTITION BY AR.RequestDeviceIdentifier ORDER BY AR.CreationDate DESC) AS auth_rn,
        ROW_NUMBER() OVER (PARTITION BY AR.RequestDeviceIdentifier, AR.UserId ORDER BY D.RevisionDate DESC, D.CreationDate DESC) AS device_rn
    FROM dbo.AuthRequestView AR
    LEFT JOIN Device D ON AR.RequestDeviceIdentifier = D.Identifier 
        AND D.UserId = AR.UserId
    WHERE AR.Type IN (0, 1)
        AND AR.CreationDate >= DATEADD(MINUTE, -@ExpirationMinutes, GETUTCDATE())
        AND AR.UserId = @UserId
)
SELECT PR.*
FROM PendingRequests PR
WHERE auth_rn = 1 
    AND (device_rn = 1 OR device_rn IS NULL) -- Get first device or handle no device found
    AND PR.Approved IS NULL;
```

Comment (rkac-bw) [reply src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql]: ok seems by design, UX_Device_UserId_Identifier  is on (UserId, Identifier)
This means same identifier can exist for different users but same user cannot have multiple devices with same identifier

so just need to add `AND D.UserId = AR.UserId` to join

File: src/Core/Auth/Models/Data/PendingAuthRequestDetails.cs (hunk 7 above)
Comment (ike-kottlowski) [reply src/Core/Auth/Models/Data/PendingAuthRequestDetails.cs:43]: I opted for the direct approach of specifying all the columns we want from the stored procedure.

File: src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql (hunk 8 above)
Comment (ike-kottlowski) [reply src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql]: addressed in https://github.com/bitwarden/server/pull/5957/commits/853551d9bd4768fb38c91cca04d3f8092c7c5529

File: src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql (hunk 6 above)
Comment (ike-kottlowski) [reply src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql]: See my [comment](https://github.com/bitwarden/server/pull/5957#discussion_r2170067495) below.

File: src/Core/Auth/Repositories/IAuthRequestRepository.cs [hunk 9]
```diff
@@ -9,6 +9,13 @@ public interface IAuthRequestRepository : IRepository<AuthRequest, Guid>
 {
     Task<int> DeleteExpiredAsync(TimeSpan userRequestExpiration, TimeSpan adminRequestExpiration, TimeSpan afterAdminApprovalExpiration);
     Task<ICollection<AuthRequest>> GetManyByUserIdAsync(Guid userId);
+    /// <summary>
+    /// Gets all active pending auth requests for a user. Each auth request in the collection will be associated with a different
+    /// device. It will be the most current request for the device.
+    /// </summary>
+    /// <param name="userId">UserId of the owner of the AuthRequests</param>
+    /// <returns>a collection Auth request details or null</returns>
```
JaredSnider-Bitwarden: ❓ :  It says it returns the collection or null, but you aren't using a nullable reference type like: 
`Task<IEnumerable<PendingAuthRequestDetails>?> GetManyPendingAuthRequestByUserId(Guid userId);`

However, isn't it best practice to return an empty collection instead of null, generally speaking? Then, we don't have to perform null checking in consuming call sites?

src/Core/Auth/Services/IAuthRequestService.cs
JaredSnider-Bitwarden, line 1: Nice new docs!

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
```diff
@@ -57,6 +62,31 @@ where ar.OrganizationId.Equals(organizationId) && ar.ResponseDate == null && ar.
...
+        var mostRecentAuthRequests = await
+            (from authRequest in dbContext.AuthRequests
+             where authRequest.Type == AuthRequestType.AuthenticateAndUnlock
+                || authRequest.Type == AuthRequestType.Unlock
+             where authRequest.UserId == userId
+             where authRequest.CreationDate.AddMinutes(expirationMinutes) > DateTime.UtcNow
+             group authRequest by authRequest.RequestDeviceIdentifier into groupedAuthRequests
+             select
+                 (from r in groupedAuthRequests
+                  join d in dbContext.Devices on r.RequestDeviceIdentifier equals d.Identifier into deviceJoin
```
JaredSnider-Bitwarden: ⚠️ : in the [stored procedure](https://github.com/bitwarden/server/pull/5957/files#diff-0cd7b3b170c33ddea2ebfe64e87fe35193a2a8fca12945c7ab3ae99bdd9ed96dR19), you now filter the device query by the identifier AND the userId.  Shouldn't we also filter by user id here?

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
```diff
@@ -57,6 +62,31 @@ where ar.OrganizationId.Equals(organizationId) && ar.ResponseDate == null && ar.
         }
     }
 
+    public async Task<IEnumerable<PendingAuthRequestDetails>> GetManyPendingAuthRequestByUserId(Guid userId)
+    {
+        var expirationMinutes = (int)_globalSettings.PasswordlessAuth.UserRequestExpiration.TotalMinutes;
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+        var mostRecentAuthRequests = await
+            (from authRequest in dbContext.AuthRequests
```
JaredSnider-Bitwarden, line 71: :question: I'm not an EF expert by any means, but we use the `AuthRequests` table directly here... but in the [stored proc](https://github.com/bitwarden/server/pull/5957/files#diff-0cd7b3b170c33ddea2ebfe64e87fe35193a2a8fca12945c7ab3ae99bdd9ed96dR17), we use the `AuthRequestView`. Is this an issue? Do we already have this view registered with EF so we can query against it instead of the table?

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
```diff
@@ -57,6 +62,31 @@ where ar.OrganizationId.Equals(organizationId) && ar.ResponseDate == null && ar.
...
+             where authRequest.Type == AuthRequestType.AuthenticateAndUnlock
+                || authRequest.Type == AuthRequestType.Unlock
+             where authRequest.UserId == userId
+             where authRequest.CreationDate.AddMinutes(expirationMinutes) > DateTime.UtcNow
```
JaredSnider-Bitwarden: ⚠️ : this is a tiny difference that isn't likely to matter in the real world, but I wanted to mention it: In the [stored proc](https://github.com/bitwarden/server/pull/5957/files#diff-0cd7b3b170c33ddea2ebfe64e87fe35193a2a8fca12945c7ab3ae99bdd9ed96dR21), you use an inclusive check, but you use a strictly greater check here.

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
```diff
@@ -66,10 +66,8 @@ public async Task DeleteExpiredAsync_Works(
         Assert.NotNull(await authRequestRepository.GetByIdAsync(notExpiredAdminApprovalRequest.Id));
         Assert.NotNull(await authRequestRepository.GetByIdAsync(notExpiredApprovedAdminApprovalRequest.Id));
 
-        // Ensure the repository responds with the amount of items it deleted and it deleted the right amount.
-        // NOTE: On local development this might fail on it's first run because the developer could have expired AuthRequests
```
JaredSnider-Bitwarden, line 70: ❓ : is this no longer a possible scenario to warn against?

JaredSnider-Bitwarden (changes requested):
Nice work on this! A few questions below:

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
ike-kottlowski: Correct, I changed the assert to be a generic `> 4` than the exact assertion it was before.

The exact assertion I felt was too restrictive and brittle. Since we already check that each of the items were removed, or not, a few lines above. Adding the strict assertion felt duplicative. Since it was brittle, I would rather have a test I can trust to fail or pass correctly and not have this warning, that requires I run the test multiple times to get the correct response.

As a side note the strict-ness of the assertion also lead to building a "cleanup" method that would remove AuthRequests after tests were finished so that this specific test would pass. That felt like a code smell and after reviewing that the test does correctly identify the items that need to be deleted or not, I felt it improved code maintainability to change the assertion.

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
ike-kottlowski: The AuthRequestsView is generally just a facade over the AuthRequests Table.

```sql
CREATE OR ALTER VIEW [dbo].[AuthRequestView]
AS
SELECT
    *
FROM
    [dbo].[AuthRequest];
```

A quick look at our EF migrations it doesn't look like we use views in the Entity Framework configurations.

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
```diff
@@ -0,0 +1,46 @@
...
+        (
+            SELECT
+                AR.*,
+                D.Id AS DeviceId,
+                ROW_NUMBER() OVER (PARTITION BY AR.RequestDeviceIdentifier ORDER BY AR.CreationDate DESC) AS rn
+            FROM dbo.AuthRequestView AR
+                LEFT JOIN Device D ON AR.RequestDeviceIdentifier = D.Identifier
+                    AND D.UserId = AR.UserId
+            WHERE AR.Type IN (0, 1) -- 0 = AuthenticateAndUnlock, 1 = Unlock
+                AND AR.CreationDate >= DATEADD(MINUTE, -@ExpirationMinutes, GETUTCDATE())
```
rkac-bw: We should move the approved into the cte to reduce data pulls
```

CREATE PROCEDURE [dbo].[AuthRequest_ReadPendingByUserId]
    @UserId UNIQUEIDENTIFIER,
    @ExpirationMinutes INT
AS
BEGIN
    SET NOCOUNT ON;

    ;
    WITH
        PendingRequests
        AS
        (
            SELECT
                AR.*,
                D.Id AS DeviceId,
                ROW_NUMBER() OVER (PARTITION BY AR.RequestDeviceIdentifier ORDER BY AR.CreationDate DESC) AS rn
            FROM dbo.AuthRequestView AR
                LEFT JOIN Device D ON AR.RequestDeviceIdentifier = D.Identifier
                    AND D.UserId = AR.UserId
            WHERE AR.Type IN (0, 1) -- 0 = AuthenticateAndUnlock, 1 = Unlock
                AND AR.CreationDate >= DATEADD(MINUTE, -@ExpirationMinutes, GETUTCDATE())
                AND AR.UserId = @UserId
		AND AR.Approved IS NULL
        )
    SELECT
        PR.Id,
        PR.UserId,
        PR.OrganizationId,
        PR.Type,
        PR.RequestDeviceIdentifier,
        PR.RequestDeviceType,
        PR.RequestIpAddress,
        PR.RequestCountryName,
        PR.ResponseDeviceId,
        PR.AccessCode,
        PR.PublicKey,
        PR.[Key],
        PR.MasterPasswordHash,
        PR.Approved,
        PR.CreationDate,
        PR.ResponseDate,
        PR.AuthenticationDate,
        PR.DeviceId
    FROM PendingRequests PR
    WHERE rn = 1;
END;

```

test/Infrastructure.IntegrationTest/Auth/Repositories/AuthRequestRepositoryTests.cs
JaredSnider-Bitwarden: Awesome! Thank you for adjusting this, and thank you for the explanation.

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
JaredSnider-Bitwarden: That's fair. We will have to be careful if we ever do add any logic into the view as the behaviors could diverge then.

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
ike-kottlowski: We put the `Approved = null` after the collection because otherwise the query will return the wrong AuthRequest.

This is because if there are 3 AuthRequests (ar, ar(-1), and ar(-2)) for the same device only one of them needs to be approved: the most recent one "ar". If the most recent one is approved then the device does not _actually_ have a pending AuthRequest, but it does have 2 other vestigial AuthRequests. If the most recent one is approved (approved != null) then this query will return the next one in line "ar(-1)", which is vestigial.

This does not meet the requirements. We want the most recent AuthRequest, and then ignore the rest if the most recent is already approved. So, I chose to aggregate the data into the most recent AuthRequests _then_ only select the ones that are `approved = null`.

src/Core/Auth/Repositories/IAuthRequestRepository.cs
ike-kottlowski: Docs here are wrong it returns an empty enumerable.

fixed. https://github.com/bitwarden/server/pull/5957/commits/eeeaefe220509174930d739d310ba2b13a5734d6

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
ike-kottlowski: Good catch, fixed. https://github.com/bitwarden/server/pull/5957/commits/52df8f65835adcc3a54b72397db9ad2b0ffb50c6

src/Infrastructure.EntityFramework/Auth/Repositories/AuthRequestRepository.cs
ike-kottlowski: Good catch, fixed. https://github.com/bitwarden/server/pull/5957/commits/52df8f65835adcc3a54b72397db9ad2b0ffb50c6

src/Sql/dbo/Auth/Stored Procedures/AuthRequest_ReadPendingByUserId.sql
rkac-bw: Ah ok, makes sense, then good

JaredSnider-Bitwarden (approved):
Nice work!

rkac-bw (approved):
lgtm

PR #5958, opened 2025-06-12
[PM-23493] Generic mailer proposal

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23493

## 📔 Objective

This implements a new Mailer service which supersedes the `HandlebarsMailService`. It allows teams to create emails without having to extend a generic service.

The `IMailer` only contains a single method, `SendEmail`, which sends an instance of `BaseMail`.

To create a new email you create two classes, and add two templates. The name of the template *must* be identical to the view class, and must exist in the same namespace.

```c#
public class TestMailView : BaseMailView
{
    public required string Token { get; init; }
    public required string Email { get; init; }
    public required string WebVaultUrl { get; init; }

    public string Url =>
        string.Format(
            "{0}/redirect-connector.html#finish-signup?token={1}&email={2}&fromEmail=true",
            WebVaultUrl,
            WebUtility.UrlEncode(Token),
            WebUtility.UrlEncode(Email)
        );
}

public class TestMail : BaseMail<TestMailView>
{
    public override string Subject { get; } = "Test Email";
}
```

And on the consuming side you would write:
```c#
var mail = new TestMail()
{
    ToEmails = ["test@bw.com"],
    View = new TestMailView() {Token = "", Email = "", WebVaultUrl = "" }
};

await mailer.SendEmail(mail);
``` 

Currently every email has to create two classes, one for the Mail and one for the view. I settled on this approach because sometimes teams sends the same email from multiple places and I wanted to ensure they are identical and you can't accidentally change the subject or metadata in one but forget the other.

This does not have to be the case and `BaseMail` could be made public, however this would mean we should heavily encourage teams to create their own methods in some service for sending emails. Which to me seems less safe than ensuring it at build.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Platform/Mailer/README.md
+++ b/src/Core/Platform/Mailer/README.md
@@ -12,7 +12,7 @@ The Mailer system consists of four main components:
 3. **BaseMailView** - Abstract base class for email template view models
 4. **IMailRenderer** - Internal interface for rendering templates (implemented by `HandlebarMailRenderer`)
 
-## How It Works
+## How To Use
 
 1. Define a view model that inherits from `BaseMailView` with properties for template data
 2. Create Handlebars templates (`.html.hbs` and `.text.hbs`) as embedded resources, preferably using the MJML pipeline,


src/Core/Platform/Mailer/README.md
```diff
@@ -0,0 +1,200 @@
...
+## Architecture
+
+The Mailer system consists of four main components:
+
+1. **IMailer** - Service interface for sending emails
+2. **BaseMail<TView>** - Abstract base class defining email metadata (recipients, subject, category)
+3. **BaseMailView** - Abstract base class for email template view models
+4. **IMailRenderer** - Internal interface for rendering templates (implemented by `HandlebarMailRenderer`)
+
+## How It Works
```
coroiu: nit: technically this section doesn't say anything about how the mailer actually works, only how it can be used
```suggestion
## How To Use
```

coroiu (approved):

PR #5960, opened 2025-06-12
[PM-22610] validate file within max length; log deletion of invalid uploads

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22610

## 📔 Objective

1. Fixes a send bug that automatically deleted file uploads whose size was > 1 MiB.
2. Logs when sends are deleted so that this behavior can be validated.
3. Ensures that allowed send file sizes never exceeds 501 MiB.

## 🦮 Reviewer guidelines


src/Core/Tools/SendFeatures/Commands/NonAnonymousSendCommand.cs
```diff
@@ -135,23 +148,31 @@ public async Task<bool> ConfirmFileSize(Send send)
...
+        var minimum = fileData.Size - SendFileSettingHelper.FILE_SIZE_LEEWAY;
+        var maximum = Math.Min(
+            fileData.Size + SendFileSettingHelper.FILE_SIZE_LEEWAY,
+            SendFileSettingHelper.MAX_FILE_SIZE
+        );
+        var (valid, size) = await _sendFileStorageService.ValidateFileAsync(send, fileData.Id, minimum, maximum);
 
-        if (!valid || realSize > SendFileSettingHelper.FILE_SIZE_LEEWAY)
+        // protect file service from upload hijacking by deleting invalid sends
+        if (!valid)
```
audreyality, line 159: 📝 The old value of this line mistakenly compared the file size against `FILE_SIZE_LEEWAY`. It should have compared against `MAX_FILE_SIZE`.

🌱 This PR incorporates the check by switching the validate method to check the file is within a specific range. 

* The maximum length constrains the range checked by the method.
* Range derivation is the command's responsibility.
* Checking the range is the storage service's responsibility.

This ensures that the range calculation is DRY, requiring no additional calculation within each storage service.

♻️ This might be made more DRY by adjusting the storage service to return the file length. The range check could then be performed exclusively by the command. I didn't implement this because it would effectively introduce a new method, which seemed a bit large for a bug report.

adudek-bw (approved):

PR #5970, opened 2025-06-16
[PM-17562] Add support for Auth on Webhook integration requests

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This PR adds optional support for auth (scheme and token) to our current webhook integration. The WebhookIntegrationHandler checks the configuration for a `Scheme` and, if present, will use the `Scheme` and `Token` to build an authentication header as part of the request. For webhooks without auth requirements, everything stays the same (both scheme and token are nullable and no Auth header is added if `Scheme` is `null`).

### Architectural Question

I've built this one to follow along our current path of configuring the `Url` at the `WebhookIntegrationConfiguration` level. This allows for the most flexibility (and multiple webhooks), but it also requires a lot of duplication and now requires token and scheme to be duplicated as well. We could alternatively move `Scheme`, `Token`, and `Url` to the `WebhookIntegration` level, which would require far less duplication on each configuration, but would also mean that Organizations could only really have one webhook (unless we relax the restriction on the `OrganizationIntegration` table that only allows for one instance of an integration type).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Models/Data/EventIntegrations/WebhookIntegrationConfiguration.cs
+++ b/src/Core/AdminConsole/Models/Data/EventIntegrations/WebhookIntegrationConfiguration.cs
@@ -2,4 +2,4 @@
 
 namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 
-public record WebhookIntegrationConfiguration(string url);
+public record WebhookIntegrationConfiguration(string? Scheme, string? Token, string Url);
--- a/src/Core/AdminConsole/Services/Implementations/EventIntegrations/WebhookIntegrationHandler.cs
+++ b/src/Core/AdminConsole/Services/Implementations/EventIntegrations/WebhookIntegrationHandler.cs
@@ -2,6 +2,7 @@
 
 using System.Globalization;
 using System.Net;
+using System.Net.Http.Headers;
 using System.Text;
 using Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 
@@ -18,8 +19,16 @@ public class WebhookIntegrationHandler(IHttpClientFactory httpClientFactory)
 
     public override async Task<IntegrationHandlerResult> HandleAsync(IntegrationMessage<WebhookIntegrationConfigurationDetails> message)
     {
-        var content = new StringContent(message.RenderedTemplate, Encoding.UTF8, "application/json");
-        var response = await _httpClient.PostAsync(message.Configuration.url, content);
+        var request = new HttpRequestMessage(HttpMethod.Post, message.Configuration.Url);
+        request.Content = new StringContent(message.RenderedTemplate, Encoding.UTF8, "application/json");
+        if (!string.IsNullOrEmpty(message.Configuration.Scheme))
+        {
+            request.Headers.Authorization = new AuthenticationHeaderValue(
+                scheme: message.Configuration.Scheme,
+                parameter: message.Configuration.Token
+            );
+        }
+        var response = await _httpClient.SendAsync(request);
         var result = new IntegrationHandlerResult(success: response.IsSuccessStatusCode, message);
 
         switch (response.StatusCode)
--- a/test/Api.Test/AdminConsole/Controllers/OrganizationIntegrationsConfigurationControllerTests.cs
+++ b/test/Api.Test/AdminConsole/Controllers/OrganizationIntegrationsConfigurationControllerTests.cs
@@ -151,7 +151,7 @@ public class OrganizationIntegrationsConfigurationControllerTests
     {
         organizationIntegration.OrganizationId = organizationId;
         organizationIntegration.Type = IntegrationType.Slack;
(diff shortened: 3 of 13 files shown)


src/Core/AdminConsole/Services/Implementations/EventIntegrations/WebhookIntegrationHandler.cs
```diff
@@ -18,8 +19,16 @@ public class WebhookIntegrationHandler(IHttpClientFactory httpClientFactory)
...
     {
-        var content = new StringContent(message.RenderedTemplate, Encoding.UTF8, "application/json");
-        var response = await _httpClient.PostAsync(message.Configuration.url, content);
+        var request = new HttpRequestMessage(HttpMethod.Post, message.Configuration.Url);
+        request.Content = new StringContent(message.RenderedTemplate, Encoding.UTF8, "application/json");
+        if (!string.IsNullOrEmpty(message.Configuration.Scheme))
+        {
+            request.Headers.Authorization = new AuthenticationHeaderValue(
+                scheme: message.Configuration.Scheme,
+                parameter: message.Configuration.Token
```
withinfocus, line 30: 💭 Is it alright if the token is null or empty when the scheme isn't? Weird situation and practically-speaking doesn't matter if we check I guess.

test/Api.Test/AdminConsole/Controllers/OrganizationIntegrationsConfigurationControllerTests.cs
```diff
@@ -350,7 +350,7 @@ public async Task PostAsync_InvalidTemplate_ThrowsBadRequestException(
     {
         organizationIntegration.OrganizationId = organizationId;
         organizationIntegration.Type = IntegrationType.Webhook;
-        var webhookConfig = new WebhookIntegrationConfiguration(url: "https://localhost");
+        var webhookConfig = new WebhookIntegrationConfiguration("Bearer", "AUTH-TOKEN", Url: "https://localhost");
```
withinfocus: ❓ Should we have tests with and without the new additions?

src/Core/AdminConsole/Models/Data/EventIntegrations/WebhookIntegrationConfiguration.cs
```diff
@@ -2,4 +2,4 @@
 
 namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 
-public record WebhookIntegrationConfiguration(string url);
+public record WebhookIntegrationConfiguration(string? Scheme, string? Token, string Url);
```
withinfocus: 🎨 I like to think of optional parameters as followers to required ones, and while it isn't necessarily impactful here the `Url` would be first.

withinfocus (approved):
This works as-is, but a couple areas of improvement.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/WebhookIntegrationHandler.cs
brant-livefront: I copied the `AuthenticationHeaderValue` approach - `scheme` is `string`, but `token` is `string?`. My thought is we mirror that so that we could ostensibly configure whatever has out of the box support.

test/Api.Test/AdminConsole/Controllers/OrganizationIntegrationsConfigurationControllerTests.cs
brant-livefront: 👍  Good call. I did add those in the handler tests, but we should have them here.

src/Core/AdminConsole/Models/Data/EventIntegrations/WebhookIntegrationConfiguration.cs
brant-livefront: I'll refactor to that. 👍

src/Core/AdminConsole/Models/Data/EventIntegrations/WebhookIntegrationConfiguration.cs
```diff
@@ -2,4 +2,4 @@
 
 namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 
-public record WebhookIntegrationConfiguration(string url);
+public record WebhookIntegrationConfiguration(string Url, string? Scheme = null, string? Token = null);
```
withinfocus, line 5: 🎨 After you changed the parameter order, it made me think about any accidental swaps of your intended values in the commit. I don't see any of that, but if this first parameter were a `Uri` it would be a compile error at least. I am not sure of the impact of that addition and it would be extra work on the parsing side (and perhaps useless).

withinfocus (approved):

BTreston (dismissed):

withinfocus:
@BTreston this is ready for final approval.

r-tome (approved):

PR #5971, opened 2025-06-16
[PM-17562] Add integration filter support

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This is a POC of an idea for adding arbitrary filters to our existing event integrations. The idea would be to give organization admins the ability to add a filter to a configuration that would filter out events (e.g. I want to listen for these key events, but only when it happens in a specific set of high priority collections). I'm new to expression trees and this is a first draft POC, so feel free to provide as much feedback as you want.

Here's the basic structure of the idea:

1. We add a Filters column in the `OrganizationIntegrationConfiguration` table - alongside where we store Template. (Note: I haven't added this piece yet since I wanted to get feedback first on this approach. The property I pull out is `string?` so for now it's fine and I'm just doing a bunch in unit tests)
2. That Filters column is a JSON structure that stores a filter group - basically a collection of Rules and whether it's `all` or `any` in evaluating them. Rules are just the property, the operation and the value. So you can say "CollectionId, In, a list of values". Or "CollectionId, Equals, exact value". Right now I have in, equals and before/after for dates.
3. When the `EventIntegrationHandler` pulls the DB configuration, this would just be another column in the object that comes back (and would get cached as well when we get to that). So no extra load on the DB or new call.
4. The `EventIntegrationHandler` serializes the Filters property and hands it to the filter service. The filter service has a set of compiled expression trees that it can use to quickly run the filter calculation. Since we have a limited set of properties on `EventMessage` we can build a handful of useful expressions to evaluate. If we get a better idea of what admins might like to be filtering, we could probably adjust that as well.
5. At runtime, we just run the compiled expression, passing in the dynamic set of values. So for example, CollectionId, In, set of values - the CollectionId property is already part of the compiled expressions - we build that ahead of time. But the list of values is provided by the Filters at runtime from the DB. So we can efficiently run the query without needing to reflect.
6. The outcome of that expression tree evaluation is a bool that comes back to `EventIntegrationHandler` - if it's false, we simply move on and never publish the message that would kick off the integration. If it's true, we proceed as we would have previously.

So net-net, this should give us somewhat arbitrary filters without really any performance change to what we're doing now. We don't add any more load on the DB and we hopefully don't add much performance load to the handler due to the expression trees.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationFilterOperation.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+
+public enum IntegrationFilterOperation
+{
+    Equals,
+    NotEquals,
+    In,
+    NotIn,
+    DateAfter,
+    DateBefore
```
withinfocus: 💭 These feel expensive / taxing to me and maybe we don't include them. If this were used in a filter it would have a questionable lifetime but a constant need for evaluation i.e. "login events after 1/1/25" actually processing all login events or "login events before 1/1/25" being evaluated for every event and doing nothing.

src/Core/AdminConsole/Models/Data/Organizations/OrganizationIntegrationConfigurationDetails.cs
```diff
@@ -12,6 +12,7 @@ public class OrganizationIntegrationConfigurationDetails
     public IntegrationType IntegrationType { get; set; }
     public EventType EventType { get; set; }
     public string? Configuration { get; set; }
+    public string? Filters { get; set; }
```
withinfocus, line 15: ℹ️ This will be on the main table object too, and obviously we need to do all the DB work (which I can handle once settled).

src/Core/AdminConsole/Services/IIntegrationFilterService.cs
```diff
@@ -0,0 +1,11 @@
+﻿#nullable enable
+
+using Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+using Bit.Core.Models.Data;
+
+namespace Bit.Core.Services;
+
+public interface IIntegrationFilterService
```
withinfocus, line 8: ❓ I only see one implementation of this and it's very static in nature -- do we even need DI?

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
```diff
@@ -31,25 +34,47 @@ public async Task HandleEventAsync(EventMessage eventMessage)
...
+            {
+                if (configuration.Filters is string filterJson)
+                {
+                    // Evaluate filters - if false, then discard and do not process
+                    var filters = JsonSerializer.Deserialize<IntegrationFilterGroup>(filterJson)
+                        ?? throw new InvalidOperationException($"Failed to deserialize Filters to FilterGroup");
+                    if (!integrationFilterService.EvaluateFilterGroup(filters, eventMessage))
+                    {
+                        continue;
+                    }
```
withinfocus, line 47: 🎨 Any and all optimizations for speed here will be important. I would think we could abstract away the validation when asking for filters so that it happens once on load and be done. Does all this happen when we introduce a cache?

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
```diff
@@ -31,25 +34,47 @@ public async Task HandleEventAsync(EventMessage eventMessage)
...
+                };
 
-            await eventIntegrationPublisher.PublishAsync(message);
+                await eventIntegrationPublisher.PublishAsync(message);
+            }
+            catch (Exception exception)
+            {
+                logger.LogError(
+                    exception,
+                    "Failed to publish Integration Message for {Type}, check record Id {RecordId} for error in Configuration or Filters",
```
withinfocus: 🎨 You can just say "configuration" with the ID -- eliminate "record".

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationFilterService.cs
```diff
@@ -0,0 +1,204 @@
...
+        AddInFilter<Guid?>("ProviderUserId");
+        AddInFilter<Guid?>("ProviderOrganizationId");
+        AddInFilter<Guid?>("ActingUserId");
+        AddInFilter<Guid?>("SecretId");
+        AddInFilter<Guid?>("ServiceAccountId");
+
+        AddDateFilter("Date", before: true);
+        AddDateFilter("Date", before: false);
+    }
+    private void AddEqualityFilter<T>(string propertyName)
```
withinfocus: ⛏️ Newlines between functions.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
 - An array of `OrganizationIntegrationConfigurationDetails` is what the `EventIntegrationHandler` fetches from
   the database to determine what to publish at the integration level.
 
+## Filtering
+
+In addition to the ability to configure integrations mentioned above, Organization admins can
```
withinfocus:
```suggestion
In addition to the ability to configure integrations mentioned above, organization admins can
```

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+also add `Filters` stored in the `OrganizationIntegrationConfiguration`. Filters are completely
+optional and as simple or complex as organization admins want to make them. These are stored in
+the database as JSON and serialized into an `IntegrationFilterGroup`. This is then passed to
+the `IntegrationFilterService`, which evaluates it to a `bool`. If it's `true`, the integration
+proceeds as above. If it's `false`, we ignore this event and do not route it to the integration
+level.
+
+### `IntegrationFilterGroup`
+- Logical AND/OR grouping of a number of rules and other subgroups.
```
withinfocus: ⛏️ Newline.

withinfocus:
```suggestion
- Logical `AND / OR` grouping of a number of rules and other subgroups.
```

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+- `AndOperator` indicates if they all must be true or if it should be true if any of them are
+  true.
+  - This applies to _both_ the inner group and the list of rules.
+  - For instance, if this group contained Rule1 and Rule2 and then Group1 and Group2:
```
withinfocus: ⛏️ Combine this into one sentence with a semicolon.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+    - `true` would means: `Rule1 && Rule2 && Group1 && Group2`
```
withinfocus:
```suggestion
    - `true` would mean: `Rule1 && Rule2 && Group1 && Group2`
```

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+    - `false` would mean `Rule1 || Rule2 || Group1 || Group2`
+- `Rules` is a list of `IntegrationFlterRule`
+  - This can be null or empty - in which case it would provide only `true` in any calculations.
```
withinfocus:
```suggestion
  - This can be null or empty, in which case it would provide only `true` in any calculations.
```

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
withinfocus: ⛏️ Periods at the end of each bullet.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+- `Groups` is a list of additional `IntegrationFlterGroup`
+    - This can be null or empty - in which case it would provide only `true` in any calculations.
```
withinfocus:
```suggestion
    - This can be null or empty, in which case it would provide only `true` in any calculations.
```

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+
+### `IntegrationFilterRule`
+- The core of the filtering framework to determine if the data in this specific `EventMessage`
```
withinfocus: ⛏️ Newline.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+  matches the data for which the filter is searching.
+- `Property` is the specific property on `EventMessage` that will be evaluated (e.g. `CollectionId`)
+- `Operation` is the type of comparison to perform on the property and value
+  - Supported Operations:
```
withinfocus:
```suggestion
  - Supported operations:
```

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -228,6 +228,67 @@ Currently, there are integrations / handlers for Slack and webhooks (as mentione
...
+
+In addition to the ability to configure integrations mentioned above, Organization admins can
+also add `Filters` stored in the `OrganizationIntegrationConfiguration`. Filters are completely
+optional and as simple or complex as organization admins want to make them. These are stored in
+the database as JSON and serialized into an `IntegrationFilterGroup`. This is then passed to
+the `IntegrationFilterService`, which evaluates it to a `bool`. If it's `true`, the integration
+proceeds as above. If it's `false`, we ignore this event and do not route it to the integration
+level.
+
+### `IntegrationFilterGroup`
```
withinfocus, line 241: 🎨 Consider using tables for this and the section below it -- would that make it read more easily you think?

withinfocus (commented):
The unit tests for the filtering help a lot because I don't necessarily feel qualified to review all that expression matching logic! I left a few comments and couldn't help myself with some documentation ⛏️s but I think this is a workable direction. Performance is gonna be crucial so I don't know if we try to figure out a cache first or after this.

Lemme know about desired next steps like doing the database work.

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationFilterOperation.cs
brant-livefront: Yeah, I agree. Since these are happening in real time, the Date is not really a useful filter option. I'll pull them out.

src/Core/AdminConsole/Services/IIntegrationFilterService.cs
brant-livefront: We are using that to mock the filter service for tests (so I can force it to return a value and make sure the handler is correctly routing the messages).

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
brant-livefront: I pushed up the above fixes, but I'll try playing with tables. I think that could work.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
brant-livefront: This is just making sure we parsed the filters correct from the DB (i.e. we didn't store a bad filter). The filters themselves are parsed and cached when the IntegrationFilterService starts.

This filters column would be part of the DB cache. I wonder if there's a way for us to cache the filters in addition where we run a filter before even looking to the DB cache? If we think a lot of operations might get filtered out, that might be a solid optimization.

I guess that also raises another question of if there should be maybe a larger scope for filters? Right now, I'm allowing for filters at the event type level which is the most flexible. But it is also one of the more granular (i.e. if I have the same filter on every event type, I'd have to process and store that on all of them with no optimization across the whole integration / organization)

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
brant-livefront: Pushed up the last commit with a shot at making these tables. It makes the editor lines exceptionally long, but I think it does make it a bit more readable. See which you prefer and I'll adjust 👍

withinfocus (changes requested):
This is good, but needs the database changes first.

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationFilterOperation.cs
```diff
@@ -0,0 +1,10 @@
+﻿#nullable enable
+namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+
+public enum IntegrationFilterOperation
+{
+    Equals,
+    NotEquals,
+    In,
+    NotIn
```
BTreston: ⛏️ We should assign the numerical value to these enum members

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationFilterService.cs
```diff
@@ -0,0 +1,159 @@
...
+{
+    private delegate bool CompiledFilter(EventMessage message, object? value);
+
+    private readonly Dictionary<string, CompiledFilter> _equalsFilters = new();
+    private readonly Dictionary<string, CompiledFilter> _inFilters = new();
+
+    public IntegrationFilterService()
+    {
+        BuildFilters();
+    }
```
BTreston: This feels like it would fit better encapsulated in its own class, including `AddEqualityFilter` and `AddInFilter`

BTreston (changes requested):
I think we should add a feature flag for this. I'm roping in @eliykat for a second set of eyes, as I think he will have some additional feedback

withinfocus:
> I think we should add a feature flag for this.

Yep, and we do. None of this is on anywhere but dev and it's not running in our cloud instance since no Service Bus is configured.

withinfocus (dismissed):

src/Core/AdminConsole/Models/Data/Organizations/OrganizationIntegrationConfigurationDetails.cs
brant-livefront: These changes should all be here now (merged in from the other PR)

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationFilterService.cs
brant-livefront: In the latest commit, I pulled out the creation of expressions into `IntegrationFilterFactory`. So now the factory is responsible for creating the expressions and the service is responsible for groups and rules (and simply executing the expressions, not building them).

withinfocus (dismissed):

withinfocus:
@BTreston or @eliykat this is ready for final approval.

eliykat:
I may still be several days away from getting to this, so I've requested a review from @r-tome and @jrmccannon in case they have the capacity to look at it before me. I definitely think it's worth the additional review though.

At a glance, is there a UI for configuring integrations (including the new filter capability), or are you just editing the db directly at this stage?

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationFilterService.cs
```diff
@@ -0,0 +1,111 @@
...
+            "CollectionId",
+            "GroupId",
+            "PolicyId",
+            "OrganizationUserId",
+            "ProviderUserId",
+            "ProviderOrganizationId",
+            "ActingUserId",
+            "SecretId",
+            "ServiceAccountId"
+        };
```
r-tome: These properties could be moved to a static readonly field

r-tome (approved):
This looks great to me, I just left a minor non-blocking suggestion

brant-livefront:
@eliykat 
> At a glance, is there a UI for configuring integrations (including the new filter capability), or are you just editing the db directly at this stage?

Sorry I missed this comment earlier. There is no UI for adding these yet, but there is an API - `OrganizationIntegrationConfigurationController`.

r-tome (approved):

r-tome (approved):

withinfocus (approved):

PR #5973, opened 2025-06-16
[PM-17562] Fix flickering unit test - WebhookIntegrationHandlerTests

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

One of the tests in `WebhookIntegrationHandlerTests` has been flickering (passing/failing on and off). This was a result of calling DateTime.UtcNow in multiple places in the test. If the test ran slightly too long, it would slip outside the window of the expected range. This PR locks down a specific value at the front of the test so that all of the values agree.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


withinfocus (dismissed):

eliykat (changes requested):
This is a little better, but is ultimately still relying on the vagaries of the system clock.

Please use `TimeProvider` so that you can mock it in tests. Here's an example:
1. In your sut, inject `TimeProvider` and use it to get the UTC: https://github.com/bitwarden/server/blob/4a1212095053bb93e936038a71f44b01ebbb3c36/src/Core/Platform/Installations/Commands/UpdateInstallationActivityDateCommand/UpdateInstallationCommand.cs#L52-L53
2. In your tests, initialize your `SutProvider` to use a fake time provider and set a constant time: https://github.com/bitwarden/server/blob/4a1212095053bb93e936038a71f44b01ebbb3c36/test/Core.Test/Platform/Installations/Commands/UpdateInstallationCommandTests.cs#L62-L68
3. Write your assertions with the guarantee that the time has not advanced unless/until you expressly advance it. You can assert an exact time without having to use `InRange` to allow for a presumed execution time.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/WebhookIntegrationHandler.cs
```diff
@@ -9,7 +9,9 @@
 
 namespace Bit.Core.Services;
 
-public class WebhookIntegrationHandler(IHttpClientFactory httpClientFactory)
+public class WebhookIntegrationHandler(
+    IHttpClientFactory httpClientFactory,
+    TimeProvider timeProvider)
```
withinfocus, line 14: ❓ Where else do we need to use this now? Anywhere?

withinfocus (commented):
I'll let Thomas provide the approval.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/WebhookIntegrationHandler.cs
brant-livefront: Good question. I'll take a deeper look at our retry timing. I think we could probably inject there as well.

brant-livefront: I found at least one other place, so I'll update those as well. 👍

brant-livefront: Updated in latest commit. 👍

test/Core.Test/AdminConsole/Services/RabbitMqIntegrationListenerServiceTests.cs
```diff
@@ -55,7 +61,7 @@ public async Task ProcessReceivedMessageAsync_FailureNotRetryable_PublishesToDea
         var cancellationToken = CancellationToken.None;
         await sutProvider.Sut.StartAsync(cancellationToken);
 
-        message.DelayUntilDate = DateTime.UtcNow.AddMinutes(-1);
+        message.DelayUntilDate = null;
```
withinfocus, line 64: ❓ Why set these `null` ones at all now?

brant-livefront: In those tests, we're testing something other than the retry logic. So setting them to null just bypasses any date checks at all. We could definitely set them to the past as I was doing, but when I looked at it I thought `null` was a better way of just ignoring that whole path. And `DelayUntilDate` should be `null` for non-retry messages, so this is a normal case - the only reason to set them explicitly null is because we're using `BitAutoData` to generate the message and it could add a future date.

test/Core.Test/AdminConsole/Services/RabbitMqIntegrationListenerServiceTests.cs
```diff
@@ -18,19 +19,24 @@ public class RabbitMqIntegrationListenerServiceTests
...
+    private readonly DateTime _now = new DateTime(2014, 3, 2, 1, 0, 0, DateTimeKind.Utc);
     private readonly IIntegrationHandler _handler = Substitute.For<IIntegrationHandler>();
     private readonly IRabbitMqService _rabbitMqService = Substitute.For<IRabbitMqService>();
 
     private SutProvider<RabbitMqIntegrationListenerService> GetSutProvider()
     {
-        return new SutProvider<RabbitMqIntegrationListenerService>()
+        var sutProvider = new SutProvider<RabbitMqIntegrationListenerService>()
             .SetDependency(_handler)
             .SetDependency(_rabbitMqService)
```
eliykat, line 30: No change required here, but for future reference: `SutProvider` will automatically populate with NSubstitute mocks when you call `Create()` for any dependencies not explicitly set. You can retrieve the mocks using `GetDependency<T>` in order to arrange them. So you really only need to call `SetDependency` if you need something more specific than that.

That will also handle reinitializing the mocks for each test. As currently written I'm not certain that your mocks are reset at the start of each test.

eliykat (approved):
Thanks!

test/Core.Test/AdminConsole/Services/RabbitMqIntegrationListenerServiceTests.cs
brant-livefront: Thanks for the note. I do know about the GetDependency and I've used that in a few places. But I sometimes would use an injected mock just for readability purposes (i.e. if I have multiple things to change/check on rabbitMqService, it's nicer than having strings of GetDependency calls). 

**But** I didn't think about the reseting in between tests. That's a really good point.

PR #5983, opened 2025-06-18
[PM-20576] OrganizationReport - Queries and Command

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20576

## 📔 Objective

An earlier PR created a table called `OrganizationReport`.
This tasks involves creating the commands and queries to Add, Get, Drop records to that table - with unit tests.

## 📸 Screenshots

No screenshots at this moment.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+        if (organization == null)
+        {
+            return new Tuple<AddOrganizationReportRequest, bool, string>(request, false, "Invalid Organization");
+        }
+
+        // ensure that we have a URL
+        if (string.IsNullOrWhiteSpace(request.ReportData))
+        {
+            return new Tuple<AddOrganizationReportRequest, bool, string>(request, false, "Report Data is required");
+        }
```
Banrion: ❓ Is this actually checking for a url in the report data or just that the report data was passed in? This question is based on the comment.

src/Core/Dirt/Reports/ReportFeatures/DropOrganizationReportCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        _organizationReportRepo = organizationReportRepository;
+    }
+
+    public async Task DropOrganizationReportAsync(DropOrganizationReportRequest request)
+    {
+        var data = await _organizationReportRepo.GetByOrganizationIdAsync(request.OrganizationId);
+        if (data == null || data.Count() == 0)
+        {
+            throw new BadRequestException("Organization does not have any records.");
```
Banrion: ⚠️  I'm not sure if this is common practice for us, but I would expect an error message for if the organization report is not found (such as a 404 error response) rather than providing information about that organization. I've often followed practice that we don't expose internal information in request responses to avoid attacks that target information gathering. This includes letting the requester know if an organization even has or doesn't have reports.

src/Core/Dirt/Reports/ReportFeatures/DropOrganizationReportCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        data.Where(_ => request.OrganizationReportIds.Contains(_.Id)).ToList().ForEach(async _ =>
+        {
+            await _organizationReportRepo.DeleteAsync(_);
+        });
```
Banrion: ❓ Following the above comment, does this handle cases where the report isn't found? Do we need to?

src/Sql/dbo/Dirt/Views/OrganizationReportView.sql
```diff
@@ -0,0 +1,2 @@
+CREATE VIEW [dbo].[OrganizationReportView] AS
```
Banrion: ❓ We have the view but I don't see it being used in the report reads like the view is being used in the organization applications. Is this intended?

test/Infrastructure.EFIntegration.Test/Infrastructure.EFIntegration.Test.csproj
```diff
@@ -30,6 +30,7 @@
 
     <ItemGroup>
       <Folder Include="AutoFixture" />
+      <Folder Include="Tools\Repositories\" />
```
Banrion: 💭  I see this is using "Tools". Just wanted to point this out in case this needs to be `dirt`.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Bit.Core.Dirt.Entities;
+using Bit.Core.Dirt.Reports.ReportFeatures.Interfaces;
+using Bit.Core.Dirt.Reports.ReportFeatures.Requests;
+using Bit.Core.Dirt.Repositories;
+using Bit.Core.Exceptions;
+using Bit.Core.Repositories;
+
+namespace Bit.Core.Dirt.Reports.ReportFeatures;
```
prograhamming, line 9: 🌱 In the future we might want to break the reports up into their own directories.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public AddOrganizationReportCommand(
+        IOrganizationRepository organizationRepository,
+        IOrganizationReportRepository organizationReportRepository)
+    {
+        _organizationRepo = organizationRepository;
+        _organizationReportRepo = organizationReportRepository;
+    }
+
+    public async Task<OrganizationReport> AddOrganizationReportAsync(AddOrganizationReportRequest request)
```
prograhamming, line 27: 🤔 In CQRS, commands typically don't return anything back to the front end unless it is a success or failure response or returning an id of the resource created.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+        var organization = await _organizationRepo.GetByIdAsync(request.OrganizationId);
+        if (organization == null)
+        {
+            return new Tuple<AddOrganizationReportRequest, bool, string>(request, false, "Invalid Organization");
+        }
+
+        // ensure that we have a URL
+        if (string.IsNullOrWhiteSpace(request.ReportData))
+        {
+            return new Tuple<AddOrganizationReportRequest, bool, string>(request, false, "Report Data is required");
```
prograhamming: ❌ I think this should be a type versus a tuple. It will make it more readable and self-documenting.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+        IOrganizationRepository organizationRepository,
+        IOrganizationReportRepository organizationReportRepository)
+    {
+        _organizationRepo = organizationRepository;
+        _organizationReportRepo = organizationReportRepository;
+    }
+
+    public async Task<OrganizationReport> AddOrganizationReportAsync(AddOrganizationReportRequest request)
+    {
+        var (req, IsValid, errorMessage) = await ValidateRequestAsync(request);
```
prograhamming: ❓Do we need to return the `AddOrganizationReportRequest` back from the `ValidateRequestAsync()` method? We don't do anything with `req` and only use the `request` parameter in this code.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+            RevisionDate = DateTime.UtcNow
+        };
+
+        organizationReport.SetNewId();
+
+        var data = await _organizationReportRepo.CreateAsync(organizationReport);
+        return data;
+    }
+
+    private async Task<Tuple<AddOrganizationReportRequest, bool, string>> ValidateRequestAsync(
```
prograhamming: ❌  same comment about types and tuples. 

❓same comment about `AddOrganizationReportRequest`

src/Core/Dirt/Reports/ReportFeatures/DropOrganizationReportCommand.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Bit.Core.Dirt.Reports.ReportFeatures.Interfaces;
+using Bit.Core.Dirt.Reports.ReportFeatures.Requests;
+using Bit.Core.Dirt.Repositories;
+using Bit.Core.Exceptions;
+
+namespace Bit.Core.Dirt.Reports.ReportFeatures;
+
+public class DropOrganizationReportCommand : IDropOrganizationReportCommand
```
prograhamming, line 9: ❌ Need to add logging through out the commands and queries. We are totally blind on the server if an issue comes up. Logging would help us give context into what is going on and where.

util/Migrator/DbScripts/2025-06-13-00_OrganizationReport.sql
```diff
@@ -0,0 +1,118 @@
...
+
+CREATE OR ALTER VIEW [dbo].[OrganizationReportView]
+AS
+    SELECT
+        *
+    FROM
+        [dbo].[OrganizationReport];
+GO
+
+CREATE PROCEDURE [dbo].[OrganizationReport_Create]
```
rkac-bw: Missing create or alter

util/Migrator/DbScripts/2025-06-13-00_OrganizationReport.sql
```diff
@@ -0,0 +1,118 @@
...
+        [CreationDate] DATETIME2 (7) NOT NULL,
+        [RevisionDate] DATETIME2 (7) NOT NULL,
+        CONSTRAINT [PK_OrganizationReport] PRIMARY KEY CLUSTERED ([Id] ASC),
+        CONSTRAINT [FK_OrganizationReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id])
+    );
+
+    CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId] ON [dbo].[OrganizationReport]([OrganizationId] ASC);
+
+    CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_Date] ON [dbo].[OrganizationReport]([OrganizationId] ASC, [Date] DESC);
+
```
rkac-bw: will want to check for existence
```
IF NOT EXISTS (
    SELECT 1
    FROM sys.indexes
    WHERE name = 'IX_OrganizationReport_OrganizationId'
      AND object_id = OBJECT_ID('dbo.OrganizationReport')
)
BEGIN
    CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId]
    ON [dbo].[OrganizationReport]([OrganizationId] ASC);
END
GO

IF NOT EXISTS (
    SELECT 1
    FROM sys.indexes
    WHERE name = 'IX_OrganizationReport_OrganizationId_Date'
      AND object_id = OBJECT_ID('dbo.OrganizationReport')
)
BEGIN
    CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_Date]
    ON [dbo].[OrganizationReport]([OrganizationId] ASC, [Date] DESC);
END
GO
```

src/Core/Dirt/Reports/ReportFeatures/DropOrganizationReportCommand.cs
voommen-livefront: In this case, if no report exists, nothing is deleted - regardless, no errors are thrown or message returned thereby keeping this endpoint idempotent. 
It was intentional in this endpoint not to check if a report exists during deletion.

src/Core/Dirt/Reports/ReportFeatures/DropOrganizationReportCommand.cs
voommen-livefront: Another excellent catch. I'll update the message to be more generic, and I will add logs to be specific.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
voommen-livefront: Good catch! I'll fix the comments and add logs

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
voommen-livefront: There was only use case for this and since it was in the same file, I used a tuple. I have removed the `request` as a return value - like you suggested - and that simplifies this line a lot more.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
voommen-livefront: Great catch! you are right - we don't need the request to be returned.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
voommen-livefront: agreed - the changes you suggested made this cleaner.

util/Migrator/DbScripts/2025-06-13-00_OrganizationReport.sql
voommen-livefront: I will fix this in the previous PR https://github.com/bitwarden/server/pull/5839
Since this PR depended on that - I have branched out of that PR.

util/Migrator/DbScripts/2025-06-13-00_OrganizationReport.sql
voommen-livefront: I figured that if the table is being created, the indexes shouldn't already exist. Therefore, I didn't do this check.

src/Core/Dirt/Reports/ReportFeatures/DropOrganizationReportCommand.cs
voommen-livefront: Excellent catch! Thank you! I completely missed logging! I will add logs to all the commands and queries.

src/Sql/dbo/Dirt/Views/OrganizationReportView.sql
voommen-livefront: I have this fixed in the PR for the DB changes - we are using this in the `OrganizationReport_ReadById` and `OrganizationReport_ReadByOrganizationId` procedures.

prograhamming (approved):

test/Infrastructure.EFIntegration.Test/Infrastructure.EFIntegration.Test.csproj
voommen-livefront: You are right - this was not needed. I have removed it. The existing reports are in `dirt/repositories`

prograhamming (approved):

PR #5988, opened 2025-06-20
[PM-17562] Add database support for integration filters

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This PR adds the database migration to support filters added in PR #5971 (which is currently still under review and the target of this PR). This should be a backwards compatible change because we are only adding a nullable column to `OrganiztionIntegrationConfiguration` (along with the Create/Update stored procedures) and recreating the `OrganiztionIntegrationConfigurationDetailsView` to add the column there as well. As such it's just one migration.

In addition, there are some code changes to the request models and such to support specifying filters when creating/updating a configuration via the API.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Sql/dbo/Tables/OrganizationIntegrationConfiguration.sql
+++ b/src/Sql/dbo/Tables/OrganizationIntegrationConfiguration.sql
@@ -4,10 +4,10 @@ CREATE TABLE [dbo].[OrganizationIntegrationConfiguration]
     [OrganizationIntegrationId] UNIQUEIDENTIFIER NOT NULL,
     [EventType] SMALLINT NOT NULL,
     [Configuration] VARCHAR (MAX) NULL,
-    [Filters] VARCHAR (MAX) NULL,
     [Template] VARCHAR (MAX) NULL,
     [CreationDate] DATETIME2 (7) NOT NULL,
     [RevisionDate] DATETIME2 (7) NOT NULL,
+    [Filters] VARCHAR (MAX) NULL,
     CONSTRAINT [PK_OrganizationIntegrationConfiguration] PRIMARY KEY CLUSTERED ([Id] ASC),
     CONSTRAINT [FK_OrganizationIntegrationConfiguration_OrganizationIntegration] FOREIGN KEY ([OrganizationIntegrationId]) REFERENCES [dbo].[OrganizationIntegration] ([Id]) ON DELETE CASCADE
 );


src/Api/AdminConsole/Models/Request/Organizations/OrganizationIntegrationConfigurationRequestModel.cs
```diff
@@ -24,9 +26,13 @@ public bool IsValidForType(IntegrationType integrationType)
             case IntegrationType.CloudBillingSync or IntegrationType.Scim:
                 return false;
             case IntegrationType.Slack:
-                return !string.IsNullOrWhiteSpace(Template) && IsConfigurationValid<SlackIntegrationConfiguration>();
+                return !string.IsNullOrWhiteSpace(Template) &&
```
withinfocus, line 29: ♻️ I just noticed that this checks for whitespace and not empty ... I would hope we never store whitespace. Nothing to do here I suppose, but it's also an unordinary check.

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationFilterOperation.cs
```diff
@@ -3,8 +3,8 @@ namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 
 public enum IntegrationFilterOperation
 {
-    Equals,
-    NotEquals,
-    In,
-    NotIn
+    Equals = 1,
```
withinfocus: ❓ Why is this needed? You _want_ to start at 1?

src/Sql/dbo/Stored Procedures/OrganizationIntegrationConfiguration_Create.sql
```diff
@@ -3,6 +3,7 @@
     @OrganizationIntegrationId UNIQUEIDENTIFIER,
     @EventType SMALLINT,
     @Configuration VARCHAR(MAX),
+    @Filters VARCHAR(MAX),
```
withinfocus: ❌ New columns always get added to the end of procs, really just in case they are invoked in a not-parameterized way. We don't do that here, but I am conditioned to it 😜. You should also support nullability so add a `= NULL` to the parameter; this is in case the proc is called before app logic is updated to supply it which is possible when we deploy database changes first.

src/Sql/dbo/Stored Procedures/OrganizationIntegrationConfiguration_Update.sql
```diff
@@ -3,6 +3,7 @@
     @OrganizationIntegrationId UNIQUEIDENTIFIER,
     @EventType SMALLINT,
     @Configuration VARCHAR(MAX),
+    @Filters VARCHAR(MAX),
```
withinfocus: ❌ Same comments as the above.

util/Migrator/DbScripts/2025-06-19_00_AddFiltersToOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,98 @@
+/* add new column "Filters", nullable to OrganizationIntegrationConfiguration */
+
+ALTER TABLE [dbo].[OrganizationIntegrationConfiguration] ADD [Filters] VARCHAR (MAX) NULL
```
withinfocus: ❌ You should check if the column is missing first, in case this script runs more than once. You can lift that from another migration that adds a column.

util/Migrator/DbScripts/2025-06-19_00_AddFiltersToOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,98 @@
...
+
+/* add column "Filters" to OrganizationIntegrationConfigurationView */
+
+IF EXISTS(SELECT *
+          FROM sys.views
+          WHERE [Name] = 'OrganizationIntegrationConfigurationDetailsView')
+    BEGIN
+        DROP VIEW [dbo].[OrganizationIntegrationConfigurationDetailsView];
+    END
+GO
```
withinfocus: 🎨 You can just `CREATE OR ALTER VIEW`.

src/Api/AdminConsole/Models/Request/Organizations/OrganizationIntegrationConfigurationRequestModel.cs
brant-livefront: Yeah this check is to _fail_ if they don't specify Template (or send us an empty string). This would avoid us storing whitespace Template - which is nullable on the DB, but we're adding this check because for this integration type it makes no sense (how do you send an empty slack message?).

The Filters check below is the inverse - you can specify nothing and we'll accept that. But if you do send a string, it needs to be a valid `IntegrationFilterGroup`. Which, actually, as I type this I'm thinking we should change that check to simply is not null and allow empty filter strings to fail rather than store them.

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationFilterOperation.cs
brant-livefront: Sorry, I had looked at a couple Enums that did start at one and assumed 1-index (and I think that's what it was defaulting to when I didn't specify). I'll adjust to 0-indexed.

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationFilterOperation.cs
```diff
@@ -3,8 +3,8 @@ namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
 
 public enum IntegrationFilterOperation
 {
-    Equals = 1,
-    NotEquals = 2,
-    In = 3,
-    NotIn = 4
+    Equals = 0,
```
withinfocus, line 6: ⛏️ You don't need to set these at all if you want the default 0-start.

src/Sql/dbo/Stored Procedures/OrganizationIntegrationConfiguration_Create.sql
```diff
@@ -16,6 +17,7 @@ BEGIN
         [OrganizationIntegrationId],
         [EventType],
         [Configuration],
+        [Filters],
```
withinfocus: ⛏️ Do the same shift to the end for everything.

src/Sql/dbo/Views/OrganizationIntegrationConfigurationDetailsView.sql
```diff
@@ -4,6 +4,7 @@ AS
         oi.[OrganizationId],
         oi.[Type] AS [IntegrationType],
         oic.[EventType],
+        oic.[Filters],
```
withinfocus: ⛏️ Same for these even. Helps with readability and seeing the changes over time too.

util/Migrator/DbScripts/2025-06-19_00_AddFiltersToOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,95 @@
...
+    BEGIN
+        ALTER TABLE
+            [dbo].[OrganizationIntegrationConfiguration]
+            ADD
+                [Filters] VARCHAR (MAX) NULL
+    END
+GO
+
+/* add column "Filters" to OrganizationIntegrationConfigurationView */
+CREATE OR ALTER VIEW [dbo].[OrganizationIntegrationConfigurationDetailsView]
```
withinfocus, line 13: ❌ I missed it last round but you need to do this for `OrganizationIntegrationConfigurationView` (the traditional view for the table) too.

util/Migrator/DbScripts/2025-06-19_00_AddFiltersToOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,95 @@
...
+    [dbo].[OrganizationIntegration] oi ON oi.[Id] = oic.[OrganizationIntegrationId]
+GO
+
+/* add column to OrganizationIntegrationConfiguration_Create */
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationIntegrationConfiguration_Create]
+    @Id UNIQUEIDENTIFIER OUTPUT,
+    @OrganizationIntegrationId UNIQUEIDENTIFIER,
+    @EventType SMALLINT,
+    @Configuration VARCHAR(MAX),
+    @Filters VARCHAR(MAX),
```
withinfocus: ❌ You missed propagating all your updates to the script.

src/Core/AdminConsole/Entities/OrganizationIntegrationConfiguration.cs
```diff
@@ -12,6 +12,7 @@ public class OrganizationIntegrationConfiguration : ITableObject<Guid>
     public Guid OrganizationIntegrationId { get; set; }
     public EventType EventType { get; set; }
     public string? Configuration { get; set; }
+    public string? Filters { get; set; }
```
withinfocus: ⛏️ Even this is a place where I like them at the end.

util/Migrator/DbScripts/2025-06-19_00_AddFiltersToOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,104 @@
...
+/* add column to OrganizationIntegrationConfiguration_Create */
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationIntegrationConfiguration_Create]
+    @Id UNIQUEIDENTIFIER OUTPUT,
+    @OrganizationIntegrationId UNIQUEIDENTIFIER,
+    @EventType SMALLINT,
+    @Configuration VARCHAR(MAX),
+    @Template VARCHAR(MAX),
+    @CreationDate DATETIME2(7),
+    @RevisionDate DATETIME2(7),
+    @Filters VARCHAR(MAX)
```
withinfocus: ❌ You missed the null assignments like you have in the main files. I like to just copy and paste the above file contents into scripts so nothing is missed.

src/Sql/dbo/Tables/OrganizationIntegrationConfiguration.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE [dbo].[OrganizationIntegrationConfiguration]
     [OrganizationIntegrationId] UNIQUEIDENTIFIER NOT NULL,
     [EventType] SMALLINT NOT NULL,
     [Configuration] VARCHAR (MAX) NULL,
+    [Filters] VARCHAR (MAX) NULL,
```
withinfocus: 😬 Just noticed this one too, which should move to the end.

withinfocus (changes requested):
One last thing!

PR #5992, opened 2025-06-20
[PM-22219] - [Vault] [Server] Exclude items in default collections from Admin Console

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22219

## 📔 Objective

This PR updates the `GetOrganizationCiphers` action in the ciphers controller to filter out the default user collections if the `ExcludeDefaultUserCollections` flag is enabled.

Summary of changes:

- add feature flag `ExcludeDefaultUserCollections`
- add sproc `CipherOrganizationDetails_ReadByOrganizationIdExcludingDefaultCollections` that gathers all the required ciphers and data and excludes the default collection ciphers

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Vault/Queries/OrganizationCiphersQuery.cs
```diff
@@ -61,4 +64,28 @@ public async Task<IEnumerable<CipherOrganizationDetailsWithCollections>> GetOrga
...
+             .GetManyByOrganizationIdAsync(orgId))
+         .GroupBy(cc => cc.CipherId)
+         .ToDictionary(g => g.Key, g => g);
+
+        return (await _cipherRepository
+                .GetManyOrganizationDetailsByOrganizationIdAsync(orgId))
+            .Where(c =>
+                // select ciphers with no collections or none of its collections are default
+                !cipherGroups.TryGetValue(c.Id, out var grp)
+                || !grp.Any(cc => defaultCollIds.Contains(cc.CollectionId))
```
shane-melton: ⚠️ We should only exclude the cipher if it _only_ belongs to default collection. If it belongs to multiple, then it should be included / visible.

(I think switching to `!grp.All(...)` should work, but there may be a better way)

src/Core/Vault/Queries/OrganizationCiphersQuery.cs
```diff
@@ -61,4 +64,28 @@ public async Task<IEnumerable<CipherOrganizationDetailsWithCollections>> GetOrga
         var allOrganizationCiphers = await GetAllOrganizationCiphers(organizationId);
         return allOrganizationCiphers.Where(c => c.CollectionIds.Intersect(managedCollectionIds).Any());
     }
+
+    public async Task<IEnumerable<CipherOrganizationDetailsWithCollections>>
+        GetAllOrganizationCiphersExcludingDefaultUserCollections(Guid orgId)
+    {
+        var defaultCollIds = (await _collectionRepository
```
shane-melton: 💭 🎨  It would probably be beneficial to put this logic in its own sproc to avoid pulling all ciphers, collection ids, and collectionCiphers for the organization from the database every time. This method will be called semi frequently for admins in the AC.

src/Api/Vault/Controllers/CiphersController.cs
```diff
@@ -306,8 +309,11 @@ public async Task<ListResponseModel<CipherMiniDetailsResponseModel>> GetOrganiza
         {
             throw new NotFoundException();
         }
-
-        var allOrganizationCiphers = await _organizationCiphersQuery.GetAllOrganizationCiphers(organizationId);
+        var allOrganizationCiphers = _featureService.IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
```
shane-melton, line 320: 💭 I know this is the only flag we have at the moment, but the name doesn't seem to properly reflect the behavior being modified here. We might want to create our own Vault flag related to these item organization changes.

shane-melton (changes requested):

jaasen-livefront:
@shane-melton Thanks for the feedback! Ready for another look when you have a moment.

src/Sql/dbo/Stored Procedures/CollectionCipher_GetOrganizationDetailsWithCollections.sql
shane-melton, line 1: 🎨 Sorry, I probably should have been more clear. I think we should have the sproc perform the logic to filter out the ciphers that only belong to default collection. (e.g. using a `CTE` or `JOIN`s on the Collection/CollectionCipher tables)

jaasen-livefront: @shane-melton Ah ok gotcha. Updated. I kept the other sproc as I think it's still useful. Let me know if you feel otherwise.

shane-melton: I think there may still be some confusion. We should have a sproc like `CipherOrganizationDetails_ReadByOrganizationIdWithoutDefaultCollections` that queries all the ciphers that belong to an organization AND excludes those ciphers only belong to default collections. 

Then the controller/query only has to make a single DB call and it gets exactly the result it needs to return without any filtering/rearranging at the API level.

rkac-bw: Maybe something like this? Would need to verify logic and test

 ```
-- View that provides organization cipher details with their collection associations
CREATE OR ALTER VIEW [dbo].[OrganizationCipherDetailsWithCollectionsView]
AS
    SELECT
        C.[Id],
        C.[UserId],
        C.[OrganizationId],
        C.[Type],
        C.[Data],
        C.[Attachments],
        C.[CreationDate],
        C.[RevisionDate],
        C.[DeletedDate],
        C.[Reprompt],
        C.[Key],
        CASE
            WHEN O.[UseTotp] = 1 THEN 1
            ELSE 0
        END AS [OrganizationUseTotp],
        CC.[CollectionId],          -- NULL when cipher has no collections (due to LEFT JOIN)
        COL.[Type] AS [CollectionType]  -- NULL when cipher has no collections OR when collection has no type set
    FROM [dbo].[Cipher] C
    INNER JOIN [dbo].[Organization] O ON C.[OrganizationId] = O.[Id]
    LEFT JOIN [dbo].[CollectionCipher] CC ON CC.[CipherId] = C.[Id]  -- LEFT JOIN ensures ciphers without collections are included
    LEFT JOIN [dbo].[Collection] COL ON CC.[CollectionId] = COL.[Id] -- LEFT JOIN to get collection details; NULL when no collection
    WHERE C.[UserId] IS NULL; -- Organization ciphers only (exclude user-owned ciphers)
GO


CREATE OR ALTER PROCEDURE
[dbo].[CipherOrganizationDetails_ReadByOrganizationIdExcludingDefaultCollections]
    @OrganizationId UNIQUEIDENTIFIER
AS
BEGIN
    SET NOCOUNT ON;

    WITH CiphersWithNoCollections AS (
        -- Ciphers with no collections at all
        SELECT
            [Id], [UserId], [OrganizationId], [Type], [Data], [Attachments],
            [CreationDate], [RevisionDate], [DeletedDate], [Reprompt], [Key], [OrganizationUseTotp]
        FROM [dbo].[OrganizationCipherDetailsWithCollectionsView]
        WHERE [OrganizationId] = @OrganizationId
            AND [CollectionId] IS NULL
    ),
    CiphersWithNullTypeCollections AS (
        -- Ciphers with collections that have NULL type (treated as non-default)
        SELECT
            [Id], [UserId], [OrganizationId], [Type], [Data], [Attachments],
            [CreationDate], [RevisionDate], [DeletedDate], [Reprompt], [Key], [OrganizationUseTotp]
        FROM [dbo].[OrganizationCipherDetailsWithCollectionsView]
        WHERE [OrganizationId] = @OrganizationId
            AND [CollectionId] IS NOT NULL
            AND [CollectionType] IS NULL
    ),
    CiphersWithNonDefaultCollections AS (
        -- Ciphers with collections that are explicitly non-default (Type != 1)
        SELECT
            [Id], [UserId], [OrganizationId], [Type], [Data], [Attachments],
            [CreationDate], [RevisionDate], [DeletedDate], [Reprompt], [Key], [OrganizationUseTotp]
        FROM [dbo].[OrganizationCipherDetailsWithCollectionsView]
        WHERE [OrganizationId] = @OrganizationId
            AND [CollectionId] IS NOT NULL
            AND [CollectionType] IS NOT NULL
            AND [CollectionType] != 1
    )
    SELECT DISTINCT
        [Id], [UserId], [OrganizationId], [Type], [Data], [Attachments],
        [CreationDate], [RevisionDate], [DeletedDate], [Reprompt], [Key], [OrganizationUseTotp]
    FROM (
        SELECT * FROM CiphersWithNoCollections
        UNION ALL  -- Use UNION ALL since we'll DISTINCT at the end
        SELECT * FROM CiphersWithNullTypeCollections
        UNION ALL
        SELECT * FROM CiphersWithNonDefaultCollections
    ) AS AllCiphers;
END
GO
```

Comment (rkac-bw) [reply src/Sql/dbo/Stored Procedures/CollectionCipher_GetOrganizationDetailsWithCollections.sql:1]: A filtered index for org only ciphers might help
```
CREATE INDEX IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly
ON [dbo].[Cipher] ([OrganizationId]) 
INCLUDE ([Id], [Type], [Data], [Attachments], [CreationDate], [RevisionDate], [DeletedDate], [Reprompt], [Key])
WHERE [UserId] IS NULL;
```

Comment (jaasen-livefront) [reply src/Sql/dbo/Stored Procedures/CollectionCipher_GetOrganizationDetailsWithCollections.sql:1]: @rkac-bw @shane-melton Thanks for the feedback. I've got a solution that I think should suffice. Let me know what you think!

File: src/Sql/dbo/Stored Procedures/CollectionCipher_ReadByOrganizationIdExcludingDefaultCollections.sql [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+        C.[UserId],
+        C.[OrganizationId],
+        C.[Type],
+        C.[Data],
+        C.[Attachments],
+        C.[CreationDate],
+        C.[RevisionDate],
+        C.[DeletedDate],
+        C.[Reprompt],
+        C.[Key],
```
rkac-bw: So following our standard - Stored procedures should select from views and views select from tables, and this seems to duplicate the logic in the view 
Separation of Concerns

  Views (Complex Logic Layer):
  - Contains JOIN logic, business rules, calculated fields
  - Defines "what data looks like" - the data contract
  - Can be reused by multiple procedures
  - Changes here don't break procedure signatures

  Procedures (Simple Access Layer):
  - Contains only parameterized filtering (WHERE clauses)
  - Defines "which data to get" - the access pattern
  - Focuses on security context and performance
  - Changes here don't affect data structure
  
   Both the view and stored procedure have the same join logic

jaasen-livefront: @rkac-bw Ah yes ofc. Oversight on my part. I've trimmed out the unnecessary joins from the proc.

jaasen-livefront:
@rkac-bw @shane-melton Ready for re-review!

rkac-bw (approved):
LGTM

src/Sql/dbo/Views/OrganizationCipherDetailsCollectionsView.sql
```diff
@@ -0,0 +1,26 @@
...
+          ELSE 0
+      END AS [OrganizationUseTotp],
+      CC.[CollectionId],
+      COL.[Type] AS [CollectionType]
+    FROM [dbo].[Cipher] C
+    INNER JOIN [dbo].[Organization] O ON C.[OrganizationId] = O.[Id]
+    LEFT JOIN [dbo].[CollectionCipher] CC ON CC.[CipherId] = C.[Id]
+    LEFT JOIN [dbo].[Collection] COL ON CC.[CollectionId] = COL.[Id]
+    WHERE C.[UserId] IS NULL; -- Organization ciphers only
+GO
```
shane-melton: ⛏️ We shouldn't have a `GO` statement at the end of these view definitions. (Typically the migration scripts need those final `GO` statements)

util/Migrator/DbScripts/2025-07-02_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,61 @@
...
+        OrganizationUseTotp,
+        CollectionId
+    FROM dbo.OrganizationCipherDetailsWithCollectionsView
+    WHERE OrganizationId = @OrganizationId
+      AND (CollectionId IS NULL       -- no collections
+           OR CollectionType <> 1);  -- or at least one non-default
+END;
+GO
+
+GO
```
shane-melton: ⛏️ Double `GO` statements, and we should have one at the end of the migration script.

src/Core/Constants.cs
```diff
@@ -206,6 +206,7 @@ public static class FeatureFlagKeys
     public const string PhishingDetection = "phishing-detection";
     public const string RemoveCardItemTypePolicy = "pm-16442-remove-card-item-type-policy";
     public const string PM22134SdkCipherListView = "pm-22134-sdk-cipher-list-view";
+    public const string ExcludeDefaultUserCollections = "pm-19467-create-default-location";
```
shane-melton: 🎨 I know I originally suggested in this PR that we might want a separate flag, but we've already used AC's `CreateDefaultLocation` flag for all the other Vault client work. We should continue using it here as well and revert creating this new flag for consistency and less combinations for QA to test.

Apologies for the back and forth!

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadByOrganizationIdExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,27 @@
...
+        Reprompt,
+        [Key],
+        OrganizationUseTotp,
+        CollectionId
+    FROM dbo.OrganizationCipherDetailsWithCollectionsView
+    WHERE OrganizationId = @OrganizationId
+      AND (CollectionId IS NULL       -- no collections
+           OR CollectionType <> 1);  -- or at least one non-default
+END;
+GO
```
shane-melton: ⛏️ No need for `GO` statement in definition

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadByOrganizationIdExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,27 @@
...
+        [Type],
+        [Data],
+        [Attachments],
+        CreationDate,
+        RevisionDate,
+        DeletedDate,
+        Reprompt,
+        [Key],
+        OrganizationUseTotp,
+        CollectionId
```
shane-melton: ⛏️ We should be consistent with wrapping these columns with brackets `[]`

src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs
```diff
@@ -653,6 +653,47 @@ await connection.ExecuteAsync(
...
+                if (!dict.TryGetValue(cipher.Id, out var details))
+                {
+                    details = new CipherOrganizationDetailsWithCollections(cipher, /*dummy*/null);
+                    dict.Add(cipher.Id, details);
+                    tempCollections[cipher.Id] = new List<Guid>();
+                }
+
+                if (cc?.CollectionId != null)
+                {
+                    var list = tempCollections[cipher.Id];
```
shane-melton: 🎨 We have a `AddIfNotExists()` helper.
e.g. `tempCollections[cipher.Id].AddIfNotExists(cc.CollectionId)`

src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs
```diff
@@ -653,6 +653,47 @@ await connection.ExecuteAsync(
...
+
+                return details;
+            },
+            new { OrganizationId = orgId },
+            splitOn: "CollectionId",
+            commandType: CommandType.StoredProcedure
+        );
+
+        // now assign each List<Guid> back to the array property in one shot
+        foreach (var kv in dict)
```
shane-melton, line 903: ⛏️ Our [C# code styles](https://contributing.bitwarden.com/contributing/code-style/csharp#control-blocks) request curly braces for all control blocks

src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs
```diff
@@ -653,6 +653,47 @@ await connection.ExecuteAsync(
...
+                {
+                    var list = tempCollections[cipher.Id];
+                    if (!list.Contains(cc.CollectionId))
+                        list.Add(cc.CollectionId);
+                }
+
+                return details;
+            },
+            new { OrganizationId = orgId },
+            splitOn: "CollectionId",
```
shane-melton, line 898: 👍 I forgot about this Dapper capability, nice one!

util/Migrator/DbScripts/2025-07-02_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
shane-melton, line 1: ⛏️ Can we update the migration script date since its a few weeks out of date now.

shane-melton (changes requested):
This is looking really good, nice work! Most comments are just small nits/style requests. 
I do think we should revert adding a new feature flag just for this, since we made all the other vault changes behind the existing AC flag.

Lastly, it would be beneficial to add some [integration tests for the CipherRepository ](https://github.com/bitwarden/server/blob/8bccf255c08b68bb97568a81d3f6a6bb84112439/test/Infrastructure.IntegrationTest/Vault/Repositories/CipherRepositoryTests.cs#L19)method added; especially since its doing some more complicated logic than simply retrieving the results of the DB sprocs.

util/Migrator/DbScripts/2025-07-02_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
jaasen-livefront: @shane-melton You bet!

jaasen-livefront:
@rkac-bw @shane-melton One last little tweak, updated the file name to today and added one more layer of filtering to the view:

`AND O.[Enabled] = 1; -- Only enabled organizations`

src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs
jaasen-livefront: @shane-melton It's pretty great with one to many relationships! Saves from having to do multiple queries.

jaasen-livefront:
@rkac-bw  @shane-melton Should be ready for another look when you have time. Ty!

src/Sql/dbo/Views/OrganizationCipherDetailsCollectionsView.sql
```diff
@@ -0,0 +1,26 @@
+CREATE VIEW [dbo].[OrganizationCipherDetailsWithCollectionsView]
+AS
```
rkac-bw, line 2: I believe The Dapper mapper expects  which has these properties - folders and favourites which are needed for CipherOrganizationDetails model (the base class inherits from Cipher)

```
CREATE VIEW [dbo].[OrganizationCipherDetailsWithCollectionsView]
  AS
      SELECT
        C.[Id],
        C.[UserId],
        C.[OrganizationId],
        C.[Type],
        C.[Data],
        C.[Favorites],     -- Required by CipherOrganizationDetails model
        C.[Folders],       -- Required by CipherOrganizationDetails model
        C.[Attachments],
        C.[CreationDate],
        C.[RevisionDate],
        C.[DeletedDate],
        C.[Reprompt],
        C.[Key],
        CASE
            WHEN O.[UseTotp] = 1 THEN 1
            ELSE 0
        END AS [OrganizationUseTotp],
        CC.[CollectionId],
        COL.[Type] AS [CollectionType]
      FROM [dbo].[Cipher] C
      INNER JOIN [dbo].[Organization] O ON C.[OrganizationId] = O.[Id]
      LEFT JOIN [dbo].[CollectionCipher] CC ON CC.[CipherId] = C.[Id]
      LEFT JOIN [dbo].[Collection] COL ON CC.[CollectionId] = COL.[Id]
      WHERE C.[UserId] IS NULL -- Organization ciphers only
        AND O.[Enabled] = 1; -- Only enabled organizations
```

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadByOrganizationIdExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,27 @@
+ -- Stored procedure that filters out ciphers that ONLY belong to default collections
+CREATE PROCEDURE [dbo].[CipherOrganizationDetails_ReadByOrganizationIdExcludingDefaultCollections]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON;
+
```
rkac-bw: Better performance to pre filter collection type - added folder and favorites which i believe the model needs:

```
CREATE OR ALTER PROCEDURE
  [dbo].[CipherOrganizationDetails_ReadByOrganizationIdExcludingDefaultCollections]
      @OrganizationId UNIQUEIDENTIFIER
  AS
  BEGIN
      SET NOCOUNT ON;

      -- Use CTE to identify ciphers that should be included
      WITH [NonDefaultCiphers] AS (
          SELECT DISTINCT [Id]
          FROM [dbo].[OrganizationCipherDetailsWithCollectionsView]
          WHERE [OrganizationId] = @OrganizationId
            AND ([CollectionId] IS NULL    -- Include ciphers with no collections
                 OR [CollectionType] <> 1)  -- Include if at least one non-default collection
      )
      -- Return cipher details with collection associations for Dapper multi-mapping
      SELECT
          V.[Id],
          V.[UserId],
          V.[OrganizationId],
          V.[Type],
          V.[Data],
          V.[Favorites],
          V.[Folders],
          V.[Attachments],
          V.[CreationDate],
          V.[RevisionDate],
          V.[DeletedDate],
          V.[Reprompt],
          V.[Key],
          V.[OrganizationUseTotp],
          V.[CollectionId]  -- For Dapper splitOn parameter
      FROM [dbo].[OrganizationCipherDetailsWithCollectionsView] V
      INNER JOIN [NonDefaultCiphers] NDC ON V.[Id] = NDC.[Id]
      WHERE V.[OrganizationId] = @OrganizationId
      ORDER BY V.[RevisionDate] DESC;
  END;
  GO
```

src/Sql/dbo/Vault/Tables/Cipher.sql
```diff
@@ -34,3 +34,8 @@ GO
 CREATE NONCLUSTERED INDEX [IX_Cipher_DeletedDate]
     ON [dbo].[Cipher]([DeletedDate] ASC);
 
+GO
+CREATE NONCLUSTERED INDEX [IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly]
+    ON [dbo].[Cipher] ([OrganizationId])
```
rkac-bw: update to include folder and favourites:
```
CREATE NONCLUSTERED INDEX IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly
      ON [dbo].[Cipher] ([OrganizationId])
      INCLUDE ([Id], [Type], [Data], [Favorites], [Folders], [Attachments], [CreationDate],
  [RevisionDate], [DeletedDate], [Reprompt], [Key])
      WHERE [UserId] IS NULL;
```

jaasen-livefront:
@rkac-bw Thanks for the feedback! Resolved in https://github.com/bitwarden/server/pull/5992/commits/6155089022a2b0fcd963c8cc7a613e4fd69c0cdc

shane-melton (changes requested):
Looks good! Though we should update the migration script date again.

jaasen-livefront:
> Looks good! Though we should update the migration script date again.

Done!

shane-melton (dismissed):
Thank you!

util/Migrator/DbScripts/2025-08-06_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,75 @@
...
+          V.[CreationDate],
+          V.[RevisionDate],
+          V.[DeletedDate],
+          V.[Reprompt],
+          V.[Key],
+          V.[OrganizationUseTotp],
+          V.[CollectionId]  -- For Dapper splitOn parameter
+      FROM [dbo].[OrganizationCipherDetailsWithCollectionsView] V
+      INNER JOIN [NonDefaultCiphers] NDC ON V.[Id] = NDC.[Id]
+      WHERE V.[OrganizationId] = @OrganizationId
```
rkac-bw: Does this need to match EF Implementation (line 1006-1007) and filter on AND (V.[CollectionId] IS NULL OR V.[CollectionType] <> 1) ?

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadByOrganizationIdExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,38 @@
+ -- Stored procedure that filters out ciphers that ONLY belong to default collections
+CREATE PROCEDURE
+  [dbo].[CipherOrganizationDetails_ReadByOrganizationIdExcludingDefaultCollections]
```
rkac-bw: File name should be CipherOrganizationDetails_ReadByOrganizationIdExcludingDefaultCollections.sql?

util/Migrator/DbScripts/2025-08-06_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
jaasen-livefront: @rkac-bw Yes I believe that should be the case. I've updated the sproc and migration script.
https://github.com/bitwarden/server/pull/5992/commits/97c8bde59d1dc0528fe7672ed15287f51583bb45

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadByOrganizationIdExcludingDefaultCollections.sql
jaasen-livefront: Done!

util/Migrator/DbScripts/2025-08-15_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,76 @@
...
+          V.[CollectionId]  -- For Dapper splitOn parameter
+      FROM [dbo].[OrganizationCipherDetailsWithCollectionsView] V
+      INNER JOIN [NonDefaultCiphers] NDC ON V.[Id] = NDC.[Id]
+      WHERE V.[OrganizationId] = @OrganizationId
+      AND (V.[CollectionId] IS NULL OR V.[CollectionType] <> 1)
+      ORDER BY V.[RevisionDate] DESC;
+  END;
+  GO
+
+CREATE NONCLUSTERED INDEX IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly
```
rkac-bw: Needs to be idempotent, should check if exists in case it needs to rerun:
```
IF NOT EXISTS (
    SELECT 1 
    FROM sys.indexes 
    WHERE name = 'IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly' 
    AND object_id = OBJECT_ID('dbo.Cipher')
)
```

src/Api/Vault/Controllers/CiphersController.cs
rkac-bw: So the PR mentions creating it's own flag - ExcludeDefaultUserCollections , why are we not doing that?

  If the feature flag is disabled after default collections already exist:
  - Would personal collections would suddenly become visible in org views?
  - No way to roll back safely?

src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
```diff
@@ -965,6 +966,54 @@ public UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(
...
+        var dbContext = GetDatabaseContext(scope);
+
+        var defaultTypeInt = (int)CollectionType.DefaultUserCollection;
+
+        //  filter out any cipher that belongs *only* to default collections
+        //    i.e. keep ciphers with no collections, or with ≥1 non-default collection
+        var query = from c in dbContext.Ciphers.AsNoTracking()
+                    where c.UserId == null
+                       && c.OrganizationId == organizationId
+                       && (
```
rkac-bw, line 1019: Should this equal sql logic?
`&& c.Organization.Enabled` 

from line 28 in view?
`AND O.[Enabled] = 1; -- Only enabled organizations`

util/Migrator/DbScripts/2025-08-15_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
jaasen-livefront: Done! https://github.com/bitwarden/server/pull/5992/commits/9794cbf314880e18263427306765c2ea290dce6a

src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
jaasen-livefront: Agree! Done. https://github.com/bitwarden/server/pull/5992/commits/9794cbf314880e18263427306765c2ea290dce6a

src/Sql/dbo/Vault/Tables/Cipher.sql
```diff
@@ -34,3 +34,9 @@ GO
 CREATE NONCLUSTERED INDEX [IX_Cipher_DeletedDate]
     ON [dbo].[Cipher]([DeletedDate] ASC);
 
+GO
+CREATE NONCLUSTERED INDEX IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly
+      ON [dbo].[Cipher] ([OrganizationId])
```
rkac-bw: Should use square brackets
[IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly]

jaasen-livefront: Done! https://github.com/bitwarden/server/pull/5992/commits/6bf2dc981fbecabe77f979663453d57df1b2ae46

src/Sql/dbo/Vault/Tables/Cipher.sql
```diff
@@ -34,3 +34,9 @@ GO
...
+      INCLUDE ([Id], [Type], [Data], [Favorites], [Folders], [Attachments], [CreationDate],
+[RevisionDate], [DeletedDate], [Reprompt], [Key])
```
rkac-bw: Contributing guide says no Mid-list line break - Guide says either single line OR vertical (not partial) for `INCLUDE` list

util/Migrator/DbScripts/2025-09-03_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
```diff
@@ -0,0 +1,86 @@
...
+    SELECT 1
+    FROM sys.indexes
+    WHERE name = 'IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly'
+    AND object_id = OBJECT_ID('dbo.Cipher')
+)
+BEGIN
+    CREATE NONCLUSTERED INDEX [IX_Cipher_OrganizationId_Filtered_OrgCiphersOnly]
+          ON [dbo].[Cipher] ([OrganizationId])
+          INCLUDE ([Id], [Type], [Data], [Favorites], [Folders], [Attachments], [CreationDate],
+    [RevisionDate], [DeletedDate], [Reprompt], [Key])
```
rkac-bw: Same issue here either 1 line or 1 item per line for include list

src/Sql/dbo/Vault/Tables/Cipher.sql
jaasen-livefront: Fixed! Along with a couple other fixes. 
https://github.com/bitwarden/server/pull/5992/commits/aa4c03e21cb230dc622cc07bc2f92f3fd9aeb9a9

util/Migrator/DbScripts/2025-09-03_00_CipherOrganizationDetailsExcludingDefaultCollections.sql
jaasen-livefront: Fixed in https://github.com/bitwarden/server/pull/5992/commits/aa4c03e21cb230dc622cc07bc2f92f3fd9aeb9a9

src/Sql/dbo/Views/OrganizationCipherDetailsCollectionsView.sql
```diff
@@ -0,0 +1,28 @@
...
+    SELECT
```
rkac-bw, line 3: Filename mismatch:
  - Filename: OrganizationCipherDetailsCollectionsView.sql (missing "With")
  - SQL Object: [dbo].[OrganizationCipherDetailsWithCollectionsView] (contains "With")

jaasen-livefront: @rkac-bw Nice catch! I renamed the sql view object.
https://github.com/bitwarden/server/pull/5992/commits/f6ae4ee0ff17d89249bf9bf50afdcde5c129a920

shane-melton (approved):

PR #5995, opened 2025-06-23
[PM-22862] Account security version

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22862

## 📔 Objective

Implement storage for account security state. This contains:
- A base64 encoded, signed account security state
- The contained account security version

the latter is meant for server side operations, without having to implement COSE parsing for the server or database queries.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/KeyManagement/Queries/UserAccountKeysQuery.cs
```diff
@@ -12,11 +12,25 @@ public class UserAccountKeysQuery(IUserSignatureKeyPairRepository signatureKeyPa
...
+        else
+        {
+            return new UserAccountKeysData
+            {
+                PublicKeyEncryptionKeyPairData = user.GetPublicKeyEncryptionKeyPair(),
+                SignatureKeyPairData = await signatureKeyPairRepository.GetByUserIdAsync(user.Id),
+                SecurityStateData = new SecurityStateData
+                {
+                    SecurityState = user.SecurityState!,
+                    SecurityVersion = (int)user.SecurityVersion!,
```
mzieniukbw: ❓  Should we use `GetSecurityVersion()` instead ? That way we can make `SecurityVersion` required and consolidate with `SecurityStateModel`

src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs
```diff
@@ -104,13 +104,15 @@ public async Task RotateV2AccountKeysAsync(RotateUserAccountKeysData model, User
...
         user.SignedPublicKey = model.AccountKeys.PublicKeyEncryptionKeyPairData.SignedPublicKey;
+        user.SecurityState = model.AccountKeys.SecurityStateData!.SecurityState;
     }
 
     public void UpgradeV1ToV2Keys(RotateUserAccountKeysData model, User user, List<UpdateEncryptedDataForKeyRotation> saveEncryptedDataActions)
     {
         ValidateV2Encryption(model);
         saveEncryptedDataActions.Add(_userSignatureKeyPairRepository.SetUserSignatureKeyPair(user.Id, model.AccountKeys.SignatureKeyPairData));
         user.SignedPublicKey = model.AccountKeys.PublicKeyEncryptionKeyPairData.SignedPublicKey;
+        user.SecurityState = model.AccountKeys.SecurityStateData!.SecurityState;
```
mzieniukbw, line 118: ❓ Shouldn't we validate this to never be null within `ValidateV2Encryption` ?

src/Api/KeyManagement/Models/Requests/SecurityStateModel.cs
```diff
@@ -0,0 +1,28 @@
+﻿#nullable enable
+using Bit.Core.KeyManagement.Models.Data;
+
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+public class SecurityStateModel
```
mzieniukbw, line 7: ❓ It is weird that we have both `SecurityStateModel` and `SecurityStateData`, that only differs by one field being validated, where other is not. Additionally, this model is used for request and response. Maybe it will be better to consolidate to one or change name of this class to be more explicit ?

src/Core/Entities/User.cs
```diff
@@ -61,6 +61,15 @@ public class User : ITableObject<Guid>, IStorableSubscriber, IRevisable, ITwoFac
...
     /// </summary>
     public string? SignedPublicKey { get; set; }
+    /// <summary>
+    /// The security version is included in the security state, but needs COSE parsing
+    /// </summary>
+    public int? SecurityVersion { get; set; }
+    /// <summary>
+    /// The security state is a signed object attesting to the version of the user's account.
+    /// </summary>
+    [MaxLength(1000)]
```
mzieniukbw: I am worried this is not enough to fit in the payload we want to pass on, hiding under COSE structured state.
I don't know what's big enough, maybe 4000 ? Unless you think 1000 is good enough.

mzieniukbw (changes requested):

src/Core/KeyManagement/Queries/UserAccountKeysQuery.cs
```diff
@@ -12,11 +12,25 @@ public class UserAccountKeysQuery(IUserSignatureKeyPairRepository signatureKeyPa
 {
     public async Task<UserAccountKeysData> Run(User user)
     {
-        var userAccountKeysData = new UserAccountKeysData
+        if (user.GetSecurityVersion() < 2)
```
mzieniukbw, line 15: Unit test coverage ?

mzieniukbw: Do we have any Integration tests too for this to cover ?

src/Core/Entities/User.cs
quexten: Here is a sample signed-security-state: `hFgepAEnAxg8BFDqa6w3IhBBPJ5/umClLr8eOgABOH8CoFgkomhlbnRpdHlJZFBO6OR2vcBJo4ULcY4pt+ZsZ3ZlcnNpb24CWEAJFi/KNSoH95Xg6shmdrhPdyc0RiSIsyd0r4cPiVhnqL52nlq5srdmqojx/AuCrZnFHh4pH0KIpUp1XV9HDEMJ`, which even encoded in inefficient b64 is <200 bytes.

The point where this wouldn't be enough is if we either switch to post-quantum signatures which are 1KB-4KB, or if we expand what the signed security state stores. But I think we can change this if we get to that point?

mzieniukbw: I thought it will be much bigger right now. That's fine then, we can increase it later when needed.

src/Core/KeyManagement/Queries/UserAccountKeysQuery.cs
quexten: I do think we do want to keep data and model separate. We may change the model but may not want to change the data. F.e if we add crypto libraries to the server, we could drop the version from the model, because it can be parsed from the COSE struct in the state. Then the model wouldn't have the version, but the data struct would still have it.

Agree on the GetSecurityVersion though, and updated it.

src/Api/KeyManagement/Models/Requests/SecurityStateModel.cs
quexten: So do we want a requestmodel and responsemodel?

The data is separate for the reason pointed out here: https://github.com/bitwarden/server/pull/5995#discussion_r2213149963

mzieniukbw: Ok, so we want Requests / Responses and (Internal) Data Models separation. That's fine then, i tried to oversimplify too much :)

mzieniukbw (approved):

src/Core/KeyManagement/Queries/UserAccountKeysQuery.cs
quexten: Covered by unit tests now. We also have the `RotateUpgradeToV2UserAccountKeysAsync_Success` integration test.

src/Sql/dbo/Tables/User.sql
```diff
@@ -43,6 +43,8 @@
     [LastEmailChangeDate]              DATETIME2 (7)    NULL,
     [VerifyDevices]                    BIT              DEFAULT ((1)) NOT NULL,
     [SignedPublicKey]                  VARCHAR (MAX)    NULL,
+    [SecurityVersion]                  INT              NULL,
+    [SecurityState]                    NVARCHAR (MAX)   NULL,
```
rkac-bw, line 46: - SQL Server: NVARCHAR(MAX) at `/server/src/Sql/dbo/Tables/User.sql:47`
   - Other DBs: varchar(1000) in EF migrations
   - C# Model: [MaxLength(1000)] at `/server/src/Core/Entities/User.cs:71`
   -Should we Consider changing SQL Server to NVARCHAR(1000) for consistency

util/MySqlMigrations/Migrations/20250623113545_UserSecurityState.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Bit.MySqlMigrations.Migrations;
+
+/// <inheritdoc />
+public partial class UserSecurityState : Migration
+{
+    /// <inheritdoc />
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        migrationBuilder.AddColumn<string>(
```
rkac-bw: - SQL Server migration adds SecurityVersion first, then SecurityState
   - EF migrations add SecurityState first, then SecurityVersion
   
   Can we make them the same ?

src/Sql/dbo/Tables/User.sql
quexten: I think we want to keep this without a limit actually, and limit only on the request model. The length of this field will increase once we get to post-quantum signatures.

Moving this to the model!

util/MySqlMigrations/Migrations/20250623113545_UserSecurityState.cs
quexten: Order is the same now.

rkac-bw (approved):
lgtm

PR #6001, opened 2025-06-24
[PM-12474] Move to authorization to attibutes/handlers/requirements

jrmccannon:
## 🎟️ Tracking
[PM-12474](https://bitwarden.atlassian.net/browse/PM-12474)

## 📔 Objective
This refactor moves our inline authorization code into the middleware stack allowing us to use the `[Authorize<Requirement>]` attribute.

It also removes an unused endpoint (`/organizations/{orgId}/users/{orgUserId}/groups`). 
Here is the client-side [PR](https://github.com/bitwarden/clients/pull/15487) removing the unused call.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-12474]: https://bitwarden.atlassian.net/browse/PM-12474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -230,29 +225,22 @@ private ListResponseModel<OrganizationUserUserDetailsResponseModel> GetResultLis
 
 
     [HttpGet("{id}/groups")]
-    public async Task<IEnumerable<string>> GetGroups(string orgId, string id)
+    [Authorize<ManageGroupsOrUsersRequirement>]
+    public async Task<IEnumerable<string>> GetGroups(string id)
     {
         var organizationUser = await _organizationUserRepository.GetByIdAsync(new Guid(id));
```
eliykat: Can you please update all instances of `string id` or `string orgId` parameters to be `Guid`s while you're here?

src/Api/AdminConsole/Authorization/Requirements/ManageGroupsOrUsersRequirement.cs
eliykat, line 1: `Permissions` are only used for custom users, so you need to expressly check for Owner and Admin roles as well (see `ManageUsersRequirement` as an example).

eliykat (changes requested):

src/Api/AdminConsole/Authorization/Requirements/ManageGroupsOrUsersRequirement.cs
eliykat, line 1: What are your thoughts on unit testing this? It's so simple that it feels like it would be just rewriting the implementation again. But I am on the fence about it.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -146,15 +147,9 @@ public async Task<OrganizationUserDetailsResponseModel> Get(Guid id, bool includ
     }
 
     [HttpGet("mini-details")]
+    [Authorize<MemberOrProviderRequirement>]
     public async Task<ListResponseModel<OrganizationUserUserMiniDetailsResponseModel>> GetMiniDetails(Guid orgId)
```
eliykat, line 162: Can you please add xmldoc on this one - both to clarify its purpose, and to make it clear that it's non-privileged data within the organization (and therefore available to all members). This is because we regularly get H1 tickets to the effect of `YoU havEn't SeCuREd yOUr endPOInts" on endpoints like these.

e.g.
> Returns a set of basic information about all members of the organization. This is available to all members of the organization to manage collections. For this reason, it contains as little information as possible and no cryptographic keys or other sensitive data.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -230,32 +225,23 @@ private ListResponseModel<OrganizationUserUserDetailsResponseModel> GetResultLis
 
 
     [HttpGet("{id}/groups")]
-    public async Task<IEnumerable<string>> GetGroups(string orgId, string id)
+    [Authorize<ManageGroupsOrUsersRequirement>]
+    public async Task<IEnumerable<string>> GetGroups(Guid id)
```
eliykat: This doesn't look like it's used anywhere. Could you please confirm this, and then delete it, along with the client-side code? (Even better than refactoring it!)

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -230,32 +225,23 @@ private ListResponseModel<OrganizationUserUserDetailsResponseModel> GetResultLis
...
-        var orgGuidId = new Guid(orgId);
-        if (!await _currentContext.ManageResetPassword(orgGuidId))
-        {
-            throw new NotFoundException();
-        }
-
-        var organizationUser = await _organizationUserRepository.GetByIdAsync(new Guid(id));
-        if (organizationUser == null || !organizationUser.UserId.HasValue)
+        var organizationUser = await _organizationUserRepository.GetByIdAsync(id);
+        if (organizationUser is not { UserId: not null })
```
eliykat: I think the double negative makes this more confusing.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
eliykat, line 1: Please cover the `Get` endpoint as well. It's a bit unusual today in that it's based off the orgUser's properties rather than the `orgId` in the route, but we can bring it into line with our usual practice. e.g.

```c#
[Authorize<ManageUsersRequirement>]
// etc
if (organizationUser.organizationId != orgId)
{
}
```

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerTests.cs
eliykat, line 1: Love the integration tests on particularly sensitive endpoints 👏 

💡 These tests only test custom users. It would also be good to add tests for other user roles (e.g. using `[InlineData]`).

eliykat (changes requested):
Great work, thanks for following up on this. Really nice to see these attributes start to be used.

src/Api/AdminConsole/Authorization/Requirements/ManageGroupsOrUsersRequirement.cs
jrmccannon: I think its worth the unit test around this. I'll add one.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
jrmccannon: Removed. Updated objective in description with link to client

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -124,10 +124,11 @@ public OrganizationUsersController(
...
     [HttpGet("{id}")]
+    [Authorize<ManageUsersRequirement>]
     public async Task<OrganizationUserDetailsResponseModel> Get(Guid id, bool includeGroups = false)
     {
         var (organizationUser, collections) = await _organizationUserRepository.GetDetailsByIdWithCollectionsAsync(id);
-        if (organizationUser == null || !await _currentContext.ManageUsers(organizationUser.OrganizationId))
+        if (organizationUser == null)
         {
             throw new NotFoundException();
         }
```
eliykat: You still need to make sure the orgUser belongs to the organization specified in the route, otherwise you could mix & match an organization ID that you own with an unrelated orgUser.

eliykat (changes requested):

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -125,10 +125,10 @@ public OrganizationUsersController(
 
     [HttpGet("{id}")]
     [Authorize<ManageUsersRequirement>]
-    public async Task<OrganizationUserDetailsResponseModel> Get(Guid id, bool includeGroups = false)
+    public async Task<OrganizationUserDetailsResponseModel> Get(Guid orgId, Guid id, bool includeGroups = false)
     {
         var (organizationUser, collections) = await _organizationUserRepository.GetDetailsByIdWithCollectionsAsync(id);
-        if (organizationUser == null)
+        if (organizationUser == null && organizationUser.OrganizationId != orgId)
```
eliykat: Typo - should be an OR.

```c#
        if (organizationUser == null || organizationUser.OrganizationId != orgId)
```

eliykat (approved):
Please ask QA how they'd like to handle testing. It can probably be handled by nightly automation but they should make that call.

PR #6022, opened 2025-07-03
[SM-1274] Adding Project Events

cd-bitwarden:
## 📔 Objective

Add project event logs 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


JimmyVo16 (dismissed):

src/Sql/dbo/Tables/Event.sql
```diff
@@ -14,11 +14,12 @@
     [IpAddress]              VARCHAR(50)      NULL,
     [Date]                   DATETIME2 (7)    NOT NULL,
     [ProviderId]             UNIQUEIDENTIFIER NULL,
-    [ProviderUserId]         UNIQUEIDENTIFIER NULL,
+[ProviderUserId]         UNIQUEIDENTIFIER NULL,
```
withinfocus: ⛏ Formatting issue.

src/Sql/dbo/Tables/Event.sql
```diff
@@ -14,11 +14,12 @@
...
     [ProviderOrganizationId] UNIQUEIDENTIFIER NULL,
     [SystemUser]             TINYINT          NULL,
     [DomainName]             VARCHAR(256)     NULL,
     [SecretId]               UNIQUEIDENTIFIER NULL,
+    [ProjectId]              UNIQUEIDENTIFIER NULL,
```
withinfocus: ❌ New columns must always be at the end.

withinfocus (changes requested):
We need a traditional SQL migration since Dapper-based repositories will use this table right? Table update, proc changes, etc.

src/Sql/dbo/Tables/Event.sql
rkac-bw: Also need migration script, in /util/Migrator/DbScripts/, to add column:

 ```
IF COL_LENGTH('[dbo].[Event]', 'ProjectId') IS NULL
  BEGIN
      ALTER TABLE [dbo].[Event]
      ADD [ProjectId] UNIQUEIDENTIFIER NULL
  END
  GO
```

Comment (rkac-bw) [reply src/Sql/dbo/Tables/Event.sql]: Need to update create_event.sql stored procedure, need too add projectid there and dapper calls:
  - src/Infrastructure.Dapper/Repositories/Repository.cs:55 (generic CreateAsync)
  - src/Infrastructure.Dapper/AdminConsole/Repositories/EventRepository.cs:95

  Additionally, the bulk insert path in EventRepository.BuildEventsTable (lines 168-245) is missing the
  ProjectId column mapping, which will cause issues for bulk event operations.
  
  Also
  MySQL Migration: No migration exists to add ProjectId column
  MySQL Database Snapshot: Missing ProjectId in the schema

File: src/Api/AdminConsole/Controllers/EventsController.cs [hunk 3]
```diff
@@ -104,6 +112,102 @@ public async Task<ListResponseModel<EventResponseModel>> GetOrganization(string
...
+            throw new NotFoundException();
+        }
+
+        var secret = await _secretRepository.GetByIdAsync(secretGuid);
+        var orgIdForVerification = secret?.OrganizationId ?? orgGuid;
+        var secretOrg = _currentContext.GetOrganization(orgIdForVerification);
+
+        if (secretOrg == null || !await _currentContext.AccessEventLogs(secretOrg.Id))
+        {
+            throw new NotFoundException();
```
withinfocus, line 134: ❓ Is this what SM does elsewhere? There can be a debate between 404 and 403.

src/Api/AdminConsole/Controllers/EventsController.cs
```diff
@@ -104,6 +112,102 @@ public async Task<ListResponseModel<EventResponseModel>> GetOrganization(string
...
+        }
+
+        bool canViewLogs = false;
+
+        if (secret == null)
```
withinfocus, line 139: ❓ Why would this happen? Is it for org-wide event queries?

src/Api/SecretsManager/Controllers/SecretsTrashController.cs
```diff
@@ -15,17 +19,23 @@ public class TrashController : Controller
...
+        IRestoreTrashCommand restoreTrashCommand,
+        IUserService userService,
+        IEventService eventService)
     {
         _currentContext = currentContext;
         _secretRepository = secretRepository;
         _emptyTrashCommand = emptyTrashCommand;
         _restoreTrashCommand = restoreTrashCommand;
+        _eventService = eventService;
+        _userService = userService;
```
withinfocus: ⛏️ You swapped the order of these two vs. the parameters higher up.

src/Core/AdminConsole/Services/Implementations/EventService.cs
```diff
@@ -464,6 +464,58 @@ public async Task LogServiceAccountSecretsEventAsync(Guid serviceAccountId, IEnu
...
     }
 
+    public async Task LogUserProjectsEventAsync(Guid userId, IEnumerable<Project> projects, EventType type, DateTime? date = null)
+    {
+        var orgAbilities = await _applicationCacheService.GetOrganizationAbilitiesAsync();
+        var eventMessages = new List<IEvent>();
+
+        foreach (var project in projects)
+        {
+            if (!CanUseEvents(orgAbilities, project.OrganizationId))
```
withinfocus, line 474: ❓ I don't know the history of usage with these -- why is this check made here vs. before the event is submitted?

src/Sql/dbo/SecretsManager/Stored Procedures/Event/Event_ReadPageByProjectId.sql
```diff
@@ -0,0 +1,23 @@
...
+    @BeforeDate DATETIME2(7),
+    @PageSize INT
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT
+        *
+    FROM
+        [dbo].[EventView]
```
withinfocus: 🎨 I always like to name what you're selecting from, in case someone absent-mindedly `JOIN`s later.

src/Sql/dbo/SecretsManager/Stored Procedures/Event/Event_ReadPageByProjectId.sql
```diff
@@ -0,0 +1,23 @@
...
+    WHERE
+        [Date] >= @StartDate
+        AND (@BeforeDate IS NOT NULL OR [Date] <= @EndDate)
```
withinfocus, line 38: ❓ Shouldn't this and the other query below use `>=` or am I misunderstanding the logic?

util/Migrator/DbScripts/2025-07-17_00_AddProjectEventLogsToEventNewColumn.sql
```diff
@@ -0,0 +1,16 @@
...
+IF OBJECT_ID('[dbo].[EventView]', 'V') IS NOT NULL
+BEGIN
+    DROP VIEW [dbo].[EventView];
+END
+GO
+
+CREATE VIEW [dbo].[EventView]
+AS
+SELECT * FROM [dbo].[Event];
+GO
```
withinfocus: ⛏️ Newline at the end.

src/Api/AdminConsole/Controllers/EventsController.cs
```diff
@@ -104,6 +112,102 @@ public async Task<ListResponseModel<EventResponseModel>> GetOrganization(string
...
+            canViewLogs = secretOrg.Type is Core.Enums.OrganizationUserType.Admin or Core.Enums.OrganizationUserType.Owner;
+        }
+        else
+        {
+            if (!_currentContext.AccessSecretsManager(secret.OrganizationId))
+            {
+                throw new NotFoundException();
+            }
+
+            var userId = _userService.GetProperUserId(User)!.Value;
```
JimmyVo16: Non-blocker: Could you extract lines 145–154 into a separate method to make the logic easier to understand?

src/Api/AdminConsole/Controllers/EventsController.cs
JimmyVo16: I believe we prefer 403 over 404 to prevent probing or resource enumeration.

cd-bitwarden: Yeah this is what we do often, especially inside of the EventsController API, so to keep it consistent with how other calls behave I kept it as NotFoundException

src/Api/AdminConsole/Controllers/EventsController.cs
cd-bitwarden: This happens when the secret has been permanently deleted, if it has been perm. deleted then the only person who can view the logs is an admin/owner. This logic is here because we decided that we still want the org admins/owners to be able to view the logs for this item even though we don't have all the details for the secret available.

src/Api/SecretsManager/Controllers/SecretsTrashController.cs
cd-bitwarden: Thanks, I fixed it :)

src/Core/AdminConsole/Services/Implementations/EventService.cs
cd-bitwarden: We don't log the event if the organization doesn't have the ability to view events, so that is why I use the orgId from the project and check if we are even going to log that event.

src/Api/AdminConsole/Controllers/EventsController.cs
```diff
@@ -104,6 +112,102 @@ public async Task<ListResponseModel<EventResponseModel>> GetOrganization(string
...
+        [FromQuery] DateTime? start = null,
+        [FromQuery] DateTime? end = null,
+        [FromQuery] string continuationToken = null)
+    {
+        if (!Guid.TryParse(id, out var projectGuid) || !Guid.TryParse(orgId, out var orgGuid))
+        {
+            throw new NotFoundException();
+        }
+
+        // Try to get project
```
JimmyVo16: Can you extract these into separate methods to make the comments unnecessary?

Here is an example, but it doesn't have to be like this. 

Calling code
```
       var project = await GetProject(projectGuid, orgGuid);

        await ValiddateOrganization(project);
```

Extracted methods
```

    private async Task ValiddateOrganization(Project project)
    {
        var org = _currentContext.GetOrganization(project.OrganizationId);

        if (org == null || !await _currentContext.AccessEventLogs(org.Id))
        {
        }
    }

    private async Task<Project> GetProject(Guid projectGuid, Guid orgGuid)
    {
        var project = await _projectRepository.GetByIdAsync(projectGuid);
        if (project != null)
        {
            return project;
        }

        var fallbackProject =  new Project
        {
            Id = projectGuid,
            OrganizationId = orgGuid
        };

        return fallbackProject;
    }

```

JimmyVo16: The outcome of this [discussion](https://github.com/bitwarden/server/pull/6022/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.sql&owned-by%5B%5D=JimmyVo16&show-viewed-files=true#r2216814462) should apply here as well.

src/Sql/dbo/SecretsManager/Stored Procedures/Event/Event_ReadPageByProjectId.sql
cd-bitwarden: This is the way the similar queries have been written to handle the pagination, it looks wonky because we are paginating backward in time.

JimmyVo16 (changes requested):
Just some minor code clean up from the AC team's perspective.

withinfocus (approved):
Leaving this to AC to finish.

src/Api/AdminConsole/Controllers/EventsController.cs
cd-bitwarden: @withinfocus @JimmyVo16 my opinion in this would be to make a new ticket to address this in all areas, because only doing it for my new code would be inconsistent.  Are you two okay with that approach?

withinfocus: I agree -- my goal is consistency across the platform.

JimmyVo16 (dismissed):

withinfocus (dismissed):

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
```diff
@@ -26,7 +26,7 @@ public SecretRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper
             var dbContext = GetDatabaseContext(scope);
             var secret = await dbContext.Secret
                                     .Include("Projects")
-                                    .Where(c => c.Id == id && c.DeletedDate == null)
+                                    .Where(c => c.Id == id)
```
justindbaur: Could there be any uses of `GetByIdAsync` that is depending on it NOT returning deleted items?

src/Api/AdminConsole/Controllers/EventsController.cs
```diff
@@ -104,6 +113,77 @@ public async Task<ListResponseModel<EventResponseModel>> GetOrganization(string
         return new ListResponseModel<EventResponseModel>(responses, result.ContinuationToken);
     }
 
+    [HttpGet("~/organization/{orgId}/secrets/{id}/events")]
+    public async Task<ListResponseModel<EventResponseModel>> GetSecrets(
+        string id, string orgId,
```
justindbaur: You can just accept these as Guids and not have to do the parsing. It does technically change the behavior of returning a 400 if they give a non-valid guid instead of a 404 though.

```suggestion
        Guid id, Guid orgId,
```

src/Api/AdminConsole/Controllers/EventsController.cs
```diff
@@ -104,6 +113,77 @@ public async Task<ListResponseModel<EventResponseModel>> GetOrganization(string
...
+        var (fromDate, toDate) = ApiHelpers.GetDateRange(start, end);
+        var result = await _eventRepository.GetManyBySecretAsync(secret, fromDate, toDate, new PageOptions { ContinuationToken = continuationToken });
+        var responses = result.Data.Select(e => new EventResponseModel(e));
+        return new ListResponseModel<EventResponseModel>(responses, result.ContinuationToken);
+    }
+
+    [HttpGet("~/organization/{orgId}/projects/{id}/events")]
+    public async Task<ListResponseModel<EventResponseModel>> GetProjects(
+        string id,
+        string orgId,
```
justindbaur: Same here

src/Core/SecretsManager/Repositories/Noop/NoopSecretRepository.cs
```diff
@@ -105,4 +105,6 @@ public Task<int> GetSecretsCountByOrganizationIdAsync(Guid organizationId, Guid
     {
         return Task.FromResult(0);
     }
+
+    public Task<IEnumerable<Secret>> GetManyTrashedSecretsByIds(IEnumerable<Guid> ids) => throw new NotImplementedException();
```
justindbaur: Is it intentionally that this one is throwing when the rest don't?

```suggestion
    public Task<IEnumerable<Secret>> GetManyTrashedSecretsByIds(IEnumerable<Guid> ids) => Task.FromResult<IEnumerable<Secret>>([]);
```

util/MySqlMigrations/Migrations/20250717164642_20250717_AddingProjectIdToEvent.cs
justindbaur, line 1: Doesn't need to be fixed here but just for the future, we don't need to add the date at the beginning of EF migrations, they are already added to the front for ordering.

justindbaur (commented):
The `string` -> `Guid` can be optional I see that it appears to be the pattern in that controller, I just don't like the pattern 😄. 

The other one is just something to think about, if you've already done that and don't see a problem then it's fine with me.

bitwarden_license/src/Commercial.Infrastructure.EntityFramework/SecretsManager/Repositories/SecretRepository.cs
cd-bitwarden: EXCELLENT POINT!

withinfocus (approved):

justindbaur (approved):

util/Migrator/DbScripts/2025-07-17_00_AddProjectEventLogsToEventNewColumn.sql
rkac-bw, line 16: EXEC sp_refreshview N'[dbo].[EventView]'; should be used here after using CREATE or ALTER VIEW

withinfocus: To note, this is an optimization and isn't as destructive as the explicit drop before it.

I am more of a `CREATE OR ALTER` person vs. using the `EXEC` but that's me.

PR #6019, opened 2025-07-01
[PM-22108] Add PolicyDetails_ReadByOrganizationId proc

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22108

## 📔 Objective

We want to create this sproc to return the same data as PolicyDetails_ReadByUserId, but for all users in the organization.

### stored proc behavior 

1. Start with the OrganizationId and policyType.
2. Find all users associated with that OrganizationId by UserId or Email on the OrganizationUser table.
3. Find all organizations that those users belong to, including the starting organization.
4. Get policies by policyType for orgs and users. 

### Use cases for getting the OrgUsers
Once we find all the OrgUsers for the desired organization, we then need to find all the OrgUsers that are associated with those individual users.

1. Joining OrgUsers with UserId to other orgs by UserID
2. Joining OrgUsers with UserId to other orgs by Email
3. Joining OrgUsers with Email to other orgs by UserID
4. Joining OrgUsers with Email to other orgs by Email

#### Simplified Data  

OrganizationUser table

| **OrganizationUserID** | **UserId** | **Email**         | **OrgId** |
| ---------------------- | ---------- | ----------------- | --------- |
| OrgUser1               | UserA      | NULL              | Org1      |
| OrgUser2               | UserB      | NULL              | Org1      |
| OrgUser3               | NULL       | UserC@example.com | Org1      |
| OrgUser4               | NULL       | UserD@example.com | Org1      |
| OrgUser5               | UserF      | NULL              | Org10     |
| OrgUser6               | UserA      | NULL              | Org2      |
| OrgUser7               | NULL       | UserB@example.com | Org3      |
| OrgUser8               | UserC      | NULL              | Org4      |
| OrgUser9               | NULL       | UserD@example.com | Org5      |
| OrgUser10              | UserA      | UserA@example.com | Org6      |

User Table

|            |                   |     |
| ---------- | ----------------- | --- |
| **UserId** | **Email**         |     |
| UserA      | UserA@example.com |     |
| UserB      | UserB@example.com |     |
| UserC      | UserC@example.com |     |
| UserD      | UserD@example.com |     |
| UserF      | UserF@example.com |     |

Policy Table

  

|     |                |           |
| --- | -------------- | --------- |
| Id  | **PolicyType** | **OrgId** |
| 1   | PolicyA        | Org1      |
| 2   | PolicyB        | Org1      |
| 3   | PolicyA        | Org2      |
| 4   | PolicyA        | Org3      |
[...]

PR #6036, opened 2025-07-03
[PM-22108] Add OrganizationPolicyValidator

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22108

## 📔 Objective

**Business Objectives**: 
We want the ability to query all users in an organization and build policy requirements for them. This is similar to the existing `PolicyRequirementQuery`, except it will operate at the organization level instead of at the user level.

This method will only be used when enabling or disabling a policy.

**Solution**:
1. Create an abstract class, `OrganizationPolicyValidator`, that will contain the new org-level query method, `GetUserPolicyRequirementsByOrganizationIdAsync`, so it will only be used within `IPolicyValidator`. This will help prevent developers from confusing the new org-level method with the existing user-level, `GetAsync`.
2. The implementations of `IPolicyRequirement` still contain the logic for policy enforcement, so any changes there will affect both user-level and org-level PolicyRequirement building.
3. Child validators of `OrganizationPolicyValidator` should work seamlessly with the existing `IPolicyValidator` code.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Infrastructure.Dapper/AdminConsole/Repositories/PolicyRepository.cs
```diff
@@ -61,6 +61,14 @@ public async Task<ICollection<Policy>> GetManyByUserIdAsync(Guid userId)
         }
     }
 
+    public Task<IEnumerable<UserPolicyDetails>> GetOrganizationPolicyDetailsByOrgId(Guid orgId, PolicyType policyType)
```
JimmyVo16: The logic for this will be fleshed out in [this PR](https://github.com/bitwarden/server/pull/6019).

src/Infrastructure.EntityFramework/AdminConsole/Repositories/PolicyRepository.cs
```diff
@@ -56,6 +56,8 @@ public PolicyRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper
         }
     }
 
+    public Task<OrganizationPolicyDetails> GetOrganizationPolicyDetailsByOrgId(Guid orgId, PolicyType policyType) => throw new NotImplementedException();
```
JimmyVo16: The logic for this will be fleshed out in https://github.com/bitwarden/server/pull/6019.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
```diff
@@ -27,6 +27,8 @@ private static void AddPolicyValidators(this IServiceCollection services)
         services.AddScoped<IPolicyValidator, ResetPasswordPolicyValidator>();
         services.AddScoped<IPolicyValidator, MaximumVaultTimeoutPolicyValidator>();
         services.AddScoped<IPolicyValidator, FreeFamiliesForEnterprisePolicyValidator>();
+        services.AddScoped<IPolicyValidator, ProtoTypeOrganizationDataOwnershipPolicyValidator>();
```
JimmyVo16, line 30: @eliykat The prototype classes are just examples of how we can use this pattern in the upcoming PR. I'll remove them before marking this as `ready for review`.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementBuilder.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Bit.Core.AdminConsole.Models.Data.Organizations.Policies;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.Implementations;
+
+public class PolicyRequirementBuilder
+{
+    public static T Build<T>(IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories, IEnumerable<PolicyDetails> policyDetails) where T : IPolicyRequirement
```
JimmyVo16: @eliykat We can turn this into a DI service and add it to both `OrganizationPolicyValidator` and `PolicyRequirementQuery`, if you'd like. Both solutions have similar trade-offs.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public abstract PolicyType Type { get; }
+
+    public abstract IEnumerable<PolicyType> RequiredPolicies { get; }
+
+
+    protected async Task<IEnumerable<T>> GetUserPolicyRequirementsByOrganizationIdAsync<T>(Guid organizationId, PolicyType policyType) where T : IPolicyRequirement
+    {
+        var policyDetails = await policyRepository.GetPolicyDetailsByOrganizationIdAsync(organizationId, policyType);
+        var policyDetailGroups = policyDetails.GroupBy(policyDetail => policyDetail.UserId);
```
JimmyVo16, line 24: @eliykat 

I really like this logic. From the consumer's perspective, they're just getting a list of IPolicyRequirement. The IPolicyRequirement objects will be the same regardless of whether they came from here or `PolicyRequirementQuery`.

It also allows the `IPolicyRequirement` implementations to contain the logic for determining which users the policy should apply to.

Finally, we only need the `UserId` here and not elsewhere, so the `OrganizationPolicyDetails` object can handle that without requiring changes to existing code.

Please review and let me know what you think of this pattern. I'll add tests and clean up some comments once we're aligned on the big picture.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        var requirements = new List<T>();
+
+        foreach (var policyDetailGroup in policyDetailGroups)
+        {
+            requirements.Add(PolicyRequirementBuilder.Build<T>(_factories, policyDetailGroup.ToList()));
+
+        }
```
eliykat: This is filtering the factories and policies (inside `Build`) and creating a new list of the `policyDetailGroup` each iteration through the foreach loop.

It is a bit of code duplication, but I suggest it's better to drop the `Builder` abstraction so you can handle it in a slightly more bespoke manner here:
1. get the factory upfront (out of the loop)
2. no need to filter for policyType, because your repository query is already scoped to the policyType you need
3. the loop only needs to call Enforce and Create.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
eliykat: This all looks good to me, you've done a good job making it conform to the existing patterns.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementBuilder.cs
eliykat: I don't think DI is needed. Per my comment below I'm not sure we need this at all, but if you do keep it then a static class is fine.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
eliykat: I appreciate the examples!

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
JimmyVo16: Another benefit of combining the logic is that it signals to future devs that the pieces are connected, but that's pretty trivial, since they can already tell by the requirement type.

JimmyVo16:
I'm closing this PR because we've restructured some tickets for better visibility and tracking.

PR #6065, opened 2025-07-08
PM-21685 fixing flaky test

prograhamming:
## 🎟️ Tracking

[PM-21685](https://bitwarden.atlassian.net/browse/PM-21685?linkSource=email)

## 📔 Objective

This PR is to fix a flaky test under the Import Ciphers codebase.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/test/Api.Test/Tools/Controllers/ImportCiphersControllerTests.cs
+++ b/test/Api.Test/Tools/Controllers/ImportCiphersControllerTests.cs
@@ -20,6 +20,7 @@ using NSubstitute;
 using NSubstitute.ClearExtensions;
 using Xunit;
 using GlobalSettings = Bit.Core.Settings.GlobalSettings;
+using ImportCiphersLimitationSettings = Bit.Core.Settings.GlobalSettings.ImportCiphersLimitationSettings;
 
 namespace Bit.Api.Test.Tools.Controllers;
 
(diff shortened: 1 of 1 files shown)


prograhamming:
Whoops! I added to the user story card in Jira and  should of added this is in the PR. 

I believe the cause of the flakiness was the `GlobalSettings.ImportCiphersLimitation`([Model is here](https://github.com/bitwarden/server/blob/main/src/Core/Settings/GlobalSettings.cs#L604)) values being assigned with random values from the `BitAutoData` attribute. The error message with the `BadRequestException` received was `"You cannot import this much data at once."` which is throw in the if statement conditional under the `PostImport` - `import-organization` endpoint [See line 69 in the file](https://github.com/bitwarden/server/blob/main/src/Api/Tools/Controllers/ImportCiphersController.cs#L69).

For organizations we store our import cipher limits in `appsettings.json`([See line 59 here](https://github.com/bitwarden/server/blob/main/src/Api/appsettings.json#L59)), so the SRE team can increase the limit quickly if they are onboarding a huge organization that goes beyond those limitations.  Also, a side note we don't do this for the personal import ciphers endpoints those are hard coded within the code ([See here](https://github.com/bitwarden/server/blob/main/src/Api/Tools/Controllers/ImportCiphersController.cs#L52)) which is why the personal import cipher tests never had this problem as well.

`_globalSettings.ImportCiphersLimitation` fields and `ImportOrganizationCiphersRequestModel`([Model is here](https://github.com/bitwarden/server/blob/main/src/Api/Tools/Models/Request/Organizations/ImportOrganizationCiphersRequestModel.cs#L9)) fields were getting assigned random values and eventually one of the **array fields** in `ImportOrganizationCiphersRequestModel` would be more than the number value in `_globalSettings.ImportCiphersLimitation`.  By setting the `_globalSettings.ImportCiphersLimitation` fields the same as they would be in `appsettings.json` ensures the randomness situation doesn't happen for import organization cipher tests and the tests use the values that would be used in PROD.

audreyality:
Watch me flap the needs-qa library because I forgot this was a fix to a unit test in between reading the explanation and accepting the PR. 🫠

PR #6066, opened 2025-07-08
Improve generated OpenAPI files

dani-garcia:
## 📔 Objective

Improve the OpenAPI generation experience in a couple of ways:
- Added a new script in the `dev` folder to generate the OpenAPI files in the same way that CI does and the way that the SDK requires. Previously this was documented on [the SDK readme](https://github.com/bitwarden/sdk-internal#api-bindings), but I had to go and look up the commands every single time. This also simplifies the CI workflow.
- Updated swashbuckle to the latest version, technically not needed, but might as well do it while i was playing with the other things. I saw no changes in the generated schemas regardless of the version anyway.
- Added a couple of filters to modify how the OpenAPI file gets generated:
    - `EncryptedStringSchemaFilter`: This will tag every field in a class that has the `#[EncryptedString]` annotation with the `x-enc-string` format type. This will allow us to use `EncString`  in the SDK instead of plain strings.
    - `SourceFileLineOperationFilter`: This will tag every operation with the source file and line, which will allow other tools to retrieve them and take you to the originating source of the operation. It also updates the description to include a permalink to the file in github, which will show in the SDK docs:
    ```rust
    /// POST /accounts/api-key
    ///
    ///  This operation is defined on file: [`https://github.com/bitwarden/server/blob/1e17dec650760da9cddf9813934d898207aabe05/src/Api/Auth/Controllers/AccountsController.cs:L522`]
    async fn accounts_api_key_post<'a>(
        &self,
        secret_verification_request_model: Option<models::SecretVerificationRequestModel>,
    ) -> Result<models::ApiKeyResponseModel, Error<AccountsApiKeyPostError>>;
    ```
  - `GitCommitDocumentFilter`: This just adds a `x-git-commit` property to the generated JSON file, useful if you're like me and forget which version of the server was used to generate the JSON file.

Note that these last two filters use `.pdb` Debug files and Git commands, so they are only enabled in development.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/SharedWeb/Swagger/EncryptedStringSchemaFilter.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            // Only apply to string properties
+            if (prop.PropertyType != typeof(string))
+                continue;
+
+            // Check if the property has the EncryptedString attribute
+            if (prop.GetCustomAttributes(typeof(EncryptedStringAttribute), true).FirstOrDefault() != null)
+            {
+                // Convert prop.Name to camelCase for JSON schema property lookup
+                var jsonPropName = char.ToLowerInvariant(prop.Name[0]) + prop.Name[1..];
```
justindbaur: Maybe this instead? Their algorithm is a _little_ different.
```suggestion
                var jsonPropName = JsonNamingPolicy.CamelCase.ConvertName(prop.Name);
```

https://github.com/dotnet/runtime/blob/17365a6255c1ffb88123ce77144e8fb3b5e54bbd/src/libraries/System.Text.Json/Common/JsonCamelCaseNamingPolicy.cs#L28-L52

src/SharedWeb/Swagger/SourceFileLineOperationFilter.cs
justindbaur, line 1: Just incredible.

justindbaur (commented):
One suggestion, otherwise fantastic!

src/SharedWeb/Swagger/EncryptedStringSchemaFilter.cs
dani-garcia: Oh nice! I've tested it and it's actually better than what I'm doing, as it deals with things like `SSN` the same way that the swagger json casing does, thanks!

ike-kottlowski (approved):
Thanks boss.

PR #6075, opened 2025-07-09
Add bulk default collection creation method

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22102

## 📔 Objective

**Intent Behavior**: We want a method to create one default collection per user, but only if they do not already have one. We also want the entire operation to run within a single database transaction to prevent time-of-check/time-of-use bugs.

**The how:**

1. Received a list of OrgUserIds, which indicates the users we want to create default collections for, along with some additional data.
2. Retrieve all users with a default collection for the organization.
3. Filter out the input OrgUserIds that already have default collections.
4. Use a bulk call to create default collections for users who need them.

## 📸 Screenshots

This method is not being used yet. Other PRs will use this method. The database integration tests should be sufficient. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
```diff
@@ -309,6 +312,212 @@ public async Task<ICollection<CollectionAccessSelection>> GetManyUsersByIdAsync(
         }
     }
 
+    public async Task CreateDefaultCollectionsAsync(Guid organizationId, IEnumerable<Guid> orgUserIds, string defaultCollectionName)
```
JimmyVo16: @eliykat for the list of `orgUserIds` should we pass in the requirement objects with the OrgUserId, just the `orgUserIds` like it is here

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
```diff
@@ -309,6 +312,212 @@ public async Task<ICollection<CollectionAccessSelection>> GetManyUsersByIdAsync(
...
+                    throw;
+                }
+            }
+        }
+    }
+
+    private async Task<HashSet<Guid>> GetOrgUserIdsWithDefaultCollectionAsync(SqlConnection connection, SqlTransaction transaction, Guid organizationId)
+    {
+        const string sql = @"
+                    SELECT
```
JimmyVo16, line 354: Here is the query plan for this. I don't see any issues. 

```
| StmtText |
| :--- |
|   |--Nested Loops(Inner Join, OUTER REFERENCES:([cu].[CollectionId])) |
|        |--Nested Loops(Inner Join, OUTER REFERENCES:([ou].[Id])) |
|        |    |--Index Seek(OBJECT:([vault_dev].[dbo].[OrganizationUser].[IX_OrganizationUser_OrganizationId] AS [ou]), SEEK:([ou].[OrganizationId]=[@OrganizationId]) ORDERED FORWARD) |
|        |    |--Index Seek(OBJECT:([vault_dev].[dbo].[CollectionUser].[IX_CollectionUser_OrganizationUserId] AS [cu]), SEEK:([cu].[OrganizationUserId]=[vault_dev].[dbo].[OrganizationUser].[Id] as [ou].[Id]) ORDERED FORWARD) |
|        |--Clustered Index Seek(OBJECT:([vault_dev].[dbo].[Collection].[PK_Collection] AS [c]), SEEK:([c].[Id]=[vault_dev].[dbo].[CollectionUser].[CollectionId] as [cu].[CollectionId]),  WHERE:([vault_dev].[dbo].[Collection].[Type] as [c].[Type]=[@CollectionType]) ORDERED FORWARD) |

```

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using Xunit;
+
+
+namespace Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories.CollectionRepository;
+
+public class CreateDefaultCollectionsTests
+{
+
+    [DatabaseTheory, DatabaseData]
+    public async Task CreateDefaultCollectionsAsync_ShouldCreateDefaultCollection_WhenUsersDoNotHaveDefaultCollection(
```
JimmyVo16, line 12: @jrmccannon can you let me know if we have any existing methods to clean up the data? I didn't see any teardown or cleanup methods for these integration tests.

jrmccannon: We do not. There is currently nothing in the testing tools that will back the transactions done during that execution context. I've just been calling the delete methods manually.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
```diff
@@ -309,6 +312,217 @@ public async Task<ICollection<CollectionAccessSelection>> GetManyUsersByIdAsync(
...
+        using (var connection = new SqlConnection(ConnectionString))
+        {
+            connection.Open();
+            using (var transaction = connection.BeginTransaction())
+            {
+                try
+                {
+                    var orgUserIdWithDefaultCollection = await GetOrgUserIdsWithDefaultCollectionAsync(connection, transaction, organizationId);
+
+                    var missingDefaultCollectionUserIds = affectedOrgUserIds.Where(orgUserId => !orgUserIdWithDefaultCollection.Contains(orgUserId)).ToList();
```
eliykat: ⛏️  I think using `Except` would be clearer here.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
eliykat, line 1: There's a lot of duplicated SqlBulkCopy logic from `CipherRepository` here. Can it be moved to this repository (as the actual repository for the domain) and then called by `CipherRepository.CreateAsync`? (Assuming it's more or less a cut & paste - I haven't compared closely.)

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
eliykat: I think nested loops are a problem (as they are in other contexts) but it's probably doing that because your local db is so small. That said, this is a very simple query so I don't expect anything too crazy in prod.

eliykat (changes requested):
Initial review just to get some feedback to you - overall it looks good.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
JimmyVo16: Yeah, it looks like we have indexes on all the join columns, so a merge join should be possible in prod.

JimmyVo16: Question: Since I don't have visibility into prod, I don't know how bad a nested join will be, because in some contexts it's fine. So moving forward, should I lean on the side of caution when I see a nested join and try to convert it to a merge join?

eliykat: This is an extremely simple SQL query - some inner joins and a WHERE clause - so just on that basis I would be very surprised if it caused any issues in prod. It's also using indexes at each step, which is good. And the final query plan is determined by the actual data in our prod db. So in this case I am not concerned about it, although feel free to discuss with Robert if you like.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
JimmyVo16: Yeah, it looks similar, but it’s not really shareable since the bulk of the work is in the build methods, like `BuildCiphersTable`. The logic here essentially involves creating variables for each table column and inserting them as a row, which is very specific to each individual table.

We could potentially use reflection to iterate through the table and get its columns, but I don’t think it’s worth tackling that here.

eliykat: The common logic is that both have to build the in-code representation of the table and copy source data into it - `BuildCollectionsTable` and `BuildCollectionUsersTable` in both repositories, which as far as I can tell are identical. Note that these relate to the same tables - CiphersRepository also inserts data into the Collection table during cipher import.

At its simplest, we could just make those a public methods on `CollectionRepository`, so that if another repository needs to bulk copy data into it, they don't need to also maintain their own logic for our domain.

However it could be slightly better encapsulated so that it handles the bulk insert for you, e.g.

```c#
    public Action<Task> BulkCopyCollections(SqlConnection connection, SqlBulkCopyOptions options,
        SqlTransaction transaction, IEnumerable<Collection> collections)
        => () =>
        {
            using var bulkCopy = new SqlBulkCopy(connection, SqlBulkCopyOptions.KeepIdentity, transaction);
            bulkCopy.DestinationTableName = "[dbo].[Collections]";
            var dataTable = BuildCollectionsTable(bulkCopy, collections);
            bulkCopy.WriteToServer(dataTable);
        };

```

This could be used by CollectionRepository to bulk insert default collections, and CipherRepository to bulk insert collections when importing.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
```diff
@@ -309,6 +312,219 @@ public async Task<ICollection<CollectionAccessSelection>> GetManyUsersByIdAsync(
...
+    {
+        if (!affectedOrgUserIds.Any())
+        {
+            return;
+        }
+
+        using (var connection = new SqlConnection(ConnectionString))
+        {
+            connection.Open();
+            using (var transaction = connection.BeginTransaction())
```
eliykat: Rider informs me this should be `await using`, because it's an `IAsyncDisposable` interface.

⛏️ I also recommend `await using var` without brackets or indentation, just to avoid excessive indentation.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
eliykat: Just saw this - the list of OrgUserIds is fine 👍

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -769,4 +775,90 @@ private static async Task ReplaceCollectionUsersAsync(DatabaseContext dbContext,
...
+            return;
+        }
+
+        using (var scope = ServiceScopeFactory.CreateScope())
+        {
+            var dbContext = GetDatabaseContext(scope);
+
+            var orgUserIdWithDefaultCollection = await GetOrgUserIdsWithDefaultCollectionAsync(dbContext, organizationId);
+
+            var missingDefaultCollectionUserIds = affectedOrgUserIds.Where(orgUserId => !orgUserIdWithDefaultCollection.Contains(orgUserId)).ToList();
```
eliykat: Can use `Except` here as well to be consistent.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -769,4 +775,90 @@ private static async Task ReplaceCollectionUsersAsync(DatabaseContext dbContext,
...
+                     dbContext.Collections,
+                     temp => temp.cu.CollectionId,
+                     c => c.Id,
+                     (temp, c) => new { temp.ou, Collection = c }
+                 )
+                 .Where(x => x.Collection.Type == CollectionType.DefaultUserCollection)
+                 .Select(x => x.ou.Id)
+                 .ToListAsync();
+
+        return results.ToHashSet();
```
eliykat, line 826: `ToListAsync` isn't required if you're going to convert it into a HashSet anyway.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        var organization = await organizationRepository.CreateAsync(new Organization
+        {
+            Id = organizationId,
+            Name = "Test name",
+            PlanType = PlanType.EnterpriseAnnually,
+            Plan = "Test",
+            BillingEmail = "billing@example.com"
+        });
+
+        var resulOrganizationUsers = await Task.WhenAll(
```
eliykat: typo in `resul`? (also in other tests)

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            ApiKey = "TEST",
+            SecurityStamp = "stamp",
+        });
+
+        var orgUser = await organizationUserRepository.CreateAsync(new OrganizationUser
+        {
+            OrganizationId = organization.Id,
+            UserId = user.Id,
+            Status = OrganizationUserStatusType.Confirmed,
+        });
```
eliykat: Check out the extension methods in [OrganizationTestHelpers](https://github.com/bitwarden/server/blob/main/test/Infrastructure.IntegrationTest/AdminConsole/OrganizationTestHelpers.cs), they may help here.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        var organizationId = Guid.NewGuid();
+
+        var organization = await organizationRepository.CreateAsync(new Organization
+        {
+            Id = organizationId,
+            Name = "Test name",
+            PlanType = PlanType.EnterpriseAnnually,
+            Plan = "Test",
+            BillingEmail = "billing@example.com"
+        });
```
eliykat: `OrganizationTestHelpers` can also help here.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        foreach (var userId in userIds)
+        {
+            if (userId == null)
+            {
+                continue;
+            }
+
+            await userRepository.DeleteAsync(new User() { Id = userId.Value });
+        }
```
eliykat: There is a `DeleteManyAsync` method.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
eliykat, line 1: Suggestion: a mixed test where 1 user already has a default collections, and 1 does not.

eliykat (changes requested):

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
JimmyVo16: Would it make sense to add a linting rule for this? It could help prevent small inconsistencies like this and reduce some mental overhead.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
JimmyVo16: Unfortunately, there isn’t an async version of `ToHashSet`, so I just went with `ToListAsync` followed by ToHashSet. That way, the database call remains non-blocking, and we still get O(1) lookups later when filtering.

A `ToDictionaryAsync` could potentially work here, but `ToHashSet` feels more semantically appropriate since a dictionary is key-value pairs, and we’re only dealing with `OrgUserId`.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
JimmyVo16: That's strange, spell check should've caught that.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
JimmyVo16: > BuildCollectionsTable and BuildCollectionUsersTable in both repositories, which as far as I can tell are identical.

Oh yeah, they are the same, so the collections table should be too.

Your proposal will work, but should we be concerned about the circular dependency? I don't think it's an issue now, but it could be in the future.

JimmyVo16: It looks like other teams already have helper methods. I think we can put it there.

Side note: it would be nice to have something that can be shared between Dapper and EF projects, but I guess the use case is very small, so the lemon is not worth the squeeze right now.

JimmyVo16: I abstracted the methods into `BulkResourceCreationService`. We can add more to it and use it in `CipherRepository` in another ticket. This will reduce scope creep and testing effort.

I was also considering the name `BulkResourceCreator`. Either way, having the word create in the name helps reinforce the single responsibility of the class.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
eliykat: We don't have any linting set up for `server` other than `dotnet format` which is pretty minimal. Bishop has wanted to add [MegaLinter](https://megalinter.io/latest/) for a while but I think there was some concern around the "kitchen sink" nature of that, so it hasn't progressed. I do wish we had a solution here though; feel free to follow it up in `#architecture` if you want.

In the meantime I find Rider's intellisense suggestions to be pretty good, but I acknowledge it's easier to miss than a linting job.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
eliykat: Good point - there is a `ToHashSetAsync` but not until EF 9 and we're still on EF 8.x. I think this is a good solution in the meantime!

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
eliykat: It's not a circular dependency, but I get what you mean - should a repository depend on another? You're probably right that it's best to avoid. The example I am following here is how [RotateUserAccountKeysCommand](https://github.com/bitwarden/server/blob/7e80e01747617f3d88168570f1ee50a90ea31772/src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs#L100-L101) builds up a set of repository actions and then executes them. That avoids the issue because the dependency is only ever from the command to the repository.

That said, we're not really in a position to implement that exact pattern (it would require factoring of Tools' import code), so I think this is a good first step.

Can you please make a follow-up ticket to change `CipherRepository` over to using this static class, to make sure we close this loop?

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        using var bulkCopy = new SqlBulkCopy(connection, SqlBulkCopyOptions.KeepIdentity, transaction);
+        bulkCopy.DestinationTableName = "[dbo].[CollectionUser]";
+        var dataTable = BuildCollectionsUsersTable(bulkCopy, collectionUsers, errorMessage);
+        await bulkCopy.WriteToServerAsync(dataTable);
+    }
+
+    private static DataTable BuildCollectionsUsersTable(SqlBulkCopy bulkCopy, IEnumerable<CollectionUser> collectionUsers, string errorMessage)
+    {
+        var collectionUser = collectionUsers.First();
```
eliykat: I suggest `FirstOrDefault` so that an empty sequence can be handled by your null check below, rather than having `First` throw an error.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -0,0 +1,129 @@
...
+        table.Columns.Add(orgUserIdColumn);
+
+        var readOnlyColumn = new DataColumn(nameof(collectionUser.ReadOnly), typeof(bool));
+        table.Columns.Add(readOnlyColumn);
+
+        var hidePasswordsColumn = new DataColumn(nameof(collectionUser.HidePasswords), typeof(bool));
+        table.Columns.Add(hidePasswordsColumn);
+
+        var manageColumn = new DataColumn(nameof(collectionUser.Manage), typeof(bool));
+        table.Columns.Add(manageColumn);
```
eliykat: I suggest using `GetType()` like the CreateCollectionsAsync method does, so it automatically tracks any future change to the types used by the entity object.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        foreach (DataColumn col in table.Columns)
+        {
+            bulkCopy.ColumnMappings.Add(col.ColumnName, col.ColumnName);
+        }
+
```
eliykat, line 44: Do we need the primary key definition as in the current CipherRepository implementation? https://github.com/bitwarden/server/blob/7e80e01747617f3d88168570f1ee50a90ea31772/src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs#L888-L892

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        using var bulkCopy = new SqlBulkCopy(connection, SqlBulkCopyOptions.KeepIdentity, transaction);
+        bulkCopy.DestinationTableName = "[dbo].[Collection]";
+        var dataTable = BuildCollectionsTable(bulkCopy, collections, errorMessage);
+        await bulkCopy.WriteToServerAsync(dataTable);
+    }
+
+    private static DataTable BuildCollectionsTable(SqlBulkCopy bulkCopy, IEnumerable<Collection> collections, string errorMessage)
+    {
+        var collection = collections.First();
```
eliykat: Same comment re: `FirstOrDefault`.

eliykat (changes requested):
A few minor points, overall I really like this, good job.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
JimmyVo16: I think this is great, but for nullable types, it’s not so straightforward. I think for those, we can stick with `typeof` unless you have a simpler solution.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
JimmyVo16: Good call. I missed this

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
JimmyVo16: [Ticket](https://bitwarden.atlassian.net/browse/PM-24233)

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
eliykat: I agree, it doesn't work as nicely for nullable. That's fine.

eliykat (commented):
LGTM but as discussed @rkac-bw should take a look as it's still db changes even though it doesn't touch a .sql file. That said, he can correct me if he doesn't need to see it.

PR #6081, opened 2025-07-12
[PM-19055] Add OTP Token Provider that is not dependent on the User entity

ike-kottlowski:
## 🎟️ Tracking

[PM-19055](https://bitwarden.atlassian.net/browse/PM-19055)

## 📔 Objective

to support email OTP protected Sends we needed a way to allow OTPs to be created without having an active `ClaimsPrincipal` (user). To do this we created a class that handles generation and management of OTPs in the same way the `EmailTokenProvider` does today.

The primary difference being that this new class can be injected into the `EmailTokenProvider` and used to accomplish the same goals. This will reduce the need to support two different OTP providers, one for email mfa and one for Send authorization.

The email token provider is untouched in this PR but there is tech debt to modify the `EmailTokenProvider` to inject this new class and use its workflows, which were lifted from the `EmailTokenProvider` to begin with.

Claude Sonnet 4 was utilized when writing this feature.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-19055]: https://bitwarden.atlassian.net/browse/PM-19055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
+++ b/src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
@@ -1,6 +1,6 @@
 # OtpTokenProvider
 
-The `OtpTokenProvider` is a token provider service for generating and validating Time-Based one-time passwords (TOTP). It provides a secure way to create temporary tokens for various authentication and verification scenarios. The provider can be configured to generate tokens specific to your use case. Currently they are hard-coded in the DI pipeline.
+The `OtpTokenProvider` is a token provider service for generating and validating Time-Based one-time passwords (TOTP). It provides a secure way to create temporary tokens for various authentication and verification scenarios. The provider can be configured to generate tokens specific to your use case by using the options pattern in the DI pipeline.
 
 ## Overview
 
@@ -68,11 +68,11 @@ If you need to modify the default options you can do so by creating an extension
 
 #### OtpTokenProviderOptions
 
-``` csharp
+```csharp
 public class DefaultOtpTokenProviderOptions
 { ... }
 
-public class UserEmailOtpTokenOptions : DefaultOtpTokenProviderOptions {}
+public class UserEmailOtpTokenOptions : DefaultOtpTokenProviderOptions { }
 ```
 
 #### Service Collection
@@ -81,28 +81,27 @@ public class UserEmailOtpTokenOptions : DefaultOtpTokenProviderOptions {}
 public static IdentityBuilder AddCustomIdentityServices(
     this IServiceCollection services, GlobalSettings globalSettings)
 {
-    // Default Implementation
-    services.Configure<DefaultOtpTokenProviderOptions>(options =>
-    {
-        options.TokenLength = 6;
-        options.TokenAlpha = false;
-        options.TokenNumeric = true;
-        options.DistributedCacheEntryOptions = new DistributedCacheEntryOptions
-        {
-            AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(5)
-        };
-    });
-    // User email Options
+    // possible customization
     services.Configure<UserEmailOtpTokenOptions>(options =>
     {
         options.TokenLength = 8;
         // The other options are left default
     });
 
-    // TryAddScoped Default options
-    services.TryAddScoped<IOtpTokenProvider<DefaultOtpTokenProviderOptions>, OtpTokenProvider<DefaultOtpTokenProviderOptions>>();
-    // TryAddScoped UserEmail options
-    services.TryAddScoped<IOtpTokenProvider<UserEmailOtpTokenOptions>, OtpTokenProvider<UserEmailOtpTokenOptions>>();
+    // TryAddTransient open generics -> this allows us to inject IOtpTokenProvider<T> without having to specify the specific type here.
+    services.TryAddTransient(typeof(IOtpTokenProvider<>), typeof(OtpTokenProvider<>);
+}
+```
+
+#### Usage
+
+```csharp
+public class UserEmailTokenProvider(
+    IOtpTokenProvider<UserEmailOtpTokenOptions> otpTokenProvider
+)
+{
+    private readonly IOtpTokenProvider<UserEmailOtpTokenOptions> _otpTokenProvider = otpTokenProvider;
+    ...
 }
 ```
 
@@ -133,11 +132,13 @@ The cache key format uses three components: `{tokenProviderName}_{purpose}_{uniq
 #### Possible Email Token Provider Example
 
 Email token provider uses:
+
 - **Token Provider Name**: `"EmailToken"` (identifies the specific use case)
 - **Purpose**: `"EmailTwoFactorAuthentication"` (specific action being verified)
 - **Unique Identifier**: `"{user.Id}_{securityStamp}"` (user-specific data)
 
 These are passed into the OTP Token Provider which creates a cache record:
+
 - Cache Key: `EmailToken_EmailTwoFactorAuthentication_guid_guid`
 
 ## Security Considerations


src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/IOtpTokenProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Generates a new one-time password (OTP) based on the configured parameters.
+    /// The generated OTP is stored in the distributed cache with a key based on the unique identifier and purpose. If the
+    /// key is already in use, it will overwrite and generate a new OTP with a refreshed TTL.
+    /// </summary>
+    /// <param name="tokenProviderName">Name of the token provider, used to distinguish different token providers that may inject this class</param>
+    /// <param name="purpose">Purpose of the OTP token, used to distinguish different types of tokens.</param>
+    /// <param name="uniqueIdentifier">Unique identifier to distinguish one request from another</param>
+    /// <returns>generated token</returns>
```
JaredSnider-Bitwarden: Nit: generated token or null since you have `string?`

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/IOtpTokenProvider.cs
JaredSnider-Bitwarden, line 1: Excellent docs!

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
```diff
@@ -0,0 +1,173 @@
...
+### Cache Options
+
+See `DistributedCacheEntryOptions` documentation for a complete list of configuration options.
+
+| Property | Default | Description |
+|----------|---------|-------------|
+| `AbsoluteExpirationRelativeToNow` | 5 minutes | How long tokens remain valid |
+
+## Cache Key Format
+todo: add docs about unique
```
JaredSnider-Bitwarden: ❓ : has this been addressed or did you plan on adding a bit more here?

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Examples:
+
+#### Possible Email Token Provider
+Email token provider uses the `user.Id`, `securityStamp`, and `purpose` to create a unique key. This key can be passed into the OTP provider along with the specific use case purpose.
+
+- Use Case Purpose: `EmailToken`
+- Unique Identifier: `guid_guid_purpose`
+
+These are passed into the Otp Token Provider which create record in the cache:
```
JaredSnider-Bitwarden: nit: `creates a record`

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
```diff
@@ -0,0 +1,173 @@
...
+- Exact string matching for validation
+- Automatic removal after successful validation
+- Returns `false` for expired or non-existent tokens
+
+## Dependency Injection
+
+The provider is registered in `ServiceCollectionExtensions.cs`:
+
+```csharp
+services.AddScoped<IOtpTokenProvider, OtpTokenProvider>();
```
JaredSnider-Bitwarden: ❓ : should we go ahead and adopt the new `TryAdd` pattern from [ADR 26](https://contributing.bitwarden.com/architecture/adr/dotnet-dependency-injection-enhancements)?

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
JaredSnider-Bitwarden, line 1: Nice docs!

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -379,6 +380,7 @@ public static IdentityBuilder AddCustomIdentityServices(
         this IServiceCollection services, GlobalSettings globalSettings)
     {
         services.AddScoped<IOrganizationDuoUniversalTokenProvider, OrganizationDuoUniversalTokenProvider>();
+        services.AddScoped<IOtpTokenProvider, OtpTokenProvider>();
```
JaredSnider-Bitwarden: ❓ : should we go ahead and adopt the new `TryAdd` pattern from [ADR 26](https://contributing.bitwarden.com/architecture/adr/dotnet-dependency-injection-enhancements)?

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/OtpTokenProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    public int TokenLength { get; protected set; } = 6;
+    /// <summary>
+    /// Sets whether the token should contain alphabetic characters. Default is false.
+    /// </summary>
+    public bool TokenAlpha { get; protected set; } = false;
+    /// <summary>
+    /// Sets whether the token should contain numeric characters. Default is true.
+    /// </summary>
+    public bool TokenNumeric { get; protected set; } = true;
```
JaredSnider-Bitwarden: ⚠️  : the defaults live in two places - on the public properties and in the `OtpTokenProviderConfigurationOptions` object which doesn't feel great. 

After reviewing this configuration pattern and doing a bit of research, I think we might have to change gears here slightly.  The runtime configuration approach doesn't allow for modifying the data without a deployment if we wanted whereas if we were to use the [Options Pattern](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-9.0) then we could do that for whatever OTP provider required that. 

1. `OtpTokenProviderConfigurationOptions` becomes `OtpTokenProviderOptions`
```C#
public class OtpTokenProviderOptions
{
    public int TokenLength { get; set; } = 6;
    public bool TokenAlpha   { get; set; } = false;
    public bool TokenNumeric { get; set; } = true;
    public DistributedCacheEntryOptions CacheEntryOptions { get; set; }
        = new DistributedCacheEntryOptions {
            AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(5)
        };
}
```

2. I prefer the marker subclass + generic pattern over the enum pattern for identifying which options should be injected. So, we would require users of the OtpTokenProvider to use marker classes to name their options:
```C#
public class UserEmailOtpOptions : OtpTokenProviderOptions { }
public class SendAccessOtpOptions : OtpTokenProviderOptions { }
```

3. Interface changes to just (w/ docs):
```C#
public interface IOtpTokenProvider<TOptions>
    where TOptions : OtpTokenProviderOptions
{
    Task<string?> GenerateTokenAsync(string providerName, string purpose, string uniqueIdentifier);
    Task<bool> ValidateTokenAsync(string token, string providerName, string purpose, string uniqueIdentifier);
}
```

4. Class changes to something like:
```C#
public class OtpTokenProvider<TOptions> : IOtpTokenProvider<TOptions>
    where TOptions : OtpTokenProviderOptions { 
    
        private readonly IDistributedCache _distributedCache = _distributedCache;
        private readonly TOptions _options;

 public OtpTokenProvider(
        [FromKeyedServices("persistent")] IDistributedCache cache,
        IOptions<TOptions> optsAccessor)
    {
        _distributedCache = cache;
        _options  = optsAccessor.Value;
    }

... 
}
```

5. DI changes look something like:
```C#
// Non-app settings example
services.Configure<UserEmailOtpOptions>(opts => {
    // we can override defaults here
    opts.TokenLength = 8;
    // …
});

// Appsettings example:
services
  .AddOptions<UserEmailOtpOptions>()
  .Bind(Configuration.GetSection("OtpProviders:UserEmail"))
  .Configure(opts => {
     // Could add custom logic if required. 
      if (opts.TokenLength < 4) opts.TokenLength = 4;
  });
```
 Using appsettings here could let us have much better control over these settings.

```C#
// Register open generic provider
services.TryAddScoped(typeof(IOtpTokenProvider<>), typeof(OtpTokenProvider<>));  // note: should we consider using transient over scoped here?  
```

Happy to discuss this if need be!

JaredSnider-Bitwarden (changes requested):
Nice work! I think a pivot to the options pattern might serve us well in setting this up for future consumption a bit better, however. See details down below 👇

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
ike-kottlowski: https://github.com/bitwarden/server/pull/6081/commits/d419582f7416f1a6b13d1db1069c5948448762cc

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
ike-kottlowski: https://github.com/bitwarden/server/pull/6081/commits/d419582f7416f1a6b13d1db1069c5948448762cc

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/readme.md
ike-kottlowski: https://github.com/bitwarden/server/pull/6081/commits/d419582f7416f1a6b13d1db1069c5948448762cc

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
ike-kottlowski: https://github.com/bitwarden/server/pull/6081/commits/d419582f7416f1a6b13d1db1069c5948448762cc

src/Core/Auth/Identity/TokenProviders/OtpTokenProvider/OtpTokenProvider.cs
ike-kottlowski: Addressed https://github.com/bitwarden/server/pull/6081/commits/d879115e49e568246fe0de1f56065093d19aa705

I adopted the options pattern per our out of band discussion.

We have a default configuration for the IOtpTokenProvider but this can be extended by using the "subclass + marker" pattern you've shown above.

``` csharp
public class DefaultOtpTokenProviderOptions { }
public class UserEmailOtpOptions : DefaultOtpTokenProviderOptions { }
```

I did a simple PoC (as a test it is not included here) where I was able to register another `IOtpTokenProvider` but with a `UserEmailOtpOptions` configuration which gives us extensibility for the future.

``` csharp
services.TryAddScoped<IOtpTokenProvider<UserEmailOtpOptions>, OtpTokenProvider<UserEmailOtpOptions>>();
```

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -379,6 +379,17 @@ public static IdentityBuilder AddCustomIdentityServices(
...
+        services.Configure<DefaultOtpTokenProviderOptions>(options =>
+        {
+            options.TokenLength = 6;
+            options.TokenAlpha = false;
+            options.TokenNumeric = true;
+            options.DistributedCacheEntryOptions = new DistributedCacheEntryOptions
+            {
+                AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(5)
+            };
+        });
```
JaredSnider-Bitwarden: ⚠️ : I'm fairly sure that we don't need to do this. Wouldn't the default values of the `DefaultOtpTokenProviderOptions` be used anytime someone injects `IOptions<DefaultOtpTokenProviderOptions>`?

I think the `services.Configure<DefaultOtpTokenProviderOptions>(...)` is only really useful if we want to (1) override the defaults for a specific class or (2) if we want to pull in configurable defaults from `appsettings.json`

Can you do a quick test to see if that is correct?

JaredSnider-Bitwarden: If correct, we would also need to update the `readme`.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -379,6 +379,17 @@ public static IdentityBuilder AddCustomIdentityServices(
...
+        services.TryAddScoped<IOtpTokenProvider<DefaultOtpTokenProviderOptions>, OtpTokenProvider<DefaultOtpTokenProviderOptions>>();
```
JaredSnider-Bitwarden: ❓ : did you have a specific reason for not using the open-generic registration which is supposed to handle any future marker types?

```C#
services.TryAddScoped(
    typeof(IOtpTokenProvider<>),
    typeof(OtpTokenProvider<>)
);
```
^ this essentially says: “Whenever anybody asks for `IOtpTokenProvider<T>`, for any `T` that satisfies the generic constraint, give them an `OtpTokenProvider<T>`.”

Here's a source which explains this a bit: https://codetosolutions.com/blog/62/how-to-register-open-generics-in-.net-core-di.

JaredSnider-Bitwarden: ❓ : should we consider using `transient` over `scoped` since the `OtpTokenProvider` is really stateless?

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
ike-kottlowski: I think my though here was to show the process of modifying the options in the code. It is just the defaults, and we don't need to have it in here, but it could serve the purpose of documentation if and when we need to modify the options for different implementations.

I'm not committed enough to that train of thought though, addressed: https://github.com/bitwarden/server/pull/6081/commits/3b6feeca16046300cd4de705699a9d3bc99b965b

Readme update: https://github.com/bitwarden/server/pull/6081/commits/3dc863c707696423215a4ba66011bae90c72f002

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
ike-kottlowski: Old dog new tricks. I want to say something about being specific and know which providers are registered rather than having to go hunt for the implementations but that would make me sound like a luddite.

Addressed: https://github.com/bitwarden/server/pull/6081/commits/3b6feeca16046300cd4de705699a9d3bc99b965b

ike-kottlowski: Missed this in the first pass.

In this context the `scoped` service would operate like a transient service I would think. The code paths using the OtpTokenProvider would only send or only verify. So, functionally they're probably the same. But probably best to use transient either way.

Addressed: https://github.com/bitwarden/server/pull/6081/commits/3b6feeca16046300cd4de705699a9d3bc99b965b

JaredSnider-Bitwarden (approved):
Nice work! Thank you for your patience in implementing the changes!

PR #6100, opened 2025-07-17
[PM-17562] Add support for null/all event type

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

This PR alters the OrganizationIntegrationConfiguration table to receive `null` for Event Type. It then interprets `null` event types as being "all events". This allows the FE calling the API a way to set a configuration in one call that will apply to all events that occur for this integration. This is especially useful for HEC/SIEM integrations where the receiver is looking to consume all events via the integration and reduces the need to specify ~80 configurations that are identical except for the event type.

To interpret nulls correctly, the cache has been updated to make two reads:
1. Read based on the specific event that occurred
2. Read based on a null event type

It then merges the two results into one array to return. This allows for multiple configurations to exist (some which specify an explicit event, and some that cover all events with `null`).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Models/Data/Organizations/OrganizationIntegrationConfigurationDetails.cs
```diff
@@ -11,7 +11,7 @@ public class OrganizationIntegrationConfigurationDetails
     public Guid OrganizationId { get; set; }
     public Guid OrganizationIntegrationId { get; set; }
```
rkac-bw, line 12: I think the view needs these columns
  
  - Line 10: public Guid Id { get; set; } - Expects oic.[Id]
  - Line 12: public Guid OrganizationIntegrationId { get; set; } - Expects oi.[Id]
  
  /src/Sql/dbo/Views/OrganizationIntegrationConfigurationDetailsView.sql
  
  - Need to add?
    - oic.[Id]
    - oi.[Id] AS [OrganizationIntegrationId]

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationConfigurationDetailsCacheService.cs
```diff
@@ -32,12 +31,22 @@ public List<OrganizationIntegrationConfigurationDetails> GetConfigurationDetails
         IntegrationType integrationType,
         EventType eventType)
     {
-        var key = new IntegrationCacheKey(organizationId, integrationType, eventType);
-        return _cache.TryGetValue(key, out var value)
-            ? value
-            : new List<OrganizationIntegrationConfigurationDetails>();
-    }
+        var specificKey = new IntegrationCacheKey(organizationId, integrationType, eventType);
+        var allEventsKey = new IntegrationCacheKey(organizationId, integrationType, null);
```
withinfocus, line 35: ❓ We'd never support both existing right? So wouldn't this be fallback logic when there isn't a result found?

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationConfigurationDetailsCacheService.cs
```diff
@@ -32,12 +31,22 @@ public List<OrganizationIntegrationConfigurationDetails> GetConfigurationDetails
...
+        {
+            results.AddRange(specificConfigs);
+        }
+        if (_cache.TryGetValue(allEventsKey, out var fallbackConfigs))
+        {
+            results.AddRange(fallbackConfigs);
+        }
 
+        return results;
+    }
```
withinfocus, line 49: ⛏️ Empty lines between methods.

src/Sql/dbo/Stored Procedures/OrganizationIntegrationConfiguration_Create.sql
```diff
@@ -1,12 +1,12 @@
...
     @Id UNIQUEIDENTIFIER OUTPUT,
     @OrganizationIntegrationId UNIQUEIDENTIFIER,
-    @EventType SMALLINT,
     @Configuration VARCHAR(MAX),
     @Template VARCHAR(MAX),
     @CreationDate DATETIME2(7),
     @RevisionDate DATETIME2(7),
-    @Filters VARCHAR(MAX) = NULL
+    @Filters VARCHAR(MAX) = NULL,
+    @EventType SMALLINT = NULL
```
withinfocus: ❌ You don't need to move the parameter (you shouldn't), and given that you're changing the table definition alone the proc doesn't have to change at all. What you do need to do though is alter the proc given the schema change (after that).

src/Sql/dbo/Stored Procedures/OrganizationIntegrationConfiguration_Update.sql
```diff
@@ -1,12 +1,12 @@
...
     @Id UNIQUEIDENTIFIER OUTPUT,
     @OrganizationIntegrationId UNIQUEIDENTIFIER,
-    @EventType SMALLINT,
     @Configuration VARCHAR(MAX),
     @Template VARCHAR(MAX),
     @CreationDate DATETIME2(7),
     @RevisionDate DATETIME2(7),
-    @Filters VARCHAR(MAX) = NULL
+    @Filters VARCHAR(MAX) = NULL,
+    @EventType SMALLINT = NULL
```
withinfocus: ❌ Same as above.

src/Core/AdminConsole/Models/Data/Organizations/OrganizationIntegrationConfigurationDetails.cs
withinfocus: The model is portable and used for multiple use cases, so when the view is used to retrieve records several properties seen here are not needed.

withinfocus (changes requested):

util/Migrator/DbScripts/2025-07-17_00_AllowNullEventTypeOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,69 @@
+ALTER TABLE [dbo].[OrganizationIntegrationConfiguration]
+    ALTER COLUMN [EventType] SMALLINT NULL
+GO
+
```
rkac-bw, line 13: for idempotent changes should check if EventType column exists and is NOT already nullable
IF EXISTS (
    SELECT 1 
    FROM sys.columns 
    WHERE object_id = OBJECT_ID(N'[dbo].[OrganizationIntegrationConfiguration]') 
    AND name = 'EventType'
    AND is_nullable = 0  -- Currently NOT NULL
)
BEGIN
    ALTER TABLE [dbo].[OrganizationIntegrationConfiguration]
        ALTER COLUMN [EventType] SMALLINT NULL
END
GO

src/Core/AdminConsole/Services/Implementations/EventIntegrations/IntegrationConfigurationDetailsCacheService.cs
brant-livefront: I'm pretty sure we would want to support this - i.e. imagine I want a channel in Slack that receives _all_ events, but an admin receives only a handful of very high value events. I realize it'd be better to not have to do 2 reads every time if we could avoid it, but it seems like a valid case to me. Remember, you are allowed to have multiple configurations for one event type / integration. So you could have 2 different slack messages that you specified with the same explicit event type.

Or we could enforce this and say that configurations are all or individual, but not a mix. You could have a config for all, or you can have a config for one or more, but you can say "do this for every event but do this for just one."

src/Sql/dbo/Stored Procedures/OrganizationIntegrationConfiguration_Create.sql
```diff
@@ -1,7 +1,7 @@
 ﻿CREATE PROCEDURE [dbo].[OrganizationIntegrationConfiguration_Create]
     @Id UNIQUEIDENTIFIER OUTPUT,
     @OrganizationIntegrationId UNIQUEIDENTIFIER,
-    @EventType SMALLINT,
+    @EventType SMALLINT = NULL,
```
withinfocus: ❌ What I meant is you don't need to touch this at all. After you modify the table you will recreate the procs and this will be handled with no awareness by the proc.

util/Migrator/DbScripts/2025-07-17_00_AllowNullEventTypeOrganizationIntegrationConfiguration.sql
```diff
@@ -0,0 +1,78 @@
...
+    FROM sys.columns
+    WHERE object_id = OBJECT_ID(N'[dbo].[OrganizationIntegrationConfiguration]')
+    AND name = 'EventType'
+    AND is_nullable = 0 -- Currently NOT NULL
+)
+BEGIN
+    ALTER TABLE [dbo].[OrganizationIntegrationConfiguration]
+        ALTER COLUMN [EventType] SMALLINT NULL
+END
+GO
```
withinfocus, line 12: ❌ Also recreate the table's view as well as the details view -- this refreshes the metadata for nullability.

withinfocus:
@JimmyVo16 this is ready for team review.

PR #6112, opened 2025-07-22
[PM-22736] Send password hasher

ike-kottlowski:
## 🎟️ Tracking

[PM-22736](https://bitwarden.atlassian.net/browse/PM-22736)

## 📔 Objective

This accomplishes enablement work needed to implement the send authorization features. We are using the password hasher that key management has built to help standardize how we manage send password hashing for authorization.

Services are not currently injected anywhere in the code base. They are also not injected into the dependency pipeline but are set up to be injected in the future.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Auth/UserFeatures/PasswordValidation/PasswordValidationConstants.cs
+++ b/src/Core/Auth/UserFeatures/PasswordValidation/PasswordValidationConstants.cs
@@ -1,0 +1,6 @@
+﻿namespace Bit.Core.Auth.PasswordValidation;
+
+public static class PasswordValidationConstants
+{
+    public const int PasswordHasherKdfIterations = 100000;
+}
--- a/src/Core/KeyManagement/Sends/ISendPasswordHasher.cs
+++ b/src/Core/KeyManagement/Sends/ISendPasswordHasher.cs
@@ -1,0 +1,7 @@
+﻿namespace Bit.Core.KeyManagement.Sends;
+
+public interface ISendPasswordHasher
+{
+    bool VerifyPasswordHash(string sendPasswordHash, string inputPasswordHash);
+    string HashPasswordHash(string clientHashedPassword);
+}
--- a/src/Core/KeyManagement/Sends/SendPasswordHasher.cs
+++ b/src/Core/KeyManagement/Sends/SendPasswordHasher.cs
@@ -1,0 +1,29 @@
+﻿using Bit.Core.Entities;
+using Microsoft.AspNetCore.Identity;
+
+namespace Bit.Core.KeyManagement.Sends;
+
+public class SendPasswordHasher(IPasswordHasher<User> passwordHasher) : ISendPasswordHasher
+{
+    /// <summary>
+    /// Verifies an existing send password hash against a new input password hash.
+    /// </summary>
+    public bool VerifyPasswordHash(string sendPasswordHash, string inputPasswordHash)
+    {
+        if (string.IsNullOrWhiteSpace(sendPasswordHash) || string.IsNullOrWhiteSpace(inputPasswordHash))
+        {
+            return false;
+        }
+        var passwordResult = passwordHasher.VerifyHashedPassword(new User(), sendPasswordHash, inputPasswordHash);
+
+        return passwordResult is PasswordVerificationResult.Success or PasswordVerificationResult.SuccessRehashNeeded;
+    }
+
+    /// <summary>
+    /// Accepts a client hashed send password and returns a server hashed password.
+    /// </summary>
+    public string HashPasswordHash(string clientHashedPassword)
+    {
+        return passwordHasher.HashPassword(new User(), clientHashedPassword);
+    }
+}
--- a/src/Core/KeyManagement/Sends/SendPasswordServiceCollectionExtensions.cs
+++ b/src/Core/KeyManagement/Sends/SendPasswordServiceCollectionExtensions.cs
@@ -1,0 +1,17 @@
+﻿using Bit.Core.Auth.PasswordValidation;
+using Bit.Core.Entities;
+using Bit.Core.KeyManagement.Sends;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class SendPasswordServiceCollectionExtensions
+{
+    public static void AddSendPasswordServices(this IServiceCollection services)
+    {
+        services.TryAddScoped<IPasswordHasher<User>, PasswordHasher<User>>();
+        services.Configure<PasswordHasherOptions>(options => options.IterationCount = PasswordValidationConstants.PasswordHasherKdfIterations);
+        services.TryAddScoped<ISendPasswordHasher, SendPasswordHasher>();
+    }
+}
--- a/test/Core.Test/KeyManagement/SendPasswordHasherTests.cs
+++ b/test/Core.Test/KeyManagement/SendPasswordHasherTests.cs
@@ -1,0 +1,188 @@
+﻿using Bit.Core.Entities;
+using Bit.Core.KeyManagement.Sends;
+using Bit.Test.Common.AutoFixture;
+using Bit.Test.Common.AutoFixture.Attributes;
+using Microsoft.AspNetCore.Identity;
+using NSubstitute;
+using Xunit;
+
+namespace Bit.Core.Test.KeyManagement.Sends;
+
+[SutProviderCustomize]
+public class SendPasswordHasherTests
+{
+    [Theory]
+    [BitAutoData(PasswordVerificationResult.Success)]
+    [BitAutoData(PasswordVerificationResult.SuccessRehashNeeded)]
+    public void VerifyPasswordHash_WithValidMatching_ReturnsTrue(
+        PasswordVerificationResult passwordVerificationResult,
+        SutProvider<SendPasswordHasher> sutProvider,
+        string sendPasswordHash,
+        string inputPasswordHash)
+    {
+        // Arrange
+        sutProvider.GetDependency<IPasswordHasher<User>>()
+            .VerifyHashedPassword(Arg.Any<User>(), sendPasswordHash, inputPasswordHash)
+            .Returns(passwordVerificationResult);
+
+        // Act
+        var result = sutProvider.Sut.VerifyPasswordHash(sendPasswordHash, inputPasswordHash);
+
+        // Assert
+        Assert.True(result);
+        sutProvider.GetDependency<IPasswordHasher<User>>()
+            .Received(1)
+            .VerifyHashedPassword(Arg.Any<User>(), sendPasswordHash, inputPasswordHash);
+    }
+
+    [Theory, BitAutoData]
+    public void VerifyPasswordHash_WithNonMatchingPasswords_ReturnsFalse(
+        SutProvider<SendPasswordHasher> sutProvider,
+        string sendPasswordHash,
+        string inputPasswordHash)
+    {
+        // Arrange
+        sutProvider.GetDependency<IPasswordHasher<User>>()
+            .VerifyHashedPassword(Arg.Any<User>(), sendPasswordHash, inputPasswordHash)
+            .Returns(PasswordVerificationResult.Failed);
+
+        // Act
+        var result = sutProvider.Sut.VerifyPasswordHash(sendPasswordHash, inputPasswordHash);
+
+        // Assert
+        Assert.False(result);
+        sutProvider.GetDependency<IPasswordHasher<User>>()
+            .Received(1)
+            .VerifyHashedPassword(Arg.Any<User>(), sendPasswordHash, inputPasswordHash);
+    }
+
+    [Theory]
+    [InlineData(null, "inputPassword")]
+    [InlineData("", "inputPassword")]
+    [InlineData("   ", "inputPassword")]
+    [InlineData("sendPassword", null)]
+    [InlineData("sendPassword", "")]
+    [InlineData("sendPassword", "   ")]
+    [InlineData(null, null)]
+    [InlineData("", "")]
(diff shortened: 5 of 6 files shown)


src/Core/KeyManagement/Sends/SendPasswordServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Bit.Core.Auth.PasswordValidation;
+using Bit.Core.Entities;
+using Bit.Core.KeyManagement.Sends;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class SendPasswordServiceCollectionExtensions
```
mzieniukbw: Let's be specific on the purpose
```suggestion
public static class SendPasswordHasherServiceCollectionExtensions
```

src/Core/KeyManagement/Sends/ISendPasswordHasher.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Bit.Core.KeyManagement.Sends;
+
+public interface ISendPasswordHasher
+{
+    bool VerifyPasswordHash(string sendPasswordHash, string inputPasswordHash);
```
mzieniukbw: 🤔 "Verify" would imply a result different than `bool` (or thrown error).
Could we pick different naming, like "Compare" or "Matches" or "Equal" ?

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Bit.Core.Entities;
+using Microsoft.AspNetCore.Identity;
+
+namespace Bit.Core.KeyManagement.Sends;
+
+public class SendPasswordHasher(IPasswordHasher<User> passwordHasher) : ISendPasswordHasher
+{
+    /// <summary>
+    /// Verifies an existing send password hash against a new input password hash.
+    /// </summary>
```
mzieniukbw: It's better to put this documentation to `ISendPasswordHasher` and replace this with `<inheritdoc />`

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
```diff
@@ -0,0 +1,29 @@
...
+            return false;
+        }
+        var passwordResult = passwordHasher.VerifyHashedPassword(new User(), sendPasswordHash, inputPasswordHash);
+
+        return passwordResult is PasswordVerificationResult.Success or PasswordVerificationResult.SuccessRehashNeeded;
+    }
+
+    /// <summary>
+    /// Accepts a client hashed send password and returns a server hashed password.
+    /// </summary>
```
mzieniukbw: It's better to put this documentation to `ISendPasswordHasher` and replace this with `<inheritdoc />`

src/Core/KeyManagement/Sends/ISendPasswordHasher.cs
```diff
@@ -0,0 +1,7 @@
...
+    string HashPasswordHash(string clientHashedPassword);
```
mzieniukbw: It's much easier to understand if the method name also explicitly says the expected data
```suggestion
    string HashClientPasswordHash(string clientHashedPassword);
```

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Bit.Core.KeyManagement.Sends;
+
+public class SendPasswordHasher(IPasswordHasher<User> passwordHasher) : ISendPasswordHasher
+{
+    /// <summary>
+    /// Verifies an existing send password hash against a new input password hash.
+    /// </summary>
+    public bool VerifyPasswordHash(string sendPasswordHash, string inputPasswordHash)
+    {
```
mzieniukbw: It's not easy to understand what is expected from the arguments for the validation to pass.
The documentation (and argument names?) does not suggest that `inputPasswordHash` is in fact client's `clientPasswordHash`.
Some better documentation would also be nice, that would explain that `sendPasswordHash` is expected to be server encrypted already while the `inputPasswordHash` should be client encrypted only, where the (verification) comparison is made by server hasing the client password hash and comparing with (server hashed) send password hash.

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
```diff
@@ -0,0 +1,29 @@
...
+        if (string.IsNullOrWhiteSpace(sendPasswordHash) || string.IsNullOrWhiteSpace(inputPasswordHash))
+        {
+            return false;
+        }
+        var passwordResult = passwordHasher.VerifyHashedPassword(new User(), sendPasswordHash, inputPasswordHash);
+
+        return passwordResult is PasswordVerificationResult.Success or PasswordVerificationResult.SuccessRehashNeeded;
```
mzieniukbw, line 25: ❓ Why is `PasswordVerificationResult.SuccessRehashNeeded` treated as success ? From what i understand, this is is returned when the send password hash is in old format (Version 2: PBKDF2 with HMAC-SHA1 and 1000 iterations, barely adds any security) or when the iteration count is lower than expected (Version 3, but not with 100_000 iterations).
It might be better to be strict to not allow for any KDF downgrade attack.
If we allow this for legacy reasons, then are we going to re-hash, as we currently do in https://github.com/bitwarden/server/blob/c7b1c7f6d57de0e619a7c7163be4cb74378344da/src/Core/Tools/SendFeatures/Services/SendAuthorizationService.cs#L46 ?

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    /// Verifies an existing send password hash against a new input password hash.
+    /// </summary>
+    public bool VerifyPasswordHash(string sendPasswordHash, string inputPasswordHash)
+    {
+        if (string.IsNullOrWhiteSpace(sendPasswordHash) || string.IsNullOrWhiteSpace(inputPasswordHash))
+        {
+            return false;
+        }
+        var passwordResult = passwordHasher.VerifyHashedPassword(new User(), sendPasswordHash, inputPasswordHash);
```
mzieniukbw: how about constructing the `new User()` once ?
`private static readonly User USER = new User();`

test/Core.Test/KeyManagement/SendPasswordHasherTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        // Act
+        sutProvider.Sut.VerifyPasswordHash(sendPasswordHash, inputPasswordHash);
+
+        // Assert
+        Assert.NotNull(capturedUser);
+        Assert.IsType<User>(capturedUser);
+    }
+
+    [Theory, BitAutoData]
+    public void VerifyPasswordHash_WithMultipleCalls_CreatesNewUserInstanceEachTime(
```
mzieniukbw: I would argue that none of the `CreatesNewUserInstance` like tests adds any value, since we don't care about the provided User object in the `PasswordHasher<User>` implementation.

test/Core.Test/KeyManagement/SendPasswordHasherTests.cs
```diff
@@ -0,0 +1,188 @@
+﻿using Bit.Core.Entities;
+using Bit.Core.KeyManagement.Sends;
+using Bit.Test.Common.AutoFixture;
+using Bit.Test.Common.AutoFixture.Attributes;
+using Microsoft.AspNetCore.Identity;
+using NSubstitute;
+using Xunit;
+
+namespace Bit.Core.Test.KeyManagement.Sends;
```
mzieniukbw, line 8: the namespace does not match the directory. The Test should be placed in `test/Core.Test/KeyManagement/Sends/SendPasswordHasherTests.cs`

src/Core/KeyManagement/Sends/ISendPasswordHasher.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Bit.Core.KeyManagement.Sends;
+
+public interface ISendPasswordHasher
```
mzieniukbw, line 3: Please add some docs for this.

mzieniukbw (changes requested):
I have few concerns from KM side.

src/Core/KeyManagement/Sends/SendPasswordServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public static void AddSendPasswordServices(this IServiceCollection services)
+    {
+        services.TryAddScoped<IPasswordHasher<User>, PasswordHasher<User>>();
+        services.Configure<PasswordHasherOptions>(options => options.IterationCount = PasswordValidationConstants.PasswordHasherKdfIterations);
+        services.TryAddScoped<ISendPasswordHasher, SendPasswordHasher>();
```
mzieniukbw: ⚠️ This is wrong, since it technically moves all Password Hashing to be Sends specific and Auth owned, which in practise is used in all over Api.
If Auth wants to own the amount of iterations for Sends, then neither `PasswordHasherOptions`, nor `IPasswordHasher` should be injected like that, hence overriding entire Api service.
You should instead use `TryAddKeyedScoped`, provide service key and inject named `IPasswordHasher` into `SendPasswordHasher`.
The `PasswordHasherOptions` should follow the same pattern (named configuration), unless it's not possible, then maybe `PasswordHasher` instance should be created manually, instead of letting it be created by DI, like:
```
    class SendsPasswordHasherOptions : IOptions<PasswordHasherOptions>
    {
        public PasswordHasherOptions Value { get; } = new PasswordHasherOptions
        {
            CompatibilityMode = PasswordHasherCompatibilityMode.IdentityV3,
            IterationCount = PasswordValidationConstants.PasswordHasherKdfIterations
        };
    }

    services.TryAddKeyedScoped<IPasswordHasher<User>, () => new PasswordHasher<User>(new SendsPasswordHasherOptions())>();
```

mzieniukbw: Not to mention that Sends KDF iterations should be owned by Tools, unless i don't know about some ownership change on Sends ?

src/Core/KeyManagement/Sends/ISendPasswordHasher.cs
quexten: `HashOfClientPasswordHash`?

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
quexten: Noting, in the conversation between @audreyality , @JaredSnider-Bitwarden  and me, I think this was one of the discussed outcomes. Unlike traditional websites, where you input a low-entropy password to the above function, in our use-case we input a high-entropy, pre-hashed secret sent by the client. Thus, we don't really care about the hash iteration count on the server side, and realistically could reduce it to a single hash without much impact.

A leaked server DB would have an attacker brute-force both server + client iteration counts. These passwords are assumed to be unique for the send. Additionally, sends only live 30 days, and passwords are one-time. It seems unlikely that we would need to re-hash within a 30 day timeframe?

src/Core/Auth/UserFeatures/PasswordValidation/PasswordValidationConstants.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Bit.Core.Auth.PasswordValidation;
```
mzieniukbw: ❓ Namespace does not match the directory (unless intended?)

src/Core/KeyManagement/Sends/SendPasswordHasher.cs
mzieniukbw: Might be something we want to document as a comment in code, so anyone reading it would not immediately jump into conclusion like me, without knowing the context ?

audreyality: > These passwords are assumed to be unique for the send.

Just a point of clarity: We don't assume the passwords are unique. We're adding an OTP option for folks that want increased security.

src/Core/KeyManagement/Sends/SendPasswordServiceCollectionExtensions.cs
audreyality: > Not to mention that Sends KDF iterations should be owned by Tools, unless i don't know about some ownership change on Sends ?

The send password logic is [moving to the auth team](https://bitwarden.atlassian.net/browse/PM-19394). Tools will own the send authentication flow. The flow will invoke auth's interfaces to authenticate and/or authorize send access.

Passwords on sends are used to authorize access to the send payload. They're unrelated to send decryption.

src/Core/KeyManagement/Sends/SendPasswordServiceCollectionExtensions.cs
ike-kottlowski: https://github.com/bitwarden/server/pull/6112/commits/ec4b0d8a4aae7e2a27081c06568443a30ce47572

test/Core.Test/KeyManagement/SendPasswordHasherTests.cs
```diff
@@ -0,0 +1,103 @@
+﻿using Bit.Core.KeyManagement.Sends;
+using Bit.Test.Common.AutoFixture;
+using Bit.Test.Common.AutoFixture.Attributes;
+using Microsoft.AspNetCore.Identity;
+using NSubstitute;
+using Xunit;
+
+namespace Bit.Core.Test.KeyManagement.SendPasswordHasherTests;
```
mzieniukbw:
```suggestion
namespace Bit.Core.Test.KeyManagement.Sends;
```

JaredSnider-Bitwarden (approved):

PR #6116, opened 2025-07-23
[PM-23987] Cannot save item to Default User collection

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23987

## 📔 Objective

Previously, when fetching collections in `CipherService.SaveDetailsAsync`, we were retrieving the organization collections. This excluded the user's default collection, which caused save failures when the user tried to add an item to it.

This PR aims to replace it with a call to retrieve the user’s collections directly. This ensures their `DefaultUserCollection` is included, fixing the original issue.

This change also addressed a bug uncovered by integration tests: we were allowing organization members to add items to collections they didn’t have access to. By switching to user-scoped collections, we now properly enforce access control.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Vault/Services/Implementations/CipherService.cs
+++ b/src/Core/Vault/Services/Implementations/CipherService.cs
@@ -136,7 +136,11 @@ public class CipherService : ICipherService
         {
             if (cipher.OrganizationId.HasValue && collectionIds != null)
             {
-                var existingCollectionIds = (await _collectionRepository.GetManyByOrganizationIdAsync(cipher.OrganizationId.Value)).Select(c => c.Id);
+                var userCollections = await _collectionRepository.GetManyByUserIdAsync(savingUserId);
+                var existingCollectionIds = userCollections
+                    .Where(c => c.OrganizationId == cipher.OrganizationId.Value)
+                    .Select(c => c.Id)
+                    .ToList();
                 if (collectionIds.Except(existingCollectionIds).Any())
                 {
                     throw new BadRequestException("Specified CollectionId does not exist on the specified Organization.");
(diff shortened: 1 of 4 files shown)


src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -136,7 +136,11 @@ public async Task SaveDetailsAsync(CipherDetails cipher, Guid savingUserId, Date
         {
             if (cipher.OrganizationId.HasValue && collectionIds != null)
             {
-                var existingCollectionIds = (await _collectionRepository.GetManyByOrganizationIdAsync(cipher.OrganizationId.Value)).Select(c => c.Id);
+                var userCollections = await _collectionRepository.GetManyByUserIdAsync(savingUserId);
+                var existingCollectionIds = userCollections
+                    .Where(c => c.OrganizationId == cipher.OrganizationId.Value)
+                    .Select(c => c.Id)
+                    .ToList();
```
eliykat, line 143: For the current bug:
- what if an admin saves a cipher to a collection they have admin permissions to (e.g. can edit any collection) but are not assigned to? I think you mentioned they don't use this method - but is that a guarantee? Should this method be able to handle this scenario?
- what if the cipher is assigned to multiple collections, some of which the user isn't assigned to? (I'm not sure how this is handled generally.)

I don't know the answers to these questions and this domain can take up a lot of your time - I recommend opening it up to Vault's review. It may be safer to go back to querying collections by OrgId rather than by user permissions, but that's your call.

For the permission check issue:
- this would also need to check the user's collection permissions. However, even then the logic is slightly different: they need edit access to _at least one_ collection it's in, not _all_ collections it's in.
- `skipPermissionCheck` is generally confusing and its use is unclear. Sometimes it's used because the controller checks the permissions, but in this case it also appears to be used because the "permission check" in the service method is only for editing a cipher, not creating one. It's unclear where a permission check should go and how it should interact with this parameter. I don't want our team to be tacking on solutions - I think this really needs to be looked at (and preferably refactored) by the code owner team.

My recommendation is to fix the current bug here, and hand off the permissions issue to Vault to look into asap.

PR #6118, opened 2025-07-23
[PM-17562] Refactor event integration methods / declarations in ServiceCollectionExtensions

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

On a previous PR, @withinfocus pointed me to [Adopt TryAdd Dependency Injection Overloads](https://contributing.bitwarden.com/architecture/adr/dotnet-dependency-injection-enhancements/). I wanted to circle back and take a look at adopting more TryAdd methods in the setup of event integrations. I also wanted to clean up some of the methods and try consolidate and clarify setup.

To that end, it's probably not helpful to look at these changes in the diff viewer on GitHub. I've moved all of the private methods to the bottom of the file and rearranged some other pieces, so it will likely confuse the diff view. If you view it as the raw file, hopefully you agree it's more readable now.

I also updated our README documentation to reflect the change of where to put supporting service (like SlackService HttpClient) declarations. Having these in one place should be an improvement (no need to define things like cache twice or introduce bugs with slightly different declarations).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Models/Response/DomainsResponseModel.cs
+++ b/src/Api/Models/Response/DomainsResponseModel.cs
@@ -8,10 +8,10 @@ using Bit.Core.Models.Api;
 
 namespace Bit.Api.Models.Response;
 
-public class DomainsResponseModel : ResponseModel
+public class DomainsResponseModel() : ResponseModel("domains")
 {
     public DomainsResponseModel(User user, bool excluded = true)
-        : base("domains")
+        : this()
     {
         if (user == null)
         {
@@ -38,13 +38,13 @@ public class DomainsResponseModel : ResponseModel
     public IEnumerable<GlobalDomains> GlobalEquivalentDomains { get; set; }
 
 
-    public class GlobalDomains
+    public class GlobalDomains()
     {
         public GlobalDomains(
             GlobalEquivalentDomainsType globalDomain,
             IEnumerable<string> domains,
             IEnumerable<GlobalEquivalentDomainsType> excludedDomains,
-            bool excluded)
+            bool excluded) : this()
         {
             Type = (byte)globalDomain;
             Domains = domains;
--- a/src/Api/Vault/Models/Response/SyncResponseModel.cs
+++ b/src/Api/Vault/Models/Response/SyncResponseModel.cs
@@ -7,6 +7,7 @@ using Bit.Api.Tools.Models.Response;
 using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.Models.Data.Provider;
 using Bit.Core.Entities;
+using Bit.Core.KeyManagement.Models.Response;
 using Bit.Core.Models.Api;
 using Bit.Core.Models.Data;
 using Bit.Core.Models.Data.Organizations;
@@ -18,7 +19,7 @@ using Bit.Core.Vault.Models.Data;
 
 namespace Bit.Api.Vault.Models.Response;
 
-public class SyncResponseModel : ResponseModel
+public class SyncResponseModel() : ResponseModel("sync")
 {
     public SyncResponseModel(
         GlobalSettings globalSettings,
@@ -37,7 +38,7 @@ public class SyncResponseModel : ResponseModel
         bool excludeDomains,
         IEnumerable<Policy> policies,
         IEnumerable<Send> sends)
-        : base("sync")
+        : this()
     {
         Profile = new ProfileResponseModel(user, organizationUserDetails, providerUserDetails,
             providerUserOrganizationDetails, userTwoFactorEnabled, userHasPremiumFromOrganization, organizationIdsClaimingingUser);
@@ -54,6 +55,23 @@ public class SyncResponseModel : ResponseModel
         Domains = excludeDomains ? null : new DomainsResponseModel(user, false);
         Policies = policies?.Select(p => new PolicyResponseModel(p)) ?? new List<PolicyResponseModel>();
         Sends = sends.Select(s => new SendResponseModel(s, globalSettings));
+        UserDecryption = new UserDecryptionResponseModel
+        {
+            MasterPasswordUnlock = user.HasMasterPassword()
+                ? new MasterPasswordUnlockResponseModel
+                {
+                    Kdf = new MasterPasswordUnlockKdfResponseModel
+                    {
+                        KdfType = user.Kdf,
+                        Iterations = user.KdfIterations,
+                        Memory = user.KdfMemory,
+                        Parallelism = user.KdfParallelism
+                    },
+                    MasterKeyEncryptedUserKey = user.Key!,
+                    Salt = user.Email.ToLowerInvariant()
+                }
+                : null
+        };
     }
 
     public ProfileResponseModel Profile { get; set; }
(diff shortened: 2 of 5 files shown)


withinfocus (dismissed):
Definitely hard to read in a code review. Tests are passing but we'll have to see how this does once deployed just to ensure no runtime issues.

justindbaur:
Definitely a big improvement, just some thoughts on how we can maybe solve the `IHostedService` problem but I am also not right up against this code so very sorry if it sends you into a wild goose chase. 

To me, it seems like the problem is that you want to send in some level of dynamic primitives into the constructors of `AzureServiceBusEventListenerService` and `AzureServiceBusIntegrationListenerService`. To fix that we could instead employ the named options pattern first to limit the amount of dynamic primitives needed to just a single string. The constructor of `AzureServiceBusIntegrationListenerService` would change to:

```c#
public class ServiceBusIntegrationListenerOptions
{
    public int MaxRetries { get; set; }
    public string TopicName { get; set; }
    public string SubscriptionName { get; set; }
}

public AzureServiceBusIntegrationListenerService<T> : BackgroundService
    where T : IIntegrationTypeAccessor
{

    public AzureServiceBusIntegrationListenerService(IIntegrationHandler handler,
        IOptionsMonitor<ServiceBusIntegrationListenerOptions> options,
        IAzureServiceBusService serviceBusService,
        ILogger<AzureServiceBusIntegrationListenerService> logger)
    {
        _handler = handler;
        _logger = logger;
        var options = options.Get(T.Type.ToString());
        _maxRetries = options.MaxRetries;
        _serviceBusService = serviceBusService;

        _processor = _serviceBusService.CreateProcessor(options.TopicName, options.SubscriptionName, new ServiceBusProcessorOptions());
    }
    // ...rest of class...
}
```

The new generic is going to need to look like:
```c#
public interface IIntegrationTypeAccessor
{
    static abstract IntegrationType Type { get; }
}
```

and will have to be implemented for each integration type, returning something unique for each one.

```c#

Now when you register the service you should be able to do something like this:

```c#
private static IServiceCollection AddAzureServiceBusIntegration<TConfig, TIntegration>(
    this IServiceCollection services,
    string eventSubscriptionName,
    string integrationSubscriptionName,
    IntegrationType integrationType,
    GlobalSettings globalSettings)
    where TConfig : class
    where TIntegration : IIntegrationTypeAccessor
{
    services.AddOptions<ServiceBusIntegrationListenerOptions>(TIntegration.Type.ToString())
        .Configure(options =>
        {
            options.SubscriptionName = eventSubscriptionName;
            // ... others ...
        });

    services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, AzureServiceBusIntegrationListenerService<TIntegration>>();
}
```

Now you have a concrete type along with the hosted service to make sure it doesn't get added multiple times but you also have different versions of it per integration type still.

jrmccannon (dismissed):

brant-livefront:
@justindbaur Thanks for the comment on this! While I did end up on a bit of a wild goose chase, I think it turned out profitable and all of the services are now TryAdd. 👍 😄 

I couldn't type the listeners directly to the integration for a couple of reasons:
- The Repository listener at the Event level doesn't map to an integration, but rather writes all events to storage.
- The listener / handler pattern was intended to decouple listeners from the handler. A listener should just know which queue to listen to and DI gives it a handler to send the events to. So making it aware of integration felt like defeating some of that decoupling.

So instead, I've build a new set of configuration classes that we can use as the types to make sure each listener is uniquely typed. This also has the virtue of getting rid of a _lot_ of the boilerplate in the ServiceCollectionExtensions. And best of all, it means that a lot of the ASB vs. RabbitMQ code can be consolidated and cleaned up. Most of the setup is now done in one place, which makes it easier to read and less likely that we setup ASB differently than RabbitMQ.

cc @withinfocus

src/Core/AdminConsole/Models/Data/EventIntegrations/EventListenerConfiguration.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+
+public interface EventListenerConfiguration
```
withinfocus:
```suggestion
public interface IEventListenerConfiguration
```

Prefix?

src/Core/AdminConsole/Models/Data/EventIntegrations/HecListenerConfiguration.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        get => _globalSettings.EventLogging.RabbitMq.HecIntegrationQueueName;
+    }
+
+    public string IntegrationRetryQueueName
+    {
+        get => _globalSettings.EventLogging.RabbitMq.HecIntegrationRetryQueueName;
+    }
+
+    public string EventSubscriotionName
```
withinfocus:
```suggestion
    public string EventSubscriptionName
```

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationListenerConfiguration.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Bit.Core.Enums;
+
+namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+
+public interface IntegrationListenerConfiguration : EventListenerConfiguration
```
withinfocus: ... and maybe use the `I` throughout. We also often have interfaces and implementations in separate folders, but not sure if I love that pattern.

src/Core/AdminConsole/Models/Data/EventIntegrations/RepositoryListenerConfiguration.cs
```diff
@@ -0,0 +1,17 @@
...
+
+public class RepositoryListenerConfiguration(GlobalSettings globalSettings)
+    : ListenerConfiguration(globalSettings), EventListenerConfiguration
+{
+    public string EventQueueName
+    {
+        get => _globalSettings.EventLogging.RabbitMq.EventRepositoryQueueName;
+    }
+
+    public string EventSubscriotionName
```
withinfocus:
```suggestion
    public string EventSubscriptionName
```

Just check all of these.

withinfocus (changes requested):

src/Core/AdminConsole/Models/Data/EventIntegrations/EventListenerConfiguration.cs
brant-livefront: Good catch. These started life as class/subclass but changed to interfaces. I forgot to prefix them. Fixed in next commit.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
```diff
@@ -399,36 +399,44 @@ These names added here are what must match the values provided in the secrets or
 in Global Settings. This must be in place (and the local ASB emulator restarted) before you can use any
 code locally that accesses ASB resources.
 
+## ListenerConfiguration
+
+New integrations will need their own subclass of `ListenerConfiguration` which also conforms to
+`IntegrationListenerConfiguration`. This class provides a way of accessing the previously configured
```
withinfocus:
```suggestion
`IIntegrationListenerConfiguration`. This class provides a way of accessing the previously configured
```

Right?

withinfocus (changes requested):

src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md
brant-livefront: 🤦  yep. Good catch 👍

jrmccannon (approved):

PR #6121, opened 2025-07-24
[PM-23229] Add extra validation to kdf changes + authentication data + unlock data

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23229
https://bitwarden.atlassian.net/browse/PM-18024
Client side PR is here: https://github.com/bitwarden/clients/pull/15748
Note: Server and clients must be merged together, since this directly changes request models.

## 📔 Objective

Extracts KDF changes to a command, and adds validation and tests. Further, this adds support for the authentication data and unlock data models.

Note: This removes the KDF request model, since it no longer adds anything to the PasswordRequestModel. Validation is moved to the change kdf command.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/KeyManagement/Models/Data/MasterPasswordUnlockAndAuthenticationData.cs
```diff
@@ -0,0 +1,34 @@
+﻿#nullable enable
+using Bit.Core.Entities;
+using Bit.Core.Enums;
+
+namespace Bit.Core.KeyManagement.Models.Data;
+
+public class MasterPasswordUnlockAndAuthenticationData
```
quexten, line 7: Git diff is not showing this but this file is not net-new but moved from "MasterPasswordUnlockData"

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -264,8 +266,8 @@ public async Task PostKdf([FromBody] KdfRequestModel model)
             throw new UnauthorizedAccessException();
         }
 
-        var result = await _userService.ChangeKdfAsync(user, model.MasterPasswordHash,
-            model.NewMasterPasswordHash, model.Key, model.Kdf.Value, model.KdfIterations.Value, model.KdfMemory, model.KdfParallelism);
+        var result = await _changeKdfCommand.ChangeKdfAsync(user, model.MasterPasswordHash,
+            model.NewMasterPasswordHash, model.Key, model.ToKdfSettings(), model.AuthenticationData.ToData(), model.UnlockData?.ToData());
```
Thomas-Avery: ⚠️  Accessing `model.AuthenticationData` can throw a null reference exception and return a `500` error.

It looks like we are expecting these to be null based on the command implementation but not the interface? I have some confusion for what the expectation is.


Can we add something like the following so it is represented as a bad request from the client if they are expected for a change KDF request? Isn't the KDF change web client only?

```
     if (model.AuthenticationData == null)
        {
            throw new BadRequestException("AuthenticationData required");
        }

        if (model.UnlockData == null)
        {
            throw new BadRequestException("UnlockData required");
        }
 ```

💭  Could be a good opportunity to make a DTO for `ChangeKdfAsync` to get rid of all these params and put the validation on constructing that object.

File: src/Api/Auth/Models/Request/Accounts/PasswordRequestModel.cs [hunk 3]
```diff
@@ -1,17 +1,21 @@
-﻿// FIXME: Update this file to be null safe and then delete the line below
-#nullable disable
+﻿#nullable enable
```
Thomas-Avery, line 1:
```suggestion
```

The Api project now has nullable enabled project wide, so this doesn't need to be declared.

src/Api/KeyManagement/Models/Requests/InnerKdfRequestModel.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+/// <summary>
+/// Note: This should not be confused with the auth owned KDF request model, which represents a change to the KDF requested by a client. This model
+/// is merely a sub-model for other request models to transport KDF information in a non-flat deeply nested object.
+/// </summary>
+public class InnerKdfRequestModel
+{
+    public required KdfType KdfType { get; init; }
+    public required int Iterations { get; init; }
```
Thomas-Avery:
```suggestion
    [Required]
    public required KdfType KdfType { get; init; }
    [Required]
    public required int Iterations { get; init; }
```

Should these match `KdfRequestModel` with having the `[Required]` annotation?

src/Core/KeyManagement/Kdf/IChangeKdfCommand.cs
```diff
@@ -0,0 +1,18 @@
+﻿#nullable enable
```
Thomas-Avery:
```suggestion
```
The Core project now has nullable enabled project wide, so this doesn't need to be declared.

src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs
```diff
@@ -0,0 +1,80 @@
+﻿#nullable enable
```
Thomas-Avery:
```suggestion
```

Same thing here.

src/Core/KeyManagement/Kdf/IChangeKdfCommand.cs
```diff
@@ -0,0 +1,18 @@
...
+
+/// <summary>
+/// Command to change the Key Derivation Function (KDF) settings for a user. This includes 
+/// changing the masterpassword authentication hash, and the masterkey encrypted userkey.
+/// The salt must not change during the KDF change.
+/// </summary>
+public interface IChangeKdfCommand
+{
+    public Task<IdentityResult> ChangeKdfAsync(User user, string masterPasswordAuthenticationHash, string newMasterPasswordAuthenticationHash,
+        string masterKeyEncryptedUserKey, KdfSettings kdf, MasterPasswordAuthenticationData authenticationData, MasterPasswordUnlockData unlockData);
```
Thomas-Avery:
```suggestion
    public Task<IdentityResult> ChangeKdfAsync(User user, string masterPasswordAuthenticationHash, string newMasterPasswordAuthenticationHash,
        string masterKeyEncryptedUserKey, KdfSettings kdf, MasterPasswordAuthenticationData? authenticationData, MasterPasswordUnlockData? unlockData);
```

Match implementation if optional is what we're wanting here.

💭  Several params, primitives, and optionals. Consider a DTO and/or, for optionals, a different method entry point. Think what will make cleanup easier when we make the properties required?

src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+            user.Kdf = kdf.KdfType;
+            user.KdfIterations = kdf.Iterations;
+            user.KdfMemory = kdf.Memory;
+            user.KdfParallelism = kdf.Parallelism;
+            await _userRepository.ReplaceAsync(user);
+            await _pushService.PushLogOutAsync(user.Id);
+            return IdentityResult.Success;
+        }
+
+        return IdentityResult.Failed(_identityErrorDescriber.PasswordMismatch());
```
Thomas-Avery:
```suggestion
        _logger.LogWarning("Change KDF failed for user {userId}.", user.Id);
        return IdentityResult.Failed(_identityErrorDescriber.PasswordMismatch());
```
Missing logging of previous implementation.

src/Core/KeyManagement/Models/Data/KdfSettings.cs
```diff
@@ -0,0 +1,40 @@
+﻿#nullable enable
```
Thomas-Avery:
```suggestion
```

src/Core/KeyManagement/Models/Data/MasterPasswordAuthenticationData.cs
```diff
@@ -0,0 +1,20 @@
+﻿#nullable enable
```
Thomas-Avery:
```suggestion
```

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -264,8 +266,8 @@ public async Task PostKdf([FromBody] KdfRequestModel model)
             throw new UnauthorizedAccessException();
         }
```
Thomas-Avery, line 268: Doesn't look like `PostKdf` has any unit test coverage. Would be good to have the null cases for `model.AuthenticationData` and `model.UnlockData`

Thomas-Avery (changes requested):
A few questions from my side on what is intended.

src/Api/Auth/Controllers/AccountsController.cs
quexten: > It looks like we are expecting these to be null based on the command implementation but not the interface? I have some confusion for what the expectation is.

Indeed, that seems wrong. To clarify our expectation here; AuthenticationData and UnlockData are new values, replacing:
- KdfRequestModel's `Kdf`, `KdfIterations`, `KdfMemory,` `KdfParallelism`
- PasswordRequestModel's `Key`, `NewMasterPasswordHash`
However, we must keep backwards compatibility for a while, especially on `PasswordRequestModel`, since not all clients implement the new way of representing this data yet. So they are - for the time being - nullable.

> Can we add something like the following so it is represented as a bad request from the client if they are expected for a change KDF request? Isn't the KDF change web client only?

Yeah, that's true, for change kdf we can at runtime enforce them to be non-null, and that makes sense. I'll add that.

src/Core/KeyManagement/Kdf/IChangeKdfCommand.cs
quexten: I dropped the primitives now, and made the auth data / unlock data non optional, since as you point out changekdf is currently only on web clients.

src/Api/KeyManagement/Models/Requests/InnerKdfRequestModel.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Core.Enums;
+using Bit.Core.KeyManagement.Models.Data;
+
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+/// <summary>
+/// Note: This should not be confused with the auth owned KDF request model, which represents a change to the KDF requested by a client. This model
+/// is merely a sub-model for other request models to transport KDF information in a non-flat deeply nested object.
```
Thomas-Avery: Should we update this note since the `KdfRequestModel` is being deleted now?

src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        // Salt is ensured to be the same as unlock data, and the value stored in the account and not updated.
+        // KDF is ensured to be the same as unlock data above and updated below.
+
+        user.Key = unlockData.MasterKeyWrappedUserKey;
+        // Salt is ensured to be the same as authentication data, and the value stored in the account, and is not updated.
+        // Kdf - These will be seperated in the future, but for now are ensured to be the same as authentication data above.
+        user.Kdf = unlockData.Kdf.KdfType;
+        user.KdfIterations = unlockData.Kdf.Iterations;
+        user.KdfMemory = unlockData.Kdf.Memory;
+        user.KdfParallelism = unlockData.Kdf.Parallelism;
```
Thomas-Avery:
```suggestion
        // Authentication
        // Note: This mutates the user but does not yet save it to DB. That is done atomically, later.
        // This entire operation MUST be atomic to prevent a user from being locked out of their account.
        // Salt is ensured to be the same as unlock data, and the value stored in the account and not updated.
        // KDF is ensured to be the same as unlock data above and updated below.
        var result = await _userService.UpdatePasswordHash(user, authenticationData.MasterPasswordAuthenticationHash);
        if (!result.Succeeded)
        {
            _logger.LogWarning("Change KDF failed for user {userId}.", user.Id);
            return result;
        }

        // Salt is ensured to be the same as authentication data, and the value stored in the account, and is not updated.
        // Kdf - These will be seperated in the future, but for now are ensured to be the same as authentication data above.
        user.Key = unlockData.MasterKeyWrappedUserKey;
        user.Kdf = unlockData.Kdf.KdfType;
        user.KdfIterations = unlockData.Kdf.Iterations;
        user.KdfMemory = unlockData.Kdf.Memory;
        user.KdfParallelism = unlockData.Kdf.Parallelism;
```

Comments look a little off here to me. The way I'm reading them is, one should be associated with the usage of `authenticationData` and the other should be associated with the usage of `unlockData`?

test/Api.Test/Models/Request/Accounts/KdfRequestModelTests.cs
```diff
@@ -1,65 +0,0 @@
...
-using Bit.Core.Enums;
-using Xunit;
-
-namespace Bit.Api.Test.Models.Request.Accounts;
-
-public class KdfRequestModelTests
-{
-    [Theory]
-    [InlineData(KdfType.PBKDF2_SHA256, 1_000_000, null, null)] // Somewhere in the middle
-    [InlineData(KdfType.PBKDF2_SHA256, 600_000, null, null)] // Right on the lower boundary
```
Thomas-Avery, line 12: 💭 It would be nice to keep coverage for these lower and upper bound checks. What about converting these over to a `KdfSettingsValidatorTests` to test the static class `KdfSettingsValidator`?

Thomas-Avery (changes requested):
Looking good, a couple of minor things to take a look at.

src/Api/Auth/Controllers/AccountsController.cs
Thomas-Avery: Nice, thanks for the clarification.

src/Api/KeyManagement/Models/Requests/InnerKdfRequestModel.cs
quexten: Let's rename this one to `KdfRequestModel` then.

test/Api.Test/Models/Request/Accounts/KdfRequestModelTests.cs
quexten: Restored the tests for the static validator

src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs
quexten: Makes sense, and yes that was the intent. I've moved them.

src/Api/KeyManagement/Models/Requests/InnerKdfRequestModel.cs
Thomas-Avery, line 1: ⛏️  `InnerKdfRequestModel.cs` → `KdfRequestModel.cs`

Can we have file name match class name?

Thomas-Avery (approved):
LGTM, one optional change.

src/Api/KeyManagement/Models/Requests/InnerKdfRequestModel.cs
quexten: Renamed!

ike-kottlowski (approved):

PR #6021, opened 2025-07-02
[PM-22103] Exclude default collections from admin apis

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22103

## 📔 Objective

Update all Admin API endpoints that retrieve user collections to exclude collections of the type `DefaultUserCollection`.

Comments added to each repository method changed to clarify where its being used.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #6122, opened 2025-07-24
[PM-23987] Fix saving to default collections by updating collection lookup

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23987

## 📔 Objective

In [CipherService.SaveDetailsAsync](https://github.com/bitwarden/server/blob/c503ecbefc56b1c242f6fcbf660de6ac0970260f/src/Core/Vault/Services/Implementations/CipherService.cs#L139) we're retrieving all the Organization collections to check if it exists when creating a cipher and it was failing because we were excluding the type `DefaultUserCollection`.

This PR addresses this by creating a specific query to retrieve only Shared collections.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Sql/dbo/Stored Procedures/Collection_ReadSharedCollectionsByOrganizationId.sql
+++ b/src/Sql/dbo/Stored Procedures/Collection_ReadSharedCollectionsByOrganizationId.sql
@@ -10,5 +10,5 @@ BEGIN
         [dbo].[CollectionView]
     WHERE
         [OrganizationId] = @OrganizationId AND
-        [Type] = 0 -- DefaultUserCollection only
+        [Type] = 0 -- SharedCollections only
 END
--- a/util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
+++ b/util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
@@ -25,7 +25,7 @@ BEGIN
         [dbo].[CollectionView]
     WHERE
         [OrganizationId] = @OrganizationId AND
-        [Type] = 0 -- DefaultUserCollection only
+        [Type] = 0 -- SharedCollections only
 END
 GO
 


BTreston (dismissed):

util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
```diff
@@ -0,0 +1,30 @@
...
+
+    SELECT
+        *
+    FROM
+        [dbo].[CollectionView]
+    WHERE
+        [OrganizationId] = @OrganizationId
+END
+GO
+
```
rkac-bw, line 15: Are Admin Console Export and Import logic going to handles default collections?

util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
```diff
@@ -0,0 +1,30 @@
+CREATE OR ALTER PROCEDURE [dbo].[Collection_ReadByOrganizationId]
```
rkac-bw, line 1: Do we need v2 of these in case of rollback of app server ?

src/Api/Controllers/CollectionsController.cs
```diff
@@ -109,7 +109,7 @@ public async Task<ListResponseModel<CollectionResponseModel>> Get(Guid orgId)
         var readAllAuthorized = (await _authorizationService.AuthorizeAsync(User, CollectionOperations.ReadAll(orgId))).Succeeded;
         if (readAllAuthorized)
         {
-            orgCollections = await _collectionRepository.GetManyByOrganizationIdAsync(orgId);
+            orgCollections = await _collectionRepository.GetManySharedCollectionsByOrganizationIdAsync(orgId);
         }
         else
```
eliykat, line 114: I believe the `else` block is used to display collections in AC when the user cannot read all collections - so they just see the collections they're assigned to. In this case, do we want the default collection included? I don't think so, but please check with Priya. You may have to filter out the default collection in the `else` block if not.

That said - I'm not sure when this actually happens - can you view the Collections tab but not view all collections?

This is not strictly required for the fix here so you can look at it separately if you want.

src/Api/Vault/Controllers/CiphersController.cs
```diff
@@ -648,7 +648,7 @@ private async Task<bool> CanEditItemsInCollections(Guid organizationId, IEnumera
         if (await CanEditAllCiphersAsync(organizationId))
         {
             // TODO: This can likely be optimized to only query the requested ciphers and then checking they belong to the org
-            var orgCollections = (await _collectionRepository.GetManyByOrganizationIdAsync(organizationId)).ToDictionary(c => c.Id);
+            var orgCollections = (await _collectionRepository.GetManySharedCollectionsByOrganizationIdAsync(organizationId)).ToDictionary(c => c.Id);
```
eliykat: This is an authorization check - I'm not sure it needs to be limited to shared collections only. You can edit items in your default collection.

src/Sql/dbo/Stored Procedures/Collection_ReadSharedCollectionsByOrganizationId.sql
```diff
@@ -0,0 +1,14 @@
...
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT
+        *
+    FROM
+        [dbo].[CollectionView]
+    WHERE
+        [OrganizationId] = @OrganizationId AND
+        [Type] != 1 -- Exclude DefaultUserCollection
```
eliykat: ⛏️ for future proofing: this reads shared collections only, so checking for `[Type] = 0` is probably safer in case we add more collection types in the future.

util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
r-tome: No, the `DefaultUserCollection` type collections are still in development and feature flagged.

eliykat: I think it's fine because the interface isn't changing so it's not breaking, and we're just changing it back to how it was before https://github.com/bitwarden/server/pull/6021.

util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
eliykat: Currently in discussion but Product and Tools Team (who own export) are aware this will need to be reviewed either way. There are no default collections in prod currently so it doesn't need to be addressed in this PR.

eliykat (changes requested):

util/Migrator/DbScripts/2025-07-24_00_ReadSharedCollections.sql
r-tome: Good question! Yes, it should handle default collections. That is another issue this is addressing.

src/Api/Controllers/CollectionsController.cs
r-tome: > That said - I'm not sure when this actually happens - can you view the Collections tab but not view all collections?

You can't. That's one condition to be able to view the Admin Collections tab.

src/Sql/dbo/Stored Procedures/Collection_ReadSharedCollectionsByOrganizationId.sql
```diff
@@ -10,5 +10,5 @@ BEGIN
         [dbo].[CollectionView]
     WHERE
         [OrganizationId] = @OrganizationId AND
-        [Type] != 1 -- Exclude DefaultUserCollection
+        [Type] = 0 -- DefaultUserCollection only
```
eliykat: Typo I think:

```suggestion
        [Type] = 0 -- SharedCollections only
```

src/Sql/dbo/Stored Procedures/OrganizationUserUserDetails_ReadByOrganizationId_V2.sql
eliykat, line 1: This is used in a lot of places so I'm a little wary of hitting the same issue where we incorrectly exclude Default Collections for some caller. However, it looks like the only caller that actually requests collection and group associations is our AC query used for the members endpoint.

💭 out of scope here, but it makes me think that we should have separate queries:
- a query that _only_ returns the user details _without_ collections and groups - which is what most people use today (particularly other teams)
- a query that _always_ returns user details with (shared) collections and groups - which is what we use internally for AC. Reducing the risk of another caller getting confused about shared vs. default collections.

eliykat (approved):

rkac-bw (approved):
lgtm

PR #6133, opened 2025-07-28
fix: remove the duplicate name field

ruyut:


djsmith85 (approved):
@ruyut Nice catch! Changes are looking good and are ready to get merged.

PR #6138, opened 2025-07-29
Billing/pm 22692/secrets manager seat and serviceaccount bug

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22692

## 📔 Objective

Corrects the logic for sending an email to organization owners. Currently the Secrets Manager Seats and Service Accounts(Machine Secrets) being paid for are used to determine if the AutoScale Limit for each has been reached. This change corrects this to look at the current amount of occupied/used seats and service accounts instead of the count being paid for.

## 📸 Screenshots

https://github.com/user-attachments/assets/e7da1b2c-0b1b-4249-80a6-857090ee6cb2

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Models/Business/SecretsManagerSubscriptionUpdate.cs
```diff
@@ -50,11 +50,7 @@ public class SecretsManagerSubscriptionUpdate
     public bool MaxAutoscaleSmSeatsChanged => MaxAutoscaleSmSeats != Organization.MaxAutoscaleSmSeats;
     public bool MaxAutoscaleSmServiceAccountsChanged =>
         MaxAutoscaleSmServiceAccounts != Organization.MaxAutoscaleSmServiceAccounts;
-    public bool SmSeatAutoscaleLimitReached => SmSeats.HasValue && MaxAutoscaleSmSeats.HasValue && SmSeats == MaxAutoscaleSmSeats;
```
sbrown-livefront, line 53: 📝 Removing these helper functions because they are only used in the `UpdateSecretsManagerSubscriptionCommand` and because the correct logic requires an external property, it seems this would fit better in the validation logic of the command.

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
```diff
@@ -100,7 +92,7 @@ private async Task SendSeatLimitEmailAsync(Organization organization)
                     OrganizationUserType.Owner))
                 .Select(u => u.Email).Distinct();
 
-            await _mailService.SendSecretsManagerMaxSeatLimitReachedEmailAsync(organization, organization.MaxAutoscaleSmSeats.Value, ownerEmails);
+            await _mailService.SendSecretsManagerMaxSeatLimitReachedEmailAsync(organization, organization.MaxAutoscaleSmSeats!.Value, ownerEmails);
```
sbrown-livefront, line 95: 📝 This and the following Max are checked before being fed into this function

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
```diff
@@ -197,7 +189,7 @@ private async Task ValidateSmSeatsUpdateAsync(SecretsManagerSubscriptionUpdate u
             throw new BadRequestException("Organization has no Secrets Manager seat limit, no need to adjust seats");
         }
 
-        if (update.Autoscaling && update.SmSeats.Value < organization.SmSeats.Value)
+        if (update.Autoscaling && update.SmSeats!.Value < organization.SmSeats.Value)
```
sbrown-livefront, line 192: 📝 I added the following null forgives because the update seats are set by the organization initially and the validation for the path guarantees that a seat amount change has occurred. This is a patch for the warnings, but am I correct in my thinking?

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
```diff
@@ -220,7 +212,7 @@ private async Task ValidateSmSeatsUpdateAsync(SecretsManagerSubscriptionUpdate u
         }
 
         // Check minimum seats included with plan
-        if (plan.SecretsManager.BaseSeats > update.SmSeats.Value)
+        if (plan.SecretsManager.BaseSeats > update.SmSeats!.Value)
```
sbrown-livefront, line 215: ❓ The Free Plan seems to be the only plan with `BaseSeats` for the Secrets Manager. It doesn't seem like we can hit this. Is it still a valid path?

amorask-bitwarden: You're right that the Free plan is the only tier with base seats (2). The client does not allow a Free plan to scale seats through the UI, but it's possible for someone to try and hit the API directly with that type of invalid request so I think we can leave it in.

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
amorask-bitwarden: Yeah I'd assume so especially given the operator is just disabling compiler warnings and if what you said was not the case, we would consistently be hitting NREs during the invocation of this method.

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpdateSecretsManagerSubscriptionCommand.cs
sbrown-livefront: Solid! Thanks for the context.

amorask-bitwarden (approved):

PR #6139, opened 2025-07-29
[PM-24278] Create Remove Individual Vault validator

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24278

## 📔 Objective

This PR adds a validator for the Remove Individual Vault policy. When the policy is enabled via the Admin Console, it ensures that all confirmed members subject to the policy have a default collection.

Changes
1. Added a validator for the Remove Individual Vault policy.
3. Implemented logic to retrieve policy requirements and upsert default collections for applicable users.
5. Introduced a new pattern for retrieving multiple policy requirements for a single organization.

## 📸 Screenshots

This code is currently not hooked up and will be consumed in PR

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,69 @@
...
+            GetUserOrgIds(policyUpdate, requirements),
+            GetDefaultUserCollectionName());
+    }
+
+    private static string GetDefaultUserCollectionName()
+    {
+        return "Default";
+    }
+
+    private static List<Guid> GetUserOrgIds(PolicyUpdate policyUpdate, IEnumerable<OrganizationDataOwnershipPolicyRequirement> requirements)
```
JimmyVo16: Note: You can do this with LINQ, but it requires three iterations over the collection. Since we're expecting potentially 10k+ records, it's wiser to just iterate through it once.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        var requirements = await GetUserPolicyRequirementsByOrganizationIdAsync<OrganizationDataOwnershipPolicyRequirement>(policyUpdate.OrganizationId, policyUpdate.Type);
+
+        await _collectionRepository.UpsertDefaultCollectionsAsync(
+            policyUpdate.OrganizationId,
+            GetUserOrgIds(policyUpdate, requirements),
+            GetDefaultUserCollectionName());
+    }
+
+    private static string GetDefaultUserCollectionName()
```
JimmyVo16, line 62: This [ticket](https://bitwarden.atlassian.net/browse/PM-24279) will flesh this method out. I'm separating it into a different method to make it easier to expand later.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -25,18 +25,24 @@ public enum OrganizationDataOwnershipState
...
+    /// <param name="organizationUserIdsByOrgId">
+    /// A dictionary with the OrganizationId as the key and the OrganizationUserId as the value.
+    /// </param>
     public OrganizationDataOwnershipPolicyRequirement(
         OrganizationDataOwnershipState organizationDataOwnershipState,
-        IEnumerable<Guid> organizationIdsWithPolicyEnabled)
+        IEnumerable<Guid> organizationIdsWithPolicyEnabled,
+        Dictionary<Guid, Guid> organizationUserIdsByOrgId)
     {
-        _organizationIdsWithPolicyEnabled = organizationIdsWithPolicyEnabled ?? [];
```
JimmyVo16, line 39: Rider said this is redundant, and we have a system-wide initiative to move away from passing null around, so I think it's fine to remove `?? []`.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,62 @@
...
+    IPolicyRepository policyRepository,
+    ICollectionRepository collectionRepository,
+    IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories)
+    : OrganizationPolicyValidator(policyRepository, factories)
+{
+    public override PolicyType Type => PolicyType.OrganizationDataOwnership;
+
+    public override IEnumerable<PolicyType> RequiredPolicies => [];
+
+    public override Task<string> ValidateAsync(PolicyUpdate policyUpdate, Policy? currentPolicy) => Task.FromResult("");
```
JimmyVo16, line 26: Note: I don't think there is any validation for this policy. The [doc](https://bitwarden.atlassian.net/wiki/spaces/QA/pages/1746436110/Item+Organization+Milestone+1+-+Create+default+location+when+Remove+Individual+Vault+policy+is+on) explicitly states
> This code will live in a new IPolicyValidator for this policy. It doesn’t do anything except trigger this side-effect.

eliykat: That's correct, we only want the validator because it's how we trigger side effects. It's a good argument for separating this out in the future.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementQuery.cs
```diff
@@ -21,7 +21,6 @@ public async Task<T> GetAsync<T>(Guid userId) where T : IPolicyRequirement
 
         var policyDetails = await GetPolicyDetails(userId);
         var filteredPolicies = policyDetails
-            .Where(p => p.PolicyType == factory.PolicyType)
```
eliykat: The repository call needs to be updated so it's still filtering by PolicyType somewhere.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
eliykat: I think this is right, looking at the code flow null should never be passed in here.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyRequirementQuery.cs
JimmyVo16: I think we have to undo this and let it get handled in the ticket that will update PolicyDetails_ReadByUserId, which is fine since this path is no longer coupled to the org-level path.

JimmyVo16: I'll make sure to account for this in the other ticket (when I find it lol)

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    private async Task UpsertDefaultCollectionsForUsersAsync(PolicyUpdate policyUpdate)
+    {
+        var requirements = await GetUserPolicyRequirementsByOrganizationIdAsync<OrganizationDataOwnershipPolicyRequirement>(policyUpdate.OrganizationId, policyUpdate.Type);
+
+        var userOrgIds = GetUserOrgIds(policyUpdate, requirements);
+
+        if (!userOrgIds.Any())
+        {
+            _logger.LogWarning($"No UserOrganizationIds found for {policyUpdate.OrganizationId}");
```
JimmyVo16: Realistically, this shouldn't happen, but when you have a complex system, bugs can hide in unexpected places. Right there, I'm asking myself, "If this case happens, how would we know?" The answer is that we probably won't, at least not until someone reports a bug or something. It would be nice to have logs like these, so we can be proactive.

Now, there are issues with leaving logs like these, too. It can make the code more complex, and it's not clear whether this is an expected use case or if we're just being defensive. Maybe we create a method or class with documentation stating that this is for defensive logging and not an expected use case?

Ultimately, it's up to the team to decide whether we want more defensive logging. Let me know what you think. @eliykat

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            var userOrgId = requirement.GetOrganizationUserId(policyUpdate.OrganizationId);
+
+            if (userOrgId.HasValue)
+            {
+                userOrgIds.Add(userOrgId.Value);
+            }
+            else
+            {
+                _logger.LogWarning("UserOrganizationId is null for organization {OrganizationId}", policyUpdate.OrganizationId);
```
JimmyVo16: [same](https://github.com/bitwarden/server/pull/6139/files#r2254902385)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
```diff
@@ -280,6 +280,6 @@ private async Task HandleConfirmationSideEffectsAsync(Guid organizationId, IEnum
         }
 
         var organizationUserIds = confirmedOrganizationUsers.Select(u => u.Id).ToList();
-        await _collectionRepository.CreateDefaultCollectionsAsync(organizationId, organizationUserIds, defaultUserCollectionName);
+        await _collectionRepository.UpsertDefaultCollectionsAsync(organizationId, organizationUserIds, defaultUserCollectionName);
```
JimmyVo16: This isn't related to my PR, but I came across it because I changed the name of my method.

It looks like we're using the bulk creation path for both single-user and multiple-user confirmations. I understand that our best practice is to have the single-user use case go through the bulk interface. Normally, this is more efficient since it reduces the number of database transactions and has no trade-offs.

However, in this case, the bulk implementation uses SqlBulkCopy, which has higher overhead compared to a regular insert, so using it all the time may not be the most optimized solution. That said, I don’t know if this will cause an issue since I can’t test it in production, but I think it’s something we should keep an eye on.

test/Core.Test/Tools/ImportFeatures/ImportCiphersAsyncCommandTests.cs
audreyality, line 1: 🤔 Are there are any new scenarios that need to be added to these tests?

JimmyVo16: The `ImportCiphersAsyncCommand.cs` file doesn’t use the new part of the requirements, so we’re essentially just updating how `OrganizationDataOwnershipPolicyRequirement` is initialized. We do have tests for the new behavior in `OrganizationDataOwnershipPolicyRequirementFactoryTests.cs`.

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommandTests.cs
```diff
@@ -476,13 +476,13 @@ public async Task ConfirmUserAsync_WithCreateDefaultLocationEnabled_WithOrganiza
             .GetByOrganizationAsync<OrganizationDataOwnershipPolicyRequirement>(organization.Id)
             .Returns(new OrganizationDataOwnershipPolicyRequirement(
                 OrganizationDataOwnershipState.Enabled,
-                [organization.Id]));
-
+                [organization.Id]
```
r-tome: :art: the comma should be at the end of this line

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommandTests.cs
```diff
@@ -508,7 +508,8 @@ public async Task ConfirmUserAsync_WithCreateDefaultLocationEnabled_WithOrganiza
             .GetByOrganizationAsync<OrganizationDataOwnershipPolicyRequirement>(org.Id)
             .Returns(new OrganizationDataOwnershipPolicyRequirement(
                 OrganizationDataOwnershipState.Enabled,
-                [org.Id]));
+                [org.Id]
```
r-tome: :art: same thing for this comma

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/CollectionRepository/CreateDefaultCollectionsTests.cs
r-tome, line 1: We should update the test names and filename for the new method name

test/Core.Test/Vault/Services/CipherServiceTests.cs
```diff
@@ -195,10 +196,13 @@ public async Task SaveDetailsAsync_PersonalVault_WithPolicyRequirementsEnabled_W
             .IsEnabled(FeatureFlagKeys.PolicyRequirements)
             .Returns(true);
 
+
```
r-tome: :art: White lines can be removed

test/Core.Test/Tools/ImportFeatures/ImportCiphersAsyncCommandTests.cs
```diff
@@ -64,7 +64,8 @@ public async Task ImportIntoIndividualVaultAsync_WithPolicyRequirementsEnabled_W
             .GetAsync<OrganizationDataOwnershipPolicyRequirement>(importingUserId)
             .Returns(new OrganizationDataOwnershipPolicyRequirement(
                 OrganizationDataOwnershipState.Disabled,
-                []));
+                []
```
r-tome: :art: comma at the end of the line

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+    public override Task<string> ValidateAsync(PolicyUpdate policyUpdate, Policy? currentPolicy) => Task.FromResult("");
+
+    public override async Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? currentPolicy)
+    {
+        if (!featureService.IsEnabled(FeatureFlagKeys.CreateDefaultLocation))
+        {
+            return;
+        }
+
+        if (currentPolicy is not { Enabled: true } && policyUpdate is { Enabled: true })
```
r-tome:
```suggestion
        if (currentPolicy?.Enabled != true && policyUpdate.Enabled)
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    }
+
+    private async Task UpsertDefaultCollectionsForUsersAsync(PolicyUpdate policyUpdate)
+    {
+        var requirements = await GetUserPolicyRequirementsByOrganizationIdAsync<OrganizationDataOwnershipPolicyRequirement>(policyUpdate.OrganizationId, policyUpdate.Type);
+
+        var userOrgIds = GetUserOrgIds(policyUpdate, requirements);
+
+        if (!userOrgIds.Any())
```
r-tome, line 50: Nitpick: I think this has better performance
```suggestion
        if (userOrgIds.Count == 0)
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
r-tome: Non-blocking idea: maybe change it to `LogError` because it not an expected scenario and if it does happen we should assure its logged somewhere.

r-tome (commented):
It looks great to me, I just have some minor suggestions

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
JimmyVo16: Are you positive? Because, from my understanding, not all enumerations have Count. If the collection doesn’t, then it will iterate through the whole collection to get a count, which turns it into O(n) versus iterating to the first element.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -52,6 +58,18 @@ public bool RequiresDefaultCollection(Guid organizationId)
...
+    /// Return the OrganizationUserId for the given OrganizationId.
+    /// </summary>
+    public Guid? GetOrganizationUserId(Guid organizationId)
+    {
+        if (_organizationUserIdsByOrgId.TryGetValue(organizationId, out var orgUserId))
+        {
+            return orgUserId;
+        }
+        return null;
+    }
```
eliykat: This is a bit of a non-intuitive interface for a PolicyRequirement. I see the problem though; you don't just want to know whether the policy applies, you also want the OrgUserId, preferably without doing a separate db call.

There is also a more functional issue, which is that the factory includes any PolicyDetails where the user is in an accepted or confirmed state. This is required for (a) removing the individual vault (which takes effect as soon as the user is confirmed) and (b) creating a default collection on confirmation (because the user is still in the Accepted state until the end of the command). However, we don't want Accepted users to get a default collection when the policy is turned on.

So I recommend something like

```c#
public (bool CreateDefaultCollection, Guid? OrganizationUserId) CreateDefaultCollectionWhenPolicyEnabled(Guid organizationId)
```

This would return confirmed users only, and the interface is a clearer expression of what it's for.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
eliykat: It's a `List`, which stores its count internally:

> Retrieving the value of this property is an O(1) operation.

([ref](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.count?view=net-9.0))

Rider often suggests this, I'm not a huge fan though. I like how expressive `Any()` is and it works on all IEnumerables. (And surely `Any()` is O(1) as well for any reasonable implementation?)

Not a big deal either way. I asked in `#code` to see if anyone has strong opinions.

EDIT: SonarCloud agrees with Rui, so maybe just make the machine happy.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
```diff
@@ -0,0 +1,47 @@
...
+    protected async Task<IEnumerable<T>> GetUserPolicyRequirementsByOrganizationIdAsync<T>(Guid organizationId, PolicyType policyType) where T : IPolicyRequirement
+    {
+        var factory = factories.OfType<IPolicyRequirementFactory<T>>().SingleOrDefault();
+        if (factory is null)
+        {
+            throw new NotImplementedException("No Requirement Factory found for " + typeof(T));
+        }
+
+        var policyDetails = await policyRepository.GetPolicyDetailsByOrganizationIdAsync(organizationId, policyType);
+        var policyDetailGroups = policyDetails.GroupBy(policyDetail => policyDetail.UserId);
```
eliykat, line 24: Just double checking, will `UserId` ever be `null`?

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirementFactoryTests.cs
eliykat, line 1: Related to my previous comment - please add tests around user status.

Also review the CodeCov results and see if there's anything else you can improve coverage on. e.g. the `OrganizationPolicyValidator`.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
r-tome: I thought about this PR when I saw that thread!

Like I said on there: using `Count` is faster because `Any() `does some type checks. The difference can be negligible because its nanoseconds difference.

I see the case for `Any()` that is easier to read and looks better. I just thought I'd mention it because why not go for the more efficient route?

audreyality (approved):
👍🏻 Tools code LGTM!

📝 There's a lot of feedback from other teams; if you need to clear my review, please wait until all feedback is addressed to tag me back in. 😄

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
JimmyVo16: OH, that's right, it's a list, which is why it's `Count` and not `Count()`.

> SonarCloud agrees with Rui, so maybe just make the machine happy.

How can I see this? I don't see it in the sonar [job](https://github.com/bitwarden/server/actions/runs/16977238260/job/48129155316?pr=6139) at a glance.

JimmyVo16: OH, I see it now. It's in this [link](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=6139&id=bitwarden_server&open=AZiqgsbBAjQ-11VnT1gf)

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
JimmyVo16: This is a really good catch. I double-checked and found that the sproc does not return the UserId. I’ve made a [fix](https://github.com/bitwarden/server/pull/6203/files) for it.

JimmyVo16:
Update: Still going over the code review comments, and will look at the CodeCov and Sonar results.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
JimmyVo16: > creating a default collection on confirmation (because the user is still in the Accepted state until the end of the command)

This is a good catch.

I’m gonna spend some time digesting this more. Definitely will keep your proposal in mind.

JimmyVo16: Let me know what you think of the [commit](https://github.com/bitwarden/server/pull/6139/commits/6714f0c75214a40f94527e173c716cf7fb5926b5) that addresses this.

2 minor things

1. Instead of having a `Guid?`, I just passed in an empty Guid. This is nice because we don’t have to handle null everywhere, but I’m not sure if this is a good pattern. The class and `ShouldCreateDefaultCollection` should provide enough context, and all the properties are read-only, so they can’t be modified later. But obviously, someone could still use the Guid without a check; at least with nullable, it forces you to handle it.
2. The class is a bit nicer than a tuple, but it’s not a big deal. If we’re going to stick with the class, can it live in this file? This follows a similar requirement pattern where the factory and requirement object are in the same class.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -60,18 +60,26 @@ public bool RequiresDefaultCollection(Guid organizationId)
...
+        return new DefaultCollectionRequest(Guid.Empty, true);
     }
 }
 
+public class DefaultCollectionRequest(Guid organizationUserId, bool shouldCreateDefaultCollection)
+{
+    public readonly bool ShouldCreateDefaultCollection = shouldCreateDefaultCollection;
+    public readonly Guid OrganizationUserId = organizationUserId;
+}
+
```
eliykat: Perfect use case for a `record`, which gives you a really nice one liner:

```c#
public record DefaultCollectionRequest(Guid organizationUserId, bool shouldCreateDefaultCollection);
```

It's OK to keep it in this file.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -60,18 +60,26 @@ public bool RequiresDefaultCollection(Guid organizationId)
...
+    /// <returns>A DefaultCollectionRequest containing the organization user ID and creation flag.</returns>
+    public DefaultCollectionRequest GetDefaultCollectionRequest(Guid organizationId)
     {
         if (_organizationUserIdsByOrgId.TryGetValue(organizationId, out var orgUserId))
         {
-            return orgUserId;
+            return new DefaultCollectionRequest(orgUserId, true);
         }
-        return null;
+        return new DefaultCollectionRequest(Guid.Empty, true);
```
eliykat: Initially I agreed with you that `null` is preferable because it makes the caller handle the `null`. But the intention here is that the caller checks the bool first, so I think it's OK.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -60,18 +60,26 @@ public bool RequiresDefaultCollection(Guid organizationId)
     }
 
     /// <summary>
-    /// Return the OrganizationUserId for the given OrganizationId.
+    /// Get the user's default collection request for the specified organization.
```
eliykat: I think this could use slightly more explanation, e.g. `...which indicates whether the user should have a default collection created for them, and if so, the relevant OrganiationUserId to create the collection for`.

Also, once you exclude `Accepted` orgUsers from this method, the name (and xmldoc) will need to reflect its specific purpose (i.e. enforcing the requirement when the policy is enabled).

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
eliykat: I like this as a solution to the interface and I've left a few comments in a new review.

Excluding users in the Accepted status still needs to be handled.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
JimmyVo16: AC team's verdict on Defensive Logging: Accepted pattern here, as the debugging benefits outweigh the concern that it might obscure intent, since code review helps address this.

JimmyVo16:
I'm looking into this failing test.

```
Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories.CollectionRepository.UpsertDefaultCollectionsTests.UpsertDefaultCollectionsAsync_ShouldNotCreateDefaultCollection_WhenUsersAlreadyHaveOne

Error: Microsoft.Data.SqlClient.SqlException : Transaction (Process ID 53) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
```

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
```diff
@@ -0,0 +1,277 @@
+﻿using Bit.Core.AdminConsole.Entities;
```
JimmyVo16, line 1: Starting this thread so others can easily reply, but it’s not related to this line of code.

```
Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories.CollectionRepository.UpsertDefaultCollectionsTests.UpsertDefaultCollectionsAsync_ShouldNotCreateDefaultCollection_WhenUsersAlreadyHaveOne

Error: Microsoft.Data.SqlClient.SqlException : Transaction (Process ID 53) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
```

The tests that use `BulkResourceCreationService.CreateCollectionsUsersAsync` is failing intermittently due to a deadlock.. It’s not always the same test that fails, and they don’t fail consistently; it depends on which one is selected as the deadlock victim. Failed jobs: [1](https://github.com/bitwarden/server/actions/runs/17104174322/job/48508599680?pr=6139) and [2](https://github.com/bitwarden/server/actions/runs/16987060511/job/48158122141)

I think the main reason is that we’re using SqlBulkCopy, which likely has high overhead, and CollectionUser gets hit frequently.

Since this code is already merged, I’ll need to create a bug ticket to address it.

JimmyVo16: Created [ticket](https://bitwarden.atlassian.net/browse/PM-25056)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
JimmyVo16: So, after this [bug](https://github.com/bitwarden/server/pull/6139#discussion_r2289203090), I strongly advise against calling this method for a single user confirmation, since that increases the risk of locking the CollectionUser.

This is outside the scope of this ticket, so if we want to address it, I can create a separate ticket.

cc @r-tome @Thomas-Avery

JimmyVo16:
@eliykat 

This PR is ready for review. There are a few sidebar conversations, but we can move them to Slack if needed. They shouldn’t block this PR.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -63,10 +88,13 @@ public override OrganizationDataOwnershipPolicyRequirement Create(IEnumerable<Po
         var organizationDataOwnershipState = policyDetails.Any()
             ? OrganizationDataOwnershipState.Enabled
             : OrganizationDataOwnershipState.Disabled;
-        var organizationIdsWithPolicyEnabled = policyDetails.Select(p => p.OrganizationId).ToHashSet();
+
+        var policyDetailsByOrganizationId = policyDetails
```
JimmyVo16: I just noticed this, but maybe this part can be moved into the method. One of the benefits is that the computation will only be done when the method is used, making it more optimized.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
rkac-bw: I think part of the problem is that the guid is not using the sequential guid method ,  Random CollectionId causes fragmentation in the first part of the composite key and needs to insert into random places for the primary key increasing locking and deadlocking,  GenerateComb() GUIDs will reliably append to the end of indexes, making them ideal for preventing fragmentation and appending to end of table pk sort order

https://github.com/bitwarden/server/blob/de1d0b5ba7f21f6b3f822d049489290f4df0557b/src/Infrastructure.Dapper/Repositories/CollectionRepository.cs#L396C12-L396C47

Currently:
`var collectionId = Guid.NewGuid();
`
Should be 
`var collectionId = GenerateComb()`

you could then sort by collectionid for insert (optionally add a tablelock)

```
  // In CollectionRepository.BuildDefaultCollectionForUsers:
  collections = collections.OrderBy(c => c.Id).ToList();
  collectionUsers = collectionUsers
      .OrderBy(cu => cu.CollectionId)
      .ThenBy(cu => cu.OrganizationUserId)
      .ToList();
```

 ```
 // In BulkResourceCreationService:
  using var bulkCopy = new SqlBulkCopy(
      connection,
      SqlBulkCopyOptions.KeepIdentity | SqlBulkCopyOptions.TableLock,
      transaction)
  {
      DestinationTableName = "[dbo].[Collection]",
      BatchSize = 5000,
      BulkCopyTimeout = 120,
      EnableStreaming = true
  };
```

 ```
 // Add order hints
  bulkCopy.ColumnOrderHints.Add("Id", SortOrder.Ascending);
```

rkac-bw: Also, not sure we need `KeepIdentity ` as these are pregenerated pk's and not using an identity value

JimmyVo16: @rkac-bw Do you mind moving this conversation to this [ticket](https://bitwarden.atlassian.net/browse/PM-25056) so whoever picks it up has the context?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
r-tome: We should definitely address it. Please share the ticket when you create it

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirementFactoryTests.cs
```diff
@@ -30,24 +31,85 @@ public void State_WithOrganizationDataOwnershipPolicies_ReturnsRestricted(
...
     [Theory, BitAutoData]
-    public void RequiresDefaultCollection_WithNoPolicies_ReturnsFalse(
-        Guid organizationId,
+    public void PolicyType_ReturnsOrganizationDataOwnership(SutProvider<OrganizationDataOwnershipPolicyRequirementFactory> sutProvider)
+    {
+        Assert.Equal(PolicyType.OrganizationDataOwnership, sutProvider.Sut.PolicyType);
+    }
+
+    [Theory, BitAutoData]
+    public void GetDefaultCollectionRequest_WithConfirmedUser(
```
r-tome: Test names should specify what is the expected outcome. In this case something like `GetDefaultCollectionRequest_WithConfirmedUser_ReturnsTrue`

r-tome (approved):
Great work, looks good to me! Just a minor suggestion on the test names

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -46,14 +47,34 @@ public OrganizationDataOwnershipPolicyRequirement(
...
+    /// <param name="organizationId">The organization ID to create the request for.</param>
+    /// <returns>A DefaultCollectionRequest containing the OrganizationUserId and a flag indicating whether to create a default collection.</returns>
+    public DefaultCollectionRequest GetDefaultCollectionRequest(Guid organizationId)
     {
-        return _organizationIdsWithPolicyEnabled.Contains(organizationId);
+        var policyDetail = _policyDetails
+            .FirstOrDefault(p => p.OrganizationId == organizationId);
+
+        if (policyDetail != null && policyDetail.HasStatus([OrganizationUserStatusType.Confirmed]))
+        {
```
rkac-bw, line 63: Is this an issue? this checks that user is confirmed but line 51 and 52 reference accepted users?

JimmyVo16: Good catch! The documentation was wrong.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
JimmyVo16: Cool. Please move all future discussions to the [ticket](https://bitwarden.atlassian.net/browse/PM-25138).

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -52,26 +47,34 @@ public OrganizationDataOwnershipPolicyRequirement(
...
-    /// Returns true if the Organization Data Ownership policy is enforced in that organization.
+    /// Gets a default collection request for enforcing the Organization Data Ownership policy.
+    /// Only confirmed users are applicable.
+    /// This indicates whether the user should have a default collection created for them when the policy is enabled,
+    /// and if so, the relevant OrganizationUserId to create the collection for.
     /// </summary>
-    public bool RequiresDefaultCollection(Guid organizationId)
+    /// <param name="organizationId">The organization ID to create the request for.</param>
+    /// <returns>A DefaultCollectionRequest containing the OrganizationUserId and a flag indicating whether to create a default collection.</returns>
+    public DefaultCollectionRequest GetDefaultCollectionRequest(Guid organizationId)
```
eliykat: I still think this method name could describe its purpose, rather than its return value (which is already described by the type system). When should this be called? What is it used for? `RequiresDefaultCollectionWhenPolicyEnabled` is wordy but more useful to callers.

The xmldoc provides some of this context but I don't think that's a substitute for a descriptive name.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
eliykat: I think your latest implementation is overall much simpler, good job.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        var featureService = Substitute.For<IFeatureService>();
+        featureService
+            .IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
+            .Returns(true);
+
+        var sut = new OrganizationDataOwnershipPolicyValidator(policyRepository, collectionRepository, [factory], featureService, logger);
+        return sut;
+    }
+
+    private static List<OrganizationPolicyDetails> GenerateOrganizationPolicyDetails(PolicyUpdate policyUpdate)
```
eliykat: This isn't the worst thing in the world, and LLMs make it way easier to smash out test data like this, but we have tooling for this in the form of the AutoData library. For example, you can inject:

```cs
        [PolicyDetails(PolicyType.OrganizationDataOwnership)] IEnumerable<PolicyDetails> policyDetails,
```

and basically get the same object (a sequence of 3 PolicyDetails), except that you'd need to set the OrgId.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        logger ??= Substitute.For<ILogger<OrganizationDataOwnershipPolicyValidator>>();
+
+        var featureService = Substitute.For<IFeatureService>();
+        featureService
+            .IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
+            .Returns(true);
+
+        var sut = new OrganizationDataOwnershipPolicyValidator(policyRepository, collectionRepository, [factory], featureService, logger);
+        return sut;
+    }
```
eliykat, line 237: This should be replaced by using `SutProvider`. See existing tests for example usage. But basically when you inject it, it creates a new sut with NSubstitute mocks as its dependency, then you can just configure the mocks.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidatorTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        var factory = Substitute.For<IPolicyRequirementFactory<TestPolicyRequirement>>();
+        factory.Create(Arg.Any<IEnumerable<PolicyDetails>>()).Returns(new TestPolicyRequirement());
+        factory.Enforce(Arg.Any<PolicyDetails>()).Returns(true);
+
+        sutProvider.GetDependency<IPolicyRepository>()
+            .GetPolicyDetailsByOrganizationIdAsync(organizationId, PolicyType.TwoFactorAuthentication)
+            .Returns(policyDetails);
+
+        var factories = new List<IPolicyRequirementFactory<IPolicyRequirement>> { factory };
+        var sut = new TestOrganizationPolicyValidator(sutProvider.GetDependency<IPolicyRepository>(), factories);
```
eliykat, line 59: Use `SutProvider.Sut` rather than initializing a new one. That's what `SutProvider` is for.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidatorTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        Guid userId1,
+        Guid userId2,
+        SutProvider<TestOrganizationPolicyValidator> sutProvider)
+    {
+        // Arrange
+        var policyDetails = new List<OrganizationPolicyDetails>
+        {
+            new() { UserId = userId1, OrganizationId = organizationId },
+            new() { UserId = userId1, OrganizationId = organizationId },
+            new() { UserId = userId2, OrganizationId = organizationId }
```
eliykat: Slight misarrangement of data: 1 user would never have 2 PolicyDetails for the same org. So I think it would be more like

```suggestion
            new() { UserId = userId1, OrganizationId = organizationId },
            new() { UserId = userId1, OrganizationId = Guid.NewGuid() },
            new() { UserId = userId2, OrganizationId = organizationId }
```

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidatorTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        var sut = new TestOrganizationPolicyValidator(sutProvider.GetDependency<IPolicyRepository>(), factories);
+
+        // Act
+        var result = await sut.TestGetUserPolicyRequirementsByOrganizationIdAsync<TestPolicyRequirement>(
+            organizationId, PolicyType.TwoFactorAuthentication);
+
+        // Assert
+        Assert.Equal(2, result.Count());
+
+        factory.Received(2).Create(Arg.Any<IEnumerable<PolicyDetails>>());
```
eliykat: It would be good to check that it grouped them correctly.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
JimmyVo16: I don't think `RequiresDefaultCollectionWhenPolicyEnabled` is a good name. The `WhenPolicyEnabled` part shouldn't really be there; it’s business logic, and now it’s duplicated in two places. If we modify the policy, this will need to be updated as well. Also, `RequiresDefaultCollection` sounds like it should be a boolean, but we're actually returning an object. The object itself doesn’t contain logic to dictate what to do.

`GetDefaultCollectionRequest` isn't great either, but the Get part makes more sense for an object. Maybe we can name the class something better than `DefaultCollectionRequest`?

Here are the side-by-side uses in code. `RequiresDefaultCollectionWhenPolicyEnabled` feels off to me...


```
        var userOrgIds = requirements
            .Select(requirement => requirement.RequiresDefaultCollectionWhenPolicyEnabled(policyUpdate.OrganizationId))
            .Where(request => request.ShouldCreateDefaultCollection)
            .Select(request => request.OrganizationUserId);
```

```
        var userOrgIds = requirements
            .Select(requirement => requirement.GetDefaultCollectionRequest(policyUpdate.OrganizationId))
            .Where(request => request.ShouldCreateDefaultCollection)
            .Select(request => request.OrganizationUserId);
```

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
JimmyVo16: `SutProvider` doesn't work since we're mocking the factories. If you look at the tests in the same class that don't involve the factory logic, I did just use `SutProvider` for those. I think [PolicyRequirementQueryTests](https://github.com/bitwarden/server/blob/e042572cfb20bc5a89b0189f59696f1c0592f9a3/test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirementQueryTests.cs#L23) ran into the same problem.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidatorTests.cs
JimmyVo16: Does `SutProvider.Sut `work for an abstract class? I was under the assumption that you still have to create a concrete class for it. By the way, this also runs into the mocking [factory problem](https://github.com/bitwarden/server/pull/6139#discussion_r2298177744).

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
JimmyVo16: This is for `OrganizationPolicyDetails`, not `PolicyDetails`, but you’re right. I decided to create a custom attribute for it, similar to `PolicyDetails`, since there will likely be more tests for this resource in the future.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
eliykat: > The WhenPolicyEnabled part shouldn't really be there; it’s business logic, and now it’s duplicated in two places. If we modify the policy, this will need to be updated as well.

I don't understand - where else is it duplicated? I thought this was the source of truth for whether a user needs a default collection created when the policy is turned on.

The reason for the `WhenPolicyEnabled` wording is that this filters by `Confirmed` status, so it's not suitable for other cases such the confirm flow (because the user is not in a confirmed state yet). It is specifically for the "enable policy" flow.

The side-by-side example is a good one though. What about `GetDefaultCollectionRequestOnPolicyEnable`? I am mostly concerned about correct usage by the caller.

JimmyVo16: >I don't understand - where else is it duplicated? I thought this was the source of truth for whether a user needs a default collection created when the policy is turned on.

I was thinking that the logic for checking whether the policy is enabled or disabled is separate from this default collection flow; therefore, appending "PolicyEnabled" seems odd. I was also under the impression that these other attributes shouldn't apply if the policy is disabled, so is that extra postfix necessary?

Actually, if the policy is disabled, it won't be returned based on the [WHERE clause](https://github.com/bitwarden/server/blob/f4341b2f3bed714e651439f3191190e1aea07997/src/Sql/dbo/Stored%20Procedures/PolicyDetails_ReadByUserId.sql#L26) in the stored procedure, so `GetDefaultCollectionRequest` will always be `false` when the policy is disabled.

With that said, I think we're in the realm where many of our premises are built on assumptions now, since it seems like something that can easily be changed in the future, and there isn't any obvious issue with using any variation of "PolicyEnabled." We can move forward with it.

> The side-by-side example is a good one though. What about GetDefaultCollectionRequestOnPolicyEnable? I am mostly concerned about correct usage by the caller.

Yeah, `GetDefaultCollectionRequestOnPolicyEnable` is better, we can go with that.

JimmyVo16: Yeah this looks reasonable

```
        var userOrgIds = requirements
            .Select(requirement => requirement.GetDefaultCollectionRequestOnPolicyEnable(policyUpdate.OrganizationId))
            .Where(request => request.ShouldCreateDefaultCollection)
            .Select(request => request.OrganizationUserId);

```

eliykat: OK I think I see our miscommunication 😁 

> I was also under the impression that these other attributes shouldn't apply if the policy is disabled, so is that extra postfix necessary?
>
> Actually, if the policy is disabled, it won't be returned based on the [WHERE clause](https://github.com/bitwarden/server/blob/f4341b2f3bed714e651439f3191190e1aea07997/src/Sql/dbo/Stored%20Procedures/PolicyDetails_ReadByUserId.sql#L26) in the stored procedure, so GetDefaultCollectionRequest will always be false when the policy is disabled.

The postfix is meant to communicate: does this user need a default collection created _at the time that the policy is enabled by an administrator_. "OnPolicyEnabled" communicates the event the method relates to (when it is called), not the state of the policy.

This has proven just how ambiguous this naming is... but I still think something is better than nothing, with the xmldoc to provide the detail.

Your current changes are fine 👍

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
eliykat: Alas! Got by reference to my own code! That's fine 😁  I think you can work around it with the `SutProvider` builder pattern but it's really not that important.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidatorTests.cs
eliykat: SutProvider needs a concrete class, it wasn't about that. But yes, same issue with the factory, so that's OK.

eliykat (approved):
:shipit:

PR #6145, opened 2025-07-30
[PM-17562] Update logs to use custom categories

brant-livefront:
## 🎟️ Tracking

## 📔 Objective

@justindbaur suggested [in a comment](https://github.com/bitwarden/server/pull/6118#pullrequestreview-3068034131) on #6118 that we might want to use custom categories for our integration logging, since we'd be using the same base listener class in a number of places and the type-based `ILogger` would roll that up into one category. 

This PR adopts that suggestion and refactors to inject an `ILoggerFactory` into the listeners and then use the queue/subscription they are listening to as part of the categoryName. This gives us much finer grain awareness when errors do occur on a specific listener.

I also noticed one or two places we could use an extra test for logging that wasn't there previously, so I added tests for those.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusEventListenerService.cs
```diff
@@ -39,6 +40,12 @@ public override async Task StopAsync(CancellationToken cancellationToken)
         await base.StopAsync(cancellationToken);
     }
 
+    private static ILogger CreateLogger(ILoggerFactory loggerFactory, TConfiguration configuration)
+    {
+        return loggerFactory.CreateLogger(
+            categoryName: $"Bit.Core.Services.AzureServiceBusEventListenerService.{configuration.EventSubscriptionName}");
```
withinfocus, line 46: 💭 If we move this file, this could be missed. Can we reflect and get this? Is that performant?

src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusIntegrationListenerService.cs
```diff
@@ -14,16 +14,17 @@ public class AzureServiceBusIntegrationListenerService<TConfiguration> : Backgro
...
         TConfiguration configuration,
         IIntegrationHandler handler,
         IAzureServiceBusService serviceBusService,
-        ILogger<AzureServiceBusIntegrationListenerService<TConfiguration>> logger)
+        ILoggerFactory loggerFactory)
     {
         _handler = handler;
-        _logger = logger;
+        _logger = loggerFactory.CreateLogger(
+            categoryName: $"Bit.Core.Services.AzureServiceBusIntegrationListenerService.{configuration.IntegrationSubscriptionName}");
```
withinfocus, line 27: 💭 Same comment for all of these.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusEventListenerService.cs
brant-livefront: Updated to be more dynamic. For these ones in the static context, I still have to provide the name, but at least it gets the whole path dynamically.

Performance shouldn't be an issue as these are singletons and it should only happen once in the constructor. 👍

src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusEventListenerService.cs
```diff
@@ -39,6 +40,12 @@ public override async Task StopAsync(CancellationToken cancellationToken)
         await base.StopAsync(cancellationToken);
     }
 
+    private static ILogger CreateLogger(ILoggerFactory loggerFactory, TConfiguration configuration)
+    {
+        return loggerFactory.CreateLogger(
+            categoryName: $"{typeof(AzureServiceBusEventListenerService<>).FullName}.{configuration.EventQueueName}");
```
withinfocus: ❓ You explicitly went from the subscription to queue name for these right?

withinfocus (approved):

src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusEventListenerService.cs
brant-livefront: Great catch. I don't know how I messed that up, but this should _absolutely_ be the Subscription since this is ASB.

r-tome (dismissed):

justindbaur:
I don't recommend reflecting for the type name, neither way unfortunately makes a very log friendly name.

```c#
// Open generic
typeof(MyType<>).FullName == "MyNamespace.MyType`1"

// Closed generics
this.GetType().FullName == "Namespace.MyType`1[[Namespace.GenericType, AssemblyName, 1.0.0]]"
```

brant-livefront:
@justindbaur Good note. I see what you mean. I had set a breakpoint to see what it returned, but I had done that on the static one where it wasn't pulling from `this`. The `this` version returns
 
 ```
Bit.Core.Services.RabbitMqIntegrationListenerService`1[[Bit.Core.AdminConsole.Models.Data.EventIntegrations.SlackListenerConfiguration, Core, Version=2025.7.2.0, Culture=neutral, PublicKeyToken=null]].integration-slack-queue 
```
Yikes.

I could see us simply chopping off the string at the backtick? Seems like the simplest way to get the benefits of it being dynamic, without inheriting all that mess. Or just roll back and go with the manual version without introspection - maybe I could add a unit test that does introspect and verify that path matches?

@withinfocus thoughts?

withinfocus:
A test would be a nice way to catch that so sure. I dislike the hard-coding but I guess it's difficult to avoid and we want the control.

brant-livefront:
@justindbaur @withinfocus Latest commit reverts back to hardcoding, but adds explicit tests that DO introspect and verify it's correct.  👍  I also noticed a few of these didn't have `#nullable enable`, so I added that and fixed some of the errors it generated.

withinfocus (approved):

PR #6158, opened 2025-08-04
PM-23754 initial commit

prograhamming:
## 🎟️ Tracking

[PM-23754](https://bitwarden.atlassian.net/browse/PM-23754)
[PM-23757](https://bitwarden.atlassian.net/browse/PM-23757])

## 📔 Objective

Change OrganizationReport table

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
+++ b/src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
@@ -39,7 +39,6 @@ public class AddOrganizationReportCommand : IAddOrganizationReportCommand
         {
             OrganizationId = request.OrganizationId,
             ReportData = request.ReportData,
-            Date = request.Date == default ? DateTime.UtcNow : request.Date,
             CreationDate = DateTime.UtcNow,
         };
 
--- a/src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
+++ b/src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
@@ -7,5 +7,4 @@ public class AddOrganizationReportRequest
 {
     public Guid OrganizationId { get; set; }
     public string ReportData { get; set; }
-    public DateTime Date { get; set; }
 }
--- a/src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
+++ b/src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
@@ -1,19 +1,24 @@
 CREATE TABLE [dbo].[OrganizationReport] (
     [Id]                       UNIQUEIDENTIFIER NOT NULL,
     [OrganizationId]           UNIQUEIDENTIFIER NOT NULL,
-    [Date]                     DATETIME2 (7)    NOT NULL,
     [ReportData]               NVARCHAR(MAX)    NOT NULL,
     [CreationDate]             DATETIME2 (7)    NOT NULL,
     [ContentEncryptionKey]     VARCHAR(MAX)     NOT NULL,
+    [SummaryData]              NVARCHAR(MAX)    NULL,
+    [ApplicationData]          NVARCHAR(MAX)    NULL,
+    [RevisionDate]             DATETIME2 (7)    NOT NULL,
     CONSTRAINT [PK_OrganizationReport] PRIMARY KEY CLUSTERED ([Id] ASC),
     CONSTRAINT [FK_OrganizationReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id])
-);
+    );
 GO
+
 
 CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId]
-    ON [dbo].[OrganizationReport]([OrganizationId] ASC);
+   ON [dbo].[OrganizationReport]([OrganizationId] ASC);
 GO
 
-CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_Date]
-    ON [dbo].[OrganizationReport]([OrganizationId] ASC, [Date] DESC);
+
+CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_RevisionDate]
+   ON [dbo].[OrganizationReport]([OrganizationId] ASC, [RevisionDate] DESC);
 GO
+
--- a/util/Migrator/DbScripts/2025-08-04_00_AlterOrganizationReport.sql
+++ b/util/Migrator/DbScripts/2025-08-04_00_AlterOrganizationReport.sql
@@ -1,0 +1,79 @@
+IF EXISTS (
+SELECT * FROM sys.indexes WHERE name = 'IX_OrganizationReport_OrganizationId_Date'
+AND object_id = OBJECT_ID('dbo.OrganizationReport')
+)
+BEGIN
+DROP INDEX [IX_OrganizationReport_OrganizationId_Date] ON [dbo].[OrganizationReport];
+END
+GO
+
+IF COL_LENGTH('[dbo].[OrganizationReport]', 'Date') IS NOT NULL
+BEGIN
+ALTER TABLE [dbo].[OrganizationReport]
+DROP COLUMN [Date];
+END
+GO
+
+IF OBJECT_ID('dbo.OrganizationReport') IS NOT NULL
+BEGIN
+ALTER TABLE [dbo].[OrganizationReport]
+    ADD [SummaryData] NVARCHAR(MAX) NULL,
+    [ApplicationData] NVARCHAR(MAX) NULL,
+    [RevisionDate] DATETIME2 (7) NOT NULL;
+END
+GO
+
+IF NOT EXISTS (
+SELECT * FROM sys.indexes WHERE name = 'IX_OrganizationReport_OrganizationId_RevisionDate'
+AND object_id = OBJECT_ID('dbo.OrganizationReport')
+)
+BEGIN
+CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId_RevisionDate]
+ ON [dbo].[OrganizationReport]([OrganizationId] ASC, [RevisionDate] DESC);
+END
+GO
+
+CREATE OR ALTER VIEW [dbo].[OrganizationReportView]
+AS
+SELECT
+    *
+FROM
+    [dbo].[OrganizationReport];
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationReport_Create]
+   @Id UNIQUEIDENTIFIER OUTPUT,
+   @OrganizationId UNIQUEIDENTIFIER,
+   @ReportData NVARCHAR(MAX),
+   @CreationDate DATETIME2(7),
+   @ContentEncryptionKey VARCHAR(MAX),
+   @SummaryData NVARCHAR(MAX),
+   @ApplicationData NVARCHAR(MAX),
+   @RevisionDate DATETIME2(7)
+AS
+BEGIN
+   SET NOCOUNT ON;
+
+INSERT INTO [dbo].[OrganizationReport](
+    [Id],
+    [OrganizationId],
+    [ReportData],
+    [CreationDate],
+    [ContentEncryptionKey],
+    [SummaryData],
+    [ApplicationData],
+[RevisionDate]
+)
+VALUES (
+    @Id,
+    @OrganizationId,
+    @ReportData,
+    @CreationDate,
+    @ContentEncryptionKey,
+    @SummaryData,
+    @ApplicationData,
+    @RevisionDate
+    );
+END
+GO
+
--- a/src/Core/Dirt/Entities/OrganizationReport.cs
+++ b/src/Core/Dirt/Entities/OrganizationReport.cs
@@ -9,11 +9,14 @@ public class OrganizationReport : ITableObject<Guid>
 {
     public Guid Id { get; set; }
     public Guid OrganizationId { get; set; }
-    public DateTime Date { get; set; }
     public string ReportData { get; set; } = string.Empty;
     public DateTime CreationDate { get; set; } = DateTime.UtcNow;
 
     public string ContentEncryptionKey { get; set; } = string.Empty;
+
+    public string? SummaryData { get; set; } = null;
+    public string? ApplicationData { get; set; } = null;
+    public DateTime RevisionDate { get; set; } = DateTime.UtcNow;
 
     public void SetNewId()
     {
--- a/src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
+++ b/src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
@@ -38,7 +38,7 @@ public class OrganizationReportRepository :
             var dbContext = GetDatabaseContext(scope);
             var result = await dbContext.OrganizationReports
                 .Where(p => p.OrganizationId == organizationId)
-                .OrderByDescending(p => p.Date)
+                .OrderByDescending(p => p.RevisionDate)
                 .Take(1)
                 .FirstOrDefaultAsync();
 
--- a/src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_Create.sql
+++ b/src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_Create.sql
@@ -1,26 +1,35 @@
 CREATE PROCEDURE [dbo].[OrganizationReport_Create]
-    @Id UNIQUEIDENTIFIER OUTPUT,
-    @OrganizationId UNIQUEIDENTIFIER,
-    @Date DATETIME2(7),
-    @ReportData NVARCHAR(MAX),
-    @CreationDate DATETIME2(7),
-    @ContentEncryptionKey VARCHAR(MAX)
+   @Id UNIQUEIDENTIFIER OUTPUT,
+   @OrganizationId UNIQUEIDENTIFIER,
+   @ReportData NVARCHAR(MAX),
+   @CreationDate DATETIME2(7),
+   @ContentEncryptionKey VARCHAR(MAX),
+   @SummaryData NVARCHAR(MAX),
+   @ApplicationData NVARCHAR(MAX),
+   @RevisionDate DATETIME2(7)
 AS
-    SET NOCOUNT ON;
+BEGIN
+   SET NOCOUNT ON;
 
-    INSERT INTO [dbo].[OrganizationReport](
-        [Id],
-        [OrganizationId],
-        [Date],
-        [ReportData],
-        [CreationDate],
-        [ContentEncryptionKey]
-    )
-    VALUES (
-        @Id,
-        @OrganizationId,
-        @Date,
-        @ReportData,
-        @CreationDate,
-        @ContentEncryptionKey
+
+INSERT INTO [dbo].[OrganizationReport](
+    [Id],
+    [OrganizationId],
+    [ReportData],
+    [CreationDate],
+    [ContentEncryptionKey],
+    [SummaryData],
+    [ApplicationData],
+[RevisionDate]
+)
+VALUES (
+    @Id,
+    @OrganizationId,
+    @ReportData,
+    @CreationDate,
+    @ContentEncryptionKey,
+    @SummaryData,
+    @ApplicationData,
+    @RevisionDate
     );
+END
--- a/util/MySqlMigrations/Migrations/20250804191034_2025-08-04_00_AlterOrganizationReport.Designer.cs
+++ b/util/MySqlMigrations/Migrations/20250804191034_2025-08-04_00_AlterOrganizationReport.Designer.cs
@@ -3,63 +3,71 @@ using System;
 using Bit.Infrastructure.EntityFramework.Repositories;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 
 #nullable disable
 
-namespace Bit.SqliteMigrations.Migrations
+namespace Bit.MySqlMigrations.Migrations
 {
     [DbContext(typeof(DatabaseContext))]
-    partial class DatabaseContextModelSnapshot : ModelSnapshot
+    [Migration("20250804191034_2025-08-04_00_AlterOrganizationReport")]
+    partial class _20250804_00_AlterOrganizationReport
     {
-        protected override void BuildModel(ModelBuilder modelBuilder)
+        /// <inheritdoc />
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
-            modelBuilder.HasAnnotation("ProductVersion", "8.0.8");
+            modelBuilder
+                .HasAnnotation("ProductVersion", "8.0.8")
+                .HasAnnotation("Relational:MaxIdentifierLength", 64);
+
+            MySqlModelBuilderExtensions.AutoIncrementColumns(modelBuilder);
 
             modelBuilder.Entity("Bit.Core.Dirt.Reports.Models.Data.OrganizationMemberBaseDetail", b =>
                 {
                     b.Property<Guid>("CipherId")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("char(36)");
 
                     b.Property<Guid?>("CollectionId")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("char(36)");
 
                     b.Property<string>("CollectionName")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("longtext");
 
                     b.Property<string>("Email")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("longtext");
 
                     b.Property<Guid?>("GroupId")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("char(36)");
 
                     b.Property<string>("GroupName")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("longtext");
 
                     b.Property<bool?>("HidePasswords")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("tinyint(1)");
 
                     b.Property<bool?>("Manage")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("tinyint(1)");
 
                     b.Property<bool?>("ReadOnly")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("tinyint(1)");
 
                     b.Property<string>("ResetPasswordKey")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("longtext");
 
                     b.Property<string>("TwoFactorProviders")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("longtext");
 
                     b.Property<Guid?>("UserGuid")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("char(36)");
 
                     b.Property<string>("UserName")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("longtext");
 
                     b.Property<bool>("UsesKeyConnector")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("tinyint(1)");
 
                     b.ToTable("OrganizationMemberBaseDetails");
                 });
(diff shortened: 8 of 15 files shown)


util/Migrator/DbScripts/2025-08-04_00_AlterOrganizationReport.sql
```diff
@@ -0,0 +1,79 @@
...
+END
+GO
+
+IF COL_LENGTH('[dbo].[OrganizationReport]', 'Date') IS NOT NULL
+BEGIN
+ALTER TABLE [dbo].[OrganizationReport]
+DROP COLUMN [Date];
+END
+GO
+
```
rkac-bw: Normally ,for edd, the date is dropped from code reference and left on table first in case we have to roll back server code. Once the code roll back would not involve date column we would then drop it

prograhamming: Since this feature isn't live and users aren't using this table yet, does this still apply?

rkac-bw: It should be good if not in use

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,265 @@
...
+    public async Task<IActionResult> GetOrganizationReportDataAsync(Guid orgId, Guid reportId)
+    {
+        GuardOrganizationAccess(orgId);
+
+        var reportData = await _getOrganizationReportDataQuery.GetOrganizationReportDataAsync(orgId, reportId);
+        return Ok(reportData);
+    }
+
+    [HttpPatch("{orgId}/data/report/{reportId}")]
+    public async Task<IActionResult> UpdateOrganizationReportDataAsync(Guid orgId, Guid reportId, [FromBody] UpdateOrganizationReportDataRequest request)
```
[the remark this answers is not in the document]
prograhamming: Updated code to remove this issue.

src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
ttalty, line 1: ❌ This should also have `SummaryData` and `ReportData`

src/Core/Dirt/Reports/ReportFeatures/Requests/UpdateOrganizationReportRequest.cs
ttalty, line 1: ❌ Update should also take into account `SummaryData` and `ApplicationData`

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
ttalty, line 1: ❌ With the changes to the `AddOrganizationReportRequest` for adding `SummaryData` and `ApplicationData` this command needs to create the `OrganizationReport` object using those new fields.

src/Core/Dirt/Reports/ReportFeatures/UpdateOrganizationReportCommand.cs
ttalty, line 1: ❌ After adding the `SummaryData` and `ApplicationData` to the `UpdateOrganizationReportRequest` these new properties can be used to set them on the `OrganizationReport` object.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,260 @@
...
+        [FromQuery] DateTime startDate,
+        [FromQuery] DateTime endDate)
+    {
+        GuardOrganizationAccess(orgId);
+
+        var summaryDataList = await _getOrganizationReportSummaryDataByDateRangeQuery
+            .GetOrganizationReportSummaryDataByDateRangeAsync(orgId, reportId, startDate, endDate);
+
+        return Ok(summaryDataList);
+    }
```
ttalty: ❌ The summary data by date range lookup should not be limited to the reportId. There will only ever be one summary for any report. 

The purpose of this endpoint is to gather the summaries from all reports ran over a specified date range.

ttalty (changes requested):
❌ Two changes

1 - Summaries by date range should not be limited to a reportId
2 - The POST for creating a whole organization report only saves report data

As noted in other comments. The controller methods are split into different use cases: 
- Whole record: `ReportData` `SummaryData` `ApplicationData`
- Summary data only: `SummaryData`
- Application data only: `ApplicationData'
- Report data only: `ReportData`

Right now the POST and PATCH for the whole report only creates/updates the record with `ReportData` it should create/update the record for all 3 columns. 

When a new report is ran, a new copy of summary and application data is created also. Having one endpoint to create and then using the data specific endpoints to modify.

src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
prograhamming: updated.

src/Core/Dirt/Reports/ReportFeatures/Requests/UpdateOrganizationReportRequest.cs
prograhamming: updated.

src/Core/Dirt/Reports/ReportFeatures/UpdateOrganizationReportCommand.cs
prograhamming: updated.

src/Core/Dirt/Reports/ReportFeatures/AddOrganizationReportCommand.cs
prograhamming: updated

src/Api/Dirt/Controllers/OrganizationReportsController.cs
prograhamming: updated.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetLatestByOrganizationId.sql
ttalty, line 1: ❌ Shouldn't this use `SELECT TOP 1` 

This currently returns a "Sequence contains more than one element" 

The `OrganizationReportRepository.GetLatestByOrganizationIdAsync` uses this query. A single `OrganizationReport` is the expected return type. Yet this query will return all records just in descending order based on `RevisionDate`

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,261 @@
...
+            throw new BadRequestException("Organization ID in the request body must match the route parameter");
+        }
+
+        var summaryDataList = await _getOrganizationReportSummaryDataByDateRangeQuery
+            .GetOrganizationReportSummaryDataByDateRangeAsync(request.OrganizationId, request.StartDate, request.EndDate);
+
+        return Ok(summaryDataList);
+    }
+
+    [HttpGet("{orgId}/data/summary/{reportId}")]
```
ttalty: ❌  `[HttpPost("{orgId}/data/summary/date-range")]` is being routed to `[HttpGet("{orgId}/data/summary/{reportId}")]` and "date-range" is being taken as the reportId. 

Error:
```
{
    "message": "The model state is invalid.",
    "validationErrors": {
        "reportId": [
            "The value 'date-range' is not valid."
        ]
    },
    "exceptionMessage": null,
    "exceptionStackTrace": null,
    "innerExceptionMessage": null,
    "object": "error" 
}
```

EDIT: We should use query string params on the `{orgId}/data/summary` route for the date range and return an error when there is at least no `fromDate` value. If there is no `toDate` then it can be assumed it is the current date.

src/Infrastructure.Dapper/Dirt/OrganizationReportRepository.cs
```diff
@@ -32,17 +33,156 @@ public async Task<ICollection<OrganizationReport>> GetByOrganizationIdAsync(Guid
...
+            await connection.ExecuteAsync(
+                $"[{Schema}].[OrganizationReport_UpdateSummaryData]",
+                parameters,
+                commandType: CommandType.StoredProcedure);
+
+            // Return the updated report
+            return await connection.QuerySingleOrDefaultAsync<OrganizationReport>(
+                $"[{Schema}].[OrganizationReport_ReadById]",
+                new { Id = reportId },
+                commandType: CommandType.StoredProcedure);
```
ttalty: ❌ The parameters should set the OrganizationId. 

`OrganizationReport_UpdateSummaryData` sql expects `@OrganizationId` as a parameter

src/Infrastructure.Dapper/Dirt/OrganizationReportRepository.cs
```diff
@@ -32,17 +33,156 @@ public async Task<ICollection<OrganizationReport>> GetByOrganizationIdAsync(Guid
...
+                Id = reportId,
+                ApplicationData = applicationData,
+                RevisionDate = DateTime.UtcNow
+            };
+
+            await connection.ExecuteAsync(
+                $"[{Schema}].[OrganizationReport_UpdateApplicationData]",
+                parameters,
+                commandType: CommandType.StoredProcedure);
+
```
ttalty, line 168: ❌ `OrganizationReport_UpdateApplicationData` stored procedure takes in only `@Id` `@OrganizationId` and `@ApplicationData` 

Parameters here are being setup with an added `RevisionDate` which is not needed since the stored procedure sets the revision date to `GETUTCDATE()` already

src/Infrastructure.Dapper/Dirt/OrganizationReportRepository.cs
```diff
@@ -32,17 +33,156 @@ public async Task<ICollection<OrganizationReport>> GetByOrganizationIdAsync(Guid
...
+            await connection.ExecuteAsync(
+                $"[{Schema}].[OrganizationReport_UpdateReportData]",
+                parameters,
+                commandType: CommandType.StoredProcedure);
+
+            // Return the updated report
+            return await connection.QuerySingleOrDefaultAsync<OrganizationReport>(
+                $"[{Schema}].[OrganizationReport_ReadById]",
+                new { Id = reportId },
+                commandType: CommandType.StoredProcedure);
```
ttalty, line 135: ❌ `OrganizationReport_UpdateReportData` parameters are `@Id` `@OrganizationId` `@ReportData` here the parameters are adding the `RevisionDate` which is not needed since the stored procedure sets the date already using `GETUTCDATE()`

ttalty (changes requested):
❌ Changes noted in comments

src/Infrastructure.Dapper/Dirt/OrganizationReportRepository.cs
```diff
@@ -32,17 +33,156 @@ public async Task<ICollection<OrganizationReport>> GetByOrganizationIdAsync(Guid
...
+        {
+            var parameters = new
+            {
+                OrganizationId = organizationId,
+                StartDate = startDate,
+                EndDate = endDate
+            };
+
+            var results = await connection.QueryAsync<OrganizationReportSummaryDataResponse>(
+                $"[{Schema}].[OrganizationReport_GetSummaryDataByDateRange]",
```
ttalty: ❌ Looks like the stored procedure name is actually `OrganizationReport_GetSummariesByDateRange'

Instead of changing the string here I think the sproc itself should change. The naming convention for application, summary, and report data has been "Get{dataType}DataBy{id or dateRange}" 

I think the sproc should be named `OrganizationReport_GetSummaryDataByDateRange`

ttalty (changes requested):

src/Core/Dirt/Reports/ReportFeatures/UpdateOrganizationReportCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+            var updatedRecord = new OrganizationReport()
+            {
+                Id = existingReport.Id,
+                OrganizationId = request.OrganizationId,
+                CreationDate = existingReport.CreationDate,
+                ContentEncryptionKey = request.ContentEncryptionKey,
+                SummaryData = request.SummaryData,
+                ApplicationData = request.ApplicationData,
+                RevisionDate = DateTime.UtcNow
+            };
```
rkac-bw: do we need the report updated with this updatedRecord var?

src/Core/Dirt/Reports/ReportFeatures/UpdateOrganizationReportCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            await _organizationReportRepo.ReplaceAsync(updatedRecord);
+
+            _logger.LogInformation(Constants.BypassFiltersEventId, "Successfully updated organization report {reportId} for organization {organizationId}",
+                request.ReportId, request.OrganizationId);
+
+            return await _organizationReportRepo.GetLatestByOrganizationIdAsync(request.ReportId);
+        }
```
rkac-bw, line 75: passing ReportId should it be  OrganizationId as per _organizationReportRepo.GetLatestByOrganizationIdAsync?

src/Api/Dirt/Controllers/OrganizationReportsController.cs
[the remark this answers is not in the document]
prograhamming: I am going to ignore this warning because when I click on 'Show more details' it references a line number that has nothing to do with the code above:

```
Method at line 1233 of /src/Api/Vault/Controllers/CiphersController.cs gets a parameter from a user request from id. This parameter value flows through the code and is eventually used to access application state altering functionality. This may enable Cross-Site Request Forgery (CSRF).
```

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,265 @@
...
+            return Ok(applicationData);
+        }
+        catch (Exception ex) when (!(ex is BadRequestException || ex is NotFoundException))
+        {
+            throw;
+        }
+    }
+
+    [HttpPatch("{orgId}/data/application")]
+    public async Task<IActionResult> UpdateOrganizationReportApplicationDataAsync(Guid orgId, [FromBody] UpdateOrganizationReportApplicationDataRequest request)
```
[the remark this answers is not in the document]
prograhamming: Same thing as the last warning message.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+        var report = await _getOrganizationReportQuery.GetOrganizationReportAsync(reportId);
+
+        if (report == null)
+        {
+            throw new NotFoundException("Report not found for the specified organization.");
+        }
+
+        if (report.OrganizationId != orgId)
+        {
+            throw new BadRequestException("Organization ID in the request body must match the route parameter");
```
ttalty: ⛏️ Maybe this should say "Invalid Report ID" also here. The current wording says the org id in the request body does not match the route. Yet the condition here checks against the returned report org id to the route org id.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+        var reportData = await _getOrganizationReportDataQuery.GetOrganizationReportDataAsync(orgId, reportId);
+
+        if (reportData == null)
+        {
+            throw new NotFoundException("Organization report data not found.");
+        }
+
+        if (reportData.OrganizationId != orgId)
+        {
+            throw new BadRequestException("Organization ID in the request body must match the route parameter");
```
ttalty: ⛏️ This isn't a mismatch of the request organization Id. It's a mismatch that the returned report's organizationId does not match the orgId in the route param. Maybe this should say "Invalid Report ID"

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+
+
+        if (summaryData == null)
+        {
+            throw new NotFoundException("Report not found for the specified organization.");
+        }
+
+        if (summaryData.OrganizationId != orgId)
+        {
+            throw new BadRequestException("Organization ID in the request body must match the route parameter");
```
ttalty: ⛏️ Maybe this should say "Invalid Report ID" also here. The current wording says the org id in the request body does not match the route. Yet the condition here checks against the returned summary data's org id to the route org id.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+            var applicationData = await _getOrganizationReportApplicationDataQuery.GetOrganizationReportApplicationDataAsync(orgId, reportId);
+
+            if (applicationData == null)
+            {
+                throw new NotFoundException("Organization report application data not found.");
+            }
+
+            if (applicationData.OrganizationId != orgId)
+            {
+                throw new BadRequestException("Organization ID in the request body must match the route parameter");
```
ttalty: ⛏️ Maybe this should say "Invalid Report ID" also here. The current wording says the org id in the request body does not match the route. Yet the condition here checks against the returned application data's org id to the route org id.

ttalty (approved):
Looks good and I tested the endpoints. Just waiting on @rkac-bw to take a closer look at any repository/integration areas.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+        }
+        catch (Exception ex) when (!(ex is BadRequestException || ex is NotFoundException))
+        {
+            throw;
+        }
+    }
+
+    #endregion
+
+    private async Task<bool> GuardOrganizationAccessAsync(Guid organizationId)
```
withinfocus: 🎨 I would name and model this after other check methods, like what we have in Secrets Manager for instance.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+    #region Whole OrganizationReport Endpoints
+
+    [HttpGet("{orgId}/latest")]
+    public async Task<IActionResult> GetLatestOrganizationReportAsync(Guid orgId)
+    {
+        var guardResult = await GuardOrganizationAccessAsync(orgId);
+
+        if (guardResult)
+        {
+            throw new NotFoundException("Access denied.");
```
withinfocus: ❌ This is a mix of response codes, a hybrid 401 / 403 / 404. Let's stick with standard rejections around either authentication or authorization.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,333 @@
...
+
+
+        if (summaryData == null)
+        {
+            throw new NotFoundException("Report not found for the specified organization.");
+        }
+
+        if (summaryData.OrganizationId != orgId)
+        {
+            throw new BadRequestException("Invalid report id");
```
withinfocus: ⛏️ Ensure consistent usage of "ID" everywhere.

src/Core/Dirt/Models/Data/OrganizationReportApplicationDataResponse.cs
```diff
@@ -0,0 +1,10 @@
+﻿#nullable enable
+
+namespace Bit.Core.Dirt.Models.Data;
+
+public class OrganizationReportApplicationDataResponse
+{
+    public Guid Id { get; set; }
+    public Guid OrganizationId { get; set; }
+    public string ApplicationData { get; set; } = string.Empty;
```
withinfocus: ❓ Why these sorts of defaults?

src/Core/Dirt/Reports/ReportFeatures/Requests/GetOrganizationReportSummaryDataByDateRangeRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿// FIXME: Update this file to be null safe and then delete the line below
```
withinfocus, line 1: ♻️ You're introducing new code, so can nullable be dealt with?

src/Core/Dirt/Reports/ReportFeatures/UpdateOrganizationReportApplicationDataCommand.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        catch (Exception ex) when (!(ex is BadRequestException || ex is NotFoundException))
+        {
+            _logger.LogError(ex, "Error updating organization report application data {reportId} for organization {organizationId}",
+                request.Id, request.OrganizationId);
+            throw;
+        }
+    }
+
+    private async Task<(bool IsValid, string errorMessage)> ValidateRequestAsync(UpdateOrganizationReportApplicationDataRequest request)
```
withinfocus: ⛏️ Your casing is inconsistent with the tuple and it can be CamelCase.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
```diff
@@ -0,0 +1,16 @@
...
+    @Id UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT
+        [Id],
+        [OrganizationId],
+        [ApplicationData]
+    FROM [dbo].[OrganizationReport]
```
withinfocus: ❌ All selects should use views, and it seems like you are in need of a custom details view for just application data? I don't see why you'd need to return what you have just used as parameters either.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
```diff
@@ -0,0 +1,16 @@
...
+    WHERE [OrganizationId] = @OrganizationId
+    AND [Id] = @Id
```
withinfocus: ❓ I don't understand this query -- why would you need to check the org ID if you have the ID? This is a primary key lookup.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
withinfocus: This comment applies in the same way to all the other custom selects I see below.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
withinfocus: This question is repeated for all similar usages below.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
```diff
@@ -1,19 +1,24 @@
...
     CONSTRAINT [PK_OrganizationReport] PRIMARY KEY CLUSTERED ([Id] ASC),
     CONSTRAINT [FK_OrganizationReport_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id])
-);
+    );
 GO
 
+
 CREATE NONCLUSTERED INDEX [IX_OrganizationReport_OrganizationId]
-    ON [dbo].[OrganizationReport]([OrganizationId] ASC);
+   ON [dbo].[OrganizationReport]([OrganizationId] ASC);
```
withinfocus:
```suggestion
   ON [dbo].[OrganizationReport] ([OrganizationId] ASC);
```

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
prograhamming: I don't think we initially would need these, but I thought it would be a good to have them in case we onboard new organization that large (100k+) and JSON for summary, report, and application could be quite large.

src/Core/Dirt/Reports/ReportFeatures/UpdateOrganizationReportApplicationDataCommand.cs
prograhamming: fixed.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
prograhamming: updated.

src/Core/Dirt/Models/Data/OrganizationReportApplicationDataResponse.cs
prograhamming: removed.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
prograhamming: Updated to mirror Secrets Manager.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
prograhamming: changed to mirror Secrets Manager and just throw `NotFoundException`

src/Core/Dirt/Reports/ReportFeatures/Requests/GetOrganizationReportSummaryDataByDateRangeRequest.cs
prograhamming: Updated the code and removed this.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
prograhamming: I added for completeness sake, but it's a good point. It doesn't matter for the queries and we don't need the `OrganizationId`

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -294,6 +294,11 @@ public async Task<IActionResult> UpdateOrganizationReportApplicationDataAsync(Gu
 
             var updatedReport = await _updateOrganizationReportApplicationDataCommand.UpdateOrganizationReportApplicationDataAsync(request);
 
+            if (updatedReport.Id != request.Id)
+            {
+                throw new BadRequestException("Report ID in the request body must match the route parameter");
+            }
+
```
ttalty: ❌ This should match what `UpdateOrganizationReportSummaryAsync` does. This method does not implement the route param for `reportId`

This error condition will never be triggered. The update is based on the report id in the request. Therefore the updated object id will always match the request id.

Add `reportId` as an input param and then add:

```
        if (request.ReportId != reportId)
        {
            throw new BadRequestException("Report ID in the request body must match the route parameter");
        }
```

Remove lines 297 -301.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -55,11 +55,9 @@ IUpdateOrganizationReportApplicationDataCommand updateOrganizationReportApplicat
     [HttpGet("{orgId}/latest")]
     public async Task<IActionResult> GetLatestOrganizationReportAsync(Guid orgId)
     {
-        var guardResult = await GuardOrganizationAccessAsync(orgId);
-
-        if (guardResult)
+        if (!await _currentContext.AccessReports(orgId))
```
withinfocus: ❓ Do all these APIs need this? Can't this become an attribute on the controller?

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
withinfocus: Coming back to this one as well -- if you're selecting a single report I would still think you select the whole row or a single column -- one or the other. I'd like to know more about the performance characteristics and why you don't need all of them at once.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
prograhamming: Fixed this line of code.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
prograhamming: I'm just following a similar pattern that other sections of the API follow. I'm happy to write a new user story to make this change. Considering our deadline for this feature and the big changes to the table design for this feature, we need to ship something.

withinfocus:
I still have https://github.com/bitwarden/server/pull/6158#discussion_r2319666181 and there's a conflict.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
prograhamming: This comment has been addressed. Current code:
```
CREATE PROCEDURE [dbo].[OrganizationReport_GetApplicationDataById]
    @Id UNIQUEIDENTIFIER
AS
BEGIN
    SET NOCOUNT ON

    SELECT
        [Id],
        [OrganizationId],
        [ApplicationData]
    FROM [dbo].[OrganizationReportView]
    WHERE [Id] = @Id
END
```

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetApplicationDataById.sql
prograhamming: My worry with a large organization (100k+) is the JSON we are storing could be quite large to be sending/receiving the entire report around. This potentially cause us to hit the request/response size limit. This may be a over optimization, but figured having it on handle would save us time in case we get an issue. 

The Id's are mainly used to check any malicious intent with a user being a part of two organizations. Let's say they get removed from Organization A and they try to use a report Id from Organization A but pass in Organization B's org Id with Organization A's report Id in the route params.

withinfocus: I said something similar 6-12 months ago and to not use the relational DB in that case. The usage of such a thing is outside this PR which is why I am asking how these would be used. I don't think it's advisable long-term to make this small optimization when if it's really true that 100k+ users will be in a report that we skip past the different storage mechanism.

Technically-speaking though and per previous comments of mine, we return all the columns of a view with select procs by design, with the exception being a single column. You are returning additional columns which suggests the need for details views in order to maintain the interface to the database as well as performance -- if you were to map these procs to the traditional domain object representing the table you will have a partially-hydrated object and that can lead to logic issues in a variety of places.

I don't understand "malicious intent" with a response from the database. App logic should handle that, and you shouldn't be querying the database at all without authorization.

withinfocus (approved):

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -0,0 +1,297 @@
...
+        _getOrganizationReportSummaryDataByDateRangeQuery = getOrganizationReportSummaryDataByDateRangeQuery;
+        _getOrganizationReportDataQuery = getOrganizationReportDataQuery;
+        _updateOrganizationReportDataCommand = updateOrganizationReportDataCommand;
+        _getOrganizationReportApplicationDataQuery = getOrganizationReportApplicationDataQuery;
+        _updateOrganizationReportApplicationDataCommand = updateOrganizationReportApplicationDataCommand;
+    }
+
+    #region Whole OrganizationReport Endpoints
+
+    [HttpGet("{orgId}/latest")]
```
withinfocus: 🎨 I left this as a commit comment and did that get lost? It was about using `organizationId` here to be consistent with other controllers, unless I am mistaken.

prograhamming: I believe so.  However, based on your other comment I added the change throughout the endpoints, commands/queries, and repositories just now. Running unit tests and will push up in about 10 minutes.

withinfocus: I don't see a push.

prograhamming: Remote branch had something on it, just merged and pushed up.

PR #6163, opened 2025-08-05
[PM-24278] poc

JimmyVo16:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/AdminConsole/Controllers/PoliciesController.cs
+++ b/src/Api/AdminConsole/Controllers/PoliciesController.cs
@@ -9,6 +9,7 @@ using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.Enums;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationDomains.Interfaces;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
 using Bit.Core.AdminConsole.Repositories;
 using Bit.Core.Auth.Models.Business.Tokenables;
 using Bit.Core.Context;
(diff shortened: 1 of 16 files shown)


src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -208,8 +209,34 @@ public async Task<PolicyResponseModel> Put(Guid orgId, PolicyType type, [FromBod
...
+    {
+        var savePolicyRequest = new SavePolicyRequest(
+            new PolicyUpdate
+            {
+                OrganizationId = Guid.Parse("aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"),
+                Type = PolicyType.OrganizationDataOwnership,
+                Enabled = true,
+                Data = ""
+            },
+            new OrganizationModelOwnershipPolicyModel("Default Collection Name", executeSideEffect: true),
```
JimmyVo16, line 236: @eliykat 
So at the controller level, we can detect the behavior of the validator. 

Of course, this is just a mock, so we should move this logic to where it makes more sense. The endpoint doesn’t need to know about this.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/SavePolicyCommand.cs
```diff
@@ -78,6 +78,30 @@ public async Task<Policy> SaveAsync(PolicyUpdate policyUpdate)
...
+                     {
+                         OrganizationId = policyUpdate.OrganizationId,
+                         Type = policyUpdate.Type,
+                         CreationDate = _timeProvider.GetUtcNow().UtcDateTime
+                     };
+
+        if (_policyValidators.TryGetValue(policyUpdate.Type, out var validator))
+        {
+
+            await validator.ProtoTypeOnSaveSideEffectsAsync(policyRequest, policy);
```
JimmyVo16, line 96: @eliykat 

A lot of this is boilerplate, but as you can see, the SaveCommand doesn’t have to be aware of any data associated with the policy metadata; it can just pass it along. In the event that we need to change something for one policy, the SaveCommand and the other policies wouldn’t be affected.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -26,6 +26,33 @@ public class OrganizationDataOwnershipPolicyValidator(
...
+        // Here we will have command data associated with this particular policy validator.
+        if (metadata!.ExecuteSideEffect)
+        {
+
+        }
+
+        return Task.CompletedTask;
+    }
+
+    public static OrganizationModelOwnershipPolicyModel? MapToOrganizationModelOwnershipPolicyModel(IPolicyMetadataModel model)
```
JimmyVo16, line 43: @eliykat Here we need a mapper to turn the generic metadata object into something the validator can use. We can easily centralize this in a switch statement or something similar.

eliykat: I'd like to see how this is actually handled in the flow from client -> controller -> command -> validator.

Given that this model depends on the policy type, the controller won't know how to instantiate it. One example of how to handle this is in the `Policy.data` property: it's a json blob, it's maintained as a `string` through the api layer, and then the policy specific code uses `CoreHelpers.LoadClassFromJsonData` to instantiate it into an object when needed.

You could follow a similar pattern here, maintaining it as a string through to the validator, then the validator creates the actual object to deal with it. However, that does tightly couple it to the request sent by the client, whereas the `executeSideEffect` property is set by the controller layer in your current code.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/SavePolicyCommand.cs
eliykat: Looks good and in keeping with the design of the validators.

JimmyVo16:
Just a POC

PR #6170, opened 2025-08-06
[PM-23845] Update cache service to handle concurrency

JimmyVo16:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Services/IApplicationCacheService.cs
+++ b/src/Core/Services/IApplicationCacheService.cs
@@ -1,4 +1,5 @@
-﻿using Bit.Core.AdminConsole.Entities;
+﻿using System.Collections.Concurrent;
+using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.Entities.Provider;
 using Bit.Core.AdminConsole.Models.Data.Provider;
 using Bit.Core.Models.Data.Organizations;
@@ -7,11 +8,11 @@ namespace Bit.Core.Services;
 
 public interface IApplicationCacheService
 {
-    Task<IDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync();
+    Task<ConcurrentDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync();
 #nullable enable
     Task<OrganizationAbility?> GetOrganizationAbilityAsync(Guid orgId);
 #nullable disable
-    Task<IDictionary<Guid, ProviderAbility>> GetProviderAbilitiesAsync();
+    Task<ConcurrentDictionary<Guid, ProviderAbility>> GetProviderAbilitiesAsync();
     Task UpsertOrganizationAbilityAsync(Organization organization);
     Task UpsertProviderAbilityAsync(Provider provider);
     Task DeleteOrganizationAbilityAsync(Guid organizationId);
--- a/src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
+++ b/src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
@@ -1,6 +1,7 @@
 ﻿// FIXME: Update this file to be null safe and then delete the line below
 #nullable disable
 
+using System.Collections.Concurrent;
 using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.Entities.Provider;
 using Bit.Core.AdminConsole.Models.Data.Provider;
@@ -15,10 +16,13 @@ public class InMemoryApplicationCacheService : IApplicationCacheService
     private readonly IOrganizationRepository _organizationRepository;
     private readonly IProviderRepository _providerRepository;
     private DateTime _lastOrgAbilityRefresh = DateTime.MinValue;
-    private IDictionary<Guid, OrganizationAbility> _orgAbilities;
+    private DateTime _lastProviderAbilityRefresh = DateTime.MinValue;
+    private ConcurrentDictionary<Guid, OrganizationAbility> _orgAbilities;
     private TimeSpan _orgAbilitiesRefreshInterval = TimeSpan.FromMinutes(10);
+    private readonly SemaphoreSlim _orgInitLock = new(1, 1);
+    private readonly SemaphoreSlim _providerInitLock = new(1, 1);
 
-    private IDictionary<Guid, ProviderAbility> _providerAbilities;
+    private ConcurrentDictionary<Guid, ProviderAbility> _providerAbilities;
 
     public InMemoryApplicationCacheService(
         IOrganizationRepository organizationRepository, IProviderRepository providerRepository)
@@ -27,7 +31,7 @@ public class InMemoryApplicationCacheService : IApplicationCacheService
         _providerRepository = providerRepository;
     }
 
-    public virtual async Task<IDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync()
+    public virtual async Task<ConcurrentDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync()
     {
         await InitOrganizationAbilitiesAsync();
         return _orgAbilities;
@@ -42,7 +46,7 @@ public class InMemoryApplicationCacheService : IApplicationCacheService
     }
 #nullable disable
 
-    public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbilitiesAsync()
+    public virtual async Task<ConcurrentDictionary<Guid, ProviderAbility>> GetProviderAbilitiesAsync()
     {
         await InitProviderAbilitiesAsync();
         return _providerAbilities;
@@ -66,37 +70,69 @@ public class InMemoryApplicationCacheService : IApplicationCacheService
 
     public virtual Task DeleteOrganizationAbilityAsync(Guid organizationId)
     {
-        _orgAbilities?.Remove(organizationId);
-
-        return Task.FromResult(0);
+        _orgAbilities?.TryRemove(organizationId, out _);
+        return Task.CompletedTask;
     }
 
     public virtual Task DeleteProviderAbilityAsync(Guid providerId)
     {
-        _providerAbilities?.Remove(providerId);
-
-        return Task.FromResult(0);
+        _providerAbilities?.TryRemove(providerId, out _);
+        return Task.CompletedTask;
     }
 
     private async Task InitOrganizationAbilitiesAsync()
     {
         var now = DateTime.UtcNow;
-        if (_orgAbilities == null || (now - _lastOrgAbilityRefresh) > _orgAbilitiesRefreshInterval)
+
+        if (_orgAbilities != null && (now - _lastOrgAbilityRefresh) <= _orgAbilitiesRefreshInterval)
         {
-            var abilities = await _organizationRepository.GetManyAbilitiesAsync();
-            _orgAbilities = abilities.ToDictionary(a => a.Id);
-            _lastOrgAbilityRefresh = now;
+            return;
+        }
+
+        await _orgInitLock.WaitAsync();
+        try
+        {
+            // Double-check after acquiring lock to avoid redundant refresh
+            now = DateTime.UtcNow;
+            if (_orgAbilities == null || (now - _lastOrgAbilityRefresh) > _orgAbilitiesRefreshInterval)
+            {
+                var abilities = await _organizationRepository.GetManyAbilitiesAsync();
+                _orgAbilities = new ConcurrentDictionary<Guid, OrganizationAbility>(
+                    abilities.ToDictionary(a => a.Id));
+                _lastOrgAbilityRefresh = now;
+            }
+        }
+        finally
+        {
+            _orgInitLock.Release();
         }
     }
 
     private async Task InitProviderAbilitiesAsync()
     {
         var now = DateTime.UtcNow;
-        if (_providerAbilities == null || (now - _lastOrgAbilityRefresh) > _orgAbilitiesRefreshInterval)
+
+        if (_providerAbilities != null && (now - _lastProviderAbilityRefresh) <= _orgAbilitiesRefreshInterval)
         {
-            var abilities = await _providerRepository.GetManyAbilitiesAsync();
-            _providerAbilities = abilities.ToDictionary(a => a.Id);
-            _lastOrgAbilityRefresh = now;
+            return;
+        }
+
+        await _providerInitLock.WaitAsync();
+        try
+        {
+            // Double-check after acquiring lock to avoid redundant refresh
+            now = DateTime.UtcNow;
+            if (_providerAbilities == null || (now - _lastProviderAbilityRefresh) > _orgAbilitiesRefreshInterval)
+            {
+                var abilities = await _providerRepository.GetManyAbilitiesAsync();
+                _providerAbilities = new ConcurrentDictionary<Guid, ProviderAbility>(
+                    abilities.ToDictionary(a => a.Id));
+                _lastProviderAbilityRefresh = now;
+            }
+        }
+        finally
+        {
+            _providerInitLock.Release();
         }
     }
 }
--- a/test/Core.Test/Helpers/OrganizationAbilityBuilder.cs
+++ b/test/Core.Test/Helpers/OrganizationAbilityBuilder.cs
@@ -1,0 +1,93 @@
+﻿using System.Collections.Concurrent;
+using Bit.Core.AdminConsole.Entities;
+using Bit.Core.Entities;
+using Bit.Core.Models.Data.Organizations;
+using Bit.Core.Models.Data.Organizations.OrganizationUsers;
+using Bit.Core.Vault.Models.Data;
+
+namespace Bit.Core.Test.Helpers;
+
+public static class OrganizationAbilityBuilder
+{
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(Guid organizationId) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    organizationId,
+                    new OrganizationAbility
+                    {
+                        Id = organizationId,
+                        LimitItemDeletion = true
+                    })
+            });
+
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(Organization organization) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    organization.Id,
+                    new OrganizationAbility(organization))
+            });
+
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(OrganizationUser orgUser) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    orgUser.OrganizationId,
+                    new OrganizationAbility
+                    {
+                        Id = orgUser.OrganizationId,
+                        UseEvents = true,
+                        Enabled = true
+                    })
+            });
+
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(Group group) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    group.OrganizationId,
+                    new OrganizationAbility
+                    {
+                        Id = group.OrganizationId,
+                        UseEvents = true,
+                        Enabled = true
+                    })
+            });
+
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(OrganizationUserOrganizationDetails orgUser, Organization organization) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    orgUser.OrganizationId,
+                    new OrganizationAbility(organization))
+            });
+
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(CipherDetails cipherDetails) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    cipherDetails.OrganizationId!.Value,
+                    new OrganizationAbility())
+            });
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(ICollection<Collection> collections) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    collections.First().OrganizationId,
+                    new OrganizationAbility
+                    {
+                        LimitCollectionCreation = true,
+                        LimitCollectionDeletion = true,
+                        AllowAdminAccessToAllCollectionItems = true
+                    })
+            });
+
+}
--- a/test/Identity.Test/Helpers/OrganizationAbilityBuilder.cs
+++ b/test/Identity.Test/Helpers/OrganizationAbilityBuilder.cs
@@ -1,0 +1,18 @@
+﻿using System.Collections.Concurrent;
+using Bit.Core.AdminConsole.Entities;
+using Bit.Core.Models.Data.Organizations;
+using Bit.Core.Models.Data.Organizations.OrganizationUsers;
+
+namespace Bit.Identity.Test.Helpers;
+
+public static class OrganizationAbilityBuilder
+{
+    public static ConcurrentDictionary<Guid, OrganizationAbility> BuildConcurrentDictionary(OrganizationUserOrganizationDetails orgUser, Organization organization) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, OrganizationAbility>(
+                    orgUser.OrganizationId,
+                    new OrganizationAbility(organization))
+            });
+}
--- a/test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
+++ b/test/Api.Test/Vault/AuthorizationHandlers/BulkCollectionAuthorizationHandlerTests.cs
@@ -8,6 +8,7 @@ using Bit.Core.Models.Data;
 using Bit.Core.Models.Data.Organizations;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
+using Bit.Core.Test.Helpers;
 using Bit.Core.Test.Vault.AutoFixture;
 using Bit.Test.Common.AutoFixture;
 using Bit.Test.Common.AutoFixture.Attributes;
@@ -1265,17 +1266,7 @@ public class BulkCollectionAuthorizationHandlerTests
         var actingUserId = Guid.NewGuid();
         var orgId = collections.First().OrganizationId;
 
-        var organizationAbilities = new Dictionary<Guid, OrganizationAbility>
-        {
-            { collections.First().OrganizationId,
-                new OrganizationAbility
-                {
-                    LimitCollectionCreation = true,
-                    LimitCollectionDeletion = true,
-                    AllowAdminAccessToAllCollectionItems = true
-                }
-            }
-        };
+        var organizationAbilities = OrganizationAbilityBuilder.BuildConcurrentDictionary(collections);
 
         var operationsToTest = new[]
         {
--- a/test/Api.Test/Vault/Controllers/CiphersControllerTests.cs
+++ b/test/Api.Test/Vault/Controllers/CiphersControllerTests.cs
@@ -12,6 +12,7 @@ using Bit.Core.Exceptions;
 using Bit.Core.Models.Data.Organizations;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
+using Bit.Core.Test.Helpers;
 using Bit.Core.Vault.Entities;
 using Bit.Core.Vault.Models.Data;
 using Bit.Core.Vault.Repositories;
@@ -79,7 +80,7 @@ public class CiphersControllerTests
         sutProvider.GetDependency<ICipherRepository>().GetByIdAsync(id, userId).ReturnsForAnyArgs(cipherDetails);
 
         sutProvider.GetDependency<ICollectionCipherRepository>().GetManyByUserIdCipherIdAsync(userId, id).Returns((ICollection<CollectionCipher>)new List<CollectionCipher>());
-        sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(new Dictionary<Guid, OrganizationAbility> { { cipherDetails.OrganizationId.Value, new OrganizationAbility() } });
+        sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(OrganizationAbilityBuilder.BuildConcurrentDictionary(cipherDetails));
         var cipherService = sutProvider.GetDependency<ICipherService>();
 
         await sutProvider.Sut.PutCollections_vNext(id, model);
@@ -95,7 +96,7 @@ public class CiphersControllerTests
         sutProvider.GetDependency<ICipherRepository>().GetByIdAsync(id, userId).ReturnsForAnyArgs(cipherDetails);
 
         sutProvider.GetDependency<ICollectionCipherRepository>().GetManyByUserIdCipherIdAsync(userId, id).Returns((ICollection<CollectionCipher>)new List<CollectionCipher>());
-        sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(new Dictionary<Guid, OrganizationAbility> { { cipherDetails.OrganizationId.Value, new OrganizationAbility() } });
+        sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(OrganizationAbilityBuilder.BuildConcurrentDictionary(cipherDetails));
 
         var result = await sutProvider.Sut.PutCollections_vNext(id, model);
 
--- a/test/Core.Test/AdminConsole/Services/EventServiceTests.cs
+++ b/test/Core.Test/AdminConsole/Services/EventServiceTests.cs
@@ -1,12 +1,11 @@
 ﻿using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.Entities.Provider;
-using Bit.Core.AdminConsole.Models.Data.Provider;
 using Bit.Core.Context;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Models.Data;
-using Bit.Core.Models.Data.Organizations;
 using Bit.Core.Services;
+using Bit.Core.Test.Helpers;
 using Bit.Test.Common.AutoFixture;
 using Bit.Test.Common.AutoFixture.Attributes;
 using Bit.Test.Common.Helpers;
@@ -27,10 +26,7 @@ public class EventServiceTests
     public async Task LogGroupEvent_LogsRequiredInfo(Group group, EventType eventType, DateTime date,
         Guid actingUserId, Guid providerId, string ipAddress, DeviceType deviceType, SutProvider<EventService> sutProvider)
     {
-        var orgAbilities = new Dictionary<Guid, OrganizationAbility>()
-        {
-            { group.OrganizationId, new OrganizationAbility() { UseEvents = true, Enabled = true } }
-        };
+        var orgAbilities = OrganizationAbilityBuilder.BuildConcurrentDictionary(group);
         sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(orgAbilities);
         sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
         sutProvider.GetDependency<ICurrentContext>().IpAddress.Returns(ipAddress);
@@ -61,10 +57,7 @@ public class EventServiceTests
     public async Task LogGroupEvent_WithEventSystemUser_LogsRequiredInfo(Group group, EventType eventType, EventSystemUser eventSystemUser, DateTime date,
         Guid actingUserId, Guid providerId, string ipAddress, DeviceType deviceType, SutProvider<EventService> sutProvider)
     {
-        var orgAbilities = new Dictionary<Guid, OrganizationAbility>()
-        {
-            { group.OrganizationId, new OrganizationAbility() { UseEvents = true, Enabled = true } }
-        };
+        var orgAbilities = OrganizationAbilityBuilder.BuildConcurrentDictionary(group);
         sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(orgAbilities);
         sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
         sutProvider.GetDependency<ICurrentContext>().IpAddress.Returns(ipAddress);
@@ -98,6 +91,8 @@ public class EventServiceTests
         await sutProvider.GetDependency<IEventWriteService>().Received(1).CreateManyAsync(Arg.Is(AssertHelper.AssertPropertyEqual<IEvent>(expected, new[] { "IdempotencyId" })));
     }
 
+
+
     [Theory]
     [BitMemberAutoData(nameof(InstallationIdTestCases))]
     public async Task LogOrganizationEvent_ProvidesInstallationId(Guid? installationId, EventType eventType,
@@ -120,10 +115,7 @@ public class EventServiceTests
     public async Task LogOrganizationUserEvent_LogsRequiredInfo(OrganizationUser orgUser, EventType eventType, DateTime date,
         Guid actingUserId, Guid providerId, string ipAddress, DeviceType deviceType, SutProvider<EventService> sutProvider)
     {
-        var orgAbilities = new Dictionary<Guid, OrganizationAbility>()
-        {
-            {orgUser.OrganizationId, new OrganizationAbility() { UseEvents = true, Enabled = true } }
-        };
+        var orgAbilities = OrganizationAbilityBuilder.BuildConcurrentDictionary(orgUser);
         sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(orgAbilities);
         sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
         sutProvider.GetDependency<ICurrentContext>().IpAddress.Returns(ipAddress);
@@ -154,10 +146,7 @@ public class EventServiceTests
     public async Task LogOrganizationUserEvent_WithEventSystemUser_LogsRequiredInfo(OrganizationUser orgUser, EventType eventType, EventSystemUser eventSystemUser, DateTime date,
         Guid actingUserId, Guid providerId, string ipAddress, SutProvider<EventService> sutProvider)
     {
-        var orgAbilities = new Dictionary<Guid, OrganizationAbility>()
-        {
-            {orgUser.OrganizationId, new OrganizationAbility() { UseEvents = true, Enabled = true } }
-        };
+        var orgAbilities = OrganizationAbilityBuilder.BuildConcurrentDictionary(orgUser);
         sutProvider.GetDependency<IApplicationCacheService>().GetOrganizationAbilitiesAsync().Returns(orgAbilities);
         sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
         sutProvider.GetDependency<ICurrentContext>().IpAddress.Returns(ipAddress);
@@ -187,10 +176,7 @@ public class EventServiceTests
     public async Task LogProviderUserEvent_LogsRequiredInfo(ProviderUser providerUser, EventType eventType, DateTime date,
         Guid actingUserId, Guid providerId, string ipAddress, DeviceType deviceType, SutProvider<EventService> sutProvider)
     {
-        var providerAbilities = new Dictionary<Guid, ProviderAbility>()
-        {
-            {providerUser.ProviderId, new ProviderAbility() { UseEvents = true, Enabled = true } }
-        };
+        var providerAbilities = ProviderAbilityBuilder.BuildConcurrentDictionary(providerUser);
         sutProvider.GetDependency<IApplicationCacheService>().GetProviderAbilitiesAsync().Returns(providerAbilities);
         sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
         sutProvider.GetDependency<ICurrentContext>().IpAddress.Returns(ipAddress);
@@ -225,10 +211,7 @@ public class EventServiceTests
             providerOrganization.ProviderId = provider.Id;
         }
 
-        var providerAbilities = new Dictionary<Guid, ProviderAbility>()
-        {
-            { provider.Id, new ProviderAbility() { UseEvents = true, Enabled = true } }
-        };
+        var providerAbilities = ProviderAbilityBuilder.BuildConcurrentDictionary(provider);
         sutProvider.GetDependency<IApplicationCacheService>().GetProviderAbilitiesAsync().Returns(providerAbilities);
         sutProvider.GetDependency<ICurrentContext>().UserId.Returns(actingUserId);
         sutProvider.GetDependency<ICurrentContext>().IpAddress.Returns(ipAddress);
@@ -251,4 +234,5 @@ public class EventServiceTests
 
         await sutProvider.GetDependency<IEventWriteService>().Received(1).CreateManyAsync(Arg.Is(AssertHelper.AssertPropertyEqual<IEvent>(expected, new[] { "IdempotencyId" })));
     }
+
 }
--- a/test/Core.Test/Helpers/ProviderAbilityBuilder.cs
+++ b/test/Core.Test/Helpers/ProviderAbilityBuilder.cs
@@ -1,0 +1,37 @@
+﻿using System.Collections.Concurrent;
+using Bit.Core.AdminConsole.Entities.Provider;
+using Bit.Core.AdminConsole.Models.Data.Provider;
+
+namespace Bit.Core.Test.Helpers;
+
+public static class ProviderAbilityBuilder
+{
+    public static ConcurrentDictionary<Guid, ProviderAbility> BuildConcurrentDictionary(ProviderUser providerUser) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, ProviderAbility>(
+                    providerUser.ProviderId,
+                    new ProviderAbility
+                    {
+                        Id = providerUser.ProviderId,
+                        UseEvents = true,
+                        Enabled = true
+                    })
+            });
+
+
+    public static ConcurrentDictionary<Guid, ProviderAbility> BuildConcurrentDictionary(Provider provider) =>
+        new(
+            new[]
+            {
+                new KeyValuePair<Guid, ProviderAbility>(
+                    provider.Id,
+                    new ProviderAbility
+                    {
+                        Id = provider.Id,
+                        UseEvents = true,
+                        Enabled = true
+                    })
+            });
+}
--- a/test/Core.Test/Services/UserServiceTests.cs
+++ b/test/Core.Test/Services/UserServiceTests.cs
@@ -16,11 +16,11 @@ using Bit.Core.Billing.Services;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Exceptions;
-using Bit.Core.Models.Data.Organizations;
 using Bit.Core.Models.Data.Organizations.OrganizationUsers;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
 using Bit.Core.Settings;
+using Bit.Core.Test.Helpers;
 using Bit.Core.Utilities;
 using Bit.Test.Common.AutoFixture;
 using Bit.Test.Common.AutoFixture.Attributes;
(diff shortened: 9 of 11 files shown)


src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
     {
-        _providerAbilities?.Remove(providerId);
-
-        return Task.FromResult(0);
+        _providerAbilities.TryRemove(providerId, out _);
+        return Task.CompletedTask;
     }
 
-    private async Task InitOrganizationAbilitiesAsync()
+    private async Task InitOrganizationAbilitiesAsync() =>
```
JimmyVo16: Note: I'm aware of Justin's nice-to-have recommendation. I'll spend some time on it, but it will go into a separate PR to keep this one small and scoped.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
+            () => _orgAbilities,
+            dict => _orgAbilities = dict,
+            () => _lastOrgAbilityRefresh,
+            dt => _lastOrgAbilityRefresh = dt,
+            _orgInitLock,
+            async () => await _organizationRepository.GetManyAbilitiesAsync(),
+            _refreshInterval,
+            ability => ability.Id);
+
+    private async Task InitProviderAbilitiesAsync() =>
```
JimmyVo16: Another approach is to eagerly initialize the two ability ConcurrentDictionary instances during class initialization, since it's guaranteed to run only once. This shifts the initialization from runtime to startup, which avoids the one-time load delay on the first user request (Justin's point in the Slack thread) and also removes the complexity of creating our own locking mechanism.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: Created a new [ticket](https://bitwarden.atlassian.net/browse/PM-24610) for this.

nick-livefront (approved):
✅ Vault changes are only in tests - altered tests match existing implementations only refactored with the ConcurrentDictionary helpers.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: Please disregard this. We’ll still need the double lock for the refresh.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// FIXME: Update this file to be null safe and then delete the line below
 #nullable disable
```
eliykat, line 2: Please update this while we're here if possible.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -10,24 +11,23 @@
...
-public class InMemoryApplicationCacheService : IApplicationCacheService
+public class InMemoryApplicationCacheService(
+    IOrganizationRepository organizationRepository,
+    IProviderRepository providerRepository)
+    : IApplicationCacheService
 {
-    private readonly IOrganizationRepository _organizationRepository;
-    private readonly IProviderRepository _providerRepository;
+    private readonly IOrganizationRepository _organizationRepository = organizationRepository;
+    private readonly IProviderRepository _providerRepository = providerRepository;
```
eliykat: No need for the class properties in this case, the ctor arguments have a class scope.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -10,24 +11,23 @@
...
-    private IDictionary<Guid, OrganizationAbility> _orgAbilities;
-    private TimeSpan _orgAbilitiesRefreshInterval = TimeSpan.FromMinutes(10);
+    private DateTime _lastProviderAbilityRefresh = DateTime.MinValue;
+    private ConcurrentDictionary<Guid, OrganizationAbility> _orgAbilities;
+    private readonly TimeSpan _refreshInterval = TimeSpan.FromMinutes(10);
+    private readonly SemaphoreSlim _orgInitLock = new(1, 1);
+    private readonly SemaphoreSlim _providerInitLock = new(1, 1);
 
-    private IDictionary<Guid, ProviderAbility> _providerAbilities;
+    private ConcurrentDictionary<Guid, ProviderAbility> _providerAbilities;
```
eliykat: ⛏️ nit: group this together with the orgAbilities definition.

Or you could group them together by org vs. provider, e.g.
```
_orgAbilities
_orgInitLock
_lastOrgAbilityRefresh

_providerAbilities
_providerInitLock
_lastProviderAbilityRefresh
```

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
     public virtual async Task UpsertProviderAbilityAsync(Provider provider)
     {
         await InitProviderAbilitiesAsync();
-        var newAbility = new ProviderAbility(provider);
-
-        _providerAbilities[provider.Id] = newAbility;
+        _providerAbilities.AddOrUpdate(
+            provider.Id,
+            _ => new ProviderAbility(provider),
+            (_, _) => new ProviderAbility(provider));
```
eliykat: The docs say you can still use the index setter with a ConcurrentDictionary:

> **To do this**
> Store a key/value pair in the dictionary unconditionally, and overwrite the value of a key that already exists
> **Use this method**
> The indexer's setter: `dictionary[key] = newValue`

ref: https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-9.0

And [this SO answer](https://stackoverflow.com/a/58649857) confirms it's thread safe (I also had a look in the decompiler myself).

The indexer seems more straightforward given that we always want to set the value.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
+            (_, _) => new OrganizationAbility(organization));
     }
 
     public virtual Task DeleteOrganizationAbilityAsync(Guid organizationId)
     {
-        _orgAbilities?.Remove(organizationId);
-
-        return Task.FromResult(0);
+        _orgAbilities.TryRemove(organizationId, out _);
+        return Task.CompletedTask;
```
eliykat: `_orgAbilities` is potentially null and this change has removed the null forgiving operator. This is where I think enabling nullable will help.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
+        Func<DateTime> getLastRefresh,
+        Action<DateTime> setLastRefresh,
+        SemaphoreSlim @lock,
+        Func<Task<IEnumerable<TAbility>>> fetchFunc,
+        TimeSpan refreshInterval,
+        Func<TAbility, Guid> getId)
     {
-        var now = DateTime.UtcNow;
-        if (_orgAbilities == null || (now - _lastOrgAbilityRefresh) > _orgAbilitiesRefreshInterval)
+        if (getCache() != null && (DateTime.UtcNow - getLastRefresh()) <= refreshInterval)
```
eliykat: Please use [TimeProvider](https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-9.0) whenever you need a DateTime.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
-        var now = DateTime.UtcNow;
-        if (_providerAbilities == null || (now - _lastOrgAbilityRefresh) > _orgAbilitiesRefreshInterval)
+        await @lock.WaitAsync();
+        try
         {
-            var abilities = await _providerRepository.GetManyAbilitiesAsync();
-            _providerAbilities = abilities.ToDictionary(a => a.Id);
-            _lastOrgAbilityRefresh = now;
+            var now = DateTime.UtcNow;
+            if (getCache() == null || (now - getLastRefresh()) > refreshInterval)
```
eliykat: This logic is the same as the previous check, but it's expressed in the opposite way (i.e. `false` skips the rest of the method, instead of `true`). It would be more consistent (and less nesting) to make it a copy + paste of the previous check and return early.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
+            {
+                var sources = await fetchFunc();
```
eliykat: Any thoughts on the timeout approach demonstrated in Robert's doc? 

```c#
        var dbTask = _organizationRepository.GetManyAbilitiesAsync();
        var timeoutTask = Task.Delay(_dbTimeout);

        if (await Task.WhenAny(dbTask, timeoutTask) == dbTask)
        {
            var abilities = await dbTask;
            _orgAbilities = abilities.ToDictionary(a => a.Id);
        }
```

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
+                var abilities = new ConcurrentDictionary<Guid, TAbility>(
+                    sources.ToDictionary(getId));
+                setCache(abilities);
+                setLastRefresh(now);
```
eliykat: Robert's example also set the updated refresh time before calling the db, so that other threads would return early without waiting for the semaphor. I'm not sure it's strictly necessary (that's what the semaphor is for) but I think it probably saves some wasted work waiting for the semaphor.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -51,52 +51,89 @@ public virtual async Task<IDictionary<Guid, ProviderAbility>> GetProviderAbiliti
...
+
+    private async Task InitAbilitiesAsync<TAbility>(
+        Func<ConcurrentDictionary<Guid, TAbility>> getCache,
+        Action<ConcurrentDictionary<Guid, TAbility>> setCache,
+        Func<DateTime> getLastRefresh,
+        Action<DateTime> setLastRefresh,
+        SemaphoreSlim @lock,
+        Func<Task<IEnumerable<TAbility>>> fetchFunc,
+        TimeSpan refreshInterval,
+        Func<TAbility, Guid> getId)
```
eliykat: This is a nice way to share the logic, and in a pretty readable fashion.

However, it makes me think that we could implement all of this logic in a generic base class, then have a separate service for org abilities and provider abilities. I think the only actual difference between the two is the type (`OrganizationAbility` vs `ProviderAbility` - which can be a generic type argument) and the `fetchFunc` (which could be a ctor argument or abstract method).

test/Core.Test/Services/InMemoryApplicationCacheServiceTests.cs
```diff
@@ -1,31 +1,272 @@
...
 
-        _sut = new InMemoryApplicationCacheService(_organizationRepository, _providerRepository);
+        // Assert
+        Assert.Equal(targetAbility, result);
     }
 
-    // Remove this test when we add actual tests. It only proves that
-    // we've properly constructed the system under test.
-    [Fact]
-    public void ServiceExists()
```
eliykat: Love that we're replacing this 😁

test/Core.Test/Services/InMemoryApplicationCacheServiceTests.cs
```diff
@@ -1,31 +1,272 @@
...
+        // Assert
+        Assert.IsType<ConcurrentDictionary<Guid, ProviderAbility>>(result);
+        Assert.Equal(providerAbilities.Count, result.Count);
+        foreach (var ability in providerAbilities)
+        {
+            Assert.True(result.TryGetValue(ability.Id, out var actualAbility));
+            Assert.Equal(ability, actualAbility);
+        }
+        await sutProvider.GetDependency<IProviderRepository>().Received(1).GetManyAbilitiesAsync();
+    }
```
eliykat: Using a single base class would also avoid having to repeat the tests for identical functionality.

test/Core.Test/Services/InMemoryApplicationCacheServiceTests.cs
```diff
@@ -1,31 +1,272 @@
...
+            {
+                var result = await sutProvider.Sut.GetOrganizationAbilitiesAsync();
+                results.Add(result);
+            });
+
+        // Assert
+        var firstResult = results.First();
+        Assert.All(results, result => Assert.Same(firstResult, result));
+        await sutProvider.GetDependency<IOrganizationRepository>().Received(1).GetManyAbilitiesAsync();
+    }
```
eliykat: Very nice 👏 

Very small suggestion: `Assert.Equal(100, results.Count);` just as an extra guarantee that the test is doing what you think it's doing.

test/Core.Test/Services/InMemoryApplicationCacheServiceTests.cs
eliykat, line 1: Potentially missing test coverage: I can't see any tests for the `refreshInterval` logic.

test/Core.Test/Helpers/OrganizationAbilityBuilder.cs
eliykat, line 1: These helpers are fairly opaque and not very flexible. They're all called `BuildConcurrentDictionary`, but the dictionary they return - and the `OrganizationAbility` object contained in that dictionary - are all different.

I suggest you really only need 1 helper: `BuildConcurrentDictionary(OrganizationAbility orgAbility)`, and let the caller construct and pass in the ability they need.

EDIT: if you expose this as an `IReadOnlyDictionary` per my comment below, you aren't forced to use a `ConcurrentDictionary` here. You could just use `Dictionary` for simplicity. And call the method `ToDictionary()`.

test/Identity.Test/Helpers/OrganizationAbilityBuilder.cs
eliykat, line 1: Ideally would use the `Core.Test` helper.

src/Core/Services/IApplicationCacheService.cs
```diff
@@ -7,11 +8,11 @@ namespace Bit.Core.Services;
 
 public interface IApplicationCacheService
 {
-    Task<IDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync();
+    Task<ConcurrentDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync();
```
eliykat: We should expose these as `IReadOnlyDictionary`s so that callers cannot modify their contents.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: I don’t think AddOrUpdate increases complexity by much, as both patterns are extremely readable. AddOrUpdate also signals to devs that it’s thread-safe without them having to confirm anything, and it makes the code easier to extend in the future.

Given that the extra complexity is trivial, I think opting for clearer intent and a small win in extensibility is worth it.

src/Core/Services/IApplicationCacheService.cs
JimmyVo16: I think it’s a good idea to set this to a type that is read-only to make the intent clear and prevent unintended manipulation, but this will require some changes downstream. This PR is starting to get pretty big in scope.

I might consider reverting the concurrentDictionary change to keep this PR small with just the double-lock mechanism.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: This operation isn’t technically atomic because if it fails, the data isn’t reset to its original state. It remains in the state where it failed. For example, if process1 updates the refresh variable and the database operation fails, all other processes will think the cache is updated when it’s not.

If we do it this way, when the database succeeds but updating the refresh variable fails, we’ll still run into the same issue, but the consequence is lower because a second database hit is better than a stale cache.

That said, I don’t think either issue is particularly significant.

JimmyVo16: > I'm not sure it's strictly necessary (that's what the semaphor is for) but I think it probably saves some wasted work waiting for the semaphor.

Yes, if we set the refresh variable first, we’ll stop the other processes at the first null check a little sooner, but there’s still a chance they could pass it, which is why the lock and the double null check are there to ensure they don’t hit the database again.

test/Core.Test/Services/InMemoryApplicationCacheServiceTests.cs
JimmyVo16: Yeah, I've noticed this issue as well.

JimmyVo16: One thing I will say is that there’s a school of thought where people test the interface of a method to assert its behavior and to provide self-documentation, even though many of its internals are tested elsewhere. So if we have a base class, we’ll still have some interface tests for each resource.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: Yeah, it’s a pretty clear separation between the two. I’m okay with the idea of breaking them up into two classes, but we will still have to put them into the cache class for the client; otherwise, the clients will have to import them individually, which isn’t a bad idea, but it is a bigger change.

If we have to inject them back into the cache class, then I don’t see much value in abstracting them.

There are quite a few things we can improve in this class and its child class, InMemoryServiceBusApplicationCacheService. Since it’s a bug, I’m not sure where we want to draw the line.

Not related, but since we’re making big changes, I might as well pull this in. What are your ideas for a feature flag here? All the logic is very interconnected, so putting feature flag checks all over the place isn’t ideal.

Maybe we just create a new, separate class and do one feature check at start-up. This would reduce complexity but would require us to restart the service for a rollback, which is not ideal at all.

test/Identity.Test/Helpers/OrganizationAbilityBuilder.cs
JimmyVo16: I can’t because it’s not a reference. I’m not sure about the patterns for our test projects, but I was trying to avoid potential circular dependencies or disorganized referencing for something so trivial.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
eliykat: Fair enough!

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -69,13 +64,13 @@ public virtual async Task UpsertOrganizationAbilityAsync(Organization organizati
 
     public virtual Task DeleteOrganizationAbilityAsync(Guid organizationId)
     {
-        _orgAbilities.TryRemove(organizationId, out _);
+        _orgAbilities?.TryRemove(organizationId, out _);
```
eliykat: You've made `_orgAbilities` non nullable (which I think is the right move), so you don't need the `?` here anymore.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -69,13 +64,13 @@ public virtual async Task UpsertOrganizationAbilityAsync(Organization organizati
...
         return Task.CompletedTask;
     }
 
     public virtual Task DeleteProviderAbilityAsync(Guid providerId)
     {
-        _providerAbilities.TryRemove(providerId, out _);
+        _providerAbilities?.TryRemove(providerId, out _);
```
eliykat: Same here

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
```diff
@@ -97,43 +92,49 @@ await InitAbilitiesAsync(
...
+            setLastRefresh(timeProvider.GetUtcNow());
         }
         finally
         {
             @lock.Release();
         }
+
+        bool SkipRefresh()
+        {
+            return getCache() != null && (timeProvider.GetUtcNow() - getLastRefresh()) <= refreshInterval;
```
eliykat: `getCache` can never return null.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
eliykat: > This operation isn’t technically atomic because if it fails, the data isn’t reset to its original state. It remains in the state where it failed. For example, if process1 updates the refresh variable and the database operation fails, all other processes will think the cache is updated when it’s not.
> 
> If we do it this way, when the database succeeds but updating the refresh variable fails, we’ll still run into the same issue, but the consequence is lower because a second database hit is better than a stale cache.
> 
> That said, I don’t think either issue is particularly significant.

I agree; I think the practical difference is:
1. update `lastRefresh` first: if the db call fails, wait 10 minutes before trying again
2. update `lastRefresh` last: if the db call fails, keep trying it continuously

I don't know if it's a huge issue, but maybe having an old cache is better than smashing the db? CC @rkac-bw

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
eliykat: > Yeah, it’s a pretty clear separation between the two. I’m okay with the idea of breaking them up into two classes, but we will still have to put them into the cache class for the client; otherwise, the clients will have to import them individually, which isn’t a bad idea, but it is a bigger change.
> 
> If we have to inject them back into the cache class, then I don’t see much value in abstracting them.

I think it's OK for callers to inject the specific cache service they want (organization or provider). These are distinct domains and areas of the codebase, so they generally don't need both. That said, you could keep the current combined interface as a facade to facilitate the change.

> Not related, but since we’re making big changes, I might as well pull this in. What are your ideas for a feature flag here? All the logic is very interconnected, so putting feature flag checks all over the place isn’t ideal.
> 
> Maybe we just create a new, separate class and do one feature check at start-up. This would reduce complexity but would require us to restart the service for a rollback, which is not ideal at all.

I agree that feature flags on startup are not a great idea for that reason. I would feature flag this by:
- creating a vNext implementation
- having each method of the current implementation redirect the call to the equivalent method of the vNext implementation when the feature flag is on

So for example:

```c#
// current class
    public virtual async Task<IDictionary<Guid, OrganizationAbility>> GetOrganizationAbilitiesAsync()
    {
        // redirect if the flag is on
        if (featureService.IsEnabled(FeatureFlagKeys.FixApplicationCacheService)
        {
            return await vNextApplicationCacheService.GetOrganizationAbilitiesAsync();
        }
        
        // otherwise continue with the current implementation
        await InitOrganizationAbilitiesAsync();
        return _orgAbilities;
    }
```

> There are quite a few things we can improve in this class and its child class, InMemoryServiceBusApplicationCacheService. Since it’s a bug, I’m not sure where we want to draw the line.

We don't have to separate the org vs. provider cache services in this PR - I'll leave that as your call. However I do suggest that if we're using the current implementation as a facade anyway (per above), it's just as easy to redirect to the calls to separate cache services.

test/Core.Test/Services/InMemoryApplicationCacheServiceTests.cs
eliykat: Good point. So if we had a base class with 2 child classes, each child class should probably have its own set of tests anyway.

That said (if you decide to split the caches), you may be able to avoid a base class and just have a single concrete class which can be instantiated with different type and ctor arguments to suit each domain. Then, only 1 class, only 1 test suite.

It's not a big deal either way. I think that would be a elegant solution but it's possible it wouldn't work for some reason, or just out of scope here.

src/Core/Services/IApplicationCacheService.cs
eliykat: Fair enough, we can call this out of scope. (although it does alarm me that we're just handing out this mutable dict to whoever wants it 0_0)

However, `ConcurrentDictionary` also implements `IDictionary`, which would match the current interface. So I think you should keep `ConcurrentDictionary` internally (given that it's related to the concurrency issues we're fixing here), but you can keep the external interface the same as it is today, which should limit impact on callers.

JimmyVo16: > (although it does alarm me that we're just handing out this mutable dict to whoever wants it 0_0)

Yeah, this is one of a couple of strange things about this class lol.

JimmyVo16: > you can keep the external interface the same as it is today, which should limit impact on callers.

This is brilliant! I can undo all the test stuff and make the interface a little better.

test/Core.Test/Helpers/OrganizationAbilityBuilder.cs
JimmyVo16: This comment is no longer applicable.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: Sorry, I missed this part. 

The intent of the code is that if the DB takes too long, we can fall back to the previous value, so it doesn’t seem like Consistency (the C in CAP) is a high priority here. 

I think this is a good approach. My concern is, what if this is the initial request and there’s no value in the cache? Also, what if it continues past 30 seconds? Would these two conditions prevent that pod from operating normally?

I think this is a valid problem, and I’ll spend some time to see if I can find a good solution.

JimmyVo16:
Update: I'm still addressing some of the code review comments. The two big things I'm planning to do are looking into exponential backoff and feature-flagging this.

withinfocus:
I plan on taking a look once all of @eliykat's comments are addressed by the way.

You mentioned backoff strategies and that feels out of scope here, and arguably impossible to do accurately without that backoff knowledge being shared across instances and therefore propagating the problem. I would like to think that the concurrency improvements alone will give this plenty of room for growth and scale.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
eliykat: Those are good points. Maybe we can leave it for now depending on @rkac-bw 's feedback, as an empty or significantly out of date cache _will_ cause problems for end users. The database layer itself should have some kind of timeout, and we can monitor latency in Datadog.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: So, thinking this through, I don't think we need this null check anymore. Since our formula `timeProvider.GetUtcNow() - getLastRefresh() <= refreshInterval` will always trigger when it first initializes, removing the null reference check reduces complexity.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: I tried adding the feature flag using IFeatureService, but I ran into an issue: the cache service is a singleton, while IFeatureService is scoped. The cache needs to stay a singleton; otherwise, there’s not much point in doing this. 

I’m concerned about making IFeatureService a singleton because it would affect the entire application, and I’m not sure if LaunchDarkly would play nice. It might cache the values, which could prevent us from turning off the feature flags. I’m not familiar enough with this area to make an informed decision.

The third option is to use IServiceProvider in the cache to resolve IFeatureService. It feels dirty and overly complex.

I'm open to suggestions.

withinfocus: The feature service lifetime cannot change and will explode if you did. Thomas' point I believe is to have a scoped wrapper that reads the flag and makes a routing decision. The cache itself as a singleton won't touch the flag logic and it's just the usage that's related.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: I think exponential backoff with jitter is a better option here. This helps the cache retry when the database times out, and the jitter will prevent accidental synchronized spikes (when all pods try again at the same time). Of course, since it's random, there will still be a chance that more pods than you would have liked hit the db at the same time. 

The only way to avoid this for sure is to have some kind of orchestration between the pods, but of course, that will add more scopes.

Then there’s the maximum request timeout for the HTTP request. I'm not sure if we have it here, but it's common to have something like that, so we have to work within that range, or else our logic is pointless.

JimmyVo16:
Hey folks, @eliykat @withinfocus  as I'm working through adding the feature flag and exponential backoff, I’ve noticed that the scope can get pretty big. Based on my conversation with Robert, the issue was that the cache logic was simply wrong in the past. The current PR addresses this problem, and we have test coverage to assert the behavior.

Can we get alignment on what we need to have in this PR? Since the problem has existed for a while, I'm guessing there’s no hard deadline for it; however, I will say that getting the fix out and continuing to iterate is better than having one giant PR. We can also continue adding more stuff in this ticket, but it can go in another PR. This will improve testing, code review, and rollback.

withinfocus:
All functionality should be feature-flagged by rule, and given that this is a hot path it needs to be protected -- unforeseen issues could bring the whole system down with this. The feature flag does need to be introduced -- I would shim the current calls to use the new or old version based on it.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: Yeah, that should address the DI problem

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
eliykat: I assumed that was your plan all along ;)

eliykat:
As discussed today, we don't need to split the service into separate Organization + Provider cache services in this PR. We can leave that as a future improvement. Similarly, exponential backoff isn't required to address the immediate problem, so let's leave that for now.

I think the main piece of feedback outstanding is to feature flag this change per [discussions above](https://github.com/bitwarden/server/pull/6170#discussion_r2279337448). Then please request my review again.

src/Core/Services/IFeatureRoutedCacheService.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Bit.Core.AdminConsole.Entities;
+using Bit.Core.AdminConsole.Entities.Provider;
+using Bit.Core.AdminConsole.Models.Data.Provider;
+using Bit.Core.Models.Data.Organizations;
+
+namespace Bit.Core.Services;
+
+public interface IFeatureRoutedCacheService
```
JimmyVo16: I would love some feedback on the class name. This seems like something we should keep long-term, as the problem of feature flagging a singleton service will likely come up again. 

That said, let's not spend too much time on this; something reasonable is fine. There will be a cleanup task to remove the feature flag later, so we can spend more time on it then if needed.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: We have a challenge here. The `InMemoryServiceBusApplicationCacheService` and `InMemoryApplicationCacheService` both use the same `IApplicationCacheService`, and `InMemoryServiceBusApplicationCacheService` inherits from `InMemoryApplicationCacheService`.

---

### Options:

1. Decouple `InMemoryServiceBusApplicationCacheService` and `InMemoryApplicationCacheService`.
   - Inject a new feature-flagged version of `InMemoryApplicationCacheService`.
   - Create a new interface for `InMemoryServiceBusApplicationCacheService` so it’s more explicit where it’s used and maintained.
   - Overall, composition is better than inheritance for this use case.
   - For the feature flag, it could be tricky since `InMemoryServiceBusApplicationCacheService` is also a singleton, so it will need its own wrapper.
2. Create another feature flag wrapper for `InMemoryServiceBusApplicationCacheService`.
   - We might have to duplicate all of `InMemoryServiceBusApplicationCacheService`’s event logic for `DeleteOrganizationAbilityAsync` and `DeleteProviderAbilityAsync` because we can’t just replace the inherited behavior.
   - This involves more duplicate work and expands the impact radius, but from the feature flag perspective, it’s probably easier.
   - Since the existing caches are being registered in an if statement, we can add the vNext version with its respective cache, with no risk of mixing them up.

---

### Other notes/considerations:

1. I just want to highlight that the new version and the old version are using different data sources. If we switch over, it’s going to hit the database again.
2. Should we consider releasing just `InMemoryApplicationCacheService` vNext first and then following up with `InMemoryServiceBusApplicationCacheService` to reduce the impact area?
   - Since we're using conditional DI registration, we only need to register one cache service, not the other.


---

@eliykat @withinfocus Please let me know your thoughts. I’m open to exploring new options or improving the existing ones.

src/Core/Services/IFeatureRoutedCacheService.cs
eliykat: I think the name is fine. My suggestion is `FeatureFlaggedCacheService` but I don't mind.

A short xmldoc explaining the purpose of this interface/class might help as well.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
```diff
@@ -0,0 +1,97 @@
...
+using Bit.Core.AdminConsole.Models.Data.Provider;
+using Bit.Core.Models.Data.Organizations;
+
+namespace Bit.Core.Services.Implementations;
+
+public class FeatureRoutedCacheService(
+    IFeatureService featureService,
+    IVNextInMemoryApplicationCacheService vNextInMemoryApplicationCacheService,
+    IApplicationCacheService inMemoryApplicationCacheService)
+    : IFeatureRoutedCacheService
```
eliykat: You want all callers to use this instead of the current implementation. With the current approach, I think you'd have to update all callers manually, but it would be better to use DI to just inject the new implementation.

You could have this class (the shim) implement current interface that everyone already injects, and then have the current and vNext implementations use new interface definitions (which would only be injected by this class).

e.g.

```c#
public class FeatureRoutedCacheService : IApplicationCacheService; // the current interface that everyone already injects

public class VCurrentInMemoryApplicationCacheService : IVCurrentInMemoryApplicationCacheService; // old implementation

public class VNextInMemoryApplicationCacheService : IVNextInMemoryApplicationCacheService; // new implementation
```

src/Core/Services/IFeatureRoutedCacheService.cs
eliykat, line 1: I just noticed none of this is in our code ownership. Could you move all new files into (for example) `Core/AdminConsole/AbilitiesCache/`?

The existing implementation will be deleted soon so I am less concerned about that.

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
eliykat: My suggestion is to decouple them as follows:
1. copy the service bus logic in `InMemoryApplicationServiceBusApplicationCacheService` into a new class that is not a subclass of anything. It only sends the service bus messages, nothing else. Remove any calls to what was formerly the base class.
2. make a no-op implementation of the new class to be used when service bus messaging is not configured.
3. update the `VNextInMemoryApplicationCacheService` to inject this new service bus interface and call it.
4. conditionally register the correct implementation (service bus or no-op) in the startup code.

The resut is that the old implementation(s) continue to use the inheritance approach; the new implementation uses a composition approach. The new service bus class is only used by the new implementation, so no additional flagging is required.

So the startup code would look something like:

```c#
        // always registered because now the conditional registration only affects the service bus service
        services.AddSingleton<IVNextInMemoryApplicationCacheService, VNextInMemoryApplicationCacheService>();

        if (CoreHelpers.SettingHasValue(globalSettings.ServiceBus.ConnectionString) &&
            CoreHelpers.SettingHasValue(globalSettings.ServiceBus.ApplicationCacheTopicName))
        {
            // existing
            services.AddSingleton<IApplicationCacheService, InMemoryServiceBusApplicationCacheService>();
            // new
            services.AddSingleton<IApplicationCacheServiceBusService, ApplicationCacheServiceBusService>();
        }
        else
        {
            // existing
            services.AddSingleton<IApplicationCacheService, InMemoryApplicationCacheService>();
            // new
            services.AddSingleton<IApplicationCacheServiceBusService, NoopApplicationCacheServiceBusService>();
        }
```

I think the alternative would be to make `FeatureRoutedCacheService` generic, which could route between any 2 implementations of an `IApplicationFeatureCacheService`. But that feels overcomplicated to me, particularly because it affects the feature flagging logic itself (which should be simple and safe). Decoupling the 2 classes is simpler which suggests it's probably a better design overall.

src/Core/Services/Implementations/VNextInMemoryApplicationCacheService.cs
```diff
@@ -0,0 +1,135 @@
...
+            () => _lastOrgAbilityRefresh,
+            dt => _lastOrgAbilityRefresh = dt,
+            _orgInitLock,
+            async () => await organizationRepository.GetManyAbilitiesAsync(),
+            _refreshInterval,
+            ability => ability.Id);
+
+    private async Task InitProviderAbilitiesAsync() =>
+       await InitAbilitiesAsync<ProviderAbility>(
+            concurrentDictionary => _providerAbilities = concurrentDictionary,
```
eliykat: I suggest more consistent naming with the previous method: `dict => _providerAbilities = dict`.

test/Core.Test/Services/VNextInMemoryApplicationCacheServiceTests.cs
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    [Theory, BitAutoData]
+    public async Task DeleteOrganizationAbilityAsync_NullSource_DoesNotThrow(
+        Guid organizationId,
+        SutProvider<VNextInMemoryApplicationCacheService> sutProvider)
+    {
+        // Act & Assert
+        await sutProvider.Sut.DeleteOrganizationAbilityAsync(organizationId);
+    }
```
eliykat: No longer relevant because the dict is never null.

test/Core.Test/Services/VNextInMemoryApplicationCacheServiceTests.cs
```diff
@@ -0,0 +1,416 @@
...
+
+    [Theory, BitAutoData]
+    public async Task DeleteProviderAbilityAsync_NullSource_DoesNotThrow(
+        Guid providerId,
+        SutProvider<VNextInMemoryApplicationCacheService> sutProvider)
+    {
+        // Act & Assert
+        await sutProvider.Sut.DeleteProviderAbilityAsync(providerId);
+    }
```
eliykat: Same here

test/Core.Test/Services/VNextInMemoryApplicationCacheServiceTests.cs
```diff
@@ -0,0 +1,416 @@
...
+        List<OrganizationAbility> updatedAbilities)
+    {
+        // Arrange
+        var fakeTimeProvider = new FakeTimeProvider();
+        var orgRepo = Substitute.For<IOrganizationRepository>();
+        var providerRepo = Substitute.For<IProviderRepository>();
+
+        orgRepo.GetManyAbilitiesAsync().Returns(organizationAbilities, updatedAbilities);
+
+        var sut = new VNextInMemoryApplicationCacheService(orgRepo, providerRepo, fakeTimeProvider);
```
eliykat: (non-blocking) Tip: you can instantiate a new SutProvider with FakeTimeProvider configured like this:

```c#
        var sutProvider = new SutProvider<VNextInMemoryApplicationCacheService>()
            .WithFakeTimeProvider()
            .Create();
```

src/Core/Services/Implementations/InMemoryApplicationCacheService.cs
JimmyVo16: I think your first idea of having a bus service and a no-op could work.

Another problem we have is this weird part of the [code](https://github.com/bitwarden/server/blob/9081c205b1bc5e9d57af747e2227dca1a50d5002/src/Core/HostedServices/ApplicationCacheHostedService.cs#L144) that directly uses `InMemoryServiceBusApplicationCacheService`.

~I think we can introduce a new interface (for `BaseUpsertOrganizationAbilityAsync` and `BaseDeleteOrganizationAbilityAsync`) for this area of the code, then have the same FeatureRoutedCacheService handle it accordingly based on the feature flag.~

Update: Yeah, I think we can just do the same existing pattern and use `as FeatureRoutedCacheService`. We should circle back on this and come up with something better for this particular problem.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
```diff
@@ -0,0 +1,150 @@
...
+/// - We need to evaluate feature flags at request time while maintaining singleton cache behavior
+///
+/// The service acts as a scoped proxy that can access the scoped IFeatureService while
+/// delegating actual cache operations to the appropriate singleton implementation.
+/// </remarks>
+public class FeatureRoutedCacheService(
+    IFeatureService featureService,
+    IVNextInMemoryApplicationCacheService vNextInMemoryApplicationCacheService,
+    IVCurrentInMemoryApplicationCacheService inMemoryApplicationCacheService,
+    IApplicationCacheServiceBusMessaging serviceBusMessaging)
```
JimmyVo16, line 28: This message will include the whole review status up to this point, so we can comment back and forth.

I think this should solve our problem with the feature flag. I haven’t had a chance to test it at runtime yet, and I still need to add tests for the IApplicationCacheServiceBusMessaging implementations and update the FeatureRoutedCacheService flag.

@eliykat Please let me know if there’s anything else I should add.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            await vNextInMemoryApplicationCacheService.UpsertOrganizationAbilityAsync(organization);
+        }
+        else
+        {
+            if (inMemoryApplicationCacheService is InMemoryServiceBusApplicationCacheService serviceBusCache)
+            {
+                await serviceBusCache.UpsertOrganizationAbilityAsync(organization);
+            }
+            else
```
JimmyVo16, line 125: So this shouldn't happen based on how we configure the DI, but it should fail hard if it does, to prevent sneaky bugs.

src/Core/HostedServices/ApplicationCacheHostedService.cs
eliykat, line 1: OK this is a problem in with the idea for the separate `ApplicationCacheServiceBusMessaging` class.

This is the _receiving_ end of the message; if it receives a message over the service bus (from another instance), it updates its own `OrganizationAbilityCache`. The way it's currently written, it calls the base method on `InMemoryServiceBusApplicationCacheService` so that it doesn't trigger a circular message.

However, with this change we no longer have this distinction; we have no way to update our local cache without sending a message over the bus, which could trigger really bad circular messaging.

I'm not sure how to solve this at the moment; I'll think on it. It's possible that keeping the previous inheritance structure and just having separate feature flagged versions of each implementation is boilerplate-y but simpler, and avoids refactoring the entire area which isn't really our goal here. Thoughts?

eliykat (commented):
My latest thinking, following on from my [previous comment](https://github.com/bitwarden/server/pull/6170#discussion_r2309409863).

I think we should revert the `ServiceBusApplicationCacheMessaging` refactor and maintain the current class hierarchy, basically to limit scope and not get caught up refactoring the service bus mechanism.

Suggestion:

- `FeatureRoutedInMemoryApplicationCacheService` routes between:
   - current implementation
   - vNext implementation
- `FeatureRoutedInMemoryServiceBusApplicationCacheService` routes between:
   - current implementation (child class of non-service bus current implementation)
   - vNext implementation (child class of non-service bus vNext implementation)
- Startup code registers the appropriate "feature routed" depending on whether we want service bus or non-service bus.

One tweak that can reduce boilerplate here is that I don't think we need separate vCurrent and vNext interface definitions: the routed classes can be registered under the existing interfaces (which is injected by consumers), and they can inject the concrete implementations they needs (vNext and vCurrent) rather than interfaces. This is what `ApplicationCacheHostedService` does when it requires a specific implementation, and I think it matches what we actually need here rather than having duplicate interface definitions just for DI purposes.

Let me know what you think, open to further discussion.

src/Core/HostedServices/ApplicationCacheHostedService.cs
JimmyVo16: > One tweak that can reduce boilerplate here is that I don't think we need separate vCurrent and vNext interface definitions: the routed classes can be registered under the existing interfaces (which is injected by consumers), and they can inject the concrete implementations they needs (vNext and vCurrent) rather than interfaces. This is what ApplicationCacheHostedService does when it requires a specific implementation, and I think it matches what we actually need here rather than having duplicate interface definitions just for DI purposes.

I'm not sure what benefit this gives us… an extra interface doesn't seem so bad. Plus, the DI configuration is declarative and simpler than if we implemented our own creation logic. We also need to figure out how to unit test it. All of this seems like a lot of work for a solved problem since the code is already implemented.

JimmyVo16: > However, with this change we no longer have this distinction; we have no way to update our local cache without sending a message over the bus, which could trigger really bad circular messaging.

I made a mistake in the code. It should be calling `serviceBusCache.BaseDeleteOrganizationAbilityAsync`. This should address our problem.

JimmyVo16: Ok, my latest changes should address the circular issue. I also added tests to assert the behavior. I had to do some funkiness, but I left comments explaining.

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
```diff
@@ -0,0 +1,135 @@
...
+            _providerInitLock,
+            async () => await providerRepository.GetManyAbilitiesAsync(),
+            _refreshInterval,
+            ability => ability.Id);
+
+
+    private async Task InitAbilitiesAsync<TAbility>(
+        Action<ConcurrentDictionary<Guid, TAbility>> setCache,
+        Func<DateTimeOffset> getLastRefresh,
+        Action<DateTimeOffset> setLastRefresh,
```
justindbaur, line 102: I think you could save yourself a couple of functions you have to pass around by just passing the ref the DateTimeOffset you want to update/read.

```suggestion
        ref DateTimeOffset lastRefresh
```

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public virtual async Task UpsertOrganizationAbilityAsync(Organization organization)
+    {
+        await InitOrganizationAbilitiesAsync();
+
+        _orgAbilities.AddOrUpdate(
+            organization.Id,
+            _ => new OrganizationAbility(organization),
+            (_, _) => new OrganizationAbility(organization));
```
justindbaur: You can use the overloads that let you give it a factory argument to avoid some extra closures. The `static` isn't technically necessary to avoid the closure but by it being there it forces you to make sure you don't rely on any outer scope.

```suggestion
        _orgAbilities.AddOrUpdate(
            organization.Id,
            static (_, organization) => new OrganizationAbility(organization),
            static (_, _, organization) => new OrganizationAbility(organization),
            organization);
```

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
```diff
@@ -0,0 +1,135 @@
...
+        return _providerAbilities;
+    }
+
+    public virtual async Task UpsertProviderAbilityAsync(Provider provider)
+    {
+        await InitProviderAbilitiesAsync();
+        _providerAbilities.AddOrUpdate(
+            provider.Id,
+            _ => new ProviderAbility(provider),
+            (_, _) => new ProviderAbility(provider));
```
justindbaur: Same idea.

```suggestion
        _providerAbilities.AddOrUpdate(
            provider.Id,
            static (_, provider) => new ProviderAbility(provider),
            static (_, _, provider) => new ProviderAbility(provider),
            provider);
```

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
```diff
@@ -0,0 +1,135 @@
...
+        Func<Task<IEnumerable<TAbility>>> fetchFunc,
+        TimeSpan refreshInterval,
+        Func<TAbility, Guid> getId)
+    {
+        if (SkipRefresh())
+        {
+            return;
+        }
+
+        await @lock.WaitAsync();
```
justindbaur, line 113: Possible future enhancement could be to take and pass along a `CancellationToken` to this method and the `fetchFunc`. Then, especially the `Update*` and `Delete*` methods should take a cancellation token and pass it here. I'm slightly less concerned about the `Get*` methods but I'm certainly not opposed to it being added there. 

My thinking about why that could be valuable is that it's only a hosted service that calls the update/delete methods and if we are wanting to kill a service then there is no longer any value in keeping these locks and queries going. We can kill them and let the service shutdown a little quicker.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
```diff
@@ -61,13 +61,12 @@ public async Task UpsertOrganizationAbilityAsync(Organization organization)
...
         {
             await vNextInMemoryApplicationCacheService.UpsertOrganizationAbilityAsync(organization);
+            await serviceBusMessaging.NotifyOrganizationAbilityUpsertedAsync(organization);
         }
         else
         {
             await inMemoryApplicationCacheService.UpsertOrganizationAbilityAsync(organization);
         }
-
-        await serviceBusMessaging.NotifyOrganizationAbilityUpsertedAsync(organization);
```
eliykat: This works for now as best I can tell, but what happens when you remove `FeatureRoutedCacheService`? I assume you'd inline both calls into all calling locations, but we can't rely on callers to know about this extra message bus service.

Reverting back to the class hierarchy (see my [comment](https://github.com/bitwarden/server/pull/6170#pullrequestreview-3174580843) above) would resolve this. But my assumption is you don't want to rework the approach again. The other alternative I can think of would be to add additional methods to `IVNextInMemoryApplicationCacheService.cs` to "update locally only" (without hitting the messaging bus) - the equivalent of the `Base` methods on the old implementation. It's a tricky problem so these are suggestions only.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
```diff
@@ -137,9 +137,11 @@ public async Task BaseDeleteOrganizationAbilityAsync(Guid organizationId)
         }
         else
         {
+            // NOTE: This is a temporary workaround InMemoryServiceBusApplicationCacheService legacy implementation.
+            // Avoid using this approach in new code.
             if (inMemoryApplicationCacheService is InMemoryServiceBusApplicationCacheService serviceBusCache)
             {
-                await serviceBusCache.DeleteOrganizationAbilityAsync(organizationId);
+                await serviceBusCache.BaseDeleteOrganizationAbilityAsync(organizationId);
```
eliykat: I don't quite understand the comment - wouldn't new code be using the new implementation (after this shim service is removed)?

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
justindbaur: Strike this, I forgot you can't pass value types by ref into async methods.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
JimmyVo16: There are a few problems with the current code that I think we should address separately from the original issue of it not being atomic.

**Problems**
1. It’s using a method that is not on the interface, which is why we had to use this [workaround](https://github.com/bitwarden/server/pull/6170#discussion_r2335444588). This makes it harder to test and causes us to lose many compile-time checking benefits.  
2. We should move away from inheritance, as you can see how difficult it is to apply a feature flag in this setup.

**Potential Solutions**
1. Create another interface for the bus, or fix the existing one by ensuring all the methods the client needs are included on the interface.  
2. Use a bus service that contains all the bus functionality, and it can have the in-memory cache as a dependency.

I can start a doc to include all these items, along with the exponential backoff and a base class that the two resources can inherit. I would strongly advise that these changes not block the solution to the original problem from going out.

Edit: I meant to say `these changes not block the solution` not `these changes block the solution`. Sorry for the confusion.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
JimmyVo16: Addressed this in the other [thread’s comment](https://github.com/bitwarden/server/pull/6170#discussion_r2337595609).

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
JimmyVo16: Yeah, that’s a good callout. Originally, when I considered adding cancellation, I thought it wasn’t worth it since we only had a database hit. But adding in the locks makes sense to me. I can add this to a doc for future improvements.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
JimmyVo16: [Doc](https://bitwarden.atlassian.net/wiki/x/FYD-dg) created

src/Core/AdminConsole/AbilitiesCache/VNextInMemoryApplicationCacheService.cs
JimmyVo16: [Doc](https://bitwarden.atlassian.net/wiki/x/FYD-dg) created

JimmyVo16:
Looks like main is broken. I’ve [notified](https://bitwarden.slack.com/archives/C05LRC3DBDK/p1757532691947189) the team that owns this code, and I’ll monitor it.

eliykat:
Please treat architecture's review as blocking - you can remove the hold label once @justindbaur has approved.

PR #6172, opened 2025-08-07
Adding concurrency rate limiter to scim endpoint.

jrmccannon:
## 🎟️ Tracking
[PM-24462](https://bitwarden.atlassian.net/browse/PM-24462)

## 📔 Objective
Since inviting users needs to be transactional and we need to return an error to the IdP for SCIM, this adds the `ConcurrencyRateLimiter` to the PostUsers SCIM endpoint.

This acts as a queue that only allows 1 request per organization to be processed. This should solve any concurrency issues we have while ensuring organizations can add users independently.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-24462]: https://bitwarden.atlassian.net/browse/PM-24462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bitwarden_license/src/Scim/Startup.cs
+++ b/bitwarden_license/src/Scim/Startup.cs
@@ -1,8 +1,10 @@
 ﻿using System.Globalization;
+using Bit.Core;
 using Bit.Core.Billing.Extensions;
 using Bit.Core.Context;
 using Bit.Core.SecretsManager.Repositories;
 using Bit.Core.SecretsManager.Repositories.Noop;
+using Bit.Core.Services;
 using Bit.Core.Settings;
 using Bit.Core.Utilities;
 using Bit.Scim.Context;
@@ -63,6 +65,9 @@ public class Startup
             });
         });
 
+        // Rate Limiter
+        services.AddScimConcurrencyRateLimiter();
+
         // Identity
         services.AddCustomIdentityServices(globalSettings);
 
@@ -95,6 +100,7 @@ public class Startup
         IApplicationBuilder app,
         IWebHostEnvironment env,
         IHostApplicationLifetime appLifetime,
+        IFeatureService featureService,
         GlobalSettings globalSettings)
     {
         app.UseSerilog(env, appLifetime, globalSettings);
@@ -128,6 +134,12 @@ public class Startup
 
         // Add current context
         app.UseMiddleware<CurrentContextMiddleware>();
+
+        // Add Rate Limiter
+        if (featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization))
+        {
+            app.UseRateLimiter();
+        }
 
         // Add MVC to the request pipeline.
         app.UseEndpoints(endpoints => endpoints.MapDefaultControllerRoute());
--- a/bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
+++ b/bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
@@ -1,0 +1,20 @@
+﻿using System.Threading.RateLimiting;
+using Bit.Core.AdminConsole.Authorization;
+
+namespace Bit.Scim.Utilities;
+
+public static class ScimConcurrencyRateLimiter
+{
+    public const string PolicyName = "ScimConcurrencyLimiter";
+
+    public static IServiceCollection AddScimConcurrencyRateLimiter(this IServiceCollection services) =>
+        services.AddRateLimiter(options => options.AddPolicy(PolicyName, context =>
+            RateLimitPartition.GetConcurrencyLimiter(context.GetOrganizationId(),
+                factory: _ => new ConcurrencyLimiterOptions
+                {
+                    PermitLimit = int.MaxValue, // How much do we want in memory
+                    QueueLimit = 1,
+                    QueueProcessingOrder = QueueProcessingOrder.OldestFirst
+                })
+        ));
+}
--- a/bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
+++ b/bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
@@ -12,8 +12,8 @@ public static class ScimConcurrencyRateLimiter
             RateLimitPartition.GetConcurrencyLimiter(context.GetOrganizationId(),
                 factory: _ => new ConcurrencyLimiterOptions
                 {
-                    PermitLimit = int.MaxValue, // How much do we want in memory
-                    QueueLimit = 1,
+                    PermitLimit = 1,
+                    QueueLimit = int.MaxValue,
                     QueueProcessingOrder = QueueProcessingOrder.OldestFirst
                 })
         ));
--- a/bitwarden_license/test/Scim.IntegrationTest/Controllers/v2/UsersControllerTests.cs
+++ b/bitwarden_license/test/Scim.IntegrationTest/Controllers/v2/UsersControllerTests.cs
@@ -596,4 +596,52 @@ public class UsersControllerTests : IClassFixture<ScimApplicationFactory>, IAsyn
         var databaseContext = _factory.GetDatabaseContext();
         Assert.Equal(_initialUserCount, databaseContext.OrganizationUsers.Count());
     }
+
+
+    [Fact]
+    public async Task Post_RateLimited_TooManyRequests()
+    {
+        // Arrange - Create a new factory instance to avoid interference with other tests
+        var localFactory = new ScimApplicationFactory();
+        localFactory.SubstituteService((IFeatureService featureService)
+            => featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization)
+                .Returns(true));
+
+        localFactory.ReinitializeDbForTests(localFactory.GetDatabaseContext());
+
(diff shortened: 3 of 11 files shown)


src/Core/Settings/GlobalSettings.cs
```diff
@@ -521,7 +522,7 @@ public class SyslogSettings
         /// The absolute path to a Certificate (DER or Base64 encoded with private key).
         /// </summary>
         /// <remarks>
-        /// The certificate path and <see cref="CertificatePassword"/> are passed into the <see cref="System.Security.Cryptography.X509Certificates.X509Certificate2.X509Certificate2(string, string)" />.
+        /// The certificate path and <see cref="CertificatePassword"/> are passed into the <see cref="X509Certificate2" />.
```
withinfocus, line 525: ❓ Unrelated change?

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Bit.Scim.Utilities;
+
+public static class ScimConcurrencyRateLimiter
+{
+    public const string PolicyName = "ScimConcurrencyLimiter";
+
+    public static IServiceCollection AddScimConcurrencyRateLimiter(this IServiceCollection services) =>
+        services.AddRateLimiter(options => options.AddPolicy(PolicyName, context =>
+            RateLimitPartition.GetConcurrencyLimiter(context.GetOrganizationId(),
```
withinfocus, line 12: ❓ `CurrentContext` isn't available at this point? Lot easier than the extension method.

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
```diff
@@ -0,0 +1,20 @@
...
+using Bit.Core.AdminConsole.Authorization;
+
+namespace Bit.Scim.Utilities;
+
+public static class ScimConcurrencyRateLimiter
+{
+    public const string PolicyName = "ScimConcurrencyLimiter";
+
+    public static IServiceCollection AddScimConcurrencyRateLimiter(this IServiceCollection services) =>
+        services.AddRateLimiter(options => options.AddPolicy(PolicyName, context =>
```
withinfocus, line 11: ❓ This won't work across instances right? So at scale this won't provide what you need.

withinfocus (commented):
This is fancy and I hadn't heard of it, but I'm not sure it will work. Can't you just use a database transaction?

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
jrmccannon: You can provide a distributed store for it to use. 

I'm concerned about how the transaction lock would work. We do a bunch of initial database calls, then figure out if we can add the user, then add the user and perform side effects.

Do we have any examples of how that would work across different calls?

src/Core/Settings/GlobalSettings.cs
jrmccannon: This was an error in rider that was bugging me so I added it 🤷

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
jrmccannon: I don't believe `CurrentContext` would be available but I can look again

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
withinfocus: There are [some](https://github.com/bitwarden/server/blob/9d05105dc0a6c06b381d46be1a5d228b97fe9297/src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs#L422) around -- look for `BeginTransaction`. You want to do the selects and commits in the one transaction. Yes, it locks for a moment, but that is by design.

@rkac-bw may dislike some of this approach and it deserves performance review, but I consider this the standard -- you need to lock your context for your selects so that they are accurate.

jrmccannon: I'm not 100% sure if this is a viable solution. 

I would have to open the transaction, get my data, do my validation against the state of the organization, then up the seat count, commit the transaction and then send out the emails.

Is that what you were thinking?

withinfocus: Yep.

bitwarden_license/src/Scim/Startup.cs
```diff
@@ -129,6 +135,12 @@ public void Configure(
         // Add current context
         app.UseMiddleware<CurrentContextMiddleware>();
 
+        // Add Rate Limiter
+        if (featureService.IsEnabled(FeatureFlagKeys.ScimInviteUserOptimization))
+        {
+            app.UseRateLimiter();
+        }
```
eliykat, line 142: I think we want to avoid using feature flags in startup code, otherwise we have to reboot the instance if we want to toggle the flag. Better to check it at the point of enforcement instead (e.g. when the request comes in). I'm not sure if/how that be configured with the middleware though.

eliykat (commented):
Seems reasonable to me, but we can also try @withinfocus 's suggestion of locking the row. If we limit it to a short lived read-validate-update operation it may not impact other requests too much.

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
```diff
@@ -0,0 +1,20 @@
...
+            RateLimitPartition.GetConcurrencyLimiter(context.GetOrganizationId(),
+                factory: _ => new ConcurrencyLimiterOptions
+                {
+                    PermitLimit = int.MaxValue, // How much do we want in memory
+                    QueueLimit = 1,
+                    QueueProcessingOrder = QueueProcessingOrder.OldestFirst
+                })
```
rkac-bw, line 18: I think this is inverted?
```
 PermitLimit = int.MaxValue  //Allows more than one concurrent request?
  QueueLimit = 1,              // Only 1 can queue, is this not where you want people to quere?


```

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
rkac-bw: We can have both but should this not return  429 (Too Many Requests) I believe default is 503, should there be a retry? possibly a backoff?

bitwarden_license/test/Scim.IntegrationTest/Controllers/v2/UsersControllerTests.cs
```diff
@@ -596,4 +596,52 @@ public async Task Delete_NotFound()
...
+        // Wait for all requests to complete
+        var contexts = await Task.WhenAll(tasks);
+
+        var successfulResponses = contexts.Where(c => c.Response.StatusCode == StatusCodes.Status201Created).ToList();
+
+        // We should also have at least one successful request
+        Assert.True(successfulResponses.Count > 0, "Expected at least one request to succeed");
+
+        // Total responses should equal our request count
+        Assert.Equal(50, contexts.Length);
```
rkac-bw, line 645: Does this test that seat limit is not exceeded even for this one node?

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
rkac-bw: Are we only interested in solving this problem in cloud, so only possibly need db locking for SQL Server hyperscale? or do we need this for EF too?

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
jrmccannon: Good catch. Corrected this.

bitwarden_license/src/Scim/Utilities/ScimConcurrencyRateLimiter.cs
withinfocus: The transaction scope has a timeout. Not sure if we even configure that beyond some default but it's worth looking into. Remember that we already use transactions for even more high-throughput operations elsewhere. You can catch the timeout or other SQL-related exceptions and turn that into several error responses, and such a thing deserves core-level error handling shared by all API responses honestly.

Unless we are really slamming this thing (which appears unlikely) the transactions will execute in time sequentially and are handled by the DBMS -- this is battle-tested and mission-critical logic that's been in systems like MSSQL for decades and it can be depended upon, plus it's very easy to implement. I accept, and have dealt with in past lives, situations where we need to pivot _away_ from that, but we are nowhere near that state today.

Both Dapper and EF can implement a database transaction with a handful of lines of code.

jrmccannon:
Decided to go a different direction

PR #6173, opened 2025-08-08
[PM-20140] Prevent accidental bulk removal of users without a Master Password

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20140
## 📔 Objective

Directory Connector (and the public api import endpoint generally) can remove users from an organization. For organizations that use Master Password-less decryption (trusted devices or key connector), this has the effect of putting removed user accounts in an unrecoverable state, because they cannot login without being a member (but also cannot re-join without logging in).

This may be desired for general user offboarding, however combined with the `overwriteExisting` option - a debugging option which removes users en masse - it can effectively brick an entire organization's membership (except for Owners). This is very easy to do.

This PR prevents the use of the `overwriteExisting` option if it would remove any user without a master password.

Normal syncs are still permitted to remove these users in order to preserve existing workflows - noting that this will generally remove a few users but is unlikely to remove every user in the organization.

It is behind a feature flag just in case it causes any unforeseen issues with customers' existing workflows. That said, it's within a command which is _already_ behind a feature flag, so ideally we will clean that up first before merging this (or at least before turning it on). The import command refactor is being turned on shortly so that should be OK.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -116,7 +116,7 @@ public static class FeatureFlagKeys
     public const string ImportAsyncRefactor = "pm-22583-refactor-import-async";
     public const string CreateDefaultLocation = "pm-19467-create-default-location";
     public const string MembersGetEndpointOptimization = "pm-23113-optimize-get-members-endpoint";
-    public const string DirectoryConnectorRemoveUsersFix = "pm-24592-directory-connector-prevent-user-removal";
+    public const string DirectoryConnectorPreventUserRemoval = "pm-24592-directory-connector-prevent-user-removal";
 
     /* Auth Team */
     public const string PM9112DeviceApprovalPersistence = "pm-9112-device-approval-persistence";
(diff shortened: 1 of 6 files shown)


src/Core/Constants.cs
```diff
@@ -116,6 +116,7 @@ public static class FeatureFlagKeys
     public const string ImportAsyncRefactor = "pm-22583-refactor-import-async";
     public const string CreateDefaultLocation = "pm-19467-create-default-location";
     public const string MembersGetEndpointOptimization = "pm-23113-optimize-get-members-endpoint";
+    public const string DirectoryConnectorRemoveUsersFix = "pm-20140-directory-connector-remove-users-fix";
```
vincentsalucci: Will you please update the key to `pm-24592-directory-connector-prevent-user-removal`

eliykat: Done!

JimmyVo16 (approved):

PR #6184, opened 2025-08-12
[PM-24192] Move account recovery logic to command

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24192

## 📔 Objective

Move the recover account logic out of `UserService` and into its own command. See ticket for more information.

This is generally a lift & shift, except for the new `CheckProviderPermissionsAsync` private method, and tests (which are all new).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -464,7 +464,7 @@ public class OrganizationUsersController : Controller
     }
 
     [HttpPut("{id}/reset-password")]
-    [Authorize<ManageAccountRecoveryRequirement>]
+    [Authorize<ManageAccountRecoveryRequirement>]   // note: the command has additional permissions checks
     public async Task PutResetPassword(Guid orgId, Guid id, [FromBody] OrganizationUserResetPasswordRequestModel model)
     {
         // Get the users role, since provider users aren't a member of the organization we use the owner check
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -476,7 +476,7 @@ public class OrganizationUsersController : Controller
             throw new NotFoundException();
         }
 
-        var result = await _adminRecoverAccountCommand.AdminResetPasswordAsync(orgUserType.Value, orgId, id, model.NewMasterPasswordHash, model.Key);
+        var result = await _adminRecoverAccountCommand.RecoverAccountAsync(orgUserType.Value, orgId, id, model.NewMasterPasswordHash, model.Key);
         if (result.Succeeded)
         {
             return;
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -11,6 +11,7 @@ using Bit.Api.Vault.AuthorizationHandlers.Collections;
 using Bit.Core;
 using Bit.Core.AdminConsole.Enums;
 using Bit.Core.AdminConsole.Models.Data.Organizations.Policies;
+using Bit.Core.AdminConsole.OrganizationFeatures.AccountRecovery;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.RestoreUser.v1;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies;
@@ -66,6 +67,7 @@ public class OrganizationUsersController : Controller
     private readonly IRestoreOrganizationUserCommand _restoreOrganizationUserCommand;
     private readonly IInitPendingOrganizationCommand _initPendingOrganizationCommand;
     private readonly IRevokeOrganizationUserCommand _revokeOrganizationUserCommand;
+    private readonly IAdminRecoverAccountCommand _adminRecoverAccountCommand;
 
     public OrganizationUsersController(IOrganizationRepository organizationRepository,
         IOrganizationUserRepository organizationUserRepository,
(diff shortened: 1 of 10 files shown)


JimmyVo16 (dismissed):

eliykat:
Reverting to draft while I implement some changes requested by Product. Specifically, we want to alert the user before they submit the dialog, so I intend to add a prevalidate endpoint to check permissions at the time the dialog opens. This avoids additional load on the Members endpoint when the page first loads.

eliykat:
@JimmyVo16 This is ready for review again. I was going to improve the UX by adding an endpoint that would prevalidate the user's permissions when the dialog first opens: see the discussion [here](https://bitwarden.atlassian.net/browse/PM-24192?focusedCommentId=106208). I moved the authorization logic into handlers so it could be reused, however in the end these changes were deprioritized. I'll create a follow up ticket to improve the UX, for now we can just get the work done to date merge, which should still cover the concerns in the original [ticket](https://bitwarden.atlassian.net/browse/PM-24192).

The other main change since your last review is to add a feature flag.

src/Api/AdminConsole/Authorization/RecoverProviderAccountAuthorizationHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+            return;
+        }
+
+        var targetUserProviderUsers =
+            await providerUserRepository.GetManyByUserAsync(targetOrganizationUser.UserId.Value);
+
+        if (targetUserProviderUsers.Any(providerUser => !currentContext.ProviderUser(providerUser.ProviderId)))
+        {
+            var failureReason = new AuthorizationFailureReason(this, "You cannot recover a provider user account.");
+            context.Fail(failureReason);
```
eliykat: Reminder on how authorization handlers work: for an action to be authorized, at least 1 handler must call `Succeed`, and _no_ handler must call `Fail`. So this handler will effectively override the other one in the case of providers, denying the action.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerPutResetPasswordTests.cs
```diff
@@ -0,0 +1,196 @@
...
+                .IsEnabled(FeatureFlagKeys.AccountRecoveryCommand)
+                .Returns(true);
+        });
+        _client = _factory.CreateClient();
+        _loginHelper = new LoginHelper(_factory, _client);
+    }
+
+    public async Task InitializeAsync()
+    {
+        _ownerEmail = $"org-user-reset-password-integration-test-{Guid.NewGuid()}@bitwarden.com";
```
JimmyVo16: Don’t we normally use example.com for test mock-ups?

src/Api/AdminConsole/Authorization/RecoverProviderAccountAuthorizationHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+            // This should be checked as part of the command validation, but it's also required
+            // for this logic to work properly, so we'll fail here if not set.
+            context.Fail();
+            return;
+        }
+
+        var targetUserProviderUsers =
+            await providerUserRepository.GetManyByUserAsync(targetOrganizationUser.UserId.Value);
+
+        if (targetUserProviderUsers.Any(providerUser => !currentContext.ProviderUser(providerUser.ProviderId)))
```
JimmyVo16: Just to confirm my understanding: if this user is a provider in any reseller, then we cannot reset their password?

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -500,6 +510,37 @@ public async Task PutResetPassword(Guid orgId, Guid id, [FromBody] OrganizationU
...
 
+    public async Task PutResetPasswordvNext(Guid orgId, Guid id, [FromBody] OrganizationUserResetPasswordRequestModel model)
+    {
+        var targetOrganizationUser = await _organizationUserRepository.GetByIdAsync(id);
+        if (targetOrganizationUser == null || targetOrganizationUser.OrganizationId != orgId)
+        {
+            throw new NotFoundException();
+        }
+
+        var authorizationResult = await _authorizationService.AuthorizeAsync(User, targetOrganizationUser, new RecoverAccountAuthorizationRequirement());
```
JimmyVo16, line 524: Just thinking out loud: this looks like it could be in a middleware that we can attach to the controller method. That way, it's cleaner and we can add it to multiple places. In some languages, you can even hydrate the request object in the middleware so the data can be passed along to other middlewares.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerPutResetPasswordTests.cs
eliykat: Yes we do and yes we should!

src/Api/AdminConsole/Authorization/RecoverProviderAccountAuthorizationHandler.cs
eliykat: The business rule is: to recover a provider user's account, you must also be a member for the same provider(s).

In effect, this code iterates through the target user's `ProviderUser` associations and calls `currentContext` to make sure the current user is also a member for the same providers.

If the current user is not a member for the same providers, that indicates a privilege escalation so we block it.

eliykat: I've added a comment and some xmldoc on `ICurrentContext` to clarify this a little.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
eliykat: This is definitely doable, but at this stage we have no cause to reuse this elsewhere, and the logic is already encapsulated in the handler. I think it's clearer to make the explicit call. As you've found with your logging case, moving logic into the middleware isn't always very clear to the reader.

That said, Billing Team do have a neat attribute that fetches and injects the organization for the current request: https://github.com/bitwarden/server/blob/main/src/Api/Billing/Controllers/VNext/OrganizationBillingVNextController.cs#L36-L44. We could do something similar to get the orgUser, make sure they exist, and make sure they match the current organization - given that that is a very common but critical bit of boilerplate. I've already fussed this PR a lot though so I'd prefer to leave it for future experimentation.

eliykat:
I will take a final look at Claude's feedback tomorrow, but all major issues have been addressed.

eliykat:
@JimmyVo16 This is ready for review. The main change I've made is to combine the 2 authorization handlers - the logic is very similar, but now it runs in a single handler in sequence, rather than both handlers being executed separately. I ultimately decided this was clearer than 2 handlers, and allows for more descriptive error messages for each failure case.

Claude's latest review failed because we hit our usage limits. You can see the last review in the comment history, but everything there seems unnecessary or wrong - let me know if you disagree.

PR #6186, opened 2025-08-12
[PM-21878] update gateway/stripe fields for business units

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21878

## 📔 Objective

When editing a business unit in the Bitwarden portal, a bug exists that prevents the billing "Gateway Customer ID" and "Gateway Subscription ID" fields from saving. This change fixes the bug by making sure the fields get copied to the save model for business units, not only for MSPs.

## 📸 Screenshots

https://github.com/user-attachments/assets/39238237-9b4c-46cb-b866-8f5825025986

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Admin/AdminConsole/Models/ProviderEditModel.cs
+++ b/src/Admin/AdminConsole/Models/ProviderEditModel.cs
@@ -90,6 +90,7 @@ public class ProviderEditModel : ProviderViewModel, IValidatableObject
         switch (Type)
         {
             case ProviderType.Msp:
+            case ProviderType.BusinessUnit:
                 existingProvider.Gateway = Gateway;
                 existingProvider.GatewayCustomerId = GatewayCustomerId;
                 existingProvider.GatewaySubscriptionId = GatewaySubscriptionId;


amorask-bitwarden (dismissed):
Nice catch

eliykat (dismissed):

src/Admin/AdminConsole/Models/ProviderEditModel.cs
```diff
@@ -90,6 +90,7 @@ public virtual Provider ToProvider(Provider existingProvider)
         switch (Type)
         {
             case ProviderType.Msp:
+            case ProviderType.BusinessUnit:
```
eliykat: 🎨  It looks like you already have a couple of extension methods to express similar logic: https://github.com/bitwarden/server/blob/54e7fac4d99b7a8e3fc83221c561d60c68194494/src/Core/Billing/Extensions/BillingExtensions.cs#L25-L37

Maybe you could also use an extension on the enum type? If we ever add another billable provider type, we don't want to have to find and update all these switch statements.

Particularly because I understand the `Status: Billable` check is/was only for migration, and in the future it should only depend on the type.

kdenney: Good suggestion! I went ahead and added an extension method and a comment explaining why we only send these types to stripe. Thanks for the feedback!

kdenney:
QA found an issue where the edit page was unable to load if you saved an invalid stripe customer id. I have updated this PR with changes to fix this issue. I decided to fix the page load issue to hopefully solve the issue on QA and also at the point of editing a provider to make sure this doesn't happen again.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
```diff
@@ -295,6 +295,22 @@ public async Task<IActionResult> Edit(Guid id, ProviderEditModel model)
...
+                await _stripeAdapter.CustomerGetAsync(model.GatewayCustomerId);
+                await _stripeAdapter.SubscriptionGetAsync(model.GatewaySubscriptionId);
+            }
+            catch (StripeException e)
+            {
+                var oldModel = await GetEditModel(id);
+                ModelState.AddModelError(nameof(model.GatewayCustomerId), $"Stripe Error: {e.Message}");
+                return View(oldModel);
+            }
+        }
```
eliykat: Suggested interface: 

```c#
if (!await _providerBillingService.ValidateGatewayCustomerId(model))
{
  var oldModel = await GetEditModel(id);
  ModelState.AddModelError(nameof(model.GatewayCustomerId), $"Stripe Error: {e.Message}");
  return View(oldModel);
}
```

src/Admin/AdminConsole/Controllers/ProvidersController.cs
```diff
@@ -382,10 +398,18 @@ private async Task<ProviderEditModel> GetEditModel(Guid id)
...
+        try
+        {
+            payByInvoice =
+                _featureService.IsEnabled(FeatureFlagKeys.PM199566_UpdateMSPToChargeAutomatically) &&
+                (await _stripeAdapter.CustomerGetAsync(provider.GatewayCustomerId)).ApprovedToPayByInvoice();
+        }
+        catch (StripeException e)
+        {
+            TempData["Error"] = $"Stripe Error: {e.Message}";
+        }
```
eliykat: Suggested interface:

```c#
var payByInvoice = await _providerBillingService.ApprovedToPayByInvoice(provider);
```

src/Admin/AdminConsole/Controllers/ProvidersController.cs
kdenney: Good suggestion! I decided to create 2 separate methods and added unit tests for them. The only downside is that technically the methods will act like the ids are invalid whenever any stripe exception is thrown. Most likely this would only give a false negative upon a fatal stripe error or similar so probably not a big issue but I am open to feedback!

src/Admin/AdminConsole/Controllers/ProvidersController.cs
kdenney: For this one, I'm going to defer to @amorask-bitwarden - I could create a method that fetches the customer and returns this property but I don't agree that method should contain the try/catch because it would not be obvious from the caller that it is doing so. If the purpose is simply to remove all calls of StripeAdapter from the controller and move those out one layer to a service, then I could create a wrapper method "GetCustomerAsync" in the ProviderBillingService class that simply calls the adapter, but I'm not sure that gains us much. I do agree that this could be cleaned up, though, so I am open to additional suggestions.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
```diff
@@ -295,6 +295,25 @@ public async Task<IActionResult> Edit(Guid id, ProviderEditModel model)
...
+                return View(oldModel);
+            }
+            if (!string.IsNullOrEmpty(model.GatewayCustomerId) &&
+                !await _providerBillingService.IsValidGatewaySubscriptionIdAsync(model.GatewaySubscriptionId))
+            {
+                var oldModel = await GetEditModel(id);
+                ModelState.AddModelError(nameof(model.GatewaySubscriptionId), $"Invalid Gateway Subscription Id: {model.GatewaySubscriptionId}");
+                return View(oldModel);
+            }
+        }
```
eliykat: (non-blocking) This is an improvement but I wonder if the interface could be further abstracted. For example, could it take the whole `model` object and validate whatever properties it needs to, and then return a result which the controller handles?

The current design requires me (in a different team) to know which properties need to be validated by Billing Team and call the appropriate interfaces one-by-one.

I am hammering this point a little but cross-team interfaces is something @amorask-bitwarden and I have been discussing semi-recently.

Again, this is non-blocking so I will leave it to you as to whether you iterate on it further.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
eliykat: I think we can rely on the application wide error handling. [You don't see this in dev mode](https://github.com/bitwarden/server/blob/18d146406c31a3452f8b94ecf51525e7dfb71437/src/Admin/Startup.cs#L151-L158), but if an unhandled error is thrown it is already displayed as a toast (here I am just throwing a random exception for demonstration purposes):

https://github.com/user-attachments/assets/208d6a88-158a-44bd-b5b4-cce1ff7139ef

You could still catch and rethrow if you wanted to throw a nicer error. But then the caller doesn't have to handle it, because Stripe being down really is an exception.

The service could encapsulate the feature flag check and the stripe adapter call.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
eliykat: (non-blocking) Not my area of expertise, but maybe you could inspect the error you get back from stripe? If it's an "invalid customer ID" error (or whatever) then you can return `false`, if it's anything else you can throw. (see my comment below re: application wide error handling)

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
```diff
@@ -881,6 +881,32 @@ await stripeAdapter.SubscriptionUpdateAsync(provider.GatewaySubscriptionId,
         }
     }
 
+    public async Task<bool> IsValidGatewayCustomerIdAsync(string customerId)
```
amorask-bitwarden: I'd like to make a couple improvements here:
1. Let's move both of these methods to the `SubscriberService`. An `ISubscriber` is an interface that represents any entity that could potentially be linked to Stripe. This includes `Provider`, but also `Organization` and `User`. That will make these operations more extensible.
3. When you move to the `SubscriberService`, let's have the method take an `ISubscriber` rather than a `string`, which gives us some free type enforcement without having to invoke the method.
4. This method should handle the `nullOrEmpty` check rather than the invoker.
5. `return true` can be placed immediately after the `StripeAdapter` invocation. 
6. We should match on the specific type of `StripeException` that gets returned. To my utter hatred, Stripe throws a `StripeException` when a resource doesn't exist, but they at least classify it using `stripeException.StripeError.Code == "resource_missing"` (https://docs.stripe.com/error-codes). As such, I'd suggest adding "resource_missing" to `StripeConstants.ErrorCodes` and then instituting a `when` case on the `StripeException` that checks for that code.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
amorask-bitwarden: @eliykat I'm a bit unclear on what's being requested here. I'm fine with moving the billing-level validation of the `ProviderEditModel` to a separate method within the same controller, but I'm hesitant to move this validation to a dedicated method in Core A.) Because we'd have to duplicate the necessary portions of the Edit Model requires maintenance and B.) Because this is the only place in the application it would ever be invoked.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
amorask-bitwarden: If we don't want to have the `StripeAdapter` in controller code, I suggest the following:
1. Replace the `StripeAdapter.CustomerGetAsync` with the `SubscriberService.GetCustomer` invocation. This returns null if the resource is missing.
2. If `Customer` is `null`, then `payByInvoice` is `false`
3. Remove the exception handling as it's not necessary.

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
kdenney: 👍 These changes have been pushed.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
kdenney: > If we don't want to have the `StripeAdapter` in controller code, I suggest the following:
> 
>     1. Replace the `StripeAdapter.CustomerGetAsync` with the `SubscriberService.GetCustomer` invocation. This returns null if the resource is missing.
> 
>     2. If `Customer` is `null`, then `payByInvoice` is `false`
> 
>     3. Remove the exception handling as it's not necessary.

I have pushed a change that makes this change. The code looks much cleaner and removes the dependency on StripeAdapter.

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
kdenney: Note that we decided against moving the "resource_missing" code to StripeError to avoid side effects from the change. We plan to refactor this in the future and at that point can move it. Let me know if I need to create a ticket or similar to track this future refactor.

src/Admin/AdminConsole/Controllers/ProvidersController.cs
eliykat: It doesn't necessarily need to move to `Core`, so maybe `ProviderBillingService` was a poor example. But I think we should move a lot of this logic out of the controller generally; both because there's too much logic in this controller (compare it to our fairly thin controllers in `Api`) and because AC Team is the codeowner of this file (so it doesn't seem like the place for Billing-related validation).

My request is really just to have a more abstracted interface between teams, and for each team to own their own code, however that looks. That is a bigger refactor though.

eliykat (approved):
Thanks for iterating on this!

PR #6192, opened 2025-08-14
Adding transaction around seat increase and user addition

jrmccannon:
## 🎟️ Tracking
[PM-24462](https://bitwarden.atlassian.net/browse/PM-24462)

## 📔 Objective
This wraps the addition of users and increasing of the seat count for an organization in a transaction. It also does a quick check to make sure we can add users to the password manager plan.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-24462]: https://bitwarden.atlassian.net/browse/PM-24462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrmccannon:
@withinfocus Here is the seat increase, user addition and a quick check of if the seats can indeed be increased wrapped in a transaction. I'm still adding some tests and validating that everything is updated correctly. Just wanted to get eyes on it early.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +262,103 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+        await using var connection = new SqlConnection(ConnectionString);
+
+        await using var transaction = await connection.BeginTransactionAsync();
+
+        var organizationResult = await connection.QueryAsync<Organization>(
+            "[dbo].[Organization_ReadById]",
+            new { Id = organizationId },
+            commandType: CommandType.StoredProcedure);
+
+        var organization = organizationResult.SingleOrDefault();
```
withinfocus, line 283: 🎨 Dapper lets you ask for a single directly doesn't it? Do we use that elsewhere?

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +262,103 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+        var occupiedSeatCount = occupiedSeatCountResult.SingleOrDefault() ?? new OrganizationSeatCounts();
+
+        var validatorRequest = new PasswordManagerSubscriptionUpdate(
+            new InviteOrganization(organization, plan),
+            occupiedSeatCount.Total,
+            organizationUsersList.Count);
+
+        var result = InviteUsersPasswordManagerValidator.ValidatePasswordManager(validatorRequest);
+
+        switch (result)
```
withinfocus, line 304: 🎨 Executing a series of queries in a repository is one thing and makes sense, but introduction of logic like this gives me pause and I think should be in a service or command -- the transaction would be created there and the repository stays focused? Thoughts?

This doesn't have to be right now as I see other usages of transactions being right in the repository, but it's a slippery slope.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +262,103 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+                }
+                catch (Exception)
+                {
+                    await transaction.RollbackAsync();
+                    throw;
+                }
+
+                break;
+            case Invalid<PasswordManagerSubscriptionUpdate> invalid:
+                await transaction.RollbackAsync();
```
withinfocus, line 357: ❓ It won't matter as of right now with this only having read data, but do we want to call for a rollback at all?

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +262,103 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+                throw new Exception(invalid.Error.Message);
```
withinfocus, line 358: ❌ We need something better than the base exception class.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +262,103 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+            default:
```
withinfocus, line 359: ❓ Per my other comment, does anything need to happen? Is this actually worthy of throwing? Can't just throw the base exception either.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +262,103 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+        Guid organizationId,
+        DateTime requestDate,
+        Plan plan,
+        IEnumerable<CreateOrganizationUser> organizationUserCollection)
+    {
+        var organizationUsersList = organizationUserCollection.ToList();
+
+        await using var connection = new SqlConnection(ConnectionString);
+
+        await using var transaction = await connection.BeginTransactionAsync();
```
withinfocus, line 276: ℹ️ Speak with @rkac-bw about what isolation level we need to use here. SQL Azure defaults now have uncommitted reads so this won't work.

withinfocus (commented):
I just looked at the Dapper implementation but yep this is the idea! Few comments on improvements though.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
jrmccannon: I did this just initially so we didn't have any floating transactions. I see now there is a `ReleaseAsync()` that could be used...unless it should be disposed once it leaves the using context?

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
jrmccannon: If the logic is elevated to the command, we wouldn't need to do any of this here. The exception was a way of short circuiting the rest of the command logic. I'll elevate the logic to the command and this won't need to be here.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
jrmccannon: I'd rather have it up in the command. I'll investigate opening the transaction and seeing if I have to pass that around since we'd need to execute against the same context/scope ? 🤷

withinfocus: In a prior life of mine passing in the transaction as a method parameter was quite common. It definitely can work, just hasn't been needed here.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
withinfocus: The `using` would let it close automatically and it invokes the release. This feels like a no-op with respect to the transaction so I _think_ you want to not do anything with it.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
jrmccannon: Thinking this through more and playing around with it a bit, we would need to abstract away the transaction from the repositories. This would be a large amount of change because both Dapper and EF use different objects for the database.  

Those objects would have to leave the repository layer in order to be passed into other methods (for the read and then the write).

Unless we would add state to the repository, which could lead to weird side-effects, uninitialized states, etc.

How should I proceed?

1. Add abstraction around transaction and db connection to be universal between the different data stores we support.

2. Add transaction state to the repository and create new methods that would support using that stored transaction.

withinfocus: The repository doesn't need state -- can't you offer method overloads that take a transaction? Use it or not.

jrmccannon: The transaction object is different between EF and Dapper.

I would need to create an abstraction, return that, then use that to pass the connection around.

withinfocus: We are breaking new ground here so this is valuable to work through. I had thought commands were in the infrastructure libraries but that's not the case, although maybe they should be. We don't need to move all the existing ones or anything, and we can call this something other than "command" too if we want, but I think that's where you start this. The commands are already interface-driven, and you'd ask for one with its respective implementation creating a transaction there. Does this not cover everything?

jrmccannon: I'll give that a try and see how it plays out. Thanks!

jrmccannon: So finally getting back to revisiting this.

> commands were in the infrastructure libraries but that's not the case, although maybe they should be. [...] The commands are already interface-driven, and you'd ask for one with its respective implementation creating a transaction there. 

After playing around with this, I have some thoughts. 

Currently, commands contain business logic and sit between the controller and the repositories. The repositories act as the abstraction layer between the server and the database. If we moved Commands into the infrastructure, we would then have to duplicate business logic between the two commands. That would lead to long term maintenance problems as changes would have to be duplicated between each.

For this specific use case, we get the organization's current seat count, get the number of occupied seats an organization currently has, validate that we can indeed invite the new organization user, then add the organization user, increase the seat count if necessary, and send the invite. All of that needs to occur in a transaction because of the `MaxAutoscaleSeats` property on Organization.  If they hit that value, we cannot invite that new user. We also have a side effect of sending out an email to the organization owners when they approach that threshold.

In order for something like this to be done with datastore specific commands, we would have to rewrite all our current logic (I believe).  Unless I'm overthinking this and you are just suggesting taking the current method (`AddUsersToPasswordManagerAsync`) out of the repository and put it in its own infrastructure class that would have access to the SqlConnection/DbContext be able to execute business logic.

In regards to passing around a transaction, the problem we have in the current Repository classes is that every repo method creates its own SqlConnection/DbContext for the database. While this does allow us to call the database in parallel (in EF's case), it stops us from being able to share a connection or transaction across calls that leave the repository level. EF uses `DbContext` and `DbContextTransacation` as its abstractions while Dapper uses the traditional `SqlConnection` and `DbTransaction`.  This stops us from being able to pass them between different repository calls because the interface must be consistent between the two implementations.

Thoughts?

withinfocus: Meh, definitely some problems with our architecture then, and I have never had to solve this with different transaction objects as you're describing.

Once we are in the infrastructure projects, we can just ask for concrete implementations right? Then we can use overloads from other repositories and pass in a transaction, SQL connection, etc.

As an alternative, and I am not able to experiment with this right now, but what if we invented an `ICompositeRepository` or similar, then implemented the logic there? Perhaps it's no better than moving and duplicating logic in the commands that are then in the infrastructure projects, but the boundaries are a bit cleaner-feeling to me.

I'll keep thinking about this.

jrmccannon:
Decided to go different direction

PR #6201, opened 2025-08-15
[PM-24233] Use BulkResourceCreationService in CipherRepository

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24233

## 📔 Objective

This refactor replaces direct `SqlBulkCopy` usage on `CipherRepository` with calls to `BulkResourceCreationService` methods when the `CipherRepositoryBulkResourceCreation` feature flag is enabled.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


JimmyVo16 (dismissed):

JimmyVo16 (approved):

src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
```diff
@@ -108,7 +108,8 @@ public async Task ImportIntoIndividualVaultAsync(
         }
 
         // Create it all
-        await _cipherRepository.CreateAsync(importingUserId, ciphers, newFolders);
+        var useBulkResourceCreationService = _featureService.IsEnabled(FeatureFlagKeys.CipherRepositoryBulkResourceCreation);
```
djsmith85, line 111: ❓  Instead of retrieving the feature flag on every usage of `_cipherRepository`, couldn't the cipherRepository check for the feature flag itself or could we swap out the cipherRepository-implementation when the flag is enabled?

This would ensure the interface for the cipher repository would stay intact and none of the calling code would need to change.

quexten (changes requested):
KM changes look good, but I agree with @djsmith85, I'm a bit confused why the feature flag is not handled inside the cipher repository. Happy to change to an approving review if we have a reason to do so.

src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
r-tome: > couldn't the cipherRepository check for the feature flag itself

That was actually my first approach (you can see it in earlier commits), but after discussing with @eliykat we agreed it goes against our architecture guidelines — feature flags shouldn’t be introduced at the repository layer.
That’s why I switched it to a boolean parameter. I can also split it into separate methods and remove the parameter if that’s preferred.

r-tome:
@quexten @djsmith85 I’ve updated the repository to have `vNext` methods for each case, following our usual pattern. The callers still need to check the feature flag to decide which method to call, but this setup keeps the repository interface clean and consistent.

All of this is only temporary — once the feature flag is removed, the legacy methods and the branching in the callers will be cleaned up quickly.

JimmyVo16 (approved):

djsmith85 (approved):

PR #6221, opened 2025-08-20
[PM-15621] Refactor delete claimed user command

eliykat:
## 🎟️ Tracking

### Primary objective
The initial implementation of `DeleteManagedOrganizationUserCommand` called `UserService` to actually delete the user. However, `UserService` doesn't support bulk deletion, so we wrote our own implementation of bulk delete.

This ticket simplifies the structure of the command so it no longer relies on `UserService` and so that it is generally easier to follow.

@JimmyVo16 did the initial implementation here, and most of the actual business logic is his work. I then tinkered a bunch with the result pattern (see below).

All changes should be feature flagged.

### Secondary objective

We have been experimenting with the result pattern in our commands, returning out a value to represent a success or failure, instead of throwing exceptions when a request is invalid according business logic. (Exceptions can still be thrown for actually exceptional cases.)

I used this ticket to iterate on this, and in particular to simplify the implementation of the current `CommandResult` and `ValidationResult` classes. As this resulted in new implementations that are inconsistent with current use, I've kept those local to the `DeleteClaimedAccountvNext` namespace.

The main goal with these implementations were:
- simple definition
- simple to instantiate and return by the called method
- simple to handle by the caller

This is largely achieved by using a [OneOf](https://github.com/mcintyre321/OneOf/tree/master) object to contain the `Error | TResult` value inside the result object. The benefit of this is:
- terse definition of the result object
- utilities to unwrap the result are all provided already by the `OneOf` library (in particular, `Match` which provides an exhaustive pattern match-like syntax)
- `OneOf` is already in use elsewhere (particularly by Billing Team) so it'll become more familiar in our codebase
- minimal boilerplate caused by generics

This also supports bulk operations: the `ValidationResult` contains the validated request, and there is a `BulkCommandResult` wrapper which also contains the `ID` associated with the request. Each of these match what the caller needs to process the result. Bulk operations are very common in AC Team's code, so I thought this was important to handle out of the box. (In other words: we can't just return `IEnumerable<OneOf<Error, None>>` because it's unclear what error belongs to what request.)

A side effect is that the `Error` class is no longer generic, making it easier to define errors and move them between layers of the application.

Lastly, the controller methods return strongly typed results, following Billing Team's [OrganizationBillingVNextController](https://github.com/bitwarden/server/blob/main/src/Api/Billing/Controllers/VNext/OrganizationBillingVNextController.cs). We could probably use something like their [BaseBillingController](https://github.com/bitwarden/server/blob/main/src/Api/Billing/Controllers/BaseBillingController.cs) to help handle results, but I drew the line here for now.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -525,11 +536,34 @@ public async Task DeleteAccount(Guid orgId, Guid id)
         await _deleteClaimedOrganizationUserAccountCommand.DeleteUserAsync(orgId, id, currentUser.Id);
     }
 
+    private async Task<IResult> DeleteAccountvNext(Guid orgId, Guid id)
+    {
+        var currentUserId = _userService.GetProperUserId(User);
+        if (currentUserId == null)
+        {
+            return TypedResults.NotFound(new ErrorResponseModel("Current user not found."));
```
r-tome: Shouldn't we return `Unauthorized` like you're doing for the Bulk method? I usually think of `NotFound` for things you're trying to retrieve

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedAccountvNext/CommandResult.cs
r-tome, line 1: What about making this be an `abstract record`? Also, sometimes we do want to return something if it succeeds.

eliykat: It's not abstract - the command uses it directly.

I've added a more general use `CommandResult`, which uses `OneOf<Error, T>` to represent the result. The one I've used here is just a specific instance of that, where `T` is the `None` type.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedAccountvNext/ValidationResult.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Creates a failed <see cref="ValidationResult{TRequest}"/> with the specified error.
+    /// </summary>
+    public static ValidationResult<T> Invalid<T>(T request, Error error) => new(request, error);
+
+    /// <summary>
+    /// Extracts successfully validated requests from a sequence of <see cref="ValidationResult{TRequest}"/>.
+    /// </summary>
+    public static List<T> ValidRequests<T>(this IEnumerable<ValidationResult<T>> results) =>
+        results
+            .Where(r => r.Error.IsT1)
```
jrmccannon: What is Error.IsT1 ?

jrmccannon (dismissed):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedAccountvNext/ValidationResult.cs
eliykat: `Error` is a `OneOf` type, where `OneOf` is actually a generic over `OneOf<T0, T1, ... Tn>`. `IsT1` is a helper method to identify the underlying type. Obviously it's not very descriptive which is why I've wrapped it in a more useful interface here.

eliykat:
Further changes:
- added tests
- slight rework to the result implementations, because it's useful to have wrappers around the `OneOf` utility methods that express the purpose - e.g. `IsValid` rather than `IsT0` (particularly for assertions in tests). This required subclassing the `OneOfBase` type, which is a slightly different approach but not a significantly more complex one in my view.

src/Core/Constants.cs
```diff
@@ -129,6 +129,7 @@ public static class FeatureFlagKeys
     public const string DirectoryConnectorPreventUserRemoval = "pm-24592-directory-connector-prevent-user-removal";
     public const string CipherRepositoryBulkResourceCreation = "pm-24951-cipher-repository-bulk-resource-creation-service";
     public const string CollectionVaultRefactor = "pm-25030-resolve-ts-upgrade-errors";
+    public const string DeleteClaimedUserAccountRefactor = "delete-claimed-user-account-refactor";
```
vincentsalucci: Please update this to reflect the key in LD: `pm-25094-refactor-delete-managed-organization-user-command`

PR #6229, opened 2025-08-21
[PM-25182] Improve swagger OperationIDs: Part 1

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25182

## 📔 Objective

This PR continues improving the Swagger specification to make the SDK generated code easier to use. The main focus is on the function names on the SDK side, which are based on the Swagger `OperationID` of each route. 

The `OperationID` needs to be unique for the whole schema, so by default it's derived based on the route's path and method. This creates _very_ unwieldy names on the SDK side:
```rust
// The function for GET {}/organizations/{id}/access-policies/people/potential-grantees
pub async fn organizations_id_access_policies_people_potential_grantees_get(...) {}

// With this line of patches it can be something sensible like
pub async fn get_people_potential_grantees() {}
```

The new `OperationID` strategy is to instead use `ControllerName_FunctionName`. This will be done automatically, but it introduces a couple of limitations for our controllers:
- Can't have multiple overloaded controller functions exposed in the schema. There's only 5 or 6 cases in the entire codebase so I've renamed them to be more descriptive.
- Can't have a function with more than one route exposed in the schema. While I'd say it's generally discouraged to expose the same function under two routes, we have a lot of cases of `DELETE /model` and `POST /model/delete` pointing to the same place, so I've split them and marked one as deprecated:
```cs
[HttpDelete("{id}")]
public async Task Delete(Guid id) {}

[HttpPost("{id}/delete")]
[Obsolete("This endpoint is deprecated. Use DELETE /{id} instead.")]
public async Task DeletePost(Guid id) 
```

To this end, this PR contains the following changes:
- Unified the filter configuration between API and Identity, which used to be duplicated, in a new `SwaggerGenOptionsExt.cs` file.
- `ActionNameOperationFilter.cs`: For each route, we store the `Action name` as an extension. This name is just the name of the function that defines the route, which will help the SDK generate better function names.
- `SwaggerExcludeDocumentFilter.cs` and `SwaggerExcludeAttribute.cs`: These are the attribute to exclude certain routes, and the filter that does so.
- Updated a few controllers to exclude routes and rename functions as needed. I've generally tried to exclude the less specific methods (so keeping DELETE/PUT instead of POST), but let me know if you'd prefer different names/methods.
- I've also included a filter that checks if there are duplicate operationIDs in the project and prints a nice error message if that's the case, but it's disabled until I fix all the conflicts (after this PR I have a couple more for other teams)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/SharedWeb/Swagger/CheckDuplicateOperationIdsDocumentFilter.cs
```diff
@@ -0,0 +1,83 @@
...
+/// but we use controller action names to generate them, which can lead to duplicates if a Controller function
+/// has multiple HTTP methods or if a Controller has overloaded functions.
+/// </summary>
+public class CheckDuplicateOperationIdsDocumentFilter(bool printDuplicates = true) : IDocumentFilter
+{
+    public bool PrintDuplicates { get; } = printDuplicates;
+
+    public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
+    {
+        var operationIdMap = new Dictionary<string, List<(string path, OpenApiPathItem pathItem, OperationType method, OpenApiOperation operation)>>();
```
justindbaur: nit of all nits but I would define the items in the tuple with `PascalCase` names. You only reference the items from deconstructing it though so it matters even less to me in this case though.

src/SharedWeb/Swagger/CheckDuplicateOperationIdsDocumentFilter.cs
```diff
@@ -0,0 +1,83 @@
+﻿#nullable enable
```
justindbaur: This shouldn't be needed now that it's enabled by default in the whole repo.

src/SharedWeb/Swagger/SwaggerExcludeAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+    public string HttpMethod { get; } = httpMethod.ToUpper();
+    public string? Path { get; } = path;
+
+    public bool Matches(string httpMethod, string? path)
+    {
+        if (!httpMethod.Equals(HttpMethod, StringComparison.CurrentCultureIgnoreCase)) return false;
+
+        if (Path == null) return true;
+
+        return path?.Contains(Path, StringComparison.CurrentCultureIgnoreCase) ?? false;
```
justindbaur: I really can't imagine when this would matter but this is more inline with other things that compare http method. https://github.com/dotnet/runtime/blob/ec2b401d9b651e3aca70f8619a01fe866a35b2ca/src/libraries/System.Net.Http/src/System/Net/Http/HttpMethod.cs#L57

```suggestion
        if (!httpMethod.Equals(HttpMethod, StringComparison.OrdinalIgnoreCase)) return false;

        if (Path == null) return true;

        return path?.Contains(Path, StringComparison.OrdinalIgnoreCase) ?? false;
```

src/SharedWeb/Swagger/SwaggerExcludeDocumentFilter.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
+    {
+        var operationsToRemove = new List<(string pathKey, OperationType operationType)>();
+
+        foreach (var apiDesc in context.ApiDescriptions)
+        {
+            // Check if the descriptor is a controller action and it has the SwaggerExclude attribute
+            if (apiDesc.ActionDescriptor is not ControllerActionDescriptor actionDesc) continue;
+            var hideAttributes = actionDesc.MethodInfo.GetCustomAttributes(typeof(SwaggerExcludeAttribute), true) as SwaggerExcludeAttribute[];
```
justindbaur: I _think_ there is a generic version of this.

```suggestion
            var hideAttributes = actionDesc.MethodInfo.GetCustomAttributes<SwaggerExcludeAttribute>(true);
```

justindbaur (commented):
Few small things.

src/SharedWeb/Swagger/SwaggerExcludeDocumentFilter.cs
dani-garcia: Seems that while `Assembly`, `MemberInfo` and `ParameterInfo` do have a generic function like you mention, the type `MethodInfo` that I'm using here only has the non-generic variant, yay for consistency 😭 

```csharp
    public abstract partial class MemberInfo : ICustomAttributeProvider
    {
        ...
 
        public abstract object[] GetCustomAttributes(bool inherit);
        public abstract object[] GetCustomAttributes(Type attributeType, bool inherit);
```

justindbaur: `MethodInfo` implements `MemberInfo` so it should be available and I'm able to use it [in this](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMA6AGQEsoBHAbj0JPICUBTAMwBt2BjYFQD2URrgCQkyUwDMJdCwDseAN55xGdOrUTxANwCGYIgFt2wABZC4ASSichRALxFgATwAO7IZwAUABXYwAGcRAEoyAHFzAFlzKzhfACJo4AA5AzMksIBCMUlDYz4IYOAhEwBBYGAwKgAjCGB2YOdTeOs7ByjzAGESssrq2oam4IAeGLcqmvrG9gA+XxqIdjD88XUNTABOXwASJIAxIWg4IgMh2dGQIhVi0vLp4bngsh6TqGBfMIBfbPyfnhAbhZPIiJMnld2EQbpCRuxVHggUj8HIMERAiERIiJABtSYAXXUqDRmGIqQyZm+2k2qEURCSSQBQKAA==).

dani-garcia: Weird, I tried that but it gives me an error here, not sure what I'm missing:
image

dani-garcia: Oh woops, you need the `using System.Reflection;`, the IDE was not suggesting it to me for reasons, probably because the name overload.

justindbaur: Oh yeah, it's an extension method, yay for discoverability of those 😢.

src/SharedWeb/Swagger/CheckDuplicateOperationIdsDocumentFilter.cs
dani-garcia: Nice, wasn't aware that tuple elements followed PascalCase, updated!

dani-garcia:
I've updated the code here with the new pattern that we talked about in https://github.com/bitwarden/contributing-docs/pull/657. This makes the changes a bit more verbose in the controllers themselves, but removes a whole filter + decorator.

justindbaur (approved):

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -654,6 +671,13 @@ public async Task SetUserVerifyDevicesAsync([FromBody] SetVerifyDevicesRequestMo
         await _userService.SaveUserAsync(user);
     }
 
+    [HttpPost("verify-devices")]
+    [Obsolete("This endpoint is deprecated. Use PUT /verify-devices instead.")]
+    public async Task PostSetUserVerifyDevicesAsync([FromBody] SetVerifyDevicesRequestModel request)
+    {
+        await SetUserVerifyDevicesAsync(request);
+    }
+
```
JaredSnider-Bitwarden, line 680: I checked through the other changes and they should all be good, but we use POST on the clients for this method so I've created https://bitwarden.atlassian.net/browse/PM-25376 to have us swap over to the PUT.

JaredSnider-Bitwarden (approved):
Auth files look good to me!

PR #6230, opened 2025-08-21
[PM-23174] ManageAccountRecovery Permission Forces Master Password Set

enmande:
## 🎟️ Tracking

[PM-23174](https://bitwarden.atlassian.net/browse/PM-23174)

## 📔 Objective

Any Trusted Device Encryption SSO Organization User should be forced to set a Master Password if they have been granted the ManageAccountRecovery permission, either opaquely through Owner and Admin roles, or transparently as a Custom user role with the "Manage Account Recovery" permission.

This should be driven specifically by the _permission_ itself as applied to the User through Organization settings.

## 📸 Screenshots

Multiple permutations of invite have been developer-tested, including the following. Two particular cases have been recorded here for illustration.

### TDE-Encryption Org (SSO/single org/account recovery), User invited with Custom role and Manage Account Recovery Permission

https://github.com/user-attachments/assets/07ca2a0a-254f-4ca6-af76-f6b01d51b13d

### TDE-Encryption Org (SSO/single org/account recovery), User invited as Admin

https://github.com/user-attachments/assets/5ec3d003-5a00-4d99-b1dd-39f5819beab4

### TDE-Encryption Org (SSO/single org/account recovery), User invited as Owner

https://github.com/user-attachments/assets/69bed495-784a-4f84-b9e0-029eb1fa2bfc

### TDE-Encryption Org (SSO/single org/account recovery), User invited as User (no extra permissions)

https://github.com/user-attachments/assets/7a4db247-4c6e-4000-8d18-7e201cb2ae10

### TDE-Encryption Org (SSO/single org/account recovery), _existing_ User with MP invited as Admin

📓 User is not required to change their MP upon acceptance to the org. 

https://github.com/user-attachments/assets/50fdf47f-a226-4d14-b451-392361745671

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines
ery

[PM-23174]: https://bitwarden.atlassian.net/browse/PM-23174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


enmande:
There remains one case in which there is some potential disconnect in the user seeing full application of the _Manage Account Recovery_ permission.

If the user is JIT-provisioned into a TDE-Encryption Organization (via SSO), then subsequently _upgraded_ to be an account recovery manager (via Admin or Owner role, or as Custom user with _Manage Account Recovery_ permission), the user is not forced to set a Master Password until their next login. If the user is logged in on their trusted device when the upgrade happens, nothing reactive happens to force their logout, or to notify them during the active session that this permission needs applied. In a TDE scenario, the session will remain active as long as the tab remains open unless otherwise configured via policy.

Subsequent sync operations will update the `ManageAccountRecovery` permission in the user's _claims_ (seen through user tokens retrieved), but it is not a full sync and does not update the client-cached `UserDecryptionOptions` which drives the business logic to enforce MP set. Could use a gut check on what an edge case this is (or isn't) before proposing we alter more wide-reaching sync behaviors or re-imagining the logic to work off of token claims, neither of which may be in scope of this ticket as imagined, and may need a separate piece of work split to consider wider-reaching refactor.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
```diff
@@ -120,26 +124,28 @@ private async Task BuildTrustedDeviceOptions()
...
-        // If sso configuration data is not null then I know for sure that ssoConfiguration isn't null
-        var organizationUser = await _organizationUserRepository.GetByOrganizationAsync(_ssoConfig.OrganizationId, _user.Id);
+        // SSO users, Organizations, and Providers with Manage Account Recovery (reset password) permission,
+        // as well as all Owners and Admins, must set a master password.
+        var hasManageResetPasswordPermission =
+            organizationUser?.GetPermissions() is { ManageResetPassword: true } ||
+            organizationUser?.Type is OrganizationUserType.Admin or OrganizationUserType.Owner ||
+            // The Organization User alone isn't enough to verify for Provider User case;
+            // // check the current context for permissions granted through this relationship.
+            await _currentContext.ManageResetPassword(_ssoConfig!.OrganizationId);
```
enmande: This is to handle _Provider User_ cases, which aren't considered in scope of the Organization User alone. I think from reviewing how this pulls out of `ICurrentContext`, this is a safe and properly-scoped check (and existed before); my chief concern is that this doesn't cast too-wide a net.

Notably, it was required to pass the existing unit test [seen here](https://github.com/bitwarden/server/blob/747e212b1ba374e4c1be9cfe255bb40c002d0ceb/test/Identity.IntegrationTest/Endpoints/IdentityServerSsoTests.cs#L430) for SSO/TDE Provider User with ManageResetPassword. As I'm still getting up to speed on _Provider_ relationships, would appreciate a gut check that this is still the right test, and is being addressed as intended.

test/Identity.Test/AutoFixture/OrganizationUserFixtures.cs
```diff
@@ -0,0 +1,26 @@
...
+using Bit.Core.Utilities;
+
+namespace Bit.Identity.Test.AutoFixture;
+
+internal class OrganizationUserWithDefaultPermissionsCustomization : ICustomization
+{
+    public void Customize(IFixture fixture)
+    {
+        fixture.Customize<OrganizationUser>(composer => composer
+            // On OrganizationUser, Permissions can be JSON data (as string) or sometimes null.
```
enmande, line 15: Particularly interested in your thoughts here, @ike-kottlowski if you don't mind. I think this is a reasonable way to accommodate `BitAutoData` needing to be nudged in the direction of "this can be null or a JSON blob, but we don't have any cases where this should be an invalid JSON string, and should be considered exceptional if it is." I didn't see any similar needs around Permissions in the `UserDecryptionOptions` space, so tried to follow similar patterns from nearby areas.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
```diff
@@ -120,26 +124,28 @@ private async Task BuildTrustedDeviceOptions()
...
         {
             var allDevices = await _deviceRepository.GetManyByUserIdAsync(_user.Id);
-            // Checks if the current user has any devices that are capable of approving login with device requests except for
-            // their current device.
-            // NOTE: this doesn't check for if the users have configured the devices to be capable of approving requests as that is a client side setting.
-            hasLoginApprovingDevice = allDevices.Any(d => d.Identifier != _device.Identifier && _loginApprovingClientTypes.TypesThatCanApprove.Contains(DeviceTypes.ToClientType(d.Type)));
+            // Checks if the current user has any devices that are capable of approving login with device requests
+            // except for their current device.
+            // NOTE: this doesn't check for if the users have configured the devices to be capable
+            // of approving requests, as that is a client-side setting.
```
JaredSnider-Bitwarden: Aside: we should remove this comment. We removed the desktop setting a while back which allowed the desktop to opt in to approving auth requests - it is now on by default. We've also added the web as an approving client, and we are in the process of adding the extension.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
JaredSnider-Bitwarden: This looks great to me! Thank you for documenting the provider use case.

test/Identity.Test/IdentityServer/UserDecryptionOptionsBuilderTests.cs
```diff
@@ -220,19 +222,28 @@ public async Task Build_WhenHasLoginApprovingDevice_ShouldApprovingDeviceFalse(
         Assert.False(result.TrustedDeviceOption?.HasLoginApprovingDevice);
     }
 
-    [Theory, BitAutoData]
+    [Theory]
+    [BitAutoData(OrganizationUserType.User)]
```
JaredSnider-Bitwarden: Note: this is an interesting situation where the test mimics a scenario which can't happen today. Only admins / owners or custom role users with the reset password permission should have that permission set, but this test confirms that, for any role that we have, if they obtain the permission, it will properly send the `HasManageResetPasswordPermission` flag down to the client so I think it is fine.

JaredSnider-Bitwarden:
> There remains one case in which there is some potential disconnect in the user seeing full application of the _Manage Account Recovery_ permission.
> 
> If the user is JIT-provisioned into a TDE-Encryption Organization (via SSO), then subsequently _upgraded_ to be an account recovery manager (via Admin or Owner role, or as Custom user with _Manage Account Recovery_ permission), the user is not forced to set a Master Password until their next login. If the user is logged in on their trusted device when the upgrade happens, nothing reactive happens to force their logout, or to notify them during the active session that this permission needs applied. In a TDE scenario, the session will remain active as long as the tab remains open unless otherwise configured via policy.
> 
> Subsequent sync operations will update the `ManageAccountRecovery` permission in the user's _claims_ (seen through user tokens retrieved), but it is not a full sync and does not update the client-cached `UserDecryptionOptions` which drives the business logic to enforce MP set. Could use a gut check on what an edge case this is (or isn't) before proposing we alter more wide-reaching sync behaviors or re-imagining the logic to work off of token claims, neither of which may be in scope of this ticket as imagined, and may need a separate piece of work split to consider wider-reaching refactor.

@enmande , I think we should be seeing an upgraded TDE user be forced to set a new MP upon the next sync based on this code here: https://github.com/bitwarden/clients/blob/main/libs/common/src/platform/sync/default-sync.service.ts#L267-L312. Can you investigate to see if that code isn't working as intended?

test/Identity.Test/AutoFixture/OrganizationUserFixtures.cs
ike-kottlowski: `BitAutodata` can be a little annoying when we have `string`s which are actually `JSON` I think this is a good approach, good job.

I would have brute forced it and just modified the object in the test class, but I think this is more elegant and actually creates a valid object instead of `"permissions_GUID"`.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
enmande: Thanks! Removed in 32b60f4cfa0cdcc54fc553dd6bfbe2a98a471be5

enmande:
> > There remains one case in which there is some potential disconnect in the user seeing full application of the _Manage Account Recovery_ permission.
> > If the user is JIT-provisioned into a TDE-Encryption Organization (via SSO), then subsequently _upgraded_ to be an account recovery manager (via Admin or Owner role, or as Custom user with _Manage Account Recovery_ permission), the user is not forced to set a Master Password until their next login. If the user is logged in on their trusted device when the upgrade happens, nothing reactive happens to force their logout, or to notify them during the active session that this permission needs applied. In a TDE scenario, the session will remain active as long as the tab remains open unless otherwise configured via policy.
> > Subsequent sync operations will update the `ManageAccountRecovery` permission in the user's _claims_ (seen through user tokens retrieved), but it is not a full sync and does not update the client-cached `UserDecryptionOptions` which drives the business logic to enforce MP set. Could use a gut check on what an edge case this is (or isn't) before proposing we alter more wide-reaching sync behaviors or re-imagining the logic to work off of token claims, neither of which may be in scope of this ticket as imagined, and may need a separate piece of work split to consider wider-reaching refactor.
> 
> @enmande , I think we should be seeing an upgraded TDE user be forced to set a new MP upon the next sync based on this code here: https://github.com/bitwarden/clients/blob/main/libs/common/src/platform/sync/default-sync.service.ts#L267-L312. Can you investigate to see if that code isn't working as intended?

**EDIT**: Addressed in [#16361](https://github.com/bitwarden/clients/pull/16361)

@JaredSnider-Bitwarden Thank you! Upon review, I am not confident that is currently solving for this case. In the referenced `sync` method, the underlying `UserDecryptionOptions` are pulled from state. While there is [a decryption options sync](https://github.com/bitwarden/clients/blob/15619c626547fb3cb38cf9c114910ce079d47e7e/libs/common/src/platform/sync/default-sync.service.ts#L172) immediately before the profile sync, it is not a full sync of user decryption options; it only syncs the [masterPasswordUnlock portion](https://github.com/bitwarden/clients/blob/main/libs/common/src/key-management/models/response/user-decryption.response.ts#L4) of decryption options. We can see that corresponding response model [in the server](https://github.com/bitwarden/server/blob/d0778a8a7b84f08934f12feee2938296973f5ff2/src/Core/KeyManagement/Models/Response/UserDecryptionResponseModel.cs#L3).

I was able to validate this with the desktop client by: 
1. Log in as a regular, confirmed user (with no roles or permissions) to a TDE-encrypted Organization on Desktop
2. Org Admin upgrades logged-in user to Custom role with _ManagePasswordReset_ permission on the Web client
3. As user on Desktop, perform File/"Sync vault"
4. Navigate the app via guarded routes.

User is not logged out or forced to set Master Password. The sync method only updated the masterPasswordUnlock option; the _ManageResetPassword_ information is not updated until (next) login, and continues to be pulled from state. The profile sync did update the organization permission, but the `UserDecryptionOptions` appear to be pulling from separate, targeted cached data in the client as provided by the server at login time, not by interpreting from permissions on the client or fully refreshing decryption options for TDE cases during the sync.

PM-23174__vault-sync

In any case, the user is forced to set a Master Password and all ends are tied up on next login, and the eventual consistency is on point. So it's not a matter of it properly applying, there's just a potential visibility gap for the TDE user _if they are currently logged in when they are upgraded_, which might be an exceptionally small case.

Patrick-Pimentel-Bitwarden:
I'm seeing 3 auth reviewers so I'm going to take myself off, if someone would like me to be on this please add me back.

JaredSnider-Bitwarden (approved):
Nice work, and thank you for the investigation.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
```diff
@@ -120,26 +124,26 @@ private async Task BuildTrustedDeviceOptions()
...
-        // If sso configuration data is not null then I know for sure that ssoConfiguration isn't null
-        var organizationUser = await _organizationUserRepository.GetByOrganizationAsync(_ssoConfig.OrganizationId, _user.Id);
+        // SSO users, Organizations, and Providers with Manage Account Recovery (reset password) permission,
+        // as well as all Owners and Admins, must set a master password.
+        var hasManageResetPasswordPermission =
+            organizationUser?.GetPermissions() is { ManageResetPassword: true } ||
+            organizationUser?.Type is OrganizationUserType.Admin or OrganizationUserType.Owner ||
+            // The Organization User alone isn't enough to verify for Provider User case;
+            // // check the current context for permissions granted through this relationship.
+            await _currentContext.ManageResetPassword(_ssoConfig!.OrganizationId);
```
eliykat: `currentContext.ManageResetPassword` includes these `organizationUser` permission checks, so they are unnecessary. This is equivalent to:

```suggestion
        var hasManageResetPasswordPermission = await _currentContext.ManageResetPassword(_ssoConfig!.OrganizationId);
```

But there's also danger with using `CurrentContext` here. `CurrentContext` uses JWT claims for the currently authenticated user, whereas this class uses a builder pattern to build user decryption options for _any_ user. There is no guarantee that the `User` we're building for matches the `CurrentContext` user.

(Perhaps this doesn't matter in practice - maybe this is only ever used for the currently logged in user - but the current design makes this defect possible.)

So you could take this in a few directions - off the top of my head:
1. go all in on relying on `CurrentContext`; use the code snippet above, and ideally fetch the user by getting `CurrentContext.User` to ensure they match
2. don't rely on `CurrentContext` at all; check the repository directly for provider relationships, using the `User` specified. We can provide a better interface here, but the quick-and-dirty implementation could mirror what [CurrentContext](https://github.com/bitwarden/server/blob/fa0c9cb38785a867cc3b12e5678051004ba6362b/src/Core/Context/CurrentContext.cs#L547) does under the hood today.
3. you could pass in provider information as part of the builder pattern, pushing it up a layer to the caller. How the caller should handle it remains an open question.

I have no opinion on which is best :)

eliykat (commented):
This is non-blocking feedback; just some AC perspective because @JaredSnider-Bitwarden previously made the mistake of asking my opinion ;)

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
enmande: Thank you for the feedback!

The behavior I am seeing when developing this is in line with the most current comments, which were updated since your suggestion to add some clarity that I think I had not fully represented initially. The repository fetches invited organization users with permissions, and we can’t get that from context as the claims aren’t present for those users first walking in. I’m happy to provide some examples from what I’m observing at debug time to rule out a misunderstanding on my part here.

Without a targeted query to interrogate provider user relationships, we need to rely on the current context for surfacing those today. Upon review with several folks, my impression was that adding code akin to what CurrentContext provides for this specific case would perhaps be adding more logic to adjust or deprecate when we get to PM-25668 since we have been able to capture that in its scope.

PM-25688 should help us solve for all these cases more elegantly while eliminating some of what we all see as technical debt right now. Having reviewed PM-25688 and the updated comments, would you agree that it’s most practical to move forward with these changes targeting the ManageResetPassword permission, and use this as incentive to prioritize PM-25688 and improve the overall story with additional, consistent impact?

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
```diff
@@ -120,26 +124,32 @@ private async Task BuildTrustedDeviceOptions()
...
+        // SSO users, Organizations, and Providers with Manage Account Recovery (reset password) permission,
+        // as well as all Owners and Admins, must set a Master Password.
+        var hasManageResetPasswordPermission =
+            // We need to interrogate the organizationUser from the repository to solve for an invited status.
+            // The claims won't exist at that time, so currentContext as it is written today won't accommodate for that.
+            // This is the edge case.
+            // TODO: PM-25668 proposes some refactor of the Organization acceptance processes to unify requirement for
+            // Organization acceptance; once it is implemented, we should be able to remove this Organization User check.
+            organizationUser?.GetPermissions() is { ManageResetPassword: true } ||
+            organizationUser?.Type is OrganizationUserType.Admin or OrganizationUserType.Owner ||
```
eliykat: This repository call will not get users in the Invited status, so this comment doesn't match the code. If the comment is wrong, then it needs to be fixed. If the code is wrong, we probably need to have more discussions around the intention here, as I have security concerns about forcing invited users to change their password.

Additionally, this repository call _will_ include users in the Revoked status, which I suggest is inappropriate because the Revoked status means that you can't access organization resources or perform administrative functions.

This leaves the Accepted and Confirmed statuses only. Here's an example of how you could query that (from our Slack conversation):

```c#
        var organizationUser =
            await _organizationUserRepository.GetByOrganizationAsync(_ssoConfig.OrganizationId, _user.Id);
        if (organizationUser == null)
        {
            throw new BadRequestException("Unable to find organization user");
        }

        // We want to include users in the accepted and confirmed status, so we have to go to the database
        // rather than CurrentContext. Note: revoked users are intentionally excluded.
        // TODO: PM-25668 proposes some refactor of the Organization acceptance processes to unify requirement for
        // Organization acceptance; once it is implemented, we should be able to remove this Organization User check.
        var hasMemberResetPasswordPermission =
            organizationUser.Status is OrganizationUserStatusType.Accepted or OrganizationUserStatusType.Confirmed &&
            (organizationUser.Type is OrganizationUserType.Owner or OrganizationUserType.Admin ||
             organizationUser.GetPermissions() is { ManageResetPassword: true });
```

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
```diff
@@ -120,26 +124,32 @@ private async Task BuildTrustedDeviceOptions()
...
+            // For Organization User in Accepted/Confirmed status, claims will be available. CurrentContext will also
+            // offer Provider User information for that case. As MSP, Provider Users will require certain permissions like
+            // ManageResetPassword to perform their role effectively.
+            await _currentContext.ManageResetPassword(_ssoConfig!.OrganizationId);
```
eliykat: This comment also doesn't match the code. `CurrentContext` only contains users in the Accepted status. But also that probably doesn't matter, because your previous db query will get those. At the very least this comment needs to be updated for accuracy.

However, given that this is really only here to check for the provider status, you can use a more precise `CurrentContext` method to check this:

```c#
await _currentContext.ProviderUserForOrgAsync(_ssoConfig!.OrganizationId);
```

This more clearly separates member and provider checks.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
eliykat: We can kick the larger architectural concerns down the road to the refactor, that's OK.

However there are still 2 main issues I see for this PR:
1. `CurrentContext` uses the current user, but this builder class takes an arbitrary `User`. Are you sure these will always be the same user? If the answer is "no" then it will not work as intended.
2. The comments incorrectly describe the behavior of the code. Are the comments incorrect, or is the code incorrect? I don't know the answer to this one. I'll leave a review shortly explaining this in more detail. (largely per our Slack conversations but it may be more useful to have it inline with the code and visible to the rest of the team). The answer here might just be to adjust the comments, but it's also about making sure that your code does what you think it does.

eliykat (commented):
This is a more targeted review just focusing on the current implementation given that you have a refactor ticket for later.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
enmande: > This repository call will not get users in the invited status, so this comment doesn’t match the code.

My experience with reading the code and validating its behavior has been different throughout this work. This permission is relevant for [TDE organizations with SSO](https://github.com/bitwarden/server/blob/3ac3b8c8d984eaf5fee59be35bb2bd4cb0c8ac61/src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs#L111) only. We notice in [Bit.Sso.Controller.AccountController.AutoProvisionUserAsync](https://github.com/bitwarden/server/blob/3ac3b8c8d984eaf5fee59be35bb2bd4cb0c8ac61/bitwarden_license/src/Sso/Controllers/AccountController.cs#L559), which is invoked [on SSO callback](https://github.com/bitwarden/server/blob/3ac3b8c8d984eaf5fee59be35bb2bd4cb0c8ac61/bitwarden_license/src/Sso/Controllers/AccountController.cs#L261), that the user is either just-in-time provisioned or linked at callback time. This seamlessly creates the relationship between user and organization user (with Id) that allows us to get the full picture from the repository at this time. Upon review, I can see in other flows how the repository might not be able to surface invited users due to timing of provisioning, therefore we might not expect them to be returned here without the callback provisioning context. Neither [the repository call](https://github.com/bitwarden/server/blob/d384c0cfe60ec02226479d3cde200e4d785a7e50/src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationUserRepository.cs#L124) nor [the sproc it invokes](https://github.com/bitwarden/server/blob/d384c0cfe60ec02226479d3cde200e4d785a7e50/src/Sql/dbo/Stored%20Procedures/OrganizationUser_ReadByOrganizationIdUserId.sql#L8) are opinionated in what returns, and there is no uncovered scenario for the purpose of this permission evaluation where that relationship is not fully recorded in the database. It would be great if these timings were more similar for reasoning about the various flows, but TDE is special.

For further illustration, I have included a debug-time video here of the behavior:


https://github.com/user-attachments/assets/fcf6b35c-8748-4768-9697-09b10e3c0f88


> I have security concerns about forcing invited users to change their password.

Understood. Users with ManageResetPassword will (and should) be required to _set_ an MP if they did not have one in order to support the requirements of the ManageResetPassword function; it requires specific keys. From what I can see and reproduce, users in situations to _change_ an MP would be those who are coerced to do so as part being invited to an organization with a Master Password policy (i.e., an existing user’s MP is of insufficient complexity), which appears somewhat separate. I don't mean to nitpick the language -- the concern may persist, and/or I may still misunderstand, and I just want to build toward better understanding. If either is true, any information on why would help with understanding and whether we need additional work to discuss that.

> Additionally, this repository call _will_ include users in the Revoked status

Fantastic catch, thank you. That is a consideration that needs accounted for. The most recent update solves for this with an inclusive check for invited, accepted, and confirmed.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
enmande: 1. Yes, `CurrentContext` user and the arbitrary user provided to the builder will always be the same _at time of writing_. We notice [here](https://github.com/bitwarden/server/blob/3ac3b8c8d984eaf5fee59be35bb2bd4cb0c8ac61/src/Core/Context/CurrentContext.cs#L120) where the context is consistently set to the claims principal user, and each path to the builder’s user can be followed back here. That said, a change in that behavior, which is outside the scope of this builder, could have unintended side effects, and being aware of all these cases at design time is unnecessarily complex. Your point about separation of concerns is salient, and the most recent update solves for that in the case of organization users by relying on the repository to close that gap. Thank you for the suggestion to use `CurrentContext` to solve for the provider user relationship.

2. I am not sure I am able to agree with this. Thorough testing, investigation of assumptions, and provided videos at checkin time support the claims. If there is more information that I still have not uncovered in these updates, I would be happy to know it. I have updated the video collection in the PR overview to help further illustrate with latest changes, as well as reworked some comments in the latest update. I will cover in more detail with the other thread about the why, especially as pertaining to the use of the repository. This permission is relevant [only for Trusted Device Encryption orgs with SSO](https://github.com/bitwarden/server/blob/3ac3b8c8d984eaf5fee59be35bb2bd4cb0c8ac61/src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs#L111), which has specific concerns and behaviors especially around timing, and I could maybe have reiterated that in the comments because TDE is a special thing. Delving deeper into the why helps us build a better shared understanding.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
enmande: Thank you for the suggestion to use `ProviderUserForOrgAsync`. It solves more cleanly than a broader interrogation of the current context.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
eliykat: Sounds good - I think we've covered this thread. I'll respond to the remaining point below.

src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs
eliykat: Thanks for all the references and video - I appreciate the in-depth response.

The SSO flow is unique in how it sets up the OrganizationUser object, and (as you identify) it is an outlier in this respect. We recently discussed this [in Slack](https://bitwarden.slack.com/archives/C04MF5S8XKN/p1758058643302329). No other flow does (or should) link an invited OrganizationUser by UserId, and indeed we want to change this flow so that it doesn't either. But... those are future changes, and again this is something that would be addressed by your refactor ticket. So that's fine; just be aware that it is not guaranteed to pick up invited users _generally_ (implicitly, only those created via that specific code path as it is currently written).

It might be worth adding a comment like, "This will include invited users created via SSO JIT because they are linked by UserId, it may not include users invited via other flows" - as all of this is quite hard to discover. (as you are experiencing first hand)

The refactor ticket you created will address a lot of this, so I would love to see that picked up soon after this work is done.

Re: the security implications - I am not well versed enough in these flows to know if there is really a security issue here. I just wanted to point out that it is rare - and potentially dangerous - to impose organization requirements on users in the invited status, because anyone can invite any user into an organization and thereby affect their account. However I will defer to the team and @JaredSnider-Bitwarden on this as the domain experts as to whether it is actually an issue in this current flow.

eliykat (commented):
Thanks for discussing and working through my feedback. My remaining comments above are non-blocking, so my feedback has been satisfied. 👍

PR #6239, opened 2025-08-22
Improve Swagger OperationIDs for Tools

dani-garcia:
## 📔 Objective

This PR is a continuation of the work started in https://github.com/bitwarden/server/pull/6229, with some of the team specific changes split out for easier review/merge. You can read that PR for a more complete explanation of the purpose, but in short:
- The SDK generates bindings from the Swagger specification, which creates function names in the format `HTTP path + HTTP method`. This leads to very unwieldy names.
- Our goal is to instead generate function names that match the functions in the controller, which means we can't have duplicate function names, and we can have more than one route per function exposed in the schema.

To fullfill those two goals, this PR does two things:
- If there are two functions with the same name in a controller, I've renamed one of them to be more descriptive.
- If a function has two route attributes, I've used the new `[SwaggerExclude("")]` to exclude one of them. In general I tried to exclude the less precise HTTP method.

The final result in the SDK looks like this:
```rust
// The current function for GET {}/organizations/{id}/access-policies/people/potential-grantees
pub async fn organizations_id_access_policies_people_potential_grantees_get(...) {}

// Once all these PRs are merged, the name will be
pub async fn get_people_potential_grantees() {}
```

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Tools/Controllers/ImportCiphersController.cs
+++ b/src/Api/Tools/Controllers/ImportCiphersController.cs
@@ -63,7 +63,7 @@ public class ImportCiphersController : Controller
     }
 
     [HttpPost("import-organization")]
-    public async Task PostImport([FromQuery] string organizationId,
+    public async Task PostImportOrganization([FromQuery] string organizationId,
         [FromBody] ImportOrganizationCiphersRequestModel model)
     {
         if (!_globalSettings.SelfHosted &&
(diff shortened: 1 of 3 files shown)


djsmith85 (approved):
@dani-garcia Nice improvement, and inline with how we have renamed the commands to distinguish them better instead of using method overloads.

PR #6244, opened 2025-08-25
[PM-19659] Clean up Notifications code

justindbaur:
## 🎟️ Tracking

Part of: https://bitwarden.atlassian.net/browse/PM-19659

## 📔 Objective

A code cleanup of the server push notifications.

- Rename all engines to have the `PushEngine` suffix instead of `PushNotificationService` since they now implement the `IPushEngine` interface NOT `IPushNotificationService`. 
- Move NotificationHub code into the `Bit.Core.Platform.Push` namespace and under platform ownership.
- Add various code docs to help people know how to add a new notification under the new guidelines. (Full README coming in a following PR). 
- Add feature registration code next to the code for the feature (and supporting tests)
- Specifically opted out of owning `PushType` so that teams don't need to request our review when they want to add a new type. Instead we own 3 tests ensuring that teams are using the enum correctly (Incrementing numbers, no duplicate numbers, and has `NotificationInfo` attribute).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Platform/PushRegistration/IPushRegistrationService.cs
```diff
@@ -1,10 +1,10 @@
-﻿#nullable enable
-
-using Bit.Core.Enums;
-using Bit.Core.NotificationHub;
+﻿using Bit.Core.Enums;
+using Bit.Core.Platform.PushRegistration;
 
+// TODO: Change this namespace to `Bit.Core.Platform.PushRegistration
```
dani-garcia, line 4: Is this meant to be resolved as part of this PR, or as a followup?

src/Core/Platform/Push/PushType.cs
```diff
@@ -0,0 +1,93 @@
+﻿using Bit.Core.Platform.Push;
+
+// TODO: This namespace should change to `Bit.Core.Platform.Push`
+namespace Bit.Core.Enums;
```
dani-garcia, line 4: Also this one?

src/Core/Platform/Push/PushServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// Adds a <see cref="IPushNotificationService"/> to the services that can be used to send push notifications to
+    /// end user devices. This method is safe to be ran multiple time provided <see cref="GlobalSettings"/> does not 
+    /// change between calls.
+    /// </summary>
+    /// <param name="services">The <see cref="IServiceCollection"/> to add services to.</param>
+    /// <param name="globalSettings">The <see cref="GlobalSettings"/> to use to configure services.</param>
+    /// <returns>The <see cref="IServiceCollection"/> for additional chaining.</returns>
+    public static IServiceCollection AddPush(this IServiceCollection services, GlobalSettings globalSettings)
+    {
+        ArgumentNullException.ThrowIfNull(services);
```
dani-garcia, line 25: Question: Can the the `this IServiceCollection services` in an extension method even be null?

src/Core/Platform/PushRegistration/IPushRegistrationService.cs
justindbaur: No, I want to do these as separate PR's because it will change a bunch of code in other teams code.

src/Core/Platform/Push/PushServiceCollectionExtensions.cs
justindbaur: It can!

```csharp
IServiceCollection services = null!;
services.AddPush();
```

does work in the sense that it would get inside the code we have defined. Since the compiler just does a simple rewrite of extension methods to be:

```csharp
IServiceCollection services = null!;
PushServiceCollectionExtensions.AddPush(services);
```

So it can be `null` just as any reference type parameter could be. I utterly hate the pattern of specifically allowing `null` as the parameter here though. [Identity Server has quite a few](https://github.com/DuendeSoftware/products/blob/ac893e1a6df45307e5e966bb56311d6c7599f832/identity-server/src/Shared/Extensions/StringExtensions.cs#L53). The usage of these look like you are calling an instance method of the item and therefore it shouldn't have a null reference and then right after you call an actual instance method and it would fail with a `NullReferenceException`. That's why I am a big believer of both not annotating it to say I support null but to also throw super early in case they do.

dani-garcia: Nice yeah agreed on it, that's an unfortunate limitation with the extension methods.

src/Core/Platform/PushRegistration/IPushRegistrationService.cs
dani-garcia: Sounds reasonable!

Issue #6234, opened 2025-08-22
bitwarden/nginx:2025.8.0 keeps restarting caused by a permission issue on default.conf


darkrain-nl:
### Steps To Reproduce

1. Updated Bitwarden from 2025.7.x (latest don't remember)
2. run `docker ps` to see the Nginx container is restarting

### Expected Result

The Nginx container starts as normal, like all the other containers

### Actual Result

The Nginx container keeps restarting and bitwarden cannot be accessed

### Additional Context

I did not modify the 'default.conf` file, it was auto generated. I tried changing something in config.yaml and run a rebuild, then start again but that didn't help. 

A few lines from `docker log bitwarden-nginx`
```
nginx: [emerg] open() "/etc/nginx/conf.d/default.conf" failed (13: Permission denied) in /etc/nginx/nginx.conf:148
nginx: [emerg] open() "/etc/nginx/conf.d/default.conf" failed (13: Permission denied) in /etc/nginx/nginx.conf:148
nginx: [emerg] open() "/etc/nginx/conf.d/default.conf" failed (13: Permission denied) in /etc/nginx/nginx.conf:148
nginx: [emerg] open() "/etc/nginx/conf.d/default.conf" failed (13: Permission denied) in /etc/nginx/nginx.conf:148
nginx: [emerg] open() "/etc/nginx/conf.d/default.conf" failed (13: Permission denied) in /etc/nginx/nginx.conf:148
```

### Build Version

2025.8.0

### Environment

Self-Hosted

### Environment Details

Operating System: Debian Trixie 13.0
Environment: Docker
Hardware: VPS 4 CPU cores 8GB ram

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

Flow86:
+1 same observed here:

```
dockerd/bitwarden/bitwarden-web/c20c89a5cccf/ghcr.io/bitwarden/web:2025.8.0[3655612]: chown: unknown user/group bitwarden:bitwarden
dockerd/bitwarden/bitwarden-web/c20c89a5cccf/ghcr.io/bitwarden/web:2025.8.0[3655612]: error: failed switching to "bitwarden:bitwarden": unable to find user bitwarden: no matching entries in passwd file
[...]

PR #6249, opened 2025-08-26
Fixed Nginx entrypoint to cp with preserve owner

ziqul:
## 🎟️ Tracking

https://github.com/bitwarden/server/issues/6234

## 📔 Objective

If user cleanly follow install instructions - Setup app will create nginx `default.conf` (and other files) with `644` permission owned by `bitwarden:bitwarden`.

During Nginx entrypoint script it copies generated `default.conf` to `/etc/nginx/conf.d/` but without `-p` flag - new file permissions would be `root:root 644`.

Then during startup Nginx will start as `bitwarden` user, which will not cause any issues by itself as `default.conf` is still readable by the world.

The issue is that for some reason some users have their Nginx config file (or sometimes even entire `bwdata` recursively) have `600` or `700` permissions. In this case Nginx will fail to start due to `default.conf` not readable by `bitwarden` user.

I assume that root cause is that some users mistakenly run `sudo chmod -R 700 /opt/bitwarden` from Linux installation guide after they have run `./bitwarden.sh install`. Or maybe some older version of Setup app where creating `default.conf` with `600` permissions and users are using very legacy installations.

Whatever may be the case I do not see any harm with copying with `-p` it even looks to me that this was the intended behaviour. This will both fix the issue for mentioned users and preserve permission structure aligned with other files.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/util/Nginx/entrypoint.sh
+++ b/util/Nginx/entrypoint.sh
@@ -30,7 +30,7 @@ mkhomedir_helper $USERNAME
 # The rest...
 
 chown -R $USERNAME:$GROUPNAME /etc/bitwarden
-cp /etc/bitwarden/nginx/*.conf /etc/nginx/conf.d/
+cp -p /etc/bitwarden/nginx/*.conf /etc/nginx/conf.d/
 mkdir -p /etc/letsencrypt
 chown -R $USERNAME:$GROUPNAME /etc/letsencrypt
 mkdir -p /etc/ssl


addisonbeck:
@ziqul Thanks for the contribution! I agree, this seems like intended functionality. If this resolves the bugs folks are seeing that's great.

PR #6253, opened 2025-08-26
[PM-24279] Add vnext policy endpoint

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24279

## 📔 Objective

This is just the next endpoint, which provides a different interface that the client will consume later. The logic for this endpoint will be fleshed out in future work.

The main purpose of this PR is to introduce the route, request body, and mapper:

1. Add endpoint  
2. Add mapper  
3. Add tests for mapper  

There are some concerns that we will not address in this PR. These concerns are documented in this [doc](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1842282517/Bulk+IPolicyRequirement+Design#Refactoring-the-Policy-Validator) and will be addressed in the near future.

## 📸 Screenshots
My frontend is having issues, so I'm testing this using Postman. Since we’re not hooking up this endpoint yet, this should be safe.

image

Confirmed that collections were created in the DB.

image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+            _ => new EmptyMetadataModel()
+        };
+    }
+
+    private IPolicyMetadataModel MapToPolicyMetadata<T>() where T : IPolicyMetadataModel, new()
+    {
+        try
+        {
+            var json = JsonSerializer.Serialize(Metadata);
+            return CoreHelpers.LoadClassFromJsonData<T>(json);
```
JimmyVo16, line 54: @eliykat Currently, if the JSON format doesn’t match the C# DTO, it just returns an empty object of that C# DTO. Is this acceptable behavior, or should we handle it differently?

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Text.Json;
+using Bit.Core.AdminConsole.Enums;
+using Bit.Core.AdminConsole.Models.Data;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
+using Bit.Core.Context;
+using Bit.Core.Utilities;
+
+namespace Bit.Api.AdminConsole.Models.Request;
+
+public class SavePolicyRequest
```
JimmyVo16, line 11: @eliykat Sonar is recommending [this](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=6253&id=bitwarden_server&open=AZjoq-3gUpsQxVADvfLg&tab=activity), but we don't do it for `/organizations/{{orgId}}/policies/{{type}}`. Please let me know if I should add it for vNext

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -212,4 +213,34 @@ public async Task<PolicyResponseModel> Put(Guid orgId, PolicyType type, [FromBod
...
+        // This logic will be fleshed out in the following PRs in PM-24279
+        var savePolicyModel = await model.ToSavePolicyModelAsync(orgId, _currentContext);
+        return new PolicyResponseModel(new Policy());
+
+        async Task ValidatePolicyContextRequest()
+        {
+            if (!_featureService.IsEnabled(FeatureFlagKeys.CreateDefaultLocation))
+            {
+                throw new NotFoundException();
+            }
```
eliykat: You can apply the `RequireFeatureAttribute` to the method and this will be handled for you.

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -212,4 +213,34 @@ public async Task<PolicyResponseModel> Put(Guid orgId, PolicyType type, [FromBod
...
+
+            if (!await _currentContext.ManagePolicies(orgId))
+            {
+                throw new NotFoundException();
+            }
```
eliykat: ...and you can use `[Authorize<ManagePoliciesRequirement>]` on the route and this will _also_ be handled for you 😎

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -212,4 +213,34 @@ public async Task<PolicyResponseModel> Put(Guid orgId, PolicyType type, [FromBod
...
+
+            if (type != model?.Policy.Type)
+            {
+                throw new BadRequestException("Mismatched policy type");
+            }
```
eliykat: I'm not sure why we include the policy type in the route and the request model. Maybe it's time to pick one or the other and just remove this error state entirely.

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
eliykat: I asked about this in Slack: https://bitwarden.slack.com/archives/C04LE162PHA/p1747006779727119

TL;DR no change required, but see if you can mark it as resolved in SonarQube somehow.

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+        return new SavePolicyModel(updatedPolicy, performedBy, metadata);
+    }
+
+    private IPolicyMetadataModel MapToPolicyMetadata()
+    {
+        return Policy?.Type switch
+        {
+            PolicyType.OrganizationDataOwnership => MapToPolicyMetadata<OrganizationModelOwnershipPolicyModel>(),
+            _ => new EmptyMetadataModel()
+        };
```
eliykat, line 46: (non-blocking) I want us to think about how this scales. It's not a huge deal right now because this is the only policy that uses this model. But as policies are added that use this pattern, this is an unexpected place that teams have to know to come and update, and it has some overhead in figuring out how it's passed through to their method.

There is also a runtime issue (I think), where your side effect method is going to cast this to get the underlying type (rather than `IPolicyMetadataModel`), but that assumes that it's been correctly instantiated as the correct type by the api layer. This creates an unexpected dependency.

There is an advantage to the `Policy.Data` pattern, where the model is passed through as a json blob and only deserialized by the code that needs it when it needs it (in this case, the side effect method).

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
eliykat: That's OK generally, except that in this case the collection name is required. So we still want some kind of check & throw if the collection name is not provided.

src/Api/AdminConsole/Controllers/PoliciesController.cs
JimmyVo16: This is just me guessing, but it seems like the desire to have the type ID in the route is because the individual policies are treated as resources, so the type IDs are similar to a resource ID, but slightly different since it also depends on the parent’s ID.

Having it in the request body seems to make mapping operations easier.

But yeah, we should stick with one approach. I’m fine either way since they both have their pros and cons.

JimmyVo16: If we decide to remove it from the route, I'll make sure to account for the order of the controller methods to prevent any clashes.

src/Api/AdminConsole/Controllers/PoliciesController.cs
JimmyVo16: Wow this is cool 😎 lol

src/Api/AdminConsole/Controllers/PoliciesController.cs
JimmyVo16: [same](https://github.com/bitwarden/server/pull/6253#discussion_r2304041892)

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
JimmyVo16: > this is an unexpected place that teams have to know to come and update

You’re right about this.

> There is an advantage to the Policy.Data pattern, where the model is passed through as a json blob and only deserialized by the code that needs it when it needs it (in this case, the side effect method).

So we can follow this pattern, but we’ll run into this [problem](https://github.com/bitwarden/server/pull/6253#discussion_r2303486081): currently, it doesn’t look like we’re validating `Policy.Data` at the controller level. If it fails, it seems like it will fail silently or return a 500 at best.

Ideally, if the client passes in invalid data, we should return a 422 with a helpful message. I’m guessing validation problems never arise because the fields in Policy.Data are optional?

#### For this PR

I think the trade-offs are either:  
1. Use the existing pattern like `Policy.Data` and not have validation, which isn’t great since we have a required field.  
2. Have the controller handle it and return a helpful error message, but this will have to live in the application layer, so it will be another place developers just have to know.  

I'm open to any other ideas

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
JimmyVo16: The outcome of this [discussion](https://github.com/bitwarden/server/pull/6253#discussion_r2304149906) will dictate this.

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -212,4 +215,16 @@ public async Task<PolicyResponseModel> Put(Guid orgId, PolicyType type, [FromBod
         var policy = await _savePolicyCommand.SaveAsync(policyUpdate);
         return new PolicyResponseModel(policy);
     }
+
+
+    [HttpPut("{type}/vnext")]
+    [RequireFeatureAttribute(FeatureFlagKeys.CreateDefaultLocation)]
+    [Authorize<ManagePoliciesRequirement>]
```
JimmyVo16, line 219: I don’t think there’s much value in adding tests to this endpoint at this stage since we moved validations to use the attributes.

We can revisit this in future PRs for this same ticket.

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
eliykat: We should have validation on `Policy.Data` as well, so that is a bit of an unsolved problem. Ideally the json deserialization library would handle this, but I haven't looked into it. I would err on the side of following existing patterns, option 1, then we can return to the validation problem later.

That said, I understand we're trying to get this shipped, so I have no hard requirement here; whatever's reasonable.

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
eliykat: Per the other thread, we can return to validation later.

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private IPolicyMetadataModel MapToPolicyMetadata<T>() where T : IPolicyMetadataModel, new()
+    {
+        try
+        {
+            var json = JsonSerializer.Serialize(Metadata);
+            return CoreHelpers.LoadClassFromJsonData<T>(json);
```
eliykat, line 54: Why do you have to serialize this first? Maybe because you're constructing a `Dictionary` out of it to begin with? If you keep it as a `string` in the api response model, you won't have to do this, you can just use the `CoreHelpers` deserializer directly.

src/Core/AdminConsole/OrganizationFeatures/Policies/Models/SavePolicyModel.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
+
+public record SavePolicyModel(PolicyUpdate PolicyUpdate, IActingUser? PerformedBy, IPolicyMetadataModel Metadata)
+{
+    public PolicyUpdate PolicyUpdate { get; init; } = PolicyUpdate;
+    public IPolicyMetadataModel Metadata { get; init; } = Metadata;
+
+    public IActingUser? PerformedBy { get; init; } = PerformedBy;
+}
```
eliykat: These declarations are redundant. This does the same thing:

```c#
public record SavePolicyModel(PolicyUpdate PolicyUpdate, IActingUser? PerformedBy, IPolicyMetadataModel Metadata);
```

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/SavePolicyCommand.cs
```diff
@@ -78,6 +80,23 @@ public async Task<Policy> SaveAsync(PolicyUpdate policyUpdate)
         return policy;
     }
 
+    public async Task<Policy> VNextSaveAsync(SavePolicyModel policyModel)
```
eliykat: I suggest a comment here, e.g.

> FIXME: this is a first pass at implementing side effects after the policy has been saved, which was not supported by the validator pattern. However, this needs to be implemented in a policy-agnostic way rather than building out switch statements in the command itself.

(my understanding at least; feel free to make it less opinionated as we're still going to discuss this as a team)

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -17,28 +16,45 @@ public class OrganizationDataOwnershipPolicyValidator(
     IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories,
     IFeatureService featureService,
     ILogger<OrganizationDataOwnershipPolicyValidator> logger)
-    : OrganizationPolicyValidator(policyRepository, factories)
+    : OrganizationPolicyValidator(policyRepository, factories), IPostSavePolicySideEffect, IPolicyValidator
```
eliykat: Suggestion: this no longer needs to implement `IPolicyValidator`, you can remove that class and its stub/noop implementations.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -17,28 +16,45 @@ public class OrganizationDataOwnershipPolicyValidator(
...
+            return;
+        }
+
+        var policyUpdate = policyRequest.PolicyUpdate;
+
+        if (postUpdatedPolicy?.Enabled == true || !policyUpdate.Enabled)
         {
-            await UpsertDefaultCollectionsForUsersAsync(policyUpdate);
+            return;
         }
```
eliykat: We want to run the side effect if the policy is changing state from disabled -> enabled. However, this method only has access to the policy **request** and the policy **after** it's been updated; you also need to know the previous state to know whether it's been toggled.

The old validator interface got the `currentPolicy` passed into it so it could determine this.

A quick fix for this would just be to return the `currentPolicy` object out of `SavePolicyCommand.SaveAsync`, e.g. `public async Task<(Policy PostUpdate, Policy PreUpdate)> SaveAsync(PolicyUpdate policyUpdate)`. That way it's in scope in your vnext method.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
 
-public abstract class OrganizationPolicyValidator(IPolicyRepository policyRepository, IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories) : IPolicyValidator
+public abstract class OrganizationPolicyValidator(IPolicyRepository policyRepository, IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories)
 {
```
eliykat: You can also remove the `IPolicyValidator` abstract declarations from this class as well now that it's no longer implementing `IPolicyValidator` (which I think is right).

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
JimmyVo16: > We want to run the side effect if the policy is changing state from disabled -> enabled. However, this method only has access to the policy request and the policy after it's been updated; you also need to know the previous state to know whether it's been toggled.

Yeah, it looks like this method needs to be completely refactored.

> A quick fix for this would just be to return the currentPolicy object out of SavePolicyCommand.SaveAsync, e.g. public async Task<(Policy PostUpdate, Policy PreUpdate)> SaveAsync(PolicyUpdate policyUpdate). That way it's in scope in your vnext method.

Yeah, unfortunately, getting the current policy is multiple layers deep in `RunValidatorAsync`. I think a simpler solution would be to abstract that logic into a private method and have both places call it.

It adds an extra request, but it’s very safe, and all the existing tests remain valid since we’re not changing the flow.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -49,20 +66,13 @@ private async Task UpsertDefaultCollectionsForUsersAsync(PolicyUpdate policyUpda
 
         if (!userOrgIds.Any())
         {
-            logger.LogError("No UserOrganizationIds found for {OrganizationId}", policyUpdate.OrganizationId);
```
JimmyVo16, line 52: Removed the error log because this is a valid use case, as an org can enable and disable users. Only confirmed users are applicable.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
JimmyVo16: Yes, I think we should remove the IPolicyValidator to prevent accidentally mapping it. As for removing the whole class, I think we should wait on that until we have a clear picture of the next step. This will reduce noise and make code reviews easier. I'll put a comment in the code so others will know

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
JimmyVo16: [Same](https://github.com/bitwarden/server/pull/6253#discussion_r2311041763)

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
JimmyVo16: I was following the current [pattern](https://github.com/bitwarden/server/blob/da66400248f3e7812467cb575ee3583f9c149269/src/Api/AdminConsole/Models/Request/PolicyRequestModel.cs#L19) . It’s JSON in the request body, so I’m not sure if we should make it a string. I feel like it will cause more semantic confusion and could mess up tools like Swagger.

JimmyVo16:
There are a few things I need to check before this is ready:

1. Address the type in the route  
2. Address the mapping and whether we should validate  
3. Make sure we have tests for all cases  
4. Do one final walkthrough.

src/Api/AdminConsole/Models/Request/SavePolicyRequest.cs
eliykat: My bad, I didn't realise we did this for `Policy.Data`. As you were...

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
eliykat: Yes I misspoke / repeated myself - I really just meant remove the `IPolicyValidator`.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyValidator.cs
```diff
@@ -6,6 +6,11 @@
 
 namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
 
+
+/// <summary>
+/// Please do not use this validator. We're currently in the process of refactoring our policy validator pattern.
+/// This is a stop-gap solution for post-policy-save side effects, but it is not the long-term solution.
+/// </summary>
 public abstract class OrganizationPolicyValidator(IPolicyRepository policyRepository, IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories)
```
eliykat, line 12: Now that this no longer implements `IPolicyValidator`, I suggest deleting unused abstract methods relating to that interface:
- PolicyType
- RequiredPolicies
- OnSaveSideEffectsAsync
- ValidateAsync

And then deleting the (also unused) `override` implementations of those in `OrganizationDataOwnershipPolicyValidator`.

I think this is a small impact on the diff and functionality (for getting this PR merged) but just tidies up some of the loose ends from this PR.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -6,25 +6,30 @@
 using Bit.Core.AdminConsole.Repositories;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
-using Microsoft.Extensions.Logging;
 
 namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
 
+/// <summary>
+/// Please do not use this validator. We're currently in the process of refactoring our policy validator pattern.
```
eliykat: Suggestion for clarity:

```suggestion
/// Please do not extend or expand this validator. We're currently in the process of refactoring our policy validator pattern.
```

Given that we are in fact going to be using it.

test/Core.Test/AdminConsole/AutoFixture/PolicyUpdateFixtures.cs
```diff
@@ -18,7 +18,7 @@ public void Customize(IFixture fixture)
     }
 }
 
-public class PolicyUpdateAttribute(PolicyType type, bool enabled = true) : CustomizeAttribute
+public class PolicyUpdateAttribute(PolicyType type = PolicyType.FreeFamiliesSponsorshipPolicy, bool enabled = true) : CustomizeAttribute
```
eliykat, line 21: This seems like an arbitrary default value, which isn't very intuitive. Why would it default to `FreeFamiliesSponsorshipPolicy` specifically? It looks like you're only using the default in 1 test - better to be explicit and just specify it there.

As an aside: if you did want this to be optional, then I think the better way would be to leave it to a random autodata value. e.g.

```c#
internal class PolicyUpdateCustomization(bool enabled, PolicyType? type = null) : ICustomization
{
    public void Customize(IFixture fixture)
    {
        fixture.Customize<PolicyUpdate>(composer => composer
            .With(o => o.Enabled, enabled)
            .With(o => o.PerformedBy, new StandardUser(Guid.NewGuid(), false)));

        if (type.HasValue)
        {
            fixture.Customize<PolicyUpdate>(composer => composer
                .With(o => o.Type, type.Value));
        }
    }
}

```

...but even then, random autodata values can also be a footgun; it could accidentally be the `OrganizationDataOwnership` policyType. Not required in this case I think - just specify the type you want in the test.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/SavePolicyCommandTests.cs
```diff
@@ -281,6 +281,90 @@ public async Task SaveAsync_ThrowsOnValidationError([PolicyUpdate(PolicyType.Sin
...
+    [BitMemberAutoData(nameof(SampleUnsupportedTestCases))]
+    public async Task VNextSaveAsync_NonOrganizationDataOwnershipPolicy_DoesNotExecutePostSaveSideEffects(
+         PolicyType policyType,
+         Policy currentPolicy,
+         [PolicyUpdate] PolicyUpdate policyUpdate)
+    {
+        // Arrange
+        policyUpdate.Type = policyType;
+        currentPolicy.Type = policyType;
+        currentPolicy.OrganizationId = policyUpdate.OrganizationId;
```
eliykat: Suggestion (non-blocking - more of an fyi): If your parameterized test data are compile-time constants, you can use `BitAutoData`. In this case:

```c#
    [Theory]
    [BitAutoData(PolicyType.SingleOrg)]
    [BitAutoData(PolicyType.TwoFactorAuthentication)]
    public async Task VNextSaveAsync_NonOrganizationDataOwnershipPolicy_DoesNotExecutePostSaveSideEffects(
```

...which is the same, but more concise.

MemberData is more useful when you have non-constant expressions, which you can't use in attribute constructors. For example, this won't work:

```c#
    [Theory]
    [BitAutoData(new Policy { Type = PolicyType.SingleOrg })]
    [BitAutoData(new Policy { Type = PolicyType.TwoFactorAuthentication })]
    public async Task VNextSaveAsync_NonOrganizationDataOwnershipPolicy_DoesNotExecutePostSaveSideEffects(
```

So that's where you'd construct your objects in a static member and then point `BitMemberAutoData` at it.

In your case however, I suggest you can probably just use `BitAutoData`.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
```diff
@@ -24,110 +23,111 @@ public class OrganizationDataOwnershipPolicyValidatorTests
     [Theory, BitAutoData]
     public async Task ExecuteSideEffectsAsync_FeatureFlagDisabled_DoesNothing(
         [PolicyUpdate(PolicyType.OrganizationDataOwnership)] PolicyUpdate policyUpdate,
-        [Policy(PolicyType.OrganizationDataOwnership, false)] Policy currentPolicy,
+        [Policy(PolicyType.OrganizationDataOwnership, false)] Policy postUpdatedPolicy,
+        [Policy(PolicyType.OrganizationDataOwnership, false)] Policy previousPolicyState,
```
eliykat: The `postUpdatedPolicy` state should match the `policyUpdate` state. Here the `policyUpdate` is enabled but the `postUpdatedPolicy` is not, which doesn't reflect real-world behavior.

Not an issue in your other tests.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
```diff
@@ -250,26 +253,28 @@ await sutProvider.GetDependency<ICollectionRepository>()
     public async Task ExecuteSideEffectsAsync_WhenDefaultCollectionsDoesNotExist_DoesNothing(
```
eliykat: Suggestion: this could use more descriptive test name, I had to look it up to see what it's testing.

```
ExecuteSideEffectsAsync_WhenDefaultCollectionNameIsInvalid_DoesNothing
```

You could also combine this with the `ExecuteSideEffectsAsync_WhenMetadataIsNull_DoesNothing` test, which is really a variant of this test case. `yield return null`.

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        _loginHelper = new LoginHelper(_factory, _client);
+    }
+
+    public async Task InitializeAsync()
+    {
+        _ownerEmail = $"integration-test{Guid.NewGuid()}@bitwarden.com";
+        await _factory.LoginWithNewAccount(_ownerEmail);
+
+        (_organization, _) = await OrganizationTestHelpers.SignUpAsync(_factory, plan: PlanType.EnterpriseAnnually2023,
+            ownerEmail: _ownerEmail, passwordManagerSeats: 10, paymentMethod: PaymentMethodType.Card);
```
eliykat: We may as well use the latest enterprise plan - `EnterpriseAnnually` (no year suffix).

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            Assert.NotNull(policy.Data);
+            var data = policy.GetDataModel<MasterPasswordPolicyData>();
+            Assert.Equal(request.Policy.Data["minComplexity"], data.MinComplexity);
+            Assert.Equal(request.Policy.Data["minLength"], data.MinLength);
+            Assert.Equal(request.Policy.Data["requireUpper"], data.RequireUpper);
+            Assert.Equal(request.Policy.Data["requireLower"], data.RequireLower);
+            Assert.Equal(request.Policy.Data["requireNumbers"], data.RequireNumbers);
+            Assert.Equal(request.Policy.Data["requireSpecial"], data.RequireSpecial);
+            Assert.Equal(request.Policy.Data["enforceOnLogin"], data.EnforceOnLogin);
```
eliykat: 🎨 Suggestion (non-blocking): `AssertHelper.AssertPropertyEqual` can help with this.

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        return;
+
+        async Task AssertResponse()
+        {
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            var responseContent = await response.Content.ReadAsStringAsync();
+
+            var result = JsonSerializer.Deserialize<JsonElement>(responseContent);
+            Assert.True(result.GetProperty("enabled").GetBoolean());
+            Assert.Equal((int)policyType, result.GetProperty("type").GetInt32());
```
eliykat: You can instantiate a model with `response.Content.ReadFromJsonAsync<T>()`, which makes assertions easier.

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        };
+
+        // Act
+        var response = await _client.PutAsync($"/organizations/{_organization.Id}/policies/{policyType}/vnext",
+            JsonContent.Create(request));
+
+        // Assert
+        await AssertResponse();
+
+        await AssertPolicy();
```
eliykat, line 92: You should also assert that the default collections were created. See [OrganizationTestHelpers](https://github.com/bitwarden/server/blob/main/test/Api.IntegrationTest/Helpers/OrganizationTestHelpers.cs) in this project for some helper methods to create other OrgUsers. You can also compare to Rui's confirm tests: https://github.com/bitwarden/server/blob/main/test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerTests.cs#L112-L113

eliykat (changes requested):

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
JimmyVo16: Question: Do you want to assert all the side effects moving forward? What is the expectation for these endpoint tests?

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
JimmyVo16: Actually, we can remove `ExecuteSideEffectsAsync_WhenMetadataIsNull_DoesNothing`, since yield return `[new EmptyMetadataModel()]` already covers that case.

test/Core.Test/AdminConsole/AutoFixture/PolicyUpdateFixtures.cs
JimmyVo16: > This seems like an arbitrary default value, which isn't very intuitive. Why would it default to FreeFamiliesSponsorshipPolicy specifically? It looks like you're only using the default in 1 test

Yeah, the goal was to have an arbitrary default value since it implies the user doesn’t care and just wants to create an object. If they do care, they’ll provide a value. What is the concern with picking an arbitrary value for an enum when the user doesn’t care about the type?

I would prefer not to use conditionals or risk the potential issues of randomness if we can avoid it.

JimmyVo16: Just looked at the test again, and I think I should also extend this to the policy attribute for consistency. I’m open to suggestions, but ideally, we should avoid conditionals or randomness if possible.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidatorTests.cs
eliykat: Not necessarily; the current implementation does (because it maps `null` to the empty model) but from a testing perspective that's not guaranteed.

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
eliykat: In my view, it depends _what_ you're testing. What do you care about? In this case, we have this new logic specific to the Organization Data Ownership Policy that triggers these side effects, the goal of these integration tests is to give us confidence about that logic, so I think it makes sense to assert that a default collection was created. Otherwise we could be getting a valid response back, but maybe the side effects never ran. In other words, the side effects are an important part of the system under test.

I think the new assertions you added covers this 👍 so this can be resolved.

test/Core.Test/AdminConsole/AutoFixture/PolicyUpdateFixtures.cs
eliykat: I have bad news for you about randomness... all the test data you're injecting is randomly generated by the [AutoFixture](https://github.com/AutoFixture/AutoFixture) library, unless you specify otherwise through various means (e.g. `BitAutoData`, `BitMemberAutoData`, or parameter attributes that customize the build chain, which is what the `PolicyUpdateAttribute` does... it's a real rabbithole).

Anyway, that's just an aside. My concern was that you usually do care about the policy type, it's not very hard to specify, and that is maximum clear. But I see here you're setting it in the body of the test, so it's not like you don't care about it, you just don't care about it _in the parameters_. I think it's just a matter of preference and I don't have a better solution at the moment, so this is OK with me.

test/Api.IntegrationTest/AdminConsole/Controllers/PoliciesControllerTests.cs
JimmyVo16: It sounds like we want to confirm the end state in the DB for these tests, which I agree with. This sets a clear expected behavior moving forward.

Some side effects can’t be easily confirmed. For those edge cases, we can handle them differently as they come up.

PR #6276, opened 2025-09-03
[PM-25381] Add env variables for controlling refresh token lifetimes

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25381

## 📔 Objective

Add a global setting for setting the `SlidingRefreshTokenLifetime` and `AbsoluteRefreshTokenLifetime` (an int in seconds) that will be used (if not null) instead of the passed in constructor values.

### AI Summary

I've successfully added global settings for controlling refresh token lifetimes in the ApiClient. Here's what was implemented:

1. Added new properties to IdentityServerSettings class in GlobalSettings.cs:
•	SlidingRefreshTokenLifetimeSeconds - Global override for sliding refresh token lifetime in seconds
•	AbsoluteRefreshTokenLifetimeSeconds - Global override for absolute refresh token lifetime in seconds
Both properties are nullable integers, so when they're null, the existing constructor parameter behavior is preserved.
2. Updated ApiClient constructor to use these global settings:
•	Modified the constructor to check if the global settings are provided (not null)
•	If the global setting is provided, it uses that value instead of the constructor parameter
•	If the global setting is null, it falls back to the existing behavior using constructor parameters

#### Key Features:

•	Backward compatibility: Existing behavior is preserved when global settings are not provided
•	Global override: When global settings are provided, they take precedence over constructor parameters
•	Null-safe: The implementation properly handles null values
•	Documentation: Added XML documentation comments explaining the purpose of the new properties

The implementation allows administrators to globally configure refresh token lifetimes across all ApiClient instances while maintaining backward compatibility for existing configurations that don't use these global settings.

## Demo

https://github.com/user-attachments/assets/58874d93-bc14-42b8-84c4-fb3c4576a0c5

### DIFF
--- a/src/Identity/IdentityServer/ApiClient.cs
+++ b/src/Identity/IdentityServer/ApiClient.cs
@@ -18,14 +18,14 @@ public class ApiClient : Client
     {
         ClientId = id;
         AllowedGrantTypes = new[] { GrantType.ResourceOwnerPassword, GrantType.AuthorizationCode, WebAuthnGrantValidator.GrantType };
-        
+
         // Use global setting: false = Sliding (default), true = Absolute
-        RefreshTokenExpiration = globalSettings.IdentityServer.UseAbsoluteRefreshTokenExpiration 
-            ? TokenExpiration.Absolute 
+        RefreshTokenExpiration = globalSettings.IdentityServer.UseAbsoluteRefreshTokenExpiration
+            ? TokenExpiration.Absolute
             : TokenExpiration.Sliding;
-        
+
         RefreshTokenUsage = TokenUsage.ReUse;
-        
+
         // Use global setting if provided, otherwise use constructor parameter
         SlidingRefreshTokenLifetime = globalSettings.IdentityServer.SlidingRefreshTokenLifetimeSeconds ?? (86400 * refreshTokenSlidingDays);
         AbsoluteRefreshTokenLifetime = globalSettings.IdentityServer.AbsoluteRefreshTokenLifetimeSeconds ?? 0; // forever


src/Identity/IdentityServer/ApiClient.cs
```diff
@@ -20,8 +20,11 @@ public ApiClient(
         AllowedGrantTypes = new[] { GrantType.ResourceOwnerPassword, GrantType.AuthorizationCode, WebAuthnGrantValidator.GrantType };
         RefreshTokenExpiration = TokenExpiration.Sliding;
         RefreshTokenUsage = TokenUsage.ReUse;
-        SlidingRefreshTokenLifetime = 86400 * refreshTokenSlidingDays;
-        AbsoluteRefreshTokenLifetime = 0; // forever
```
Patrick-Pimentel-Bitwarden: ❓ We discussed this on the auth team and to satisfy the client request to force users to login every day wouldn't we also want to set this to absolute and use an absolute refresh token lifetime?

Patrick-Pimentel-Bitwarden (commented):
I think we _could_ be missing something here with setting the sliding property to absolute. Could we set these values to be like a minute and record a demo of being logged out at the specified time limit?

If we think we don't need the changes to the `RefreshTokenExpiration` would still be good to demonstrate this is operating correctly.

src/Identity/IdentityServer/ApiClient.cs
kspearrin: Good call. Updated to add another setting for controlling the token expiration enum.

kspearrin: ref: https://docs.duendesoftware.com/identityserver/reference/models/client/#refresh-token

Patrick-Pimentel-Bitwarden: @kspearrin thank you for linking the docs! did you run these changes and observe that the absolute timing works the way we think? could we also make a recording demonstrating that this works as expected?

kspearrin: @Patrick-Pimentel-Bitwarden I added a demo to the PR description.

PR #6279, opened 2025-09-03
[PM-26222] OTP and email providers to use time-constant equality operators

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26222

## 📔 Objective

Email and OTP providers should use fixed-time comparisons of OTP codes to prevent the remote possibility of timing attacks.

## 💻 Screenshots

https://github.com/user-attachments/assets/e0ce7e6e-dda9-4d3d-b49d-65f5d2362b86


withinfocus:
@ike-kottlowski can this get some quick feature branch QA just to validate that? Small change.

trmartin4:
@withinfocus we were waiting to have a screen recording added to the PR, but I went ahead and verified the fix and added a recording.  Ike will be back on Monday and it should be a very quick review.

PR #6291, opened 2025-09-05
[PM-24192] Add OrganizationContext in API project

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24192

## 📔 Objective

This is not directly related to the linked ticket, but is a blocking refactor for its main PR #6184.

A while ago I added some static methods to:
- parse a `CurrentContextOrganization` object from claims ([link](https://github.com/bitwarden/server/blob/96fe09af893006195e6ecc50e7f6c8b786dea25d/src/Api/AdminConsole/Authorization/OrganizationClaimsExtensions.cs#L11))
- identify a ProviderUser for an organization ([link](https://github.com/bitwarden/server/blob/41001fefaeacaaf9a63740711f36fee27445fa0d/src/Api/AdminConsole/Authorization/HttpContextExtensions.cs#L39))

These were intended to replace similar logic in `CurrentContext`, given that we don't own that class and it has an enormous scope and set of responsibilities.

The static method implementation works well for constructing `CurrentContextOrganization` because it doesn't need any outside dependencies. However, the provider logic requires that you pass in an entire repository and httpContext, as well do some prior parsing of the claims to get the UserId - so this was difficult to reuse.

This PR creates a new `OrganizationContext` class and re-implements the provider logic in a non-static context so that it can inject what it needs. This results in a simpler interface.

It also provides a convenience wrapper for `GetCurrentContextOrganization` - these are related methods so I think they should stay together, and unlike a static method it can be mocked, which is easier for testing dependencies.

Nothing here is hooked up or in use yet. It will be used in #6184 (you can look at that draft PR to see example usage). Once it is proven there, I can update existing logic to use the new methods and delete any old (unused) static methods.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


jrmccannon (approved):
This looks good to me overall.

Only question would be:
Do we want to start using `OneOf<T>` for pushing None around instead of returning nullables?

PR #6307, opened 2025-09-09
[PM-22740] Update current context to jive with Send Access Tokens

ike-kottlowski:
## 🎟️ Tracking

[PM-22740](https://bitwarden.atlassian.net/browse/PM-22740)

## 📔 Objective

For Send access tokens we don't support the idea of a "Bitwarden User". Send access is anonymous but protected by access token policy. Since there is a token that means the `ClaimsPrincipal` is present in the request so the system tried to create the current context of the HTTP request which usually contains a [dbo].[user] as the `ClaimsPrincipal`. The build process for the non-Send flow hydrates this user.

The changes here allow the current context to be created without trying to force the Send access token to fit into the `[dbo].[user]` shape. This is accomplished by a quick return if the `ClientType` is `Send` when the context is built.

### Generative AI usage
Claude wrote the tests for current context. I had to make modifications to make sure they worked as well as making adjustments to make sure all the tests made sense.

We may need to revisit the tests to ensure they are functioning as we want them to.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-22740]: https://bitwarden.atlassian.net/browse/PM-22740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Core.Test/Context/CurrentContextTests.cs
+++ b/test/Core.Test/Context/CurrentContextTests.cs
@@ -1,0 +1,733 @@
+﻿using System.Security.Claims;
+using Bit.Core.AdminConsole.Context;
+using Bit.Core.AdminConsole.Entities.Provider;
+using Bit.Core.AdminConsole.Enums.Provider;
+using Bit.Core.AdminConsole.Repositories;
+using Bit.Core.Auth.Identity;
+using Bit.Core.Context;
+using Bit.Core.Enums;
+using Bit.Core.Models.Data;
+using Bit.Core.Models.Data.Organizations.OrganizationUsers;
+using Bit.Core.Repositories;
+using Bit.Test.Common.AutoFixture;
+using Bit.Test.Common.AutoFixture.Attributes;
+using Microsoft.AspNetCore.Http;
+using NSubstitute;
+using Xunit;
+
+namespace Bit.Core.Test.Context;
+
+[SutProviderCustomize]
+public class CurrentContextTests
+{
+    #region BuildAsync(HttpContext) Tests
+
+    [Theory, BitAutoData]
+    public async Task BuildAsync_HttpContext_SetsHttpContext(
+        SutProvider<CurrentContext> sutProvider)
+    {
+        var httpContext = new DefaultHttpContext();
+        var globalSettings = new Core.Settings.GlobalSettings();
+        // Act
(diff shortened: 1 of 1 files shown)


src/Core/Context/CurrentContext.cs
```diff
@@ -137,6 +129,24 @@ public virtual Task SetContextAsync(ClaimsPrincipal user)
...
+                IdentityClientType = c;
+            }
+        }
+
+        if (IdentityClientType == IdentityClientType.Send)
+        {
+            // For the Send client, we don't need to set any User specific properties on the context
+            // so just short circuit and return here.
+            return Task.FromResult(0);
+        }
```
JaredSnider-Bitwarden, line 148: ❓ should we feature flag this?

src/Core/Context/CurrentContext.cs
```diff
@@ -137,6 +129,24 @@ public virtual Task SetContextAsync(ClaimsPrincipal user)
...
+
         var subject = GetClaimValue(claimsDict, "sub");
```
JaredSnider-Bitwarden, line 150: The send access token technically has a subject - the send id.... does it make sense to move this above the return as well? We expect consumers to reference the send_id claim instead of the subject claim anyway.

JaredSnider-Bitwarden (commented):
Two questions below

src/Core/Context/CurrentContext.cs
ike-kottlowski: Can't, causes a circular dependency. see: https://github.com/bitwarden/server/pull/6307/commits/8b96735c1dfafac7e3e49a97d9bebc831f3a7522

src/Core/Context/CurrentContext.cs
ike-kottlowski: This `var subject` is used to set the `CurrentContext.UserId` on the next line.  

I don't think it makes sense to set the `CurrentContext.UserId` to a `SendId`, so I think this return is in the correct place.

JaredSnider-Bitwarden: Oh yeah, good call.

PR #6308, opened 2025-09-09
[PM-25138] Reduce db locking when creating default collections

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25138

## 📔 Objective

We currently use the same bulk method to create default collections for 1 user or for many users. However, that bulk creation method is optimized for the top end of town: it uses SqlBulkCopy, intended for large orgs. This is causing deadlock issues in tests. dbops advice is that it's best for >500 inserts at once, so we need to rethink about where and how this is used.

This PR switches the "single user" code paths back to using single user sprocs to avoid this overhead. This reverts some of the changes in #6153 which switched over to the bulk method.

It also uses `GenerateComb` and pre-sorts SqlBulkCopy data per dbops advice.

There is a follow-up ticket to write a bulk sproc (not using SqlBulkCopy) for the 1-500 range.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Enums/PolicyType.cs
+++ b/src/Core/AdminConsole/Enums/PolicyType.cs
@@ -18,6 +18,7 @@ public enum PolicyType : byte
     FreeFamiliesSponsorshipPolicy = 13,
     RemoveUnlockWithPin = 14,
     RestrictedItemTypesPolicy = 15,
+    UriMatchDefaults = 16,
 }
 
 public static class PolicyTypeExtensions
@@ -46,6 +47,7 @@ public static class PolicyTypeExtensions
             PolicyType.FreeFamiliesSponsorshipPolicy => "Remove Free Bitwarden Families sponsorship",
             PolicyType.RemoveUnlockWithPin => "Remove unlock with PIN",
             PolicyType.RestrictedItemTypesPolicy => "Restricted item types",
+            PolicyType.UriMatchDefaults => "URI match defaults",
         };
     }
 }
--- a/src/Core/Repositories/ICollectionCipherRepository.cs
+++ b/src/Core/Repositories/ICollectionCipherRepository.cs
@@ -8,6 +8,7 @@ public interface ICollectionCipherRepository
 {
     Task<ICollection<CollectionCipher>> GetManyByUserIdAsync(Guid userId);
     Task<ICollection<CollectionCipher>> GetManyByOrganizationIdAsync(Guid organizationId);
+    Task<ICollection<CollectionCipher>> GetManySharedByOrganizationIdAsync(Guid organizationId);
     Task<ICollection<CollectionCipher>> GetManyByUserIdCipherIdAsync(Guid userId, Guid cipherId);
     Task UpdateCollectionsAsync(Guid cipherId, Guid userId, IEnumerable<Guid> collectionIds);
     Task UpdateCollectionsForAdminAsync(Guid cipherId, Guid organizationId, IEnumerable<Guid> collectionIds);
--- a/src/Core/Vault/Queries/OrganizationCiphersQuery.cs
+++ b/src/Core/Vault/Queries/OrganizationCiphersQuery.cs
@@ -24,7 +24,7 @@ public class OrganizationCiphersQuery : IOrganizationCiphersQuery
         var orgCiphers = ciphers.Where(c => c.OrganizationId == organizationId).ToList();
         var orgCipherIds = orgCiphers.Select(c => c.Id);
 
-        var collectionCiphers = await _collectionCipherRepository.GetManyByOrganizationIdAsync(organizationId);
+        var collectionCiphers = await _collectionCipherRepository.GetManySharedByOrganizationIdAsync(organizationId);
         var collectionCiphersGroupDict = collectionCiphers
             .Where(c => orgCipherIds.Contains(c.CipherId))
             .GroupBy(c => c.CipherId).ToDictionary(s => s.Key);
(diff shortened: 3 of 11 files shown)


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
```diff
@@ -245,14 +250,54 @@ private async Task<IEnumerable<string>> GetUserDeviceIdsAsync(Guid userId)
...
 
     /// <summary>
-    /// Handles the side effects of confirming an organization user.
-    /// Creates a default collection for the user if the organization
-    /// has the OrganizationDataOwnership policy enabled.
+    /// Creates a default collection for a single user if required by the Organization Data Ownership policy.
+    /// </summary>
+    /// <param name="organizationUser">The organization user who has just been confirmed.</param>
+    /// <param name="defaultUserCollectionName">The encrypted default user collection name.</param>
+    private async Task CreateDefaultCollectionAsync(OrganizationUser organizationUser, string defaultUserCollectionName)
```
eliykat, line 257: There is some code duplication here with the `CreateManyDefaultCollectionsAsync` method, but this is the simplest solution for now. I think we should be able to align the code paths once we've built out better bulk sprocs for policy enforcement and default collection creation.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -10,8 +10,21 @@ public static class BulkResourceCreationService
...
+            .ThenBy(cu => cu.OrganizationUserId)
+            .ToList();
+
         using var bulkCopy = new SqlBulkCopy(connection, SqlBulkCopyOptions.KeepIdentity, transaction);
         bulkCopy.DestinationTableName = "[dbo].[CollectionUser]";
+        bulkCopy.BatchSize = 500;
+        bulkCopy.BulkCopyTimeout = 120;
+        bulkCopy.EnableStreaming = true;
+        bulkCopy.ColumnOrderHints.Add("CollectionId", SortOrder.Ascending);
+        bulkCopy.ColumnOrderHints.Add("OrganizationUserId", SortOrder.Ascending);
```
eliykat: Note that this will apply to the "import vault" code path as well, but these seem like sensible defaults for both cases.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
eliykat, line 1: @rkac-bw this implements your suggestions to pre-sort the data. However, I am getting test failures:

> Microsoft.Data.SqlClient.SqlException : Cannot bulk load. The bulk data stream was incorrectly specified as sorted or the data violates a uniqueness constraint imposed by the target table. Sort order incorrect for the following two rows: primary key of first row: (0564afea-b76a-427d-bcf5-b354000c067b), primary key of second row: (aad77ce5-5dd6-444b-b11a-b354000c067b).

The [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlbulkcopycolumnorderhint?view=sqlclient-dotnet-core-6.0) say that sort order must match the clustered index definition, but best I can tell it does (`Collection` uses `Id` as its clustered index, `CollectionUser` uses `OrganizationId` and `OrganizationUserId`).

However, just through trial and error I have found that sorting (and providing `SqlBulkCopyColumnOrderHint`s) for just `Collection.OrganizationId` and _either_ `CollectionUser.OrganizationId` or `OrganizationUserId` (but not both) works fine. This doesn't seem to match the documentation though. Any advice?

eliykat: This has been resolved with the SqlGuid extension methods discussed in Slack.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -10,8 +11,21 @@ public static class BulkResourceCreationService
     private const string _defaultErrorMessage = "Must have at least one record for bulk creation.";
     public static async Task CreateCollectionsUsersAsync(SqlConnection connection, SqlTransaction transaction, IEnumerable<CollectionUser> collectionUsers, string errorMessage = _defaultErrorMessage)
     {
+        // Offload some work from SQL Server by pre-sorting before insert.
+        // This lets us use the SqlBulkCopy.ColumnOrderHints to improve performance and reduce deadlocks.
+        collectionUsers = collectionUsers
```
jrmccannon: Non-blocking: assign to new variable instead up updating the variable in place.

src/Infrastructure.Dapper/AdminConsole/Helpers/BulkResourceCreationService.cs
```diff
@@ -98,8 +112,17 @@ private static DataTable BuildCollectionsUsersTable(SqlBulkCopy bulkCopy, IEnume
 
     public static async Task CreateCollectionsAsync(SqlConnection connection, SqlTransaction transaction, IEnumerable<Collection> collections, string errorMessage = _defaultErrorMessage)
     {
+        // Offload some work from SQL Server by pre-sorting before insert.
+        // This lets us use the SqlBulkCopy.ColumnOrderHints to improve performance and reduce deadlocks.
+        collections = collections.OrderBySqlGuid(c => c.Id).ToList();
```
jrmccannon: Non-blocking: assign to new variable instead up updating the variable in place.

jrmccannon (dismissed):
Read over the thread related to this and it looks good. Just had a couple non-blocking nits about variable assignment.

Its good to merge without them though

jrmccannon (approved):

PR #6309, opened 2025-09-10
Alter integration template processing to remove keys when encountering null values

brant-livefront:
## 📔 Objective

As we began experimenting with `Template` in real world scenarios, it started to seem awkward to have keys appear in the output string if the property was null.

For example, a JSON template with `#UserId#` returned the literal key when `UserId` was `null`:

```json
{
  "ActingUser": "acting-user-12345",
  "User": "#Userld#",
  "Event": "Cipher_ClientViewed"
}
```

This PR changes the behavior to return `""` for null values when the key is recognized. If the Template contains a key that is not found on the object, it will still return the key literal (in order to help when debugging custom templates). 

Our above example would then return:
```json
{
  "ActingUser": "acting-user-12345",
  "User": "",
  "Event": "Cipher_ClientViewed"
}
```

Additionally, I moved the JSON serialization property (`EventMessage`) into the `IntegrationTemplateContext` instead of special-casing it inside the `IntegrationTemplateProcessor`. It cleans up the processing function to not have to worry about a special case in the middle of everything. I've also added tests for `IntegrationTemplateContext` to make sure it's working as expected (and removed the special case test from `IntegrationTemplateProcessor`)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationTemplateContext.cs
+++ b/src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationTemplateContext.cs
@@ -1,4 +1,5 @@
-﻿using Bit.Core.AdminConsole.Entities;
+﻿using System.Text.Json;
+using Bit.Core.AdminConsole.Entities;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Models.Data;
@@ -22,6 +23,8 @@ public class IntegrationTemplateContext(EventMessage eventMessage)
     public Guid? CollectionId => Event.CollectionId;
     public Guid? GroupId => Event.GroupId;
     public Guid? PolicyId => Event.PolicyId;
+
+    public string EventMessage => JsonSerializer.Serialize(Event);
 
     public User? User { get; set; }
     public string? UserName => User?.Name;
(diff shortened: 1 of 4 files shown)


src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationTemplateContext.cs
```diff
@@ -23,6 +24,8 @@ public class IntegrationTemplateContext(EventMessage eventMessage)
     public Guid? GroupId => Event.GroupId;
     public Guid? PolicyId => Event.PolicyId;
 
+    public string EventMessage => JsonSerializer.Serialize(Event);
```
withinfocus, line 27: ❓ Is there any risk this blows up upon construction and we have issues when loading integrations?

JimmyVo16 (approved):
Looks good from the AC team.

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationTemplateContext.cs
brant-livefront: I think we're probably fine:
- It's a lazy accessor, so it would only effect templates that use it
- It's actually the same thing we were doing before (i.e. at least no _new_ risk)
- The object is EventMessage, which is known ahead of time and is pretty flat (DateTime, strings, enums and Guids - no circular concerns)

That said, I'm happy to wrap it if we want to be fully cautious (and probably return ""). If it were _input_, I'd definitely argue we need to wrap it. But given that it's a known, simple object, I think this is safe.

withinfocus: True, the laziness makes me feel good enough about it.

PR #6317, opened 2025-09-11
PM-13632: Add support for configuring multiple allowed origins

abergs:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13632

## 📔 Objective

This PR contains enablement work for the browser extension work, tracked by https://bitwarden.atlassian.net/browse/PM-13632.

It enables us to configure multiple allowed origins, so that we can list our domain (https://bitwarden.com etc) as well as our browser extension (chrome-extension://aabbbccc).

It loads additional Origins from the configuraiton to allow us to update it dynamically or use different values in different environments.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -687,7 +687,14 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
         {
             options.ServerDomain = new Uri(globalSettings.BaseServiceUri.Vault).Host;
             options.ServerName = "Bitwarden";
-            options.Origins = new HashSet<string> { globalSettings.BaseServiceUri.Vault, };
+
+            IEnumerable<string> allowedOrigins = [globalSettings.BaseServiceUri.Vault];
```
withinfocus: 🎨 I would suggest an `if` / `else` here to use what's in configuration, otherwise use the base vault URI. If someone is configuring this we should expect that collection be the total and to not attempt a union. This makes intent more clear to me.

abergs: Adressed

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -694,8 +694,17 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
         {
             options.ServerDomain = new Uri(globalSettings.BaseServiceUri.Vault).Host;
             options.ServerName = "Bitwarden";
-            options.Origins = new HashSet<string> { globalSettings.BaseServiceUri.Vault, };
             options.TimestampDriftTolerance = 300000;
+
+            if (globalSettings.Fido2.Origins.Any())
```
withinfocus: ⚠️ What if this is `null`?

abergs: yeah I think this is the reason the test are failing, I have an unpushed commit that fixes it. Just going to confirm tests locally first.

src/Core/Settings/GlobalSettings.cs
```diff
@@ -742,4 +743,9 @@ public class WebPushSettings : IWebPushSettings
     {
         public string VapidPublicKey { get; set; }
     }
+
+    public class Fido2Settings
+    {
+        public HashSet<string> Origins { get; set; }
```
withinfocus, line 773: ℹ️ I am assuming the configuration readers in .NET can work with this type but an array or list of strings is more typical.

withinfocus (commented):
This is fine with me, but Auth to provide formal review.

JaredSnider-Bitwarden (approved):
LGTM!

abergs:
We don't need a separata QA test for this, right? I'm good to merge as is? No behaviour should change unless we start adding configuration.

JaredSnider-Bitwarden:
> We don't need a separata QA test for this, right? I'm good to merge as is? No behaviour should change unless we start adding configuration.

Since it is not feature flagged, I would recommend a QA smoke test per our standard practices.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -694,8 +694,17 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
...
-            options.Origins = new HashSet<string> { globalSettings.BaseServiceUri.Vault, };
             options.TimestampDriftTolerance = 300000;
+
+            if (globalSettings.Fido2?.Origins?.Any() == true)
+            {
+                options.Origins = new HashSet<string>(globalSettings.Fido2.Origins);
+            }
+            else
+            {
+                options.Origins = new HashSet<string> { globalSettings.BaseServiceUri.Vault };
```
kspearrin: Add hardcoded extension ids and such

trmartin4: Would you prefer to have those hard-coded at compile-time or inserted via environment variables at runtime?  I assume that by hardcoding you mean to put them into `Constants.cs` so that we don't have to ask self-hosted customers to configure their extensions as well?  We do have a [precedent](https://github.com/bitwarden/clients/blob/f04b24851c69867dd450775f910a6c48a0991ed8/apps/desktop/src/main/native-messaging.main.ts#L343) for hard-coding those on `clients` but no need up to this point on `server`.

Is there any need to provide an avenue for override/extension by third-party extension developers to add to the list for a self-hosted scenario? If these extension IDs were compile-time constants, and we have the Vault URL, then we could eliminate the need to expose the ability to configure the `Origins` via environment variables (other than the Vault URI).

kspearrin: I am trying to eliminate the need for customers to have to configure this on their end. Client IDs/origins don't really change, and if they need to break away from them for some reason (custom client, QA environment, etc), they could override it with their own ENV variable global setting that is exposed.

abergs: I've added chromium constants in https://github.com/bitwarden/server/pull/6317/commits/26e8cf44d1eb0a40ee2e5878b30f12e5ef96bc5a.

I also included the development extension so that "Login with passkey" would work for local dev / testing out of the box too.

I verified that I was able to use the feature successfully without any custom config.

src/Core/Constants.cs
```diff
@@ -70,6 +70,18 @@ public static class CountryAbbreviations
         /// </summary>
         public const string UnitedStates = "US";
     }
+
+
+    /// <summary>
+    /// Constats for our browser extensions ID's.
```
withinfocus:
```suggestion
    /// Constants for our browser extensions IDs
```

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -694,8 +695,24 @@ public static void AddWebAuthn(this IServiceCollection services, GlobalSettings
         {
             options.ServerDomain = new Uri(globalSettings.BaseServiceUri.Vault).Host;
             options.ServerName = "Bitwarden";
-            options.Origins = new HashSet<string> { globalSettings.BaseServiceUri.Vault, };
             options.TimestampDriftTolerance = 300000;
+
+            if (globalSettings.Fido2?.Origins?.Any() == true)
+            {
+                options.Origins = new HashSet<string>(globalSettings.Fido2.Origins);
```
JaredSnider-Bitwarden, line 703: ❓ : What real world scenario do we expect any values to be set in `globalSettings.Fido2.Origins`?  With the recent change to hardcode extension ids down below... I'm not seeing when this would be used?

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
trmartin4: @abergs I was thinking that we would provide defaults/overrides at the extension ID level, rather than wholesale "override all values or use the defaults" as done here.

Perhaps the `globalSettings` could be extended with an `ExtensionIdentifierSettings` class, like we do for [web push](https://github.com/bitwarden/server/blob/ff092a031eb0489672095a76c8f4d4564222b835/src/Core/Settings/GlobalSettings.cs#L745), with default values set to the production extension IDs.  That way if you want to override the Chrome extension to use the Developer version, you could make _just_ that change in your `appSettings.Development.json` and not have to also override all the other values.  It also means that we could avoid having to list the developer extension along with the others.

Part of this ties in with @JaredSnider-Bitwarden's question [here](https://github.com/bitwarden/server/pull/6317#discussion_r2375820979), as I'm not 100% sure what the use-case is for the override and so not absolutely sure how the override API could be designed to fit the anticipated case.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
abergs: @JaredSnider-Bitwarden Any browser extension that is not already hard coded ("beta", new browser etc) or custom domain / extension-id in a selfhost/enterprise deployment.

As well as the ability to remove the support for some extension, for whatever reason that may be.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
trmartin4: Also, we believe that the hard-coding of the development extension is not valid.  From what QA has told us, when you side-load an extension you get a different, random ID for each time you do so.  So hard-coding that value is just capturing what that random GUID is for you and will not work for others.

abergs: @trmartin4 Happy to remove the development extension. I do think it's enough to only have one globalsetting-origins. It covers the balance of being able to remove and customize as much as your want - while keeping things simple.

Let me know if you'd really want me to make a more elaborate configuration and i'll tweak it, but I think the current escape hatch should serve us well.

trmartin4: Understood.  So the expectation is that if you take responsibility for providing any overrides, you must provide _all_ origins.  For example, if QA wants to test the development Chrome extension against production Edge, they cannot just override the Chrome origin, they would need to take responsibility for defining the entire array (web vault URL, and all extension origins that they want to include).  As you noted [here](https://github.com/bitwarden/server/pull/6317#discussion_r2376540211), we intend for the configuration to be used for "any browser extension that is not already hard-coded" - does that imply that we'd want to be able to include those without having to re-define those hard-coded values in your environment settings?

If we are OK with that additional complexity, then I think that's a valid approach, and perhaps we could see if this way of exposing the overrides is a burden for the QA process before officially documenting it for external use.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
JaredSnider-Bitwarden: > Any browser extension that is not already hard coded 

I'm sorry for the confusion here, but doesn't that imply that we would want the config to be additive?
` options.Origins = [...hardCodedOrigins, ...globalSettings.Fido2.Origins]` (not working code just illustrative)?

abergs: @JaredSnider-Bitwarden That was the original implementation. Bishop [suggested](https://github.com/bitwarden/server/pull/6317#discussion_r2344200357) we make it explicit.

If we make it additive, our defaults would always be allowed which in theory could be a 'thing'. At least this way the config have maximum flexibility. I'm Ok either way.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
abergs: > does that imply that we'd want to be able to include those without having to re-define those hard-coded values in your environment settings?

My comment from the other review thread applies here, so I'm re-iterating it:

Yes, the original implementation was additive, Bishop [suggested](https://github.com/bitwarden/server/pull/6317#discussion_r2344200357) we make it explicit.

If we make it additive, our defaults would always be allowed which in theory could be a 'thing'. At least this way the config have maximum flexibility with the cost of having to be explicit. I'm Ok either way though.

JaredSnider-Bitwarden (dismissed):
Thank you for this work and for the extra clarifications for our questions!

abergs:
@JaredSnider-Bitwarden @trmartin4 I'm moving this back in to ready for QA.

JaredSnider-Bitwarden (approved):

PR #6320, opened 2025-09-12
[PM-24357] Do not purge ciphers in the default collection

nick-livefront:
## 🎟️ Tracking

[PM-24357](https://bitwarden.atlassian.net/browse/PM-24357)

## 📔 Objective

When a cipher is a part of the "My Items"/"Default" collection the cipher should not be deleted when the organization vault is purged.  

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/9881a454-e3dc-432c-a9f8-63a11931519b" />

## 🦮 Reviewer guidelines

[PM-24357]: https://bitwarden.atlassian.net/browse/PM-24357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/Infrastructure.IntegrationTest/Vault/Repositories/CipherRepositoryTests.cs
shane-melton, line 1: 👍  Thanks for the integration test!

shane-melton (dismissed):
Changes look good!

jaasen-livefront (dismissed):
💯

nick-livefront:
@bitwarden/dept-dbops Just a nudge - could this one get a review?

util/Migrator/DbScripts/2025-09-12_00_ExcludeDefaultCiphersFromPurge.sql
mkincaid-bw, line 1: There are a few things wrong with the original proc that should be addressed.

1. The first delete statement from the `CollectionCipher` table is not needed because it has a foreign key to the Cipher `table` with `ON CASCADE DELETE`. This means that when the records are deleted from `Cipher`, those deletes will cascade down to the `CollectionCipher` table. In my testing, this unnecessary delete can slow the proc down by ~50%

2. The batch size of 100 is small and also causes a slowdown. I tested with various values from 100 - 20,000 and setting the batch size to 1,000 seems to have the best performance while minimizing lock escalations.

3. The existing delete statement does not guarantee any sort of delete order. While I don't think this proc will run that often, controlling the delete order can help eliminate potential deadlocks.

4. The transactions as written aren't doing anything and should be updated to properly assure that the ciphers are deleted all or nothing.

I've re-written the proc to address these. You should confirm that the business logic still function as expected but this should improve performance and still give the desired results.

```sql
CREATE OR ALTER PROCEDURE [dbo].[Cipher_DeleteByOrganizationId]
    @OrganizationId AS UNIQUEIDENTIFIER
AS
BEGIN
    SET NOCOUNT ON;
    DECLARE @BatchSize INT = 1000;
    
    BEGIN TRY
        BEGIN TRANSACTION Cipher_DeleteByOrganizationId;
        
        WHILE @BatchSize > 0
        BEGIN

            DELETE 
            FROM
                [dbo].[Cipher]
            WHERE 
                Id IN
                (   SELECT TOP (@BatchSize)
                        C.Id 
                    FROM 
                        [dbo].[Cipher] C
                    LEFT JOIN 
                        [dbo].[CollectionCipher] CC ON CC.[CipherId] = C.[Id]
                    LEFT JOIN 
                        [dbo].[Collection] Col ON Col.[Id] = CC.[CollectionId] AND Col.[Type] = 1
                    WHERE 
                        C.[OrganizationId] = @OrganizationId
                        AND Col.[Id] IS NULL
                    ORDER BY 
                        C.[Id]
                );

            SET @BatchSize = @@ROWCOUNT;
        END
        
        EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationId] @OrganizationId;
        COMMIT TRANSACTION Cipher_DeleteByOrganizationId;
        
        EXEC [dbo].[Organization_UpdateStorage] @OrganizationId;
        
    END TRY
    BEGIN CATCH
        IF @@TRANCOUNT > 0
            ROLLBACK TRANSACTION Cipher_DeleteByOrganizationId;
        THROW; -- Re-raise the error
    END CATCH
END
GO
```

mkincaid-bw (changes requested):
Change requested to improve performance.

util/Migrator/DbScripts/2025-09-12_00_ExcludeDefaultCiphersFromPurge.sql
nick-livefront: Thanks, @mkincaid-bw! Very detailed feedback and goes beyond my knowledge in some aspects, very helpful. 

One thing that is missing with the procedure you wrote is when a cipher is in multiple collections. When a cipher is in multiple collections, one of which has a Type of 1 it is still be deleted when it should be kept.

nick-livefront: >This means that when the records are deleted from Cipher, those deletes will cascade down to the CollectionCipher table.

When a cipher is assigned to multiple collections, this scenario does not occur. We want to delete from `CollectionCipher` but not `Cipher`.

nick-livefront: I altered the procedure to include ciphers that exist in multiple collections. [bc168f7](https://github.com/bitwarden/server/pull/6320/commits/bc168f73393d90ef07ce94e24f3d8ae800f1563d)

Happy to address any feedback with how I altered the code you had

src/Sql/dbo/Vault/Stored Procedures/Cipher/Cipher_DeleteByOrganizationId.sql
```diff
@@ -2,48 +2,64 @@
...
-            [dbo].[CollectionCipher] CC
-        INNER JOIN
-            [dbo].[Collection] C ON C.[Id] = CC.[CollectionId]
-        WHERE
-            C.[OrganizationId] = @OrganizationId
+            SET @BatchSize = @@ROWCOUNT;
+        END
 
-        SET @BatchSize = @@ROWCOUNT
+        -- Clean up remaining CollectionCipher relationships for Type = 0 collections
```
shane-melton: ⛏️ Can we clarify what the types are here. Specifically Type = 0 are _normal_ shared collections and Type = 1 are user specific default collections. (and copy to the migration script)

shane-melton (dismissed):
One small comment nit, otherwise looks good

util/Migrator/DbScripts/2025-09-12_00_ExcludeDefaultCiphersFromPurge.sql
rkac-bw: Trying to avoid potential deadlocks by deleting in order of clustered keys, which should leave us with blocking only:

 ```
CREATE OR ALTER PROCEDURE [dbo].[Cipher_DeleteByOrganizationId]
      @OrganizationId UNIQUEIDENTIFIER
  AS
  BEGIN
      SET NOCOUNT ON;

      DECLARE @BatchSize INT = 1000;

      BEGIN TRY
          BEGIN TRANSACTION;

          ---------------------------------------------------------------------
          -- 1. Delete organization ciphers that are NOT in any default
          --    user collection (Collection.Type = 1).
          ---------------------------------------------------------------------
          WHILE 1 = 1
          BEGIN
              ;WITH Target AS
              (
                  SELECT TOP (@BatchSize) C.Id
                  FROM dbo.Cipher C
                  WHERE C.OrganizationId = @OrganizationId
                    AND NOT EXISTS (
                        SELECT 1
                        FROM dbo.CollectionCipher CC2
                        INNER JOIN dbo.Collection Col2
                          ON Col2.Id = CC2.CollectionId
                          AND Col2.Type = 1  -- Default user collection
                        WHERE CC2.CipherId = C.Id
                    )
                  ORDER BY C.Id  -- Deterministic ordering (matches clustered index)
              )
              DELETE C
              FROM dbo.Cipher C
              INNER JOIN Target T ON T.Id = C.Id;

              IF @@ROWCOUNT = 0 BREAK;
          END

          ---------------------------------------------------------------------
          -- 2. Remove remaining CollectionCipher rows that reference
          --    non-default (Type = 0 / shared) collections, for ciphers
          --    that were preserved because they belong to at least one
          --    default (Type = 1) collection.
          ---------------------------------------------------------------------
          SET @BatchSize = 1000;
          WHILE 1 = 1
          BEGIN
              ;WITH ToDelete AS
              (
                  SELECT TOP (@BatchSize)
                         CC.CipherId,
                         CC.CollectionId
                  FROM dbo.CollectionCipher CC
                  INNER JOIN dbo.Collection Col
                          ON Col.Id = CC.CollectionId
                         AND Col.Type = 0  -- Non-default collections
                  INNER JOIN dbo.Cipher C
                          ON C.Id = CC.CipherId
                  WHERE C.OrganizationId = @OrganizationId
                  ORDER BY CC.CollectionId, CC.CipherId  -- Matches clustered index
              )
              DELETE CC
              FROM dbo.CollectionCipher CC
              INNER JOIN ToDelete TD
                  ON CC.CipherId = TD.CipherId
                 AND CC.CollectionId = TD.CollectionId;

              IF @@ROWCOUNT = 0 BREAK;
          END

          ---------------------------------------------------------------------
          -- 3. Bump revision date (inside transaction for consistency)
          ---------------------------------------------------------------------
          EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationId] @OrganizationId;

          COMMIT TRANSACTION ;

          ---------------------------------------------------------------------
          -- 4. Update storage usage (outside the transaction to avoid
          --    holding locks during long-running calculation)
          ---------------------------------------------------------------------
          EXEC [dbo].[Organization_UpdateStorage] @OrganizationId;
      END TRY
      BEGIN CATCH
          IF @@TRANCOUNT > 0
              ROLLBACK TRANSACTION; 
          THROW;
      END CATCH
  END
  GO
```

src/Sql/dbo/Vault/Stored Procedures/Cipher/Cipher_DeleteByOrganizationId.sql
nick-livefront: Addressed by @rkac-bw's feedback above ➕ [e9193c5](https://github.com/bitwarden/server/pull/6320/commits/e9193c566d194a398fe9153ab4189c6814da1fbe)

util/Migrator/DbScripts/2025-09-12_00_ExcludeDefaultCiphersFromPurge.sql
nick-livefront: 💯 Tested with:
- Ciphers in both Default, Non Default collections
- Ciphers in Unassigned collections
- Empty organizations with only ciphers in Default collections

[e9193c5](https://github.com/bitwarden/server/pull/6320/commits/e9193c566d194a398fe9153ab4189c6814da1fbe)

mkincaid-bw (approved):

shane-melton (approved):

PR #6338, opened 2025-09-15
Policy pattern

JimmyVo16:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyValidator.cs
@@ -41,3 +41,129 @@ public interface IPolicyValidator
     /// <param name="currentPolicy">The current policy, if any</param>
     public Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? currentPolicy);
 }
+
+public interface IPolicyUpdateEvent
+{
+    /// <summary>
+    /// The PolicyType that this definition relates to.
+    /// </summary>
+    public PolicyType Type { get; }
+}
+
+public interface IOnPolicyPostSaveEvent : IPolicyUpdateEvent
+{
+    public Task PostSideEffectsAsync(PolicyUpdate policyUpdate, Policy? currentPolicy);
+}
+
+public interface IOnPolicyPreSaveEvent : IPolicyUpdateEvent
+{
+    /// <summary>
+    /// Performs side effects after a policy is validated but before it is saved.
+    /// For example, this can be used to remove non-compliant users from the organization.
+    /// Implementation is optional; by default it will not perform any side effects.
+    /// </summary>
+    /// <param name="policyUpdate">The policy update request</param>
+    /// <param name="currentPolicy">The current policy, if any</param>
+    public Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? currentPolicy);
+}
+
+public interface IEnforceDependentPoliciesEvent : IPolicyUpdateEvent
+{
+    /// <summary>
+    /// PolicyTypes that must be enabled before this policy can be enabled, if any.
+    /// These dependencies will be checked when this policy is enabled and when any required policy is disabled.
+    /// </summary>
+    public IEnumerable<PolicyType> RequiredPolicies { get; }
+}
+
+public interface IPolicyValidationEvent : IPolicyUpdateEvent
+{
+    /// <summary>
+    /// Validates a policy before saving it.
+    /// Do not use this for simple dependencies between different policies - see <see cref="RequiredPolicies"/> instead.
+    /// Implementation is optional; by default it will not perform any validation.
+    /// </summary>
+    /// <param name="policyUpdate">The policy update request</param>
+    /// <param name="currentPolicy">The current policy, if any</param>
+    /// <returns>A validation error if validation was unsuccessful, otherwise an empty string</returns>
+    public Task<string> ValidateAsync(PolicyUpdate policyUpdate, Policy? currentPolicy);
+}
+
+public class NoOpPolicyHandler :
+    IPolicyValidationEvent,
+    IEnforceDependentPoliciesEvent,
+    IOnPolicyPreSaveEvent,
+    IOnPolicyPostSaveEvent
+{
+    private readonly PolicyType _policyType;
+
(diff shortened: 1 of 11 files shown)


src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyValidator.cs
```diff
@@ -41,3 +41,67 @@ public interface IPolicyValidator
     /// <param name="currentPolicy">The current policy, if any</param>
     public Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? currentPolicy);
 }
+
+public interface IPolicyUpdateEvent
```
JimmyVo16, line 45: Here’s a very rough draft of what I’m thinking:

We break `IPolicyValidator` into smaller sub-interfaces, each one associated with a specific event. All of them implement `IPolicyUpdateEvent`, so we know they belong to policy events. The `PolicyType` property is then used to determine which implementation corresponds to which policy.

Each policy handler can implement only the events it needs.  

**Benefits**
1. Each handler implements only the event interfaces it needs.
2. New event interfaces can be added without affecting existing interfaces or handlers.  
   - This fixes the issue with `IPolicyValidator`, where validators had to implement methods they didn’t need, and adding new methods impacted all existing validators.  
3. The consumer code (e.g., `SaveCommand`) can just use the event interfaces it needs without caring about the rest.  
4. A single handler can implement multiple event interfaces.  

**Cons**
1. You have to add the DI configuration for each event interface you implement.

src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyValidator.cs
```diff
@@ -41,3 +41,67 @@ public interface IPolicyValidator
...
+    /// </summary>
+    public IEnumerable<PolicyType> RequiredPolicies { get; }
+}
+
+public interface IPolicyValidationEvent : IPolicyUpdateEvent
+{
+    public Task<string> ValidateAsync(PolicyUpdate policyUpdate, Policy? currentPolicy);
+}
+
+public interface IPolicyEventHandlerFactory
```
JimmyVo16, line 78: Just a small factory to easily get the policy’s implementation of that interface. 

I tried creating a default one to avoid passing around `null`, but I ran into a few problems and it didn’t provide much value, so I decided not to go through with it. It’s something we can revisit later though.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/RequireSsoPolicyValidator.cs
```diff
@@ -8,16 +8,17 @@
 
 namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
 
-public class RequireSsoPolicyValidator : IPolicyValidator
+public class RequireSsoOnPolicyEventEventEnsureEventValidator : IEnforceDependentPoliciesEvent, IPolicyValidationEvent
```
JimmyVo16: Here are some examples of refactoring existing validators into handlers, as you can see, it’s pretty trivial. This example also implements multiple event interfaces.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/SavePolicyCommand.cs
```diff
@@ -144,12 +158,168 @@ private async Task RunValidatorAsync(IPolicyValidator validator, PolicyUpdate po
...
+        GetCurrentPolicyStateAsync(PolicyUpdate policyUpdate)
     {
         var savedPolicies = await _policyRepository.GetManyByOrganizationIdAsync(policyUpdate.OrganizationId);
         // Note: policies may be missing from this dict if they have never been enabled
         var savedPoliciesDict = savedPolicies.ToDictionary(p => p.Type);
         var currentPolicy = savedPoliciesDict.GetValueOrDefault(policyUpdate.Type);
         return (savedPoliciesDict, currentPolicy);
     }
+
+    public async Task<Policy> V3SaveAsync(SavePolicyModel policyModel)
```
JimmyVo16, line 171: Here is the new `Save` method implementation. It’s very similar to the existing `Save` method, except:  

- It breaks the validations into separate methods for different types of validations.  
- It includes pre- and post-side-effect methods.  
- All these methods are on the same level, making it easy for their output to be passed to the next one.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyHandler.cs
```diff
@@ -9,7 +9,7 @@ namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
 /// Please do not use this validator. We're currently in the process of refactoring our policy validator pattern.
 /// This is a stop-gap solution for post-policy-save side effects, but it is not the long-term solution.
 /// </summary>
-public abstract class OrganizationPolicyValidator(IPolicyRepository policyRepository, IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories)
+public abstract class OrganizationPolicyHandler(IPolicyRepository policyRepository, IEnumerable<IPolicyRequirementFactory<IPolicyRequirement>> factories)
```
JimmyVo16, line 12: If we want to restrict certain methods to just the update policy flow, we can define them in an abstract class, requiring child classes to implement them.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationPolicyHandler.cs
```diff
@@ -9,7 +9,7 @@ namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
...
 {
     protected async Task<IEnumerable<T>> GetUserPolicyRequirementsByOrganizationIdAsync<T>(Guid organizationId, PolicyType policyType) where T : IPolicyRequirement
```
JimmyVo16, line 14: We can still use the `IPolicyRequirement` to enforce policy on users, but it will only work when the policy is enabled. Which means it will only work for pre-policy updates when disabling a policy and post-policy updates when enabling a policy. 

I’m not sure if this makes the pattern easier or harder to use.

PR #6349, opened 2025-09-18
[PM-23249] Prevent log-out when changing KDF settings

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23249

## 📔 Objective

Prevent log-out when changing KDF settings, when `pm-23995-no-logout-on-kdf-change` feature flag is on.
This is regardless if the KDF settings got rotated by user manually or later automatically (force upgrade KDF settings)

When the feature flag is enabled:
1. Remove security stamp rotation when changing KDF - to not to invalidate the session
2. Send a Logout push notification with a reason - new clients will not to log out, when feature flag is enabled and reason matches. This is to make sure (very) old clients connected to newer server, with feature flag enables, will not go into bad state. As of not log out and not be able to pickup new KDF config. We want to log out those users.
3. Send a SyncSettings push notification in order to force clients to sync new MasterPasswordUnlockData - new KDF settings will be picked up by all client.

Note for Platform: See point 2. for reasoning behind the Logout push notification payload change. I have checked all clients and we do not use the "Date" field, hence removed it.

A client's followup: [#16775](https://github.com/bitwarden/clients/pull/16775)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


Patrick-Pimentel-Bitwarden (dismissed):
👍 Auth changes look good.

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
```diff
@@ -35,4 +85,353 @@ public async Task GetAccountsProfile_success()
         Assert.NotNull(content.PrivateKey);
         Assert.NotNull(content.SecurityStamp);
     }
+
+    [Theory]
+    [BitAutoData(KdfType.PBKDF2_SHA256, 600001, null, null)]
+    [BitAutoData(KdfType.Argon2id, 4, 65, 5)]
+    public async Task PostKdf_ValidRequestNoLogoutOnKdfChangeFeatureFlagOff_SuccessLogout(KdfType kdf,
```
Thomas-Avery:
```suggestion
    public async Task PostKdf_ValidRequestLogoutOnKdfChangeFeatureFlagOff_SuccessLogout(KdfType kdf,
```

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
```diff
@@ -35,4 +85,353 @@ public async Task GetAccountsProfile_success()
...
+            _passwordHasher.VerifyHashedPassword(user, user.MasterPassword!, _newMasterPasswordHash));
+
+        // Validate push notification
+        await _pushNotificationService.Received(1).PushLogOutAsync(user.Id);
+    }
+
+    [Theory]
+    [BitAutoData(KdfType.PBKDF2_SHA256, 600001, null, null)]
+    [BitAutoData(KdfType.Argon2id, 4, 65, 5)]
+    public async Task PostKdf_ValidRequestNoLogoutOnKdfChangeFeatureFlagOn_SuccessNoLogout(KdfType kdf,
```
Thomas-Avery:
```suggestion
    public async Task PostKdf_ValidRequestLogoutOnKdfChangeFeatureFlagOn_SuccessNoLogout(KdfType kdf,
```

src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs
```diff
@@ -88,7 +97,15 @@ public async Task<IdentityResult> ChangeKdfAsync(User user, string masterPasswor
...
         await _userRepository.ReplaceAsync(user);
-        await _pushService.PushLogOutAsync(user.Id);
+        if (logoutOnKdfChange)
+        {
+            await _pushService.PushLogOutAsync(user.Id);
+        }
+        else
+        {
+            await _pushService.PushSyncSettingsAsync(user.Id);
+        }
```
Thomas-Avery, line 110: ❓  We still need to send a logout push to older clients that can't handle the sync settings push right?

Or worded differently from the ticket "Update logout push notification with a way for newer clients to ignore."

Thomas-Avery (changes requested):
Have some concerns with how this will work on older clients with the flag turned on.

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
mzieniukbw: Fixed in f5965c5f

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
mzieniukbw: Fixed in f5965c5f

src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs
mzieniukbw: I understood this as a note, not as a task, since once the feature flag is enabled, we will target specific client's version. Older client's will still log out and newer clients will not.
If we are on very old server, then the feature flag would not exist anyway, even if enabled in LaunchDarkly, so we will logout. Although that scenario is technically unsupported, since we are targeting 3 major versions of server to client compatibility, hence this is expected to be broken - logout.
If the client is old, hence it does not know about the feature flag, it will logout.

mzieniukbw: Adressed in 88acd349.
We have to handle this scenario by making sure older clients get's logged out and prevent them going into bad state.

src/Core/Models/PushNotification.cs
```diff
@@ -97,3 +97,9 @@ public class ProviderBankAccountVerifiedPushNotification
     public Guid ProviderId { get; set; }
     public Guid AdminId { get; set; }
 }
+
+public class LogOutPushNotification
+{
+    public Guid UserId { get; set; }
+    public string? Reason { get; set; }
```
Thomas-Avery: I would suggest an enum for the `Reason` to keep things consistent instead of magic strings.

We will always want to match on the client for making a decision based on a specific reason.

mzieniukbw: Fixed in [1befbb8](https://github.com/bitwarden/server/pull/6349/commits/1befbb8f478dee1161a646256caeac302bf4b732)

mzieniukbw:
Something went wrong during merging, resolving that now.

PR #6350, opened 2025-09-18
removing status from org name.

jrmccannon:
## 🎟️ Tracking
[PM-25934](https://bitwarden.atlassian.net/browse/PM-25934)

## 📔 Objective
This removes invited from showing up appeneded to the end of the subject.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Models/Mail/OrganizationUserInvitedViewModel.cs
+++ b/src/Core/Models/Mail/OrganizationUserInvitedViewModel.cs
@@ -31,7 +31,7 @@ public class OrganizationUserInvitedViewModel : BaseTitleContactUsMailModel
                     ? string.Empty
                     : CoreHelpers.SanitizeForEmail(orgInvitesInfo.OrganizationName, false),
             TitleThird = orgInvitesInfo.IsFreeOrg ? string.Empty : " on Bitwarden and start securing your passwords!",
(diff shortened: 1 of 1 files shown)


jrmccannon:
> Should we do the fix to guard this behind the feature flag here as well? Or do we just want to address the blocker and do that in a followup?
The previous implementation used a variable from a higher scope to build the subject string.

HandlebarsMailService.cs
```
    public async Task SendOrganizationInviteEmailsAsync(OrganizationInvitesInfo orgInvitesInfo)
    {
        MailQueueMessage CreateMessage(string email, object model)
        {
            var message = CreateDefaultMessage($"Join {orgInvitesInfo.OrganizationName}", email);
            return new MailQueueMessage(message, "OrganizationUserInvited", model);
        }
```

Here it uses the orgInvitesInfo parameter in the local function scope to build the subject for the email. While we could go back and move this to be within the scope of the feature flag, I'm not sure its worth the time. I went and looked at what OrganizationName was being used for in the code previously and I can't find a use for it. It was not used in the template and nor the Url builder. It previously would have been the orgName concatenated with the status (invited). I wouldn't have had a space at all.  🤷 

I can go back and leave the previous function and create a new v2 function that does all the things we need it to and completely wrap the change in a feature flag if that's what we wish. I have no strong feelings.

PR #6354, opened 2025-09-18
Marked structured fields as obsolete and add Data field to the reques…

gbubemismith:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Vault/Models/Request/CipherRequestModel.cs
+++ b/src/Api/Vault/Models/Request/CipherRequestModel.cs
@@ -40,11 +40,26 @@ public class CipherRequestModel
     // TODO: Rename to Attachments whenever the above is finally removed.
     public Dictionary<string, CipherAttachmentModel> Attachments2 { get; set; }
 
+    [Obsolete("This property is deprecated and will be removed in an upcoming release.")]
     public CipherLoginModel Login { get; set; }
+
+    [Obsolete("This property is deprecated and will be removed in an upcoming release.")]
     public CipherCardModel Card { get; set; }
+
+    [Obsolete("This property is deprecated and will be removed in an upcoming release.")]
     public CipherIdentityModel Identity { get; set; }
+
+    [Obsolete("This property is deprecated and will be removed in an upcoming release.")]
     public CipherSecureNoteModel SecureNote { get; set; }
+
+    [Obsolete("This property is deprecated and will be removed in an upcoming release.")]
     public CipherSSHKeyModel SSHKey { get; set; }
+
+    /// <summary>
+    /// Opaque JSON blob containing all cipher-specific data
+    /// </summary>
+    [StringLength(50000)]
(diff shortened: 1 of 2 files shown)


src/Api/Vault/Models/Request/CipherRequestModel.cs
```diff
@@ -40,11 +38,26 @@ public class CipherRequestModel
...
+    [Obsolete("Use Data instead.")]
     public CipherSecureNoteModel SecureNote { get; set; }
+
+    [Obsolete("Use Data instead.")]
     public CipherSSHKeyModel SSHKey { get; set; }
+
+    /// <summary>
+    /// JSON string containing cipher-specific data
+    /// </summary>
+    [StringLength(50000)]
```
nikwithak: Is this string length enforced elsewhere, or is there an edge case if a user stores a reaaaally long data string? 

I just tested locally, and was able to exceed 50k characters by creating a note with the max characters (10k), and several custom fields with 5k characters each. Do we have an actual enforced limit of the data string defined anywhere, or would this introduce one?

gbubemismith: I was trying to limit the data string to 50kb @shane-melton do you know if users get above this limit? maybe we can increase to 100kb

shane-melton: We have some pretty generous limits on the cipher models being deprecated, and after looking over them, 50k is likely not enough. 

For example: https://github.com/bitwarden/server/blob/50461518e7be230a50319193b74244380e2b3575/src/Api/Vault/Models/CipherLoginModel.cs#L84-L89

Looking at the `Cipher` table, the `Data` column is defined as `NVARCHAR (MAX)` which [looks appears to hold ~2Gb](https://learn.microsoft.com/en-us/sql/t-sql/data-types/nchar-and-nvarchar-transact-sql?view=sql-server-ver17#nvarchar---n--max--) of data. 

~All that to say, do we even need a limit on `Data` now?~ Since we're deprecating these request models and their per-field limits entirely, we'll need a new reasonable limit on this `Data` property.

PJHOMEZ: ok

shane-melton (approved):

src/Api/Vault/Models/Request/CipherRequestModel.cs
roelzkie15: 👋🏼  Noob question _(new to exploring Bitwarden codebase)_. Is there a reason why the `Data` field doesn't have an `EncryptedString` decoration _(sorry, I'm not a C# expert but very curious)_?

Or maybe each value of the JSON properties is already encrypted.

Thank you in advance for taking the time to check my comment. 🙏🏼

PR #6363, opened 2025-09-22
[PM-25463] Work towards complete usage of Payments domain

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25463

## 📔 Objective

This PR makes the `server`-side changes necessary to support the `clients` changes in this [PR](https://github.com/bitwarden/clients/pull/16532).

Those include:
1. Updating the provider setup process to use Billing's modern payment request models. This includes modifying the `ProvidersController`, the request model, the `ProviderService` and the `ProviderBillingService`. 
2. Introducing the `PreviewOrganizationTaxCommand` in order to consolidate the locations where we'll make tax estimations on behalf of an organization. The 3 current scenarios wired up are listed below. They directly support specific tax scenarios called out in the linked `clients` PR.
    - Purchasing an organization subscription
    - Changing an organization's plan
    - Updating an organization's subscription
3. Moving the Restart Subscription logic into its own command
4. Removing the `pm-21881-manage-payment-details-outside-checkout` feature flag

There are also many instances where I remove `#nullable enable` from files as it's no longer needed, which has inflated the changed file count.

## 📸 Screenshots

Associated Clients PR: https://github.com/bitwarden/clients/pull/16532

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
```diff
@@ -38,6 +37,9 @@
 
 namespace Bit.Commercial.Core.Billing.Providers.Services;
 
+using static Constants;
```
sbrown-livefront, line 40: Good call bringing these statics in

src/Api/Billing/Controllers/TaxController.cs
```diff
@@ -1,33 +1,73 @@
...
+[Route("billing/tax")]
 public class TaxController(
-    IPreviewTaxAmountCommand previewTaxAmountCommand) : BaseBillingController
+    IPreviewOrganizationTaxCommand previewOrganizationTaxCommand,
+    IPreviewPremiumTaxCommand previewPremiumTaxCommand) : BaseBillingController
 {
-    [HttpPost("preview-amount/organization-trial")]
-    public async Task<IResult> PreviewTaxAmountForOrganizationTrialAsync(
-        [FromBody] PreviewTaxAmountForOrganizationTrialRequestBody requestBody)
+    [HttpPost("organizations/subscriptions/purchase")]
```
sbrown-livefront, line 18: Should this be subscription instead of subscriptions?

src/Api/Billing/Models/Requests/Payment/TokenizedPaymentMethodRequest.cs
```diff
@@ -1,31 +1,15 @@
-﻿#nullable enable
-using System.ComponentModel.DataAnnotations;
-using Bit.Api.Billing.Attributes;
-using Bit.Core.Billing.Payment.Models;
+﻿using Bit.Core.Billing.Payment.Models;
 
 namespace Bit.Api.Billing.Models.Requests.Payment;
 
-public class TokenizedPaymentMethodRequest
+public class TokenizedPaymentMethodRequest : MinimalTokenizedPaymentMethodRequest
```
sbrown-livefront, line 5: Is this for backwards compatibility?

src/Api/Billing/Controllers/TaxController.cs
amorask-bitwarden: Had a hard time with this one, but since there's no `organizationId` being passed (it's for purchasing organization subscriptions), I think plural **might** make sense? The entire tax collection paradigm's a bit of an oddball because  our implementations require they be coupled to their scenarios. TL/DR I think this probably works for the time being but we might want to revisit; especially in the case we can consolidate down to a single, extensible tax endpoint.

src/Api/Billing/Models/Requests/Payment/TokenizedPaymentMethodRequest.cs
amorask-bitwarden: No it's for composability. An actual payment method update actually sends a minimal billing address along with the tokenized payment method because you can let Stripe know the billing address for that specific payment method. 

However, many other cases require both a tokenized payment method and a full or checkout billing address, so we needed a representation that could support that (the `MinimalTokenizedPaymentMethodRequest`).

Once we had that, there was no sense in duplicating the Properties and `ToDomain` for the request models if one was a superclass of the other.

amorask-bitwarden:
@sbrown-livefront Thanks for catching the test miss! Added those.

src/Api/Billing/Controllers/TaxController.cs
sbrown-livefront: Ah, then yeah I definitely agree plural makes more sense here.

sbrown-livefront (approved):

PR #6366, opened 2025-09-23
[PM-26050] Migrate all DefaultUserCollection when claimed user is deleted

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26050

## 📔 Objective

Unclaimed users already had their `DefaultUserCollections` migrated when their `OrganizationUser` entries were deleted via `OrganizationUserRepository.DeleteAsync` and `OrganizationUserRepository.DeleteManyAsync`.

However, deleting a claimed user removes the `User` entry entirely through `UserRepository.DeleteAsync` or `UserRepository.DeleteManyAsync`, that were not updated to handle `DefaultUserCollection` migration.

This PR updates the `UserRepository` Delete logic to ensure that `DefaultUserCollections` are also migrated when a claimed user is deleted.

## 📸 Screenshots

https://github.com/user-attachments/assets/5c61ec49-f0d7-40e6-8fc8-96297fc1bdb9

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
r-tome, line 1: I had to change the `Remove` methods to `ExecuteDeleteAsync` because I was getting some concurrency issues when running the integration tests.

test/Infrastructure.IntegrationTest/Auth/Repositories/UserRepositoryTests.cs
eliykat, line 1: Please use the `OrganizationTestHelpers` extension methods in this repository. e.g.

```c#
var user = organizationUserRepository.CreateTestUserAsync();
```

test/Infrastructure.IntegrationTest/Auth/Repositories/UserRepositoryTests.cs
```diff
@@ -96,4 +97,171 @@ await userRepository.DeleteManyAsync(new List<User>
...
+        await collectionRepository.UpdateUsersAsync(defaultUserCollection.Id, new List<CollectionAccessSelection>()
+        {
+            new CollectionAccessSelection
+            {
+                Id = orgUser.Id,
+                HidePasswords = false,
+                ReadOnly = false,
+                Manage = true
+            },
+        });
```
eliykat: There's an overload of `collectionRepository.CreateAsync` that also takes access: https://github.com/bitwarden/server/blob/780400fcf9ed8a7f4650e01ac825a9a480e49121/src/Core/Repositories/ICollectionRepository.cs#L57

Same for next test method

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -283,6 +283,26 @@ public override async Task DeleteAsync(Core.Entities.User user)
...
+                foreach (var collection in defaultCollections)
+                {
+                    collection.Type = Core.Enums.CollectionType.SharedCollection;
+                    collection.DefaultUserCollectionEmail = collection.DefaultUserCollectionEmail ?? dbContext.Users
+                        .Where(u => u.Id == user.Id)
+                        .Select(u => u.Email)
+                        .FirstOrDefault();
+                    collection.RevisionDate = DateTime.UtcNow;
+                }
+            }
```
eliykat: I think this could be simplified. At the moment it has nested `foreach` loops, and then a db query inside the inner loop. Also nested `Where` and `Any`.

While I generally like the LINQ method syntax, I do think the query syntax is easier for joins. As a quick example (please verify):

```c#
            // Migrate all DefaultUserCollection to SharedCollection
            var defaultCollections = (from c in dbContext.Collections
                                      join cu in dbContext.CollectionUsers on c.Id equals cu.CollectionId
                                      join ou in dbContext.OrganizationUsers on cu.OrganizationUserId equals ou.Id
                                      where c.Type == Core.Enums.CollectionType.DefaultUserCollection && ou.UserId == user.Id
                                      select c)
                                     .ToList();

            foreach (var collection in defaultCollections)
            {
                collection.Type = Core.Enums.CollectionType.SharedCollection;
                collection.DefaultUserCollectionEmail = collection.DefaultUserCollectionEmail ?? user.Email;
                collection.RevisionDate = DateTime.UtcNow;
            }
```

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -329,21 +349,39 @@ public async Task DeleteManyAsync(IEnumerable<Core.Entities.User> users)
...
+                foreach (var collection in defaultCollections)
+                {
+                    collection.Type = Core.Enums.CollectionType.SharedCollection;
+                    collection.DefaultUserCollectionEmail = collection.DefaultUserCollectionEmail ?? dbContext.Users
+                        .Where(u => u.Id == orgUser.UserId)
+                        .Select(u => u.Email)
+                        .FirstOrDefault();
+                    collection.RevisionDate = DateTime.UtcNow;
+                }
+            }
```
eliykat: Similar to previous comment:

```c#
            // Migrate all DefaultUserCollection to SharedCollection
            var defaultCollections = (from c in dbContext.Collections
                                      join cu in dbContext.CollectionUsers on c.Id equals cu.CollectionId
                                      join ou in dbContext.OrganizationUsers on cu.OrganizationUserId equals ou.Id
                                      join u in dbContext.Users on ou.UserId equals u.Id
                                      where c.Type == Core.Enums.CollectionType.DefaultUserCollection
                                         && targetIds.Contains(u.Id)
                                      select new { Collection = c, UserEmail = u.Email }).ToList();

            foreach (var item in defaultCollections)
            {
                item.Collection.Type = Core.Enums.CollectionType.SharedCollection;
                item.Collection.DefaultUserCollectionEmail ??= item.UserEmail;
                item.Collection.RevisionDate = DateTime.UtcNow;
            }
```

eliykat: I suggest this could be moved into its own private method and used by both delete user methods.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -329,21 +349,39 @@ public async Task DeleteManyAsync(IEnumerable<Core.Entities.User> users)
...
-                             join ou in dbContext.OrganizationUsers on gu.OrganizationUserId equals ou.Id
-                             where targetIds.Contains(ou.UserId ?? default)
-                             select gu;
-            dbContext.GroupUsers.RemoveRange(groupUsers);
+            await dbContext.CollectionUsers
+                .Where(cu => dbContext.OrganizationUsers.Any(ou => ou.Id == cu.OrganizationUserId && targetIds.Contains(ou.UserId ?? default)))
+                .ExecuteDeleteAsync();
+            await dbContext.GroupUsers
+                .Where(gu => dbContext.OrganizationUsers.Any(ou => ou.Id == gu.OrganizationUserId && targetIds.Contains(ou.UserId ?? default)))
+                .ExecuteDeleteAsync();
```
eliykat, line 358: It would be better to avoid this nested `Where`/`Any` for both performance and readability. Either by wrapping the previous query syntax in brackets (admittedly a bit awkward), or using `.Join` instead of the nesting.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -354,15 +392,9 @@ where targetIds.Contains(ou.UserId ?? default)
...
-            foreach (var u in users)
-            {
-                var mappedUser = Mapper.Map<User>(u);
-                dbContext.Users.Remove(mappedUser);
-            }
-
+            await dbContext.Users.Where(u => targetIds.Contains(u.Id)).ExecuteDeleteAsync();
 
             await transaction.CommitAsync();
-            await dbContext.SaveChangesAsync();
```
eliykat, line 365: I think we still want to call `SaveChangesAsync` before commit to make sure they're tracked by EF.

r-tome: I replaced `Remove` with `ExecuteDeleteAsync` so its not really needed anymore but I'll add it back again.

util/Migrator/DbScripts/2025-09-23_00_MigrateDefaultCollectionsOnUserDelete.sql
```diff
@@ -0,0 +1,325 @@
...
+            [dbo].[Cipher]
+        WHERE
+            [UserId] = @Id
+
+        SET @BatchSize = @@ROWCOUNT
+
+        COMMIT TRANSACTION User_DeleteById_Ciphers
+    END
+
+    BEGIN TRANSACTION User_DeleteById
```
mkincaid-bw, line 25: Just wanted to point out that the transaction handling in these procs do not look correct, but likely aren't a huge issue either (seems they've been this way for a while). The `BEGIN/COMMIT TRANSACTION` in the loop aren't doing anything since the only committable statement is the delete statement, and that would be handled implicitly anyway. What should likely be changed is those statements should be removed, and the `BEGIN TRANSACTION` here should be moved above the loop. This would ensure that the entire proc completes all-or-nothing, leaving things in a consistent state.

With the current logic, if this larger transaction rolls back, all of the deleted rows in the above loop will have still been committed, and I'm not sure if that logic is correct.

However, I realize that those types of changes are outside the scope of this PR and I'm not requesting changes, but I did want to point it out for potential future changes.

r-tome: Thats a valuable assessment, thanks for pointing it out! I agree its out of scope for this PR but I have created a ticket to track it https://bitwarden.atlassian.net/browse/PM-26393

PR #6368, opened 2025-09-23
[PM-20439] Remove flagging logic (BrowserExtensionLoginApproval)

rr-bw:
## 🎟️ Tracking

[PM-20439](https://bitwarden.atlassian.net/browse/PM-20439)

Client PR: https://github.com/bitwarden/clients/pull/16568

## 📔 Objective

- Removes flagging logic for the `BrowserExtensionLoginApproval` feature flag.
- Does NOT remove the flag itself from `FeatureFlagKeys` in `Contants.cs`

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-20439]: https://bitwarden.atlassian.net/browse/PM-20439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Api/Auth/Controllers/AuthRequestsController.cs
+++ b/src/Api/Auth/Controllers/AuthRequestsController.cs
@@ -3,7 +3,6 @@
 
 using Bit.Api.Auth.Models.Response;
 using Bit.Api.Models.Response;
-using Bit.Core;
 using Bit.Core.Auth.Enums;
 using Bit.Core.Auth.Models.Api.Request.AuthRequest;
 using Bit.Core.Auth.Services;
@@ -11,7 +10,6 @@ using Bit.Core.Exceptions;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
 using Bit.Core.Settings;
-using Bit.Core.Utilities;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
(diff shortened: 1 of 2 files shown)


src/Api/Auth/Controllers/AuthRequestsController.cs
```diff
@@ -54,7 +52,6 @@ public async Task<AuthRequestResponseModel> Get(Guid id)
     }
 
     [HttpGet("pending")]
-    [RequireFeature(FeatureFlagKeys.BrowserExtensionLoginApproval)]
```
enmande, line 57: 🤔  To ensure I understand the correct order of operations for these types of PRs (first one of these I've looked at)
We are not removing the key from Constants yet because the server needs to be able to support providing flags to clients (given flags are consumed on clients in a given piece of work) in an [_n_-2 releases support window](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1152417888/Feature+Flags+Lifecycle#Step-2%3A-Remove-flag-from-server), the related clients PR will phase out its use in _latest_ upon release, and other sub-tasks on this issue like [PM-20443](https://bitwarden.atlassian.net/browse/PM-20443) and [PM-20444](https://bitwarden.atlassian.net/browse/PM-20444) are provided so we can arrange the timing of removing the constant and the data source.
Is that correct?

rr-bw: @enmande Per discussion: we use the server flags to feed into LaunchDarkly (even if the server does not have any logic that uses the flag). Those flags get sent to the client. If we were to immediately remove the flag itself from the server in this PR, then any user who is on a previous client that still has flagging logic would get that feature turned off. So we wait a bit before removing the flag itself from the server.

enmande (approved):

PR #6371, opened 2025-09-24
Begin pilot program for Claude code reviews with initial system prompt

theMickster:
## 🎟️ Tracking

TBD

## 📔 Objective

To begin a pilot program to evaluate, refine, train, and iterate towards meaningful Claude code reviews. The vault team is the first team through the wall, and the committed code is the starting point. 

I'm not looking for the _perfect_ Claude system prompt today, simply a valid, workable pilot file. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,215 @@
+# Bitwarden Server - Claude Code Configuration
+
+## Project Overview
+
+The Bitwarden Server project contains the APIs, database, and other core infrastructure items needed for the "backend" of all bitwarden client applications. The server project is written in C# using .NET Core with ASP.NET Core. The database is written in T-SQL/SQL Server. The codebase can be developed, built, run, and deployed cross-platform on Windows, macOS, and Linux distributions.
+
+This is a security-focused password management backend that handles sensitive user data and cryptographic operations. All code changes must prioritize security, performance, and reliability.
+
+## Technology Stack
+
+- **Backend Framework**: ASP.NET Core (.NET Core)
+- **Language**: C#
+- **Database**: SQL Server with T-SQL
+- **Container Platform**: Docker
+- **Architecture**: Distributed Monolith with some API delineation between the various APIs
+- **Authentication**: Custom identity system with cryptographic vault operations
+
+## Development Environment Setup
+
+### Prerequisites
+
+- .NET Core SDK (latest LTS version)
+- SQL Server or SQL Server Express
+- Docker and Docker Compose
+- Git with pre-commit hooks enabled
+
+### Environment Commands
+
+```bash
+# Set up git pre-commit hooks for automatic formatting
+git config --local core.hooksPath .git-hooks
+
+# Docker development environment
+docker-compose up -d
+
+# Database migrations (if applicable)
+dotnet ef database update
+
+# Run tests
+dotnet test
+
+# Build solution
+dotnet build
+
+# Run specific project
+dotnet run --project src/Api
+```
+
+## Project Structure
+
+```
+/src/
+├── Api/                # Main API endpoints
+├── Core/               # Business logic and domain models
+├── Infrastructure/     # Data access and external services
+├── Identity/           # Authentication and authorization
+├── Admin/              # Administrative interfaces
+├── Notifications/      # Push notifications and messaging
+├── Events/             # Event logging and auditing
+└── Sql/                # Database scripts and migrations
+
+/util/                  # Utility scripts and tools
+/dev/                   # Development configuration
+/scripts/               # Deployment and maintenance scripts
+```
+
+## Security Definitions
+
+Apply [Bitwarden security definitions](https://contributing.bitwarden.com/architecture/security/definitions).
+
+## Code Style
+
+Follow [Bitwarden code style standards](https://contributing.bitwarden.com/contributing/code-style/).
+
+## Requirements
+
+- Bitwarden CLI (`bw`) installed and configured
+- Node.js 22 with ES modules
+- BW_SESSION environment variable
+
+## File Boundaries & Restrictions
+
+### Safe to Edit
+
+- `/src/` - All source code files
+- `/test/` - Unit and integration tests
+- `/util/` - Utility scripts
+- `README.md` and documentation files
+- `/dev/` - Development configuration
+
+### Never Edit
+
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+**Architecture**: CQRS pattern with feature-based organization
+**Framework**: .NET 8.0, ASP.NET Core
+**Database**: SQL Server primary, EF Core supports PostgreSQL, MySQL/MariaDB, SQLite
+**Testing**: xUnit 2.4.1, NSubstitute
+**Container**: Docker, Docker Compose, Kubernetes/Helm deployable
+
+## Technology Stack
+
+- **Backend Framework**: ASP.NET Core & .NET Core written in C#
+- **Services**: Api, Identity, Admin, Notifications, Events, EventsProcessor
```
withinfocus: ❌ This is not the whole list. I am not sure it's needed though.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+- **Auth**: Custom identity system, JWT tokens, PBKDF2-SHA256
+- **Encryption**: AES-256-CBC, RSA-2048, zero-knowledge architecture
+- **CI/CD**: GitHub Actions - build, test (xUnit), lint (dotnet format), security (Checkmarx)
```
withinfocus: 🎨 Let's get part of this elaborated upon. You can leave the "GitHub Actions" part here but then have a little more detail in the section below for what should run locally -- really the lint, build, and test steps. Speaking about what's in the Actions themselves seems unnecessary.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+docker compose --profile cloud --profile mail up -d
+
+# Migrate database
+cd dev
+pwsh migrate.ps1
+
+# Run services
+cd src/Api && dotnet run
+cd src/Identity && dotnet run
+```
```
withinfocus: 🎨 Just link to the contributing docs getting started section for server.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Project Structure
```
withinfocus: 💭 There's a lot more to this. Wondering how much we explain.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+- Error handling implemented
+- Breaking changes documented
+- CI passes: build, test, lint
+
+## References
+
+- Architecture: https://contributing.bitwarden.com/architecture/server/
+- Setup Guide: https://contributing.bitwarden.com/getting-started/server/guide/
+- Code Style: https://contributing.bitwarden.com/contributing/code-style/
+- Security: https://bitwarden.com/help/bitwarden-security-white-paper/
```
withinfocus: 🎨 Embed these as links where they are referred to instead of having this section. See https://github.com/bitwarden/mcp-server/blob/main/CLAUDE.md as one example. The code style and security definitions are critical.

CLAUDE.md
theMickster: I originally had that in the file (based upon your work in the mcp server repo), however, upon my research (and working with Claude itself) I came to the conclusion that keeping-it-simple-silly with limited instructions would reduce token usage. In your workings with Claude, did you see any issues with linking to our wikis (or any)? Also trying to limit scope and prompt bloat, too. Thoughts?

CLAUDE.md
theMickster: 💯 agree that there's way more, but my plan was to work with subject matter experts in the coming weeks to get those into the specific project folders.  My thought was to build off the knowledge in the contributing docs + ReadMe docs + Slack threads with teammates to build strong, focused ones deeper in the repo. 

Thoughts on this for the first run?

CLAUDE.md
theMickster: I agree and see both sides. I trimmed down things to keep it limited in scope while still trying to give Claude direction on what's to be found in the repo. I also read through several _starred_ repos in GitHub (and negotiating with Claude itself, lol) and settled on this list. 
Would these be better to have down deeper in the project structure in the more specific Claude.md? Or are you thinking just remove them and see if Claude figures this out on its own?

CLAUDE.md
```diff
@@ -1,215 +1,98 @@
...
-- Documentation updated accordingly
-- Breaking changes properly communicated
-
-## Additional Notes
-
-- This is an open-source project with enterprise customers
-- Security vulnerabilities should be reported privately
-- Please refer to the Server Setup Guide in the Contributing Documentation for build instructions, recommended tooling, code style tips, and lots of other great information to get you started.
-- Consider the impact on self-hosted deployments when making infrastructure changes
-- Maintain backward compatibility for existing API clients
```
ariane-bitwarden: (Moving my comment to the PR.)
These notes seemed important to include and are not summarized or captured elsewhere:
- This is an open-source project with enterprise customers
- Consider the impact on self-hosted deployments when making infrastructure changes
- Maintain backward compatibility for existing API clients

theMickster: I went back and forth on whether or not to keep or remove these statements. I see their value, but should they be included with every PR evaluation, or would Claude be smart enough now (or in the future) to just grasp those important details?

CLAUDE.md
withinfocus: More the latter -- it will discover them and then find lower-level Claude instructions so let's omit the list here.

CLAUDE.md
withinfocus: I haven't seen issues linking out, but for the monolith we'll just have to see how chatty (and expensive) it gets. For a repo this large I think we'll have to see how far it goes, and I am not sure how far it should get on every change -- do we really want it trying to run migrations? Therefore a link to setup is perhaps enough? Not yet sure.

CLAUDE.md
withinfocus: I think I mostly dislike -- and per my other comment -- that we're providing a selective list. List every source folder and explain its purpose here, then eliminate the service list higher up. Alternatively, don't explain _any_ of the source folders because we can expect eventual development of Claude files inside each. I think the alternative is more manageable.

theMickster: Removed

CLAUDE.md
theMickster: Removed

PR #6382, opened 2025-09-25
[PM-25982] Restrict Ciphers being assigned to Default from Shared collections

nick-livefront:
## 🎟️ Tracking

[PM-25982](https://bitwarden.atlassian.net/browse/PM-25982)
Client PR: https://github.com/bitwarden/clients/pull/16591

## 📔 Objective

When assigning a cipher to a collection, the cipher should not be allowed to be placed in a default collection when it is already only assigned to Shared Collections.
- The UI won't have this ability after https://github.com/bitwarden/clients/pull/16591 is merged, this PR is to protect from hitting the API directly. 

## 📸 Screenshots

Using the UI without restrictions for ease of testing, the UI selection and error will not be a flow that users hit: 
<video src="https://github.com/user-attachments/assets/6b6fa224-d417-4694-a116-dc9eb2a9c5ea" />

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Vault/Services/Implementations/CipherService.cs
+++ b/src/Core/Vault/Services/Implementations/CipherService.cs
@@ -170,6 +170,7 @@ public class CipherService : ICipherService
         {
             ValidateCipherLastKnownRevisionDateAsync(cipher, lastKnownRevisionDate);
             cipher.RevisionDate = DateTime.UtcNow;
+            await ValidateChangeInCollectionsAsync(cipher, collectionIds, savingUserId);
             await ValidateViewPasswordUserAsync(cipher);
             await _cipherRepository.ReplaceAsync(cipher);
             await _eventService.LogCipherEventAsync(cipher, Bit.Core.Enums.EventType.Cipher_Updated);
@@ -539,6 +540,7 @@ public class CipherService : ICipherService
         try
         {
             await ValidateCipherCanBeShared(cipher, sharingUserId, organizationId, lastKnownRevisionDate);
+            await ValidateChangeInCollectionsAsync(cipher, collectionIds, sharingUserId);
 
             // Sproc will not save this UserId on the cipher. It is used limit scope of the collectionIds.
             cipher.UserId = sharingUserId;
@@ -678,6 +680,7 @@ public class CipherService : ICipherService
         {
             throw new BadRequestException("Cipher must belong to an organization.");
         }
+        await ValidateChangeInCollectionsAsync(cipher, collectionIds, savingUserId);
 
         cipher.RevisionDate = DateTime.UtcNow;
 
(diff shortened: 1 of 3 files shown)


src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -1038,6 +1050,37 @@ private async Task ValidateViewPasswordUserAsync(Cipher cipher)
         }
     }
 
+    // Validates that a cipher is not being added to a default collection when it is only currently only in shared collections
+    private async Task ValidateChangeInCollectionsAsync(Cipher updatedCipher, IEnumerable<Guid> newCollectionIds, Guid userId)
+    {
+        var currentCollectionsForCipher = await _collectionCipherRepository.GetManyByUserIdCipherIdAsync(userId, updatedCipher.Id);
+
+        if (updatedCipher.Id == Guid.Empty || !updatedCipher.OrganizationId.HasValue || !currentCollectionsForCipher.Any())
```
shane-melton: ⛏️ We can check for an empty cipherId or organizationId before we make the repository call above. `_collectionCipherRepository.GetManyByUserIdCipherIdAsync`

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -1038,6 +1050,37 @@ private async Task ValidateViewPasswordUserAsync(Cipher cipher)
...
+        // a default collection, because it is already there.
+        if (currentCollectionsContainDefault)
+        {
+            return;
+        }
+
+        var newCollections = await _collectionRepository.GetManyByManyIdsAsync(newCollectionIds);
+        var newCollectionsContainDefault = newCollections.Any(c => c.Type == CollectionType.DefaultUserCollection);
+
+        if (newCollectionsContainDefault && !currentCollectionsContainDefault)
```
shane-melton: ⛏️ (non-blocking) `!currentCollectionsContainDefault` will always be true due to early return above that checks for `currentCollectionsContainDefault`

shane-melton (changes requested):
Looks good, just two small ⛏️s

src/Core/Vault/Services/Implementations/CipherService.cs
nick-livefront: [b1e1b5c](https://github.com/bitwarden/server/pull/6382/commits/b1e1b5c73bd918019eba37595306b6ad4bd745fd) - I think I'm following the request here. This avoids any unneeded DB calls correct?

src/Core/Vault/Services/Implementations/CipherService.cs
nick-livefront: True! [5e2debe](https://github.com/bitwarden/server/pull/6382/commits/5e2debe81d3dc550804af8d08785d00f3f9b54d4)

src/Core/Vault/Services/Implementations/CipherService.cs
shane-melton: Correct, thanks!

shane-melton (approved):

PR #6385, opened 2025-09-26
[PM-25923] Simplify and align ProfileOrganizationResponseModel and ProfileProviderOrganizationResponseModel

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25923

## 📔 Objective

Ensure `ProfileOrganizationResponseModel` and `ProfileProviderOrganizationResponseModel` have consistent properties and simplify their implementation.

Enforced consistency on both models through shared `BaseProfileOrganizationResponseModel` inheritance.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/AdminConsole/Models/Response/BaseProfileOrganizationResponseModel.cs
+++ b/src/Api/AdminConsole/Models/Response/BaseProfileOrganizationResponseModel.cs
@@ -1,0 +1,139 @@
+﻿// FIXME: Update this file to be null safe and then delete the line below
+#nullable disable
+
+using System.Text.Json.Serialization;
+using Bit.Core.AdminConsole.Enums.Provider;
+using Bit.Core.AdminConsole.Models.Data;
+using Bit.Core.Auth.Enums;
+using Bit.Core.Auth.Models.Data;
+using Bit.Core.Billing.Enums;
+using Bit.Core.Billing.Extensions;
+using Bit.Core.Enums;
+using Bit.Core.Models.Api;
+using Bit.Core.Models.Data;
+using Bit.Core.Utilities;
+
+namespace Bit.Api.AdminConsole.Models.Response;
+
+/// <summary>
+/// Base class for profile organization response models that ensures all properties are fully populated.
+/// This class cannot be instantiated with an empty constructor to prevent incomplete hydration.
+/// </summary>
+public abstract class BaseProfileOrganizationResponseModel : ResponseModel
+{
+    protected BaseProfileOrganizationResponseModel(string type, BaseUserOrganizationDetails organization) : base(type)
+    {
+        Id = organization.OrganizationId;
+        Name = organization.Name;
+        Enabled = organization.Enabled;
+        Identifier = organization.Identifier;
+        ProductTierType = organization.PlanType.GetProductTier();
+        UsePolicies = organization.UsePolicies;
+        UseSso = organization.UseSso;
+        UseKeyConnector = organization.UseKeyConnector;
+        UseScim = organization.UseScim;
+        UseGroups = organization.UseGroups;
+        UseDirectory = organization.UseDirectory;
+        UseEvents = organization.UseEvents;
+        UseTotp = organization.UseTotp;
+        Use2fa = organization.Use2fa;
+        UseApi = organization.UseApi;
+        UseResetPassword = organization.UseResetPassword;
+        UsersGetPremium = organization.UsersGetPremium;
+        UseCustomPermissions = organization.UseCustomPermissions;
+        UseActivateAutofillPolicy = organization.PlanType.GetProductTier() == ProductTierType.Enterprise;
+        UseRiskInsights = organization.UseRiskInsights;
+        UseOrganizationDomains = organization.UseOrganizationDomains;
+        UseAdminSponsoredFamilies = organization.UseAdminSponsoredFamilies;
+        SelfHost = organization.SelfHost;
+        Seats = organization.Seats;
+        MaxCollections = organization.MaxCollections;
+        MaxStorageGb = organization.MaxStorageGb;
+        Key = organization.Key;
+        HasPublicAndPrivateKeys = organization.PublicKey != null && organization.PrivateKey != null;
+        ProviderId = organization.ProviderId;
+        ProviderName = organization.ProviderName;
+        ProviderType = organization.ProviderType;
+        LimitCollectionCreation = organization.LimitCollectionCreation;
+        LimitCollectionDeletion = organization.LimitCollectionDeletion;
+        LimitItemDeletion = organization.LimitItemDeletion;
+        AllowAdminAccessToAllCollectionItems = organization.AllowAdminAccessToAllCollectionItems;
+        SsoEnabled = organization.SsoEnabled ?? false;
+        if (organization.SsoConfig != null)
+        {
+            var ssoConfigData = SsoConfigurationData.Deserialize(organization.SsoConfig);
+            KeyConnectorEnabled = ssoConfigData.MemberDecryptionType == MemberDecryptionType.KeyConnector && !string.IsNullOrEmpty(ssoConfigData.KeyConnectorUrl);
+            KeyConnectorUrl = ssoConfigData.KeyConnectorUrl;
+            SsoMemberDecryptionType = ssoConfigData.MemberDecryptionType;
+        }
+    }
+
+    public Guid Id { get; set; }
+    [JsonConverter(typeof(HtmlEncodingStringConverter))]
+    public string Name { get; set; }
+    public bool Enabled { get; set; }
+    public string Identifier { get; set; }
+    public ProductTierType ProductTierType { get; set; }
+    public bool UsePolicies { get; set; }
+    public bool UseSso { get; set; }
+    public bool UseKeyConnector { get; set; }
+    public bool UseScim { get; set; }
+    public bool UseGroups { get; set; }
+    public bool UseDirectory { get; set; }
+    public bool UseEvents { get; set; }
+    public bool UseTotp { get; set; }
+    public bool Use2fa { get; set; }
+    public bool UseApi { get; set; }
+    public bool UseResetPassword { get; set; }
+    public bool UseSecretsManager { get; set; }
+    public bool UsePasswordManager { get; set; }
+    public bool UsersGetPremium { get; set; }
+    public bool UseCustomPermissions { get; set; }
+    public bool UseActivateAutofillPolicy { get; set; }
+    public bool UseRiskInsights { get; set; }
+    public bool UseOrganizationDomains { get; set; }
+    public bool UseAdminSponsoredFamilies { get; set; }
+    public bool SelfHost { get; set; }
+    public int? Seats { get; set; }
+    public short? MaxCollections { get; set; }
+    public short? MaxStorageGb { get; set; }
+    public string Key { get; set; }
+    public bool HasPublicAndPrivateKeys { get; set; }
+    public bool SsoBound { get; set; }
+    public bool ResetPasswordEnrolled { get; set; }
+    public bool LimitCollectionCreation { get; set; }
+    public bool LimitCollectionDeletion { get; set; }
+    public bool LimitItemDeletion { get; set; }
+    public bool AllowAdminAccessToAllCollectionItems { get; set; }
+    public Guid? ProviderId { get; set; }
+    [JsonConverter(typeof(HtmlEncodingStringConverter))]
+    public string ProviderName { get; set; }
+    public ProviderType? ProviderType { get; set; }
+    public bool SsoEnabled { get; set; }
+    public bool KeyConnectorEnabled { get; set; }
+    public string KeyConnectorUrl { get; set; }
+    public MemberDecryptionType? SsoMemberDecryptionType { get; set; }
+    public string FamilySponsorshipFriendlyName { get; set; }
+    public bool FamilySponsorshipAvailable { get; set; }
+    public DateTime? FamilySponsorshipLastSyncDate { get; set; }
+    public DateTime? FamilySponsorshipValidUntil { get; set; }
+    public bool? FamilySponsorshipToDelete { get; set; }
+    public bool IsAdminInitiated { get; set; }
+    public bool AccessSecretsManager { get; set; }
+    public Guid? UserId { get; set; }
+    public Guid? OrganizationUserId { get; set; }
+    public OrganizationUserStatusType Status { get; set; }
+    public OrganizationUserType Type { get; set; }
+    public Permissions Permissions { get; set; }
+    public bool UserIsClaimedByOrganization { get; set; }
+
+    /// <summary>
+    /// Obsolete property for backward compatibility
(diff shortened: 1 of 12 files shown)


src/Api/AdminConsole/Models/Response/BaseProfileOrganizationResponseModel.cs
```diff
@@ -0,0 +1,139 @@
...
+        LimitItemDeletion = organization.LimitItemDeletion;
+        AllowAdminAccessToAllCollectionItems = organization.AllowAdminAccessToAllCollectionItems;
+        SsoEnabled = organization.SsoEnabled ?? false;
+        if (organization.SsoConfig != null)
+        {
+            var ssoConfigData = SsoConfigurationData.Deserialize(organization.SsoConfig);
+            KeyConnectorEnabled = ssoConfigData.MemberDecryptionType == MemberDecryptionType.KeyConnector && !string.IsNullOrEmpty(ssoConfigData.KeyConnectorUrl);
+            KeyConnectorUrl = ssoConfigData.KeyConnectorUrl;
+            SsoMemberDecryptionType = ssoConfigData.MemberDecryptionType;
+        }
```
jrmccannon: Would sso config be null if Sso Is enabled?

r-tome: No, it will have a value https://github.com/bitwarden/server/blob/a36340e9ad08920e85bed9c200135158fa1ed9f2/src/Api/AdminConsole/Controllers/OrganizationsController.cs#L557

src/Api/AdminConsole/Models/Response/BaseProfileOrganizationResponseModel.cs
eliykat, line 1: This aligns the definition of the models, but we still need to make sure they are consistently populated:

1. The base class shouldn't contain properties only relevant to 1 of the models. In particular, `OrganizationUserId` and `UserIsClaimedByOrganization` don't apply to providers because there is no OrgUser involved (it joins through the ProviderUser table, not OrganizationUser). So their definition and initialization would be moved to the `ProfileOrganizationResponseModel`. At the moment they are defined on the base but only populated in 1 of the concrete models, which is what we want to avoid.
2. The base class should initialize all properties relevant to both models. At the moment, there is a bunch of organization information only initialized in the `ProfileOrganizationResponseModel` (almost everything in that constructor). We need to make sure that both queries return this information and it's populated on both models.

I think the outcome would be that all properties are defined initialized in the base class, except for `OrganizationUserId` and `UserIsClaimedByOrganization`.

src/Api/AdminConsole/Models/Response/BaseProfileOrganizationResponseModel.cs
```diff
@@ -0,0 +1,136 @@
...
+using Bit.Core.Enums;
+using Bit.Core.Models.Api;
+using Bit.Core.Models.Data;
+using Bit.Core.Utilities;
+
+namespace Bit.Api.AdminConsole.Models.Response;
+
+/// <summary>
+/// Base class for profile organization response models that ensures all properties are fully populated.
+/// This class cannot be instantiated with an empty constructor to prevent incomplete hydration.
```
eliykat: Suggestion: this could provide clearer instruction. e.g.

> Contains organization properties for both OrganizationUsers and ProviderUsers.
> Any organization properties in sync data should be added to this class so they are populated for both
> members and providers.

src/Api/AdminConsole/Models/Response/ProfileOrganizationResponseModel.cs
eliykat, line 1: Suggest this is renamed `ProfileMemberOrganizationResponseModel`.

Please add xmldoc for each, explaining their purpose:

> Sync data for organization members and their organization.
> Note: see ProfileProviderOrganizationResponseModel for organization sync data received by providers.

src/Core/AdminConsole/Models/Data/BaseUserOrganizationDetails.cs
eliykat, line 1: I've been told not to use inheritance in data models before. I don't feel strongly about it, but it does make sense to me to avoid inheritance and polymorphism etc. in DTOs that just represent query results. Another way to structure it would be to have a common interface that both classes implement. Or feel free to check with dbops if you want another opinion.

util/Migrator/DbScripts/2025-10-22_00_ProviderUserOrganizationSsoData.sql
eliykat, line 1: Any queries that read from this view will need to be refreshed as well.

src/Api/AdminConsole/Models/Response/BaseProfileOrganizationResponseModel.cs
r-tome: I intended to achieve both those points but I clearly missed. Thanks for pointing this out!

src/Api/AdminConsole/Models/Response/ProfileOrganizationResponseModel.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        Status = organizationDetails.Status;
+        Type = organizationDetails.Type;
+        OrganizationUserId = organizationDetails.OrganizationUserId;
+        UserIsClaimedByOrganization = organizationIdsClaimingUser.Contains(organizationDetails.OrganizationId);
+        Permissions = CoreHelpers.LoadClassFromJsonData<Permissions>(organizationDetails.Permissions);
+        FamilySponsorshipAvailable = (organizationDetails.FamilySponsorshipFriendlyName == null || IsAdminInitiated) &&
+            StaticStore.GetSponsoredPlan(PlanSponsorshipType.FamiliesForEnterprise)
+            .UsersCanSponsor(organizationDetails);
+        AccessSecretsManager = organizationDetails.AccessSecretsManager;
```
eliykat, line 32: These properties are still defined on the base class but only initialized in this subclass. [Same issue as before.](https://github.com/bitwarden/server/pull/6385#discussion_r2389844684)

Double check me, but:
- `FamilySponsorshipAvailable` - I _think_ this indicates that the user has a family sponsorship they can redeem. Providers don't get family sponsorships so I think it's probably right that it only lives on this model. Check with Billing Team if unsure.
- `AccessSecretsManager` should be on the base class because it relates to the organization. But I think that providers don't get access to SM, so it should be set in the member model but overridden to `false` on the provider model. Check with Priya if unsure.

In both cases please add explanatory comments.

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -339,15 +339,15 @@ await _providerUserRepository.GetManyOrganizationDetailsByUserAsync(user.Id,
     }
 
     [HttpGet("organizations")]
-    public async Task<ListResponseModel<ProfileOrganizationResponseModel>> GetOrganizations()
+    public async Task<ListResponseModel<ProfileMemberOrganizationResponseModel>> GetOrganizations()
```
eliykat: Ah, sorry about the code review bingo triggered by this naming change. You can undo it if you want.

src/Core/AdminConsole/Models/Data/IProfileMemberOrganizationDetails.cs
```diff
@@ -0,0 +1,60 @@
+﻿using Bit.Core.AdminConsole.Enums.Provider;
+using Bit.Core.Billing.Enums;
+
+namespace Bit.Core.AdminConsole.Models.Data;
+
+/// <summary>
+/// Interface defining common organization details properties shared between
+/// regular organization users and provider organization users for profile endpoints.
+/// </summary>
+public interface IProfileMemberOrganizationDetails
```
eliykat: It applies to both members and providers, so `IProfileOrganizationDetails` is probably a more accurate name.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepository/OrganizationUserRepositoryTests.cs
```diff
@@ -541,6 +541,74 @@ public async Task GetManyDetailsByUserAsync_Works(IUserRepository userRepository
...
+        Assert.Null(orgWithoutSsoDetails.SsoEnabled);
+        Assert.Null(orgWithoutSsoDetails.SsoConfig);
+
+        // Cleanup
+        await ssoConfigRepository.DeleteAsync(ssoConfig);
+        await organizationUserRepository.DeleteAsync(orgUserWithSso);
+        await organizationUserRepository.DeleteAsync(orgUserWithoutSso);
+        await organizationRepository.DeleteAsync(organizationWithSso);
+        await organizationRepository.DeleteAsync(organizationWithoutSso);
+        await userRepository.DeleteAsync(user);
```
eliykat: I suggest we don't worry about cleaning up test data: https://bitwarden.slack.com/archives/C03UJFF9E9M/p1759209081233459

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepository/OrganizationUserRepositoryTests.cs
```diff
@@ -541,6 +541,74 @@ public async Task GetManyDetailsByUserAsync_Works(IUserRepository userRepository
         Assert.Equal(organization.UseAdminSponsoredFamilies, result.UseAdminSponsoredFamilies);
     }
 
+    [Theory, DatabaseData]
+    public async Task GetManyDetailsByUserAsync_ShouldPopulateSsoPropertiesCorrectly(
```
eliykat, line 542: I also think a general test would be worthwhile, which checks all properties, similar to the provider test.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/ProviderUserRepositoryTests.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        // Cleanup
+        await ssoConfigRepository.DeleteAsync(ssoConfig);
+        await providerOrganizationRepository.DeleteAsync(providerOrganizationWithSso);
+        await providerOrganizationRepository.DeleteAsync(providerOrganizationWithoutSso);
+        await providerUserRepository.DeleteAsync(providerUser);
+        await providerRepository.DeleteAsync(provider);
+        await organizationRepository.DeleteAsync(organizationWithSso);
+        await organizationRepository.DeleteAsync(organizationWithoutSso);
+        await userRepository.DeleteAsync(user);
```
eliykat: Same here as above (if you agree)

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/ProviderUserRepositoryTests.cs
```diff
@@ -0,0 +1,151 @@
...
+using Bit.Core.Entities;
+using Bit.Core.Repositories;
+using Xunit;
+
+namespace Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories;
+
+public class ProviderUserRepositoryTests
+{
+    [Theory, DatabaseData]
+    public async Task GetManyOrganizationDetailsByUserAsync_ShouldPopulateSsoPropertiesCorrectly(
```
eliykat: Naming: this is specific to SsoProperties, but it checks all properties. I think it's good to check all properties but the test name should reflect this (or have separate tests).

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/ProviderUserRepositoryTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        await providerOrganizationRepository.DeleteAsync(providerOrganizationWithSso);
+        await providerOrganizationRepository.DeleteAsync(providerOrganizationWithoutSso);
+        await providerUserRepository.DeleteAsync(providerUser);
+        await providerRepository.DeleteAsync(provider);
+        await organizationRepository.DeleteAsync(organizationWithSso);
+        await organizationRepository.DeleteAsync(organizationWithoutSso);
+        await userRepository.DeleteAsync(user);
+    }
+
+    private static void AssertProviderOrganizationDetails(
```
eliykat, line 92: Looking at this again, the test helpers create a very basic organization without many properties set. That means most properties would default to null or false, which is indistinguishable from a sproc that is missing data (e.g. missing a column).

I suggest using Autodata to create the organization object so that it's populated with identifiable data. You can do this with `new Fixture().Create<T>();`.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/ProviderUserRepositoryTests.cs
r-tome: It had that name because initially it just checked SSO properties but I added the other ones afterwards

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/ProviderUserRepositoryTests.cs
r-tome: I've added a private method that initializes all properties. To use `new Fixture()` I would have to add the `AutoFixture` package and it seemed to be out of scope here, I can add it later if you think its worth it.
Edit: I went back and edited `CreateTestOrganizationAsync` to set all the properties

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepository/OrganizationUserRepositoryTests.cs
r-tome: `GetManyDetailsByUserAsync_Works` already did that, I added the missing properties to it

src/Api/AdminConsole/Models/Response/ProfileOrganizationResponseModel.cs
r-tome: I double checked and you're right! The `FamilySponshorship` properties are in [OrganizationUserOrganizationDetailsView](https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Views/OrganizationUserOrganizationDetailsView.sql) but not in [ProviderUserProviderOrganizationDetailsView](https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Views/ProviderUserProviderOrganizationDetailsView.sql)

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/ProviderUserRepositoryTests.cs
eliykat: The benefit of autofixture is that when a new property is added in the future, autofixture will automatically detect it and populate it. Otherwise, it relies on the dev to update the helper method to populate it, which they won't do if they've forgotten about it (and that's what we're trying to protect against).

I'm OK with leaving it for now though.

PR #6388, opened 2025-09-26
Refactor Slack Callback Mechanism

brant-livefront:
## 📔 Objective

I discovered that there was a problem in how we were configuring Slack previously. I had the organizationId passed as part of the callback URL, but Slack needs the callback URL to be known ahead of time. Instead, this PR creates an empty Slack `OrganizationIntegration` for the organization and passes the id of the integration through the state. When Slack calls us back, they return the state so we can validate that the request matches what was originally initiated. If we find the empty `OrganizationIntegration` row, we update it to include the details from Slack. If we don't find a pre-existing row, then we ignore it.

This PR also introduces a OrganizationIntegrationStatus to help navigate scenarios like this. While Slack is awaiting the return OAuth callback, it is in an `Initiated` status. When we've received the callback and stored the token, its status changes to `Completed`. That way we know that even though there's a record, it is not yet complete and is waiting on further information. We plan on extending the use of this status when we get to the Microsoft Teams integration as well.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/AdminConsole/Controllers/SlackIntegrationController.cs
```diff
@@ -18,25 +15,53 @@
...
+            protocol: currentContext.HttpContext.Request.Scheme,
+            host: currentContext.HttpContext.Request.Host.ToUriComponent()
+        );
+        if (string.IsNullOrEmpty(callbackUrl))
+        {
+            throw new BadRequestException("Unable to build callback Url");
+        }
+
+        var integrations = await integrationRepository.GetManyByOrganizationAsync(organizationId);
+        var integration = integrations.FirstOrDefault(i => i.Type == IntegrationType.Slack);
```
withinfocus, line 46: 💭 I don't love the exposure from this and its assumptions. We haven't declared yet that you can only have one Slack integration and we should maybe check that if one _is_ returned that its configuration is null.

src/Api/AdminConsole/Controllers/SlackIntegrationController.cs
```diff
@@ -18,25 +15,53 @@
...
+        if (integration is null)
+        {
+            integration = await integrationRepository.CreateAsync(new OrganizationIntegration
+            {
+                OrganizationId = organizationId,
+                Type = IntegrationType.Slack,
+                Configuration = null,
+            });
+        }
+        if (integration.Configuration is not null)
```
withinfocus: ⛏️ Logic might change per my above comment but this could be an `else`.

src/Api/AdminConsole/Controllers/SlackIntegrationController.cs
```diff
@@ -46,38 +71,39 @@ public async Task<IActionResult> RedirectAsync(Guid organizationId)
...
 
-    [HttpGet("create", Name = nameof(CreateAsync))]
-    public async Task<IActionResult> CreateAsync(Guid organizationId, [FromQuery] string code)
+    [HttpGet("integrations/slack/create", Name = nameof(CreateAsync))]
+    [AllowAnonymous]
+    public async Task<IActionResult> CreateAsync([FromQuery] string code, [FromQuery] string state)
     {
-        if (!await currentContext.OrganizationOwner(organizationId))
+        // Fetch existing Initiated record
+        var integration = await integrationRepository.GetByIdAsync(Guid.Parse(state));
```
withinfocus: 🎨 Need to try parsing and use other safety mechanisms -- people will know this API exists and could send junk at it.

src/Api/AdminConsole/Controllers/SlackIntegrationController.cs
```diff
@@ -46,38 +71,39 @@ public async Task<IActionResult> RedirectAsync(Guid organizationId)
...
+        if (integration is null)
         {
-            throw new NotFoundException();
+            throw new BadRequestException("No record found for given state.");
```
withinfocus: ⚠️ Too revealing to me -- just keep returning not found.

src/Api/AdminConsole/Models/Response/Organizations/OrganizationIntegrationResponseModel.cs
```diff
@@ -21,4 +19,29 @@ public OrganizationIntegrationResponseModel(OrganizationIntegration organization
     public Guid Id { get; set; }
     public IntegrationType Type { get; set; }
     public string? Configuration { get; set; }
+
+    public OrganizationIntegrationStatus Status => Type switch
+    {
+        // Not yet implemented, shouldn't be present, leave Initiated for now
+        IntegrationType.CloudBillingSync => OrganizationIntegrationStatus.Initiated,
+        IntegrationType.Scim => OrganizationIntegrationStatus.Initiated,
```
withinfocus: 🎨 With these out of scope really I'd let them fall into a `default` case, and that should probably be assumed as `Completed`. Or, do we return a status of "Invalid" for both this and a default?

src/Core/AdminConsole/Enums/OrganizationIntegrationStatus.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Bit.Api.AdminConsole.Models.Response.Organizations;
+
+public enum OrganizationIntegrationStatus : int
+{
+    Initiated = 1,
```
withinfocus: ⛏️ Pretty sure we like to start with zero and you just won't set values here as that's the default, plus no one should use the numeric value.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/SlackService.cs
```diff
@@ -73,9 +74,18 @@ public async Task<string> GetDmChannelByEmailAsync(string token, string email)
         return await OpenDmChannel(token, userId);
     }
 
-    public string GetRedirectUrl(string redirectUrl)
+    public string GetRedirectUrl(string callbackUrl, string state)
     {
-        return $"https://slack.com/oauth/v2/authorize?client_id={_clientId}&scope={_scopes}&redirect_uri={redirectUrl}";
+        var builder = new UriBuilder(_slackOAuthBaseUri);
+        var query = HttpUtility.ParseQueryString(builder.Query);
```
withinfocus, line 80: ❓ Does this need any safety?

withinfocus (changes requested):

src/Api/AdminConsole/Models/Response/Organizations/OrganizationIntegrationResponseModel.cs
brant-livefront: I was trying to avoid a default, because if we explicitly handle every case, the compiler will yell at us as soon as we've added a new Type and we forget to update it here. 

But it might be worth simply adding and `Invalid` status and mapping everything to that (i.e. new cases would respond with Invalid until fixed, instead of compile time erroring). I do like the idea of having Invalid here since some of these cases (i.e. HEC, Datadog) I'm picking Initiated just to have a fallback when in reality they shouldn't be allowed to be setup like that. I'll take a stab at doing that.

withinfocus: Well, for our known types where we might want to show it in the UI in a positive way, something like "NotApplicable" could be softer.

src/Api/AdminConsole/Controllers/SlackIntegrationController.cs
brant-livefront: We do have a constraint on the DB such that you can only have one integration per type per organization. So the three scenarios here are:
1. There's nothing stored and this is the first time:  Create the empty row and redirect
2. There's a previously stored empty row: Return the id and kick off the redirect again (maybe something just went bad in the process and we don't want to enforce one-shot-or-nothing)
3. There's fully completed slack config: Stop and fail. We don't want to overwrite the existing slack config

I do think though (given your comment below) that I probably over-optimized a bit. I'll refactor to a more clear if / else so it's a bit more clear in code.

src/Api/AdminConsole/Controllers/SlackIntegrationController.cs
brant-livefront: I added a bit more safety to this state param. Might be overkill but I wanted to make it much less guessable than just the guid. If we do like this, it'd be easy to pickup and reuse on the Teams integration.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/SlackService.cs
brant-livefront: Should be safe - we're using a private const as the base URL and it doesn't have an existing query string.

src/Api/AdminConsole/Models/Response/Organizations/OrganizationIntegrationResponseModel.cs
```diff
@@ -21,4 +19,29 @@ public OrganizationIntegrationResponseModel(OrganizationIntegration organization
     public Guid Id { get; set; }
     public IntegrationType Type { get; set; }
     public string? Configuration { get; set; }
+
+    public OrganizationIntegrationStatus Status => Type switch
```
withinfocus, line 23: ℹ️ I anticipate us changing this in the future and once we get a UI in place. Something still isn't quite right to me around N/A.

src/Api/AdminConsole/Models/Response/Organizations/OrganizationIntegrationResponseModel.cs
```diff
@@ -21,4 +19,29 @@ public OrganizationIntegrationResponseModel(OrganizationIntegration organization
...
+
+        // HEC and Datadog should only be allowed to be created non-null.
+        // If they are null, they are NotApplicable
+        IntegrationType.Hec => string.IsNullOrWhiteSpace(Configuration)
+            ? OrganizationIntegrationStatus.NotApplicable
+            : OrganizationIntegrationStatus.Completed,
+        IntegrationType.Datadog => string.IsNullOrWhiteSpace(Configuration)
+            ? OrganizationIntegrationStatus.NotApplicable
+            : OrganizationIntegrationStatus.Completed,
+    };
```
withinfocus, line 46: One example here is that I don't see those states as N/A but invalid. Do we need both?

src/Core/AdminConsole/Models/Data/EventIntegrations/IntegrationOAuthState.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System.Security.Cryptography;
+using System.Text;
+using Bit.Core.AdminConsole.Entities;
+
+namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+
+public class IntegrationOAuthState
+{
+    private const int _orgHashLength = 12;
+    private static readonly TimeSpan _maxAge = TimeSpan.FromMinutes(20);
```
withinfocus, line 10: ❓ Is this Slack-specific?

brant-livefront: No, I just wanted to pick an appropriate timeout for our own purposes. I had in mind that we'd re-use this for Microsoft Teams so I didn't want it tied to their timing. Even if we think the state is valid, there's still the possibility that Slack rejects the code when we use it.

withinfocus (approved):

withinfocus:
@jrmccannon all yours.

PR #6401, opened 2025-09-30
PM-2035: PRF Unlock

abergs:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-2035

## 📔 Objective

This PR is complimentary to the client changes proposed in https://github.com/bitwarden/clients/pull/16662.

It adds all PRF enabled passkeys to the DecryptionOptions response in our SyncResponse.

We previously thought we needed to return them in the WebAuthnGrant (IdTokenResponse) too, but that idea has been dropped and that code flow is unchanged (Except for the addition of the transports paramter)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Identity/IdentityServer/RequestValidators/WebAuthnGrantValidator.cs
```diff
@@ -98,7 +101,8 @@ public async Task ValidateAsync(ExtensionGrantValidationContext context)
         }
 
         var (user, credential) = await _assertWebAuthnLoginCredentialCommand.AssertWebAuthnLoginCredential(token.Options, deviceResponse);
-        UserDecryptionOptionsBuilder.WithWebAuthnLoginCredential(credential);
+        var allCredentials = await _webAuthnCredentialRepository.GetManyByUserIdAsync(user.Id);
+        UserDecryptionOptionsBuilder.WithWebAuthnLoginCredentials(allCredentials);
```
abergs: @trmartin4 Was there a decision by Auth to not send all of these as part of the sign in and just rely on sync?

If so - I will remove the array changes for the Auth response models. In theory, we could also remove the transports from Auth's flow. 

(cc: @jlf0dev)

abergs: The only reason to keep sending all down is if we expect the user to sign in with something else and the decrypt using a passkey?

Or do we perform a sync in-between the sign in and the next unlock step?

abergs: I'll note that we could meet "half way" and just return a single [credential] but still change the type to be an array.

We're currently sharing the `UserDecryptionOptions` between auth/KM on the client, which would require both to be lists. OR, we duplicate the `UserDecryptionOptions` on the client where KM's implementation has an array of credentials, and auth only one. But it does seems a bit overcomplicated to do that.

trmartin4: Yes, we agreed that we did not want to send it down after authentication and just rely on sync.  We perform sync immediately after logging in, and don't direct the user to the vault until sync has completed, and so any subsequent unlocks will have the full list of potential credentials to use.  I would support not duplicating objects as well, and just keep an array type with a single item.

abergs: For the record: We decided to handle the array mapping on the client side, causing us to not having to do a breaking change server side.

quexten (approved):
Approving for KM, will re-review the clients changes soon.

ike-kottlowski (approved):
Updates to the UserDecryptionOptions look good, thank you.

abergs:
Fixed another conflict...

quexten (dismissed):

src/Api/Vault/Models/Response/SyncResponseModel.cs
```diff
@@ -57,6 +61,16 @@ public SyncResponseModel(
...
         Sends = sends.Select(s => new SendResponseModel(s, globalSettings));
+        var webAuthnPrfOptions = webAuthnCredentials
+            .Where(c => c.GetPrfStatus() == WebAuthnPrfStatus.Enabled)
+            .Select(c => new WebAuthnPrfDecryptionOption(
+                c.EncryptedPrivateKey,
+                c.EncryptedUserKey,
+                c.CredentialId,
+                [] // transports as empty array
+            ))
+            .ToArray();
```
abergs, line 72: consider adding a feature flag to always return a empty list?

PR #6403, opened 2025-09-30
Add default values for MaxConcurrentCalls and PrefetchCount across event and integration ASB listeners

brant-livefront:
## 📔 Objective

When looking at the performance on prod of the new event integrations, it became clear that we had some sort of bottleneck in getting messages out of ASB fast enough. Most of the message processing code and the code that does the table storage piece are identical across the new version and the Azure queue hosted service. In digging into ASB performance, we found two possible settings that could be causing problems:

- **MaxConcurrentCalls**
  - Determines the number of asynchronously operating concurrent processors of messages
  - Default is 1, so everything was processed serially (i.e. one listener could handle 1 message at a time)
- **PrefetchCount**
  - Determines how many rows an instance should fetch ahead of processing
  - Instances fetch their prefetch count, cache locally, work through their cache until they've exhausted it, and then prefetch another batch
  - This lowers the amount of network latency per instance, so they can process their queue faster without needing to fetch records as often
  - Default is 0, meaning _no_ prefetch at all - each message required a fetch from ASB

This PR adds some code to allow for us to configure these at the ListenerConfiguration level. That mans we could later introduce individual global settings properties and tune to the level of each event or integration listener (if we need that granularity). As a starting point, however, it simply introduces global settings properties for default values. The `ListenerConfiguration` base class then applies those defaults to everything. This gives us a quick way to get started, but a path forward for more precise tuning later.

Also note: I initialized global settings with the same defaults as ASB has (i.e. 1 for MaxConcurrentCalls, 0 for PrefetchCount). My thinking was this gives us a way to safely deploy this code without having to worry about changing anything by accident. The expectation is that we'd override these with precise values as we deploy.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Models/Data/EventIntegrations/IEventListenerConfiguration.cs
+++ b/src/Core/AdminConsole/Models/Data/EventIntegrations/IEventListenerConfiguration.cs
@@ -5,4 +5,6 @@ public interface IEventListenerConfiguration
     public string EventQueueName { get; }
     public string EventSubscriptionName { get; }
     public string EventTopicName { get; }
+    public int EventPrefetchCount { get; }
+    public int EventMaxConcurrentCalls { get; }
 }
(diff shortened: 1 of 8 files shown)


withinfocus:
@r-tome all yours.

r-tome (approved):

PR #6410, opened 2025-10-03
Add Microsoft Teams integration

brant-livefront:
## 📔 Objective

This PR adds an integration with Microsoft Teams.

This is a multi-step process that we will need to take into account when building the UI for this.

1. The organization owner kicks off an OAuth flow via the redirect
2. Microsoft authenticates the user and our request
3. Microsoft sends the user back to our create url, passing along a code and our state
4. We verify the state and use the code provided via OAuth to fetch (via Microsoft's Graph API) the TenantId and list of Teams for the owner who kicked off the OAtuh flow. We store these into an `Initiated` record in `OrganizationIntegration`
5. The owner installs our official app into their Teams instance and picks the channel in which to receive messages from the bot
6. Microsoft sends us an incoming bot event about the app installation
7. We parse the information from that incoming installation event, pulling out the tenant id, team id, conversation id, and serviceUrl
8. We lookup in our OrganizationIntegration table to see if there was a previously stored `Initiated` record with a matching tenant id and team id. If it is present, we update that record to include the serviceUrl and conversation id of the channel, thus moving the Teams integration to `Completed`. If there is not a matching record, we disregard the request
9. At this point the Teams integration is ready to send event information. The only renaming piece is to setup `OrganizationIntegrationConfiguration` records for the desired event types and template strings. These records will all contain `null` for the `Configuration` at this level because all of the necessary configuration is provided in the `OrganizationIntegration` configuration object.

### Setup tasks
- I'm currently using a minimalist app containing a minimal bot just as a POC. We will need an official app to be made in order to launch this. 
  - I'll write up more notes on the specific steps but overall we need to register an app, register a bot, put together a manifest/package, and submit it to Microsoft's store
  - It is possible to side load it directly in a Team, which is what I'm doing to test currently
- This PR adds a new integration, which will need new Subscriptions to be created in ASB
  - If it launches _before_ we turn the feature flag on, then we can create these subscriptions ahead of time
  - If it launches _after_, we'll need to follow [the process outlined in the docs](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Services/Implementations/EventIntegrations/README.md#azure-service-bus-1) (i.e. create the subscription with a false filter before deploying the code, then remove the filter after the code is live).
  - The subscriptions / defaults are:
    - `globalSettings.EventLogging.AzureServiceBus.TeamsEventSubscriptionName` with a default name of `events-teams-subscription`
    - `globalSettings.EventLogging.AzureServiceBus.TeamsIntegrationSubscriptionName` with a default name of `integration-teams-subscription`

### Notes
- This PR builds on #6388 and specifically the `IntegrationOAuthState` to do the OAuth validation piece. This helps validate the information coming back on the create method was actually generated by a legitimate actor and for the correct organization. As the API is public and not behind auth, we need to make sure that the requests received are valid
- Additionally, the bot callback that we receive on app install is not behind our auth. However, Microsoft does send their own JWT which is validated by the `TeamsBotCredentialProvider`. It will validate that the requests are issued by Microsoft and contain our client id as the audience. Additionally, this API can only add information to an existing, matching `Initiated` record - you must have a matching tenant id and teams id. and t can't modify a `Completed` record
- Microsoft does [not allow bots to post to private channels](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/conversations/channel-and-group-conversations?tabs=dotnet). This is a known limitation for which I couldn't find a workaround.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


dev/servicebusemulator_config.json
```diff
@@ -3,22 +3,6 @@
...
-              "DuplicateDetectionHistoryTimeWindow": "PT20S",
-              "ForwardDeadLetteredMessagesTo": "",
-              "ForwardTo": "",
-              "LockDuration": "PT1M",
-              "MaxDeliveryCount": 3,
-              "RequiresDuplicateDetection": false,
-              "RequiresSession": false
-            }
-          }
-        ],
```
brant-livefront, line 21: This is just a cleanup. We don't use queues for our local test so I thought it'd be cleaner to remove this.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/Queries/OrganizationIntegrationReadByTeamsConfigurationTenantIdTeamIdQuery.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public IQueryable<OrganizationIntegration> Run(DatabaseContext dbContext)
+    {
+        var query =
+            from oi in dbContext.OrganizationIntegrations
+            where oi.Type == IntegrationType.Teams &&
+                  oi.Configuration != null &&
+                  oi.Configuration.Contains($"\"TenantId\":\"{_tenantId}\"") &&
+                  oi.Configuration.Contains($"\"id\":\"{_teamId}\"")
```
brant-livefront, line 26: @withinfocus I couldn't find a good way to do a JSON query in EF and keep it generic, so I opted to use this Contains approach instead. I think it's functionally equivalent.

eliykat:
I've requested a review from the team as I'm trying to wrap up work before going on leave.

src/Api/AdminConsole/Controllers/TeamsIntegrationController.cs
```diff
@@ -0,0 +1,147 @@
...
+        );
+        if (string.IsNullOrEmpty(callbackUrl))
+        {
+            throw new BadRequestException("Unable to build callback Url");
+        }
+
+        var token = await teamsService.ObtainTokenViaOAuth(code, callbackUrl);
+        if (string.IsNullOrEmpty(token))
+        {
+            throw new BadRequestException("Invalid response from Slack.");
```
withinfocus:
```suggestion
            throw new BadRequestException("Invalid response from Teams.");
```

src/Api/AdminConsole/Controllers/TeamsIntegrationController.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        var teams = await teamsService.GetJoinedTeamsAsync(token);
+
+        if (teams.Count == 0)
```
withinfocus: 🎨 `Any`?

src/Core/AdminConsole/Models/Teams/TeamsApiResponse.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Text.Json.Serialization;
```
withinfocus, line 1: ⛏️ I'd use `Api` instead of `API` in this filename.

src/Core/AdminConsole/Models/Teams/TeamsBotCredentialProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public Task<bool> IsAuthenticationDisabledAsync()
+    {
+        return Task.FromResult(false);
+    }
+
+    public Task<bool> ValidateIssuerAsync(string issuer)
+    {
+        return Task.FromResult(issuer == "https://api.botframework.com");
```
withinfocus: ❓ They don't provide a constant for this?

src/Core/AdminConsole/Repositories/IOrganizationIntegrationRepository.cs
```diff
@@ -5,4 +5,6 @@ namespace Bit.Core.Repositories;
 public interface IOrganizationIntegrationRepository : IRepository<OrganizationIntegration, Guid>
 {
     Task<List<OrganizationIntegration>> GetManyByOrganizationAsync(Guid organizationId);
+
+    Task<OrganizationIntegration?> GetByTenantIdTeamId(string tenantId, string teamId);
```
withinfocus: 💭 This naming is configuration-specific and kinda bothers me since it's so pointed. What about something that has "configuration data" in it, and the type? Perhaps `GetByTeamsConfigurationTenantIdTeamId`?

util/Migrator/DbScripts/2025-10-3_00_AddOrganizationIntegration_ReadByTenantIdTeamIdStoredProcedure.sql
```diff
@@ -0,0 +1,22 @@
...
+    @TeamId NVARCHAR(200)
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+SELECT TOP 1
+    [Id],
+    [OrganizationId],
+    [Type],
+    [Configuration]
```
withinfocus: ❌ You should select * from the view.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/TeamsService.cs
```diff
@@ -0,0 +1,183 @@
...
+        string conversationId,
+        Uri serviceUrl,
+        string teamId,
+        string tenantId)
+    {
+        var integration = await _integrationRepository.GetByTenantIdTeamId(
+            tenantId: tenantId,
+            teamId: teamId);
+
+        if (integration is null)
```
withinfocus: ❓ These sorts of things shouldn't actively throw exceptions?

src/Core/Core.csproj
```diff
@@ -38,6 +38,8 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.10" />
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.52.0" />
     <PackageReference Include="Microsoft.Azure.NotificationHubs" Version="4.2.0" />
+    <PackageReference Include="Microsoft.Bot.Builder" Version="4.23.0" />
+    <PackageReference Include="Microsoft.Bot.Connector" Version="4.23.0" />
```
withinfocus, line 42: ℹ️ These save a lot of time and code, but I don't love adding dependencies so strongly for an integration.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/Queries/OrganizationIntegrationReadByTeamsConfigurationTenantIdTeamIdQuery.cs
withinfocus: This works, but you could also parse it as JSON and make the checks a bit more cleanly. It doesn't have to all be in the where clause.

src/Sql/dbo/Stored Procedures/OrganizationIntegration_ReadByTenantIdTeamId.sql
```diff
@@ -0,0 +1,21 @@
...
+    @TeamId NVARCHAR(200)
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+SELECT TOP 1
+    [Id],
+    [OrganizationId],
+    [Type],
+    [Configuration]
```
withinfocus: ❌ You should select * from the view.

withinfocus:
Oh, and do you also want to update the docs?

src/Core/AdminConsole/Models/Teams/TeamsBotCredentialProvider.cs
brant-livefront: Looks like they do 👍  adjusted in next commit

src/Core/AdminConsole/Repositories/IOrganizationIntegrationRepository.cs
brant-livefront: That makes sense. Adjusting to this name in the next commit.

src/Core/Core.csproj
brant-livefront: I agree 100% in general. I did all of the graph api calls natively since they were just normal REST. But the incoming bot stuff seemed to push heavily in the dependency direction.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/TeamsService.cs
brant-livefront: This is the incoming app install callback. Any time someone installs the official app we'll get this callback, whether or not they've setup their organization to use the app. So we could get a decent amount of noise on this one where we don't have a matching integration, but we get this callback. Also, I don't think bubbling up an exception or returning an error state for the call changes what Microsoft does for handling (and could encourage them to re-send if we sent back say a 500 error).

Given all that, I thought it best to simply no-op and move on.

brant-livefront:
@withinfocus Looking at what needs to change in the docs now, but wanted to push up the code changes I have. I think I've addressed all your comments.  👍

withinfocus (approved):
Documentation can be added here in another commit and I'll re-review but this all works. I still have one open comment about the EF query getting cleaned up, but what you have works.

withinfocus (approved):
Left two comments on the [commit](https://github.com/bitwarden/server/commit/b5376d528a8fa41a8a67d54a2fa20644e6e5063d#comments) (that often gets me when I click on email notifications) with one being a ⛏️ you can quickly fix, but this is good to go at this point.

src/Api/AdminConsole/Controllers/TeamsIntegrationController.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        var teams = await teamsService.GetJoinedTeamsAsync(token);
+
+        if (!teams.Any())
+        {
+            throw new BadRequestException("No teams were found.");
+        }
+
+        var teamsIntegration = new TeamsIntegration(TenantId: teams.First().TenantId, Teams: teams);
```
jrmccannon:
```suggestion
        var teamsIntegration = new TeamsIntegration(TenantId: teams[0].TenantId, Teams: teams);
```
To get rid of warning.

src/Api/AdminConsole/Controllers/TeamsIntegrationController.cs
```diff
@@ -0,0 +1,147 @@
...
+            throw new NotFoundException();
+        }
+
+        // Verify Organization matches hash
+        if (!oAuthState.ValidateOrg(integration.OrganizationId))
+        {
+            throw new NotFoundException();
+        }
+
+        string? callbackUrl = Url.RouteUrl(
```
jrmccannon:
```suggestion
        var callbackUrl = Url.RouteUrl(
```
To remove warning.

src/Api/AdminConsole/Controllers/TeamsIntegrationController.cs
```diff
@@ -0,0 +1,147 @@
...
+{
+    [HttpGet("{organizationId:guid}/integrations/teams/redirect")]
+    public async Task<IActionResult> RedirectAsync(Guid organizationId)
+    {
+        if (!await currentContext.OrganizationOwner(organizationId))
+        {
+            throw new NotFoundException();
+        }
+
+        string? callbackUrl = Url.RouteUrl(
```
jrmccannon:
```suggestion
        var callbackUrl = Url.RouteUrl(
```
To remove warning.

src/Core/AdminConsole/Models/Data/EventIntegrations/TeamsIntegration.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Bit.Core.AdminConsole.Models.Data.EventIntegrations;
+
+public record TeamsIntegration(
+    string TenantId,
+    IReadOnlyList<TeamInfo> Teams,
+    string? ChannelId = null,
+    Uri? ServiceUrl = null)
+{
+    public bool isVerified => (!string.IsNullOrEmpty(ChannelId) && ServiceUrl is not null);
```
jrmccannon:
```suggestion
    public bool IsVerified => !string.IsNullOrEmpty(ChannelId) && ServiceUrl is not null;
```
To remove warning

src/Core/AdminConsole/Models/Teams/TeamsApiResponse.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Bit.Core.Models.Teams;
+
+public class TeamsOAuthResponse
```
jrmccannon, line 8: Could we provide XML documentation for these classes along with links to the Teams API documentation around these responses?

src/Core/AdminConsole/Services/ITeamsService.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Bit.Core.Models.Teams;
+
+namespace Bit.Core.Services;
+
+public interface ITeamsService
```
jrmccannon, line 9: Can we put some XML comments explaining each of these methods?

src/Core/AdminConsole/Services/Implementations/EventIntegrations/TeamsService.cs
```diff
@@ -0,0 +1,183 @@
...
+
+namespace Bit.Core.Services;
+
+public class TeamsService(
+    IHttpClientFactory httpClientFactory,
+    IOrganizationIntegrationRepository integrationRepository,
+    GlobalSettings globalSettings,
+    ILogger<TeamsService> logger) : ActivityHandler, ITeamsService
+{
+    private readonly IOrganizationIntegrationRepository _integrationRepository = integrationRepository;
```
jrmccannon: Couldn't we just use the integration repository from the primary constructor?

test/Core.Test/AdminConsole/Services/IntegrationTypeTests.cs
```diff
@@ -5,26 +5,45 @@ namespace Bit.Core.Test.Services;
...
+    public void ToRoutingKey_CloudBillingSync_ThrowsException()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => IntegrationType.CloudBillingSync.ToRoutingKey());
+    }
+
+    [Fact]
+    public void ToRoutingKey_Scim_ThrowsException()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => IntegrationType.Scim.ToRoutingKey());
+    }
```
jrmccannon, line 18: What's prompting these tests? I'm not seeing any code related to those.

brant-livefront: The routing key is meant to be used for sending/filtering integration messages (i.e. in an ASB subscription). CloudBillingSync and Scim were copied over from `OrganizationConnectionType` in case we wanted to unify those into this architecture at some point. As such, there's no implementations for either and so it's really an error to be calling ToRoutingKey on either of those. So these tests are just to pin that down.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/TeamsService.cs
brant-livefront: 💯 fixed in next commit

src/Core/AdminConsole/Models/Teams/TeamsApiResponse.cs
brant-livefront: Added in latest. 👍

src/Core/AdminConsole/Services/ITeamsService.cs
brant-livefront: Added in latest. 👍

PR #6418, opened 2025-10-06
Billing/pm 25379/clean up org metadata usage

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25379

## 📔 Objective

This work was a refactor, so no functional or visual changes should be present. Note: these changes are only applied with the feature flag `pm-25379-use-new-organization-metadata-structure`

Here’s a list of changes made for this issue.

1. All properties that were not being used anywhere in clients were removed from the models.
1. The cloud-hosted organization subscription page:
    1.  This page used to look at the organization metadata (provided by the server API) to determine if it should show the self hosting section. The property checked was `isEligibleForSelfHost`. This was updated to instead simply check that the org’s plan tier is either families or enterprise and the matching property was removed from the metadata models. (This is equivalent to the logic used by the server API before, but does not require any additional data to be fetched.)
    1. This page also used `isManaged` in org metadata to switch between two variations of UI for the “manage subscription” section. This was updated to instead use `hasBillableProvider` which the client already has access to on the org object and the property was removed from the metadata models. The UI switches:
        1. When the org is managed by a consolidated billing MSP.
        1. When the org is not managed by a consolidated billing MSP.
1. The organization members page uses organization metadata fetched from the API in the following cases:
    1. `isOnSecretsManagerStandalone` is used to force the secrets manager access checkbox when inviting a new member or editing an existing member.
    1. `occupiedSeatCount` is used in the following cases:
        1. to restrict inviting additional members if the seat limit is reached
        1. to show the remaining seats on the invite dialog

This PR is of course the server-side changes. It includes the new feature flag, removing of unused properties, refactoring the `GetMetadata` function into a new query class, and a new vnext endpoint to use that query.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/Billing/Controllers/VNext/OrganizationBillingVNextController.cs
```diff
@@ -113,6 +116,24 @@ public async Task<IResult> RestartSubscriptionAsync(
...
+        [BindNever] Organization organization)
+    {
+        var metadata = await getOrganizationMetadataQuery.Run(organization);
+
+        if (metadata == null)
+        {
+            return TypedResults.NotFound();
+        }
+
+        var response = OrganizationMetadataResponse.From(metadata);
```
amorask-bitwarden: ⛏️ I'm fine with removing this layer of response models when they do nothing but copy the Core models if you want to. Your call.

amorask-bitwarden (approved):

src/Api/Billing/Controllers/VNext/OrganizationBillingVNextController.cs
kdenney: Great suggestion; thanks! Especially now that it's only 2 properties, this class seems very superfluous now 😆 Removed in https://github.com/bitwarden/server/pull/6418/commits/ddbf08472e09930ca5627d2fcb9e72943498f3d7

amorask-bitwarden (dismissed):

PR #6421, opened 2025-10-07
[PM-23134] Update PolicyDetails sprocs for performance

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23134

## 📔 Objective

Modified `OrganizationUser_ReadByUserIdWithPolicyDetails` to use a CTE and the new stored procedure `UserProviderAccessView`, as outlined in the dbops update.

Updated `PolicyRequirementQuery` to use `PolicyDetails_ReadByUserIdsPolicyType`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByUserIdWithPolicyDetails.sql
```diff
@@ -4,31 +4,53 @@
...
+            OU.[Id],
+            OU.[OrganizationId],
+            OU.[Type],
+            OU.[Status],
+            OU.[Permissions]
+        FROM
+            [dbo].[OrganizationUserView] OU
+        WHERE
+            (OU.[Status] <> 0 AND OU.[UserId] = @UserId)
+            OR (OU.[Status] = 0 AND OU.[Email] = @UserEmail AND @UserEmail IS NOT NULL)
```
JimmyVo16: We should use UNION instead of OR. Here’s an [example](https://github.com/bitwarden/server/blob/03327cb082a35ede6f61191afa140ddf2297aa97/src/Sql/dbo/Stored%20Procedures/PolicyDetails_ReadByOrganizationId.sql#L15): maybe we can turn this UNION into a view and use it for both stored procedures.

r-tome: @bitwarden/dept-dbops what do you think?

mkincaid-bw: I did a quick test and the UNION does perform slightly better than the OR clause, which is what I expected.

eliykat:
@JimmyVo16 has already been looking at this so I'll let him handle the review.

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByUserIdWithPolicyDetails.sql
```diff
@@ -4,31 +4,70 @@
...
-            FROM [dbo].[ProviderUserView] PU
-            INNER JOIN [dbo].[ProviderOrganizationView] PO ON PO.[ProviderId] = PU.[ProviderId]
-            WHERE PU.[UserId] = OU.[UserId] AND PO.[OrganizationId] = P.[OrganizationId]
-        ) THEN 1 ELSE 0 END AS IsProvider
-FROM [dbo].[PolicyView] P
-INNER JOIN [dbo].[OrganizationUserView] OU
-    ON P.[OrganizationId] = OU.[OrganizationId]
-WHERE P.[Type] = @PolicyType AND
+
+    DECLARE @UserEmail NVARCHAR(320)
```
mkincaid-bw: ⛏️ Purely cosmetic here but the email column in the DB is `NVARCHAR(256)`, so `NVARCHAR(320)` is larger than needed. However, SQL Server only allocates memory to hold the actual value, so no need to change unless you want to be consistent with the table column size.

util/Migrator/DbScripts/2025-10-02_00_RefactorPolicyDetailsQueries.sql
```diff
@@ -0,0 +1,85 @@
...
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationUser_ReadByUserIdWithPolicyDetails]
+    @UserId UNIQUEIDENTIFIER,
+    @PolicyType TINYINT
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @UserEmail NVARCHAR(320)
```
mkincaid-bw: ⛏️ Purely cosmetic here but the email column in the DB is `NVARCHAR(256)`, so `NVARCHAR(320)` is larger than needed. However, SQL Server only allocates memory to hold the actual value, so no need to change unless you want to be consistent with the table column size.

PR #6426, opened 2025-10-07
Ac/pm 25823/vnext policy upsert pattern

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25823

## 📔 Objective

Add a new version of SavePolicyCommand to enable 
1. Individual handlers to access metadata properties
2. Support pre and post side effects.
3. Refactor to clean up policy dependencies and validation logic.
4. Enable each policy to implement only the events relevant to it.
6. Add unit tests

This code isn’t consumed yet and therefore isn’t testable.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IVNextSavePolicyCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IVNextSavePolicyCommand.cs
@@ -1,6 +1,6 @@
-﻿#nullable enable
-using Bit.Core.AdminConsole.Entities;
+﻿using Bit.Core.AdminConsole.Entities;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
+using Microsoft.Azure.NotificationHubs.Messaging;
 
 namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyUpdateEvents.Interfaces;
 
@@ -24,7 +24,7 @@ public interface IVNextSavePolicyCommand
     /// </summary>
     /// <param name="policyRequest">Policy data, acting user, and metadata.</param>
     /// <returns>The saved policy with updated revision and applied changes.</returns>
-    /// <exception cref="Exception">
+    /// <exception cref="BadRequestException">
     /// Thrown if:
     /// - The organization can’t use policies
     /// - Dependent policies are missing or block changes


src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,209 @@
...
+    IEnumerable<IEnforceDependentPoliciesEvent> policyValidationEventHandlers,
+    TimeProvider timeProvider,
+    IPolicyEventHandlerFactory policyEventHandlerFactory)
+    : IVNextSavePolicyCommand
+{
+    private readonly IReadOnlyDictionary<PolicyType, IEnforceDependentPoliciesEvent> _policyValidationEvents = MapToDictionary(policyValidationEventHandlers);
+
+    private static Dictionary<PolicyType, IEnforceDependentPoliciesEvent> MapToDictionary(IEnumerable<IEnforceDependentPoliciesEvent> policyValidationEventHandlers)
+    {
+        var policyValidationEventsDict = new Dictionary<PolicyType, IEnforceDependentPoliciesEvent>();
```
JimmyVo16, line 25: I’m not sure if this is needed since it’s from the old pattern. I think it was meant to prevent duplicates, but that was when everything was in one interface. Now we have four interfaces, so I’m not sure it makes sense anymore. I’m happy to remove it.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,209 @@
...
+        await policyRepository.UpsertAsync(policy);
+
+        return policy;
+    }
+
+    private async Task ValidateTargetedPolicyAsync(SavePolicyModel policyRequest,
+        Policy? currentPolicy)
+    {
+        await ExecutePolicyEventAsync<IPolicyValidationEvent>(
+            currentPolicy!.Type,
```
r-tome: Why do we always expect the `currentPolicy` to not be null? What happens when we're setting up a policy the first time?

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,209 @@
...
+            case { Count: 1 }:
+                throw new BadRequestException($"Turn off the {dependentPolicyTypes.First().GetName()} policy because it requires the {validator.Type.GetName()} policy.");
+            case { Count: > 1 }:
+                throw new BadRequestException($"Turn off all of the policies that require the {validator.Type.GetName()} policy.");
+        }
+    }
+
+    private static void ValidateEnablingRequirements(
+        IEnforceDependentPoliciesEvent validator,
+        Policy? currentPolicy,
```
r-tome: `currentPolicy` is not used

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
JimmyVo16: Yeah, that should have used the type from the request object.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IOnPolicyPreUpdateEvent.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Bit.Core.AdminConsole.Entities;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyUpdateEvents.Interfaces;
+
+public interface IOnPolicyPreUpdateEvent : IPolicyUpdateEvent
+{
+    /// <summary>
+    /// Performs side effects before a policy is upserted.
```
JimmyVo16, line 9: Hey, I’m addressing your comments here so we can have a thread on it.


Original comment
> add xmldoc to every new interface clarifying what it represents and when its called

I added documentation to the method in the interface, would that be sufficient? I’m happy to adjust it if needed, but I’d prefer to avoid duplicating the same text in two places.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            if (!policyValidationEventsDict.TryAdd(policyValidationEvent.Type, policyValidationEvent))
+            {
+                throw new Exception($"Duplicate PolicyValidationEvent for {policyValidationEvent.Type} policy.");
+            }
+        }
+        return policyValidationEventsDict;
+    }
+
+    public async Task<Policy> SaveAsync(SavePolicyModel policyRequest)
```
JimmyVo16, line 36: Hey, I’m addressing your comments here so we can have a thread on it.

Original comment

> [add xmldoc] VNextSavePolicyCommand explain the execution flow

What do you think would make the most sense here? The method is pretty declarative each operation is abstracted and has a meaningful name. I try to use method names to convey intent whenever possible, and the overall flow is fairly straightforward. There isn’t much cyclomatic complexity or any deeply nested paths.

I’m happy to add more documentation, but I’m not sure where it would be most useful and it doesn’t seem beneficial to just repeat the same steps in the XML doc.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        ValidatePolicyDependencies(policyUpdateRequest, currentPolicy, savedPoliciesDict);
+
+        await ValidateTargetedPolicyAsync(policyRequest, currentPolicy);
+
+        await ExecutePreUpsertSideEffectAsync(policyRequest, currentPolicy);
+
+        var upsertedPolicy = await UpsertPolicyAsync(policyUpdateRequest);
+
+        await eventService.LogPolicyEventAsync(upsertedPolicy, EventType.Policy_Updated);
```
r-tome, line 55: This can be moved inside `UpsertPolicyAsync`

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
r-tome: > The method is pretty declarative each operation is abstracted and has a meaningful name

I agree that the method itself is clear and the flow is easy to follow, but you still have to look into the implementation to really understand how the workflow fits together. 

Each interface should have a short summary describing what it represents, when it’s used, and then each method should specify what it does and what it returns (if anything). 

My concern is that other devs shouldn’t have to dig through the code to piece together how everything is orchestrated — the XML docs should give enough context to guide them.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IVNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,10 @@
+﻿#nullable enable
+using Bit.Core.AdminConsole.Entities;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyUpdateEvents.Interfaces;
+
```
r-tome, line 6:
```suggestion

/// <summary>
/// Handles creating or updating organization policies with validation and side effect execution.
/// </summary>
/// <remarks>
/// Workflow:
/// 1. Validates organization can use policies  
/// 2. Validates required and dependent policies  
/// 3. Runs policy-specific validation (<see cref="IPolicyValidationEvent"/>)  
/// 4. Executes pre-save logic (<see cref="IOnPolicyPreUpdateEvent"/>)  
/// 5. Saves the policy  
/// 6. Logs the event  
/// 7. Executes post-save logic (<see cref="IOnPolicyPostUpdateEvent"/>)  
/// </remarks>
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IVNextSavePolicyCommand.cs
```diff
@@ -0,0 +1,10 @@
...
+public interface IVNextSavePolicyCommand
+{
+    Task<Policy> SaveAsync(SavePolicyModel policyRequest);
```
r-tome, line 33:
```suggestion
/// <summary>
    /// Performs necessary validations, saves the policy and any side effects
    /// </summary>
    /// <param name="policyRequest">Policy data, acting user, and metadata.</param>
    /// <returns>The saved policy with updated revision and applied changes.</returns>
    /// <exception cref="BadRequestException">
    /// Thrown if:
    /// - The organization can’t use policies  
    /// - Dependent policies are missing or block changes  
    /// - Custom validation fails  
    /// </exception>
    Task<Policy> SaveAsync(SavePolicyModel policyRequest);
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IPolicyEventHandlerFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿#nullable enable
+
+using Bit.Core.AdminConsole.Enums;
+using OneOf;
+using OneOf.Types;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyUpdateEvents.Interfaces;
+
```
r-tome, line 8:
```suggestion

/// <summary>
/// Provides policy-specific event handlers used during the save workflow in <see cref="IVNextSavePolicyCommand"/>.
/// </summary>
/// <remarks>
/// Supported handlers:  
/// - <see cref="IEnforceDependentPoliciesEvent"/> for dependency checks  
/// - <see cref="IPolicyValidationEvent"/> for custom validation  
/// - <see cref="IOnPolicyPreUpdateEvent"/> for pre-save logic  
/// - <see cref="IOnPolicyPostUpdateEvent"/> for post-save logic  
/// </remarks>
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/Interfaces/IPolicyEventHandlerFactory.cs
```diff
@@ -0,0 +1,12 @@
...
+public interface IPolicyEventHandlerFactory
+{
+    OneOf<T, None> GetHandler<T>(PolicyType policyType) where T : IPolicyUpdateEvent;
```
r-tome, line 29:
```suggestion
/// <summary>
    /// Gets the event handler for the given policy type and handler interface.
    /// </summary>
    /// <typeparam name="T">Handler type implementing <see cref="IPolicyUpdateEvent"/>.</typeparam>
    /// <param name="policyType">The policy type to resolve.</param>
    /// <returns>
    /// <see cref="OneOf{T, None}"/> — the handler if available, or None if not implemented.
    /// </returns>
    OneOf<T, None> GetHandler<T>(PolicyType policyType) where T : IPolicyUpdateEvent;
```

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
JimmyVo16: Yeah, the current implementation fires the event inside `UpsertPolicyAsync`. I moved it up since firing the event is a separate operation from saving the policy. They’re independent of each other. I moved it up to `SaveAsync` because this is the method that wants to fire off an event. This way, it's visible and it's not nested in some random method. 

This is how I approach methods, but I’m open to change. Generally, I think a method should live in the scope that needs it. For example, `policyRepository.GetByOrganizationIdTypeAsync` is only needed by `UpsertPolicyAsync`, so it makes sense to place it there. But `ExecutePostUpsertSideEffectAsync` is independent of `UpsertPolicyAsync` and is something that `SaveAsync` wants to happen, so it lives in `SaveAsync`.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
JimmyVo16: >My concern is that other devs shouldn’t have to dig through the code to piece together how everything is orchestrated — the XML docs should give enough context to guide them.

Yeah, that’s a fair statement.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
JimmyVo16: Chatted over a call. This implementation is fine since logging an event is more of a side effect.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/VNextSavePolicyCommand.cs
JimmyVo16: Chatted over a call. Interfaces should have documentation so that the client using them doesn’t need to know the implementation, which is the main goal of an interface. Implementations, however, don’t need documentation if the code provides enough context.

PR #6435, opened 2025-10-09
PM-26727 - First iteration of Claude Code Review tuning

theMickster:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26727

## 📔 Objective

Incorporate learnings from deeper Claude SDK research and feedback from teammates for the first iteration of changes. 
✅ The consensus from folks is that multiple Claude comments are too verbose and clog up person-to-person chat. Therefore, we're going to attempt to setup the _sticky_ comment functionality so Claude only updates one comment throughout the PR lifecycle
✅ Claude **only** recessively loads `CLAUDE.md` files. We should set instructions on the most important files for it to load before prompting for a code review. See [Manage Claude's memory](https://docs.claude.com/en/docs/claude-code/memory) for more in-depth details.
✅ Instruct Claude to leverage our reviewer guidelines & emoji callouts
✅ Limit the amount of praise about a code change (Claude is really good at that 😂), and focus more on improvements
✅ Instruct Claude to correctly review subsequent commits for any reopened or new issues.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,7 +47,7 @@
 - **Database update**: `pwsh dev/migrate.ps1`
 - **Generate OpenAPI**: `pwsh dev/generate_openapi_files.ps1`
 
-## Code Review Checklist
+## Development Workflow
 
 - Security impact assessed
 - xUnit tests added / updated


CLAUDE.md
theMickster, line 1: Goal is to attempt to move us towards more consistency between our Claude.md files in our repos. Our top-level files should have consistent high-level instructions. We should avoid obvious commands (Claude knows how to build a project using the dotnet CLI..).

CLAUDE.md
```diff
@@ -1,23 +1,28 @@
...
+
 - **NEVER** log or expose sensitive data: No PII, passwords, keys, or vault data in logs or error messages
+
 - **ALWAYS** use secure communication channels: Enforce confidentiality, integrity, and authenticity
+
 - **ALWAYS** encrypt sensitive data: All vault data must be encrypted at rest, in transit, and in use
-- **ALWAYS** prioritize cryptographic integrity and data protection
-- **ALWAYS** add unit tests (with mocking) for any new feature development
 
-## Project Context
```
theMickster, line 14: I do not think that these Project Context notes are necessary. 
I do not think that these will help Claude make better decisions about our code. 
All information is easily inferred.

CLAUDE.md
```diff
@@ -34,39 +39,15 @@
...
 - **Database update**: `pwsh dev/migrate.ps1`
 - **Generate OpenAPI**: `pwsh dev/generate_openapi_files.ps1`
 
-## Code Review Checklist
-
-- Security impact assessed
-- xUnit tests added / updated
-- Performance impact considered
-- Error handling implemented
-- Breaking changes documented
```
theMickster, line 51: This is redundant because we are prompting Claude to specifically do these in the review-code.yml.
As we implement additional actions in the future, we do not want these here because the instructions do not necessarily apply to every type of action we would want to run (e.g. when we have Claude evaluate an Issue, then a Code Review Checklist does not apply).

CLAUDE.md
```diff
@@ -34,39 +39,15 @@
...
-- CODEOWNERS file respected
-
 ### Key Architectural Decisions
 
 - Use .NET nullable reference types (ADR 0024)
 - TryAdd dependency injection pattern (ADR 0026)
 - Authorization patterns (ADR 0022)
 - OpenTelemetry for observability (ADR 0020)
 - Log to standard output (ADR 0021)
-
```
theMickster, line 63: We have established that there is too much information in these wikis to be of use to Claude. 
We also now understand that Claude does not pull these into memory when it runs. 
We will work towards more specific code examples from these documents as we refine our Claude markdown. But for now, they are being removed.

CLAUDE.md
withinfocus: This file will be used by Claude when developing though. I have found this kind of thing to be essential to guide it through its work. It should not be labeled as a "code review checklist" but it's what's expected to be done during development too.

CLAUDE.md
withinfocus: How was it determined that Claude doesn't use these? I have personally seen it do so.

theMickster: Then I either misread or misunderstood the documentation around how Claude Code fetches outside such as these. And even asking Claude itself multiple times whether or not these references are fetched without a specific command to fetch straight-up told me that is doesn't 🤷🏼

Either way, there is a lot of information in those wikis that Claude could be pulling to perform a code review. Are we able to determine how costly it is to use them? 

So do you want to keep them in all of the repos or leave them out? I think we need to be consistent across repos

withinfocus: To reiterate, this file is used for development activities as well as review. I have asked Claude Code in VS Code to make a certain change and seen it retrieve coding style standards to do it. All repos therefore should have these kinds of things in their instructions.

CLAUDE.md
theMickster: Chalk it up to naive thoughts on how an AI LLM trained to develop code should work I guess, but the instructions feel obvious and redundant to me 🤷🏼

withinfocus: There's a nuance to it for sure, and prompts change a lot of its work too, but what is most important to me here is that a change may not be holistic / total if you either don't tell it with precision what to do or not have this kind of content. We want an end engineer not necessarily having to repetitively say to write unit tests ... using xUnit, etc.

CLAUDE.md
```diff
@@ -34,25 +39,11 @@
...
-## Common Commands
+## Non-Obvious Commands
 
-- **Build**: `dotnet build`
-- **Test**: `dotnet test`
-- **Run locally**: `dotnet run --project src/Api`
 - **Database update**: `pwsh dev/migrate.ps1`
 - **Generate OpenAPI**: `pwsh dev/generate_openapi_files.ps1`
 
-## Code Review Checklist
```
withinfocus, line 45: 🎨 I have called this "Development Workflow" and then listed out steps to perform. That may work better.

theMickster: Renamed :-)

PR #6443, opened 2025-10-10
[PM-26316] Prevent users from sharing archived cipher

nick-livefront:
## 🎟️ Tracking

[PM-26316](https://bitwarden.atlassian.net/browse/PM-26316)

## 📔 Objective

Throw an error when a user attempts to share an archived cipher to an organization.
- Archiving is only available for personal items. 

## 📸 Screenshots

Testable via the cli:
<video src="https://github.com/user-attachments/assets/86e73a80-25e8-420b-845c-ba6330724194" />

## 🦮 Reviewer guidelines

[PM-26316]: https://bitwarden.atlassian.net/browse/PM-26316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Api/Vault/Controllers/CiphersController.cs
+++ b/src/Api/Vault/Controllers/CiphersController.cs
@@ -1285,6 +1285,11 @@ public class CiphersController : Controller
 
             ValidateClientVersionForFido2CredentialSupport(existingCipher);
 
+            if (existingCipher.ArchivedDate.HasValue)
+            {
+                throw new BadRequestException("Cannot move archived items to an organization.");
+            }
+
             shareCiphers.Add((cipher.ToCipherDetails(existingCipher), cipher.LastKnownRevisionDate));
         }
 
--- a/test/Api.Test/Vault/Controllers/CiphersControllerTests.cs
+++ b/test/Api.Test/Vault/Controllers/CiphersControllerTests.cs
@@ -1821,6 +1821,51 @@ public class CiphersControllerTests
     }
 
     [Theory, BitAutoData]
+    public async Task PutShareMany_ExistingCipherArchived_ThrowsBadRequestException(
+        Guid organizationId,
+        Guid userId,
+        CipherWithIdRequestModel request,
+        SutProvider<CiphersController> sutProvider)
+    {
+        // Request model does not have ArchivedDate (only the existing cipher does)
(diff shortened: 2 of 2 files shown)


src/Api/Vault/Controllers/CiphersController.cs
```diff
@@ -1263,6 +1268,11 @@ public async Task<ListResponseModel<CipherMiniResponseModel>> PutShareMany([From
                 _logger.LogError("Cipher was not encrypted for the current user. CipherId: {CipherId}, CurrentUser: {CurrentUserId}, EncryptedFor: {EncryptedFor}", cipher.Id, userId, cipher.EncryptedFor);
                 throw new BadRequestException("Cipher was not encrypted for the current user. Please try again.");
             }
+
+            if (cipher.ArchivedDate.HasValue)
+            {
+                throw new BadRequestException("Cannot move archived items to an organization.");
+            }
```
nikwithak, line 1275: Here you're validating the ArchivedDate passed from the request (`cipher in model.Ciphers`), which [doesn't look like it gets saved downstream](https://github.com/bitwarden/server/blob/vault/pm-26316/share-archived-cipher/src/Infrastructure.Dapper/Vault/Repositories/CipherRepository.cs#L451-L460) when written to the DB.

Whereas in  `PutShare` you're validating against [the cipher from `_cipherRepository`](https://github.com/bitwarden/server/pull/6443/files#diff-2131dc80a94eb2fe7b30c7ea3d0bb0623e1a1d4743773fb354da670414680146R740), which guarantees the `ArchivedDate` isn't set in the backend.

In this case I think the cipher could enter an inconsistent state if the Cipher in the DB has `ArchivedDate` set, but it's not in the request.

nick-livefront: 👏 👏 Nice catch! I added a check for the stored cipher `ArchiveDate` as well. [6b1ba79](https://github.com/bitwarden/server/pull/6443/commits/6b1ba792d55f293d8fbb749e170397b581f20af4)

I left the initial check even though the request model isn't stored, avoiding any future bugs if the downstream logic were to change.

PR #6465, opened 2025-10-17
Azure Service Bus Local Batching

brant-livefront:
## 📔 Objective

When turning on the new events integration in production, we were witnessing latency spikes on various services that need to make calls into EventWriteSerivce (mostly for single events via CreateAsync). The spikes seemed to be due to the fact that we couldn't write to ASB fast enough. Part of that may have been throttling on the standard tier ASB instance, but we also wanted to make sure we were handling writes as fast as possible.

This PR implements a local batching solution that will batch messages up before sending them as a batch in a single AMQP frame. ASB has slightly higher latency than the previous Azure Queues because it has to deal with AMQP overhead that is slightly more than the simpler REST calls for Azure Queue. However, there is support in ASB for batching multiple messages into a single frame, which should allow us to accept writes faster but still deliver everything quickly to ASB. It should also reduce network overhead since each individual message no longer requires a separate call. 

In essence, this is the same principle we used in #6403 to reduce network on the listener side, just applied to the sender. There is no built in support for building batches locally, however, so we have to use a concurrent queue and handle that ourselves.

**Note:** I actually implemented this first entirely inside of AzureServiceBusService using a Timer based flush. After having finished that version, I realized that the queue is likely to be constantly filled with new messages coming in and it would simplify things a bit to just have a separate, persistent worker batching and sending constantly (i.e. we don't have to worry about a semaphore or locking or timing, just keep sending batches as long as we have something to send). I pushed this up as the latest commit. But the first commit is still there if we think it's a better approach.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusEventBatchBackgroundService.cs
+++ b/src/Core/AdminConsole/Services/Implementations/EventIntegrations/AzureServiceBusEventBatchBackgroundService.cs
@@ -1,0 +1,96 @@
+﻿using System.Collections.Concurrent;
+using Azure.Messaging.ServiceBus;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Bit.Core.Services;
+
+public class AzureServiceBusEventBatchBackgroundService(ServiceBusSender sender, int batchSize, ILogger logger) : BackgroundService
+{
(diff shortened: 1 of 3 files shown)


brant-livefront:
Closing in favor of #6477 - we are going to use partitioning to increase throughput instead of local batching. With partitioning, we can't locally batch (because batches can't contain multiple partition keys).

PR #6472, opened 2025-10-21
[PM-26430] Remove Type property from PolicyRequestModel to use route parameter only

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26430

## 📔 Objective

This PR removes the `PolicyType` field from the Policy update endpoint request models (`PolicyRequestModel` and `SavePolicyRequest`). The policy type is now taken from the URL (`{type}`) instead of the request body.

What changed:
- `PolicyType` removed from `PolicyRequestModel` and `SavePolicyRequest`
- `ToPolicyUpdateAsync()` and `ToSavePolicyModelAsync()` now take `PolicyType` as a parameter instead of reading it from the request
- Controller passes `PolicyType` from the route parameter to the request methods

I chose to remove the property from the request body and keep it in the URL path because this approach provides better backwards compatibility. Existing API consumers can continue using the same endpoint URLs without needing to modify their request payloads. 

[Clients PR](https://github.com/bitwarden/clients/pull/16960).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/test/Api.Test/Controllers/PoliciesControllerTests.cs
+++ b/test/Api.Test/Controllers/PoliciesControllerTests.cs
@@ -487,14 +487,14 @@ public class PoliciesControllerTests
             .Returns(policy);
 
         // Act
-        var result = await sutProvider.Sut.PutVNext(orgId, model);
+        var result = await sutProvider.Sut.PutVNext(orgId, policy.Type, model);
 
         // Assert
         await sutProvider.GetDependency<IVNextSavePolicyCommand>()
             .Received(1)
             .SaveAsync(Arg.Is<SavePolicyModel>(
                 m => m.PolicyUpdate.OrganizationId == orgId &&
-                     m.PolicyUpdate.Type == model.Policy.Type &&
+                     m.PolicyUpdate.Type == policy.Type &&
                      m.PolicyUpdate.Enabled == model.Policy.Enabled &&
                      m.PerformedBy.UserId == userId &&
                      m.PerformedBy.IsOrganizationOwnerOrProvider == true));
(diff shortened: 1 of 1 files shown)


src/Api/AdminConsole/Models/Request/PolicyRequestModel.cs
```diff
@@ -9,20 +9,18 @@ namespace Bit.Api.AdminConsole.Models.Request;
 
 public class PolicyRequestModel
 {
-    [Required]
-    public PolicyType? Type { get; set; }
```
JimmyVo16, line 13: Question: How will this affect the response object? I think right now both the request and response share the same class. If we’re removing it from the request, then that might cause issues since the type seems important for consumers.

r-tome: It won't affect the response objects at all, they're separate classes: https://github.com/bitwarden/server/blob/main/src/Api/AdminConsole/Public/Models/Response/PolicyResponseModel.cs and https://github.com/bitwarden/server/blob/main/src/Api/AdminConsole/Models/Response/Organizations/PolicyResponseModel.cs

Client response model https://github.com/bitwarden/clients/blob/main/libs/common/src/admin-console/models/response/policy.response.ts

JimmyVo16 (dismissed):

src/Api/AdminConsole/Authorization/AuthorizationHandlerCollectionExtensions.cs
r-tome, line 1: This was changed by `dotnet format`

PR #6477, opened 2025-10-22
Refactor Azure Service Bus to use the organization id as a partition key

brant-livefront:
## 📔 Objective

This PR adds a Partition Key to the Azure Service Bus messages. By using the Organization Id for the partition key, we consistently route all of an organization's events to the same partition. This gives us a best-effort shot at preserving the order of events (i.e. sending them out in the order they arrived). Using ASB Sessions would give us more guaranteed FIFO ordering, but it adds a lot of complexity and could add concurrency / latency issues.

By using partitioning in production, ASB will shard our topics across multiple partitions. This allows us greater aggregate throughput than simply using one message broker / store. ASB will choose the correct partition for us based on an internal hash of the organization id. We're abstracted away from the details, but the important part is that any message with the same partition key will be hashed consistently to the same partition.

**Note**: This is probably an alternative approach to #6465. If we choose to go with partitions, it's unlikely we'd be able to batch messages locally into a single frame because every message would need to have the same partition key. That would add a tremendous amount of complexity (multiple local stores based on organization ids) and it might not have any benefits (if we receive lots of messages for different organizations, we'd end up sending them one at a time anyway). The hope would be that the greater throughput of partitions and scaling messaging units would give us the reduced latency that the local batching was hoping to provide.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationEventWriteService.cs
```diff
@@ -14,15 +14,21 @@ public EventIntegrationEventWriteService(IEventIntegrationPublisher eventIntegra
     public async Task CreateAsync(IEvent e)
     {
         var body = JsonSerializer.Serialize(e);
-        await _eventIntegrationPublisher.PublishEventAsync(body: body);
+        await _eventIntegrationPublisher.PublishEventAsync(body: body, organizationId: e.OrganizationId?.ToString() ?? string.Empty);
```
withinfocus: ❓ Is an empty string appropriate? Or `null`?

brant-livefront: Good catch. yeah, I think `null` is better in this case. Empty string may end up being hashed and we send _all_ of these to one partition, whereas `null` should default to round robin. Changed in latest commit.

PR #6481, opened 2025-10-22
[PM-21153] Recovery Code Not Functioning for SSO-required Users

enmande:
## 🎟️ Tracking

[PM-21153](https://bitwarden.atlassian.net/browse/PM-21153)

## 📔 Objective

> [!NOTE]
> This is supported by clients pr [17016](https://github.com/bitwarden/clients/pull/17016)

### Background

As part of `VerifyTwoFactorAsync`, the `TwoFactorAuthenticationValidator` will remove 2FA methods and fast-forward a non-SSO user through to 2FA management. This proves authentication, encourages them to revisit their 2FA posture after recovery, and presents an immediate opportunity to obtain their new recovery code.

This is not currently possible for users constrained by an organization policy of SSO Required; SSO validity is assessed ahead of 2FA validity, so SSO-required users are failed out of the validation process by SSO requirement.

### Changes

🎏 This refactor is feature-flagged to preserve existing validation logic for safety during production rollout. Flag should be safely disable-able without affecting user state.

This updates `BaseRequestValidator` to allow for definable and composable validation scenarios for edge-cases such as these. These edge-cases are expected to be extraordinary and uncommon, but the composition pattern puts clarity out front, as did the linear procedure before.

The compositional nature of the update makes it somewhat easier to walk into desired future patterns of refactoring validation scenarios into their own validators as part of iterative improvement.

### Non-Org User, Without Pending Client Update in Linked PR
Showing here, support for enablement in older clients. 

https://github.com/user-attachments/assets/2c273d2e-9da0-4038-9727-40093c20e230

### Non-Org User, With Pending Client Update in Linked PR

https://github.com/user-attachments/assets/8b42cd62-dbd7-4485-85f8-3dc27fec501d

### Org User With SSO Requirement, Without Pending Client Update in Linked PR
Showing here, support for enablement in older clients. Not as elegant as the redirect, but a middle ground which surfaces information to the user.

https://github.com/user-attachments/assets/0baf7f73-29a1-4e0e-b615-3559deed3108

### Org User With SSO Requirement, With Pending Client Update in Linked PR

https://github.com/user-attachments/assets/e73ee7c3-23c4-4726-b96b-9aca312c4b2d

### TDE User login (/token)
Showing here, TDE login and device approval, an unaffected validation scheme route, remains unaffected.

https://github.com/user-attachments/assets/94958312-3b9d-4178-94de-c1640f75bd9f

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-21153]: https://bitwarden.atlassian.net/browse/PM-21153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Identity.Test/AutoFixture/RequestValidationFixtures.cs
+++ b/test/Identity.Test/AutoFixture/RequestValidationFixtures.cs
@@ -1,6 +1,7 @@
 ﻿using System.Reflection;
 using AutoFixture;
 using AutoFixture.Xunit2;
+using Bit.Identity.IdentityServer;
 using Duende.IdentityServer.Validation;
 
 namespace Bit.Identity.Test.AutoFixture;
@@ -8,7 +9,8 @@ namespace Bit.Identity.Test.AutoFixture;
 internal class ValidatedTokenRequestCustomization : ICustomization
 {
     public ValidatedTokenRequestCustomization()
-    { }
+    {
+    }
 
     public void Customize(IFixture fixture)
     {
@@ -22,10 +24,45 @@ internal class ValidatedTokenRequestCustomization : ICustomization
 public class ValidatedTokenRequestAttribute : CustomizeAttribute
 {
     public ValidatedTokenRequestAttribute()
-    { }
+    {
+    }
 
     public override ICustomization GetCustomization(ParameterInfo parameter)
     {
         return new ValidatedTokenRequestCustomization();
+    }
+}
+
+internal class CustomValidatorRequestContextCustomization : ICustomization
+{
+    public CustomValidatorRequestContextCustomization()
+    {
+    }
+
+    /// <summary>
+    /// Specific context members like <see cref="CustomValidatorRequestContext.RememberMeRequested" />,
+    /// <see cref="CustomValidatorRequestContext.TwoFactorRecoveryRequested"/>, and
+    /// <see cref="CustomValidatorRequestContext.SsoRequired" /> should initialize false,
+    /// and are made truthy in context upon evaluation of a request. Do not allow AutoFixture to eagerly make these
+    /// truthy; that is the responsibility of the <see cref="Bit.Identity.IdentityServer.RequestValidators.BaseRequestValidator{T}" />
+    /// </summary>
+    public void Customize(IFixture fixture)
+    {
+        fixture.Customize<CustomValidatorRequestContext>(composer => composer
+            .With(o => o.RememberMeRequested, false)
+            .With(o => o.TwoFactorRecoveryRequested, false)
+            .With(o => o.SsoRequired, false));
+    }
+}
+
+public class CustomValidatorRequestContextAttribute : CustomizeAttribute
+{
+    public CustomValidatorRequestContextAttribute()
+    {
+    }
+
+    public override ICustomization GetCustomization(ParameterInfo parameter)
+    {
+        return new CustomValidatorRequestContextCustomization();
     }
 }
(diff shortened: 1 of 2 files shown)


src/Identity/IdentityServer/CustomValidatorRequestContext.cs
```diff
@@ -42,10 +42,23 @@ public class CustomValidatorRequestContext
...
-
     /// <summary>
     /// A validated auth request
     /// <see cref="AuthRequest.IsValidForAuthentication"/>
     /// </summary>
     public AuthRequest ValidatedAuthRequest { get; set; }
+    /// <summary>
+    /// Validation methods, by name as string, which have passed validation in the current context.
+    /// </summary>
+    public List<string> CompletedValidationSchemes { get; set; } = [];
```
ike-kottlowski: ❓ : This doesn't seem to be used anywhere other than assignments after successful completion of validators. Is there a future purpose you envision this array being used for? Otherwise, I'm not sure it's needed in the current body of work.

PJHOMEZ (approved):

src/Identity/IdentityServer/CustomValidatorRequestContext.cs
enmande: @ike-kottlowski we do use it once, [here](https://github.com/bitwarden/server/blame/09b5a8c5580434033aecfc3e4f83832cdec53e40/src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs#L369) for branching logic/context awareness as part of this specific fix.

That said, that is its full purpose currently. I would totally understand if the recommendation were to re-think this marker as a flag the way `rememberMe` functions today. I wrote it this way trying to balance context awareness for this newer pattern against continuing to add flags to the context. I just didn't want to encourage the context to become confusing with situational flags that could become cross-purposes at some (undefined, uncertain) point in the future.

There is a compelling argument that specific flags bring clarity, and it could be considered a case of bike-shedding to use this completed schemes since we aren't logging, reporting , or doing anything else with that information at this time, and don't have plans for it beyond "use this as a tool to calculate edge-cases when they present."

ike-kottlowski: I think using the flag in the context is the simplest step to take forward. Adding the array is a good idea but maybe adds too much noise at this current junction.

The flag `context.TwoFactorRequired = true` already tells us the `TwoFactorAuthenticationValidator` has ran _and_ passed. This is an inference, so it is not as direct as the array approach, which is explicit, and a better approach IMO. 

I think a well-documented tech-debt task is the best way to move toward tracking the state of Authentication Validators. This way we can implement it in all places (2FA, NDV, SSO Policy) at one time. Doing it in the context of this bug feels like we have too many half-implemented solutions (2FA validator being the prime example) which makes maintaining more difficult.

enmande: Thank you for the considered recommendation! This is already moving a lot around, and in context of the ripples of implementation you highlighted here, it's a tidy and workable path forward. I've updated the business logic around this flag use in 9a3450e536585747f1b960322c68b02fab0763d5, updated the test suite, and re-run my screenshot scenarios locally to make sure it's still on the rails. I will see about getting something on the board in the next day for the proposed Tech Debt task as well.

ike-kottlowski (approved):

PR #6488, opened 2025-10-23
[PM-26636] - Auto Confirm Org User Command

jrmccannon:
## 🎟️ Tracking
[PM-26636](https://bitwarden.atlassian.net/browse/PM-26636)

## 📔 Objective
Add the automatic organization user confirmation feature that allows client applications to automatically confirm organization users. The implementation includes:

  - A new API endpoint (POST /organizations/{orgId}/users/{id}/auto-confirm)
  - A command pattern implementation (AutomaticallyConfirmOrganizationUserCommand) that handles the confirmation workflow including validation, key exchange, event logging, email notifications, device registration cleanup, and optional default collection creation
  - Comprehensive validation logic (AutomaticallyConfirmOrganizationUsersValidator) that enforces organizational policies including Single Organization Policy and Two-Factor Authentication requirements
  - Idempotent confirmation operation that safely handles duplicate requests
  - A new OrganizationUser_AutomaticallyConfirmed event type for audit logging
  - Integration with the v2 utilities framework for standardized command/validation patterns
  - Full test coverage including unit tests, integration tests, and controller tests

  The feature is protected behind the AutomaticConfirmUsers feature flag and requires the ManageUsers authorization requirement.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26636]: https://bitwarden.atlassian.net/browse/PM-26636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUsers/AutomaticallyConfirmOrganizationUsersValidatorTests.cs
+++ b/test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUsers/AutomaticallyConfirmOrganizationUsersValidatorTests.cs
@@ -228,7 +228,7 @@ public class AutomaticallyConfirmOrganizationUsersValidatorTests
     [BitAutoData(OrganizationUserType.Owner)]
     [BitAutoData(OrganizationUserType.Custom)]
     [BitAutoData(OrganizationUserType.Admin)]
-    public async Task ValidateAsync_WithAdminType_ReturnsUserIsNotUserTypeError(
+    public async Task ValidateAsync_WithNonUserType_ReturnsUserIsNotUserTypeError(
         OrganizationUserType userType,
         SutProvider<AutomaticallyConfirmOrganizationUsersValidator> sutProvider,
         Organization organization,


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -691,6 +699,42 @@ public async Task PatchBulkEnableSecretsManagerAsync(Guid orgId,
...
+    [Authorize<ManageUsersRequirement>]
+    [HttpPost("{id:guid}/auto-confirm")]
+    [RequireFeature(FeatureFlagKeys.AutomaticConfirmUsers)]
+    public async Task AutomaticallyConfirmOrganizationUserAsync(Guid orgId, Guid id, [FromBody] OrganizationUserConfirmRequestModel model)
+    {
+        var userId = _userService.GetProperUserId(User);
+
+        if (userId is null || userId.Value == Guid.Empty)
+        {
+            throw new UnauthorizedAccessException();
```
r-tome: Could possibly be an attribute? Just something to think about

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -691,6 +699,42 @@ public async Task PatchBulkEnableSecretsManagerAsync(Guid orgId,
         await BulkEnableSecretsManagerAsync(orgId, model);
     }
 
+
+    [Authorize<ManageUsersRequirement>]
+    [HttpPost("{id:guid}/auto-confirm")]
```
r-tome: I remember a discussion about this https://bitwarden.slack.com/archives/C02LPDKPML6/p1641389870010400?thread_ts=1641386236.008400&cid=C02LPDKPML6

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -691,6 +699,42 @@ public async Task PatchBulkEnableSecretsManagerAsync(Guid orgId,
...
+                PerformedBy = new StandardUser(userId.Value, await _currentContext.OrganizationOwner(orgId)),
+                PerformedOn = _timeProvider.GetUtcNow()
+            });
+
+        if (result is { IsError: true })
+        {
+            switch (result.AsError)
+            {
+                case OrganizationNotFound organizationNotFound:
+                    throw new NotFoundException(organizationNotFound.Message);
```
r-tome: How does the return message look like? https://github.com/bitwarden/server/blob/ff4b3eb9e5896f74aa4499da7378d66a90b314f6/src/Api/Utilities/ExceptionHandlerFilterAttribute.cs#L121

Should we throw an exception on the controller or `return NotFound()`

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IAutomaticallyConfirmOrganizationUserCommand.cs
r-tome, line 1: Xmldoc missing

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
jrmccannon: Ah. I see now. Well that's not great. 

Should we allow for the overriding of that not found message? Or is that something that needs to be run by architecture?

r-tome: I think its wise to run by architecture

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Interfaces/IAutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.AutoConfirmUser;
+using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.DeleteClaimedAccount;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
+
+/// <summary>
+/// Command to automatically confirm an organization user.
+/// </summary>
```
eliykat, line 8: More context around what autoconfirm is would be helpful - `<remarks>` is good for this. e.g.

> The autoconfirm feature sends push notifications to eligible logged in client apps to automatically confirm OrganizationUsers without manual intervention by administrators. When a client app receives the push notification, it performs the key exchange and sends an autoconfirm request to the server. This command is used to process those requests. It should not be used to confirm users in any other context.

I think a `README` in the `AutoConfirmUser` folder would also be useful to document the feature in more detail. I'll raise it in the dev channel.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,200 @@
...
+    IFeatureService featureService,
+    IPolicyRequirementQuery policyRequirementQuery,
+    ICollectionRepository collectionRepository,
+    ILogger<AutomaticallyConfirmOrganizationUserCommand> logger) : IAutomaticallyConfirmOrganizationUserCommand
+{
+    public async Task<CommandResult> AutomaticallyConfirmOrganizationUserAsync(AutomaticallyConfirmOrganizationUserRequest request)
+    {
+        var requestData = await RetrieveDataAsync(request);
+
+        if (requestData.IsError) return requestData.AsError;
```
eliykat: Here, I got you these: `{` and `}`

(Please avoid single line `if` statements.)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,200 @@
...
+
+        var validatedData = await validator.ValidateAsync(requestData.AsSuccess);
+
+        if (validatedData.IsError) return validatedData.AsError;
+
+        var validatedRequest = validatedData.Request;
+
+        var successfulConfirmation = await organizationUserRepository.ConfirmOrganizationUserAsync(validatedRequest.OrganizationUser);
+
+        if (!successfulConfirmation) return new None();
```
eliykat: Please add comments to explain our approach here

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,200 @@
...
+                    OrganizationId = request.Organization.Id,
+                    Name = request.DefaultUserCollectionName,
+                    Type = CollectionType.DefaultUserCollection
+                },
+                groups: null,
+                [new CollectionAccessSelection
+                {
+                    Id = request.OrganizationUser.Id,
+                    Manage = true
+                }]);
```
eliykat, line 94: 💭 (non-blocking) Thoughts on making this an idempotent repository method as well? e.g. `CreateDefaultCollectionAsync`. Given that we are creating it in multiple places and creating default collections properly is quite important. (not this PR though)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,200 @@
...
+    {
+        var organizationUser = await GetOrganizationUserAsync(request.OrganizationUserId);
+
+        if (organizationUser.IsError) return organizationUser.AsError;
+
+        var organization = await GetOrganizationAsync(request.OrganizationId);
+
+        if (organization.IsError) return organization.AsError;
+
+        return new AutomaticallyConfirmOrganizationUserRequestData
```
eliykat: I suggest that it's clearer to call this the `ValidationRequest`; it's an enriched DTO required by the validator. The overall flow of data then looks like: API request -> Command Request -> Validation Request -> Validation Result -> Command Result -> API response... which is nice and symmetrical. The `Data` suffix is not very clear by comparison. (it's not a data model)

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -691,6 +699,56 @@ public async Task PatchBulkEnableSecretsManagerAsync(Guid orgId,
...
     }
 
+
+    [Authorize<ManageUsersRequirement>]
+    [HttpPost("{id}/auto-confirm")]
+    public async Task<IActionResult> AutomaticallyConfirmOrganizationUserAsync([FromRoute] Guid orgId,
+        [FromRoute] Guid id,
+        [FromBody] OrganizationUserConfirmRequestModel model)
+    {
+        if (!_featureService.IsEnabled(FeatureFlagKeys.AutomaticConfirmUsers))
```
r-tome: This should be replaced with `[RequireFeature(FeatureFlagKeys.AutomaticConfirmUsers)]`

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
r-tome: I would add a comment here to explain that the only reason this would not be successful is if the user is already confirmed. Otherwise it seems like silent failure.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,200 @@
...
+            .MapAsync(PushSyncOrganizationKeysAsync)
+            .ToResultAsync();
+    }
+
+    private async Task<CommandResult<AutomaticallyConfirmOrganizationUserRequestData>> CreateDefaultCollectionsAsync(
+        AutomaticallyConfirmOrganizationUserRequestData request)
+    {
+        try
+        {
+            if (!featureService.IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
```
r-tome: Extract all these conditions into a private method

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserRequest.cs
r-tome, line 1: Filename is misspelled

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -691,6 +699,56 @@ public async Task PatchBulkEnableSecretsManagerAsync(Guid orgId,
...
+                OrganizationUserId = id,
+                Key = model.Key,
+                DefaultUserCollectionName = model.DefaultUserCollectionName,
+                PerformedBy = new StandardUser(userId.Value, await _currentContext.OrganizationOwner(orgId)),
+                PerformedOn = _timeProvider.GetUtcNow()
+            });
+
+        if (result is { IsError: true })
+        {
+            return result.AsError switch
```
r-tome: The Billing team has a helper method in `BaseBillingController.Handle`, we should do something similar

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
r-tome, line 1: We should add unit tests for this command

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
eliykat: I think it's also worth explaining here that this is idempotent and designed to handle multiple simultaneous requests from logged in clients.

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
```diff
@@ -0,0 +1,53 @@
...
+        commandResult.Match<IResult>(
+            error => error switch
+            {
+                BadRequestError badRequest => TypedResults.BadRequest(new ErrorResponseModel(badRequest.Message)),
+                NotFoundError notFound => TypedResults.NotFound(new ErrorResponseModel(notFound.Message)),
+                InternalError internalError => TypedResults.Json(
+                    new ErrorResponseModel(internalError.Message),
+                    statusCode: StatusCodes.Status500InternalServerError),
+                _ => TypedResults.Json(
+                    new ErrorResponseModel(commandResult.AsError.Message),
```
eliykat: `AsError` is not typesafe, sometimes it can't be helped but here you can use the inner variable `error` to access this in a typesafe way. (as is the intention of `Match`)

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
```diff
@@ -0,0 +1,53 @@
...
+                    statusCode: StatusCodes.Status500InternalServerError
+                )
+            },
+            _ => TypedResults.NoContent()
```
eliykat: This is the "non error" case, so presumably the command has succeeded; should this be an OK response instead?

eliykat: 💭 (non-blocking) Also, given that this is `CommandResult<T>`, it suggests that the response to the client should include content. Maybe it could take a `Func<T, R>` to handle the conversion to the api response model in the non-error case.

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
```diff
@@ -0,0 +1,53 @@
...
+        commandResult.Result.Match<IResult>(
+           error => error switch
+           {
+               NotFoundError notFoundError => TypedResults.NotFound(new ErrorResponseModel(notFoundError.Message)),
+               _ => TypedResults.BadRequest(new ErrorResponseModel(error.Message))
+           },
+           _ => TypedResults.NoContent()
+        );
+
+    protected static IResult Handle(CommandResult commandResult) =>
```
eliykat, line 10: Same feedback as above for `Handle(CommandResult<T>)`.

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    protected static IResult Handle(BulkCommandResult commandResult) =>
+        commandResult.Result.Match<IResult>(
+           error => error switch
+           {
+               NotFoundError notFoundError => TypedResults.NotFound(new ErrorResponseModel(notFoundError.Message)),
+               _ => TypedResults.BadRequest(new ErrorResponseModel(error.Message))
+           },
+           _ => TypedResults.NoContent()
+        );
```
eliykat: This is not generally how we'd handle a `BulkCommandResponse`; see the `BulkDeleteAccount` endpoint as a better example. (i.e. it's usually mapped to a BulkResponseModel with an OK response because there is no single success or failure.)

I see that you're currently using this on the (non-bulk) `DeleteAccount` endpoint which is why it "works", but I think that indicates that the non-bulk endpoint shouldn't be handling a `BulkCommandResponse`.

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
eliykat, line 1: 🎉  Thanks for starting this!

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -544,14 +551,7 @@ public async Task<IResult> DeleteAccount(Guid orgId, Guid id)
...
 
-        var commandResult = await _deleteClaimedOrganizationUserAccountCommand.DeleteUserAsync(orgId, id, currentUserId.Value);
-
-        return commandResult.Result.Match<IResult>(
-            error => error is NotFoundError
-                ? TypedResults.NotFound(new ErrorResponseModel(error.Message))
-                : TypedResults.BadRequest(new ErrorResponseModel(error.Message)),
-            TypedResults.Ok
-        );
+        return Handle(await _deleteClaimedOrganizationUserAccountCommand.DeleteUserAsync(orgId, id, currentUserId.Value));
```
eliykat: 🎨 (non-blocking) assigning the result to a `var commandResult` variable makes for easier debugging.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -691,6 +691,32 @@ public async Task PatchBulkEnableSecretsManagerAsync(Guid orgId,
...
+
+        return Handle(await _automaticallyConfirmOrganizationUserCommand.AutomaticallyConfirmOrganizationUserAsync(
+            new AutomaticallyConfirmOrganizationUserRequest
+            {
+                OrganizationId = orgId,
+                OrganizationUserId = id,
+                Key = model.Key,
+                DefaultUserCollectionName = model.DefaultUserCollectionName,
+                PerformedBy = new StandardUser(userId.Value, await _currentContext.OrganizationOwner(orgId)),
+                PerformedOn = _timeProvider.GetUtcNow()
```
eliykat: Passing in the date from the controller level seems unexpected to me. It's not information from the request or otherwise from the API layer (like claims in `currentContext`). Is there any reason the command can't get this for itself? That would also make the date more accurate because it could be fetched at the time of performing the operation. (even though the difference is probably minimal)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        if (!successfulConfirmation) return new None();
+
+        return await validatedRequest.ToCommandResultAsync()
+            .MapAsync(CreateDefaultCollectionsAsync)
+            .MapAsync(LogOrganizationUserConfirmedEventAsync)
+            .MapAsync(SendConfirmedOrganizationUserEmailAsync)
+            .MapAsync(DeleteDeviceRegistrationAsync)
+            .MapAsync(PushSyncOrganizationKeysAsync)
+            .ToResultAsync();
```
eliykat: It's worth considering error handling here. You do handle all errors, but then each private method here is returning an `Error` object, which will shortcut all following steps. But do we want to skip sending emails just because the event log failed; or skip logging the event because we couldn't create the default collection?

I tend to think that all these methods here should log but otherwise swallow errors. We've already confirmed the user, we should do as many side effects that we can. (This should be noted in a comment.)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserRequest.cs
```diff
@@ -0,0 +1,33 @@
...
+    public required string DefaultUserCollectionName { get; init; }
+    public required IActingUser PerformedBy { get; init; }
+    public required DateTimeOffset PerformedOn { get; init; }
+}
+
+public record AutomaticallyConfirmOrganizationUserRequestData
+{
+    public required OrganizationUser OrganizationUser { get; init; }
+
+    public Guid UserId => OrganizationUser.UserId ?? throw new ArgumentNullException(nameof(OrganizationUser.UserId));
```
eliykat: I'm not sure about this; it's a getter that might throw an error, not when the class is created, but when it's called at a later point in time? That seems worse than a Guid? that the caller can handle.
I assume you're trying to avoid representing invalid types, but I don't think this is doing that.

(non-blocking, probably out of scope but planting the seed 🌱 )  It's pretty common that we require a certain user state in certain code flows; I think the better move is to start making specific API and domain models to represent this, e.g. AcceptedOrganizationUser where UserId cannot be null. Arguably our current models are too closely tied to the database schema.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
eliykat, line 1: We are only auto-confirming `User` roles - this needs to be added as validation.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+        return await OrganizationUserOwnershipValidationAsync(request)
+            .ThenAsync(OrganizationUserConformsToOrganizationPoliciesAsync);
+    }
+
+    private async Task<ValidationResult<AutomaticallyConfirmOrganizationUserRequestData>> OrganizationUserOwnershipValidationAsync(
+        AutomaticallyConfirmOrganizationUserRequestData request) =>
+        request.Organization.PlanType is not PlanType.Free
+        || await organizationUserRepository.GetCountByFreeOrganizationAdminUserAsync(request.UserId) == 0
+            ? Valid(request)
+            : Invalid(request, new UserToConfirmIsAnAdminOrOwnerOfAnotherFreeOrganization());
```
eliykat: This needs to check that they are an Admin/Owner of _this_ organization. But if we limit this to User roles then we probably don't need this check at all.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    private async Task<ValidationResult<AutomaticallyConfirmOrganizationUserRequestData>>
+        OrganizationUserConformsToOrganizationPoliciesAsync(AutomaticallyConfirmOrganizationUserRequestData request)
+    {
+        return await OrganizationUserConformsToTwoFactorRequiredPolicyAsync(request)
+            .ThenAsync(OrganizationUserConformsToSingleOrgPolicyAsync);
```
eliykat: 🎨 (non-blocking) this is slightly too much nesting; I suggest this could be moved to the chain in the parent method.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+            return Invalid(request, new UserIsNotAccepted());
+        }
+
+        if (orgUser.OrganizationId != request.Organization.Id)
+        {
+            return Invalid(request, new OrganizationUserIdIsInvalid());
+        }
+
+        return await OrganizationUserOwnershipValidationAsync(request)
+            .ThenAsync(OrganizationUserConformsToOrganizationPoliciesAsync);
```
eliykat: (non-blocking clarification) Why the mixed approach? I guess because `ThenAsync` only accepts Tasks, and the first few checks are synchronous?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+                .Any(p => p.OrganizationId == request.Organization.Id)
+                    ? Invalid(request, new UserDoesNotHaveTwoFactorEnabled())
+                    : Valid(request);
+        }
+
+        return (await policyRequirementQuery.GetAsync<RequireTwoFactorPolicyRequirement>(request.UserId))
+            .IsTwoFactorRequiredForOrganization(request.Organization.Id)
+                ? Invalid(request, new UserDoesNotHaveTwoFactorEnabled())
+                : Valid(request);
+    }
```
eliykat: Oh man this nesting of feature flags is rough, but also my fault. `PolicyRequirement`s is used in some places but not in others (this flag was never turned on), and it's likely going to change substantially before it's rolled out for all policies. For now I want us to avoid continuing to embed this `PolicyRequirements` feature flag switch. Either:

- use the `policyService` legacy implementation, or
- use `policyRequirements` but don't feature flag it - this work is already behind its own feature flag and will be QA'd before release. We can treat the `PolicyRequirements` flag as for legacy code only.

I am not sure which is better. We've already started using the new system, so maybe it's better to keep using it for new code, so at least we are being consistently inconsistent 🙃  (the second option) But I am open to other ideas here.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/IAutomaticallyConfirmOrganizationUsersValidator.cs
eliykat, line 1: 💭 (non-actionable) We may want to make some reusable interfaces for validators at some stage, as I expect they will all end up looking like `Task<ValidationResult<T>> ValidateAsync(T request);`

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    private async Task<ValidationResult<AutomaticallyConfirmOrganizationUserRequestData>> OrganizationUserConformsToSingleOrgPolicyAsync(
```
eliykat: Does this remove the single org exemption for owners and admins? This probably needs a bit of thought, depending on whether you use `policyService` or `policyRequirement`s. At the moment I think both code paths maintain the exemption. I think `PolicyRequirement` gives you slightly finer grained control over this without affecting other policies.

src/Core/AdminConsole/Utilities/v2/Results/CommandResult.cs
eliykat, line 1: 👏 thank you for moving this. I should've used the `v2` convention to begin with.

src/Core/AdminConsole/Utilities/v2/Results/CommandResult.cs
```diff
@@ -40,3 +40,27 @@ public record BulkCommandResult<T>(Guid Id, CommandResult<T> Result);
...
+        return Task.FromResult<CommandResult<T>>(value);
+    }
+
+    public static async Task<CommandResult<TOut>> MapAsync<TIn, TOut>(
+        this Task<CommandResult<TIn>> inputTask,
+        Func<TIn, Task<CommandResult<TOut>>> next)
+    {
+        var input = await inputTask;
+        if (input.IsError) return input.AsError;
+        return await next(input.AsSuccess);
```
eliykat: Could this use `Match` to take advantage of the `OneOf` library utils? (same for the other util methods)

test/Api.IntegrationTest/Helpers/OrganizationTestHelpers.cs
```diff
@@ -170,6 +170,25 @@ public static async Task EnableOrganizationDataOwnershipPolicyAsync<T>(
         await policyRepository.CreateAsync(policy);
     }
 
+    /// <summary>
+    /// Enables the Organization Auto Confirm policy for the specified organization.
+    /// </summary>
+    public static async Task EnableOrganizationAutoConfirmPolicyAsync<T>(
```
eliykat: I think a generic `EnablePolicyAsync(PolicyType policyType)` method would be helpful, otherwise we're going to end up with a helper method for every policy.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerTests.cs
eliykat, line 1: Controller tests are going to get very long; thoughts on making separate files per set of functionality? e.g. autoconfirm endpoints would have their own test file. I did this with unit tests for `OrganizationUserControllerPutTests`.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerTests.cs
```diff
@@ -474,4 +474,117 @@ private async Task VerifyMultipleUsersConfirmedAsync(List<(OrganizationUser orgU
...
+            {
+                Key = testKey,
+                DefaultUserCollectionName = defaultCollectionName
+            });
+
+        Assert.Equal(HttpStatusCode.Forbidden, result.StatusCode);
+    }
+
+    [Fact]
+    public async Task AutoConfirm_WhenOwnerInvitesValidUser_ThenShouldReturnNoContent()
```
eliykat: Test name: we're testing acceptance, not invitation.

This test should also check the db to verify the user has moved into the correct state (that is, they've actually been confirmed).

eliykat (changes requested):
Nice work!

eliykat:
Also... enjoy your merge errors 😎

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
jrmccannon: They should make a linter for this.

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
jrmccannon: For this one, its a `CommandResult` which returns a None.  Do we want to return the none? or just return the Ok/NoContent directly since None implies no return object.

Also, for this one, would the NoContent be more accurate as its a 200 code (success) but gives the additional information of telling the client that there's no body to look for? (I know we don't do this else where but is a good API practice.)

I'll change for consistency if that's better.

src/Core/AdminConsole/Utilities/v2/Results/CommandResult.cs
jrmccannon: I'm unsure of how to do this since the `Task<>` is involved here. I don't  think you can await inside the match.

jrmccannon: I figured it out.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
jrmccannon: Say no more. Policy Requirements all the way!

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
jrmccannon: Made some more methods. Should be all chained now !

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerTests.cs
jrmccannon: Migrated.

src/Api/AdminConsole/Controllers/BaseAdminConsoleController.cs
eliykat: Returning NoContent makes sense to me.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
eliykat: I'll make a ticket to follow this up.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -87,15 +96,28 @@ [new CollectionAccessSelection
...
-    private async Task<CommandResult<AutomaticallyConfirmOrganizationUserRequestData>> PushSyncOrganizationKeysAsync(AutomaticallyConfirmOrganizationUserRequestData request)
+    /// <summary>
+    /// Determines whether a default collection should be created for an organization user during the confirmation process.
+    /// </summary>
+    /// <param name="request">
+    /// The validation request containing information about the user, organization, and collection settings.
+    /// </param>
+    /// <returns>The result is a boolean value indicating whether a default collection should be created.</returns>
+    private async Task<bool> ShouldCreateDefaultCollectionAsync(AutomaticallyConfirmOrganizationUserValidationRequest request) =>
+        featureService.IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
```
eliykat: We should avoid nesting feature flags. We can assume this will be on by the time this feature is released, so I suggest this check is removed.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
eliykat: Unaddressed feedback ([ref](https://github.com/bitwarden/server/pull/6488/files#r2483028476))

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        _ = await validatedRequest.ApplyAsync([
+            CreateDefaultCollectionsAsync,
+            LogOrganizationUserConfirmedEventAsync,
+            SendConfirmedOrganizationUserEmailAsync,
+            DeleteDeviceRegistrationAsync,
+            PushSyncOrganizationKeysAsync
+        ]);
+
+        return new None(); // Operation is idempotent. If this is false, then the user is already confirmed.
```
eliykat: Comment doesn't seem to apply here. Copy/paste error from above maybe?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            return new None(); // Operation is idempotent. If this is false, then the user is already confirmed.
+        }
+
+        _ = await validatedRequest.ApplyAsync([
+            CreateDefaultCollectionsAsync,
+            LogOrganizationUserConfirmedEventAsync,
+            SendConfirmedOrganizationUserEmailAsync,
+            DeleteDeviceRegistrationAsync,
+            PushSyncOrganizationKeysAsync
```
eliykat: I suggest that `ApplyAsync` is overdesigned here. Compared to this, which would be very familiar to anyone:

```c#
        await Task.WhenAll(
            // each method returns a Task because we don't care about the result
            CreateDefaultCollectionsAsync(validatedRequest),
            LogOrganizationUserConfirmedEventAsync(validatedRequest),
            SendConfirmedOrganizationUserEmailAsync(validatedRequest),
            DeleteDeviceRegistrationAsync(validatedRequest),
            PushSyncOrganizationKeysAsync(validatedRequest)
        );
```

I like FP patterns when they can make our code simpler or clearer - I think result objects do that. That in turn suggests some amount of railroad programming might be useful when dealing with lots of chained result objects. But in cases where we can use standard methods to achieve the same thing, that seems clearer.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,227 @@
...
+        ]);
+
+        return new None(); // Operation is idempotent. If this is false, then the user is already confirmed.
+    }
+
+    private async Task<CommandResult<AutomaticallyConfirmOrganizationUserValidationRequest>> CreateDefaultCollectionsAsync(
+        AutomaticallyConfirmOrganizationUserValidationRequest request)
```
eliykat: As I hinted at above, these could just return `Task`s because they are pure side effect methods.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
eliykat: `DeleteDeviceRegistrationAsync` and `PushSyncOrganizationKeysAsync` seem related. Both affect the push notification system. I assume that the device registration may need to be deleted before the push is sent, just based on current usage, although I don't know for sure. Can you please check the dependency between these and move them into the same method (to guarantee order of execution) if there's any doubt.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,227 @@
...
+            return new FailedToDeleteDeviceRegistration();
+        }
+    }
+
+    private async Task<CommandResult<AutomaticallyConfirmOrganizationUserValidationRequest>> RetrieveDataAsync(
+        AutomaticallyConfirmOrganizationUserRequest request)
+    {
+        var organizationUser = await GetOrganizationUserAsync(request);
+
+        if (organizationUser.IsError) return organizationUser.AsError;
```
eliykat: 💭 (non-blocking but we should think about this) In C# land, this is actually less typesafe than using nullable reference types.

The compiler won't stop you from accessing `AsError` or `AsSuccess` without checking it first (in which case you'll get a runtime error if it isn't what you think it is). That's why `Match` is better, but it's not always convenient to use. By comparison, the compiler _will_ make sure you check for null, and with nullable enabled by default and repository methods returning `T?` you get this type safety out of the box.

I suspect this is FP patterns hitting up against imperative code flow. Let me know what you think.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    private async Task<CommandResult<AcceptedOrganizationUser>> GetOrganizationUserAsync(AutomaticallyConfirmOrganizationUserRequest request)
+    {
+        var organizationUser = await organizationUserRepository.GetByIdAsync(request.OrganizationUserId);
+
+        return organizationUser switch
+        {
+            null or { UserId: null } => new UserNotFoundError(),
+            { Status: not OrganizationUserStatusType.Accepted } => new UserIsNotAccepted(),
```
eliykat: This is hiding some validation in the command even though you have a separate validator.

I suggest "user must have a userId" and "user must be in the accepted state" is part of validation logic, not fetching logic.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+        request.OrganizationUser.OrganizationId == request.Organization.Id
+            ? Valid(request)
+            : Invalid(request, new OrganizationUserIdIsInvalid());
+
+    private async Task<ValidationResult<AutomaticallyConfirmOrganizationUserValidationRequest>> OrganizationUserOwnershipValidationAsync(
+        AutomaticallyConfirmOrganizationUserValidationRequest request) =>
+        request.Organization.PlanType is not PlanType.Free
+        || await organizationUserRepository.GetCountByFreeOrganizationAdminUserAsync(request.OrganizationUser.UserId) == 0
+            ? Valid(request)
+            : Invalid(request, new UserToConfirmIsAnAdminOrOwnerOfAnotherFreeOrganization());
```
eliykat: Does not apply because this will never confirm owners or admins, so there's no risk of them becoming owner/admin of more than 1 free org. Also, autoconfirm is enterprise only, so would not confirm them to a free org.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
eliykat, line 1: 👏 I really like the way you structured these policy checks. Very clear to read.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/Errors.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Bit.Core.AdminConsole.Utilities.v2;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.AutoConfirmUser;
+
+public record OrganizationNotFound() : NotFoundError("Invalid organization");
+public record FailedToWriteToEventLog() : InternalError("Failed to write to event log");
+public record FailedToSendConfirmedUserEmail() : InternalError("Failed to send confirmed user email");
+public record FailedToDeleteDeviceRegistration() : InternalError("Failed to delete device registration");
+public record FailedToPushOrganizationSyncKeys() : InternalError("Failed to push organization sync keys");
+public record UserIsNotUserType() : BadRequestError("Only organization users with the User can be automatically confirmed");
```
eliykat: Missing a

```suggestion
public record UserIsNotUserType() : BadRequestError("Only organization users with the User role can be automatically confirmed");
```

src/Core/AdminConsole/Models/Data/OrganizationUsers/AcceptedOrganizationUser.cs
eliykat, line 1: I don't think this should subclass the data model, it's too different. e.g.

- doesn't need a Status, because it only represents an accepted user
- doesn't need an Email, because you join through the UserId to get their email
- UserId is never null (although you did handle this with `new`)
- Key is always null (they get the key once they're confirmed, so that would be on the Confirmed user model)

Check the xmldoc on `OrganizationUser.cs` for some of this business logic.

However, then you do end up with a bunch of mapping code to/from `OrganizationUser` and even between different user state models as we move an orgUser through different states. I don't have a good solution for that now, maybe that's part of the cost of this pattern.

We don't need to tackle it now if it makes this PR too large, it's up to you.

src/Core/AdminConsole/Utilities/v2/Results/CommandResult.cs
eliykat, line 1: Please move the CommandResult extension methods to a separate file so that they're not overwhelming the basic result definition.

(non-blocking) My preference is also to add these as we need them so we have a clear use case. But if you want to hold onto them, putting them in a separate file is also fine.

src/Core/AdminConsole/Utilities/v2/Errors.cs
```diff
@@ -0,0 +1,43 @@
...
+        return (errors.All(e => e is InternalError),
+                errors.All(e => e is NotFoundError),
+                errors.All(e => e is BadRequestError)) switch
+        {
+            (true, _, _) => new AggregateInternalError(messages),
+            (_, true, _) => new AggregateNotFoundError(messages),
+            (_, _, true) => new AggregateBadRequestError(messages),
+            _ => new AggregateError(messages)
+        };
+    }
```
eliykat: I don't think we should include this yet without a clear use case. We almost never want to concatenate arbitrary errors; we either return a single error or an `IEnumerable<OrganizationUserBulkResponseModel>` with properly structured error data. Given that our errors are generally displayed in toasts in the UI this is unlikely to be displayed properly.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
eliykat, line 1: This file has an odd layout, it seems reversed?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
eliykat: I see you have a separate PR for this, all good.

src/Core/AdminConsole/Utilities/v2/Errors.cs
eliykat: Also because this PR is now very large, so if we can remove some code we don't need right now that will make it more manageable to get merged.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
jrmccannon: My personal preference was to grab the time when the request comes in from the Controller. That pattern also allowed for the entire request to share the same date for saving database records. 

I'll move this down to be a bit more consistent

test/Api.IntegrationTest/Helpers/OrganizationTestHelpers.cs
jrmccannon: This is unused. Removing

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUserCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+                {
+                    return new None();
+                }
+
+                await Task.WhenAll(
+                    CreateDefaultCollectionsAsync(validatedData.Request),
+                    LogOrganizationUserConfirmedEventAsync(validatedData.Request),
+                    SendConfirmedOrganizationUserEmailAsync(validatedData.Request),
+                    SyncOrganizationKeysAsync(validatedData.Request)
+                );
```
eliykat, line 60: Something I just thought of - you want to make sure that the default collection is created before you push a full sync to the client, otherwise they won't pull it down.

To be safe:


```suggestion
                // Create their default collection first to make sure it's there before we push a sync
                await CreateDefaultCollectionsAsync(validatedData.Request);

                await Task.WhenAll(
                    LogOrganizationUserConfirmedEventAsync(validatedData.Request),
                    SendConfirmedOrganizationUserEmailAsync(validatedData.Request),
                    SyncOrganizationKeysAsync(validatedData.Request)
                );
```

I do otherwise think it's fine to parallelize these operations, and something we could benefit from doing more often.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
eliykat, line 1: This came out really well and is really easy to read, great job iterating on it 👏

src/Core/AdminConsole/Models/Data/OrganizationUsers/AcceptedOrganizationUserToConfirm.cs
eliykat, line 1: Also a good solution to this problem 👍

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
```diff
@@ -0,0 +1,174 @@
...
+                .IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
+                .Returns(true);
+        });
+        _client = _factory.CreateClient();
+        _loginHelper = new LoginHelper(_factory, _client);
+    }
+
+    public async Task InitializeAsync()
+    {
+        _ownerEmail = $"org-user-integration-test-{Guid.NewGuid()}@bitwarden.com";
```
eliykat: Claude picked this up on me the other day, we should use `@example.com` for all our tests.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
```diff
@@ -0,0 +1,174 @@
...
+    private string _ownerEmail = null!;
+
+    public OrganizationUserControllerAutoConfirmTests(ApiApplicationFactory apiFactory)
+    {
+        _factory = apiFactory;
+        _factory.SubstituteService<IFeatureService>(featureService =>
+        {
+            featureService
+                .IsEnabled(FeatureFlagKeys.CreateDefaultLocation)
+                .Returns(true);
```
eliykat, line 36: Shouldn't be required anymore.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            featureService
+                .IsEnabled(FeatureFlagKeys.AutomaticConfirmUsers)
+                .Returns(true);
+        });
+        var localClient = localFactory.CreateClient();
+        var localLoginHelper = new LoginHelper(localFactory, localClient);
+
+        return (localFactory, localClient, localLoginHelper);
+    }
```
eliykat: Why is this required in addition to client and factory set up in InitializeAsync? If it's only to set the feature flag differently, I suggest we can rely on the declarative `RequireFeature` attribute without testing it every time we use it, and that will simplify your test setup requirements. (In other words - just delete the test case with the flag off.)

Note that for tests that use this private method, they are bootstrapping 2 api clients: `_client` and `localClient`, which is not ideal.

If we do need it - you also need to make sure you dispose of this client. So perhaps it is better moved into the initialize/dispose methods to make sure its lifecycle is properly handled. That exacerbates the "bootstrapping 2 separate clients" problem though, which is why it's maybe easier just to get rid of it.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
eliykat, line 1: Additional validation I just thought of:
- can the organization use autoconfirm? (use the orgAbility cache)
- is the autoconfirm policy enabled?

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
eliykat, line 1: Additional api integration test suggestions:
- confirming that the default collection is created
- given that concurrency is a concern here, it would be good to have a test that fires off say 10 requests simultaneously and then checks that only 1 was processed. Not sure how to verify this though - maybe check that only 1 default collection was created? MailService might also be a mock that you could use in your assertions.

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUsers/AutomaticallyConfirmOrganizationUsersValidatorTests.cs
```diff
@@ -0,0 +1,681 @@
...
+        var result = await sutProvider.Sut.ValidateAsync(request);
+
+        // Assert
+        Assert.True(result.IsError);
+        Assert.IsType<OrganizationUserIdIsInvalid>(result.AsError);
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task ValidateAsync_WithInvitedStatus_ReturnsUserIsNotAcceptedError(
```
eliykat: Tests that check the user's status can be parameterized, e.g.

```
[BitAutoData(OrganizationUserStatusType.Invited)]
[BitAutoData(OrganizationUserStatusType.Confirmed)]
[BitAutoData(OrganizationUserStatusType.Revoked)]
```

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUsers/AutomaticallyConfirmOrganizationUsersValidatorTests.cs
```diff
@@ -0,0 +1,681 @@
...
+        var result = await sutProvider.Sut.ValidateAsync(request);
+
+        // Assert
+        Assert.True(result.IsError);
+        Assert.IsType<UserIsNotAccepted>(result.AsError);
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task ValidateAsync_WithAdminType_ReturnsUserIsNotUserTypeError(
```
eliykat: Tests that check the user's role can also be parameterized.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
r-tome: +1! I appreciate the comments

test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUsers/AutomaticallyConfirmUsersCommandTests.cs
```diff
@@ -0,0 +1,730 @@
...
+            .Received(1)
+            .ConfirmOrganizationUserAsync(Arg.Is<AcceptedOrganizationUserToConfirm>(o =>
+                o.OrganizationUserId == organizationUser.Id && o.Key == request.Key));
+
+        await AssertSuccessfulOperationsAsync(sutProvider, organizationUser, organization, user, key);
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task AutomaticallyConfirmOrganizationUserAsync_WithValidationError_ReturnsValidationError(
```
r-tome:
```suggestion
    public async Task AutomaticallyConfirmOrganizationUserAsync_WithInvalidUserOrgId_ReturnsOrganizationUserIdIsInvalidError(
```

r-tome (commented):
Great work!

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
jrmccannon: Added both tests. Used the default collection creation as way to validate only 1 request got past the idempotent check.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -62,6 +69,11 @@ public async Task<ValidationResult<AutomaticallyConfirmOrganizationUserValidatio
         return Valid(request);
     }
 
+    private async Task<bool> OrganizationHasAutomaticallyConfirmUsersPolicyEnabledAsync(
+        AutomaticallyConfirmOrganizationUserValidationRequest request) =>
+        await policyRepository.GetByOrganizationIdTypeAsync(request.OrganizationId,
+            PolicyType.AutomaticUserConfirmation) is { Enabled: true };
```
eliykat: Should also check `orgAbility.UsePolicies && orgAbility.UseAutomaticUserConfirmation`, to ensure that the organization's current plan supports this functionality. (To protect against plan downgrades.)

[Example usage.](https://github.com/bitwarden/server/blob/e11458196c7f649091f2e6c896a9b9bca9c8e856/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs#L325-L330)

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUserControllerAutoConfirmTests.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        await _loginHelper.LoginAsync(userToConfirmEmail);
+        var organizationUser = await OrganizationTestHelpers.CreateUserAsync(
+            _factory,
+            organization.Id,
+            userToConfirmEmail,
+            OrganizationUserType.User,
+            false,
+            new Permissions(),
+            OrganizationUserStatusType.Accepted);
```
eliykat: Setup is a bit confused here:
- `userToConfirm` appears to be the user who is _doing_ the confirmation, not the user being confirmed
- multiple calls to login methods are possibly redundant (e.g. you login as the owner but then immediately login as the user; you use both `_factory.LoginWithNewAccount` and `_loginHelper.LoginAsync`).

Other tests in this file have similar issues.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
jrmccannon: I shouldn't need to use the cache right? I can just use the Org record I retrieved eariler correct?

eliykat: Yes you're right.

eliykat (approved):

PR #6519, opened 2025-10-30
[PM-26967] Added new metric properties

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26967

## 📔 Objective

Add  properties for metrics within the Repo
Allow the controller to accept metric values and update the database

## 📸 Screenshots

None

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_UpdateMetrics.sql
+++ b/src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_UpdateMetrics.sql
@@ -17,22 +17,23 @@ AS
 BEGIN
     SET NOCOUNT ON;
 
-    UPDATE dbo.OrganizationReport 
+    UPDATE 
+        [dbo].[OrganizationReport]
     SET
-        ApplicationCount = @ApplicationCount,
-        ApplicationAtRiskCount = @ApplicationAtRiskCount,
-        CriticalApplicationCount = @CriticalApplicationCount,
-        CriticalApplicationAtRiskCount = @CriticalApplicationAtRiskCount,
-        MemberCount = @MemberCount,
-        MemberAtRiskCount = @MemberAtRiskCount,
-        CriticalMemberCount = @CriticalMemberCount,
-        CriticalMemberAtRiskCount = @CriticalMemberAtRiskCount,
-        PasswordCount = @PasswordCount,
-        PasswordAtRiskCount = @PasswordAtRiskCount,
-        CriticalPasswordCount = @CriticalPasswordCount,
-        CriticalPasswordAtRiskCount = @CriticalPasswordAtRiskCount,
-        RevisionDate = @RevisionDate
+        [ApplicationCount] = @ApplicationCount,
+        [ApplicationAtRiskCount] = @ApplicationAtRiskCount,
+        [CriticalApplicationCount] = @CriticalApplicationCount,
+        [CriticalApplicationAtRiskCount] = @CriticalApplicationAtRiskCount,
+        [MemberCount] = @MemberCount,
+        [MemberAtRiskCount] = @MemberAtRiskCount,
+        [CriticalMemberCount] = @CriticalMemberCount,
+        [CriticalMemberAtRiskCount] = @CriticalMemberAtRiskCount,
+        [PasswordCount] = @PasswordCount,
+        [PasswordAtRiskCount] = @PasswordAtRiskCount,
+        [CriticalPasswordCount] = @CriticalPasswordCount,
+        [CriticalPasswordAtRiskCount] = @CriticalPasswordAtRiskCount,
+        [RevisionDate] = @RevisionDate
     WHERE 
-        Id = @Id
-        
+        [Id] = @Id
+
 END
--- a/util/Migrator/DbScripts/2025-10-30_00_OrganizationReport_UpdateMetrics.sql
+++ b/util/Migrator/DbScripts/2025-10-30_00_OrganizationReport_UpdateMetrics.sql
@@ -17,22 +17,23 @@ AS
 BEGIN
     SET NOCOUNT ON;
 
-    UPDATE dbo.OrganizationReport 
+    UPDATE 
+        [dbo].[OrganizationReport]
     SET
-        ApplicationCount = @ApplicationCount,
-        ApplicationAtRiskCount = @ApplicationAtRiskCount,
-        CriticalApplicationCount = @CriticalApplicationCount,
-        CriticalApplicationAtRiskCount = @CriticalApplicationAtRiskCount,
-        MemberCount = @MemberCount,
-        MemberAtRiskCount = @MemberAtRiskCount,
-        CriticalMemberCount = @CriticalMemberCount,
-        CriticalMemberAtRiskCount = @CriticalMemberAtRiskCount,
-        PasswordCount = @PasswordCount,
-        PasswordAtRiskCount = @PasswordAtRiskCount,
-        CriticalPasswordCount = @CriticalPasswordCount,
-        CriticalPasswordAtRiskCount = @CriticalPasswordAtRiskCount,
-        RevisionDate = @RevisionDate
+        [ApplicationCount] = @ApplicationCount,
+        [ApplicationAtRiskCount] = @ApplicationAtRiskCount,
+        [CriticalApplicationCount] = @CriticalApplicationCount,
+        [CriticalApplicationAtRiskCount] = @CriticalApplicationAtRiskCount,
+        [MemberCount] = @MemberCount,
+        [MemberAtRiskCount] = @MemberAtRiskCount,
+        [CriticalMemberCount] = @CriticalMemberCount,
+        [CriticalMemberAtRiskCount] = @CriticalMemberAtRiskCount,
+        [PasswordCount] = @PasswordCount,
+        [PasswordAtRiskCount] = @PasswordAtRiskCount,
+        [CriticalPasswordCount] = @CriticalPasswordCount,
+        [CriticalPasswordAtRiskCount] = @CriticalPasswordAtRiskCount,
+        [RevisionDate] = @RevisionDate
     WHERE 
-        Id = @Id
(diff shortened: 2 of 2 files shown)


src/Api/Dirt/Models/Response/OrganizationReportResponseModel.cs
```diff
@@ -0,0 +1,38 @@
...
+    public Guid OrganizationId { get; set; }
+    public string ReportData { get; set; } = string.Empty;
+    public string ContentEncryptionKey { get; set; } = string.Empty;
+    public string SummaryData { get; set; } = string.Empty;
+    public string ApplicationData { get; set; } = string.Empty;
+    public int PasswordCount { get; set; } = 0;
+    public int PasswordAtRiskCount { get; set; } = 0;
+    public int MemberCount { get; set; } = 0;
+    public DateTime? CreationDate { get; set; } = null;
+    public DateTime? RevisionDate { get; set; } = null;
```
withinfocus: 🎨 I wouldn't want defaults on these, and null is not empty (use null). Those strings are nullable too, as are the ints -- why not represent the true default? I wouldn't want code later down the line to see a zero and indicate that's a positive when it's just undefined for example.

withinfocus: Take a look at all the nullable usage across the PR too.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_UpdateMetrics.sql
```diff
@@ -0,0 +1,39 @@
...
+        CriticalMemberCount = @CriticalMemberCount,
+        CriticalMemberAtRiskCount = @CriticalMemberAtRiskCount,
+        PasswordCount = @PasswordCount,
+        PasswordAtRiskCount = @PasswordAtRiskCount,
+        CriticalPasswordCount = @CriticalPasswordCount,
+        CriticalPasswordAtRiskCount = @CriticalPasswordAtRiskCount,
+        RevisionDate = @RevisionDate
+    WHERE 
+        Id = @Id
+        AND OrganizationId = @OrganizationId
```
withinfocus: ❌ If you have an ID for this exact row I don't see any point in the organization ID being used. I left similar comments in the past -- auth is out of scope for this and the rights to change the row should already be established.

src/Api/Dirt/Models/Response/OrganizationReportResponseModel.cs
voommen-livefront: updated the usage of nulls. Only using string.Empty where it is required in the code. You are correct - when a request is made, most values may be null.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_UpdateMetrics.sql
voommen-livefront: Thanks for catching this!

src/Api/Dirt/Models/Response/OrganizationReportResponseModel.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        Id = organizationReport.Id;
+        OrganizationId = organizationReport.OrganizationId;
+        ReportData = organizationReport.ReportData;
+        ContentEncryptionKey = organizationReport.ContentEncryptionKey;
+        SummaryData = organizationReport.SummaryData;
+        ApplicationData = organizationReport.ApplicationData;
+        PasswordCount = organizationReport.PasswordCount ?? 0;
+        PasswordAtRiskCount = organizationReport.PasswordAtRiskCount ?? 0;
+        MemberCount = organizationReport.MemberCount ?? 0;
```
withinfocus: ⚠️ It still doesn't seem right to default these to me.

src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
```diff
@@ -1,16 +1,15 @@
...
+    public string? ContentEncryptionKey { get; set; }
 
-    public string ContentEncryptionKey { get; set; }
+    public string? SummaryData { get; set; }
 
-    public string SummaryData { get; set; }
+    public string? ApplicationData { get; set; }
 
-    public string ApplicationData { get; set; }
+    public OrganizationReportMetricsRequest Metrics { get; set; } = new OrganizationReportMetricsRequest();
```
withinfocus: ⚠️ Also should not be set.

src/Core/Dirt/Models/Data/OrganizationReportMetricsData.cs
```diff
@@ -0,0 +1,40 @@
...
+    public int? CriticalApplicationCount { get; set; } = null;
+    public int? CriticalApplicationAtRiskCount { get; set; } = null;
+    public int? MemberCount { get; set; } = null;
+    public int? MemberAtRiskCount { get; set; } = null;
+    public int? CriticalMemberCount { get; set; } = null;
+    public int? CriticalMemberAtRiskCount { get; set; } = null;
+    public int? PasswordCount { get; set; } = null;
+    public int? PasswordAtRiskCount { get; set; } = null;
+    public int? CriticalPasswordCount { get; set; } = null;
+    public int? CriticalPasswordAtRiskCount { get; set; } = null;
```
withinfocus: ⚠️ Also should not be set.

withinfocus (commented):
The DB changes are fine now but the not-owned-by-me changes still have concerns.

src/Api/Dirt/Models/Response/OrganizationReportResponseModel.cs
voommen-livefront: I removed the defaults - this time even more carefully reading through the PR - and I think I caught everything, excetpt those that were required by the ITableObject

src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
voommen-livefront: here too - I have removed it as much as possible, except when I absolutely needed it.

src/Core/Dirt/Models/Data/OrganizationReportMetricsData.cs
voommen-livefront: Thanks for catching this.

src/Api/Dirt/Models/Response/OrganizationReportResponseModel.cs
prograhamming: I agree that these should be changed. I would make `PasswordCount`, `PasswordAtRiskCount`,  and `MemberCount` to have type `int?` and remove the null-coalescing. Defaulting this to zero creates ambiguity: does  0 mean "we calculated and found zero passwords" or does it mean "we haven't calculated this yet?"

src/Core/Dirt/Models/Data/OrganizationReportMetricsData.cs
prograhamming: Let's remove setting null. By default in C# value types (e.g. int, bool, etc.) cannot be null. By using the `?` operator we are making this a nullable type which allows it to hold null in addition to its normal values. It is really syntactic sugar for `Nullable<int>` which wraps the value type in a struct that tracks whether a value exists.

src/Core/Dirt/Reports/ReportFeatures/Requests/AddOrganizationReportRequest.cs
prograhamming: I would make the `OrganizationReportMetricRequest` nullable and remove it being set.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_UpdateMetrics.sql
```diff
@@ -0,0 +1,38 @@
...
+    @PasswordCount INT,
+    @PasswordAtRiskCount INT,
+    @CriticalPasswordCount INT,
+    @CriticalPasswordAtRiskCount INT,
+    @RevisionDate DATETIME2(7)
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    UPDATE dbo.OrganizationReport
```
mkincaid-bw: ⛏️  Per the [Code Style docs](https://contributing.bitwarden.com/contributing/code-style/sql#update-statements), `UPDATE` and the table name should be on separate lines. 

Also, all object names should include square brackets, ie:

```sql
UPDATE 
    [dbo].[OrganizationReport]
SET
    [ApplicationCount] = @ApplicationCount,
    [ApplicationAtRiskCount] = @ApplicationAtRiskCount,
...
WHERE 
    [Id] = @Id
```

util/Migrator/DbScripts/2025-10-30_00_OrganizationReport_UpdateMetrics.sql
```diff
@@ -0,0 +1,38 @@
...
+    @PasswordCount INT,
+    @PasswordAtRiskCount INT,
+    @CriticalPasswordCount INT,
+    @CriticalPasswordAtRiskCount INT,
+    @RevisionDate DATETIME2(7)
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    UPDATE dbo.OrganizationReport
```
mkincaid-bw: Same as above.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_UpdateMetrics.sql
voommen-livefront: updated formatting. Sorry I missed this earlier.

mkincaid-bw (approved):
DB Changes LGTM

prograhamming (approved):

PR #6524, opened 2025-10-30
[PFI-1019] Add groups to the "list all collections" api responses

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PFI-1019

## 📔 Objective

Group information was missing on collection object returned from `GET /public/collections` API. Added it in using data on the `collectionRepository.GetManyByOrganizationIdWithAccessAsync()` query.

### DIFF
--- a/src/Api/Public/Controllers/CollectionsController.cs
+++ b/src/Api/Public/Controllers/CollectionsController.cs
@@ -65,10 +65,10 @@ public class CollectionsController : Controller
     [ProducesResponseType(typeof(ListResponseModel<CollectionResponseModel>), (int)HttpStatusCode.OK)]
     public async Task<IActionResult> List()
     {
-        var collections = await _collectionRepository.GetManySharedCollectionsByOrganizationIdAsync(
+        var collections = await _collectionRepository.GetManyByOrganizationIdWithAccessAsync(
             _currentContext.OrganizationId.Value);
-        // TODO: Get all CollectionGroup associations for the organization and marry them up here for the response.
-        var collectionResponses = collections.Select(c => new CollectionResponseModel(c, null));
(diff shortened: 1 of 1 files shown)


src/Api/Public/Controllers/CollectionsController.cs
```diff
@@ -65,10 +65,10 @@ public async Task<IActionResult> Get(Guid id)
     [ProducesResponseType(typeof(ListResponseModel<CollectionResponseModel>), (int)HttpStatusCode.OK)]
     public async Task<IActionResult> List()
     {
-        var collections = await _collectionRepository.GetManySharedCollectionsByOrganizationIdAsync(
+        var collections = await _collectionRepository.GetManyByOrganizationIdWithAccessAsync(
             _currentContext.OrganizationId.Value);
-        // TODO: Get all CollectionGroup associations for the organization and marry them up here for the response.
-        var collectionResponses = collections.Select(c => new CollectionResponseModel(c, null));
+        // TODO: Do we need to filter out DefaultUserCollection types here like we do in Get above?
```
kspearrin, line 70: @r-tome Shouldn't we do this here too? I see you added it to the GET by id above in https://github.com/bitwarden/server/pull/6021. Not sure why it wasn't added here too.

r-tome: `_collectionRepository.GetManySharedCollectionsByOrganizationIdAsync` already filters out `DefaultUserCollection` types. 
You're right that if we use `GetManyByOrganizationIdWithAccessAsync` then we need to filter it out here

kspearrin:
Fixed with https://github.com/bitwarden/server/pull/6713

PR #6528, opened 2025-10-31
Add template properties for Datadog

brant-livefront:
## 📔 Objective

We recently got a few more specifics from the Datadog team on the existing pull-based integration. In order to support identical data being sent as part of our new push-based integration, we need a few more dynamic fields to enrich the data.

This PR adds these new dynamic fields.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/AdminConsole/Models/Response/Organizations/PolicyResponseModel.cs
+++ b/src/Api/AdminConsole/Models/Response/Organizations/PolicyResponseModel.cs
@@ -30,6 +30,7 @@ public class PolicyResponseModel : ResponseModel
         {
             Data = JsonSerializer.Deserialize<Dictionary<string, object>>(policy.Data);
         }
+        RevisionDate = policy.RevisionDate;
     }
 
     public Guid Id { get; set; }
@@ -37,4 +38,5 @@ public class PolicyResponseModel : ResponseModel
     public PolicyType Type { get; set; }
     public Dictionary<string, object> Data { get; set; }
     public bool Enabled { get; set; }
+    public DateTime RevisionDate { get; set; }
 }
--- a/src/Core/AdminConsole/Enums/PolicyType.cs
+++ b/src/Core/AdminConsole/Enums/PolicyType.cs
@@ -21,6 +21,7 @@ public enum PolicyType : byte
     UriMatchDefaults = 16,
     AutotypeDefaultSetting = 17,
     AutomaticUserConfirmation = 18,
+    BlockClaimedDomainAccountCreation = 19,
 }
 
 public static class PolicyTypeExtensions
@@ -52,6 +53,7 @@ public static class PolicyTypeExtensions
             PolicyType.UriMatchDefaults => "URI match defaults",
             PolicyType.AutotypeDefaultSetting => "Autotype default setting",
             PolicyType.AutomaticUserConfirmation => "Automatically confirm invited users",
+            PolicyType.BlockClaimedDomainAccountCreation => "Block account creation for claimed domains",
         };
     }
 }
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyServiceCollectionExtensions.cs
@@ -53,6 +53,7 @@ public static class PolicyServiceCollectionExtensions
         services.AddScoped<IPolicyUpdateEvent, FreeFamiliesForEnterprisePolicyValidator>();
         services.AddScoped<IPolicyUpdateEvent, OrganizationDataOwnershipPolicyValidator>();
         services.AddScoped<IPolicyUpdateEvent, UriMatchDefaultPolicyValidator>();
+        services.AddScoped<IPolicyUpdateEvent, BlockClaimedDomainAccountCreationPolicyValidator>();
         services.AddScoped<IPolicyUpdateEvent, AutomaticUserConfirmationPolicyEventHandler>();
     }
 
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/BlockClaimedDomainAccountCreationPolicyValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/BlockClaimedDomainAccountCreationPolicyValidator.cs
@@ -1,0 +1,59 @@
+﻿#nullable enable
+
+using Bit.Core.AdminConsole.Entities;
+using Bit.Core.AdminConsole.Enums;
+using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationDomains.Interfaces;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Models;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyUpdateEvents.Interfaces;
+using Bit.Core.Services;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
+
+public class BlockClaimedDomainAccountCreationPolicyValidator : IPolicyValidator, IPolicyValidationEvent
+{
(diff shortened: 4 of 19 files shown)


eliykat (commented):
I assume my review is not required because this is merging into a feature branch (so codeowner requirements don't apply) and Architecture Team are handling this work. Please re-request my review if this is not the case.

src/Sql/dbo/Stored Procedures/OrganizationUserUserDetails_ReadByOrganizationIdUserId.sql
```diff
@@ -0,0 +1,18 @@
+CREATE PROCEDURE [dbo].[OrganizationUserUserDetails_ReadByOrganizationIdUserId]
+    @OrganizationId UNIQUEIDENTIFIER,
+    @UserId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+SELECT
+    TOP 1 *
```
mkincaid-bw: ❓ Curious the purpose of the `TOP 1` here? With no `ORDER BY` clause, this query is non-deterministic and can return inconsistent results any time it's run.

brant-livefront: I think I was being a bit belt + suspenders. There's really no way this query should return more than one. I can remove it so it doesn't cause confusion.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -962,4 +962,20 @@ public async Task<bool> ConfirmOrganizationUserAsync(Core.Entities.OrganizationU
...
+    public async Task<OrganizationUserUserDetails?> GetDetailsByOrganizationIdUserIdAsync(Guid organizationId, Guid userId)
+    {
+        using (var scope = ServiceScopeFactory.CreateScope())
+        {
+            var dbContext = GetDatabaseContext(scope);
+            var view = new OrganizationUserUserDetailsViewQuery();
+            var entity = await view.Run(dbContext).FirstOrDefaultAsync(ou => ou.OrganizationId == organizationId && ou.UserId == userId);
+            return entity;
+        }
+    }
```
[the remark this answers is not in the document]
brant-livefront: Unfortunately, I need nullable turned on to match the declaration in the interface and dapper (i.e. with the optional `?`). Also, I think it's probably best to support nullable where we can now so that the path is easier when other methods are migrated to support nullable.

src/Sql/dbo/Stored Procedures/OrganizationUserUserDetails_ReadByOrganizationIdUserId.sql
mkincaid-bw: @brant-livefront FWIW, I checked the database and there is one (and only one) instance with multiple records with the same UserId and OrgId. The data in both rows are identical (except for the Id), so that leads me to believe that the duplicate row was entered inadvertently. If you leave the `TOP 1` in there, it should be fine assuming you don't care about the Id for the OrganizationUser table, since in this case they will be different. Every other User/Org combination has only one row (currently). I was initially concerned about performance but compared execution plans and that doesn't appear to be an issue so it's fine if you want to leave it in. Just be aware that it is currently possible that more than one row can be returned for a given User/Org, so you'd either need the `TOP 1` or filter it out in code.

brant-livefront: @mkincaid-bw Thanks for checking on this! I think the best course is to put the `TOP 1` back in - we don't care about order (or which id that one row returns) because we're only looking up things like Name, Email and Type here.

Our code currently would fail if it returned more than 1, so given that there's a possible edge case, the `TOP 1` is probably a good defense against that rather than code around a case that really shouldn't be there. I'll add it back in.

util/Migrator/DbScripts/2025-11-05_00_OrganizationUserUserDetails_ReadByOrganizationIdUserId.sql
```diff
@@ -0,0 +1,17 @@
...
+SELECT
+    TOP 1 *
+FROM
+    [dbo].[OrganizationUserUserDetailsView]
+WHERE
+    [OrganizationId] = @OrganizationId
+  AND
+    [UserId] = @UserId
+END
+go
```
withinfocus:
```suggestion
GO
```

⛏️ ... the literal last line of the PR! 😜

src/Sql/dbo/Stored Procedures/OrganizationUserUserDetails_ReadByOrganizationIdUserId.sql
withinfocus: Mark and team are discussing the strange backstory with this table, so this may need to stay open for a bit as they research why there's not a unique index.

withinfocus (approved):
Conditionally. Wait for DBOps.

src/Sql/dbo/Stored Procedures/OrganizationUserUserDetails_ReadByOrganizationIdUserId.sql
brant-livefront: I discussed this with Matt in Slack and I moved back (in latest commit) to dropping the `TOP 1` and assuming uniqueness. We won't merge until that's a safe assumption.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
```diff
@@ -89,19 +91,35 @@ private async Task<IntegrationTemplateContext> BuildContextAsync(EventMessage ev
...
         }
 
         if (IntegrationTemplateProcessor.TemplateRequiresActingUser(template) && eventMessage.ActingUserId.HasValue)
         {
-            context.ActingUser = await userRepository.GetByIdAsync(eventMessage.ActingUserId.Value);
+            context.ActingUser = await organizationUserRepository.GetDetailsByOrganizationIdUserIdAsync(
+                organizationId: organizationId,
+                userId: eventMessage.ActingUserId.Value
+            );
         }
```
eliykat, line 118: (non-blocking) I don't have full context here, but is it worth checking that the `UserId` and `ActingUserId` are different, so you're not making 2 db calls for the same user?

eliykat (approved):

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
brant-livefront: Good question. The only reason we didn't optimize that here is because we are [adding caching in this PR](https://github.com/bitwarden/server/pull/6608). So that would mean if it is the exact same user, then the second call would definitely be hitting cache.

src/Sql/dbo/Stored Procedures/OrganizationUserUserDetails_ReadByOrganizationIdUserId.sql
mkincaid-bw: There was only one instance of duplicate data from 2017. Robert did some analysis on code changes and found that this occurred before some additional checks were put in place to handle the unique check.

I just cleaned up the duplicate data so there are currently no dups. After discussion, we are opting to NOT add a unique index at this time, primarily because we can't account for possible duplicates in self-hosted clients. We feel this is pretty safe since there was only one instance of duplicate data in 8 years, before additional checks in code were added.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
eliykat: Makes sense! ty

PR #6537, opened 2025-11-04
[PM-27766] Add policy for blocking account creation from claimed domains.

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27766

## 📔 Objective

Enterprise organizations should have a new policy that they can enable to block users from creating individual accounts outside of their organization. We use claimed domains to enforce this ownership.

## 📸 Screenshots (video demo)

https://github.com/user-attachments/assets/9ef8e493-fdbb-4b66-b811-dd064e00b51c

### DIFF
--- a/src/Core/Auth/UserFeatures/Registration/Implementations/RegisterUserCommand.cs
+++ b/src/Core/Auth/UserFeatures/Registration/Implementations/RegisterUserCommand.cs
@@ -111,9 +111,12 @@ public class RegisterUserCommand : IRegisterUserCommand
         string orgInviteToken, Guid? orgUserId)
     {
         TryValidateOrgInviteToken(orgInviteToken, orgUserId, user);
-        var orgUser = await SetUserEmail2FaIfOrgPolicyEnabledAsync(orgUserId, user) ??
+        var orgUser = await SetUserEmail2FaIfOrgPolicyEnabledAsync(orgUserId, user);
+        if (orgUser == null && orgUserId.HasValue)
+        {
             throw new BadRequestException("Invalid organization user invitation.");
-        await ValidateEmailDomainNotBlockedAsync(user.Email, orgUser.OrganizationId);
+        }
+        await ValidateEmailDomainNotBlockedAsync(user.Email, orgUser?.OrganizationId);
 
         user.ApiKey = CoreHelpers.SecureRandomString(30);
 


util/Migrator/DbScripts/2025-11-04_00_BlockClaimedDomainAccountCreationPolicy.sql
```diff
@@ -0,0 +1,45 @@
+-- Add stored procedure for checking if a domain has the BlockClaimedDomainAccountCreation policy enabled
+-- This supports the BlockClaimedDomainAccountCreation policy (Type = 19) which prevents users from
+-- creating personal accounts using email addresses from domains claimed by organizations.
+-- The optional @ExcludeOrganizationId parameter allows excluding a specific organization from the check,
+-- enabling users to join the organization that owns their email domain.
+
+IF OBJECT_ID('[dbo].[OrganizationDomain_HasVerifiedDomainWithBlockPolicy]') IS NOT NULL
```
mkincaid-bw: ⛏️  Using `CREATE OR ALTER` instead of `DROP/CREATE` is a safer approach.
https://contributing.bitwarden.com/contributing/code-style/sql#functions-and-stored-procedures

jrmccannon (changes requested):
Left a comment.

kspearrin:
> Left a comment.

I didn't see any comment from you @jrmccannon

src/Core/Auth/UserFeatures/Registration/Implementations/RegisterUserCommand.cs
```diff
@@ -357,4 +387,26 @@ private RegistrationEmailVerificationTokenable ValidateRegistrationEmailVerifica
...
+        var isDomainBlocked = await _organizationDomainRepository.HasVerifiedDomainWithBlockClaimedDomainPolicyAsync(
+            emailDomain, excludeOrganizationId);
+        if (isDomainBlocked)
+        {
+            _logger.LogInformation(
+                "User registration blocked by domain claim policy. Domain: {Domain}, ExcludedOrgId: {ExcludedOrgId}",
+                emailDomain,
+                excludeOrganizationId);
+            throw new BadRequestException("This email address is claimed by an organization using Bitwarden.");
+        }
```
jrmccannon, line 429: This code does very similar things to the code in `SendVerificationEmailForRegistrationCommand`. 

It should probably be extracted out into a `Query` class with an overload methods. It could simply return a `bool` as a result. 

```c#
public interface IHasVerifiedDomainPolicyQuery {
        public Task<bool> IsDomainClaimedByOrganizationPolicy(string email, Guid excludeOrganization);
        public Task<bool> IsDomainClaimedByOrganizationPolicy(string email);
```

That way all email domain, logging, repo calls, other logic are encapsulated all in one spot.

jrmccannon: @kspearrin there it goes. the new github review experience is nicer...but a little inconsistent :(

jrmccannon: This logic also might be a good candidate for a Policy Requirement.

Enabling the policy could go get Verified domains to capture the domain and store it into the PolicyData field?

Adding a new domain could also subsequently update the policy data field. This would allow us to do the string comparison in memory on the server rather than doing it in SQL server. I know that's been an issue in the past.

@eliykat might have another opinion on this as well.

kspearrin: Do you have an example of the query class idea? I piped it into Claude and it came up with a solution, but it feels pretty weird.

Copying the claimed domain data to the policy feels odd too. I don't think there are significant performance concerns with the SQL query being added here, given that the registration endpoints also do not have high throughput.

jrmccannon: I was thinking something like [this](https://github.com/bitwarden/server/blob/main/src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs) for the query.  

By adding the domains to the policy directly, it would tie the two features together and give a convenient location to look up the Domains claimed and reserved by the organization.  

With the `PolicyRequirement`, it allows for us to pull policy enforcement logic into a common abstraction for all policies.  This then enables easier modifications to the policy logic when it changes later on.

src/Core/Constants.cs
eliykat, line 1: Who is managing the lifecycle of this feature flag? (enabling it and cleaning it up later)

If AC Team will be handling it then I'll let Vince know and we can kick off the usual process. That will also provide an updated feature flag key to put here.

src/Infrastructure.EntityFramework/Repositories/OrganizationDomainRepository.cs
```diff
@@ -177,5 +177,23 @@ where organizationIds.Contains(d.OrganizationId) && d.VerifiedDate != null
...
+        var query = from od in dbContext.OrganizationDomains
+                    join o in dbContext.Organizations on od.OrganizationId equals o.Id
+                    join p in dbContext.Policies on o.Id equals p.OrganizationId
+                    where od.DomainName == domainName
+                        && od.VerifiedDate != null
+                        && o.Enabled
+                        && (!excludeOrganizationId.HasValue || o.Id != excludeOrganizationId.Value)
+                        && p.Type == Core.AdminConsole.Enums.PolicyType.BlockClaimedDomainAccountCreation
+                        && p.Enabled
+                    select od;
```
eliykat, line 196: We also need to check `Organization.UsePolicies` so that we don't enforce policies for an organization who has downgraded their plan. EDIT: and `Organization.UseOrganizationDomains`. This depends on multiple features.

src/Sql/dbo/Stored Procedures/OrganizationDomain_HasVerifiedDomainWithBlockPolicy.sql
```diff
@@ -0,0 +1,32 @@
...
+        INNER JOIN [dbo].[Organization] O
+            ON OD.OrganizationId = O.Id
+        INNER JOIN [dbo].[Policy] P
+            ON O.Id = P.OrganizationId
+        WHERE OD.DomainName = @DomainName
+            AND OD.VerifiedDate IS NOT NULL
+            AND O.Enabled = 1
+            AND (@ExcludeOrganizationId IS NULL OR O.Id != @ExcludeOrganizationId)
+            AND P.Type = 19  -- BlockClaimedDomainAccountCreation
+            AND P.Enabled = 1
```
eliykat, line 25: `O.UsePolicies` here as well. (EDIT: and `O.UseOrganizationDomains`)

test/Core.Test/Utilities/EmailValidationTests.cs
```diff
@@ -0,0 +1,35 @@
...
+
+public class EmailValidationTests
+{
+    [Theory]
+    [InlineData("user@Example.COM", "example.com")]
+    [InlineData("user@EXAMPLE.COM", "example.com")]
+    [InlineData("user@example.com", "example.com")]
+    [InlineData("user@Example.Com", "example.com")]
+    [InlineData("User@DOMAIN.CO.UK", "domain.co.uk")]
+    public void GetDomain_WithMixedCaseEmail_ReturnsLowercaseDomain(string email, string expectedDomain)
```
eliykat, line 15: I suggest copying across the happy path test cases from [StrictEmailAddressAttributeTests](https://github.com/bitwarden/server/blob/main/test/Core.Test/Utilities/StrictEmailAddressAttributeTests.cs#L9-L14), so that we know our email address parsing/validation is consistent across our utils.

src/Core/Auth/UserFeatures/Registration/Implementations/RegisterUserCommand.cs
eliykat: (responding to the thread above)

There are a few ideas being discussed here.

**Repeating similar code vs. extracting it into a dedicated class (DRY):** the code being repeated is fairly minimal boilerplate, especially once the feature flag check is cleaned up, and it's only in 2 places. All the substantive business logic is already encapsulated and reused in the db query. I think it's OK as is. (If there was no new db query and the substantive logic was being implemented in C# then I agree that extracting it to a separate class would be the way to go.)

**Copying claimed domain info into the policy data:** I am usually against doing this because it's denormalizing (duplicating) data in the database, then we have to keep the policy data up-to-date with the claimed domains table. Also `Policy.Data` is generally for configuration data, which is a bit different to the use case here.

**PolicyRequirements:** for context, we have been working to centralize policy enforcement through a pattern called PolicyRequirements. But this is a very unusual kind of policy: it's enforced against an unauthenticated user with no relationship to the organization. It's not looking up a policy type by organization or user as we do today, really it's looking up a domain and working backwards from there. At this stage I don't see how we could accommodate it in our existing patterns, so I'm OK with the current bespoke solution. Even though doing so much at the db level feels a bit unusual to me, it does save us multiple calls.

TL;DR I don't have any better ideas for standardizing this logic.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationDomainRepositoryTests.cs
```diff
@@ -383,4 +385,329 @@ public async Task GetVerifiedDomainsByOrganizationIdsAsync_ShouldVerifiedDomains
         Assert.Null(otherOrganizationDomain);
         Assert.Null(unverifiedDomain);
     }
+
+    [DatabaseTheory, DatabaseData]
```
eliykat: `DatabaseTheory` is deprecated, you can just use the regular `Theory` attribute.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationDomainRepositoryTests.cs
eliykat, line 1: Please use helper methods in `OrganizationTestHelpers` (the one in this project) to create users and orgs in common cases. It saves on repetitive data arrangement. Similarly, if we could benefit from any new helper methods (e.g. to create a claimed org domain), please add them.

If you need specific data arrangement unlikely to be reused between tests then it's OK to leave it in the test itself.

eliykat (changes requested):
Overall looks good, minor comments only.

src/Core/Constants.cs
kspearrin: My guess is the AC team should manage the flag.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationDomainRepositoryTests.cs
kspearrin: Feedback from Claude on this comment:

> Note on OrganizationTestHelpers:
>   - The CreateTestOrganizationAsync helper method already sets UsePolicies = true by default (line 53 in OrganizationTestHelpers.cs)
>   - For these specific tests, we kept the inline organization creation because:
>     a. The tests need specific configurations (different Enabled states, different organization names, etc.)
>     b. The helper creates Enterprise organizations with many properties set, which is overkill for these focused tests
>     c. Inline creation makes it clearer what's being tested in each specific scenario

src/Core/Constants.cs
eliykat: I've let Vince know and he will handle the admin to set this up tomorrow.

eliykat (approved):
The latest changes look good, the only remaining issue I see is getting the updated feature flag key from Vince, which we should have tomorrow. I can approve in the meantime though as the actual implementation is done.

src/Core/Constants.cs
```diff
@@ -143,6 +143,7 @@ public static class FeatureFlagKeys
     public const string AutomaticConfirmUsers = "pm-19934-auto-confirm-organization-users";
     public const string PM23845_VNextApplicationCache = "pm-24957-refactor-memory-application-cache";
     public const string AccountRecoveryCommand = "pm-25581-prevent-provider-account-recovery";
+    public const string BlockClaimedDomainAccountCreation = "block-claimed-domain-account-creation";
```
eliykat: Here is the feature flag string in LaunchDarkly:

```suggestion
    public const string BlockClaimedDomainAccountCreation = "pm-28297-block-uninvited-claimed-domain-registration";
```

eliykat (changes requested):
This feature flag has been created, here's the updated key.

JaredSnider-Bitwarden (approved):

PR #6538, opened 2025-11-04
Remove EventBasedOrganizationIntegrations feature flag

brant-livefront:
## 📔 Objective

The `EventBasedOrganizationIntegrations` feature flag has been turned on in Production since last week. This PR removes the feature flag.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
+++ b/bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
@@ -35,8 +35,9 @@ public class ProviderService : IProviderService
 {
     private static readonly PlanType[] _resellerDisallowedOrganizationTypes = [
         PlanType.Free,
-        PlanType.FamiliesAnnually,
-        PlanType.FamiliesAnnually2019
+        PlanType.FamiliesAnnually2025,
+        PlanType.FamiliesAnnually2019,
+        PlanType.FamiliesAnnually
     ];
 
     private readonly IDataProtector _dataProtector;
--- a/src/Billing/Controllers/FreshsalesController.cs
+++ b/src/Billing/Controllers/FreshsalesController.cs
@@ -158,6 +158,7 @@ public class FreshsalesController : Controller
                 planName = "Free";
                 return true;
             case PlanType.FamiliesAnnually:
+            case PlanType.FamiliesAnnually2025:
             case PlanType.FamiliesAnnually2019:
                 planName = "Families";
                 return true;
--- a/src/Core/Billing/Enums/PlanType.cs
+++ b/src/Core/Billing/Enums/PlanType.cs
@@ -18,8 +18,8 @@ public enum PlanType : byte
     EnterpriseAnnually2019 = 5,
     [Display(Name = "Custom")]
     Custom = 6,
-    [Display(Name = "Families")]
-    FamiliesAnnually = 7,
+    [Display(Name = "Families 2025")]
+    FamiliesAnnually2025 = 7,
     [Display(Name = "Teams (Monthly) 2020")]
     TeamsMonthly2020 = 8,
     [Display(Name = "Teams (Annually) 2020")]
@@ -48,4 +48,6 @@ public enum PlanType : byte
     EnterpriseAnnually = 20,
     [Display(Name = "Teams Starter")]
     TeamsStarter = 21,
+    [Display(Name = "Families")]
+    FamiliesAnnually = 22,
 }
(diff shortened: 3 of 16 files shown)


JimmyVo16 (approved):

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -521,42 +521,31 @@ public static GlobalSettings AddGlobalSettingsServices(this IServiceCollection s
...
+            if (IsAzureServiceBusEnabled(globalSettings))
             {
                 services.TryAddSingleton<IEventIntegrationPublisher, AzureServiceBusService>();
-                services.TryAddKeyedSingleton<IEventWriteService, EventIntegrationEventWriteService>("broadcast");
+                services.TryAddSingleton<IEventWriteService, EventIntegrationEventWriteService>();
             }
             else
             {
-                services.TryAddKeyedSingleton<IEventWriteService, NoopEventWriteService>("broadcast");
+                services.TryAddSingleton<IEventWriteService, NoopEventWriteService>();
```
brant-livefront: @withinfocus This bug ([PM-27880](https://bitwarden.atlassian.net/browse/PM-27880)) got me thinking about this specific case with the NoopService. This should never occur in Cloud Prod (we already have ASB configured correctly). 

But if something misconfigured it, would it be preferable to crash here? Revert to the Azure Queue service? It feels somewhat dangerous to write nothing (no warnings or errors and no events).

withinfocus: I don't really like the current branching honestly -- allow _anyone_ to use ASB, then Rabbit, then the legacy queueing, then direct database, then nothing. It's difficult for me to parse this all in a small PR, and that change can be introduced in a subsequent one, but we don't need to limit access.

Per your question though, in a cloud environment (just us as Bitwarden) I think legacy code paths should still be allowed. We do not have an intention to actually remove the queueing code I'd say, at least for the medium-term future.

withinfocus (commented):
Claude's warning is valid, is it not? Should the no-op service be registered in the self-hosted path?

Let's also do an unused references check per its other comment.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
brant-livefront: This has been cleaned up in the latest commit. It's a much cleaner set of if conditions:

ASB -> RabbitMQ -> Azure Queues -> Repository (direct to DB, only for self hosted) -> No op if all else fails

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -524,42 +524,33 @@ public static GlobalSettings AddGlobalSettingsServices(this IServiceCollection s
...
+        if (globalSettings.SelfHosted)
         {
-            services.TryAddKeyedSingleton<IEventWriteService, NoopEventWriteService>("storage");
-            services.TryAddKeyedSingleton<IEventWriteService, NoopEventWriteService>("broadcast");
+            services.TryAddSingleton<IEventWriteService, RepositoryEventWriteService>();
+            return services;
         }
 
-        services.TryAddScoped<IEventWriteService, EventRouteService>();
+        services.TryAddSingleton<IEventWriteService, NoopEventWriteService>();
```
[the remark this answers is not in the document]
brant-livefront: This is cleaned up in the latest commit. The No-op service is the last resort, only after falling through 2 levels that are/were previously configured correctly (ASB and Azure Queues).

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -524,42 +524,33 @@ public static GlobalSettings AddGlobalSettingsServices(this IServiceCollection s
...
-                services.TryAddSingleton<IEventIntegrationPublisher, RabbitMqService>();
-                services.TryAddKeyedSingleton<IEventWriteService, EventIntegrationEventWriteService>("broadcast");
-            }
-            else
-            {
-                services.TryAddKeyedSingleton<IEventWriteService, NoopEventWriteService>("broadcast");
-            }
+        if (CoreHelpers.SettingHasValue(globalSettings.Events.ConnectionString))
+        {
+            services.AddSingleton<IEventWriteService, AzureQueueEventWriteService>();
```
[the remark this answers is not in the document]
brant-livefront: Fixed in latest

withinfocus (approved):

PR #6546, opened 2025-11-05
[PM-27731] Updated organization licenses to save the correct values from the token

cturnbull-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27731

## 📔 Objective

Fix a bug where self-hosted organizations updating their license files were not using the Token JWT as the authoritative source of truth.

The `UpdateOrganizationLicenseCommand` was only extracting 1 out of 42 properties from Token-based (JWT claims) licenses. The other 41 properties were read from the deserialized JSON file properties instead of the authoritative Token claims. This created an architectural inconsistency where the Token (signed by the cloud) was not the single source of truth, and stale or incorrect values in the JSON properties could override the Token claims.

This PR ensures the Token JWT is the single source of truth by extracting all organization properties from claims before validation and database update. Also fixes the same issue in user license updates and adds missing properties to `SelfHostedOrganizationDetails.ToOrganization()`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/test/Core.Test/Billing/Licenses/LicenseConstantsTests.cs
+++ b/test/Core.Test/Billing/Licenses/LicenseConstantsTests.cs
@@ -1,4 +1,4 @@
-using System.Reflection;
+﻿using System.Reflection;
 using Bit.Core.Billing.Licenses;
 using Bit.Core.Billing.Organizations.Models;
 using Xunit;
--- a/test/Core.Test/Billing/Licenses/Services/Implementations/OrganizationLicenseClaimsFactoryTests.cs
+++ b/test/Core.Test/Billing/Licenses/Services/Implementations/OrganizationLicenseClaimsFactoryTests.cs
@@ -1,4 +1,4 @@
-using System.Reflection;
+﻿using System.Reflection;
 using Bit.Core.AdminConsole.Entities;
 using Bit.Core.Billing.Licenses;
 using Bit.Core.Billing.Licenses.Models;


r-tome (approved):

src/Core/Billing/Organizations/Commands/UpdateOrganizationLicenseCommand.cs
```diff
@@ -46,6 +47,46 @@ public async Task UpdateLicenseAsync(SelfHostedOrganizationDetails selfHostedOrg
         }
 
         var claimsPrincipal = _licensingService.GetClaimsPrincipalFromLicense(license);
+
+        // If the license has a Token (claims-based), extract all properties from claims BEFORE validation
+        // This ensures that CanUseLicense validation has access to the correct values from claims
+        // Otherwise, fall back to using the properties already on the license object (backward compatibility)
+        if (claimsPrincipal != null)
```
cyprain-okeke, line 55: I am thinking we should check for when the claimsPrincipal is null  and log a warning message or throw new BadRequestException("Invalid license token") ; is required.

test/Core.Test/Billing/Organizations/Commands/UpdateOrganizationLicenseCommandTests.cs
```diff
@@ -99,6 +100,139 @@ await sutProvider.GetDependency<IOrganizationService>()
         }
     }
 
+    [Theory, BitAutoData]
```
cyprain-okeke, line 107: maybe we should add more test  like :

- Null ClaimsPrincipal
- Wrong installation ID in claims
- Expired license with valid claims

cyprain-okeke (changes requested):
Nice Work sir, Left some few comments

src/Core/Billing/Organizations/Commands/UpdateOrganizationLicenseCommand.cs
connerbw: Thanks for the review! I don't think we can add a null check here because a null `claimsPrincipal` is a valid backward compatibility case, not an error condition.

The code supports two license types:
1. **Token-based licenses** (newer): JWT with claims -> claimsPrincipal is populated
2. **Legacy licenses** (older): No token -> claimsPrincipal is null, properties read from JSON

The `if (claimsPrincipal != null)` at line 54 already handles both cases. Invalid licenses (regardless of type) are caught by the `CanUse` validation at lines 90-96, which already throws `BadRequestException` with appropriate error messages for issues like wrong installation ID, disabled orgs, etc.

Adding an error for null `claimsPrincipal` would break backward compatibility with existing license files that don't have tokens. In the future, when we drop support for the older system, this is a good test case we could come back in and add.

test/Core.Test/Billing/Organizations/Commands/UpdateOrganizationLicenseCommandTests.cs
connerbw: Added these new test cases, except for the first!

r-tome (dismissed):

src/Core/Billing/Organizations/Commands/UpdateOrganizationLicenseCommand.cs
```diff
@@ -46,6 +47,46 @@ public async Task UpdateLicenseAsync(SelfHostedOrganizationDetails selfHostedOrg
...
+            license.LicenseKey = claimsPrincipal.GetValue<string>(OrganizationLicenseConstants.LicenseKey);
+            license.UsePasswordManager = claimsPrincipal.GetValue<bool>(OrganizationLicenseConstants.UsePasswordManager);
+            license.UseSecretsManager = claimsPrincipal.GetValue<bool>(OrganizationLicenseConstants.UseSecretsManager);
+            license.SmSeats = claimsPrincipal.GetValue<int?>(OrganizationLicenseConstants.SmSeats);
+            license.SmServiceAccounts = claimsPrincipal.GetValue<int?>(OrganizationLicenseConstants.SmServiceAccounts);
+            license.UseRiskInsights = claimsPrincipal.GetValue<bool>(OrganizationLicenseConstants.UseRiskInsights);
+            license.UseOrganizationDomains = claimsPrincipal.GetValue<bool>(OrganizationLicenseConstants.UseOrganizationDomains);
+            license.UseAdminSponsoredFamilies = claimsPrincipal.GetValue<bool>(OrganizationLicenseConstants.UseAdminSponsoredFamilies);
+            license.UseAutomaticUserConfirmation = claimsPrincipal.GetValue<bool>(OrganizationLicenseConstants.UseAutomaticUserConfirmation);
+        }
```
cyprain-okeke, line 100: The code extracts 29 properties, but two constants appear to be missing from OrganizationLicenseConstants:
  - UseDisableSmAdsForUsers
  - UsePhishingBlocker.  Is this intentional?

connerbw: Nope, it wasn't. Those two seem to be new additions from when I opened this PR. I'll get them added, good catch

connerbw: Good catch @cyprain-okeke , I just pushed a commit addressing this. I also added a set of unit tests to guide future developers through the process of adding a property to the organization license

r-tome (approved):

PR #6571, opened 2025-11-12
[PM-27599] remove hardcoded storage values, fetch from pricing service instead

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27599

## 📔 Objective

Remove all hardcoded 1GB values for included storage, instead pulling from the plan itself.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Billing/Models/StaticStore/Plan.cs
+++ b/src/Core/Billing/Models/StaticStore/Plan.cs
@@ -97,7 +97,7 @@ public abstract record Plan
         public decimal PremiumAccessOptionPrice { get; init; }
         public short? MaxSeats { get; init; }
         // Storage
-        public short? BaseStorageGb { get; init; }
+        public short BaseStorageGb { get; init; }
         public bool HasAdditionalStorageOption { get; init; }
         public decimal AdditionalStoragePricePerGb { get; init; }
         public string StripeStoragePlanId { get; init; }
--- a/src/Core/Billing/Pricing/Organizations/PlanAdapter.cs
+++ b/src/Core/Billing/Pricing/Organizations/PlanAdapter.cs
@@ -99,7 +99,7 @@ public record PlanAdapter : Core.Models.StaticStore.Plan
             _ => true);
         var baseSeats = GetBaseSeats(plan.Seats);
         var maxSeats = GetMaxSeats(plan.Seats);
-        var baseStorageGb = (short?)plan.Storage?.Provided;
+        var baseStorageGb = (short)(plan.Storage?.Provided ?? 0);
         var hasAdditionalStorageOption = plan.Storage != null;
         var additionalStoragePricePerGb = plan.Storage?.Price ?? 0;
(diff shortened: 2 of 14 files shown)


src/Core/Billing/Models/StaticStore/Plan.cs
```diff
@@ -97,7 +97,7 @@ public record PasswordManagerPlanFeatures
         public decimal PremiumAccessOptionPrice { get; init; }
         public short? MaxSeats { get; init; }
         // Storage
-        public short? BaseStorageGb { get; init; }
```
cyprain-okeke, line 100: if existing database records have NULL BaseStorageGb or a NULL BaseStorageGb is introduce, this will cause breaking change.

src/Core/Billing/Pricing/Organizations/PlanAdapter.cs
```diff
@@ -99,7 +99,7 @@ private static PasswordManagerPlanFeatures ToPasswordManagerPlanFeatures(Plan pl
             _ => true);
         var baseSeats = GetBaseSeats(plan.Seats);
         var maxSeats = GetMaxSeats(plan.Seats);
-        var baseStorageGb = (short?)plan.Storage?.Provided;
+        var baseStorageGb = (short)(plan.Storage?.Provided ?? 0);
```
cyprain-okeke, line 102: this will default to 0GB if storage configuration is missing from pricing service and could cause validation errors. either use the existing or the best option is to validate and throw - don't allow invalid states

```
  var baseStorageGb = plan.Storage?.Provided
      ?? throw new InvalidOperationException("Storage config missing"); // ✅ GOOD

  if (baseStorageGb <= 0)
      throw new InvalidOperationException($"Invalid storage: {baseStorageGb}GB");

  return (short)baseStorageGb;
```

src/Core/Billing/Models/StaticStore/Plan.cs
kdenney: To my knowledge and from looking at all usages of this class, plans themselves are not stored in the database so this would not be a concern. Is that not true?

src/Core/Billing/Pricing/Organizations/PlanAdapter.cs
kdenney: I felt like assigning a zero base storage was actually more accurate here, when the plan from the pricing service has a null storage defined. (Null storage on a plan is actually a valid case as with the Free organization plan - no storage is provided.) So, this code in PlanAdapter is specifically saying that if the pricing service returns a plan with null storage, we will treat that as 0GB as the base, which imo is the most accurate way of representing that and also has the benefit of not needing to deal with a null baseStorageGb elsewhere in the code. But I'm happy to discuss if you disagree!

cyprain-okeke (approved):

PR #6588, opened 2025-11-17
feat(auth-validator): [Auth/PM-22975] Client Version Validator

Patrick-Pimentel-Bitwarden:
# **NEEDS TESTING ON FEATURE BRANCH**

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22975

## 📔 Objective

The objective of this ticket is to create a client version validator that checks the header data on a token request and prevent users who have migrated to v2 from logging in with older client versions.

1. Extracted out KM utilities for checking isV2Encryption from RotateUserAccountKeysCommand
2. Added some clarity and comments in BaseRequestValidator
3. Installed the validation for client version in the appropriate place to prevent account enumeration

## 📸 Screenshots

https://github.com/user-attachments/assets/cf4bc61a-3ec8-4d28-a1c1-19690b755f6a

https://github.com/user-attachments/assets/36f51ae3-5bb0-4b65-9df3-595048b1cfbf

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
 
-    <Version>2026.2.0</Version>
+    <Version>2026.2.1</Version>
 
     <RootNamespace>Bit.$(MSBuildProjectName)</RootNamespace>
     <ImplicitUsings>enable</ImplicitUsings>
--- a/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
+++ b/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
@@ -227,7 +227,7 @@ public class OrganizationUsersController : BaseAdminConsoleController
     public async Task<OrganizationUserResetPasswordDetailsResponseModel> GetResetPasswordDetails(Guid orgId, Guid id)
     {
         var organizationUser = await _organizationUserRepository.GetByIdAsync(id);
-        if (organizationUser is null || organizationUser.UserId is null)
+        if (organizationUser is null || organizationUser.OrganizationId != orgId || organizationUser.UserId is null)
         {
             throw new NotFoundException();
         }
--- a/test/Api.Test/AdminConsole/Controllers/OrganizationUsersControllerTests.cs
+++ b/test/Api.Test/AdminConsole/Controllers/OrganizationUsersControllerTests.cs
@@ -342,6 +342,72 @@ public class OrganizationUsersControllerTests
 
     [Theory]
     [BitAutoData]
+    public async Task GetResetPasswordDetails_WhenOrganizationUserNotFound_ThrowsNotFound(
+        Guid orgId, Guid orgUserId,
+        SutProvider<OrganizationUsersController> sutProvider)
+    {
+        // Arrange
+        sutProvider.GetDependency<IOrganizationUserRepository>().GetByIdAsync(orgUserId).Returns((OrganizationUser)null);
+
+        // Act & Assert
+        await Assert.ThrowsAsync<NotFoundException>(() => sutProvider.Sut.GetResetPasswordDetails(orgId, orgUserId));
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task GetResetPasswordDetails_WhenOrganizationIdMismatch_ThrowsNotFound(
+        Guid orgId, Guid orgUserId, OrganizationUser organizationUser,
+        SutProvider<OrganizationUsersController> sutProvider)
+    {
+        // Arrange
+        organizationUser.OrganizationId = Guid.NewGuid(); // Different org ID
+        organizationUser.UserId = Guid.NewGuid();
+        sutProvider.GetDependency<IOrganizationUserRepository>().GetByIdAsync(orgUserId).Returns(organizationUser);
+
+        // Act & Assert
+        await Assert.ThrowsAsync<NotFoundException>(() => sutProvider.Sut.GetResetPasswordDetails(orgId, orgUserId));
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task GetResetPasswordDetails_WhenUserIdIsNull_ThrowsNotFound(
+        Guid orgId, Guid orgUserId, OrganizationUser organizationUser,
+        SutProvider<OrganizationUsersController> sutProvider)
+    {
+        // Arrange
+        organizationUser.OrganizationId = orgId;
+        organizationUser.UserId = null;
+        sutProvider.GetDependency<IOrganizationUserRepository>().GetByIdAsync(orgUserId).Returns(organizationUser);
+
+        // Act & Assert
+        await Assert.ThrowsAsync<NotFoundException>(() => sutProvider.Sut.GetResetPasswordDetails(orgId, orgUserId));
+    }
+
+    [Theory]
+    [BitAutoData]
+    public async Task GetResetPasswordDetails_WhenValid_ReturnsDetails(
+        Guid orgId, Guid orgUserId, OrganizationUser organizationUser, User user, Organization org,
+        SutProvider<OrganizationUsersController> sutProvider)
+    {
+        // Arrange
+        organizationUser.OrganizationId = orgId;
+        organizationUser.UserId = user.Id;
+        sutProvider.GetDependency<IOrganizationUserRepository>().GetByIdAsync(orgUserId).Returns(organizationUser);
+        sutProvider.GetDependency<IUserService>().GetUserByIdAsync(user.Id).Returns(user);
+        sutProvider.GetDependency<IOrganizationRepository>().GetByIdAsync(orgId).Returns(org);
+
+        // Act
+        var response = await sutProvider.Sut.GetResetPasswordDetails(orgId, orgUserId);
+
+        // Assert
+        Assert.Equal(organizationUser.Id, response.OrganizationUserId);
+        Assert.Equal(user.Kdf, response.Kdf);
+        Assert.Equal(user.KdfIterations, response.KdfIterations);
+        Assert.Equal(org.PrivateKey, response.EncryptedPrivateKey);
+    }
+
+    [Theory]
+    [BitAutoData]
     public async Task DeleteAccount_WhenCurrentUserNotFound_ReturnsUnauthorizedResult(
         Guid orgId, Guid id, SutProvider<OrganizationUsersController> sutProvider)
     {
--- a/util/Seeder/Factories/OrganizationSeeder.cs
+++ b/util/Seeder/Factories/OrganizationSeeder.cs
@@ -3,18 +3,19 @@ using Bit.Core.Billing.Enums;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Utilities;
+using Bit.Seeder.Services;
 
 namespace Bit.Seeder.Factories;
 
 internal static class OrganizationSeeder
 {
-    internal static Organization Create(string name, string domain, int seats, string? publicKey = null, string? privateKey = null, PlanType planType = PlanType.EnterpriseAnnually)
+    internal static Organization Create(string name, string domain, int seats, IManglerService manglerService, string? publicKey = null, string? privateKey = null, PlanType planType = PlanType.EnterpriseAnnually)
     {
         var org = new Organization
         {
             Id = CoreHelpers.GenerateComb(),
-            Identifier = domain,
-            Name = name,
+            Identifier = manglerService.Mangle(domain),
+            Name = manglerService.Mangle(name),
             BillingEmail = $"billing@{domain}",
             Seats = seats,
             Status = OrganizationStatusType.Created,
--- a/util/Seeder/Recipes/OrganizationWithUsersRecipe.cs
+++ b/util/Seeder/Recipes/OrganizationWithUsersRecipe.cs
@@ -26,7 +26,7 @@ public class OrganizationWithUsersRecipe(
         // Generate organization keys
         var orgKeys = RustSdkService.GenerateOrganizationKeys();
         var organization = OrganizationSeeder.Create(
-            name, domain, seats, orgKeys.PublicKey, orgKeys.PrivateKey);
+            name, domain, seats, manglerService, orgKeys.PublicKey, orgKeys.PrivateKey);
 
         // Create owner with SDK-generated keys
         var ownerUser = UserSeeder.Create($"owner@{domain}", passwordHasher, manglerService);
--- a/util/Seeder/Recipes/OrganizationWithVaultRecipe.cs
+++ b/util/Seeder/Recipes/OrganizationWithVaultRecipe.cs
@@ -61,7 +61,7 @@ public class OrganizationWithVaultRecipe(
 
         // Create organization via factory
         var organization = OrganizationSeeder.Create(
-            options.Name, options.Domain, seats, orgKeys.PublicKey, orgKeys.PrivateKey, options.PlanType);
+            options.Name, options.Domain, seats, manglerService, orgKeys.PublicKey, orgKeys.PrivateKey, options.PlanType);
 
         // Create owner user via factory
         var ownerEmail = $"owner@{options.Domain}";
--- a/util/Seeder/Seeds/README.md
+++ b/util/Seeder/Seeds/README.md
@@ -13,21 +13,18 @@ Hand-crafted JSON fixtures for Bitwarden Seeder test data.
 ```
 Seeds/
 ├── fixtures/           Your seed data goes here
-│   ├── ciphers/        Vault items (logins, cards, identities, notes)
+│   ├── ciphers/        Vault items
 │   ├── organizations/  Organization definitions
 │   ├── rosters/        Users, groups, collections, permissions
 │   └── presets/        Complete seeding scenarios
 ├── schemas/            JSON Schema validation (auto-checked by editors)
 ├── templates/          Starter files - copy these
-│   └── CONTRIBUTING.md Detailed guide for contributors
 └── README.md           This file
 ```
 
 ## Fixtures Overview
 
 ### Ciphers
-
-Vault items - logins, cards, identities, secure notes.
 
 | Type         | Required Object | Description                |
 | ------------ | --------------- | -------------------------- |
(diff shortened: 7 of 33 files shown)


src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
```diff
@@ -119,6 +123,13 @@ await BuildSuccessResultAsync(validatorContext.User, context, validatorContext.D
                 return;
             }
 
+            // 1.5 Now check the version number of the client. Do this after ValidateContextAsync so that
+            // we prevent account enumeration. If we were to do this before ValidateContextAsync, then attackers
+            // could use a known invalid client version and make a request for a user (before we know if they have
+            // demonstrated ownership of the account via correct credentials) and identify if they exist by getting
+            // an error response back from the validator saying the user is not compatible with the client.
```
Patrick-Pimentel-Bitwarden: 💭 Could someone double check my explanation here?

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
```diff
@@ -119,6 +123,13 @@ await BuildSuccessResultAsync(validatorContext.User, context, validatorContext.D
...
+            await ValidateClientVersionAsync(context, validatorContext);
```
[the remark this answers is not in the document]
Patrick-Pimentel-Bitwarden: fixed with [91af02b](https://github.com/bitwarden/server/pull/6588/commits/91af02b9d28381430b1d617b94498a9e7d8d3ff1)

src/Core/KeyManagement/Utilities/EncryptionParsing.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public static class EncryptionParsing
+{
+    /// <summary>
+    /// Helper method to convert an encryption type string to an enum value.
+    /// </summary>
+    public static EncryptionType GetEncryptionType(string encString)
+    {
+        var parts = encString.Split('.');
+        if (parts.Length == 1)
```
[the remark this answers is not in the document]
Patrick-Pimentel-Bitwarden: This is code copied from KM, I'm hesitant to make more changes to their work.

test/Identity.IntegrationTest/Login/ClientVersionGateTests.cs
```diff
@@ -0,0 +1,117 @@
...
+    public async Task TokenEndpoint_GrantTypePassword_V2User_OnOldClientVersion_Blocked(RegisterFinishRequestModel requestModel)
+    {
+        var localFactory = new IdentityApplicationFactory();
+        var server = localFactory.Server;
+        var user = await localFactory.RegisterNewIdentityFactoryUserAsync(requestModel);
+
+        // Make user V2: set private key to COSE and add signature key pair
+        var db = localFactory.GetDatabaseContext();
+        var efUser = await db.Users.FirstAsync(u => u.Email == user.Email);
+        efUser.PrivateKey = "7.cose";
```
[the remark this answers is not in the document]
Patrick-Pimentel-Bitwarden: Read and not relevant

Patrick-Pimentel-Bitwarden:
I am seeing 100 failures in tests, double checking what is going on, consider holding off on approving.

src/Core/KeyManagement/Queries/IsV2EncryptionUserQuery.cs
```diff
@@ -0,0 +1,31 @@
...
+using Bit.Core.KeyManagement.Queries.Interfaces;
+using Bit.Core.KeyManagement.Repositories;
+using Bit.Core.KeyManagement.Utilities;
+
+namespace Bit.Core.KeyManagement.Queries;
+
+public class IsV2EncryptionUserQuery(IUserSignatureKeyPairRepository userSignatureKeyPairRepository)
+    : IIsV2EncryptionUserQuery
+{
+    public async Task<bool> Run(User user)
```
quexten: Nice, this query makes sense and is well re-usable.

src/Core/KeyManagement/Queries/IsV2EncryptionUserQuery.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(user);
+
+        var hasSignatureKeyPair = await userSignatureKeyPairRepository.GetByUserIdAsync(user.Id) != null;
+        var isPrivateKeyEncryptionV2 =
+            !string.IsNullOrWhiteSpace(user.PrivateKey) &&
+            EncryptionParsing.GetEncryptionType(user.PrivateKey) == EncryptionType.XChaCha20Poly1305_B64;
+
+        return hasSignatureKeyPair switch
```
quexten: To be honest, this is nit-picky, but this would be more readable as an if/else. If this were rust matching this would be different but it took me a second to get this.
i.e 
```
if (isPrivateKeyEncryptionV2 && hasSignatureKeyPair) {

} else if (!isPrivateKeyEncryptionV2 && !hasSignatureKeyPair) {

} else {
  throw
}
```

src/Core/KeyManagement/Queries/IsV2EncryptionUserQuery.cs
```diff
@@ -0,0 +1,31 @@
...
+
+public class IsV2EncryptionUserQuery(IUserSignatureKeyPairRepository userSignatureKeyPairRepository)
+    : IIsV2EncryptionUserQuery
+{
+    public async Task<bool> Run(User user)
+    {
+        ArgumentNullException.ThrowIfNull(user);
+
+        var hasSignatureKeyPair = await userSignatureKeyPairRepository.GetByUserIdAsync(user.Id) != null;
+        var isPrivateKeyEncryptionV2 =
```
quexten: Note: This logic is fine, but a simpler way would be to check the user's security version property. If it is >=2 it is a encryption v2 user, else (null) a V1 user. That way you could even just implement a function on the user class "IsV2Encryption()" and call user.IsV2Encryption()

src/Core/KeyManagement/Utilities/EncryptionParsing.cs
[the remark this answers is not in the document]
quexten: Yeah, this code was inherited when KM formed. But looks reasonable.

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
quexten: Is the validatorContext.User value even set before validating? (I'm not sure)

Explanation makes sense though and is very helpful.

src/Core/KeyManagement/Utilities/EncryptionParsing.cs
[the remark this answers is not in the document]
quexten: Agree, best to ignore claude here, you are just extracting existing code.

src/Core/KeyManagement/Queries/IsV2EncryptionUserQuery.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Bit.Core.Entities;
+using Bit.Core.Enums;
+using Bit.Core.KeyManagement.Queries.Interfaces;
+using Bit.Core.KeyManagement.Repositories;
+using Bit.Core.KeyManagement.Utilities;
+
+namespace Bit.Core.KeyManagement.Queries;
+
+public class IsV2EncryptionUserQuery(IUserSignatureKeyPairRepository userSignatureKeyPairRepository)
```
quexten: Ok, so having thought about this a bit, I think we should remove this query. We can do either `user.GetSecurityVersion() >= 2` or implement "IsV2User" on the user class which does the same.

Patrick-Pimentel-Bitwarden: So I took this feedback and tweaked it a little. I introduced a narrower scoped check:

`user.IsSecurityVersionTwo()`

I felt this was more defensive because now we have breadcrumbed places where we are doing branching logic based on a V2 encryption requirement.

While I am happy to be using this simpler approach than the IsV2User query I want to recognize this is a deviation from the ticket's original tech breakdown and should this be discussed to confirm this wouldn't create any issues?

See the changes in [753670d](https://github.com/bitwarden/server/pull/6588/commits/753670d26fee9314f7502b70ccf8324ee480168e)

src/Core/KeyManagement/Queries/IsV2EncryptionUserQuery.cs
Patrick-Pimentel-Bitwarden: No longer needed with using the SecurityVersion column on the user.

src/Core/KeyManagement/Queries/IsV2EncryptionUserQuery.cs
Patrick-Pimentel-Bitwarden: So after a round of discussion around the software engineering approach to this solution we arrived at the work that was done [f719763](https://github.com/bitwarden/server/pull/6588/commits/f719763a85a489951dfb3a5ad4d20d5cb7138823)

Have a look @quexten and let me know what you think. This approach keeps some attachment to looking at the private key structure which we feel is a more defensive approach than just looking at the SecurityVersion column.

quexten: I guess it's a little confusing, but a V2 is every version after and including 2, not just 2. But it's OK to leave it as is right now, since we don't yet have a direct plan for implementing one of the features that would migrate a V2 user to a security version that is larger than 2.

I'm ok with the solution as-is, thanks for spending the time here!

test/Core.Test/KeyManagement/Queries/GetMinimumClientVersionForUserQueryTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        Assert.Equal(Core.KeyManagement.Constants.MinimumClientVersionForV2Encryption, version);
+    }
+
+    [Fact]
+    public async Task Run_ReturnsNull_ForV1User()
+    {
+        var sut = new GetMinimumClientVersionForUserQuery();
+        var version = await sut.Run(new User
+        {
+            SecurityVersion = 1,
```
quexten: Could we add a test for security version being null?

Patrick-Pimentel-Bitwarden: good catch!

Patrick-Pimentel-Bitwarden: Added that test and more in [cff2f5d](https://github.com/bitwarden/server/pull/6588/commits/cff2f5df6dd50bf1acbb77e1d971df97108424bb)

src/Core/KeyManagement/Queries/GetMinimumClientVersionForUserQuery.cs
JaredSnider-Bitwarden, line 1: 🎨 : I'm not sure this query is worth the boilerplate. It's not injecting a repository and it doesn't hit the database. It references a `User` object encapsulated method `IsSetupForV2Encryption` and a constant `MinimumClientVersionForV2Encryption` - both of which can be directly accessed in its only use within the `ClientVersionValidator`. Was there something discussed with KM about this needing to be in a re-usable location vs a private method on the `ClientVersionValidator` itself?

src/Core/Entities/User.cs
```diff
@@ -211,6 +212,25 @@ public int GetSecurityVersion()
         return SecurityVersion ?? 1;
     }
 
+    public bool IsSetupForV2Encryption()
```
JaredSnider-Bitwarden: 🎨 (nit): I find this mostly clear, but there is a small amount of ambiguity in this name. `IsSetupFor` can imply they are setup for a potential migration to V2 encryption, but we actually mean that they are already on V2 encryption if this method returns true. Something like `HasV2Encryption` might clarify this but also a `<summary>` on what the method is intending could work as well.

JaredSnider-Bitwarden: Aside: the composition of user methods is exactly what I was looking for. Thank you!

src/Core/KeyManagement/Utilities/EncryptionParsing.cs
JaredSnider-Bitwarden, line 1: 🎨 : Since we have constants for known v1 and v2 key types, can we write a quick unit test suite for this helper?

src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs
JaredSnider-Bitwarden: This is a great explanation and a good note to leave here so we don't re-order this and open that door in the future.

src/Core/KeyManagement/Queries/GetMinimumClientVersionForUserQuery.cs
JaredSnider-Bitwarden: I just looked at the ticket and saw this:

> Guidance from Architecture has asked us to consider if we can make this an abstraction/utility such that it can be used in other cases as well where we may have a need in the future to prevent authentication on incompatible versions for different reasons.

```
VersionFeatureCompatibility
{
  MethodToCallToCheckIfFeatureIsEnabledForUser(User u) => boolean { get; set; }
  MinimumVersion => string { get; set; }
}
```

I think a shared utility like this would be cleaner for re-use vs trying to integrate it with the CQS DB approach we have.

src/Identity/IdentityServer/RequestValidators/ClientVersionValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+public class ClientVersionValidator(
+    ICurrentContext currentContext,
+    IGetMinimumClientVersionForUserQuery getMinimumClientVersionForUserQuery)
+    : IClientVersionValidator
+{
+    private static readonly string UpgradeMessage = "Please update your app to continue using Bitwarden";
+
+    public async Task<bool> ValidateAsync(User? user, CustomValidatorRequestContext requestContext)
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: @Patrick-Pimentel-Bitwarden , can you take a look at this feedback? Are requests without a user at this point valid?

test/Common/Constants/TestEncryptionConstants.cs
JaredSnider-Bitwarden, line 1: Thank you for making these for easy re-use in our tests!

test/Common/Constants/TestEncryptionConstants.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Bit.Test.Common.Constants;
+
+public static class TestEncryptionConstants
+{
+    // V1-style encrypted strings (AES-CBC-HMAC formats) accepted by validators
+    public const string V1EncryptedBase64 = "2.QmFzZTY0UGFydA==|QmFzZTY0UGFydA==|QmFzZTY0UGFydA==";
```
JaredSnider-Bitwarden: ❓ : could we try to come up with a better name here or add to the summary to indicate which types of request fields this is expected to be used in? I had to go look at the tests to understand how it is supposed to be used.

test/Identity.IntegrationTest/Login/ClientVersionGateTests.cs
JaredSnider-Bitwarden, line 1: Great tests and test coverage!

src/Core/KeyManagement/Queries/GetMinimumClientVersionForUserQuery.cs
Patrick-Pimentel-Bitwarden: Ah, yeah see it was originally injecting a repository but then following guidance from KM and conversation with you we just attached it to checking properties on the user. I can take it out of query form.

Patrick-Pimentel-Bitwarden: Spending some more time thinking about this, I remembered at refinement time I do believe we moved away from the generic VersionFeatureCompatibility as it could also grow the scope of this work more towards platform tooling. Also the VersionFeatureCompatibility is absent from the technical breakdown we wrote. If we are to do a more generic VersionFeatureCompatibility it very may well require repositories in the future and I think it would keep its value better following the CQS DB pattern no?

Happy to sync on this one if you want.

JaredSnider-Bitwarden: Ok, that's fair to reduce the scope down to just the validator we have. I still would like to request that we please move the checks into the validator itself as any generalization in the future wouldn't have the same shape as the query anyhow.

test/Common/Constants/TestEncryptionConstants.cs
quexten: I suggest: "AES256_CBC_HMAC_Encstring", and putting the information that this is meant for V1 users into a comment. Technically, even V2 users can have AES256_CBC_HMAC encstrings in their account, for anything that is not directly encrypted by the user keys. Organization ciphers are currently also all AES256_CBC_HMAC.

Patrick-Pimentel-Bitwarden: fixed with [2264056](https://github.com/bitwarden/server/pull/6588/commits/226405609e4e5f935eb0c819170c07dc94d7fe9d)

src/Core/KeyManagement/Queries/GetMinimumClientVersionForUserQuery.cs
Patrick-Pimentel-Bitwarden: Addressed comments in [2264056](https://github.com/bitwarden/server/pull/6588/commits/226405609e4e5f935eb0c819170c07dc94d7fe9d)

src/Core/KeyManagement/Utilities/EncryptionParsing.cs
Patrick-Pimentel-Bitwarden: added in the latest commits [36e7b1c](https://github.com/bitwarden/server/pull/6588/commits/36e7b1c65ec21ef5b4b541fa493c91ef7d7fda3f) and [55bfb71](https://github.com/bitwarden/server/pull/6588/commits/55bfb71bef5a981f15b2153afcd437f886b834c0)

src/Identity/IdentityServer/RequestValidators/ClientVersionValidator.cs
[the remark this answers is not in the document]
Patrick-Pimentel-Bitwarden: Addressed!

test/Common/Constants/TestEncryptionConstants.cs
```diff
@@ -5,6 +5,15 @@ public static class TestEncryptionConstants
     // Intended for use as a V1 encrypted string, accepted by validators
     public const string AES256_CBC_HMAC_Encstring = "2.QmFzZTY0UGFydA==|QmFzZTY0UGFydA==|QmFzZTY0UGFydA==";
 
+    // Simple stubs for other encstring versions used by parsing tests
+    public const string AES256_CBC_B64_Encstring = "0.stub";
+    public const string AES128_CBC_HMACSHA256_B64_Encstring = "1.stub";
```
quexten: Noting: We don't support this encstring type anymore, it does not exist in the real world and client / sdk support was removed.

test/Common/Constants/TestEncryptionConstants.cs
```diff
@@ -5,6 +5,15 @@ public static class TestEncryptionConstants
...
+    public const string AES256_CBC_HMAC_EmptySuffix = "2.";
+    public const string RSA2048_OAEPSHA256_B64_Encstring = "3.stub";
+    public const string RSA2048_OAEPSHA1_B64_Encstring = "4.stub";
+    public const string RSA2048_OAEPSHA256_HMACSHA256_B64_Encstring = "5.stub";
```
quexten: Noting: Between the RSA types, only type 4 is known to exist in the real world.

test/Common/Constants/TestEncryptionConstants.cs
```diff
@@ -5,6 +5,15 @@ public static class TestEncryptionConstants
...
+    public const string RSA2048_OAEPSHA1_HMACSHA256_B64_Encstring = "6.stub";
```
quexten: Could we add a Type 7 string?

test/Common/Constants/TestEncryptionConstants.cs
Patrick-Pimentel-Bitwarden: Would you like me to leave this comment amongst the testing constants? Or is there another action you would like me to take?

JaredSnider-Bitwarden: Let's only add ones that we expect to use

test/Common/Constants/TestEncryptionConstants.cs
JaredSnider-Bitwarden: Let's remove it then

test/Common/Constants/TestEncryptionConstants.cs
JaredSnider-Bitwarden: Or we can just keep these as examples

test/Common/Constants/TestEncryptionConstants.cs
JaredSnider-Bitwarden: Or mark it as deprecated

quexten: Imo: Mark AES256_CBC_B64_Encstring as deprecated, remove the AES128_CB_HMAC256_B64 encstring.

test/Common/Constants/TestEncryptionConstants.cs
quexten: I think it makes sense to only add the ones we expect to see. We probably eventually want to remove any reference to these other types from both clients and servers, and there is no reason to add new uses of it.

Patrick-Pimentel-Bitwarden: fixed with [50298fb](https://github.com/bitwarden/server/pull/6588/commits/50298fbbce4227bbd5c9f90d28fcf4ae30d3edab)

test/Common/Constants/TestEncryptionConstants.cs
Patrick-Pimentel-Bitwarden: added with [79256d7](https://github.com/bitwarden/server/pull/6588/commits/79256d73ee355229ed1e9ecec4885402877449d6)

JaredSnider-Bitwarden (dismissed):

quexten (dismissed):

prograhamming (approved):

ike-kottlowski (approved):

trmartin4:
Resolved merge conflicts again.

ike-kottlowski (approved):

quexten (approved):

prograhamming (approved):

PR #6599, opened 2025-11-19
[PM-25913] Fix owners unable to rename provider-managed organization

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25913

## 📔 Objective

Fix a bug where the owner of an organization managed by a provider could not update the organization name.

**Root cause:** changing an organization name triggers a Stripe update, which checks whether the owner has permission to manage billing. Owners cannot manage billing if the organization is managed by a provider (it's managed by the provider), so the change fails.

**Solution:** separate authorization checks from business logic. Previously this was tightly coupled because `shouldUpdateBilling` was evaluated in the controller and passed into the service method. With this PR, the service method is moved to a command, and the command logic is separated from authorization checks in the controller. This allows owners to change the name, even though this will trigger a Stripe update, and even though they do not have permission to make Stripe changes otherwise.

**Other changes:** There was some accumulated tech debt here I kept tripping over:
- the business logic was still in `OrganizationService` rather than a command, and was a mix of AC and Billing logic. AC logic has been moved into a command and Billing logic into one of their services (as discussed with @amorask-bitwarden)
- deprecated/unused properties in the request model. These have been removed.
- `OrganizationsController` did not use SutProvider, requiring manual updates to dependencies every time. This has been updated.
- this flow has a lot of optionality to it: you can update any combination of the organization name, billing email, and public and private keys on cloud, or only some of these on self-host. This was not well documented. I have updated the models and the private API to use nullable properties because this is a simpler API to communicate "no changes required for this property". The flow of the command also makes the self-host/cloud split clear.
- no tests, which have now been added.

Web PR: https://github.com/bitwarden/clients/pull/17482

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/OrganizationFeatures/Organizations/UpdateOrganizationCommand.cs
```diff
@@ -0,0 +1,142 @@
...
+    /// If we remove this migration code in the future, this command and endpoint can become cloud only.
+    /// </summary>
+    private async Task<Organization> UpdateSelfHostedAsync(Organization organization, UpdateOrganizationRequest request)
+    {
+        UpdatePublicPrivateKeyPair(organization, request);
+        await organizationService.ReplaceAndUpdateCacheAsync(organization, EventType.Organization_Updated);
+        return organization;
+    }
+
+    private static void UpdateOrganizationDetails(Organization organization, UpdateOrganizationRequest request)
```
JimmyVo16: Not a dealbreaker, but I’m always uneasy about modifying input parameters. It isn’t obvious from looking at the method, and it can lead to race conditions.

Based on what you’re trying to do, I think using an extension method would make more sense here and provide more idiomatic meaning.

src/Core/AdminConsole/OrganizationFeatures/Organizations/UpdateOrganizationCommand.cs
```diff
@@ -0,0 +1,142 @@
...
+            organization.BillingEmail = request.BillingEmail.ToLowerInvariant().Trim();
+        }
+    }
+
+    /// <summary>
+    /// Updates the organization public and private keys if provided and not already set.
+    /// This is legacy code for old organizations that were not created with a public/private keypair. It is a soft
+    /// migration that will silently migrate organizations when they change their details.
+    /// </summary>
+    private static void UpdatePublicPrivateKeyPair(Organization organization, UpdateOrganizationRequest request)
```
JimmyVo16: [same](https://github.com/bitwarden/server/pull/6599/files#r2547930217)

src/Core/AdminConsole/OrganizationFeatures/Organizations/UpdateOrganizationCommand.cs
JimmyVo16: With that said, the impact here is pretty small if the scope is only within this class.

test/Core.Test/Billing/Services/OrganizationBillingServiceTests.cs
```diff
@@ -353,4 +354,97 @@ await sutProvider.GetDependency<IStripeAdapter>()
...
+        Assert.Equal(organization.DisplayName(), customField.Value);
+    }
+
+    [Theory, BitAutoData]
+    public async Task UpdateOrganizationNameAndEmail_WhenNameIsLong_TruncatesTo30Characters(
+        Organization organization,
+        SutProvider<OrganizationBillingService> sutProvider)
+    {
+        // Arrange
+        organization.Name = "This is a very long organization name that exceeds thirty characters";
```
JimmyVo16, line 400: Non-blocker: consider this option to ensure that the string is guaranteed to meet the intended character count.

```c#
        var fixture = new Fixture();
        var nameAboveCharacterLimit = fixture.CreateMany<char>(30).ToArray().ToString();
        organization.Name = nameAboveCharacterLimit;
```

test/Core.Test/Billing/Services/OrganizationBillingServiceTests.cs
```diff
@@ -353,4 +354,97 @@ await sutProvider.GetDependency<IStripeAdapter>()
     }
 
     #endregion
+
+    [Theory, BitAutoData]
+    public async Task UpdateOrganizationNameAndEmail_UpdatesStripeCustomer(
+        Organization organization,
+        SutProvider<OrganizationBillingService> sutProvider)
+    {
+        organization.Name = "Short name";
```
JimmyVo16, line 363: Also non-blocker: [same](https://github.com/bitwarden/server/pull/6599/files#r2548034430) but with a different name for this case.

JimmyVo16 (dismissed):

src/Core/AdminConsole/OrganizationFeatures/Organizations/UpdateOrganizationCommand.cs
eliykat: Done

src/Core/AdminConsole/OrganizationFeatures/Organizations/UpdateOrganizationCommand.cs
eliykat: Done

test/Core.Test/Billing/Services/OrganizationBillingServiceTests.cs
eliykat: That works, but I think a string literal is the simpler option. The string literal will always be over 30 chars as well ;)

JimmyVo16: Yeah, it definitely is, that’s why I didn’t think it’s a blocker lol. 

The last time I ran into this problem was years ago, but I could’ve sworn there was a simpler idiom to accomplish this. Maybe that was because the company I worked for had some custom helper methods.

PR #6649, opened 2025-11-26
[PM-27131] Auto confirm policy requirement

jrmccannon:
## 🎟️ Tracking
[PM-27131](https://bitwarden.atlassian.net/browse/PM-27131)
[PM-28187](https://bitwarden.atlassian.net/browse/PM-28187)

## 📔 Objective
This will add the auto confirm policy requirement and enforcement query for enforcing the auto confirm policy for a user and organization.

This mainly is a different version of the Single Organzation Policy Requirement, except that it enforces that the user may not be a member of other organizations regardless of their status or type (Single org exempts admin, owners and providers and users in the status of invited and revoked.)

This also blocks providers from joining organizations with the auto confirm policy on and enforces its own single org policy when users attempt to join the organization.  

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27131]: https://bitwarden.atlassian.net/browse/PM-27131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-28187]: https://bitwarden.atlassian.net/browse/PM-28187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Core.Test/AdminConsole/EventIntegrations/EventIntegrationServiceCollectionExtensionsTests.cs
+++ b/test/Core.Test/AdminConsole/EventIntegrations/EventIntegrationServiceCollectionExtensionsTests.cs
@@ -9,6 +9,7 @@ using Bit.Core.Services;
 using Bit.Core.Settings;
 using Bit.Core.Utilities;
 using Microsoft.Bot.Builder;
+using Microsoft.Bot.Builder.Integration.AspNet.Core;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;


bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -412,6 +418,17 @@ public async Task AddOrganization(Guid providerId, Guid organizationId, string k
...
+        if (_featureService.IsEnabled(FeatureFlagKeys.AutomaticConfirmUsers))
+        {
+            var organizationAutoConfirmPolicyRequirement = await _policyRequirementQuery
+                .GetManyByOrganizationIdAsync<AutomaticUserConfirmationPolicyRequirement>(organizationId);
+
+            if (organizationAutoConfirmPolicyRequirement.Any())
+            {
+                throw new BadRequestException(new ProviderUsersCannotJoin().Message);
+            }
+        }
```
eliykat: This method handles linking a provider and an organization, which is OK even for auto-confirm orgs.

What we want to do is to prevent a providerUser from becoming an organizationUser, and vice versa. See https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2212790295/ACTB+Auto-confirm+security+requirements and https://bitwarden.atlassian.net/browse/PM-27131. That would be covered in the accept and confirm methods of this class (which handle joining a provider).

Given this is a separate flow, I suggest it could be done in a separate PR for https://bitwarden.atlassian.net/browse/PM-28187, so we can focus on the existing single org flows here.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -87,8 +100,7 @@ private async Task<bool> OrganizationUserConformsToTwoFactorRequiredPolicyAsync(
             .IsTwoFactorRequiredForOrganization(request.Organization!.Id);
     }
 
-    private async Task<Error?> OrganizationUserConformsToSingleOrgPolicyAsync(
-        AutomaticallyConfirmOrganizationUserValidationRequest request)
+    private async Task<Error?> OrganizationUserConformsToAutomaticUserConfirmationPolicyAsync(AutomaticallyConfirmOrganizationUserValidationRequest request)
```
eliykat: Check my thinking - and if you agree, please add a comment...
For this organization: we know that autoconfirm is enabled, because we check it above; otherwise we wouldn't be in the autoconfirm flow. So we don't need to check the (regular) Single Org Policy, only autoconfirm (which is a stricter set of requirements).
For other organizations: we don't need to enforce their policies (whether Single Org or Autoconfirm), because if there are any other organizations, the prior check will always fail.
Is that right? I am just making sure that removing the current single org check is OK.

src/Core/AdminConsole/OrganizationFeatures/Policies/Enforcement/AutoConfirm/AutomaticUserConfirmationPolicyEnforcementQuery.cs
```diff
@@ -0,0 +1,41 @@
...
+using static Bit.Core.AdminConsole.Utilities.v2.Validation.ValidationResultHelpers;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.Enforcement.AutoConfirm;
+
+public class AutomaticUserConfirmationPolicyEnforcementQuery(
+    IPolicyRequirementQuery policyRequirementQuery)
+    : IAutomaticUserConfirmationPolicyEnforcementQuery
+{
+    public async Task<ValidationResult<AutomaticUserConfirmationPolicyEnforcementRequest>> IsCompliantAsync(
+        AutomaticUserConfirmationPolicyEnforcementRequest request)
```
eliykat: This returns a ValidationResult, I think `AutomaticUserConfirmationPolicyEnforcementValidator` would be a more accurate name. It's basically a validator that is shared by multiple flows.

src/Core/AdminConsole/OrganizationFeatures/Policies/Enforcement/AutoConfirm/AutomaticUserConfirmationPolicyEnforcementRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <summary>
+    /// Request object for <see cref="AutomaticUserConfirmationPolicyEnforcementQuery"/>.
+    /// </summary>
+    /// <remarks>
+    /// This record is used to encapsulate the data required for handling the automatic confirmation policy enforcement.
+    ///
+    /// </remarks>
+    /// <param name="organizationUserToValidate">The organization user to be validated within the current organization context.</param>
+    /// <param name="organizationUsersForOtherOrganizations">THIS MUST BE POPULATED CORRECTLY. A collection of organization user records that match the provided user.</param>
```
eliykat: I agree with Claude that the "shouting documentation" (lol) indicates a fragile API. Asking the caller to pre-filter their data like this puts too much on them.

Either:
- the validator could fetch its own data - although I assume this probably duplicates some db calls, which you want to avoid
- the caller could pass in _all_ organizationUsers associated with that user - including the one to be confirmed. This doesn't make much change to the validator implementation but means the caller can just pass the retrieved data along with less work.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/AutomaticUserConfirmationPolicyRequirement.cs
```diff
@@ -0,0 +1,43 @@
...
+/// <li>Cannot be a provider</li>
+/// </ul>
+/// </remarks>
+/// <param name="policyDetails">Collection of policy details that apply to this user id</param>
+public class AutomaticUserConfirmationPolicyRequirement(IEnumerable<PolicyDetails> policyDetails) : IPolicyRequirement
+{
+    public bool IsEnabled(Guid organizationId) => policyDetails.Any(p => p.OrganizationId == organizationId);
+
+    public bool IsEnabledAndUserIsAProvider(Guid organizationId) =>
+        policyDetails.Any(p => p.OrganizationId == organizationId && p.IsProvider);
```
eliykat: I don't think this quite works:

- returns true if user is already a providerUser for the current organization, blocking acceptance ✅ 
- returns true if user is already a providerUser for _another_ organization, blocking acceptance ❌ 

I think you need a separate db call to fetch all providerUsers for the user, like you do in the `AutomaticallyConfirmOrganizationUserValidator`. This would probably be evaluated in the enforcement query rather than here.

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
eliykat, line 1: Are these changes necessary if we're now calling the new AutoConfirmPolicyRequirement in all these flows? This is effectively modifying Single Org behaviour which isn't the approach in the rest of this PR. I may be missing something though - are there places where this `policyService` method is called and the new policyRequirement/enforce query is not?

jrmccannon:
@eliykat Question for PolicyService:

Should we throw exceptions or call the requirement [here](https://github.com/bitwarden/server/blob/18a8829476230649e5b981a500b737c56a5ae7be/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L77) or [here](https://github.com/bitwarden/server/blob/18a8829476230649e5b981a500b737c56a5ae7be/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L83)?

Anyone who calls those methods with the Auto Confirm policy type would not get the correct answer. Hopefully we would be involved if anyone is checking for auto confirm, but should we throw and inform them of the requirement or enforcement classes or just call the requirement?

I don't really think we have the context to answer the question they're attempting to solve for in the future so throwing might be a safe bet.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
jrmccannon: I agree. I've now isolated single org from auto confirm as much as possible.

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
jrmccannon: They are not, but I have a question about what happens when someone later tries to call PolicyService with Auto Confirm policy type?

Should we throw exceptions or call the requirement [here](https://github.com/bitwarden/server/blob/18a8829476230649e5b981a500b737c56a5ae7be/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L77) or [here](https://github.com/bitwarden/server/blob/18a8829476230649e5b981a500b737c56a5ae7be/src/Core/AdminConsole/Services/Implementations/PolicyService.cs#L83)?

Anyone who calls those methods with the Auto Confirm policy type would not get the correct answer. Hopefully we would be involved if anyone is checking for auto confirm, but should we throw and inform them of the requirement or enforcement classes or just call the requirement?

I don't really think we have the context to answer the question they're attempting to solve for in the future so throwing might be a safe bet.

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -299,6 +311,19 @@ public async Task<List<Tuple<ProviderUser, string>>> ConfirmUsersAsync(Guid prov
...
+                    var organizationAutoConfirmPolicyRequirement = await _policyRequirementQuery
+                        .GetAsync<AutomaticUserConfirmationPolicyRequirement>(user.Id);
+
+                    if (organizationAutoConfirmPolicyRequirement
+                        .UserBelongsToOrganizationWithAutomaticUserConfirmationEnabled())
+                    {
+                        result.Add(Tuple.Create(providerUser, new ProviderUsersCannotJoin().Message));
+                        continue;
+                    }
+                }
```
eliykat, line 338: This database call is in a loop. We have a bulk interface for `IPolicyRequirementQuery` you can use to get all of them upfront.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
eliykat: The logic here looks good to me, but please add a comment explaining why we're not checking single org. (basically my explanation above.) I ask because this is a difference between the autoconfirm flow and the regular flow, it may otherwise look like an oversight.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/AutomaticallyConfirmOrganizationUsersValidator.cs
```diff
@@ -98,19 +99,18 @@ private async Task<bool> OrganizationUserConformsToTwoFactorRequiredPolicyAsync(
             return null;
         }
```
eliykat: This early return could be moved inside the `automaticUserConfirmationPolicyEnforcementValidator` to better encapsulate it.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/AutomaticUserConfirmationPolicyRequirement.cs
```diff
@@ -19,10 +19,11 @@ namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements
 /// <param name="policyDetails">Collection of policy details that apply to this user id</param>
 public class AutomaticUserConfirmationPolicyRequirement(IEnumerable<PolicyDetails> policyDetails) : IPolicyRequirement
 {
-    public bool IsEnabled(Guid organizationId) => policyDetails.Any(p => p.OrganizationId == organizationId);
+    public bool CanBeGrantedEmergencyAccess() => policyDetails.Any();
+
+    public bool UserBelongsToOrganizationWithAutomaticUserConfirmationEnabled() => policyDetails.Any();
```
eliykat: `IPolicyRequirement`s work best when they express what the policy requires in business logic terms. This isn't always possible, but here this could be (for example) `CanJoinProvider()`. Then that's the interface that the `ProviderService` calls.

src/Core/AdminConsole/OrganizationFeatures/Organizations/CloudOrganizationSignUpCommand.cs
```diff
@@ -236,6 +240,17 @@ private static void ValidatePlan(Plan plan, int additionalSeats, string productT
 
     private async Task ValidateSignUpPoliciesAsync(Guid ownerId)
     {
+        if (featureService.IsEnabled(FeatureFlagKeys.AutomaticConfirmUsers))
+        {
+            var requirement = await policyRequirementQuery.GetAsync<AutomaticUserConfirmationPolicyRequirement>(ownerId);
+
+            if (requirement.UserBelongsToOrganizationWithAutomaticUserConfirmationEnabled())
```
eliykat: Similarly, here this interface could be `CanCreateNewOrganization()`.

The idea being that this command doesn't need to know about org membership and policies. Only that this requirement doesn't block what the user is trying to do.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
```diff
@@ -196,19 +196,19 @@ private async Task CheckPoliciesAsync(Guid organizationId, User user,
 
         if (_featureService.IsEnabled(FeatureFlagKeys.AutomaticConfirmUsers))
         {
-            var error = (await _automaticUserConfirmationPolicyEnforcementQuery.IsCompliantAsync(
+            var error = (await _automaticUserConfirmationPolicyEnforcementValidator.IsCompliantAsync(
```
eliykat, line 199: This could also use the bulk db query - this method is called in a loop. (Although it currently has looped db queries in it 😭 let's not make it worse if we can help)

src/Core/AdminConsole/Services/Implementations/PolicyService.cs
eliykat: That's a fair point. Given how we are treating it here - as a separate policy that should be checked separately - I think the answer is that the future caller should know to check autoconfirm as well, and this method goes unchanged.

That is maybe fragile. But the idea is that we now have more policies that affect these flows, Autoconfirm becomes just another policy that should be checked when a user tries to do an action affected by the policy. The caller needs to know to check it, but that's the case with all policies. (I think this is an argument for having another layer of abstraction for policy enforcement like we've discussed, but out of scope here.)

Also note that I am going to return to the PolicyRequirements migration in the next few sprints so we can remove this old code - which should help with PolicyService confusion: https://bitwarden.atlassian.net/browse/PM-28958

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
jrmccannon: That bulk interface is using the OrganizationId which we don't have at this point.  We would need another method to take a list of user Ids or organization Ids (from the org users retrieved above) and then go to the database with that.

eliykat: Ah that's a good point. We can leave it for now then.

src/Core/AdminConsole/OrganizationFeatures/Policies/Enforcement/AutoConfirm/AutomaticUserConfirmationPolicyEnforcementValidator.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        if ((await providerUserRepository.GetManyByUserAsync(request.User.Id)).Count != 0)
+        {
+            return Invalid(request, new ProviderUsersCannotJoin());
+        }
+
+        if (request.AllOrganizationUsers.Count == 1)
+        {
+            return Valid(request);
+        }
```
eliykat: You're missing a check that the policy applies to this user for the current org. e.g. `automaticUserConfirmationPolicyRequirement.IsEnabled(request.OrganizationId)`. (bonus points if you can get away from the `IsEnabled` wording but whatever works.)

As it is currently written these requirements will be enforced whether or not the policy is enabled.

bitwarden_license/src/Commercial.Core/AdminConsole/Services/ProviderService.cs
```diff
@@ -89,6 +94,7 @@ public ProviderService(IProviderRepository providerRepository, IProviderUserRepo
         _providerBillingService = providerBillingService;
         _pricingClient = pricingClient;
         _providerClientOrganizationSignUpCommand = providerClientOrganizationSignUpCommand;
+        _policyRequirementQuery = policyRequirementQuery;
     }
 
     public async Task<Provider> CompleteSetupAsync(Provider provider, Guid ownerUserId, string token, string key, TokenizedPaymentMethod paymentMethod, BillingAddress billingAddress)
```
eliykat, line 101: This also needs a guard to prevent a user from setting up a new provider.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/Policies/Enforcement/AutoConfirm/AutomaticUserConfirmationPolicyEnforcementValidator.cs
jrmccannon: For this one, I think we do actually mean IsEnabled

eliykat: I agree, sometimes it is just easier to think about in enabled/disabled terms.

src/Core/AdminConsole/OrganizationFeatures/Policies/Enforcement/AutoConfirm/AutomaticUserConfirmationPolicyEnforcementValidator.cs
```diff
@@ -26,14 +26,17 @@ public async Task<ValidationResult<AutomaticUserConfirmationPolicyEnforcementReq
...
-            return Valid(request);
+            if ((await providerUserRepository.GetManyByUserAsync(request.User.Id)).Count != 0)
+            {
+                return Invalid(request, new ProviderUsersCannotJoin());
+            }
+
+            if (request.AllOrganizationUsers.Count == 1)
+            {
+                return Valid(request);
+            }
```
eliykat: This needs to be inverted, otherwise if `AllOrganizationUsers.Count > 1`, it will fall through to returning `Valid` at the end of the method. It's also generally clearer to check the prohibited cases.

```c#
            if (request.AllOrganizationUsers.Any(ou => ou !== currentOrganizationUser))
            {
                return Invalid(request, new WhateverYourErrorMessageIs());
            }
```

eliykat (changes requested):
One last thing I spotted.

eliykat (approved):
:shipit:

eliykat (approved):

PR #6650, opened 2025-11-26
Use extended cache for caching integration configuration details

brant-livefront:
## 📔 Objective

This PR moves the most high-traffic piece of event integrations over to the new extended cache. This part is the processing that happens every time when receive an event. We need to fetch the integrations from the database and determine if an integration should be fire. Previously, we had a home-grown database that would load all integrations into memory on a fixed interval. This PR replaces that with extended cache.

There are a number of benefits to having this in extended cache:
- We don't need to pre-load any data. We can fetch from the database as events first come in (with stampede protection) and the results will exist in the cache
- Due to extended cache's redis support, that newly cached item will be in redis for other nodes that see the same event type / organization / integration type later. 
- As a result of the backplane, we can invalidate the cache (flushing it from L2 / redis) as soon as an admin changes any integration or integration configurations and this will be propagated to all nodes (flushing their L1 / memory cache). The next event that would need this will then load the brand new data on first request and refill the L1 / L2 caches.
- Since we no longer need to worry about stale data, we can use a very long (1 day) TTL on these records which allows us to lean even more heavily on cache.
- We are relying on known implementations and don't need to worry about maintaining our own hand-coded in-memory cache.

In addition I've cleaned up the docs and fixed an error that I noticed while testing this (Trying to create an integration of the same type as an existing one would 500 error. It now returns a `BadRequestException` - and we have a new test to validate that)

Finally, I have updated the method / stored procedure / query for fetching configurations. It now includes configurations where EventType is `null`, which match all events. So asking for all the configurations that match a specific event type, integration type and organization id will retrieve:
1. Those specific rows that have that event type explicitly (this is the same as before)
2. Any rows that match integration type and organization id, but have a null event type (this is the new addition). Again, nulls are considered as wildcards so they are now included here.

**Note**: I've moved the changes for Controllers (the _write_ side of this PR) into 2 separate PRs since the changes were large. They will follow this PR shortly.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
```diff
@@ -132,6 +123,49 @@ internal async Task<IntegrationTemplateContext> BuildContextAsync(EventMessage e
...
+            key: EventIntegrationsCacheConstants.BuildCacheKeyForOrganizationIntegrationConfigurationDetails(
+                organizationId: organizationId,
+                integrationType: integrationType,
+                eventType: eventMessage.Type),
+            factory: async _ => await configurationRepository.GetConfigurationDetailsAsync(
+                organizationId: organizationId,
+                integrationType: integrationType,
+                eventType: eventMessage.Type
+            ),
+            options: new FusionCacheEntryOptions(duration: TimeSpan.FromDays(1)),
```
withinfocus: 🎨 Cache constant for all these.

util/Migrator/DbScripts/2025-11-21_00_OrganizationIntegrationConfigurationDetails_ReadManyWildcardByOrganizationIdIntegrationType.sql
```diff
@@ -0,0 +1,19 @@
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationIntegrationConfigurationDetails_ReadManyWildcardByOrganizationIdIntegrationType]
```
mkincaid-bw: Is this stored procedure name appropriate? With 3 predicates in the where clause, will there be more than 1 result per `OrganiztionId` and `IntegrationType`? Also, in `ReadManyWildcard`, we don't use `Wildcard` anywhere else in the DB. It's fine if it's applicable, just haven't seen it before.

withinfocus: I was good with the name when I first read it to make the openness clear, but yeah we can probably just get rid of "wildcard". Yes, there will be many.

brant-livefront: Yeah, I wasn't 100% sold on "Wildcard" either. I landed on that to call out that we're fetching ones where EventType is explicitly null, which we treat as a wildcard event - they match **all** events. I'd be up for dropping it if we think it's misleading.

withinfocus: Yeah, just get rid of it.

brant-livefront: Latest commit has a refactor to remove references to "wildcard". 👍

brant-livefront:
Claude's issues 1 and 5 are both addressed in the follow on PR for this: #6675 

The other Claude comments I'm fine with as-is, but let me know if anyone has concerns.

brant-livefront:
@withinfocus et al - I've pulled all of the Controller / API related changes out of this PR. I'll put those up (with the CQRS changes) in two separate PRs. But I wanted to slim this one down to just the DB changes, the EventIntegrationHandler changes and some fixes / supporting changes for those.

util/Migrator/DbScripts/2025-11-21_00_OrganizationIntegrationConfigurationDetails_ReadManyConfigurationDetailsByOrganizationIdIntegrationType.sql
```diff
@@ -0,0 +1,19 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON
+
+SELECT
+    oic.*
+FROM
+    [dbo].[OrganizationIntegrationConfigurationDetailsView] oic
+WHERE
+    oic.[EventType] IS NULL
```
withinfocus: ❌ This is a hidden detail that isn't revealed in any of the app logic naming. Regardless, we can optimize the whole process and depend on this new query (with this line removed) for rows with or without an event type, and then the naming will be accurate.

src/Core/AdminConsole/Services/Implementations/EventIntegrations/EventIntegrationHandler.cs
```diff
@@ -132,6 +123,51 @@ internal async Task<IntegrationTemplateContext> BuildContextAsync(EventMessage e
...
+            organizationId,
+            integrationType
+        );
+
+        configurations.AddRange(await cache.GetOrSetAsync<List<OrganizationIntegrationConfigurationDetails>>(
+            key: EventIntegrationsCacheConstants.BuildCacheKeyForOrganizationIntegrationConfigurationDetails(
+                organizationId: organizationId,
+                integrationType: integrationType,
+                eventType: eventMessage.Type),
+            factory: async _ => await configurationRepository.GetConfigurationDetailsAsync(
```
withinfocus: 🎨 This repository method can be renamed to be more clear that it's for the three parameters. Wordy but clear.

brant-livefront:
I spoke with @withinfocus and we've altered the strategy a bit, and the latest commit executes on this direction. Instead of making 2 calls to the cache (one to fetch matching records that explicitly match the event type and a second call to fetch the records that match with a null event type) we are simply combining this into one call. The net-net is I've removed all of my new stored procedure / method / query setup (which fetched only null event types). Instead, this PR now updates the existing stored procedure to include the "wildcard" matches that have null event type. So that becomes the one call we are making to the cache and then to the repository.

Before:
1. Call to retrieve specific event type matches
2. Call to retrieve null event type matches

After:
1. Call to retrieve configs that match via direct event type match _and_ and configs that match via null event type wildcard match

This is obviously effect the cache clearing, but that is being handled in the PRs that are dependent on this. I'm making those updates next.

PR #6653, opened 2025-11-27
[BEEEP] [PM-28808] Fix invalid identity URL in Swagger

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28808

## 📔 Objective

We have a swagger.json and SwaggerUI hosted on our Help Center for the public API: https://bitwarden.com/help/api/.

However it currently has the following issues:
- only supports targeting the US server - not EU
- the identity endpoint is set to `localhost` - so the Authorize button doesn't work and you can't use the "Try it out" functionality

This PR makes the following changes:
- in SwaggerGen configuration - configure US and EU servers and corresponding authorization options. This is the default configuration that will be used to produce the `swagger.json` build artifact, which is used by the Help Center.
  - note: you cannot automatically associate the server with the authorization option (or otherwise parameterize it etc) so the best we can do is try to clearly explain which authorization to use
  - this should be permitted by CORS because our cloud servers permit requests from the product website: [ref](https://github.com/bitwarden/server/blob/main/src/Core/Utilities/CoreHelpers.cs#L659).
- in the API project Swagger configuration - override the default config with the details of the locally running server. This will be used for development servers and self-host servers.

We used to specify the `--host` option in `generate_openapi_files.ps1` but that is no longer needed (and in any case didn't work because you can't specify the authorization config from the command line).

I think this approach is reasonable but I am open to other ways to achieve this. Platform should probably be the one to review this.

## 📸 Screenshots

Preview of how it should look on the Help Center:

https://github.com/user-attachments/assets/24d4b3e3-c1ae-4dff-9fb1-870815bb4da1

Preview of how it should look in dev and self hosted:

https://github.com/user-attachments/assets/08f3ae78-f0d4-46a9-87d8-26fdad3caec2

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -46,36 +51,20 @@ [this document](https://bitwarden.com/help/vault-management-api/).
...
-                            Id = "oauth2-client-credentials"
-                        },
-                    },
-                    new[] { ApiScopes.ApiOrganization }
-                }
-            });
+            // Configure Bitwarden cloud US and EU servers. These will appear in the swagger.json build artifact
+            // used for our help center. These are overwritten with the local server when running in self-hosted
+            // or dev mode (see Api Startup.cs).
+            config.AddSwaggerServerWithSecurity(
```
dani-garcia, line 57: I wonder if it wouldn't be easier to change this code to be:
```cs
if (globalSettings.SelfHosted)
{
    // Configure self-hosted server
    config.AddSwaggerServerWithSecurity(
        serverId: "self_hosted_server",
        serverUrl: globalSettings.BaseServiceUri.Api,
        identityTokenUrl: $"{globalSettings.BaseServiceUri.Identity}/connect/token",
        serverDescription: "Self-hosted");
}
else
{
    // Configure Bitwarden cloud US and EU servers. These will appear in the swagger.json build artifact
    // used for our help center. These are overwritten with the local server when running in self-hosted
    // or dev mode (see Api Startup.cs).
    config.AddSwaggerServerWithSecurity(
        serverId: "US_server",
        serverUrl: "https://api.bitwarden.com",
        identityTokenUrl: "https://identity.bitwarden.com/connect/token",
        serverDescription: "US server");

    config.AddSwaggerServerWithSecurity(
        serverId: "EU_server",
        serverUrl: "https://api.bitwarden.eu",
        identityTokenUrl: "https://identity.bitwarden.eu/connect/token",
        serverDescription: "EU server");
}
```

And then remove the manual modification of the servers that is currently being done using `config.PreSerializeFilters` in `Startup.Cs`. Seems to generate the same servers in the schema when run locally.

eliykat:
Thanks @dani-garcia ! I did mean to ask about how this would affect the SDK so thank you for checking that.

I tried your suggestion, but the pipeline generates the swagger jsons using the [development environment configuration](https://github.com/bitwarden/server/blob/main/dev/generate_openapi_files.ps1#L3) which means it outputs the `localhost` configuration. If I change that to using Production, it fails because no server certificates are configured; using the Development environment seems like a way to bypass the configuration required for a real server. Maybe that in itself needs to be revisited (do we need a CI environment type? not sure) but I thought that was out of scope for this PR.

dani-garcia (approved):

PR #6663, opened 2025-12-01
[PM-28662] Fix Individual Premium automatically disabled due to duplicate subscription leftover from failed payment

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28662

## 📔 Objective

Recent behavior (past few weeks)
When a user subscribes to Premium and the initial payment fails, the system temporarily enables Premium and leaves the invoice open. Within 1–2 days, the subscription automatically cancels.

As a result, users commonly:
Attempt to add Account Credit (which does not apply to the open invoice), or

Retry the subscribe flow, which creates a new subscription

This behavior is related to PM-28357: New Premium Membership failing due to charge attempt with old payment method instead of new payment info.
Needs Triage
, where checkout attempts sometimes charge the old default payment method instead of the new card entered during checkout.

This often leads to multiple attempts where the account toggles between enabled and disabled states because of the active/canceled subscriptions.

New behavior (starting this weekend, may be around Saturday)
A new bug is occurring where:
The initial subscription from a failed payment (possibly successful as well for 1 case) remains on the account as a duplicate subscription instead of auto-cancelling.

A new subscription is created when the user retries checkout and becomes the primary subscription.

Despite the successful new subscription, the user’s reaches out to us about their Premium status being disabled.

## 📸 Screenshots

https://github.com/user-attachments/assets/5652bbdf-e088-4e03-8bf3-8d6796056845

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


amorask-bitwarden (commented):
Can you add a video of this process working? I can't determine whether or not a subscription would just immediately move to `incomplete_expired` after a failed payment attempt.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -111,7 +111,7 @@ public async Task HandleAsync(Event parsedEvent)
                         break;
                     }
 
-                    if (subscription.Status is StripeSubscriptionStatus.Unpaid &&
+                    if ((subscription.Status is StripeSubscriptionStatus.Unpaid or StripeSubscriptionStatus.IncompleteExpired) &&
```
amorask-bitwarden: Claude is correct here in that this check is now redundant as this code block could not be hit without the status being either `unpaid` or `incomplete_expired`. 

Additionally, this code is checking for a hard-coded premium price ID. We want to change this to use the `PricingClient` to list the premium plans and check if any of the subscription item Price IDs match the premium price IDs.

PR #6669, opened 2025-12-02
Add UpdateAccountCryptographicState repository function

quexten:
## 📔 Objective

Unblock v2 account registration related tasks.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Infrastructure.Dapper/Repositories/UserRepository.cs
+++ b/src/Infrastructure.Dapper/Repositories/UserRepository.cs
@@ -288,7 +288,10 @@ public class UserRepository : Repository<User, Guid>, IUserRepository
         UnprotectData(user);
     }
 
-    public async Task SetV2AccountCryptographicStateAsync(Guid userId, UserAccountKeysData accountKeysData)
+    public async Task SetV2AccountCryptographicStateAsync(
+        Guid userId,
+        UserAccountKeysData accountKeysData,
+        IEnumerable<UpdateUserData>? updateUserDataActions = null)
     {
         if (!accountKeysData.IsV2Encryption())
         {
@@ -299,27 +302,47 @@ public class UserRepository : Repository<User, Guid>, IUserRepository
         var signatureKeyPairId = CoreHelpers.GenerateComb();
 
         await using var connection = new SqlConnection(ConnectionString);
-        await using var cmd = new SqlCommand("[dbo].[User_UpdateAccountCryptographicState]", connection);
-        cmd.CommandType = CommandType.StoredProcedure;
-        cmd.Parameters.Add("@Id", SqlDbType.UniqueIdentifier).Value = userId;
-        cmd.Parameters.Add("@PublicKey", SqlDbType.NVarChar).Value = accountKeysData.PublicKeyEncryptionKeyPairData.PublicKey;
-        cmd.Parameters.Add("@PrivateKey", SqlDbType.NVarChar).Value = accountKeysData.PublicKeyEncryptionKeyPairData.WrappedPrivateKey;
-        cmd.Parameters.Add("@SignedPublicKey", SqlDbType.NVarChar).Value =
-            accountKeysData.PublicKeyEncryptionKeyPairData.SignedPublicKey;
-        cmd.Parameters.Add("@SecurityState", SqlDbType.NVarChar).Value =
-            accountKeysData.SecurityStateData!.SecurityState;
-        cmd.Parameters.Add("@SecurityVersion", SqlDbType.Int).Value =
-            accountKeysData.SecurityStateData!.SecurityVersion;
-        cmd.Parameters.Add("@SignatureKeyPairId", SqlDbType.UniqueIdentifier).Value = signatureKeyPairId;
-        cmd.Parameters.Add("@SignatureAlgorithm", SqlDbType.TinyInt).Value = accountKeysData.SignatureKeyPairData!.SignatureAlgorithm;
-        cmd.Parameters.Add("@SigningKey", SqlDbType.VarChar).Value =
-            accountKeysData.SignatureKeyPairData!.WrappedSigningKey;
-        cmd.Parameters.Add("@VerifyingKey", SqlDbType.VarChar).Value =
-            accountKeysData.SignatureKeyPairData!.VerifyingKey;
-        cmd.Parameters.Add("@RevisionDate", SqlDbType.DateTime2).Value = timestamp;
-        cmd.Parameters.Add("@AccountRevisionDate", SqlDbType.DateTime2).Value = timestamp;
         await connection.OpenAsync();
-        await cmd.ExecuteNonQueryAsync();
+
+        await using var transaction = connection.BeginTransaction();
+        try
+        {
+            await connection.ExecuteAsync(
+                "[dbo].[User_UpdateAccountCryptographicState]",
+                new
+                {
+                    Id = userId,
+                    PublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.PublicKey,
+                    PrivateKey = accountKeysData.PublicKeyEncryptionKeyPairData.WrappedPrivateKey,
+                    SignedPublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.SignedPublicKey,
+                    SecurityState = accountKeysData.SecurityStateData!.SecurityState,
+                    SecurityVersion = accountKeysData.SecurityStateData!.SecurityVersion,
+                    SignatureKeyPairId = signatureKeyPairId,
+                    SignatureAlgorithm = accountKeysData.SignatureKeyPairData!.SignatureAlgorithm,
+                    SigningKey = accountKeysData.SignatureKeyPairData!.WrappedSigningKey,
+                    VerifyingKey = accountKeysData.SignatureKeyPairData!.VerifyingKey,
+                    RevisionDate = timestamp,
+                    AccountRevisionDate = timestamp
+                },
+                transaction: transaction,
+                commandType: CommandType.StoredProcedure);
+
+            //  Update user data that depends on cryptographic state
+            if (updateUserDataActions != null)
+            {
+                foreach (var action in updateUserDataActions)
+                {
+                    await action(connection, transaction);
+                }
+            }
+
+            await transaction.CommitAsync();
+        }
+        catch
+        {
+            await transaction.RollbackAsync();
+            throw;
+        }
     }
 
     public async Task<IEnumerable<User>> GetManyAsync(IEnumerable<Guid> ids)
--- a/src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
+++ b/src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
@@ -242,7 +242,10 @@ public class UserRepository : Repository<Core.Entities.User, User, Guid>, IUserR
         await transaction.CommitAsync();
     }
 
-    public async Task SetV2AccountCryptographicStateAsync(Guid userId, UserAccountKeysData accountKeysData)
+    public async Task SetV2AccountCryptographicStateAsync(
+        Guid userId,
+        UserAccountKeysData accountKeysData,
+        IEnumerable<UpdateUserData>? updateUserDataActions = null)
     {
         if (!accountKeysData.IsV2Encryption())
         {
@@ -301,6 +304,15 @@ public class UserRepository : Repository<Core.Entities.User, User, Guid>, IUserR
         }
 
         await dbContext.SaveChangesAsync();
+
+        // Update additional user data within the same transaction
+        if (updateUserDataActions != null)
+        {
+            foreach (var action in updateUserDataActions)
+            {
+                await action();
+            }
+        }
         await transaction.CommitAsync();
     }
 
--- a/src/Sql/dbo/Stored Procedures/User_UpdateAccountCryptographicState.sql
+++ b/src/Sql/dbo/Stored Procedures/User_UpdateAccountCryptographicState.sql
@@ -15,62 +15,51 @@ AS
 BEGIN
     SET NOCOUNT ON
 
-    BEGIN TRANSACTION
+    UPDATE
+        [dbo].[User]
+    SET
+        [PublicKey] = @PublicKey,
+        [PrivateKey] = @PrivateKey,
+        [SignedPublicKey] = @SignedPublicKey,
+        [SecurityState] = @SecurityState,
+        [SecurityVersion] = @SecurityVersion,
+        [RevisionDate] = @RevisionDate,
+        [AccountRevisionDate] = @AccountRevisionDate
+    WHERE
+        [Id] = @Id
 
-    BEGIN TRY
-        UPDATE
-            [dbo].[User]
+    IF EXISTS (SELECT 1 FROM [dbo].[UserSignatureKeyPair] WHERE [UserId] = @Id)
+    BEGIN
+        UPDATE [dbo].[UserSignatureKeyPair]
         SET
-            [PublicKey] = @PublicKey,
-            [PrivateKey] = @PrivateKey,
-            [SignedPublicKey] = @SignedPublicKey,
-            [SecurityState] = @SecurityState,
-            [SecurityVersion] = @SecurityVersion,
-            [RevisionDate] = @RevisionDate,
-            [AccountRevisionDate] = @AccountRevisionDate
+            [SignatureAlgorithm] = @SignatureAlgorithm,
+            [SigningKey] = @SigningKey,
+            [VerifyingKey] = @VerifyingKey,
+            [RevisionDate] = @RevisionDate
         WHERE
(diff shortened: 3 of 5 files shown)


src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+        user.RevisionDate = user.AccountRevisionDate = DateTime.UtcNow;
+        // Update the signature key pair data
+        if (accountKeysData.SignatureKeyPairData != null && accountKeysData.SecurityStateData != null)
+        {
+            user.SignedPublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.SignedPublicKey;
+            user.SecurityState = accountKeysData.SecurityStateData.SecurityState;
+            user.SecurityVersion = accountKeysData.SecurityStateData.SecurityVersion;
+            await _userSignatureKeyPairRepository.UpsertAsync(new UserSignatureKeyPair
+            {
+                Id = CoreHelpers.GenerateComb(),
```
mzieniukbw: With the `Id` set, this will trigger Update, instead of Create / Insert, is it intended ?
Note: when Id is omitted, upsert would trigger Create and set the Id automatically.

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+        user.PrivateKey = accountKeysData.PublicKeyEncryptionKeyPairData.WrappedPrivateKey;
+        user.PublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.PublicKey;
+        user.RevisionDate = user.AccountRevisionDate = DateTime.UtcNow;
+        // Update the signature key pair data
+        if (accountKeysData.SignatureKeyPairData != null && accountKeysData.SecurityStateData != null)
+        {
+            user.SignedPublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.SignedPublicKey;
+            user.SecurityState = accountKeysData.SecurityStateData.SecurityState;
+            user.SecurityVersion = accountKeysData.SecurityStateData.SecurityVersion;
+            await _userSignatureKeyPairRepository.UpsertAsync(new UserSignatureKeyPair
```
mzieniukbw: This will fail for Dapper, since the `IUserSignatureKeyPairRepository` implementation does not have`UserSignatureKeyPair_Create`, nor `UserSignatureKeyPair_Update` stored procedures.

mzieniukbw: Since we want to use this for user registration for now, it would make sense to call CreateAsync directly.

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+        _userRepository = userRepository;
+        _userSignatureKeyPairRepository = userSignatureKeyPairRepository;
+    }
+
+    public async Task SetAccountKeysForUserAsync(Guid userId, AccountKeysRequestModel accountKeys)
+    {
+        var user = await _userRepository.GetByIdAsync(userId);
+        if (user == null)
+        {
+            throw new ArgumentException("User not found", nameof(userId));
```
mzieniukbw: While this is true for SSO JIT, user registration with master password will not have user created yet (check https://github.com/bitwarden/server/blob/5fb69e42b08f49083a39729705b9b38fa6e1d8c4/src/Identity/Controllers/AccountsController.cs#L157).
Unless the intention is to ensure that user is created first without the keys, and then the keys are updated with this command. Does not feel great.

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+        IUserRepository userRepository,
+        IUserSignatureKeyPairRepository userSignatureKeyPairRepository)
+    {
+        _userRepository = userRepository;
+        _userSignatureKeyPairRepository = userSignatureKeyPairRepository;
+    }
+
+    public async Task SetAccountKeysForUserAsync(Guid userId, AccountKeysRequestModel accountKeys)
+    {
+        var user = await _userRepository.GetByIdAsync(userId);
```
mzieniukbw: 🤔 For the case where user is created, maybe it would make more sense to provide `User user` instead of Guid ?

src/Core/KeyManagement/Commands/Interfaces/ISetAccountKeysForUserCommand.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Bit.Core.KeyManagement.Models.Api.Request;
+
+namespace Bit.Core.KeyManagement.Commands.Interfaces;
+
+public interface ISetAccountKeysForUserCommand
```
mzieniukbw: some docs would be useful

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
quexten: I have changed this to CreateAsync, and added docs clarifying this command should only be used for setting the initial keys, not updating.

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
quexten: As mentioned in the other comment, I've now updated the comment and the command is now meant for just setting initial keys, so JIT password / regular MP have to find a different solution / write their own command that atomically sets all their data.

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
quexten: Added an SP for creation

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
quexten: Makes sense. Updated to take a user. I also added a comment since the user that is passed in does get mutated in the process, which may be unintuitive.

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
```diff
@@ -0,0 +1,50 @@
...
+                Id = CoreHelpers.GenerateComb(),
+                UserId = user.Id,
+                SignatureAlgorithm = accountKeysData.SignatureKeyPairData.SignatureAlgorithm,
+                SigningKey = accountKeysData.SignatureKeyPairData.WrappedSigningKey,
+                VerifyingKey = accountKeysData.SignatureKeyPairData.VerifyingKey,
+                CreationDate = DateTime.UtcNow,
+                RevisionDate = DateTime.UtcNow,
+            });
+        }
+        await _userRepository.ReplaceAsync(user);
```
Thomas-Avery: ⚠️ 

These DB writes aren't atomic. For v2 users if the `await _userRepository.ReplaceAsync(user);` fails we have written data into the DB from the `await _userSignatureKeyPairRepository.CreateAsync(` and won't roll it back.

I would recommend creating atomic repository calls. `await _userRepository.ReplaceAsync(user);` can be kept for v1 processing and a new one that handles all v2 in one transaction.

Thomas-Avery (changes requested):

src/Core/KeyManagement/Commands/SetAccountKeysForUserCommand.cs
quexten: @Thomas-Avery Can we merge as-is, and then fix this before rolling out any of the flows? The reason I'm suggesting that is because that will unblock the follow-up work quicker. The request models in each of the subsequent server PR need to be merged in order for us to be able to update the SDK models, thus blocking the SDK work.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -241,6 +242,66 @@ public async Task UpdateUserKeyAndEncryptedDataV2Async(Core.Entities.User user,
...
+
+        userEntity.RevisionDate = user.RevisionDate;
+        userEntity.AccountRevisionDate = user.AccountRevisionDate;
+
+        // V1 + V2 user crypto changes
+        userEntity.PublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.PublicKey;
+        userEntity.PrivateKey = accountKeysData.PublicKeyEncryptionKeyPairData.WrappedPrivateKey;
+
+        // V2 only changes
+        if (accountKeysData.IsV2Encryption())
```
Thomas-Avery: With the repository pattern it is discourage to do business logic like this inside the repository layer.  A write operation in the repository layer should be concerned with "how to persist" not "what to do". 

This applies to stored procedures more so where we are doing a similar v1 or v2 user check. We have a small snippet of guidance here https://contributing.bitwarden.com/contributing/code-style/sql#error-handling. 

> Business logic should not be in stored procedures, but there may be times when it makes sense to do error handling in other scripts (migrations, one-off data scrubs)


I would recommend closer to what you originally had in the command. A switch if a v1 user write to the user repository via `await _userRepository.ReplaceAsync(user);`. If a v2 user use this new`UpdateAccountCryptographicStateAsync` and make that only for v2 users. We try to make our repositories as "dumb" as possible because they can be challenging to test and debug. This also helps enforce separation of concerns.


Having this in a command will also make implementing a future state where we no longer want to support creating v1 users easier to implement and more readable of the intent. For this PR I also think another valid approach would be to just implement this repository method as v2 and have the registration commands we write handle v1 vs v2. Or even I think we can have v1 branch at the controller level right?

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -288,6 +287,46 @@ await connection.ExecuteAsync(
         UnprotectData(user);
     }
 
+    public async Task UpdateAccountCryptographicStateAsync(User user, UserAccountKeysData accountKeysData)
+    {
+        await using var connection = new SqlConnection(ConnectionString);
+        await connection.OpenAsync();
+
+        await using var transaction = await connection.BeginTransactionAsync();
```
Thomas-Avery: When we are just wanting to call one stored procedure we will have the transaction in the TSQL stored procedure.

We use the dapper transaction when we want to call multiple stored procedures together and roll them all back if one fails. 

For example deleting organization by ID.

Uses the generic delete repository for dapper without trasnactions https://github.com/bitwarden/server/blob/acc25293533c8da1b9a6d03e3a2270a18fb37a66/src/Infrastructure.Dapper/Repositories/Repository.cs#L86-L95

Has the transaction handling in the stored procedure. https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Stored%20Procedures/Organization_DeleteById.sql

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -241,6 +242,66 @@ public async Task UpdateUserKeyAndEncryptedDataV2Async(Core.Entities.User user,
...
+                existingKeyPair.SignatureAlgorithm = accountKeysData.SignatureKeyPairData!.SignatureAlgorithm;
+                existingKeyPair.SigningKey = accountKeysData.SignatureKeyPairData.WrappedSigningKey;
+                existingKeyPair.VerifyingKey = accountKeysData.SignatureKeyPairData.VerifyingKey;
+                existingKeyPair.RevisionDate = user.RevisionDate;
+            }
+            else
+            {
+                var newKeyPair = new UserSignatureKeyPair
+                {
+                    Id = Guid.NewGuid(),
```
Thomas-Avery:
```suggestion
                    Id = CoreHelpers.GenerateComb(),
```

⚠️  New Ids for tables need to use `CoreHelpers.GenerateComb()`.

Normally we would use the `newKeyPair.SetNewId();`.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -241,6 +242,66 @@ public async Task UpdateUserKeyAndEncryptedDataV2Async(Core.Entities.User user,
...
+        var userEntity = await dbContext.Users.FindAsync(user.Id);
+        if (userEntity == null)
+        {
+            throw new ArgumentException("User not found", nameof(user));
+        }
+
+        // Update public key encryption key pair
+
+        userEntity.RevisionDate = user.RevisionDate;
+        userEntity.AccountRevisionDate = user.AccountRevisionDate;
```
Thomas-Avery: 💭 

This will require callers to pass in an updated User model that has it's `RevisionDate` and `AccountRevisionDate` updated. But we are ignoring all of it's other properties and everything else is grabbed from the `accountKeysData`. This is unintuitive to me. I would suggest using a `(Guid userId, UserAccountKeysData accountKeysData)` parameters and grabbing the current UTC in the repository layer.


This parameter set better reads userId (target of operation) accountKeysData (the data to persist).

src/Sql/dbo/Stored Procedures/User_UpdateAccountCryptographicState.sql
```diff
@@ -0,0 +1,68 @@
...
+                [UserId],
+                [SignatureAlgorithm],
+                [SigningKey],
+                [VerifyingKey],
+                [CreationDate],
+                [RevisionDate]
+            )
+            VALUES
+            (
+                NEWID(),
```
Thomas-Avery: ⚠️  This needs to be `CoreHelpers.GenerateComb()` passed in.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -241,6 +242,66 @@ public async Task UpdateUserKeyAndEncryptedDataV2Async(Core.Entities.User user,
...
+        if (accountKeysData.IsV2Encryption())
+        {
+            userEntity.SecurityState = accountKeysData.SecurityStateData!.SecurityState;
+            userEntity.SecurityVersion = accountKeysData.SecurityStateData.SecurityVersion;
+            userEntity.SignedPublicKey = accountKeysData.PublicKeyEncryptionKeyPairData.SignedPublicKey;
+
+            // Replace existing keypair if it exists
+            var existingKeyPair = await dbContext.UserSignatureKeyPairs
+                .FirstOrDefaultAsync(x => x.UserId == user.Id);
+            if (existingKeyPair != null)
```
Thomas-Avery: 🎨  not required; informational

In the EF world we can make navigation properties. https://learn.microsoft.com/en-us/ef/core/modeling/relationships/navigations. This isn't required for this PR and would need some setup. I do think they can be helpful for some work though.

We can do things like 

```
var userEntity = await dbContext.Users.Include(x => x.UserSignatureKeyPair)
.Where(e => e.Id == user.Id).FirstOrDefaultAsync();

.....

var existingKeyPair = userEntity.UserSignatureKeyPair;
```

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
quexten: If `UpdateAccountCryptographicStateAsync` is only for V2 users, but we also have no type safety preventing callers from calling it with a V1 cryptographic state, what do we do in the failure case? Is it acceptable to just throw?

(In that case i'd also rather name it `SetV2AccountCryptographicStateAsync` to make that clear).

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
quexten: Updated to use setNewId

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
quexten: Applied this.

src/Sql/dbo/Stored Procedures/User_UpdateAccountCryptographicState.sql
quexten: Applied

src/Infrastructure.Dapper/Repositories/UserRepository.cs
quexten: Moved the transaction into the TSQL. It is unfortunate that these patterns differ between dapper and EF.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
quexten: @Thomas-Avery After contemplating, I don't think we need a command. The specific use-case for this change in the first place is purely V2 commands. So therefore, we are fine with just having the SetV2AccountCryptographicState function on the repos - without V1 support.

Thomas-Avery (approved):

PR #6677, opened 2025-12-03
[PM-28555] Create idempotent sproc for creating a My Items collection

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28555
## 📔 Objective

A user must only have a single My Items collection, however this is not enforced by db constraints, so it's particularly important that we enforce it at the time of creation. This implements a specific database method to create a My Items collection for a user, ensuring that:

- the Collection and CollectionUser entries are always created properly per business rules (rather than relying on the caller)
- duplicate collections are not created, because the db checks for it within the same transaction (TOCTOU)

The database changes required some trial-and-error to get a proper lock on the rows (using the `SERIALIZABLE` option) and to gracefully handle deadlocks. In particular, this required changes to system-wide EF configuration, so dbops please take a close look at this and let me know if there is a better / preferred way to achieve this.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
+++ b/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
@@ -10,7 +10,6 @@ using Bit.Core.Enums;
 using Bit.Core.Models.Data.Organizations;
 using Bit.Core.Models.Data.Organizations.OrganizationUsers;
 using Bit.Core.Repositories;
-using Bit.Infrastructure.EntityFramework.Models;
 using LinqToDB.Tools;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
--- a/util/MySqlMigrations/Migrations/20251226224144_DefaultCollectionOwnerIdColumn.cs
+++ b/util/MySqlMigrations/Migrations/20251226224144_DefaultCollectionOwnerIdColumn.cs
@@ -1,52 +1,50 @@
-﻿using System;
-using Microsoft.EntityFrameworkCore.Migrations;
+﻿using Microsoft.EntityFrameworkCore.Migrations;
 
 #nullable disable
 
-namespace Bit.MySqlMigrations.Migrations
+namespace Bit.MySqlMigrations.Migrations;
+
+/// <inheritdoc />
+public partial class DefaultCollectionOwnerIdColumn : Migration
 {
     /// <inheritdoc />
-    public partial class DefaultCollectionOwnerIdColumn : Migration
+    protected override void Up(MigrationBuilder migrationBuilder)
     {
-        /// <inheritdoc />
-        protected override void Up(MigrationBuilder migrationBuilder)
-        {
-            migrationBuilder.AddColumn<Guid>(
-                name: "DefaultCollectionOwnerId",
-                table: "Collection",
-                type: "char(36)",
-                nullable: true,
-                collation: "ascii_general_ci");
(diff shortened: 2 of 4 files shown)


eliykat:
@bitwarden/dept-dbops Can you please review my approach to locks/concurrency here and let me know if I'm on the right path? See the PR description for more info. My current approach has my tests working, but has broken every other instance where we're manually creating an EF transaction. I can fix those by explicitly using a strategy, but better to get some feedback before I spend more time on it.

eliykat:
@claude I disagree with "1. Missing @CollectionId Parameter in Dapper Implementation" - double check that you've read this correctly.

src/Core/Entities/Collection.cs
```diff
@@ -8,16 +8,66 @@ namespace Bit.Core.Entities;
...
+    /// </summary>
+    /// <remarks>
+    /// This is used to enforce uniqueness so that an OrganizationUser can only have 1 default collection
+    /// in a given organization.
+    /// It should NOT be used for any other purpose, used in the application code, or exposed to the front-end.
+    /// In particular, refer to the <see cref="CollectionUser"/> to evaluate user assignment and permissions.
+    /// Only populated for collections of type DefaultUserCollection.
+    /// Set to null when the OrganizationUser entry is deleted.
+    /// </remarks>
+    public Guid? DefaultCollectionOwner { get; set; }
```
r-tome:
```suggestion
    public Guid? DefaultCollectionOwnerId { get; set; }
```

util/Migrator/DbScripts/2025-12-20_00_Collection_AddDefaultCollectionOwnerAndConstraint.sql
r-tome, line 1: I think we need to add the new column to the `GROUP BY` in `Collection_ReadByIdWithPermissions` and `Collection_ReadByOrganizationIdWithPermissions`

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
```diff
@@ -821,6 +821,63 @@ public async Task UpsertDefaultCollectionsAsync(Guid organizationId, IEnumerable
...
+        {
+            // Collection already exists, return false
+            return false;
+        }
+    }
+
+    private static bool IsUniqueConstraintViolation(DbUpdateException ex)
+    {
+        // Check if the inner exception is a SqlException with error 2601 or 2627
+        return ex.InnerException is Microsoft.Data.SqlClient.SqlException sqlEx
```
r-tome: This only works for SQL, we should either add the other supported DB types or just return false for any `DbUpdateException`

src/Core/Entities/Collection.cs
eliykat: Done!

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
eliykat: I did this but I ended up removing the sproc from this PR. It became too large. I'm still figuring out the best order to do things in.

util/Migrator/DbScripts/2025-12-20_00_Collection_AddDefaultCollectionOwnerAndConstraint.sql
eliykat: This is a great catch!

eliykat:
I ended up going with the approach in https://github.com/bitwarden/server/pull/6791 instead.

PR #6682, opened 2025-12-03
Upgrade ExtendedCache to support non-Redis distributed cache

brant-livefront:
## 📔 Objective

In revising some of the documentation around extended cache, we realized that while `FusionCache` allows _any_ `IDistributedCache` to be used for L2 cache, our implementation was highly focused on just Redis (which is what we use in cloud, but is not required). This PR adds support for using shared and separate keyed distributed caches that are not Redis. 

Changes:
- Migrate setting from `UseSharedRedisCache` to `UseSharedDIstributedCache`
- If Redis string is present, we still re-use the multiplexer and create the backplane
- If Redis string is not present, it assumes that some other `IDistributedCache` is already registered. `FusionCache` will look for this service and use it if it's there (failing gracefully to just L1 / memory if not). In this case, there is no backplane support, due to the fact that Redis' pub/sub is necessary for backplane)
- Both of these scenarios work in shared or keyed scenarios depending on if `UseSharedDIstributedCache` is true.
- Added tests to cover all these new scenarios
- Updated xml docs and inline comments to be clearer about what is happening / expected

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Identity/IdentityServer/RequestValidators/CustomTokenRequestValidator.cs
+++ b/src/Identity/IdentityServer/RequestValidators/CustomTokenRequestValidator.cs
@@ -161,16 +161,16 @@ public class CustomTokenRequestValidator : BaseRequestValidator<CustomTokenReque
         }
 
         // Key connector data should have already been set in the decryption options
-        // for backwards compatibility we set them this way too. We can eventually get rid of this
-        // when all clients don't read them from the existing locations.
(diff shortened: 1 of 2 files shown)


withinfocus:
@claude we have an ADR describing this shift located at https://github.com/bitwarden/contributing-docs/pull/722. Backwards compatibility is not needed at this time as the functionality is new. The `CACHING.md` does need to be updated though.

PR #6684, opened 2025-12-04
[PM-29186] Remove 2FA user interface from premium method signatures

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29186

## 📔 Objective

During the refinement for upcoming changes to some of our 2FA setup logic, the Auth team discovered that there is a misleading signature on the `CanAccessPremium()` and `HasPremiumFromOrganization()` methods on `UserService`.  These methods as written accepted an `ITwoFactorProvidersUser`, but the premium checks have evolved over time to be consumed at various other points in the code that have nothing to do with 2FA.

This PR proposes to change those signatures to accept a `User` instead.  This does not break any existing behavior, because `User` implements `ITwoFactorProvidersUser`.  The only thing that would break would be if other implementors of `ITwoFactorProvidersUser` - namely `OrganizationUserUserDetails` - needed to be used as input to `CanAccessPremium()` or `HasPremiumFromOrganization()`.  However, a check of the code shows that this user representation does not consume those methods.

If we were to need to have an abstraction for an entity that has premium, we should create an `IPremiumUser` interface at that point to maintain the proper abstraction.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Services/IUserService.cs
+++ b/src/Core/Services/IUserService.cs
@@ -72,8 +72,8 @@ public interface IUserService
     /// This primarily serves as a means to communicate to the client when a user has inherited their premium status
     /// through an organization. Feature gating logic probably should not be behind this check.
     /// </summary>
-    /// <param name="user"></param>
-    /// <returns></returns>
+    /// <param name="user">user being acted on</param>
+    /// <returns>true if they can access premium because of organization membership; false otherwise.</returns>
     Task<bool> HasPremiumFromOrganization(User user);
     Task<string> GenerateSignInTokenAsync(User user, string purpose);
 


eliykat (dismissed):

eliykat (approved):

src/Core/Services/IUserService.cs
```diff
@@ -60,11 +59,22 @@ Task<UserLicense> GenerateLicenseAsync(User user, SubscriptionInfo subscriptionI
...
-    Task<bool> HasPremiumFromOrganization(ITwoFactorProvidersUser user);
+    Task<bool> CanAccessPremium(User user);
+
+    /// <summary>
+    /// Checks if the user has inherited access to premium features through an organization.
+    ///
+    /// This primarily serves as a means to communicate to the client when a user has inherited their premium status
+    /// through an organization. Feature gating logic probably should not be behind this check.
+    /// </summary>
+    /// <param name="user"></param>
```
enmande:
```suggestion
    /// <param name="user">user being acted on</param>
```

src/Core/Services/IUserService.cs
```diff
@@ -60,11 +59,22 @@ Task<UserLicense> GenerateLicenseAsync(User user, SubscriptionInfo subscriptionI
...
+    /// <returns></returns>
```
enmande:
```suggestion
    /// <returns>true if they can access premium because of organization membership; false otherwise.</returns>
```

src/Core/Services/IUserService.cs
trmartin4: Addressed in [`99338aa` (#6684)](https://github.com/bitwarden/server/pull/6684/commits/99338aa1d6234521874eff4363226fbc77848d29).

src/Core/Services/IUserService.cs
trmartin4: Addressed in [`99338aa` (#6684)](https://github.com/bitwarden/server/pull/6684/commits/99338aa1d6234521874eff4363226fbc77848d29).

PR #6688, opened 2025-12-04
[PM-21411] Refactor interface for determining premium status and features

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21411

## 📔 Objective

Introduce `IHasPremiumAccessQuery` to centralize premium-access checks and make the distinction clearer between having a personal premium subscription (`User.Premium`) and actually having access to premium features (personal subscription or org membership).

This new query uses a new database view (`UserPremiumAccessView`) and a new stored procedure (`User_ReadPremiumAccessByIds`) to efficiently check premium status in bulk. 

The implementation is gated behind the `PremiumAccessQuery` feature flag.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Auth/UserFeatures/PremiumAccess/IPremiumAccessQuery.cs
```diff
@@ -0,0 +1,50 @@
...
+    Task<bool> CanAccessPremiumAsync(User user);
+
+    /// <summary>
+    /// Checks if a user has access to premium features (personal subscription or organization).
+    /// Use this overload when you already know the personal premium status and only need to check organization premium.
+    /// </summary>
+    /// <param name="userId">The user ID to check for premium access</param>
+    /// <param name="hasPersonalPremium">Whether the user has a personal premium subscription</param>
+    /// <returns>True if user can access premium features; false otherwise</returns>
+    Task<bool> CanAccessPremiumAsync(Guid userId, bool hasPersonalPremium);
```
eliykat: The job of this query is to encapsulate this premium logic, so asking the caller if the user `hasPersonalPremium` is a leaky abstraction. I recommend removing this in favour of the other overload that just takes a `User`. It looks like this is only used in 1 place and only because you're trying to avoid a database call; I think if you simplified that section (which I'll comment on below) you could remove this interface.

src/Core/Billing/Premium/Queries/IHasPremiumAccessQuery.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <returns>True if user can access premium features; false otherwise</returns>
+    Task<bool> CanAccessPremiumAsync(Guid userId, bool hasPersonalPremium);
+
+    /// <summary>
+    /// Checks if a user has access to premium features through organization membership only.
+    /// This is useful for determining the source of premium access (personal vs organization).
+    /// </summary>
+    /// <param name="userId">The user ID to check for organization premium access</param>
+    /// <returns>True if user has premium access through any organization; false otherwise</returns>
+    Task<bool> HasPremiumFromOrganizationAsync(Guid userId);
```
eliykat: This is currently unused, I recommend removing it if we don't have a clear use case for it.

src/Core/Auth/UserFeatures/PremiumAccess/IPremiumAccessQuery.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <returns>True if user has premium access through any organization; false otherwise</returns>
+    Task<bool> HasPremiumFromOrganizationAsync(Guid userId);
+
+    /// <summary>
+    /// Checks if multiple users have access to premium features (optimized bulk operation).
+    /// Uses cached organization abilities and minimizes database queries.
+    /// </summary>
+    /// <param name="users">The users to check for premium access</param>
+    /// <returns>Dictionary mapping user IDs to their premium access status (personal or through organization)</returns>
+    Task<Dictionary<Guid, bool>> CanAccessPremiumBulkAsync(IEnumerable<User> users);
```
eliykat: 🎨 I think this is a good use case for an overload rather than adding `Bulk` to the name.

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
eliykat, line 1: I think trying to adjust the logic of the existing methods in-line has made this messy. I suggest writing separate vNext methods with your new implementation.

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
```diff
@@ -71,13 +106,43 @@ public async Task<bool> TwoFactorIsEnabledAsync(ITwoFactorProvidersUser user)
...
+            else
+            {
+                // Fallback: fetch the User entity
+                var fetchedUser = await _userRepository.GetByIdAsync(userId.Value);
+                if (fetchedUser == null)
+                {
+                    return false;
+                }
+                hasPersonalPremium = fetchedUser.Premium;
+            }
```
eliykat: This is trying to do the job of your new query, which is meant to be the only implementation of this logic. Some options:
- maybe Auth were wrong to remove the `GetPremium` method from `ITwoFactorProvidersUser`? It was unused, but it seems like you have a use for it here. Although if you go this way, I would do it like this:

```c#
public interface IPremiumUser
{
  public bool HasPremium { get; set; } // move this to its own interface
}

public interface ITwoFactorProvidersUser : IPremiumUser // this inherits it
{
  // snip
}

public interface IPremiumAccessQuery
{
      Task<bool> CanAccessPremiumAsync(IPremiumUser user); // accepts the interface instead of any particular object

  // snip
}

// and now in this method, you can just pass the argument along...
_premiumAccessQuery.CanAccessPremiumAsync(user);
```
 
Other options:
- require the caller to pass in a `User` to begin with
- just always fetch the `User`, don't worry about it

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
```diff
@@ -20,15 +36,34 @@ public class TwoFactorIsEnabledQuery(IUserRepository userRepository) : ITwoFacto
...
+            var userDetails = await _userRepository.GetManyWithCalculatedPremiumAsync([.. userIds]);
+            foreach (var userDetail in userDetails)
+            {
+                result.Add(
+                    (userDetail.Id,
+                     await TwoFactorEnabledAsync(userDetail.GetTwoFactorProviders(),
+                                    () => Task.FromResult(userDetail.HasPremiumAccess))
+                    )
+                );
+            }
```
eliykat: We shouldn't call an async method in a loop, even though it's a kind of fake async (using `Task.FromResult`), it's too liable to be misused. This is where I think writing a fresh implementation will help.

src/Core/Entities/User.cs
```diff
@@ -69,6 +69,10 @@ public class User : ITableObject<Guid>, IStorableSubscriber, IRevisable, ITwoFac
     /// The security state is a signed object attesting to the version of the user's account.
     /// </summary>
     public string? SecurityState { get; set; }
+    /// <summary>
+    /// Indicates whether the user has a personal premium subscription.
+    /// Does not include premium access from organizations.
```
eliykat:
```suggestion
    /// Does not include premium access from organizations -
    /// do not use this to check whether the user can access premium features.
```

src/Core/Auth/UserFeatures/PremiumAccess/PremiumAccessQuery.cs
```diff
@@ -0,0 +1,97 @@
...
+                userOrgs.Any(ou =>
+                    orgAbilities.TryGetValue(ou.OrganizationId, out var orgAbility) &&
+                    orgAbility.UsersGetPremium &&
+                    orgAbility.Enabled);
+
+            result[user.Id] = hasPremiumFromOrg;
+        }
+
+        return result;
+    }
```
eliykat: This has a database call + cache + a nested loop to get information that ultimately comes from the database anyway.

I suggest a simpler implementation would be a new sproc for this purpose (untested):

```sql
    SELECT
        U.[Id],
        MAX(U.[Premium]) PersonalPremium,
        MAX(O.[UsersGetPremium]) OrganizationPremium
    FROM
        [dbo].[UserView] U
        JOIN [dbo].[OrganizationUserView] OU ON U.Id = OU.UserId
        JOIN [dbo].[Organization] O ON OU.OrganizationId = O.Id
    WHERE
        U.[Id] IN (SELECT [Id] FROM @Ids)
    GROUP BY
        U.[Id]
```

And then this code could just return `PersonalPremium || OrganizationPremium` for each user.

I think this is better than the current sproc because it's not returning this weird `UserWithCalculatedPremium` object, it's very clear what its purpose is, and it's only used by this query.

src/Core/Billing/Premium/Queries/IHasPremiumAccessQuery.cs
r-tome: It will eventually replace https://github.com/bitwarden/server/blob/18a8829476230649e5b981a500b737c56a5ae7be/src/Core/Services/Implementations/UserService.cs#L1118

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
r-tome: I went with the simplest option of just having the callers pass in User objects. I tried to go with the IPremiumUser solution but it was more complex than it needed to because we also needed the Id for the bulk method

src/Core/Auth/UserFeatures/PremiumAccess/PremiumAccessQuery.cs
r-tome: I'm using the existing `UserWithCalculatedPremium` for now. I'll refactor that later here on in a separate branch

r-tome: You're totally right, btw. I think initially I didn't retrieve the many users and just used the cache but for some reason this evolved and I kept that logic.

src/Core/Billing/Premium/Queries/IHasPremiumAccessQuery.cs
```diff
@@ -41,10 +30,12 @@ public interface IPremiumAccessQuery
...
     /// </summary>
-    /// <param name="users">The users to check for premium access</param>
+    /// <param name="userIds">The user IDs to check for premium access</param>
     /// <returns>Dictionary mapping user IDs to their premium access status (personal or through organization)</returns>
-    Task<Dictionary<Guid, bool>> CanAccessPremiumBulkAsync(IEnumerable<User> users);
+    Task<Dictionary<Guid, bool>> HasPremiumAccessAsync(IEnumerable<Guid> userIds);
 }
 
+
+
```
eliykat: ⛏️ Extra newlines at end of file

src/Core/Repositories/IUserRepository.cs
eliykat, line 1: These methods are no longer obsolete, because your new query is using it. Although you may still want to mention that this is for internal use only, and devs should use the query instead.

All of that said - I think this sproc could be simplified per my last review, and now is a good time to do it. (i.e. by making a new sproc used by the new code, then deleting this when you clean up your feature flag). I'll leave it up to you though. It could also be a separate ticket.

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public HasPremiumAccessQuery(IUserRepository userRepository)
+    {
+        _userRepository = userRepository;
+    }
+
+    public async Task<bool> HasPremiumAccessAsync(Guid userId)
+    {
+        var user = await _userRepository.GetCalculatedPremiumAsync(userId);
+        return user?.HasPremiumAccess ?? false;
```
eliykat: If there is no `user`, something has gone wrong. I think it should throw an exception in this case.

src/Core/Billing/Premium/Queries/IHasPremiumAccessQuery.cs
eliykat: Does it need to be wired up in that method though? It's not called from anywhere at the moment (as far as I can see).

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public async Task<Dictionary<Guid, bool>> HasPremiumAccessAsync(IEnumerable<Guid> userIds)
+    {
+        var usersWithPremium = await _userRepository.GetManyWithCalculatedPremiumAsync(userIds);
```
eliykat: To be consistent with the other method, and with the interface for this query (which promises a key for each `userId`), I think this should also check and throw for missing users.

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
```diff
@@ -0,0 +1,44 @@
...
+    public async Task<bool> HasPremiumFromOrganizationAsync(Guid userId)
+    {
+        var user = await _userRepository.GetCalculatedPremiumAsync(userId);
+        if (user == null)
+        {
+            return false;
+        }
+
+        // Has org premium if has premium access but not personal premium
+        return user.HasPremiumAccess && !user.Premium;
```
eliykat: This is subtly different to the current implementation in `UserService.HasPremiumFromOrganization`: that current logic will return false if the user is not a part of any orgs, but will _not_ return false if they have both personal premium and premium from an org. That doesn't seem right either, but it's unclear what the intent is.

That said, I have no idea why we need this logic: it's only synced to clients but doesn't seem to be used there either. I would be interested to know if we could remove it.

Any ideas @amorask-bitwarden ?

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
eliykat, line 1: It's so simple now 👏

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
```diff
@@ -72,12 +83,113 @@ public async Task<bool> TwoFactorIsEnabledAsync(ITwoFactorProvidersUser user)
...
+        {
+            return false;
+        }
+
+        // If there are only premium two factor options then check premium access
+        var onlyHasPremiumTwoFactor = enabledProviderKeys.All(TwoFactorProvider.RequiresPremium);
+        if (onlyHasPremiumTwoFactor)
+        {
+            return hasPremiumAccess;
+        }
```
eliykat: I think this was previously done to avoid the db call, which was passed in as a delegate to defer that work. But now we're doing our database call upfront, I think this whole method could be simplified to:

```c#
        return providers != null && providers.Values.Any(p => p.Enabled);
```

which means you probably don't need the private method at all. I think that's a great simplification of this logic.

That said, Auth might have greater insight into any performance requirements here re: that database call (and this logic generally).

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
eliykat, line 1: I think you still need to wire up the new methods here. (preferably just have the old methods call them and return the result)

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
r-tome: I refactored this quite a bit and avoided checking premium status unless we need to.

src/Core/Repositories/IUserRepository.cs
r-tome: I have implemented the new query so there are obsolete!

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
```diff
@@ -65,19 +90,146 @@ await TwoFactorEnabledAsync(userDetail.GetTwoFactorProviders(),
...
+            var userId = user.GetUserId();
+            if (!userId.HasValue)
+            {
+                return false;
+            }
+
+            var userEntity = user as User ?? await _userRepository.GetByIdAsync(userId.Value);
+            if (userEntity == null)
+            {
+                return false;
```
eliykat: I think this should throw - there should be a valid `User` in the database.

r-tome:
> I just noticed that User has a PremiumExpirationDate. Do we need to check if premium has expired? Or is the premium column automatically updated when the expiration date passes?

@eliykat the Premium field is automatically set to false. For example in cloud instances its done via a Stripe event https://github.com/bitwarden/server/blob/main/src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
r-tome: I have updated the new implementation to match the logic in `UserService`.

`HasPremiumFromOrganization` is used through `SyncController` for billing logic.

src/Core/Auth/UserFeatures/TwoFactorAuth/TwoFactorIsEnabledQuery.cs
r-tome: Agreed! Updated.

src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserUserDetails.cs
```diff
@@ -20,6 +20,12 @@ public class OrganizationUserUserDetails : IExternal, ITwoFactorProvidersUser, I
     public string Email { get; set; }
     public string AvatarColor { get; set; }
     public string TwoFactorProviders { get; set; }
+    /// <summary>
+    /// Indicates whether the user has a personal premium subscription.
+    /// Does not include premium access from organizations -
+    /// do not use this to check whether the user can access premium features.
+    /// Null when the organization user is in Invited status (UserId is null).
+    /// </summary>
     public bool? Premium { get; set; }
```
ike-kottlowski: 🎨 : Do we want to add an `[Obsolete()]` here with direction to the new `IHasPremiumAccessQuery`?

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        return usersWithPremium.ToDictionary(u => u.Id, u => u.HasPremiumAccess);
+    }
+
+    public async Task<bool> HasPremiumFromOrganizationAsync(Guid userId)
+    {
+        var user = await _userRepository.GetPremiumAccessAsync(userId);
+        if (user == null)
+        {
+            throw new NotFoundException();
```
cyprain-okeke, line 44: Should we return False here before throw NotFoundException

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        _userRepository = userRepository;
+    }
+
+    public async Task<bool> HasPremiumAccessAsync(Guid userId)
+    {
+        var user = await _userRepository.GetPremiumAccessAsync(userId);
+        if (user == null)
+        {
+            throw new NotFoundException();
```
cyprain-okeke, line 20: Should we return False here before throw NotFoundException

cyprain-okeke (changes requested):
Nice Work, I just have some few clarifcations

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
r-tome: If we return `false` we would imply that the user exists and does not have premium access. Whereas by throwing an exception, we're indicating something went wrong - we shouldn't be checking premium for a user that doesn't exist in the first place.

src/Core/Billing/Premium/Queries/HasPremiumAccessQuery.cs
r-tome: Same answer as above

util/Migrator/DbScripts/2025-12-09_00_UserPremiumAccessView.sql
```diff
@@ -0,0 +1,53 @@
...
+    SELECT * FROM sys.indexes 
+    WHERE name = 'IX_Organization_Enabled' 
+    AND object_id = OBJECT_ID('[dbo].[Organization]')
+)
+BEGIN
+    DROP INDEX [IX_Organization_Enabled] ON [dbo].[Organization];
+END
+GO
+
+CREATE NONCLUSTERED INDEX [IX_Organization_Enabled]
```
mkincaid-bw: 🌱 In this case, dropping and re-creating this index is likely okay since the Organization table is relatively small. For larger tables, it's better to create the index with `DROP_EXISTING = ON`. This allows SQL server to use the existing index to build the new one, and also keeps the existing one in place so it can be used for queries until the new one is ready.

```sql
IF EXISTS (
    SELECT * FROM sys.indexes 
    WHERE name = 'IX_Organization_Enabled' 
    AND object_id = OBJECT_ID('[dbo].[Organization]')
)
BEGIN
    CREATE NONCLUSTERED INDEX [IX_Organization_Enabled]
    ON [dbo].[Organization]([Id] ASC, [Enabled] ASC)
    INCLUDE ([UseTotp], [UsersGetPremium])
    WITH (DROP_EXISTING = ON);
END
ELSE
BEGIN
    CREATE NONCLUSTERED INDEX [IX_Organization_Enabled]
    ON [dbo].[Organization]([Id] ASC, [Enabled] ASC)
    INCLUDE ([UseTotp], [UsersGetPremium]);
END
```

util/Migrator/DbScripts/2025-12-12_00_UserPremiumAccessView.sql
```diff
@@ -0,0 +1,53 @@
...
+                AND O.[Enabled] = 1
+        ) THEN 1 
+        ELSE 0 
+    END AS BIT) AS [OrganizationPremium]
+FROM
+    [dbo].[User] U;
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[User_ReadPremiumAccessByIds]
+    @Ids [dbo].[GuidIdArray] READONLY
```
mkincaid-bw, line 46: ❓ How many users do you anticipate being passed into this proc? I tested your view with different amounts of random users (100 users, 1,000 users, 10,000 users, etc). With 10,000 users, this query took almost 10 seconds against a prod backup DB. 

The correlated subquery with the EXISTS statement won't scale well. If performance is a concern, perhaps consider re-writing the view with joins instead. 

```sql
CREATE OR ALTER VIEW [dbo].[UserPremiumAccessView]
AS
SELECT 
    U.[Id],
    U.[Premium] AS [PersonalPremium],
    CAST(
            MAX(CASE 
                    WHEN O.[Id] IS NOT NULL THEN 1 
                    ELSE 0 
                END
                ) 
            AS BIT
        ) AS [OrganizationPremium]
FROM 
    [dbo].[User] U
LEFT JOIN 
    [dbo].[OrganizationUser] OU ON OU.[UserId] = U.[Id]
LEFT JOIN 
    [dbo].[Organization] O ON O.[Id] = OU.[OrganizationId]
    AND O.[UsersGetPremium] = 1
    AND O.[Enabled] = 1
GROUP BY 
    U.[Id], U.[Premium];
```

I had Claude convert the query so you should verify that the logic is still correct, but this query scales much better over large data sets.

util/Migrator/DbScripts/2025-12-09_00_UserPremiumAccessView.sql
r-tome: Good to know that! Updated

util/Migrator/DbScripts/2025-12-12_00_UserPremiumAccessView.sql
r-tome: > How many users do you anticipate being passed into this proc?

Most common use case is one user, the sproc will be called for every sync operation.
For multiple users it'll be called when organization admins view the members page. Definitely not as often as the individual sync operations but then it depends on the organization size.

Yes, performance is a concern so thanks for checking that! Your performance testing with 10k users confirms we need the JOIN-based approach
I updated that query and also updated the integration tests to make sure they all pass.

mkincaid-bw (approved):
LGTM

util/MySqlMigrations/Migrations/20251212171212_OrganizationUsersGetPremiumIndex.cs
eliykat, line 1: Just double checking that this empty migration is OK/intentional?
(same for Sqlite)

r-tome: Yes, MySql and Sqlite do not support `include` columns. We could add those columns to the index key but that would be overkill

cyprain-okeke (approved):

Patrick-Pimentel-Bitwarden (approved):
👍 These changes are acceptable. Good work and thank you for adding good tests.

PR #6714, opened 2025-12-09
Refactor event integration service collection extensions into their own extension

brant-livefront:
## 📔 Objective

As the event integration services grew, they were simply being added to the large ServiceCollectionExtensions in SharedWeb. This PR refactors them into their own self-contained extension. This gives us the advantage of not having so much done in one file, but also added an easy place to add extensive tests for all of these (where there previously had been no tests).

This is primarily simply a migration. However there are 2 functional changes:
1. We were previously calling `AddDistributedCache` inside of the event integration functions. This is not possible when moving outside of SharedWeb. Instead, we are now calling `AddDistributedCache` in the Startup.cs in Events and EventsProcessor (which follows the pattern in other apps - Api, Billing, etc). This has been documented as a requirement in the XMLDoc for these methods
2. In writing tests I realized that we supported launching listeners for _both_ Azure Service Bus _and_ RabbitMQ if you had both of them configured. This is likely a misconfiguration and is out of step with how we handle precedence on the _write_ side. I've updated the logic to explicitly prefer ASB and register only ASB if they are both configured (as well as adding a test to exercise this scenario explicitly).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Core.csproj
```diff
@@ -41,6 +41,7 @@
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.52.0" />
     <PackageReference Include="Microsoft.Azure.NotificationHubs" Version="4.2.0" />
     <PackageReference Include="Microsoft.Bot.Builder" Version="4.23.0" />
+    <PackageReference Include="Microsoft.Bot.Builder.Integration.AspNet.Core" Version="4.23.0" />
```
jrmccannon, line 44: Just to refresh my memory, the bot builder libraries are required for the Teams integration right?

brant-livefront: yep, exactly.

brant-livefront:
@jrmccannon @withinfocus This is ready for a re-approval after being re-targeted to main (should be no new changes)

jrmccannon (approved):

PR #6715, opened 2025-12-09
feat(register): [PM-27084] Account Register Uses New Data Types

Patrick-Pimentel-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27084

## 📔 Objective

1. Update register to use new data types
2. Added comments so that future removal of old properties is clear
3. Made accounts controller not permit nullish ambiguity
4. Added tests to demonstrate payloads result in same outcomes for the user model

## 📸 Screenshots

https://github.com/user-attachments/assets/bcf4b3de-a993-4d28-aef6-36969e1c4242

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/test/Core.Test/Auth/AutoFixture/RegisterFinishRequestModelFixtures.cs
+++ b/test/Core.Test/Auth/AutoFixture/RegisterFinishRequestModelFixtures.cs
@@ -29,7 +29,9 @@ internal class RegisterFinishRequestModelCustomization : ICustomization
             .With(o => o.OrgInviteToken, OrgInviteToken)
             .With(o => o.OrgSponsoredFreeFamilyPlanToken, OrgSponsoredFreeFamilyPlanToken)
             .With(o => o.AcceptEmergencyAccessInviteToken, AcceptEmergencyAccessInviteToken)
-            .With(o => o.ProviderInviteToken, ProviderInviteToken));
+            .With(o => o.ProviderInviteToken, ProviderInviteToken)
+            .Without(o => o.MasterPasswordAuthentication)
+            .Without(o => o.MasterPasswordUnlock));
     }
 }
 


Patrick-Pimentel-Bitwarden:
Adding @eligrubb to review because he has work he is branching off of this to do!

Patrick-Pimentel-Bitwarden:
There are remote test failures that I cannot produce locally, feel free to hold off on approving.

Patrick-Pimentel-Bitwarden:
@quexten Is claude's comment correct? See full review above.

🔴 Critical Issue: Cryptographic Inconsistency Risk
Location: RegisterFinishRequestModel.cs:70-77 and AccountsController.cs:159

When both MasterPasswordAuthenticationData and MasterPasswordUnlockData are provided, there's no validation that their KDF settings match. This creates a critical security issue:

The authentication hash is computed using MasterPasswordAuthenticationData.Kdf
But the stored user gets MasterPasswordUnlockData.Kdf settings
If these differ, authentication will fail or could accept incorrectly computed hashes
Attack scenario: A client could send:

MasterPasswordAuthenticationData with PBKDF2 100k iterations
MasterPasswordUnlockData with Argon2id 3 iterations
This bypasses KDF iteration requirements and violates cryptographic integrity.

Required fix: Add validation in RegisterFinishRequestModel.Validate():

if (MasterPasswordAuthenticationData != null && MasterPasswordUnlockData != null)
{
    if (!MasterPasswordAuthenticationData.Kdf.Equals(MasterPasswordUnlockData.Kdf))
    {
        yield return new ValidationResult(
            "KDF settings must match between authentication and unlock data.");
    }
    if (MasterPasswordAuthenticationData.Salt != MasterPasswordUnlockData.Salt)
    {
        yield return new ValidationResult(
            "Salt must match between authentication and unlock data.");
    }
}

src/Identity/Controllers/AccountsController.cs
```diff
@@ -145,40 +141,85 @@ public async Task<IActionResult> PostRegisterVerificationEmailClicked([FromBody]
...
-                identityResult =
-                    await _registerUserCommand.RegisterUserViaEmailVerificationToken(user, model.MasterPasswordHash,
-                        model.EmailVerificationToken);
+                if (string.IsNullOrEmpty(model.EmailVerificationToken))
+                    throw new BadRequestException("Email verification token absent when processing register/finish.");
 
+                identityResult = await _registerUserCommand.RegisterUserViaEmailVerificationToken(
+                    user,
+                    masterPasswordHash,
+                    model.EmailVerificationToken);
```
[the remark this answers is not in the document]
quexten: This is not a security concern but a stability concern.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -21,19 +21,32 @@ public class RegisterFinishRequestModel : IValidatableObject
     public required string Email { get; set; }
     public string? EmailVerificationToken { get; set; }
 
+    public MasterPasswordAuthenticationData? MasterPasswordAuthenticationData { get; set; }
```
mzieniukbw: 🎨  KM team drops the "Data" suffix from the name, which i think is clearer.
```suggestion
    public MasterPasswordAuthenticationData? MasterPasswordAuthentication { get; set; }
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -21,19 +21,32 @@ public class RegisterFinishRequestModel : IValidatableObject
...
+    public MasterPasswordUnlockData? MasterPasswordUnlockData { get; set; }
```
mzieniukbw: 🎨  KM team drops the "Data" suffix from the name, which i think is clearer.

```suggestion
    public MasterPasswordUnlockData? MasterPasswordUnlock { get; set; }
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -21,19 +21,32 @@ public class RegisterFinishRequestModel : IValidatableObject
...
     [StringLength(50)]
     public string? MasterPasswordHint { get; set; }
 
-    public required string UserSymmetricKey { get; set; }
+    // PM-28143 - Remove property below (made optional during migration to MasterPasswordUnlockData)
+    // Made optional but there will still be a thrown error if it does not exist either here or
+    // in the MasterPasswordAuthenticationData.
+    public string? UserSymmetricKey { get; set; }
 
     public required KeysRequestModel UserAsymmetricKeys { get; set; }
```
mzieniukbw, line 41: ℹ️  This is outside of the scope of this PR, but fyi, for account registration V2, instead of `UserAsymmetricKeys` (will become optional for V1 compatibility), we will use `AccountKeysRequestModel`

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
mzieniukbw: ⚠️  The Data model does not have any validation, please use `MasterPasswordAuthenticationDataRequestModel`.
Note, this request model live inside the `Api` project, which KM team wanted to keep, but since we now have a case for `Api` and `Identity`, i think this is worth considering to move out to `Core`. @bitwarden/team-key-management-dev  opinions ?
On a side note, i do not understand why `RegisterFinishRequestModel` is in `Core` project, i would move it out to `Identity`.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
mzieniukbw: ⚠️  The Data model does not have any validation, please use `MasterPasswordUnlockDataRequestModel`.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -95,6 +121,30 @@ public RegisterFinishTokenType GetTokenType()
 
     public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
     {
-        return KdfSettingsValidator.Validate(Kdf, KdfIterations, KdfMemory, KdfParallelism);
+        MasterPasswordUnlockData.ThrowIfExistsAndNotMatchingAuthenticationData(MasterPasswordAuthenticationData, MasterPasswordUnlockData);
+
+        // PM-28143 - Remove line below
+        var kdf = MasterPasswordUnlockData?.Kdf.KdfType
```
mzieniukbw: We either have both authentication and unlock data or we don't. It would be much cleaner to separate out the logic to something like:
```
if (MasterPasswordUnlockData != null && MasterPasswordAuthenticationData != null) {
  // Validate only from new data models
} else {
  // Validate only from old fields.
}
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -50,15 +63,28 @@ public class RegisterFinishRequestModel : IValidatableObject
...
     public User ToUser()
     {
+        // PM-28143 - Remove line below
+        // When we process this request to a user object, check if the unlock and authentication
+        // data has been passed through, and if so they should have matching values.
+        MasterPasswordUnlockData.ThrowIfExistsAndNotMatchingAuthenticationData(MasterPasswordAuthenticationData, MasterPasswordUnlockData);
+
+        // PM-28143 - Remove line below
+        MasterPasswordAuthenticationData.ThrowIfExistsAndHashIsNotEqual(MasterPasswordAuthenticationData, MasterPasswordHash);
+
```
mzieniukbw: ❓  I think it's clearer to move it out to `Validate`

```suggestion
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -50,15 +63,28 @@ public class RegisterFinishRequestModel : IValidatableObject
...
         {
             Email = Email,
             MasterPasswordHint = MasterPasswordHint,
-            Kdf = Kdf,
-            KdfIterations = KdfIterations,
-            KdfMemory = KdfMemory,
-            KdfParallelism = KdfParallelism,
-            Key = UserSymmetricKey,
+            Kdf = MasterPasswordUnlockData?.Kdf.KdfType ?? Kdf
+                ?? throw new Exception("KdfType couldn't be found on either the MasterPasswordUnlockData or the Kdf property passed in."),
```
mzieniukbw: ❓  We already handle the validation in `Validate` and this is request model, so the expectation should be clear, that `Validate` is always called first. Is there any point in doing it twice ?
```suggestion
            Kdf = MasterPasswordUnlockData?.Kdf.KdfType ?? Kdf!
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -50,15 +63,28 @@ public class RegisterFinishRequestModel : IValidatableObject
...
+            KdfIterations = MasterPasswordUnlockData?.Kdf.Iterations ?? KdfIterations
+                ?? throw new Exception("KdfIterations couldn't be found on either the MasterPasswordUnlockData or the KdfIterations property passed in."),
+            // KdfMemory and KdfParallelism are optional (only used for Argon2id)
+            KdfMemory = MasterPasswordUnlockData?.Kdf.Memory ?? KdfMemory,
+            KdfParallelism = MasterPasswordUnlockData?.Kdf.Parallelism ?? KdfParallelism,
+            // PM-28827 To be added when MasterPasswordSalt is added to the user column
+            // MasterPasswordSalt = MasterPasswordUnlockData?.Salt ?? Email.ToLower().Trim(),
```
mzieniukbw: ❓  This might be just my opinon, but this is future proofing, which is not great. Why not leave it to PM-28827, especially considering account registration V2 will be next in priority and i can't predict how this request model will look like later.

src/Core/KeyManagement/Models/Data/MasterPasswordAuthenticationData.cs
```diff
@@ -16,4 +16,17 @@ public void ValidateSaltUnchangedForUser(User user)
             throw new BadRequestException("Invalid master password salt.");
         }
     }
+
+    public static void ThrowIfExistsAndHashIsNotEqual(
```
mzieniukbw: Since this is temporary and is going to be removed with https://bitwarden.atlassian.net/browse/PM-28143, i would prefer to move out this validation into `RegisterFinishRequestModel` itself or at minimum to add a comment stating the intended removal in docs (optionally with added `[Obsolete]` attribute)

src/Core/KeyManagement/Models/Data/MasterPasswordUnlockData.cs
```diff
@@ -16,4 +16,31 @@ public void ValidateSaltUnchangedForUser(User user)
...
+            }
+        }
+    }
+
+    private static bool MatchesAuthenticationData(
+        MasterPasswordUnlockData unlockData,
+        MasterPasswordAuthenticationData authenticationData)
+    {
+        var kdfMatches = unlockData.Kdf.Equals(authenticationData.Kdf);
+        var saltMatches = unlockData.Salt == authenticationData.Salt;
```
mzieniukbw: 🤔 non blocking: This ties authentication and unlock salt's together, which might not be a plan longer run. I think @quexten did more investigation on this, opinions ?

src/Core/KeyManagement/Models/Data/MasterPasswordUnlockData.cs
```diff
@@ -16,4 +16,31 @@ public void ValidateSaltUnchangedForUser(User user)
...
+    {
+        if (unlockData != null && authenticationData != null)
+        {
+            var matches = MatchesAuthenticationData(
+                unlockData,
+                authenticationData);
+
+            if (!matches)
+            {
+                throw new Exception("KDF settings and salt must match between authentication and unlock data.");
```
mzieniukbw: Use Core's `BadRequestException` directly.
```suggestion
                throw new BadRequestException("KDF settings and salt must match between authentication and unlock data.");
```

src/Core/KeyManagement/Models/Data/MasterPasswordUnlockData.cs
```diff
@@ -16,4 +16,31 @@ public void ValidateSaltUnchangedForUser(User user)
             throw new BadRequestException("Invalid master password salt.");
         }
     }
+
+    public static void ThrowIfExistsAndNotMatchingAuthenticationData(
+        MasterPasswordAuthenticationData? authenticationData,
+        MasterPasswordUnlockData? unlockData)
+    {
+        if (unlockData != null && authenticationData != null)
```
mzieniukbw: Since both fields will become required with https://bitwarden.atlassian.net/browse/PM-28143, it would make more sense to make both required, change name to `ThrowIfNotMatchingAuthenticationData` and do the nullability check within the `RegisterFinishRequestModel` validation.

src/Core/KeyManagement/Models/Data/MasterPasswordAuthenticationData.cs
```diff
@@ -16,4 +16,17 @@ public void ValidateSaltUnchangedForUser(User user)
...
+        MasterPasswordAuthenticationData? authenticationData,
+        string? hash)
+    {
+        if (authenticationData != null && hash != null)
+        {
+            if (authenticationData.MasterPasswordAuthenticationHash != hash)
+            {
+                throw new Exception("Master password hash and hash are not equal.");
```
mzieniukbw: Use Core's `BadRequestException` instead
```suggestion
                throw new BadRequestException("Master password hash and hash are not equal.");
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
mzieniukbw: Use Core's `BadRequestException` directly.
This applies to all validation in this request model

src/Core/Utilities/KdfSettingsValidator.cs
```diff
@@ -36,6 +37,34 @@ public static IEnumerable<ValidationResult> Validate(KdfType kdfType, int kdfIte
         }
     }
 
+    // PM-28143 - Will be used in the referenced ticket.
+    public static IEnumerable<ValidationResult> Validate(MasterPasswordUnlockData masterPasswordUnlockData)
```
mzieniukbw: Since this is validation for requests, i would advocate to move this out to `KdfRequestModel`, which should be used in all request models involving KDF.

src/Identity/Controllers/AccountsController.cs
```diff
@@ -145,40 +141,85 @@ public async Task<IActionResult> PostRegisterVerificationEmailClicked([FromBody]
...
-        var user = model.ToUser();
+        User user;
+
+        try
+        {
+            user = model.ToUser();
+        }
+        catch (Exception e)
+        {
+            throw new BadRequestException(e.Message);
```
mzieniukbw: The validation should be done within the `Validate` and not `ToUser`. If there is a case where an exception needs to be thrown out within `ToUser` (which i have not found), then we should already expect `ValidationResult` or `BadRequestException` exception type.

src/Identity/Controllers/AccountsController.cs
```diff
@@ -145,40 +141,85 @@ public async Task<IActionResult> PostRegisterVerificationEmailClicked([FromBody]
...
+        }
 
         // Users will either have an emailed token or an email verification token - not both.
-        IdentityResult identityResult = null;
+        IdentityResult? identityResult = null;
+
+        // PM-28143 - Just use the MasterPasswordAuthenticationData.MasterPasswordAuthenticationHash
+        string masterPasswordHash = model.MasterPasswordAuthenticationData?.MasterPasswordAuthenticationHash
+                                 ?? model.MasterPasswordHash ?? throw new BadRequestException("MasterPasswordHash couldn't be found on either the MasterPasswordAuthenticationData or the MasterPasswordHash property passed in.");
```
mzieniukbw: ⛏️  non blocking: I would rely on request validation, rather than duplicating the validation within the controller.

src/Identity/Controllers/AccountsController.cs
```diff
@@ -145,40 +141,85 @@ public async Task<IActionResult> PostRegisterVerificationEmailClicked([FromBody]
...
 
         switch (model.GetTokenType())
         {
             case RegisterFinishTokenType.EmailVerification:
-                identityResult =
-                    await _registerUserCommand.RegisterUserViaEmailVerificationToken(user, model.MasterPasswordHash,
-                        model.EmailVerificationToken);
+                if (string.IsNullOrEmpty(model.EmailVerificationToken))
```
mzieniukbw: ⛏️  non blocking: I would move it to request validation. Same for all token types

mzieniukbw (changes requested):
Functionally ok from KM perspective, but i have reservations about the placement of validation and used data models in the request.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
Patrick-Pimentel-Bitwarden: Thank you for sharing that knowledge, helps me understand future changes.

src/Identity/Controllers/AccountsController.cs
Patrick-Pimentel-Bitwarden: fixed with [a705ea4](https://github.com/bitwarden/server/pull/6715/commits/a705ea4c579f7cfd2c0f77d7337e1f5a7428a058)

src/Identity/Controllers/AccountsController.cs
Patrick-Pimentel-Bitwarden: fixed with [a705ea4](https://github.com/bitwarden/server/pull/6715/commits/a705ea4c579f7cfd2c0f77d7337e1f5a7428a058)

src/Identity/Controllers/AccountsController.cs
Patrick-Pimentel-Bitwarden: fixed with [a705ea4](https://github.com/bitwarden/server/pull/6715/commits/a705ea4c579f7cfd2c0f77d7337e1f5a7428a058)

src/Core/Utilities/KdfSettingsValidator.cs
Patrick-Pimentel-Bitwarden: fixed with [a705ea4](https://github.com/bitwarden/server/pull/6715/commits/a705ea4c579f7cfd2c0f77d7337e1f5a7428a058)

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
Patrick-Pimentel-Bitwarden: fixed with [a705ea4](https://github.com/bitwarden/server/pull/6715/commits/a705ea4c579f7cfd2c0f77d7337e1f5a7428a058)

src/Core/KeyManagement/Models/Data/MasterPasswordAuthenticationData.cs
Patrick-Pimentel-Bitwarden: fixed with [a705ea4](https://github.com/bitwarden/server/pull/6715/commits/a705ea4c579f7cfd2c0f77d7337e1f5a7428a058)

src/Core/KeyManagement/Models/Data/MasterPasswordUnlockData.cs
Patrick-Pimentel-Bitwarden: This check needed to be removed as we want to support the authentication and unlock data having different values.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -61,30 +62,56 @@ public class RegisterFinishRequestModel : IValidatableObject
 
     public Guid? ProviderUserId { get; set; }
 
+    // Strongly-typed accessors for post-validation usage to satisfy nullability
+    [System.Text.Json.Serialization.JsonIgnore]
+    [Newtonsoft.Json.JsonIgnore]
+    public string EmailVerificationTokenRequired =>
+        EmailVerificationToken ?? throw new BadRequestException("Email verification token absent when processing register/finish.");
```
mzieniukbw: Since we already do the validation of this, why do we need this ? If you want to avoid the nullability check, just use the [null forgiving operator](https://contributing.bitwarden.com/contributing/code-style/csharp#null-forgiving-operator)

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -118,33 +145,104 @@ public RegisterFinishTokenType GetTokenType()
...
-
-        // PM-28143 - Remove line below
-        var kdfIterations = MasterPasswordUnlockData?.Kdf.Iterations
-                            ?? KdfIterations
-                            ?? throw new Exception($"{nameof(KdfIterations)} not found on RequestModel");
+        // PM-28143 - Remove this check
+        ThrowIfExistsAndHashIsNotEqual(MasterPasswordAuthentication, MasterPasswordHash);
+
+        // Ensure exactly one registration token type is provided
+        var hasEmailVerification = !string.IsNullOrWhiteSpace(EmailVerificationToken);
```
mzieniukbw: The validation of the token's is already done in more details few lines below with `switch (GetTokenType())`, this is redundant.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -118,33 +145,104 @@ public RegisterFinishTokenType GetTokenType()
...
+                    throw new BadRequestException("Provider user id absent when processing register/finish.");
+                }
+                break;
+            default:
+                throw new BadRequestException("Invalid registration finish request");
+        }
 
-        // PM-28143 - Remove line below in favor of using the unlock data.
-        return KdfSettingsValidator.Validate(kdf, kdfIterations, kdfMemory, kdfParallelism);
+        return kdfValidationResults;
```
mzieniukbw: Since the last thing we do is KDF validation, i would move the actual KDF checks (lines 173 to 185) to be the last.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -118,33 +145,104 @@ public RegisterFinishTokenType GetTokenType()
...
+    // PM-28143 - Remove function
+    private static void ThrowIfExistsAndHashIsNotEqual(
+        MasterPasswordAuthenticationDataRequestModel? authenticationData,
+        string? hash)
+    {
+        if (authenticationData != null && hash != null)
+        {
+            if (authenticationData.MasterPasswordAuthenticationHash != hash)
+            {
+                throw new BadRequestException("Master password hash and hash are not equal.");
```
mzieniukbw: This message is somewhat confusing, maybe we should say what we are validating, as of: `MasterPasswordHash` and `MasterPasswordAuthenticationHash` ?

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -61,30 +62,56 @@ public class RegisterFinishRequestModel : IValidatableObject
...
         var user = new User
         {
             Email = Email,
             MasterPasswordHint = MasterPasswordHint,
-            Kdf = MasterPasswordUnlockData?.Kdf.KdfType ?? Kdf
-                ?? throw new Exception("KdfType couldn't be found on either the MasterPasswordUnlockData or the Kdf property passed in."),
-            KdfIterations = MasterPasswordUnlockData?.Kdf.Iterations ?? KdfIterations
-                ?? throw new Exception("KdfIterations couldn't be found on either the MasterPasswordUnlockData or the KdfIterations property passed in."),
+            Kdf = MasterPasswordUnlock?.Kdf.KdfType ?? Kdf
+                ?? throw new BadRequestException("KdfType couldn't be found on either the MasterPasswordUnlockData or the Kdf property passed in."),
```
mzieniukbw: The error message still mentions `MasterPasswordUnlockData`, but should be `MasterPasswordUnlock` now, since the field name is different.

src/Core/KeyManagement/Models/Data/KdfRequestModel.cs
```diff
@@ -1,10 +1,10 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using Bit.Core.Enums;
-using Bit.Core.KeyManagement.Models.Data;
+using Bit.Core.Utilities;
 
-namespace Bit.Api.KeyManagement.Models.Requests;
+namespace Bit.Core.KeyManagement.Models.Data;
```
mzieniukbw: This is a request model, hence it belongs to `Models.Api.Request` namespace.

src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
```diff
@@ -1,8 +1,12 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using Bit.Core.KeyManagement.Models.Data;
 
-namespace Bit.Api.KeyManagement.Models.Requests;
+namespace Bit.Core.KeyManagement.Models.Api.Request;
 
+/// <summary>
+/// Use this datatype when interfacing with requests to create a separation of concern.
```
mzieniukbw, line 7: I don't know if this comment adds anything. The request and data models is rather known separation of concern, but not documented anywhere in contributing docs. Maybe that's the right place to put it ? Opinions @bitwarden/team-key-management-dev ?

src/Core/KeyManagement/Models/Data/MasterPasswordUnlockAndAuthenticationData.cs
```diff
@@ -4,6 +4,12 @@
 
 namespace Bit.Core.KeyManagement.Models.Data;
 
+/// <summary>
+/// Probably shouldn't be used as we want to make sure that the unlock data and authentication data
+/// can use separate kdf settings.
+///
+/// Should be cleaned up in the near future.
+/// </summary>
```
mzieniukbw: This comment does not belong here. It's better to raise a Jira ticket for anything actionable. Also the unlock data and authentication dats separation is done by design, but without any immediate plan to actually separate Kdf or Salt. Please leave this to KM team, i will discuss this with the team on the priority.

Patrick-Pimentel-Bitwarden:
@claude Do a review of my work please!

src/Identity/Controllers/AccountsController.cs
```diff
@@ -191,10 +205,11 @@ private RegisterFinishResponseModel ProcessRegistrationResult(IdentityResult res
...
 
-        foreach (var error in result.Errors.Where(e => e.Code != "DuplicateUserName"))
-        {
-            ModelState.AddModelError(string.Empty, error.Description);
-        }
+        if (result.Errors != null)
+            foreach (var error in result.Errors.Where(e => e.Code != "DuplicateUserName"))
+            {
+                ModelState.AddModelError(string.Empty, error.Description);
+            }
```
mzieniukbw: Coding style dictates we use brackets:
```
if (result.Errors != null)
{
    // foreach here
}
```

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -72,29 +91,112 @@ public RegisterFinishTokenType GetTokenType()
...
+        // PM-28143 - Remove this check
+        ThrowIfExistsAndHashIsNotEqual(MasterPasswordAuthentication, MasterPasswordHash);
+
+        // 1. Access Token Presence Verification Check
+        switch (GetTokenType())
+        {
+            case RegisterFinishTokenType.EmailVerification:
+                if (string.IsNullOrEmpty(EmailVerificationToken))
+                {
+                    throw new BadRequestException("Email verification token absent when processing register/finish.");
```
mzieniukbw: ⛏️  Like claude already suggested, which i completely forgot, we should return `ValidationResult` indicating which field is incorrect, instead of returning `BadRequestException` in `Validate`.
This also applies to `ThrowIfExistsAndHashIsNotEqual`

mzieniukbw (dismissed):
Looks good from KM side, with few nit comments to look at.

src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
quexten: Agreed, this comment doesn't really add anything.

src/Core/KeyManagement/Models/Data/MasterPasswordUnlockAndAuthenticationData.cs
quexten: +1 from my side. It would be OK to call it deprecated, but jira ticket is the right way to go. This DTO is left over as tech debt from before the separation. I also think that we should not add "Probably". Either it should or shouldn't.

src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
mzieniukbw: Also, worth to note, there is a PR that mentions this in contributing docs, so this comment and all other about the separation of concern is now redundant. https://github.com/bitwarden/contributing-docs/pull/740/changes#diff-360b9394971135ffb743be955a3db753636cafe84b67bcfdd052b77903511556R164

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -72,29 +91,112 @@ public RegisterFinishTokenType GetTokenType()
...
+        else
+        {
+            kdfValidationResults = KdfSettingsValidator.Validate(
+                Kdf ?? throw new BadRequestException($"{nameof(Kdf)} not found on RequestModel"),
+                KdfIterations ?? throw new BadRequestException($"{nameof(KdfIterations)} not found on RequestModel"),
+                KdfMemory,
+                KdfParallelism);
+        }
+
+        return kdfValidationResults;
```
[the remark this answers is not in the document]
Patrick-Pimentel-Bitwarden: Not an actual issue, it is designed to be this way. Unlock and Authentication can use separate salts.

src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs
```diff
@@ -72,29 +91,112 @@ public RegisterFinishTokenType GetTokenType()
...
+            default:
+                throw new BadRequestException("Invalid registration finish request");
+        }
+
+        // 2. Validate kdf settings.
+
+        IEnumerable<ValidationResult> kdfValidationResults;
+        if (MasterPasswordUnlock != null && MasterPasswordAuthentication != null)
+        {
+            kdfValidationResults = KdfSettingsValidator.Validate(MasterPasswordUnlock.ToData().Kdf);
```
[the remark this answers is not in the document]
Patrick-Pimentel-Bitwarden: not an issue.

src/Identity/Controllers/AccountsController.cs
```diff
@@ -145,40 +141,55 @@ public async Task<IActionResult> PostRegisterVerificationEmailClicked([FromBody]
...
         switch (model.GetTokenType())
         {
             case RegisterFinishTokenType.EmailVerification:
-                identityResult =
-                    await _registerUserCommand.RegisterUserViaEmailVerificationToken(user, model.MasterPasswordHash,
-                        model.EmailVerificationToken);
-
+                identityResult = await _registerUserCommand.RegisterUserViaEmailVerificationToken(
+                    user,
+                    masterPasswordHash,
```
quexten: After discussion within KM, one of the issues here is that "masterPasswordHash" is a very confusing naming, since there is three different values this could refer to, the client-local hash, the shared hash that the client sends to the server, and then the database stored server-hash (of the shared hash). The solution here is to pass through the masterPasswordAuthenticationData until the lowest layer. 

Can we either change the `RegisterUserViaEmailVerificationToken` function to work with the high level struct, or at least make a ticket and link it?

src/Identity/Controllers/AccountsController.cs
```diff
@@ -145,40 +141,55 @@ public async Task<IActionResult> PostRegisterVerificationEmailClicked([FromBody]
...
     {
-        var user = model.ToUser();
+        User user = model.ToUser();
 
         // Users will either have an emailed token or an email verification token - not both.
-        IdentityResult identityResult = null;
+        IdentityResult? identityResult = null;
+
+        // PM-28143 - Just use the MasterPasswordAuthenticationData.MasterPasswordAuthenticationHash
+        string masterPasswordHash = model.MasterPasswordAuthentication?.MasterPasswordAuthenticationHash
```
quexten: Please rename to "masterPasswordAuthenticationHash", or just use the MasterPasswordAuthentication struct all the way through. Even if it is not on the request model, you can still construct it from the email and kdf that are on the request model.

src/Identity/Controllers/AccountsController.cs
eligrubb: @quexten My [PR for V2 account registration](https://github.com/bitwarden/server/pull/6798), which stacks on this PR, replaces the masterPasswordHash with a high level data structure that includes the MasterPasswordAuthenticationData struct. Along with the naming improvements you mention, this data is also used to update the User entity when [calling `SetV2AccountCryptographicStateAsync`](https://github.com/bitwarden/server/pull/6798/changes#diff-45af4bd4ee2a2573f21c467d392896aabf3e3e862619c17e5a5e23e7073e87c3R342).

I actually [removed the masterPasswordHash variable](https://github.com/bitwarden/server/pull/6798/changes#diff-d23fef2a55cc8602622925f7dca275305672a576384e77b1d1e616282e1b6999L150) in the PR.

quexten: Thanks, in that case i'm closing the comment here!

test/IntegrationTestCommon/Factories/IdentityApplicationFactory.cs
```diff
@@ -239,14 +240,13 @@ public async Task<User> RegisterNewIdentityFactoryUserAsync(
         {
             var unlock = requestModel.MasterPasswordUnlock;
             // PM-28143 - Once UserSymmetricKey is removed and UnlockData is required, delete the fallback to UserSymmetricKey below.
-            var masterKeyWrappedUserKey = !string.IsNullOrWhiteSpace(unlock.MasterKeyWrappedUserKey)
-                ? unlock.MasterKeyWrappedUserKey
-                : (string.IsNullOrWhiteSpace(requestModel.UserSymmetricKey) ? "user_symmetric_key" : requestModel.UserSymmetricKey);
+            // Always force a valid encrypted string for tests to avoid model validation failures.
+            var masterKeyWrappedUserKey = DefaultEncryptedString;
```
mzieniukbw: ⚠️ This doesn't sound correct, to override all of the test's user key, just because the tests themselves have invalid encrypted string.
This requires fixing the tests that use this function.

src/Identity/Controllers/AccountsController.cs
Patrick-Pimentel-Bitwarden: fixed with [93c9631](https://github.com/bitwarden/server/pull/6715/commits/93c9631a75d9844a18af023eff4aa4bceee745f4)

Patrick-Pimentel-Bitwarden: I don't think I want to reconstruct it right now in the interest of trying to keep things simpler. Is that okay with you to just do the rename?

As an aside, would you mind pointing me to where I would look to see an example of how to reconstruct it in the server so I can learn?

test/IntegrationTestCommon/Factories/IdentityApplicationFactory.cs
Patrick-Pimentel-Bitwarden: Yes I agree it shouldn't override. However the tests will be addressed in PM-28143 and will have to be updated to use the new data types then.

src/Identity/Controllers/AccountsController.cs
quexten: Sounds reasonable.

What I meant by reconstruction is that the authentication data is just the "MasterPasswordHash", KDF, salt (lower-cased and trimmed email) packed into a struct. You can just directly make it with something like:
```
MasterPasswordAuthenticationData
        {
            Kdf = Kdf {...}
            MasterPasswordAuthenticationHash = MasterPasswordHash,
            Salt = email.toLower().trim()
        };
```

mzieniukbw (approved):

ike-kottlowski (approved):

mzieniukbw (approved):

PR #6739, opened 2025-12-16
[PM-29555] Add self-revoke endpoint for declining organization data ownership policy

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29555

## 📔 Objective

Adds `PUT /organizations/{orgId}/users/revoke-self` endpoint that allows organization users to self-revoke when declining the `OrganizationDataOwnership` policy.

The endpoint verifies that the policy is enabled, requires the user to be in a Confirmed state, and blocks Owners and Admins from self-revoking. Adds a new `OrganizationUser_SelfRevoked` event for audit logging.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -635,6 +639,19 @@ public async Task RevokeAsync(Guid orgId, Guid id)
         await RestoreOrRevokeUserAsync(orgId, id, _revokeOrganizationUserCommand.RevokeUserAsync);
     }
 
+    [HttpPut("revoke-self")]
+    [Authorize<MemberOrProviderRequirement>]
```
eliykat: You have to be an organization member to be revoked, so this authorization requirement shouldn't include providers. You may have to make a `MemberRequirement` if we don't have one already - copy/paste this one but omit the provider check.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/SelfRevokeUser/SelfRevokeOrganizationUserCommand.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        if (organizationUser.Type is OrganizationUserType.Owner or OrganizationUserType.Admin)
+        {
+            throw new BadRequestException("Owners and Admins are exempt from the organization data ownership policy.");
+        }
+
+        if (organizationUser.Status is not OrganizationUserStatusType.Confirmed)
+        {
+            throw new BadRequestException("User must be confirmed to self-revoke.");
+        }
```
eliykat: I suggest moving this to a `PolicyRequirement` method, e.g.

```c#
    public bool EligibleForMigration(Guid organizationId)
    {
        return _policyDetails.FirstOrDefault(p => organizationId == p.OrganizationId)
            ?.HasStatus([OrganizationUserStatusType.Confirmed]) ?? false;
    }
```

Noting that this will also handle the policy being disabled and exempt roles.

The benefit here is that when we remove the owner/admin exemption, we won't need to update this command as well.

Then this command would look like:

```c#
// get policy requirement, then:
if (!policyRequirement.EligibleForMigration(orgId))
{
  // fail
}
```

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/SelfRevokeUser/SelfRevokeOrganizationUserCommand.cs
eliykat, line 1: Please have this command return a CommandResult. See https://github.com/bitwarden/server/blob/1eb396cb401f2c30732671d2de792f0fd9e6a776/src/Core/AdminConsole/Utilities/v2/Results/CommandResult.cs#L23-L31.

Note the implicit operators, so this method can just return `new None()` or `new SomeError("MyErrorMessage")` and it'll be wrapped automatically.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/SelfRevokeUser/SelfRevokeOrganizationUserCommand.cs
eliykat: That said, now that we allow revoking oneself, and we know this logic is going to change in the future, we should make sure that the last owner cannot revoke themselves. That would leave them in a stuck state (not able to decline the migration but also not wanting to accept), but it should never happen because the owner chose to enable this policy. Nevertheless I think that is better than accidentally revoking the last owner and bricking the organization.

You may be able to use the `HasConfirmedOwnersExceptQuery` for this. See the v2 revoke validator.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -72,6 +72,21 @@ public bool RequiresDefaultCollectionOnConfirm(Guid organizationId)
     {
         return _policyDetails.Any(p => p.OrganizationId == organizationId);
     }
+
+    /// <summary>
+    /// Determines if a user is eligible for self-revocation under the Organization Data Ownership policy.
```
eliykat, line 88: Worth noting something like: "Self-revoke is used to opt out of migrating the user's personal vault to the organization pursuant to this policy." Just to explain how self-revoke relates to this policy.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/SelfRevokeUser/ISelfRevokeOrganizationUserCommand.cs
```diff
@@ -7,7 +9,12 @@
 public interface ISelfRevokeOrganizationUserCommand
 {
     /// <summary>
-    /// Revokes a user from an organization. Validates policy is enabled and user is eligible (not Owner/Admin).
+    /// Revokes a user from an organization.
+    /// Validates the OrganizationDataOwnership policy is enabled and applies to the user (currently Owners/Admins are exempt),
+    /// the user is a confirmed member, and prevents the last owner from revoking themselves.
```
eliykat: I worry about including this much implementation detail in the xmldoc; generally callers don't need to know what checks are done internally. However if you think they're useful to include, I think the `<remarks>` section is better for this.

eliykat (dismissed):
Some non-blocking comments on comments. 😀

eliykat (approved):

PR #5976, opened 2025-06-17
BRE-917 Update to Alpine base

aj-bw:
## 🎟️ Tracking
[BRE-917](https://bitwarden.atlassian.net/browse/BRE-917)

## 📔 Objective
update the base container image used throughout our server project to use an alpine base

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[BRE-917]: https://bitwarden.atlassian.net/browse/BRE-917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6745, opened 2025-12-17
[bre-1427] resolve intermittent admin build timeouts

aj-bw:
## 🎟️ Tracking
[bre-1427](https://bitwarden.atlassian.net/browse/bre-1427)

## 📔 Objective
In #5976, the new Admin service node-build stage was missing the `--platform=$BUILDPLATFORM` flag. This causes the ARM64 and ARMv7 builds to run npm ci under QEMU emulation on the AMD64 GitHub Actions runner, rather than executing natively like all other build stages.

  The emulated npm installation is significantly slower (5-10x) and intermittently times out after 45 minutes, causing workflow cancellations. This affects only the Admin service, as it's the only one with a Node.js build stage. The https://github.com/bitwarden/server/actions/workflows/build.yml?query=is%3Acancelled show this occurs multiple times daily, creating confusion and delays for QA.

### DIFF
--- a/src/Admin/Dockerfile
+++ b/src/Admin/Dockerfile
@@ -1,7 +1,7 @@
 ###############################################
 #           Node.js build stage               #
 ###############################################
-FROM node:20-alpine3.21 AS node-build
+FROM --platform=$BUILDPLATFORM node:20-alpine3.21 AS node-build
 
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


aj-bw:
@claude can you provide a PR review for this

addisonbeck (approved):

PR #6751, opened 2025-12-18
[PM-20109] Increase 2FA WebAuthn Security Key Limit

enmande:
## 🎟️ Tracking

[PM-20109](https://bitwarden.atlassian.net/browse/PM-20109)

🔗 This coordinates with `clients` PR [18040](https://github.com/bitwarden/clients/pull/18040)

## 📔 Objective

- Refactor WebAuthn credential limit for 2FA to be configurable. 
  - _Default_ configurations are supplied at design time: 10 for premium, 5 for non-premium.
  - `GlobalSettings.WebAuthn` can also be used to override these defaults.
- Introduce boundary and persistence-time validation of credential limits, allowing the client to be made unaware of limits.
- Scope of work involves updating the `UserService`; test cases are added.

### Limit Exceeded

When a user at their credential limit attempts to register a new credential, boundary checking will provide an error.

🪪 Max credentials configured for current user: **2**.

https://github.com/user-attachments/assets/40efb83c-3449-4153-95f5-f59e0412d71f

### All Registered Credentials Function

When a user is in a state where they have more credentials registered than are allowed, they can not register additional credentials, but all registered credentials continue to function for 2FA.

🪪 Max credentials configured for current user type: **2**.
🪪 Number of credentials currently configured for user: **3**.

1. 3 Max were configured.
2. User set a third credential. 
3. 2 Max was configured.
4. User is able to log in successfully.

https://github.com/user-attachments/assets/bd63149a-498d-4867-ab81-414653cf322d

### All Registered Credentials Are Removable

When a user is in a state where they have more credentials registered than are allowed, including _n + 2_ states where removing a single credential would still be above the allowed limit, any credential can be removed.

🪪 Max credentials configured for current user type: **1**.

https://github.com/user-attachments/assets/68617ae7-6329-4e16-ac97-bcd19f8dda5a

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-20109]: https://bitwarden.atlassian.net/browse/PM-20109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Services/Implementations/UserService.cs
+++ b/src/Core/Services/Implementations/UserService.cs
@@ -374,8 +374,8 @@ public class UserService : UserManager<User>, IUserService
         var maximumAllowedCredentialCount = await _hasPremiumAccessQuery.HasPremiumAccessAsync(user.Id)
             ? _globalSettings.WebAuthn.PremiumMaximumAllowedCredentials
             : _globalSettings.WebAuthn.NonPremiumMaximumAllowedCredentials;
-        // Allow 'pending' requests -- including the incoming.
-        if (provider.MetaData.Count(metadata => !metadata.Key.Equals("pending", StringComparison.OrdinalIgnoreCase)) >=
+        // Count only saved credentials ("Key{id}") toward the limit.
+        if (provider.MetaData.Count(k => k.Key.StartsWith("Key")) >=
             maximumAllowedCredentialCount)
         {
             throw new BadRequestException("Maximum allowed WebAuthn credential count exceeded.");
@@ -423,8 +423,8 @@ public class UserService : UserManager<User>, IUserService
         var maximumAllowedCredentialCount = await _hasPremiumAccessQuery.HasPremiumAccessAsync(user.Id)
             ? _globalSettings.WebAuthn.PremiumMaximumAllowedCredentials
             : _globalSettings.WebAuthn.NonPremiumMaximumAllowedCredentials;
-        // Allow 'pending' requests -- including the incoming.
-        if (provider.MetaData.Count(metadata => !metadata.Key.Equals("pending", StringComparison.OrdinalIgnoreCase)) >=
+        // Count only saved credentials ("Key{id}") toward the limit.
+        if (provider.MetaData.Count(k => k.Key.StartsWith("Key")) >=
             maximumAllowedCredentialCount)
         {
             throw new BadRequestException("Maximum allowed WebAuthn credential count exceeded.");


test/Core.Test/Services/UserServiceTests.cs
```diff
@@ -594,6 +598,255 @@ private static void SetupUserAndDevice(User user,
             user.MasterPassword = null;
         }
     }
+
+    [Theory, BitAutoData]
```
ike-kottlowski: 🎨 : Possible to combine Premium tests
``` csharp
[Theory]
[BitAutoData(true)]
[BitAutoData(false)]
public async Task StartWebAuthnRegistrationAsync_ExceedsLimitForPremiumStatus_ThrowsBadRequestException(
  bool HasPremium, 
  SutProvider<UserService> sutProvider, 
  User user)
{
  ...
  user.Premium = HasPremium;
  ...
}
```

src/Core/Services/Implementations/UserService.cs
ike-kottlowski, line 1: ❓: 
Scenario: When a user no longer has premium then tries to authenticate MFA with a passkey out of available range. Will a user at any point not be able to use a previously registered passkey to authenticate MFA?

I don't see anything that indicates this would happen, but I wanted to double check.

enmande: Good question. This is exactly the scenario tested in "All Registered Credentials Function" video above. At no point should any credential (regardless of range) cease to function once it has been successfully registered. The enforcement is entirely around registration of a credential based on the configuration "snapshot" at configuration time.

test/Core.Test/Services/UserServiceTests.cs
enmande: Great suggestion, thank you! Addressed in f09e1a6.

src/Core/Services/Implementations/UserService.cs
```diff
@@ -364,6 +370,17 @@ public async Task<CredentialCreateOptions> StartWebAuthnRegistrationAsync(User u
             provider.MetaData = new Dictionary<string, object>();
         }
 
+        // Boundary validation to provide a better UX. There is also second-level enforcement at persistence time.
+        var maximumAllowedCredentialCount = await _hasPremiumAccessQuery.HasPremiumAccessAsync(user.Id)
+            ? _globalSettings.WebAuthn.PremiumMaximumAllowedCredentials
+            : _globalSettings.WebAuthn.NonPremiumMaximumAllowedCredentials;
+        // Allow 'pending' requests -- including the incoming.
+        if (provider.MetaData.Count(metadata => !metadata.Key.Equals("pending", StringComparison.OrdinalIgnoreCase)) >=
+            maximumAllowedCredentialCount)
```
JaredSnider-Bitwarden: ❓ : I was reviewing the clients PR and I wanted to understand the early validation put in place upon read start so I came here to see this. I'm not sure I understand the comment above this line. The code is checking if non pending keys are greater than or equal to the max allowed credential limit. Thus, it's technically excluding pending requests against the limit. 

Could we rewrite this to say something like: "// Exclude 'pending' requests from count (including the incoming pending request)."

enmande: @JaredSnider-Bitwarden Thanks, this is a great point, and it got me thinking.
It would be better if this were a _positive_ check for saved credentials, not a _negative_ check for specific requests. This also more closely follows the filter for the `fido2` credential operation invocations, is clearer, and increases maintainability (i.e., no latent bugs if someone later adds a different key type).
So I have updated this at both sites in ff15e55.
Additionally, while I had written the lambda as `metadata`, I've re-spelled it to `k` to be more consistent with the surrounding code.
Let me know what you think.

JaredSnider-Bitwarden: I find that much clearer! Nice improvement! Thank you!

PR #6759, opened 2025-12-18
[PM-28747] Storage limit bypass for enforce organization ownership policy

shane-melton:
## 🎟️ Tracking

[PM-28747](https://bitwarden.atlassian.net/browse/PM-28747)

## 📔 Objective

Check the organization policies for the `OrganizationDataOwnership` during share operations and skip the storage limit checks when the policy is enabled. This reduces frictions for users that have attachments that need to be transferred to an enforcing organization during the migration period. Organization administrators will need to deal with any storage exceedance afterwards by either removing attachments or increasing their storage plan limit.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28747]: https://bitwarden.atlassian.net/browse/PM-28747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Core.Test/Vault/Services/CipherServiceTests.cs
+++ b/test/Core.Test/Vault/Services/CipherServiceTests.cs
@@ -1279,6 +1279,68 @@ public class CipherServiceTests
             Arg.Is<IEnumerable<Cipher>>(arg => !arg.Except(ciphers).Any()));
     }
 
+    [Theory, BitAutoData]
+    public async Task ShareManyAsync_StorageLimit_Enforced_WhenFeatureFlagDisabled(SutProvider<CipherService> sutProvider,
+        IEnumerable<CipherDetails> ciphers, Guid organizationId, List<Guid> collectionIds)
+    {
+        sutProvider.GetDependency<IOrganizationRepository>().GetByIdAsync(organizationId)
+            .Returns(new Organization
+            {
+                Id = organizationId,
+                PlanType = PlanType.EnterpriseAnnually,
+                UsePolicies = true,
+                MaxStorageGb = 3,
+                Storage = 3221225472 // 3 GB used, so 0 remaining
+            });
+        ciphers.FirstOrDefault().Attachments =
+            "{\"attachment1\":{\"Size\":\"250\",\"FileName\":\"superCoolFile\","
+            + "\"Key\":\"superCoolFile\",\"ContainerName\":\"testContainer\",\"Validated\":false}}";
+
+        var cipherInfos = ciphers.Select(c => (c,
+            (DateTime?)c.RevisionDate));
+        var sharingUserId = ciphers.First().UserId.Value;
+
+        sutProvider.GetDependency<IFeatureService>().IsEnabled(FeatureFlagKeys.MigrateMyVaultToMyItems).Returns(false);
+
+        var exception = await Assert.ThrowsAsync<BadRequestException>(() =>
+            sutProvider.Sut.ShareManyAsync(cipherInfos, organizationId, collectionIds, sharingUserId)
+        );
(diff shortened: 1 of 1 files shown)


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/OrganizationDataOwnershipPolicyRequirement.cs
```diff
@@ -72,6 +72,14 @@ public bool RequiresDefaultCollectionOnConfirm(Guid organizationId)
     {
         return _policyDetails.Any(p => p.OrganizationId == organizationId);
     }
+
+    /// <summary>
+    /// Determines if the policy is enforced by the specified organization.
+    /// </summary>
+    public bool EnforcedByOrg(Guid organizationId)
```
shane-melton: ℹ️ This seemed like the appropriate spot to check/confirm the specific organization that is enforcing the policy in the new requirements API. If there is another preferred approach let me know!

eliykat: This is correct! However I would make the getter more specific to your use case, then you can move your comment here as well. The idea is that these `IPolicyRequirement`s describe the effect of the policy.

We should also filter by Confirmed users.

Full example

```c#
    /// <summary>
    /// Ignore storage limits if the organization has data ownership policy enabled.
    /// Allows users to seamlessly migrate their data into the organization without being blocked by storage limits.
    /// Organization admins will need to manage storage after migration should overages occur.
    /// </summary>
    public bool IgnoreStorageLimitsOnMigration(Guid organizationId)
    {
        return _policyDetails.Any(p => p.OrganizationId == organizationId &&
                                       p.OrganizationUserStatus == OrganizationUserStatusType.Confirmed);
    }
```

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -999,18 +1000,53 @@ private async Task ValidateCipherCanBeShared(
...
+                   requirement.EnforcedByOrg(organization.Id);
         }
 
-        ValidateCipherLastKnownRevisionDate(cipher, lastKnownRevisionDate);
+        var policies =
+            await _policyService.GetPoliciesApplicableToUserAsync(userId,
+                PolicyType.OrganizationDataOwnership);
+
+        return policies.Any(p => p.OrganizationId == organization.Id);
     }
```
eliykat: This can be simplified:

- put it behind your migration flag, not the PolicyRequirements flag (that is for migrating existing code, but this is new code, so it can go behind the flag for the feature it relates to)
- therefore, you can also drop the `policyService` call, just use the new PolicyRequirementQuery
- no need to check the requirement.State, just use the new method you added

shane-melton: Thanks for the suggestions! I wasn't sure the policy requirements API was ready for use since the flag appeared to be disabled still. Implemented changes in [1f690c6](https://github.com/bitwarden/server/pull/6759/commits/1f690c6d2280f51f14d1da8b7a0da6754a6a1717)

eliykat (approved):
Nicely done.

src/Core/Vault/Services/Implementations/CipherService.cs
eliykat: Yep that flag needs to be cleaned up! This looks good, thanks.

PR #6787, opened 2025-12-31
[PM-29598] Create Subscription Upgrade Endpoint

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29598

## 📔 Objective

Objective: Create a backend endpoint to process the upgrade from a Premium subscription to a new Organization plan, ensuring the existing subscription is modified rather than creating a new one.

- **Endpoint: POST account/billing/vnext/upgrade**
- **Command: UpgradePremiumToOrganizationCommand**
- **Implementation Details:**
  1. The command will receive the user's ID and the details of the target organization plan.
  2. It will first validate that the user has an active, valid Premium subscription.
  3. Using the Stripe API, it will fetch the user's current subscription.
  4. It will then add new subscription items (the same currently added for creating the new organization subscription) and marking premium subscription items for delete
  5. A trial period will be applied to the updated subscription by setting the trial_end property date
  6. premium_upgrade_metadata will be added to the Stripe subscription object. This metadata should store the ID of the original Premium plan to facilitate a potential reversion.

**Given** a user has an active Premium subscription

**When** they submit a request to upgrade to an organization

**Then** the system updates their Stripe subscription, applies a trial, and adds reversion metadata.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationConfirmation/SendOrganizationConfirmationCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationConfirmation/SendOrganizationConfirmationCommand.cs
@@ -13,7 +13,7 @@ public class SendOrganizationConfirmationCommand(IMailer mailer, GlobalSettings 
     private const string _titleThird = "!";
 
     private static string GetConfirmationSubject(string organizationName) =>
-        $"You Have Been Confirmed To {organizationName}";
+        $"You can now access items from {organizationName}";
     private string GetWebVaultUrl(bool accessSecretsManager) => accessSecretsManager
         ? globalSettings.BaseServiceUri.VaultWithHashAndSecretManagerProduct
         : globalSettings.BaseServiceUri.VaultWithHash;
--- a/test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationConfirmation/SendOrganizationConfirmationCommandTests.cs
+++ b/test/Core.Test/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationConfirmation/SendOrganizationConfirmationCommandTests.cs
@@ -240,6 +240,6 @@ public class SendOrganizationConfirmationCommandTests
         }
     }
 
-    private static string GetSubject(string organizationName) => $"You Have Been Confirmed To {organizationName}";
+    private static string GetSubject(string organizationName) => $"You can now access items from {organizationName}";
 
 }


src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Core.Billing.Enums;
+
+namespace Bit.Api.Billing.Models.Requests.Premium;
+
+public class UpgradePremiumToOrganizationRequest
+{
+    [Required]
+    public required PlanType PlanType { get; set; }
```
amorask-bitwarden: ❌ Our newer client-side code is not operating off of `PlanType`. Let's send `Tier` and `Cadence` like we do here: https://github.com/bitwarden/server/blob/main/src/Api/Billing/Models/Requests/Organizations/OrganizationSubscriptionPurchaseRequest.cs

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    [Range(1, int.MaxValue)]
+    public int Seats { get; set; }
```
amorask-bitwarden: ❌ When upgrading from Premium, we're only going to have 1 seat. The [designs](https://www.figma.com/design/nuFrzHsgEoEk2Sm8fWOGuS/Premium---business-upgrade-flows?node-id=3053-119055&m=dev) do not support seat increases at the moment.

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    public bool PremiumAccess { get; set; } = false;
```
amorask-bitwarden: ❓ What current plan still has premium access?

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    [Range(0, 99)]
+    public int Storage { get; set; } = 0;
```
amorask-bitwarden: ❌ This can be removed as the [designs](https://www.figma.com/design/nuFrzHsgEoEk2Sm8fWOGuS/Premium---business-upgrade-flows?node-id=3053-119055&m=dev) also don't support storage selection in the upgrade modal.

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    public DateTime? TrialEndDate { get; set; }
```
amorask-bitwarden: ❓ Why are we passing this in?

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+        // Validate that the user has an active Premium subscription
+        if (!user.Premium)
+        {
+            return new BadRequest("User does not have an active Premium subscription.");
+        }
+
+        if (string.IsNullOrEmpty(user.GatewaySubscriptionId))
+        {
+            return new BadRequest("User does not have a Stripe subscription.");
+        }
```
amorask-bitwarden: ⛏️ `if (user is { Premium: true, GatewaySubscriptionId: not null and not "" })`

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        if (trialEndDate.HasValue && trialEndDate.Value < DateTime.UtcNow)
+        {
+            return new BadRequest("Trial end date cannot be in the past.");
+        }
+
+        // Fetch the current Premium subscription from Stripe
+        var currentSubscription = await subscriberService.GetSubscriptionOrThrow(user, new SubscriptionGetOptions
+        {
+            Expand = ["items.data.price"]
```
amorask-bitwarden: ❌ This is not expandable: https://docs.stripe.com/api/subscriptions/object?api-version=2025-08-27.basil&lang=cli#subscription_object-items-data-price

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+            return new BadRequest("Seats must be at least 1.");
+        }
+
+        if (trialEndDate.HasValue && trialEndDate.Value < DateTime.UtcNow)
+        {
+            return new BadRequest("Trial end date cannot be in the past.");
+        }
+
+        // Fetch the current Premium subscription from Stripe
+        var currentSubscription = await subscriberService.GetSubscriptionOrThrow(user, new SubscriptionGetOptions
```
amorask-bitwarden: ⛏️ If this is the only use of the `SubscriberService` you can use `StripeAdapter` since that will automatically throw an exception if the subscription does not exist.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+            subscriptionUpdateOptions.TrialEnd = trialEndDate.Value;
+        }
+
+        // Update the subscription in Stripe
+        await stripeAdapter.UpdateSubscriptionAsync(currentSubscription.Id, subscriptionUpdateOptions);
+
+        // Update user record
+        user.PremiumExpirationDate = trialEndDate ?? currentSubscription.GetCurrentPeriodEnd();
+        user.RevisionDate = DateTime.UtcNow;
+        await userService.SaveUserAsync(user);
```
amorask-bitwarden: ❌ This results in a `User` record being attached to an `Organization` subscription, which would not be renderable or usable on the client. The idea is to create an `Organization` record and attach the modified subscription to that `Organization` while removing the subscription from the `User`.

src/Api/Billing/Controllers/VNext/AccountBillingVNextController.cs
```diff
@@ -88,4 +89,16 @@ public async Task<IResult> GetLicenseAsync(
...
         return TypedResults.Ok(response);
     }
+
+    [HttpPost("upgrade")]
+    [InjectUser]
+    public async Task<IResult> UpgradePremiumToOrganizationAsync(
+        [BindNever] User user,
+        [FromBody] UpgradePremiumToOrganizationRequest request)
+    {
+        var (planType, seats, premiumAccess, storage, trialEndDate) = request.ToDomain();
```
sbrown-livefront: :x: `trailEndDate` can be removed here too since we won't be retrieving it from the request.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            return new BadRequest("User does not have an active Premium subscription.");
+        }
+
+        if (seats < 1)
+        {
+            return new BadRequest("Seats must be at least 1.");
+        }
+
+        if (trialEndDate.HasValue && trialEndDate.Value < DateTime.UtcNow)
```
sbrown-livefront: ❓ Do premium subscriptions have trails that this would be required? If this is to be used for the new organization, I think we would calculated the trial period in this command itself, not from the request.

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+using Bit.Core.Billing.Enums;
+
+namespace Bit.Api.Billing.Models.Requests.Premium;
+
+public class UpgradePremiumToOrganizationRequest
```
sbrown-livefront, line 7: ❓ Would we not include seats in the request(even though we assume it will be 1 for now)?

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        if (storage is > 0 && string.IsNullOrEmpty(targetPlan.PasswordManager.StripeStoragePlanId))
+        {
+            return new BadRequest("The selected plan does not support additional storage.");
+        }
+
+        // Build the list of subscription item updates
+        var subscriptionItemOptions = new List<SubscriptionItemOptions>();
+
+        // Mark existing Premium subscription items for deletion
+        foreach (var item in currentSubscription.Items.Data)
```
sbrown-livefront: ❌ We may need to be more precise in identifying subscription items by narrowing our search specifically to the items we want to delete related to premium (plan id, etc.) I'm not sure if there are some items we'd want to retain.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+            Metadata = new Dictionary<string, string>(currentSubscription.Metadata ?? new Dictionary<string, string>())
+            {
+                ["premium_upgrade_metadata"] = currentSubscription.Items.Data
+                    .Select(item => item.Price.Id)
+                    .FirstOrDefault() ?? "premium"
+            }
+        };
+
+        // Apply trial period if specified
+        if (trialEndDate.HasValue)
```
sbrown-livefront: ❓ Shouldn't this date be calculated using the `TrialPeriodDays` property of the target plan?

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // Build the subscription update options
+        var subscriptionUpdateOptions = new SubscriptionUpdateOptions
+        {
+            Items = subscriptionItemOptions,
+            ProrationBehavior = StripeConstants.ProrationBehavior.None,
+            Metadata = new Dictionary<string, string>(currentSubscription.Metadata ?? new Dictionary<string, string>())
+            {
+                ["premium_upgrade_metadata"] = currentSubscription.Items.Data
```
sbrown-livefront: ❌ We also want the premium sub end date in case of reversion: "previous_period_end_date": "period-end-date"

test/Core.Test/Billing/Premium/Commands/UpgradePremiumToOrganizationCommandTests.cs
```diff
@@ -0,0 +1,516 @@
...
+    {
+        // Arrange
+        user.Premium = false;
+
+        // Act
+        var result = await _command.Run(user, PlanType.TeamsAnnually, 5, false, null, null);
+
+        // Assert
+        Assert.True(result.IsT1);
+        var badRequest = result.AsT1;
```
sbrown-livefront, line 165: 📝 Alex made some great helper methods that are more descriptive [here](https://github.com/bitwarden/server/blob/410e754cd9295327de386b46607b5db8a6e7a2c2/src/Core/Billing/Payment/Models/PaymentMethod.cs#L13). This is out of scope for this PR, but I wonder if we could clean up command logic a bit with it.

sbrown-livefront (changes requested):
A few questions and changes

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
```diff
@@ -0,0 +1,32 @@
...
+                ? PlanType.TeamsMonthly
+                : PlanType.TeamsAnnually,
+            ProductTierType.Enterprise => Cadence == PlanCadenceType.Monthly
+                ? PlanType.EnterpriseMonthly
+                : PlanType.EnterpriseAnnually,
+            _ => throw new InvalidOperationException("Cannot upgrade to an Organization subscription that isn't Families, Teams or Enterprise.")
+        };
+
+    public (PlanType, int, bool, int?, DateTime?) ToDomain() =>
+        (PlanType, 1, false, null, null);
```
amorask-bitwarden: ❌ We don't need to hard-code all these values. The only real thing we need from the request is either the combination of the `Tier` and `Cadence` or the derived `PlanType`.

src/Api/Billing/Controllers/VNext/AccountBillingVNextController.cs
amorask-bitwarden: The only thing we need here is `planType`.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        int? storage,
+        DateTime? trialEndDate) => HandleAsync<None>(async () =>
+    {
+        // Validate that the user has an active Premium subscription
+        if (user is not { Premium: true, GatewaySubscriptionId: not null and not "" })
+        {
+            return new BadRequest("User does not have an active Premium subscription.");
+        }
+
+        if (seats < 1)
```
amorask-bitwarden: `Seats` parameter is not necessary - it will always be 1 seat.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
amorask-bitwarden: ^ correct, this parameter is not necessary.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // Fetch the current Premium subscription from Stripe
+        var currentSubscription = await stripeAdapter.GetSubscriptionAsync(user.GatewaySubscriptionId);
+
+        // Get the target organization plan
+        var targetPlan = await pricingClient.GetPlanOrThrow(targetPlanType);
+
+        // Validate plan supports requested features
+        if (premiumAccess && string.IsNullOrEmpty(targetPlan.PasswordManager.StripePremiumAccessPlanId))
```
amorask-bitwarden: ❓ I'm not sure if my previous question was answered, but what current plans support premium access that would make this necessary?

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            return new BadRequest("The selected plan does not support premium access.");
+        }
+
+        if (storage is > 0 && string.IsNullOrEmpty(targetPlan.PasswordManager.StripeStoragePlanId))
```
amorask-bitwarden: Storage cannot be purchased during an upgrade from Individual to Organization so this parameter is unnecessary.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        if (premiumAccess)
+        {
+            subscriptionItemOptions.Add(new SubscriptionItemOptions
+            {
+                Price = targetPlan.PasswordManager.StripePremiumAccessPlanId,
+                Quantity = 1
+            });
+        }
+
+        if (storage is > 0)
```
amorask-bitwarden: Unneeded after storage param is removed.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
amorask-bitwarden: Can we also get these into the `StripeConstants.MetadataKeys` class?

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+        await stripeAdapter.UpdateSubscriptionAsync(currentSubscription.Id, subscriptionUpdateOptions);
+
+        // Save the organization
+        await organizationRepository.CreateAsync(organization);
+
+        // Create organization API key
+        await organizationApiKeyRepository.CreateAsync(new OrganizationApiKey
+        {
+            OrganizationId = organization.Id,
+            ApiKey = CoreHelpers.SecureRandomString(30),
```
amorask-bitwarden, line 195: ❓ I'm not 100% familiar with this process, but are we supposed to be creating an API key on the fly? What happens during the standard organization creation process?

src/Api/Billing/Models/Requests/Premium/UpgradePremiumToOrganizationRequest.cs
amorask-bitwarden: @sbrown-livefront I recommended not doing so - since the business requirements are for a 1-seat upgrade, I feel like that would primarily just give the client extra, unnecessary responsibility.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        // Build the list of subscription item updates
+        var subscriptionItemOptions = new List<SubscriptionItemOptions>();
+
+        // Mark existing Premium subscription items for deletion
+        // Only delete Premium and storage items, not other potential subscription items
+        foreach (var item in currentSubscription.Items.Data)
+        {
+            var priceId = item.Price.Id;
+            var isPremiumItem = priceId == StripeConstants.Prices.PremiumAnnually;
```
amorask-bitwarden: ❌ This won't work if some of our upcoming changes go out before this does. Like our other newer commands, you'll need to pull the prices from the `PricingClient` and get the Price IDs that way.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
amorask-bitwarden: ❌ I still don't see a need for this block. Storage is always 0. It can be removed.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+                Quantity = storage
+            });
+        }
+
+        // Build the subscription update options
+        var subscriptionUpdateOptions = new SubscriptionUpdateOptions
+        {
+            Items = subscriptionItemOptions,
+            ProrationBehavior = StripeConstants.ProrationBehavior.None,
+            Metadata = new Dictionary<string, string>(currentSubscription.Metadata ?? new Dictionary<string, string>())
```
amorask-bitwarden: ❌ I don't think this is doing what we need it to do. `metadata` is not nullable in Stripe's API, so this will always set the update option's metadata to the current subscription's metadata, which is incorrect.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+                    .FirstOrDefault() ?? "premium",
+                [StripeConstants.MetadataKeys.PreviousPeriodEndDate] = currentSubscription.GetCurrentPeriodEnd()?.ToString("O") ?? string.Empty
+            }
+        };
+
+        // Create the Organization entity
+        var organization = new Organization
+        {
+            Id = CoreHelpers.GenerateComb(),
+            Name = $"{user.Email}'s Organization",
```
amorask-bitwarden: ❌ Looks like we missed something here. In the [designs](https://www.figma.com/design/nuFrzHsgEoEk2Sm8fWOGuS/Premium---business-upgrade-flows?node-id=3053-119055&m=dev), the Organization's name will be passed by the client. We need to update the whole chain to pass that through.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
amorask-bitwarden: Same question still stands.

amorask-bitwarden (changes requested):

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
cyprain-okeke: I think, creating the OrganizationApiKey is necessary. This is a standard pattern across all organization creation commands in the codebase:

  - CloudOrganizationSignUpCommand.cs:269-275 - Creates API key
  - ProviderClientOrganizationSignUpCommand.cs:138-144 - Creates API key
  - SelfHostedOrganizationSignUpCommand - Creates API key
  - ResellerClientOrganizationSignUpCommand - Creates API key

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        // Generate organization ID early to include in metadata
+        var organizationId = CoreHelpers.GenerateComb();
+
+        // Build the subscription update options
+        var subscriptionUpdateOptions = new SubscriptionUpdateOptions
+        {
+            Items = subscriptionItemOptions,
+            ProrationBehavior = StripeConstants.ProrationBehavior.None,
+            Metadata = new Dictionary<string, string>
```
amorask-bitwarden, line 137: ❌ `userId` needs to be removed from the Metadata to unlink the subscription from the `User`.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        // Update cache
+        await applicationCacheService.UpsertOrganizationAbilityAsync(organization);
+
+        // Create OrganizationUser for the upgrading user as owner
+        var organizationUser = new OrganizationUser
+        {
+            OrganizationId = organization.Id,
+            UserId = user.Id,
+            Key = null, // Will need to be set by client
```
amorask-bitwarden: ❌ The creation of the `OrganizationApiKey` looks correct, but this is not. I don't think this is set by the client, I think it's passed in by the client. See the `[CloudOrganizationSignupCommand](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/OrganizationFeatures/Organizations/CloudOrganizationSignUpCommand.cs#L287)`.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,210 @@
...
+            UseTotp = targetPlan.HasTotp,
+            Use2fa = targetPlan.Has2fa,
+            UseApi = targetPlan.HasApi,
+            UseResetPassword = targetPlan.HasResetPassword,
+            SelfHost = targetPlan.HasSelfHost,
+            UsersGetPremium = targetPlan.UsersGetPremium,
+            UseCustomPermissions = targetPlan.HasCustomPermissions,
+            UseScim = targetPlan.HasScim,
+            Plan = targetPlan.Name,
+            Gateway = null,
```
amorask-bitwarden: Last thing: This needs to be `Stripe` in order for the Admin Portal links to the `Customer` and `Subscription` to work.

amorask-bitwarden (changes requested):

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        // Hardcode seats to 1 for upgrade flow
+        const int seats = 1;
+
+        // Fetch the current Premium subscription from Stripe
+        var currentSubscription = await stripeAdapter.GetSubscriptionAsync(user.GatewaySubscriptionId);
+
+        // Get the premium plan to identify which price IDs to delete
+        var premiumPlan = await pricingClient.GetAvailablePremiumPlan();
```
amorask-bitwarden: ❌ I missed this, but what happens if the `User` is not on the currently available premium plan?

cyprain-okeke: This is an excellent catch

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+        const int seats = 1;
+
+        // Fetch the current Premium subscription from Stripe
+        var currentSubscription = await stripeAdapter.GetSubscriptionAsync(user.GatewaySubscriptionId);
+
+        // Get ALL premium plans (including legacy plans) to identify which price IDs to delete
+        var premiumPlans = await pricingClient.ListPremiumPlans();
+        var availablePremiumPlan = premiumPlans.FirstOrDefault(p => p.Available);
+
+        if (availablePremiumPlan == null)
```
amorask-bitwarden: I don't think this is necessary.  Really what we're trying to do here is find the specific Premium plan the user is on. We do that by looking for the subscription item for Password Manager that matches one of the premium plans. Then, when we have that subscription item, we can delete it because it will be transitioning to an organization. Let me know if this makes sense. I think you did something similar on your recent storage command you can reference.

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        // Build the subscription update options
+        var subscriptionUpdateOptions = new SubscriptionUpdateOptions
+        {
+            Items = subscriptionItemOptions,
+            ProrationBehavior = StripeConstants.ProrationBehavior.None,
+            Metadata = new Dictionary<string, string>
+            {
+                [StripeConstants.MetadataKeys.OrganizationId] = organizationId.ToString(),
+                [StripeConstants.MetadataKeys.PreviousPremiumPriceId] = availablePremiumPlan.Seat.StripePriceId,
```
amorask-bitwarden: This is what I mean: The user might not actually have been on this plan. We need the plan the user was on from the premium plans that come back from the Pricing service.

amorask-bitwarden (approved):

src/Core/Billing/Premium/Commands/UpgradePremiumToOrganizationCommand.cs
```diff
@@ -0,0 +1,223 @@
...
+        {
+            Items = subscriptionItemOptions,
+            ProrationBehavior = StripeConstants.ProrationBehavior.None,
+            Metadata = new Dictionary<string, string>
+            {
+                [StripeConstants.MetadataKeys.OrganizationId] = organizationId.ToString(),
+                [StripeConstants.MetadataKeys.PreviousPremiumPriceId] = usersPremiumPlan.Seat.StripePriceId,
+                [StripeConstants.MetadataKeys.PreviousPeriodEndDate] = currentSubscription.GetCurrentPeriodEnd()?.ToString("O") ?? string.Empty,
+                [StripeConstants.MetadataKeys.UserId] = string.Empty // Remove userId to unlink subscription from User
+            }
```
sbrown-livefront, line 145: We will want one more metadata key if the user has additional storage that we will need if they revert. The key can be `PreviousAdditionalStorage` and the value can be the amount we get from `storageItem` quantity above if it exists or 0.

Alex was also right. We'll want to put a `PreviousPremiumUserId` as a metadata key also to easily replace the `UserId` on revert rom the `User` object above.

cyprain-okeke: Added PreviousPremiumUserId to metadata

PR #6791, opened 2026-01-01
[PM-28555] Add DefaultCollectionSemaphore table

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28555
## 📔 Objective

Users are only permitted 1 default user collection ("My Items") for an organization. Today, this is enforced by querying the database and skipping users who already have such a collection, however this is susceptible to time-of-check time-of-use bugs. There is no atomic check or constraint that prevents duplicate collections from being created. This PR adds this guarantee.

A significant challenge is that the unique columns are spread across 2 tables: `Collection.Type` (to identify default collections only) and `CollectionUser.OrganizationUserId`. Additionally, we have many complex sprocs for the `Collection` table, which makes it hard to modify without making a large impact throughout the codebase.

I went with a solution suggested by @mkincaid-bw, which was a semaphore table. Advantages:

- optimistic rather than pessimistic (i.e. reduces locking)
- no impact on existing uses of the Collection table
- easy to migrate existing data to semaphores
- EDD compliant: if we roll back the deployment, the old code will ignore the semaphore table

Changes in this PR:

- create the semaphore table 
- migrate existing data to semaphores (Dapper/MSSQL only, as this feature has not been released for self-hosted yet; if this PR is not merged by 9 Jan we will have to add EF migrations as well)
- add a new repository method to insert _n_ default collections with semaphores; existing uses of `ICollectionRepository.CreateAsync` are migrated to this
- update existing `SqlBulkCopy` method to use semaphores
- DRY up default collection arrangement logic (currently repeated between repository implementations)

Design decisions worth pointing out:
- today, the repository filters out users who already have a default collection. This PR makes that the caller's responsibility, mostly so that I could properly test the semaphore behaviour. However, I also think that the repository method was doing too much, so simplifying the repository method and letting callers handle this where needed makes sense.
- the semaphore table contains the minimal amount of columns needed: only the `OrganizationUserId`, and a creation date for debugging purposes. We _could_ have a FK to the Collection as well, but it's not strictly required. I haven't included it for now.
- even though we don't expect any unique constraint violations, this is caught and transformed into a consistent exception for a user-facing error and easier debugging
- we are left with 2 repository methods, one which uses a sproc and one which uses SqlBulkCopy. They are used inconsistently, and this needs to be addressed later; ideally we would use the sproc for 1-1000 insertions and SqlBulkCopy for 1000+. That is out of scope - the only goal here was to get everyone using the semaphore table.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
```diff
@@ -369,90 +369,120 @@ public async Task UpsertDefaultCollectionsAsync(Guid organizationId, IEnumerable
...
-            {
-                return;
-            }
+        await using var connection = new SqlConnection(ConnectionString);
+        connection.Open();
+        await using var transaction = connection.BeginTransaction();
 
+        try
+        {
+
```
r-tome, line 426: :art: Extra whiteline

src/Core/AdminConsole/OrganizationFeatures/Collections/CollectionUtils.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    /// <summary>
+    /// Arranges semaphore, Collection and CollectionUser objects to create default user collections.
+    /// </summary>
+    /// <param name="organizationId">The organization ID.</param>
+    /// <param name="organizationUserIds">The IDs for organization users who need default collections.</param>
+    /// <param name="defaultCollectionName">The encrypted string to use as the default collection name.</param>
+    /// <returns></returns>
+    public static (IEnumerable<DefaultCollectionSemaphore> semaphores,
+        IEnumerable<Collection> collection,
```
r-tome:
```suggestion
        IEnumerable<Collection> collections,
```

src/Core/AdminConsole/OrganizationFeatures/Collections/CollectionUtils.cs
```diff
@@ -0,0 +1,62 @@
...
+
+public static class CollectionUtils
+{
+    /// <summary>
+    /// Arranges semaphore, Collection and CollectionUser objects to create default user collections.
+    /// </summary>
+    /// <param name="organizationId">The organization ID.</param>
+    /// <param name="organizationUserIds">The IDs for organization users who need default collections.</param>
+    /// <param name="defaultCollectionName">The encrypted string to use as the default collection name.</param>
+    /// <returns></returns>
```
r-tome:
```suggestion
    /// <returns>A tuple containing the semaphores, collections, and collection users.</returns>
```

src/Sql/dbo/AdminConsole/Stored Procedures/Collection_CreateDefaultCollections.sql
```diff
@@ -0,0 +1,97 @@
...
+                                 );
+
+    INSERT INTO @CollectionsToInsert
+    (
+        [OrganizationUserId],
+        [CollectionId]
+    )
+    SELECT
+        ou.Id,
+        NEWID()
```
r-tome: ❓ : @bitwarden/dept-dbops is it ok to not use a sequential ID here?

r-tome (dismissed):
Great work! This is a good approach given your explanation

src/Sql/dbo/AdminConsole/Stored Procedures/Collection_CreateDefaultCollections.sql
mkincaid-bw: Correct, we do not use `NEWID()` in our database codebase. Instead, the Guids are created in application code using `CoreHelpers.GenerateComb` and passed into the stored procs. This proc can be updated to use a `TwoGuidIdArray` for the `OrganizationUserIds` and `CollectionIds` to handle this.

src/Sql/dbo/AdminConsole/Stored Procedures/Collection_CreateDefaultCollections.sql
```diff
@@ -0,0 +1,97 @@
...
+    FROM
+        @OrganizationUserIds ou
+
+    BEGIN TRANSACTION;
+
+    BEGIN TRY
+        -- Insert semaphore entries first to obtain the "lock"
+        -- If this fails due to duplicate key, the entire transaction will be rolled back
+        INSERT INTO [dbo].[DefaultCollectionSemaphore]
```
mkincaid-bw: If you update the proc to take a `TwoGuidIdArray` with the `OrganizationUserId` and `CollectionId` (from my comment above), this in memory table can go away and the input parameter can be used directly below.

src/Sql/dbo/AdminConsole/Stored Procedures/Collection_CreateDefaultCollections.sql
```diff
@@ -0,0 +1,97 @@
...
+        [OrganizationUserId],
+        [CollectionId]
+    )
+    SELECT
+        ou.Id,
+        NEWID()
+    FROM
+        @OrganizationUserIds ou
+
+    BEGIN TRANSACTION;
```
mkincaid-bw: 💭 I may have missed it but I don't see where this proc is being called in code. The main concern I have is if Dapper/EF explicitly creates a transaction before calling the proc. If so, this `BEGIN TRANSACTION;` will create a nested transaction and errors/rollbacks might not operate as expected.

src/Infrastructure.Dapper/Repositories/CollectionRepository.cs
eliykat: Addressed ✅

src/Core/AdminConsole/OrganizationFeatures/Collections/CollectionUtils.cs
eliykat: Addressed ✅

src/Core/AdminConsole/OrganizationFeatures/Collections/CollectionUtils.cs
eliykat: Addressed ✅

src/Sql/dbo/AdminConsole/Stored Procedures/Collection_CreateDefaultCollections.sql
eliykat: Good catch! I also didn't know about `TwoGuidIdArray`, that's perfect. I've added that, plus a helper method to build the table in code.

src/Sql/dbo/AdminConsole/Stored Procedures/Collection_CreateDefaultCollections.sql
eliykat: It's called from the Dapper `CollectionRepository.CreateDefaultCollectionsAsync` implementation. I wasn't expressly creating a transaction in the code layer, but I _did_ have try/catch blocks in both the code and sproc, which is a bit messy. I have updated this so that all transaction and error handling is in the code layer only, rather than it being spread between both the code and the sproc.

withinfocus:
I reached out to @bitwarden/dept-dbops this morning and haven't heard back yet but I am not initially understanding this pattern and usage of the database. Logistically-speaking I think it's better to name such a thing regarding its association and to drop the "default" language (we always lead with domain too) but fundamentally why is the database used for this vs. a persistent cache? I'd like to hear about rationale and what was assessed to address this problem; we don't do this anywhere else that I know of and personally I have never used a relational database table for this kind of lock. We should strive to keep logic out of the data tier.

eliykat:
@withinfocus The naming can be adjusted no problem. In terms of the fundamental approach, the thinking was that a database can enforce uniqueness constraints as a matter of data integrity. Same as enforcing a uniqueness constraint on the `User.Email` column. The difficulty here is that the relevant columns are spread across multiple tables, hence the use of the separate table, but that's just a different way of enforcing the same principle.

Let me know how you go with dbops - if you want to revisit this approach we can, but probably best that we have a meeting about it to get alignment before spending more time on it.

eliykat:
After some discussions, we are going to simplify this approach and not seek a database-level guarantee to avoid duplicates. We are going to rely on querying for existing default collections first. If we experience TOCTOU issues in practice then we can look at increasing the isolation level of the transaction, but in any case, we are not going with the semaphore table.

I will open a new PR so there is a tidier commit history and better context for everyone.

PR #6794, opened 2026-01-03
[BEEEP] Strongly typed OrganizationUser objects

eliykat:
## 🎟️ Tracking

N/A - draft/experimental work only

## 📔 Objective

The `OrganizationUser` object is ambiguous.
- It can represent invited, accepted and confirmed users, all of which use different combinations of properties. This is handled through nullable properties, but it's left to the developer to know which are or should be populated.
- It can represent revoked users. Today we treat revoked as a separate status, but it is really an overlay on the other statuses. This is because the OrganizationUser object retains its characteristics of its underlying status. (e.g. an OrgUser that is invited and then revoked still has the arrangement of an invited user, and will look different to an OrgUser who is confirmed and then revoked.)

This PR is a proof of concept for how we might have stronger typed classes for representing OrganizationUsers in the code.

Core ideas:

- separate classes to represent invited, accepted, and confirmed states. These _only_ have the properties that are relevant to that status.
- make revoked a boolean flag on each class, rather than a separate class. This makes it easy to keep track of the "real" status of the user, while still recording whether they are revoked. (I would like to make this change in the database as well but that is out of scope here.)
- mapping methods to transform between each state - for the accepted/confirmed commands
- mapping methods to transform between each state and the OrganizationUser database object - to smooth over saving and loading from the database

For feedback from the team.

Possible additional improvements:
- querying by userId will (should) never return an invited user, so there should be a more restrictive interface or `OneOf` that excludes invited users in that case
- we need additional query classes for each main way to query an orgUser object (by id, by userId, by orgId, etc)
- may consider moving the mapping methods off the objects themselves to prevent general mis/use

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
@@ -3,6 +3,7 @@
 
 using Bit.Core.AdminConsole.Enums;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
+using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Models;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Enforcement.AutoConfirm;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
@@ -23,6 +24,7 @@ public class ConfirmOrganizationUserCommand : IConfirmOrganizationUserCommand
 {
     private readonly IOrganizationRepository _organizationRepository;
     private readonly IOrganizationUserRepository _organizationUserRepository;
+    private readonly IGetOrganizationUserQuery _getOrganizationUserQuery;
     private readonly IUserRepository _userRepository;
     private readonly IEventService _eventService;
     private readonly IMailService _mailService;
@@ -39,6 +41,7 @@ public class ConfirmOrganizationUserCommand : IConfirmOrganizationUserCommand
     public ConfirmOrganizationUserCommand(
         IOrganizationRepository organizationRepository,
         IOrganizationUserRepository organizationUserRepository,
+        IGetOrganizationUserQuery getOrganizationUserQuery,
         IUserRepository userRepository,
         IEventService eventService,
         IMailService mailService,
@@ -54,6 +57,7 @@ public class ConfirmOrganizationUserCommand : IConfirmOrganizationUserCommand
     {
         _organizationRepository = organizationRepository;
         _organizationUserRepository = organizationUserRepository;
+        _getOrganizationUserQuery = getOrganizationUserQuery;
         _userRepository = userRepository;
         _eventService = eventService;
         _mailService = mailService;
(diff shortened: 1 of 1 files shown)


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Models/AcceptedOrganizationUser.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        if (!isRevoked && entity.Status != OrganizationUserStatusType.Accepted)
+        {
+            throw new InvalidOperationException($"Cannot create AcceptedOrganizationUser from entity with status {entity.Status}");
+        }
+
+        if (isRevoked)
+        {
+            // Validate that the revoked user's pre-revoked status is Accepted
+            var preRevokedStatus = OrganizationService.GetPriorActiveOrganizationUserStatusType(entity);
```
jrmccannon, line 136: Not something to do now, but this method should be moved to the extension methods

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Models/AcceptedOrganizationUser.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (preRevokedStatus != OrganizationUserStatusType.Accepted)
+            {
+                throw new InvalidOperationException($"Cannot create AcceptedOrganizationUser from revoked entity with pre-revoked status {preRevokedStatus}");
+            }
+        }
+
+        if (!entity.UserId.HasValue)
+        {
+            throw new InvalidOperationException("Cannot create AcceptedOrganizationUser from entity with null UserId");
+        }
```
jrmccannon, line 146: Maybe move this block to a validation method?

Maybe make the validation method on the interface and force each class to implement their own/call common static methods?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/GetOrganizationUserQuery.cs
```diff
@@ -0,0 +1,50 @@
...
+    public async Task<IEnumerable<ITypedOrganizationUser>> GetManyOrganizationUsersAsync(IEnumerable<Guid> organizationUserIds)
+    {
+        var organizationUsers = await organizationUserRepository.GetManyAsync(organizationUserIds);
+
+        return organizationUsers
+            .Select(ConvertToStronglyTypedModel)
+            .ToList();
+    }
+
+    private static ITypedOrganizationUser ConvertToStronglyTypedModel(OrganizationUser organizationUser)
```
jrmccannon, line 34: Could this be made public in order to be used in places outside this query? Especially in places where we might need these types in other queries that return Org users.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Models/AcceptedOrganizationUser.cs
```diff
@@ -0,0 +1,170 @@
...
+using Bit.Core.Services;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Models;
+
+/// <summary>
+/// Represents a user who has accepted their invitation to join an organization but has not yet been confirmed
+/// by an organization administrator. At this stage, the user is linked to a User account but does not yet have
+/// access to encrypted organization data.
+/// </summary>
+public class AcceptedOrganizationUser : ITypedOrganizationUser
```
JimmyVo16, line 13: I’m trying to summarize this at a high level. The main goal is that the consuming code can interact with just these types and call methods to update the status without knowing the underlying tables or fields. For example, if they have an AcceptedOrganizationUser, they can’t even access the email property because it won’t be available.

eliykat: Yes, that's exactly it. Today, you need to read the xmldoc and be familiar with the domain to understand when certain properties will be null. This change uses the type system to enforce this instead. You get a smaller, more focused type that more directly represents the available properties.

This converts run-time errors into compile-time errors. For example, you might do this to find a confirmed user:

```c#
var orgUser = organizationUsers.FirstOrDefault(ou => ou.Email == email);
```

- current state (`OrganizationUser`): this will compile but never find a match at runtime, because `Email` is nulled out
- this PR (`ConfirmedOrganizationUser`): this will not compile because the `Email` property does not exist. The feedback loop is immediate

PR #6797, opened 2026-01-05
[PM-30391] fix for org context on sso provisioning

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30391

## 📔 Objective

Fixed the QA bug where users getting onboarded via SSO were incorrectly blocked when the MjmlWelcomeEmailTemplates feature flag was off. Root Cause: The SSO AccountController had branching logic that called RegisterUser(newUser) without organization context when the feature flag was off, causing domain validation to fail for users joining their own organization. Fix: Simplified the SSO AccountController to always call RegisterSSOAutoProvisionedUserAsync(newUser, organization), ensuring organization context is always available for domain validation.

Changes Made:
AccountController.cs:683-698 - Removed feature flag branching, always use RegisterSSOAutoProvisionedUserAsync


kspearrin:
Note: I don't think `RegisterUser(User user)` is used anywhere or needed anymore, so should it be deleted? Someone from the auth team needs to understand that and make a follow up PR to do it, if so. Should have been done as part of https://github.com/bitwarden/server/pull/6479

ike-kottlowski (approved):

PR #6810, opened 2026-01-08
[PM-22236] Fix invited accounts stuck in intermediate claimed status

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22236

## 📔 Objective

There is a [known defect](https://bitwarden.atlassian.net/browse/PM-22405) where user provisioning via the SSO `AccountController` incorrectly creates an `OrganizationUser` in the `Invited` status, but _also_ links it to a `UserId`. This violates the expected contract for `OrganizationUser` objects. 

This is generally not a problem because the user accepts the invite immediately and moves into a valid accepted state. However, it's not an atomic process, so if it goes off the happy path for any reason, the user will be left in this invalid invited state.

One side effect is that if the user's domain matches a domain claimed by the organization, they will be considered "claimed" in some contexts even though they are in the invited status. This is because we join through the `UserId` assuming that this will be null for invited OrganizationUsers; but the bug above invalidates this assumption. The result is that the user is stuck in a state that is simultaneously claimed and not claimed, and requires manual intervention by our CS team.

The solution here is to expressly exclude invited users from these queries. The queries are:
  - `Organization_ReadByClaimedUserEmailDomain` - generally used to block user-driven actions such as deleting their account or purging their vault
  - `OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2` - used to display the list of members in the Admin Console with the correct actions

Note that this does not help users stuck in this invalid state and then revoked. Because revoked is its own status, this requires a workaround to detect the underlying status - probably checking for revoked + userId not null + email not null, similar to the [OrganizationService](https://github.com/bitwarden/server/blob/99e132603926ccea532500553b62de09310bd4aa/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1182) method. However that is creating more workarounds to clean up later so I would rather leave this unhandled for now, noting that it does not pose any security risks. Instead we will work with Auth team to fix this edge case and stop these invalid states from arising ([Jira ticket](bitwarden.atlassian.net/browse/PM-22405)).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


util/Migrator/DbScripts/2026-01-14_01_ExcludeInvitedUsersFromClaimedDomains_V2.sql
```diff
@@ -0,0 +1,29 @@
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationUser_ReadByOrganizationIdWithClaimedDomains_V2]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    WITH OrgUsers AS (
+        SELECT *
```
JimmyVo16, line 8: Just an observation: It looks like this is existing code, so it’s not your change. That said, in general I think we should move away from `*` and only select what we actually need.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -325,7 +325,8 @@ join od in dbContext.OrganizationDomains on ou.OrganizationId equals od.Organiza
                     where ou.UserId == userWithDomain.UserId &&
                           od.DomainName == userWithDomain.EmailDomain &&
                           od.VerifiedDate != null &&
-                          o.Enabled == true
+                          o.Enabled == true &&
+                          ou.Status != OrganizationUserStatusType.Invited
```
JimmyVo16, line 329: Just thinking out loud here lol: 

I’ve noticed we have a few cases where we exclude certain collection types or user types. I agree that we should do filtering in the database to leverage the query optimizer and reduce sending unnecessary data over the network. However, I wonder if we could configure the sproc's parameters to include only the types we want returned. That way, we can control the types from the app in the future, reduce the need for an extra deployment, and avoid potential backward compatibility issues.

That said, this does add upfront complexity to the sproc for a situation that might never happen.

JimmyVo16 (dismissed):

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
eliykat: We do that in some sprocs, but here the status is closely tied to the purpose of the sproc: it's dealing with claimed accounts, and you can't claim an invited user's account. I would say this is a case where it should not be parameterized.

src/Sql/dbo/Stored Procedures/Organization_ReadByClaimedUserEmailDomain.sql
```diff
@@ -19,4 +19,5 @@ BEGIN
     WHERE OD.[VerifiedDate] IS NOT NULL
       AND CU.EmailDomain = OD.[DomainName]
       AND O.[Enabled] = 1
+      AND OU.[Status] != 0
```
mkincaid-bw: It would be handy to have the status documented in a comment like you did in the other proc
[https://contributing.bitwarden.com/contributing/code-style/sql/#comments-and-documentation](https://contributing.bitwarden.com/contributing/code-style/sql/#comments-and-documentation)

util/Migrator/DbScripts/2026-01-14_01_ExcludeInvitedUsersFromClaimedDomains_V2.sql
eliykat: I agree! For now this returns a whole OrgUser object so we need to select the whole row, but we should move towards more targeted data objects.

I did find one unnecessary `SELECT *` that I removed though.

JimmyVo16 (approved):

PR #6823, opened 2026-01-09
[SM-1805] Removing the new license item from the organizationLicense

cd-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/SM-1805

## 📔 Objective

Removing usedisablesmadsforusers from the files it should not be in. The new method of adding to the token requires that we do not add to these files as well.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/AdminConsole/Entities/Organization.cs
```diff
@@ -343,7 +343,6 @@ public void UpdateFromLicense(OrganizationLicense license, IFeatureService featu
         UseRiskInsights = license.UseRiskInsights;
         UseOrganizationDomains = license.UseOrganizationDomains;
         UseAdminSponsoredFamilies = license.UseAdminSponsoredFamilies;
-        UseDisableSmAdsForUsers = license.UseDisableSmAdsForUsers;
```
amorask-bitwarden, line 346: The new feature should remain on the `Organization` record so the self-hosted DB row that's created via the license respects the feature. It just doesn't need to be a direct property on the license itself.

src/Core/AdminConsole/Services/OrganizationFactory.cs
```diff
@@ -115,7 +115,6 @@ public static Organization Create(
             UseOrganizationDomains = license.UseOrganizationDomains,
             UseAdminSponsoredFamilies = license.UseAdminSponsoredFamilies,
             UseAutomaticUserConfirmation = license.UseAutomaticUserConfirmation,
-            UseDisableSmAdsForUsers = license.UseDisableSmAdsForUsers,
```
amorask-bitwarden, line 118: See comment on `Organization`.

src/Core/AdminConsole/Entities/Organization.cs
cd-bitwarden: I am confused about this suggestion because the way you are suggesting to do this, it appears in the downloaded bitwarden_organization_license.json file above the token always false, AND in the token with the correct value. 
It can't be on Organization.cs and not also be on organization license, because Organization.cs uses license.UseDisableSMAdsForUsers.

amorask-bitwarden: I don't think that's the case. When an organization owner downloads their license, the client will invoke [GET /api/organizations/{organizationId}/license](https://github.com/bitwarden/server/blob/main/src/Api/Billing/Controllers/OrganizationsController.cs#L91-L110).

That method, in turn, calls the [`CloudGetOrganizationLicenseQuery.GetLicenseAsync`](https://github.com/bitwarden/server/blob/main/src/Core/Billing/Organizations/Queries/GetCloudOrganizationLicenseQuery.cs#L44-L57) method, which ultimately returns an `OrganizationLicense`, not an `Organization`. 

While yes, that `OrganizationLicense` is initially constructed from the cloud `Organization` because of our old, deprecated approach to constructing the license, there's no reason `UseDisableSMAdsForUsers` needs to be copied from `Organization` to `OrganizationLicense` because it's in the `Token` property, which is constructed on line 55 of that method via the invocation of [`LicensingService.CreateOrganizationTokenAsync`](https://github.com/bitwarden/server/blob/main/src/Core/Billing/Services/Implementations/LicensingService.cs#L347-L359).

Then, when the user uploads their license on their self-hosted instance to create their `Organization` record, the row that gets stored in their self-hosted database will be constructed from the `Token` via the [`OrganizationFactory.Create`](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Services/OrganizationFactory.cs#L14-L68) method; not from the license's deprecated, root properties.

cd-bitwarden: So basically it seems like if I remove it from OrganizationLicense.cs I have to remove it from Organization.cs

PR #6831, opened 2026-01-13
[PM-28795] Refactor organization accept-init flow with atomic database operations and upfront validation

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28795

## 📔 Objective

Refactor the `AcceptInit` endpoint to consolidate the three separate operations (initialize org, accept user, confirm user) into a single command with upfront validation and a transactional database write, preventing the organization and user from entering an unrecoverable partial state if any step fails.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -143,7 +143,7 @@ public static class FeatureFlagKeys
     public const string RefactorMembersComponent = "pm-29503-refactor-members-inheritance";
     public const string BulkReinviteUI = "pm-28416-bulk-reinvite-ux-improvements";
     public const string UpdateJoinOrganizationEmailTemplate = "pm-28396-update-join-organization-email-template";
-    public const string RefactorOrgAcceptInit = "pm-28795-refactor-org-accept-init";
+    public const string RefactorOrgAcceptInit = "pm-33082-refactor-org-accept-init";
 
     /* Architecture */
     public const string DesktopMigrationMilestone1 = "desktop-ui-migration-milestone-1";


src/Core/AdminConsole/OrganizationFeatures/Organizations/Interfaces/IInitPendingOrganizationCommand.cs
```diff
@@ -9,5 +10,14 @@ public interface IInitPendingOrganizationCommand
...
+    [Obsolete("Use InitPendingOrganizationVNextAsync for consolidated flow with upfront validation. This method will be removed.")]
     Task InitPendingOrganizationAsync(User user, Guid organizationId, Guid organizationUserId, string publicKey, string privateKey, string collectionName, string emailToken);
+
+    /// <summary>
+    /// Initializes a pending organization and confirms the first owner with upfront validation.
+    /// </summary>
+    /// <returns>A CommandResult indicating success or specific validation errors.</returns>
+    Task<CommandResult> InitPendingOrganizationVNextAsync(
+        User user, Guid organizationId, Guid organizationUserId, string publicKey,
+        string privateKey, string collectionName, string emailToken, string userKey);
```
eliykat: This should use a request model so that these can be named properties rather than a list of parameters. Avoids mixing up different `string`s, for example.

src/Core/AdminConsole/Repositories/IOrganizationRepository.cs
```diff
@@ -62,4 +62,25 @@ public interface IOrganizationRepository : IRepository<Organization, Guid>
...
+    /// <param name="collectionName">Optional name for the default collection</param>
+    /// <returns>A task representing the asynchronous operation</returns>
+    Task InitializePendingOrganizationAsync(
+        Guid organizationId,
+        string publicKey,
+        string privateKey,
+        Guid organizationUserId,
+        Guid userId,
+        string userKey,
+        string collectionName);
```
eliykat: This should also take an object.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +255,137 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
             new { OrganizationId = organizationId, SeatsToAdd = increaseAmount, RequestDate = requestDate },
             commandType: CommandType.StoredProcedure);
     }
+
+    public async Task InitializePendingOrganizationAsync(
```
eliykat: This is a tricky one. A few thoughts...
- I think performing all write operations in one transaction makes sense - you want them to succeed or fail together
- however, with default isolation levels, I don't think there's any benefit to re-fetching the data within the transaction. And it moves a lot of the data arrangement down into the repository layer, giving it more responsibility beyond just interacting with the db.
- this touches on 4 different tables: Organization, OrganizationUser, User and Collections. But it's located in the Organization repository. We don't have any great way of handling this today, the best I've seen is how [RotateUserAccountKeysCommand](https://github.com/bitwarden/server/blob/main/src/Core/KeyManagement/UserKey/Implementations/RotateUserAccountkeysCommand.cs) builds up the `saveEncryptedDataActions` list of delegates, which are defined by different repositories, then executes them in the same transaction. It has some boilerplate, but it keeps each repository in charge of its own table, which I think is worth it.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -251,4 +255,137 @@ await connection.ExecuteAsync("[dbo].[Organization_IncrementSeatCount]",
...
+                user.RevisionDate = DateTime.UtcNow;
+
+                await connection.ExecuteAsync(
+                    "[dbo].[User_Update]",
+                    user,
+                    commandType: CommandType.StoredProcedure,
+                    transaction: transaction);
+            }
+
+            // 4. Create default collection if name provided
```
eliykat: Unfortunate that we confused the naming for this...

```suggestion
            // 4. Create default collection if name provided
            // (this is the initial collection for the organization, not a My Items collection)
```

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerAcceptInitTests.cs
eliykat, line 1: Most of these integration tests are for the current implementation, which will be removed soon, and only 2 are for the new implementation. It's OK to have some tests for the current code (e.g. to ensure parity) but we should have at least as much coverage for the new implementation.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -147,4 +175,151 @@ private bool ValidateInviteToken(OrganizationUser orgUser, User user, string ema
 
         return tokenValid;
     }
+
+    public async Task<CommandResult> InitPendingOrganizationVNextAsync(
```
eliykat: It may be worth clarifying/noting something like...

> The user initializing the organization is the first user to access it - there is no existing owner or provider who can change its settings. Therefore, the validation in this command assumes a default state - for example, it does not enforce policies for this organization, because none will be enabled.

This took me a little bit of staring at the screen to realise.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
eliykat, line 1: This command has a lot of validation, I think it would help to move it into a separate validator class. It makes for easier testing.

And especially if you move data arrangement back into the command, it separates the validation vs. the action of the command.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -147,4 +175,151 @@ private bool ValidateInviteToken(OrganizationUser orgUser, User user, string ema
...
+
+        if (org.PlanType == PlanType.Free &&
+            (orgUser.Type == OrganizationUserType.Owner || orgUser.Type == OrganizationUserType.Admin))
+        {
+            var adminCount = await _organizationUserRepository.GetCountByFreeOrganizationAdminUserAsync(user.Id);
+            if (adminCount > 0)
+            {
+                return new FreeOrgAdminLimitError();
+            }
+        }
```
eliykat: nit: this is in `ValidatePoliciesAsync` but this is not a policy.

src/Core/AdminConsole/OrganizationFeatures/Organizations/Interfaces/IInitPendingOrganizationCommand.cs
```diff
@@ -9,5 +10,14 @@ public interface IInitPendingOrganizationCommand
     /// <remarks>
     /// This method must target a disabled Organization that has null keys and status as 'Pending'.
     /// </remarks>
+    [Obsolete("Use InitPendingOrganizationVNextAsync for consolidated flow with upfront validation. This method will be removed.")]
     Task InitPendingOrganizationAsync(User user, Guid organizationId, Guid organizationUserId, string publicKey, string privateKey, string collectionName, string emailToken);
+
+    /// <summary>
+    /// Initializes a pending organization and confirms the first owner with upfront validation.
```
eliykat: The business context for this is useful, including a link to the command earlier in this overall flow:

```suggestion
    /// Initializes a pending organization created via the Bitwarden Portal on behalf of a Reseller.
    /// See <see cref="ResellerClientOrganizationSignUpCommand"/>.
    /// It also confirms the first owner.
```

src/Core/AdminConsole/Repositories/IOrganizationRepository.cs
r-tome: I considered this but I couldn't find that pattern elsewhere in our repositories, so keeping separate parameters seems more consistent (though I agree there are too many). Open to being convinced if you feel strongly about it though!

eliykat: Maybe the issue is that there are just too many parameters here. Repository calls usually take a database object (entity) or just a couple of params. This may parameters indicates that it's too complex. Once you split this up per my comment below, it'll probably be OK because it'll be comprised of 4 separate repository methods that will only have a few parameters each.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
r-tome: Removed this as part of a refactor

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerAcceptInitTests.cs
r-tome: All tests now run for both implementations

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -319,10 +313,19 @@ public async Task<IResult> AcceptInit(Guid orgId, Guid organizationUserId, [From
...
-                model.Keys.PublicKey, model.Keys.EncryptedPrivateKey,
-                model.CollectionName, model.Token, model.Key);
+            var request = new InitPendingOrganizationRequest
+            {
+                User = user,
+                OrganizationId = orgId,
+                OrganizationUserId = organizationUserId,
+                PublicKey = model.Keys.PublicKey,
+                PrivateKey = model.Keys.EncryptedPrivateKey,
+                CollectionName = model.CollectionName,
```
eliykat, line 323: There is a reusable class for representing org keys in Core: `PublicKeyEncryptionKeyPairData`. The request model has a method that maps to it, e.g. `model.Keys.ToPublicEncryptionKeyPairData()`.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -319,10 +313,19 @@ public async Task<IResult> AcceptInit(Guid orgId, Guid organizationUserId, [From
...
+                EmailToken = model.Token,
+                UserKey = model.Key
```
eliykat: As you noted in the xmldoc, this is the UserKey-encrypted organization key. UserKey would refer to the user's account encryption key, which is not the case here. A more accurate name would be `EncryptedOrganizationSymmetricKey`. But it might also be a good idea to check with Bernd whether there is a specific model or naming convention we can use.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationValidator.cs
eliykat, line 1: We don't have any documented rules for this yet - I should do that - but where validator classes have worked well, we have a single public interface: `ValidateAsync`. It also takes a request object, which is usually similar to the command request object, but enriched with additional data so it's not being fetched twice (by both the command and the validator). It returns a `ValidationResult`.

The benefit of this is that the command doesn't know anything about what needs to be validated - it just calls `ValidateAsync`.

For an example, see [DeleteClaimedOrganizationUserAccountValidator](https://github.com/bitwarden/server/blob/1eb396cb401f2c30732671d2de792f0fd9e6a776/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/DeleteClaimedAccount/DeleteClaimedOrganizationUserAccountValidator.cs).

It also makes testing the command super easy, because you can just mock a valid or invalid result, without having to mock every single validation check. Each class becomes single responsibility: command handles fetching all the data and performing the action, validator handles the validation.

Jared M has done a bit of work with these so he's also a good resource if he lines up better with your timezone.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -168,146 +171,127 @@ private async Task ValidateSignUpPoliciesAsync(Guid ownerId)
...
-        {
-            return new EmailMismatchError();
-        }
-
-        return null;
+        org.Enabled = true;
+        org.Status = OrganizationStatusType.Created;
+        org.PublicKey = request.PublicKey;
+        org.PrivateKey = request.PrivateKey;
+        org.RevisionDate = DateTime.UtcNow;
```
eliykat: Should use `TimeProvider`.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationValidator.cs
```diff
@@ -0,0 +1,164 @@
...
+        if (twoFactorReq.IsTwoFactorRequiredForOrganization(organizationId) &&
+            !await _twoFactorIsEnabledQuery.TwoFactorIsEnabledAsync(user))
+        {
+            return new TwoFactorRequiredError();
+        }
+
+        return null;
+    }
+
+    public async Task<Error?> ValidateBusinessRulesAsync(User user, Organization org, OrganizationUser orgUser)
```
eliykat: "Business rules" is a broad term. I suggest `ValidateFreeOrganizationLimit`.

That said, I don't think you can create a free org via this flow. But it also doesn't hurt to leave it as protection, as there's nothing inherent in this command that would prevent it from being used for a free org.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationRequest.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public required string PublicKey { get; init; }
+
+    /// <summary>
+    /// The organization's encrypted private key.
+    /// </summary>
+    public required string PrivateKey { get; init; }
+
+    /// <summary>
+    /// The name of the default collection to create. Optional - if null or empty, no collection is created.
```
eliykat, line 33: Worth noting something like:

> NOTE: this is not the default user collection. It is the first collection created when the organization is created, which is a standard shared collection.

eliykat: Would this ever be null? I think clients should always send it (but please check me).

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -140,11 +171,139 @@ private async Task ValidateSignUpPoliciesAsync(Guid ownerId)
...
+            _organizationRepository.BuildUpdateOrganizationAction(org),
+            _organizationUserRepository.BuildConfirmOrganizationUserAction(orgUser),
+            _userRepository.BuildVerifyUserEmailAction(request.User.Id),
+        ];
+
+        if (!string.IsNullOrWhiteSpace(request.CollectionName))
+        {
+            var defaultCollectionAction = CreateDefaultCollectionAction(request);
+            updateActions.Add(defaultCollectionAction);
+        }
```
eliykat, line 245: This is what I requested and I think it's the right pattern. But it does have some overhead in the boilerplate required, especially on seemingly unrelated repositories (like `UserRepository`).

I'm thinking again about what we're trying to achieve here. This needs to be an atomic update - but maybe not all of it needs to be atomic.

- the organization and orgUser updates need to happen together or not at all. We can't be left with an initialized org but no confirmed orgUser.
- verifying the user's email is a "nice to have". If it failed, the user can still verify their email, it's just annoying. No data is lost or invalid.
- same for the default (org) collection. If that failed for some reason, the user would just have to manually create their first collection.

So I suggest:
- `var initializeOrgUser = organizationUserRepository.InitializeForPendingOrganization` - this is your new organizationUserRepository method but with a more explicit name
- `await organizationRepository.InitializeOrganization(org, initializeOrgUser)` - combines your 2 new organizationRepository methods into a single call, because they'll always go together anyway
- call the existing userRepository method outside of the transaction
- call the existing collectionRepository outside of the transaction

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -140,11 +171,139 @@ private async Task ValidateSignUpPoliciesAsync(Guid ownerId)
...
+        var collectionUsers = new[]
+        {
+            new CollectionAccessSelection
+            {
+                Id = request.OrganizationUserId,
+                HidePasswords = false,
+                ReadOnly = false,
+                Manage = true
+            }
+        };
```
eliykat: (non-blocking) I think we do this in a few places, is it worth moving to a shared static method?

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
r-tome: I forgot about that! Nice catch

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
r-tome: From what I've seen we usually name it `Key` but that is too generic so I took your suggested name

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationValidator.cs
r-tome: That makes sense! I have refactored the whole file

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationRequest.cs
r-tome: I'm not sure why it was made nullable in the original implementation. We can address that later and also adding `[Required]` in the API model

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
r-tome: It is but I'll leave that for a separate PR.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -139,4 +167,114 @@ private bool ValidateInviteToken(OrganizationUser orgUser, User user, string ema
...
+        if (org is null)
+        {
+            return new OrganizationNotFoundError();
+        }
+
+        var validationRequest = new InitPendingOrganizationValidationRequest
+        {
+            User = request.User,
+            OrganizationId = request.OrganizationId,
+            OrganizationUserId = request.OrganizationUserId,
```
eliykat, line 193: You can remove `OrganizationId` and `OrganizationUserId` because you're already including those whole objects.

EDIT: oh I see you're extending the command request class - that's a reasonable way to minimize boilerplate, so that's OK.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
eliykat: Yes we used to do that a lot, but we're moving towards more explicit naming for our cryptographic objects. Thanks for fixing it!

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
eliykat, line 1: This command is looking really tidy now!

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationValidator.cs
eliykat, line 1: This is also really straightforward to read, great job.

src/Core/AdminConsole/Repositories/IOrganizationRepository.cs
```diff
@@ -2,11 +2,21 @@
...
 namespace Bit.Core.Repositories;
 
+/// <summary>
+/// Represents a database action that can be executed within a shared transaction
+/// during organization initialization. Used to atomically update the organization
+/// and confirm the first org user together.
+/// </summary>
+public delegate Task OrganizationInitializationAction(SqlConnection? connection = null,
+    SqlTransaction? transaction = null,
+    object? context = null);
```
eliykat: Now that this is only used in a couple of locations, I suggest using the more modern approach of a `Func<SqlConnection, SqlTransaction, Task>`. 

You can also be more specific about the required parameters - it looks like the connection and transaction can be non-nullable, and you don't need the context object.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -464,4 +464,38 @@ await dbContext.Organizations
...
+            {
+                efOrganization.Enabled = organization.Enabled;
+                efOrganization.Status = organization.Status;
+                efOrganization.PublicKey = organization.PublicKey;
+                efOrganization.PrivateKey = organization.PrivateKey;
+                efOrganization.RevisionDate = organization.RevisionDate;
+
+                await dbContext.SaveChangesAsync();
+            }
+
```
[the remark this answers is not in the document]
eliykat: I think Claude is right here. ([context](https://github.com/bitwarden/server/pull/6831/changes#r2841234551))

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -978,6 +979,25 @@ public async Task<bool> ConfirmOrganizationUserAsync(AcceptedOrganizationUserToC
...
     }
+
+    public OrganizationInitializationAction BuildConfirmOwnerAction(Core.Entities.OrganizationUser organizationUser)
+    {
+        return async (SqlConnection? _, SqlTransaction? _, object? context) =>
+        {
+            var dbContext = (DatabaseContext)context!;
+
+            var efOrganizationUser = await dbContext.OrganizationUsers.FindAsync(organizationUser.Id);
+            if (efOrganizationUser != null)
```
eliykat: Same here - if there's no orgUser, something's gone wrong.

src/Infrastructure.EntityFramework/Repositories/CollectionRepository.cs
eliykat, line 1: Mysterious whitespace 😁

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
eliykat, line 1: Whitespace here as well

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
```diff
@@ -139,4 +167,114 @@ private bool ValidateInviteToken(OrganizationUser orgUser, User user, string ema
...
+        await _collectionRepository.CreateAsync(
+            obj: defaultCollection,
+            groups: null,
+            users: defaultOwnerAccess);
+    }
+
+    private async Task SendNotificationsAsync(Organization org, OrganizationUser orgUser, User user)
+    {
+        await _eventService.LogOrganizationUserEventAsync(orgUser, EventType.OrganizationUser_Confirmed);
+        await _mailService.SendOrganizationConfirmedEmailAsync(org.DisplayName(), user.Email, orgUser.AccessSecretsManager);
```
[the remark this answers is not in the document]
eliykat: @r-tome should we just assume that the feature flags will be turned on in order? i.e. by the time you're turning on yours, the email changes will be released? ([context](https://github.com/bitwarden/server/pull/6831#discussion_r2852506009)) It might be a good idea to check the progress of those changes. It would be best to avoid nesting feature flags.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerAcceptInitTests.cs
```diff
@@ -0,0 +1,561 @@
...
+        Assert.Equal(collections.First().Id, collectionAccess.First().Id);
+        Assert.True(collectionAccess.First().Manage);
+        Assert.False(collectionAccess.First().ReadOnly);
+        Assert.False(collectionAccess.First().HidePasswords);
+    }
+
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task AcceptInit_WithoutCollectionName_InitializesOrganizationWithoutCreatingCollection(bool featureFlagEnabled)
```
eliykat: This is not an intended flow and I'm not sure it ever happens at the moment - I don't think we need an integration test for it.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerAcceptInitTests.cs
eliykat, line 1: Some of the validation-related tests are getting very granular for integration tests (because they are slow to run, we don't wan to test every path). I suggest just relying on unit tests for tests like validating the organization state.

test/Core.Test/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommandTests.cs
```diff
@@ -166,4 +156,159 @@ public string CreateToken(OrganizationUser orgUser, Guid orgUserId, SutProvider<
...
+        var requestWithoutCollection = request with { CollectionName = null };
+        SetupSuccessfulValidation(org, orgUser, requestWithoutCollection, sutProvider);
+
+        var result = await sutProvider.Sut.InitPendingOrganizationVNextAsync(requestWithoutCollection);
+
+        Assert.False(result.IsError);
+
+        var organizationRepository = sutProvider.GetDependency<IOrganizationRepository>();
+        await organizationRepository.Received(1).InitializeOrganizationAsync(
+            org, Arg.Any<OrganizationInitializationAction>());
```
eliykat: Please verify the state of the organization - at the moment this test does not check that the organization was correctly updated, only that the same object-by-reference was passed in.

Same in the next test.

eliykat (changes requested):
Great work! Feedback is minor only.

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
r-tome: > that's a reasonable way to minimize boilerplate, so that's OK

Exactly! 😄

src/Core/AdminConsole/Repositories/IOrganizationRepository.cs
r-tome: Great suggestion, it looks cleaner!

src/Core/AdminConsole/OrganizationFeatures/Organizations/InitPendingOrganizationCommand.cs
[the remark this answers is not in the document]
r-tome: I ended up nesting the feature flags. I know its not ideal but this way we don't have to deal with worrying about the order of the flags.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerAcceptInitTests.cs
r-tome: You're right! I just left the success test and authorization test.

src/Core/Constants.cs
```diff
@@ -143,6 +143,7 @@ public static class FeatureFlagKeys
     public const string RefactorMembersComponent = "pm-29503-refactor-members-inheritance";
     public const string BulkReinviteUI = "pm-28416-bulk-reinvite-ux-improvements";
     public const string UpdateJoinOrganizationEmailTemplate = "pm-28396-update-join-organization-email-template";
+    public const string RefactorOrgAcceptInit = "pm-28795-refactor-org-accept-init";
```
vincentsalucci: Please update this key to `pm-33082-refactor-org-accept-init`

PR #6516, opened 2025-10-29
[PM-27815] Remove deprecated logging methods

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27815

## 📔 Objective

Removes deprecated logging methods in accordance with https://contributing.bitwarden.com/architecture/adr/standard-output-logging and these methods having been deprecated for 3 months https://bitwarden.com/help/releasenotes/#2025-7-3

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #6860, opened 2026-01-16
[PM-30858] Fix excessive logs

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30858

## 📔 Objective

In #6516 we removed legacy logging methods and did some cleanup to no longer rely on the `inclusionPredicate` with the idea that we can customize the log levels per category through configuration. The one problem with that is that those customization were only respected if you started using the new location to define the log path `Logging:PathFormat`. If you were still using `globalSettings__logDirectory` which all self-hosters are still using then those levels won't be used. That is a problem but not actually the main issue found, the main issue is that the `inclusionPredicate` had its own level customization mechanism which for the most part only logged warnings or errors. I had seen [this line of code](https://github.com/bitwarden/server/commit/fdfec0ac4df89b33896aadf532e95404f01747c0#diff-bb8f810ec8e5de0457b5bbc1027b2c5f8b4fe3861c25228b827e1e68c1a1f733L65) (`MinimumLevel.Verbose()`) so I thought I would be changing the default from `Verbose` to `Information` but the inclusion predicate is what filtered it further. This PR makes 2 changes, it makes it so that log levels can be customized alongside `globalSettings__logDirectory` and it now changes our default production level customization to log a lot less. 

This should not affect cloud production since we further customize log levels for the `Console` provider which is what we use in cloud. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/test/Core.Test/Utilities/LoggerFactoryExtensionsTests.cs
+++ b/test/Core.Test/Utilities/LoggerFactoryExtensionsTests.cs
@@ -91,30 +91,12 @@ public class LoggerFactoryExtensionsTests
     }
 
     [Fact]
-    public async Task AddSerilogFileLogging_LegacyConfig_InfoLogs_DoNotFillUpFile()
-    {
-        await AssertSmallFileAsync((tempDir, config) =>
-        {
-            config["GlobalSettings:LogDirectory"] = tempDir;
-        });
-    }
-
-    [Fact]
     public async Task AddSerilogFileLogging_LegacyConfig_WithLevelCustomization_InfoLogs_DoNotFillUpFile()
     {
         await AssertSmallFileAsync((tempDir, config) =>
         {
             config["GlobalSettings:LogDirectory"] = tempDir;
             config["Logging:LogLevel:Microsoft.AspNetCore"] = "Warning";
-        });
-    }
-
-    [Fact]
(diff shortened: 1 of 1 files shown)


prograhamming (approved):

bitwarden_license/src/Scim/appsettings.Production.json
```diff
@@ -23,11 +23,9 @@
     }
   },
   "Logging": {
-    "IncludeScopes": false,
```
justindbaur, line 26: `IncludeScopes` is not valid right underneath `Logging`, it is never read and we already have the setting in `Logging:Console`.

bitwarden_license/src/Scim/appsettings.Production.json
```diff
@@ -23,11 +23,9 @@
...
     "LogLevel": {
-      "Default": "Debug",
-      "System": "Information",
-      "Microsoft": "Information"
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
```
justindbaur, line 28: These new settings are what .NET will generate by default from project templates. https://github.com/dotnet/aspnetcore/blob/552f3f47f8c7e463d7d6b073cfa1be24023cecd4/src/ProjectTemplates/Web.ProjectTemplates/content/WebApi-CSharp/appsettings.json#L60-L64

PR #6871, opened 2026-01-19
[PM-29955] User reset password key can be empty string during key rotation

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29955

## 📔 Objective

User reset password key can be empty string. In such state, a key rotation fails with error `All existing reset password keys must be included in the rotation.`, due to a discrepancy between how request is build in client's and how it's validated in server.

If an organization is enrolled into reset password (but not the automatic enrolment for new members), then user enrols and un-enrols from the reset password account recovery, the reset password key is set to `""` empty string. https://github.com/bitwarden/clients/blob/44bdaf71b33c54172f5f2d9dca1d265c6739cbc9/apps/web/src/app/vault/individual-vault/vault-filter/components/organization-options.component.ts#L250
Note, this behaviour changed with https://github.com/bitwarden/clients/pull/13889 and have been around for almost 1 year. I suspect this was accidental due to `@ts-strict-ignore` removal in the file.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Api/KeyManagement/Validators/OrganizationUserRotationValidator.cs
```diff
@@ -34,8 +34,7 @@ public async Task<IReadOnlyList<OrganizationUser>> ValidateAsync(User user,
         }
 
         // Exclude any account recovery that do not have a key.
-        existing = existing.Where(o => o.ResetPasswordKey != null).ToList();
```
quexten, line 37: This will fix the immediate error, but I'm not sure if we actually want to make it part of the API contract in the medium / long term. We should ideally fix it and add a server-side migration to clear the values.

At any rate, we should probably document this better.

mzieniukbw: Agreed. Raised https://bitwarden.atlassian.net/browse/PM-31001

PR #6876, opened 2026-01-21
[PM-24759] Add Method for Retrieving Policy Requirements for Multiple Users

sven-bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-24759

## 📔 Objective
Following existing patterns, we want to expose a `GetAsync` method in the query class that supports single and collection parameters. This PR opens PolicyRequirementQuery's ability to load requirements given a list of user IDs. 

We're also now delegating requests for single users to this collection-based request, unifying logic across both methods.

Tested by validating the MasterPasswordPolicyRequirement was enforced against a user upon login.
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyRequirementQuery.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyRequirementQuery.cs
@@ -21,11 +21,11 @@ public interface IPolicyRequirementQuery
     /// The policy requirement represents how one or more policy types should be enforced against the users.
     /// </summary>
     /// <returns>
-    /// A list of applicable policy requirements in corresponding order of the submitted user IDs.
+    /// A collection of tuples pairing each user ID with their corresponding policy requirement.
     /// </returns>
     /// <param name="userIds">The users that you need to enforce the policy against.</param>
     /// <typeparam name="T">The IPolicyRequirement that corresponds to the policy you want to enforce.</typeparam>
-    Task<IEnumerable<T>> GetAsync<T>(IEnumerable<Guid> userIds) where T : IPolicyRequirement;
+    Task<IEnumerable<(Guid UserId, T Requirement)>> GetAsync<T>(IEnumerable<Guid> userIds) where T : IPolicyRequirement;
 
     /// <summary>
     /// Get all organization user IDs within an organization that are affected by a given policy type.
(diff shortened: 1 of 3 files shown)


test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyRequirementQueryTests.cs
```diff
@@ -11,25 +11,6 @@ namespace Bit.Core.Test.AdminConsole.OrganizationFeatures.Policies;
 [SutProviderCustomize]
 public class PolicyRequirementQueryTests
 {
-    [Theory, BitAutoData]
-    public async Task GetAsync_IgnoresOtherPolicyTypes(Guid userId)
```
sven-bitwarden, line 15: The underlying sproc now does the filtering, so we no longer need to test what would happen if it didn't.

src/Core/AdminConsole/OrganizationFeatures/Policies/IPolicyRequirementQuery.cs
```diff
@@ -16,6 +16,17 @@ public interface IPolicyRequirementQuery
...
+    /// <summary>
+    /// Get a policy requirement for a list of users.
+    /// The policy requirement represents how one or more policy types should be enforced against the users.
+    /// <returns>
+    /// A list of applicable policy requirements in corresponding order of the submitted user IDs.
+    /// </returns>
+    /// </summary>
+    /// <param name="userIds">The users that you need to enforce the policy against.</param>
+    /// <typeparam name="T">The IPolicyRequirement that corresponds to the policy you want to enforce.</typeparam>
+    Task<IEnumerable<T>> GetAsync<T>(IEnumerable<Guid> userIds) where T : IPolicyRequirement;
```
sven-bitwarden: Is it awkward to return just Requirement[] here? The expected use case might be something like -

```cs
var userPolicyRequirements = (await _requirementQuery.GetAsync(userIds)).Zip(userIds);
foreach (var (requirement, userId) in userPolicyRequirements) ...
```

jrmccannon: Do you mean changing the return type from `IEnumerable<T>` to `T[]`? 

That would probably be fine. I doubt we'd have any need to add or remove anything to the results here in the future. This would allow for your use case that requires retaining order. :+1: 

Probably a good idea.

sven-bitwarden: This method takes in a Guid[] UserIds, but the return type doesn't give back the UserId that the PolicyRequirement applies to, so the only way to match it up to the callers is by subsequently iterating both with something like `.Zip()`. That might be fine, but was calling it out

jrmccannon: Ah. Now I understand. Apologies for that.

We definitely should provide a way for them to identify which `IPolicyRequirement` matches up to which `OrganizationUserId`. 

I would just include it in the new return value. Either a new type altogether or just a`IEnumerable<(Guid, T)>`.

Issue #6893, opened 2026-01-24
WebAuthn devices beyond previous max do not work as 2FA


jolness1:
### Steps To Reproduce

1. Register 6 or more WebAuthn devices at [vault.bitwarden.com](https://vault.bitwarden.com/#/settings/security/two-factor)
2. Log out
3. Attempt to login
4. When prompted, attempt to use 6th or later device registered as 2FA security key

### Expected Result

Website accepts the key and user successfully logs in.

### Actual Result

Get an error "Try a different security key. You're using a security key that's not registered with this website".

### Additional Context

All previously registered security keys (up to the previous limit of 5) work as expected still.
Behavior is identical across Brave and firefox on macOS, Debian 13 and Windows 11. 
Keys register and save successfully but all keys past the former 5 key limit do not work as a second factor 
Not sure about passkey login, my hunch is the behavior is the same. 
Happy to test that hypothesis and work this issue if needed. Would love a chance to work on some FIDO2/WebAuthn stuff/contribute to one of my favorite OSS projects.

### Build Version

latest

### Environment

Cloud (bitwarden.com)

### Environment Details

N/A

EDIT: Clarity/Accuracy
### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

pamperer562580892423:
Is your BW cloud account a free account?

SergeantConfused:
Hello @jolness1,

Thank you for your report. I have flagged this report to the Engineering department; please feel free to post additional information, such as screenshots or a screen video recordings, if you wish.

By the way, does that individual Bitwarden account have an individual Premium subscription?

Thank you again,

jolness1:
@SergeantConfused I have it fixed. Looks like someone enabled checking premium accounts for adding more WebAuthn keys but not for login. Super minor change but wanted to run tests/do manual QA. 

PR is here:  [https://github.com/bitwarden/server/pull/6894](https://github.com/bitwarden/server/pull/6894)

Thanks for responding late on a Friday night by the way! 

——— Videos  demonstrating behavior ———

Possible to add more than 5 keys on premium account(pre and post-fix):

https://github.com/user-attachments/assets/adc406a8-266c-44c0-b538-698774fa77b2

6th+ key does not work as 2nd factor: 

[...]

PR #6894, opened 2026-01-26
[PM-31212] Change hardcoded 5 key WebAuthn limit for login to check if premium

jolness1:
## 🎟️ Tracking

[https://github.com/bitwarden/server/issues/6893](https://github.com/bitwarden/server/issues/6893)

[https://bitwarden.atlassian.net/browse/PM-31212](https://bitwarden.atlassian.net/browse/PM-31212)

## 📔 Objective

Post-change to the new 10 WebAuthn key limit for premium users, it was possible to _register_ more than 5 keys but any keys registered past that point did not work for login because the old hardcoded limit of 5 was left in place rather than dynamically checking if user had a premium subscription.  

**_Note:_** This does not force the login to not load all 10 keys on free accounts since it is already enforced during registration.
 
Reason: If a user loses their first 5 WebAuthn keys and also has had their premium subscription lapse, I would think we wouldn't want to lock them out of their account entirely. I do not have strong feelings on this but enforcing on registration seems like enough. Definitely an edge case, though I wanted to explain my logic as it was a deliberate choice. 

I can see why it might be better to prevent login via keys beyond the free limit — users should have their recovery code, not doing so would mean that some small subset of users who pay for a year and let it lapse can continue to use more then 5 security keys etc. 

Happy to make a change for a stricter check to defend against what seems like a super remote edge case if reviewers feel that's needed! 

## 📸 Screenshots

Possible to add more than 5 keys on premium account(behavior consistent before and after fix):

https://github.com/user-attachments/assets/adc406a8-266c-44c0-b538-698774fa77b2

6th+ key does not work as 2nd factor (pre-fix, undesirable behavior): 

https://github.com/user-attachments/assets/ebd0fea4-0f34-407e-9fe3-42b68176f29a

Key under the old 5 key limit still works (behavior consistent before and after fix):

https://github.com/user-attachments/assets/7c1ecadf-ed48-408b-a690-cf1a1937423b

Behavior after fix is what would be expected:

https://github.com/user-attachments/assets/3d4141c2-cbcc-4f86-a187-1c45fa92dd7d

Free account still limited to 5 key maximum (sanity check — no changes to registration flow):

https://github.com/user-attachments/assets/6c123fcc-c87a-4e03-b1c4-242cd26667d1

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


src/Core/Auth/Identity/TokenProviders/WebAuthnTokenProvider.cs
```diff
@@ -147,8 +147,11 @@ private bool HasProperMetaData(TwoFactorProvider provider)
             return keys;
         }
 
-        // Support up to 5 keys
-        for (var i = 1; i <= 5; i++)
+        // Load all WebAuthn credentials stored in metadata. The number of allowed credentials
+        // is controlled by PremiumMaximumAllowedCredentials and NonPremiumMaximumAllowedCredentials
+        // in GlobalSettings.WebAuthn (10 for premium, 5 for non-premium).
+        var maximumCredentials = _globalSettings.WebAuthn.PremiumMaximumAllowedCredentials;
+        for (var i = 1; i <= maximumCredentials; i++)
```
enmande:
```suggestion
        for (var i = 1; i <=  provider.MetaData.Count(k => k.Key.StartsWith("Key")); i++)
```
This would more closely parallel the enforcement design around [WebAuthn registration](https://github.com/bitwarden/server/blob/afb087161a99549e8dea09afcd75997e59a96d7a/src/Core/Services/Implementations/UserService.cs#L380), and should always ensure one execution per key which exists on the provider metadata (one key, one evaluation).

jolness1:
> @jolness1 Thank you so much for you contribution! ✨ Very excited to see this coming in and for the engagement.
> 
> 🎨 If I may suggest a small change, the spirit of which is below. We'd just like from a design perspective to keep the related settings confined to registration time for these credentials; doing something like the below would let the keys "speak for themselves" here and not worry about the setting.

Looking forward to contributing ~~more~~ further. I ❤️ Bitwarden. 
Accepted. That change makes sense to me. Not super familiar with the codebase so I will happily defer to you all on that!

src/Core/Auth/Identity/TokenProviders/WebAuthnTokenProvider.cs
```diff
@@ -147,8 +147,11 @@ private bool HasProperMetaData(TwoFactorProvider provider)
             return keys;
         }
 
-        // Support up to 5 keys
-        for (var i = 1; i <= 5; i++)
+        // Load all WebAuthn credentials stored in metadata. The number of allowed credentials
+        // is controlled by PremiumMaximumAllowedCredentials and NonPremiumMaximumAllowedCredentials
+        // in GlobalSettings.WebAuthn (10 for premium, 5 for non-premium).
+        var maximumCredentials = _globalSettings.WebAuthn.PremiumMaximumAllowedCredentials;
```
enmande:
```suggestion
```
Check for `*MaximumAllowedCredentials` is no longer needed.

src/Core/Auth/Identity/TokenProviders/WebAuthnTokenProvider.cs
```diff
@@ -147,8 +147,11 @@ private bool HasProperMetaData(TwoFactorProvider provider)
             return keys;
         }
 
-        // Support up to 5 keys
-        for (var i = 1; i <= 5; i++)
+        // Load all WebAuthn credentials stored in metadata. The number of allowed credentials
+        // is controlled by PremiumMaximumAllowedCredentials and NonPremiumMaximumAllowedCredentials
+        // in GlobalSettings.WebAuthn (10 for premium, 5 for non-premium).
```
enmande:
```suggestion
        // Load all WebAuthn credentials stored in metadata. The number of allowed credentials
        // is controlled by credential registration.
```

src/Core/Auth/Identity/TokenProviders/WebAuthnTokenProvider.cs
enmande:
```suggestion
        foreach (var kvp in provider.MetaData.Where(k => k.Key.StartsWith("Key")))
        {
            var key = new TwoFactorProvider.WebAuthnData((dynamic)kvp.Value);
            keys.Add(new Tuple<string, TwoFactorProvider.WebAuthnData>(kvp.Key, key));
        }
```
`clients` should be gap-filling the indexes (e.g., `{n}` of `Key{n}`), but that shouldn't be assumed, and isn't controlled by this separate context. Ultimately, if it is `Key*`-named, it is a registered credential for the user. This will be more supportable and robust long-term.

We see this same patterning in [`UserService.StartWebAuthnRegistrationAsync`](https://github.com/bitwarden/server/blob/afb087161a99549e8dea09afcd75997e59a96d7a/src/Core/Services/Implementations/UserService.cs#L393).

jolness1:
> @jolness1 Thanks for your engagement, and sorry for the second round; I should have been more specific in my initial review with the suggested changes. Here's what I think we should have for its "final form" 🙇‍♂️
> 
> I've verified this locally and am including a brief screen recording for your convenience.
> 
>  pm-31212__refactor.mov 
> 📓 _disclaimer_ I'm not thrilled with the line numbering looking at the patches here for these requested changes, especially the `foreach`, and to be clear I'm not sure if that's a GitHub thing or a me thing in this instance. I _think_ they should come in fine, but if they don't one of us may need to tinker locally and push again. I'm trying to do this in the web UI so this PR remains yours and I'm just pulling some threads together from another spot in the codebase 😉 .

No, I should have been more thorough, I was a little distracted and should have come back to it when I had a bit more time to think through your edits. Not like this is some critical patch that needs to make it to prod to restore critical functionality but still managed to have a small version of the same panic as when I was at my first SWE job and blew up the kafka service on a Friday afternoon.  😬

Those changes make sense to me, I will implement them and then make sure we are good on linting/testing etc and then we can take another pass at it. 

Thanks for dealing with all this from the GH WebUI. From experience, it is not pleasant to work from.

jolness1:
@enmande The for loop did get stuck in there (GH webui is a nightmare to do editing in 🫠) but I removed that, code is linted and tests pass. 

Should be ready for another look when you have a moment!

jolness1:
@enmande It was my pleasure. BW is one of the most useful pieces of software in my life so to contribute even in a small way is great. Plus, I am very excited to be able to have all my Yubikeys registered, had catastrophic (at the time it didn't seem like it buy BTC was worthless. Was way more angry about my expensive SSD) data loss about 15 years ago and ever since I tend to go for "maximum reasonable backup plan".

Thanks for all your help! I think the changes will definitely help make maintaining this service easier in the future. Tech debt accumulates despite our best efforts so when there is a chance to avoid it, always best to do so.  

I'll keep an eye out for other issues — hopefully there is something else I can pick up soon. 🙌🏼  Need to do a bit more looking to see how the ticketing process works and how things are assigned, every project is a bit different. 

Thanks again!

enmande:
@jolness1 Thanks again for your submission~✨
It has been tested and merged to `main`. It will be included in an upcoming release.
Keep being awesome, and hope to see you again soon.

jolness1:
> @jolness1 Thanks again for your submission~✨ It has been tested and merged to `main`. It will be included in an upcoming release. Keep being awesome, and hope to see you again soon.

@enmande Thanks for all your help with this! Glad to be able to contribute to a great piece of software in a tiny way. 
I will keep my eyes out for other tickets to pick up and any bugs or features I think deserve attention. 

Seriously, thanks again!

PR #6896, opened 2026-01-26
Add cipher seeding with Rust SDK encryption to enable cryptographically correct test data generation

theMickster:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31013

## 📔 Objective

Continue to build on the initial success of the Seeder with incremental progress towards a solid tool.

1. Replaces hardcoded string arrays with a structured data system — enabling filtered and targeted test data generation
2. Adds 50 real companies across 3 geographic regions with domain, category, and type metadata 
3. Introduces deterministic username generation to login ciphers
4. Implements 3 organizational structure models (Traditional, Spotify, Modern) for _realistic_ collection hierarchies 
5. Enhanced the CLI to allow for different organizational types
6. Encrypts all collection names via the Rust SDK 
7. Establishes extensibility patterns for future cipher types 

Please take a moment to read the `README.md` files, the `CLAUDE.md`, and a few of the TODOs in XML comments. 
Those will help guide you towards where we will be going with the work.

🤖 Built with Claude Code

### DIFF
--- a/util/Seeder/Factories/OrganizationSeeder.cs
+++ b/util/Seeder/Factories/OrganizationSeeder.cs
@@ -7,12 +7,10 @@ namespace Bit.Seeder.Factories;
 
 public class OrganizationSeeder
 {
-    /// <summary>
-    /// Creates an enterprise organization without encryption keys.
-    /// Keys should be generated dynamically using RustSdkService.GenerateOrganizationKeys()
-    /// and assigned to PublicKey/PrivateKey after creation.
-    /// </summary>
-    public static Organization CreateEnterprise(string name, string domain, int seats)
+    private static readonly string _defaultPublicKey = "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAmIJbGMk6eZqVE7UxhZ46Weu2jKciqOiOkSVYtGvs61rfe9AXxtLaaZEKN4d4DmkZcF6dna2eXNxZmb7U4pwlttye8ksqISe6IUAZQox7auBpjopdCEPhKRg3BD/u8ks9UxSxgWe+fpebjt6gd5hsl1/5HOObn7SeU6EEU04cp3/eH7a4OTdXxB8oN62HGV9kM/ubM1goILgjoSJDbihMK0eb7b8hPHwcA/YOgKKiu/N3FighccdSMD5Pk+HfjacsFNZQa2EsqW09IvvSZ+iL6HQeZ1vwc/6TO1J7EOfJZFQcjoEL9LVI693efYoMZSmrPEWziZ4PvwpOOGo6OObyMQIDAQAB";
+    private static readonly string _defaultPrivateKey = "2.6FggyKVyaKQsfohi5yqgbg==|UU2JeafOB41L5UscGmf4kq15JGDf3Bkf67KECiehTODzbWctVLTgyDk0Qco8/6CMN6nZGXjxR2A4r5ExhmwRNsNxd77G+MprkmiJz+7w33ROZ1ouQO5XjD3wbQ3ssqNiTKId6yAUPBvuAZRixVApauTuADc8QWGixqCQcqZzmU7YSBBIPf652/AEYr4Tk64YihoE39pHiK8MRbTLdRt3EF4LSMugPAPM24vCgUv3w1TD3Fj6sDg/6oi3flOV9SJZX4vCiUXbDNEuD/p2aQrEXVbaxweFOHjTe7F4iawjXw3nG3SO8rUBHcxbhDDVx5rjYactbW5QvHWiyla6uLb6o8WHBneg2EjTEwAHOZE/rBjcqmAJb2sVp1E0Kwq8ycGmL69vmqJPC1GqVTohAQvmEkaxIPpfq24Yb9ZPrADA7iEXBKuAQ1FphFUVgJBJGJbd60sOV1Rz1T+gUwS4wCNQ4l3LG1S22+wzUVlEku5DXFnT932tatqTyWEthqPqLCt6dL1+qa94XLpeHagXAx2VGe8n8IlcADtxqS+l8xQ4heT12WO9kC316vqvg1mnsI56faup9hb3eT9ZpKyxSBGYOphlTWfV1Y/v64f5PYvTo4aL0IYHyLY/9Qi72vFmOpPeHBYgD5t3j+H2CsiU1PkYsBggOmD7xW8FDuT6HWVvwhEJqeibVPK0Lhyj6tgvlSIAvFUaSMFPlmwFNmwfj/AHUhr9KuTfsBFTZ10yy9TZVgf+EofwnrxHBaWUgdD40aHoY1VjfG33iEuajb6buxG3pYFyPNhJNzeLZisUKIDRMQpUHrsE22EyrFFran3tZGdtcyIEK4Q1F0ULYzJ6T9iY25/ZgPy3pEAAMZCtqo3s+GjX295fWIHfMcnjMgNUHPjExjWBHa+ggK9iQXkFpBVyYB1ga/+0eiIhiek3PlgtvpDrqF7TsLK+ROiBw2GJ7uaO3EEXOj2GpNBuEJ5CdodhZkwzhwMcSatgDHkUuNVu0iVbF6/MxVdOxWXKO+jCYM6PZk/vAhLYqpPzu2T2Uyz4nkDs2Tiq61ez6FoCrzdHIiyIxVTzUQH8G9FgSmtaZ7GCbqlhnurYgcMciwPzxg0hpAQT+NZw1tVEii9vFSpJJbGJqNhORKfKh/Mu1P/9LOQq7Y0P2FIR3x/eUVEQ7CGv2jVtO5ryGSmKeq/P9Fr54wTPaNiqN2K+leACUznCdUWw8kZo/AsBcrOe4OkRX6k8LC3oeJXy06DEToatxEvPYemUauhxiXRw8nfNMqc4LyJq2bbT0zCgJHoqpozPdNg6AYWcoIobgAGu7ZQGq+oE1MT3GZxotMPe/NUJiAc5YE9Thb5Yf3gyno71pyqPTVl/6IQuh4SUz7rkgwF/aVHEnr4aUYNoc0PEzd2Me0jElsA3GAneq1I/wngutOWgTViTK4Nptr5uIzMVQs9H1rOMJNorP8b02t1NDu010rSsib9GaaJJq4r4iy46laQOxWoU0ex26arYnk+jw4833WSCTVBIprTgizZ+fKjoY0xwXvI2oOvGNEUCtGFvKFORTaQrlaXZIg1toa2BBVNicyONbwnI3KIu3MgGJ2SlCVXJn8oHFppVHFCdwgN1uDzGiKAhjvr0sZTUtXin2f2CszPTbbo=|fUhbVKrr8CSKE7TZJneXpDGraj5YhRrq9ESo206S+BY=";
+
+    public static Organization CreateEnterprise(string name, string domain, int seats, string? publicKey = null, string? privateKey = null)
     {
         return new Organization
         {
@@ -45,7 +43,8 @@ public class OrganizationSeeder
             SyncSeats = true,
             Status = OrganizationStatusType.Created,
             MaxStorageGb = 10,
-            // PublicKey and PrivateKey intentionally not set - caller must generate and assign
+            PublicKey = publicKey ?? _defaultPublicKey,
+            PrivateKey = privateKey ?? _defaultPrivateKey,
         };
     }
 }
@@ -53,8 +52,8 @@ public class OrganizationSeeder
 public static class OrganizationExtensions
 {
     /// <summary>
-    /// Creates an OrganizationUser with hardcoded keys (no SDK calls).
-    /// Used by OrganizationWithUsersRecipe for fast user creation without encryption needs.
+    /// Creates an OrganizationUser with fields populated based on status.
+    /// For Invited status, only user.Email is used. For other statuses, user.Id is used.
     /// </summary>
     public static OrganizationUser CreateOrganizationUser(
         this Organization organization, User user, OrganizationUserType type, OrganizationUserStatusType status)
@@ -99,5 +98,4 @@ public static class OrganizationExtensions
             Status = status
         };
     }
-
 }
--- a/util/Seeder/Recipes/OrganizationWithVaultRecipe.cs
+++ b/util/Seeder/Recipes/OrganizationWithVaultRecipe.cs
@@ -50,9 +50,8 @@ public class OrganizationWithVaultRecipe(
         var orgKeys = sdkService.GenerateOrganizationKeys();
 
         // Create organization via factory
-        var organization = OrganizationSeeder.CreateEnterprise(options.Name, options.Domain, seats);
-        organization.PublicKey = orgKeys.PublicKey;
-        organization.PrivateKey = orgKeys.PrivateKey;
+        var organization = OrganizationSeeder.CreateEnterprise(
+            options.Name, options.Domain, seats, orgKeys.PublicKey, orgKeys.PrivateKey);
 
         // Create owner user via factory
         var ownerUser = UserSeeder.CreateUserWithSdkKeys($"owner@{options.Domain}", sdkService, passwordHasher);


util/DbSeederUtility/README.md
```diff
@@ -28,10 +28,34 @@ DbSeeder.exe <command> [options]
...
+
+# Generate a small test organization with ciphers for manual testing
+DbSeeder.exe organization -n DevOrg -u 2 -d dev.local -c 10
 ```
 
+### Options
+
+| Option | Description |
+|--------|-------------|
+| `-n, --name` | Organization name |
```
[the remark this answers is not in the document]
theMickster: Good catch. This has been resolved.

withinfocus:
This looks awesome and after a quick read makes sense to me, but @Hinton can perform something more thorough and so that you're aligned.

util/Seeder/Recipes/CiphersRecipe.cs
Hinton, line 1: I need to get around to improve the docs here.

A recipe is a cohesive unit, similar to baking a cake. You generally don't make a three layer cake by grabbing three recipes and stack them. The individual steps in a recipe can be composed from factories though.

I.e. A recipe SHALL have a `Seed` method which executes the recipe. It SHALL NOT have any variant of other methods.

util/Seeder/Data/README.md
Hinton, line 1: Can we just plug in something like https://github.com/bchavez/Bogus instead?

util/Seeder/Data/README.md
```diff
@@ -0,0 +1,144 @@
...
+- **Deterministic.** Seeded randomness means same org ID → same test data → reproducible debugging.
+- **Filterable.** `Companies.Filter(type, region, category)` for targeted data selection.
+
+---
+
+## Current Capabilities
+
+### Login Ciphers
+
+- 50 real companies across 3 regions with metadata (category, type, domain)
```
Hinton, line 20: **issue:** I'm not really comfortable using real company names here, can we make slight parodies of them?

There is a high probability these will end up in screenshots in PRs and using real company names there does not seem great.

util/Seeder/.claude/CLAUDE.md
```diff
@@ -0,0 +1,199 @@
...
+### EncString Format
+
+All encrypted fields use the EncString format:
+
+```
+2.{base64_iv}|{base64_data}|{base64_mac}
+│ └──────────┘ └──────────┘ └──────────┘
+│     IV         Ciphertext      HMAC
+└─ Type 2 = AES-256-CBC-HMAC-SHA256
+```
```
Hinton: **issue:** Claude does not need to know EncString, it's an internal abstraction in the SDK which may change at any moment.

util/Seeder/CLAUDE.md
Hinton, line 1: **question:** Can we just lift this up out of the .claude dir?

util/Seeder/.claude/CLAUDE.md
```diff
@@ -0,0 +1,199 @@
...
+
+```json
+{
+  "Name": "2.abc...",
+  "Username": "2.def...",
+  "Password": "2.ghi..."
+}
+```
+
+The `CipherSeeder.TransformToServerCipher()` method performs this flattening.
```
Hinton: This will change in the near future. Are we concerned about this document getting out of date?

util/Seeder/.claude/CLAUDE.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Rust SDK FFI
+
+### Available Functions
+
+| Function                     | Input                 | Output                      |
+| ---------------------------- | --------------------- | --------------------------- |
+| `encrypt_cipher`             | CipherView JSON + key | Cipher JSON                 |
+| `decrypt_cipher`             | Cipher JSON + key     | CipherView JSON             |
+| `generate_organization_keys` | (none)                | Org symmetric key + keypair |
```
Hinton: This is not an extensive list and feels like it will get. out of date (it already is)

util/RustSdk/rust/src/lib.rs
Hinton, line 1: We should split up this file, it's getting too big. Move cipher stuff into a `cipher.rs`?

util/Seeder/CLAUDE.md
theMickster: Sure. that would match what we did in `sdk-internal` in our crates.

util/Seeder/Recipes/CiphersRecipe.cs
theMickster: All good, I will incorporate that into the `README.md` in this pull request as I worked with Claude Code to craft some better project taxonomy and domain knowledge. Then you can provide feedback to that end.

theMickster: @Hinton 
Do you want these existing recipes that do not currently follow your`Seed()` pattern fixed now (links to existing code off main)? I made quite a few changes to them. 

- [OrganizationDomainRecipe.cs](https://github.com/bitwarden/server/blob/main/util/Seeder/Recipes/OrganizationDomainRecipe.cs)
- [GroupsRecipe.cs](https://github.com/bitwarden/server/blob/main/util/Seeder/Recipes/GroupsRecipe.cs)
- [CollectionsRecipe.cs](https://github.com/bitwarden/server/blob/main/util/Seeder/Recipes/CollectionsRecipe.cs)

theMickster: I am considering throwing away all of the changes that I made to these recipes and just go 100% net new recipe. Less reviewers needed and the other teams can go clean up their own work at their leisure

theMickster: I have refactored everything; no need for those other teams to be affected. They can clean up their own work.

util/Seeder/Data/README.md
theMickster: No, I do not think that a use of Bogus only is best for us. 
  1. Real companies = better vault testing - Testing password health against stripe.com is more realistic than
  kuphal-group.net. Users recognize real services. 
  2. Org structures are Bitwarden specific - The Traditional/Spotify/Modern models can map directly to how real customers setup their enterprises and organize collections. We can expand and control the org structures whenever we wan. Bogus has no concept of this.
  
For some scenarios, I think that Bogus could provide assistance and is the better choice {Random user data (names, emails, and other identity data). Plus, it appears to be great with Finance mock data generation. 

For the password generation, the one I crafted with Claude is superior because:
  1. We are in control the exact strength distribution (25% VeryWeak, 30% Weak, etc.). 
  2. We can intentionally reuse passwords (common in real-world breaches) 
  3. I intentionally aligned with the [zxcvbn tool](https://dropbox.tech/security/zxcvbn-realistic-password-strength-estimation) that we mentioned here [Bitwarden Help - Weak Password](https://bitwarden.com/help/reports/#weak-passwords)
  4. Regex in Bogus.. 🤮

util/Seeder/Data/README.md
theMickster: I hear your concern; let's bring that to a wider audience of leaders for a decision. I think that it's outside the scope of those on this pull request.

theMickster: I asked Clayton Adams if we had a company policy on the matter. 

> Hey there! Nope no real policy on it. In fact, in some instances we want to use specific companies for things like autofill to ensure the popular sites work on our product. As long as we aren’t using any production PII data we should be fine. I guess I could maybe see some awkwardness to make a sell if like Delta wants to use Bitwarden but they see American in some screenshots or something, but honestly, as long as it's just in GitHub I wouldn't worry about it too much.

util/Seeder/README.md
```diff
@@ -1,18 +1,192 @@
...
+| -------------- | ---------------------------------------------------- | ------------- |
+| **CipherView** | Plaintext/decrypted form. Human-readable data.       | Never         |
+| **Cipher**     | Encrypted form. All sensitive fields are EncStrings. | Yes           |
 
-Factories are helper classes for creating domain entities and populating them with realistic data. This assist in
-decreasing the amount of boilerplate code needed to create test data in recipes.
+The "View" suffix always denotes plaintext. No suffix means encrypted.
 
-### Recipes
+### EncString Format
```
SaintPatrck: ❓ This was removed from `CLAUDE.md` per [Oscar's observation](https://github.com/bitwarden/server/pull/6896#discussion_r2731351954).

Should it be removed from here too?

theMickster: Yes, removing now

util/Seeder/Data/README.md
theMickster: We acknowledge the comment and concern, but for now, we will be using real company names in vault ciphers. All other data will be bogus / self-mocked.

util/RustSdk/rust/src/cipher.rs
```diff
@@ -0,0 +1,435 @@
...
+/// Both pointers must be valid null-terminated strings.
+#[no_mangle]
+pub unsafe extern "C" fn decrypt_cipher(
+    cipher_json: *const c_char,
+    symmetric_key_b64: *const c_char,
+) -> *const c_char {
+    let cipher_json = match CStr::from_ptr(cipher_json).to_str() {
+        Ok(s) => s,
+        Err(_) => return error_response("Invalid UTF-8 in cipher_json"),
+    };
```
dani-garcia: A lot of these matches can be simplified a bit by using let-else:
```rust
    let Ok(cipher_view_json) = CStr::from_ptr(cipher_view_json).to_str() else {
        return error_response("Invalid UTF-8 in cipher_view_json");
    };
```

util/Seeder/Factories/CipherSeeder.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        return EncryptAndTransform(cipherView, orgKeyBase64, organizationId);
+    }
+
+    private Cipher EncryptAndTransform(CipherViewDto cipherView, string keyBase64, Guid organizationId)
+    {
+        var viewJson = JsonSerializer.Serialize(cipherView, SdkJsonOptions);
+        var encryptedJson = _sdkService.EncryptCipher(viewJson, keyBase64);
+
+        if (encryptedJson.Contains("\"error\""))
```
dani-garcia: This way of handling errors seems to leak the internal structure that we're using to declare errors through FFI, which seems a bit fragile. I think this check would be better inside `EncryptCipher` directly, (or maybe rename `TakeAndDestroyRustString` as `ParseResponse` and throw from there if you receive an error so all functions benefit?)

util/DbSeederUtility/README.md
```diff
@@ -28,10 +28,38 @@ DbSeeder.exe <command> [options]
...
 # Generate an organization called "seeded" with 10000 users using the @large.test email domain.
-# Login using "admin@large.test" with password "asdfasdfasdf"
+# Login using "owner@large.test" with password "asdfasdfasdf"
 DbSeeder.exe organization -n seeded -u 10000 -d large.test
+
+# Generate an organization with 5 users and 100 encrypted ciphers
+DbSeeder.exe organization -n TestOrg -u 5 -d test.com -c 100
+
+# Generate with Spotify-style collections (tribes, chapters, guilds)
+DbSeeder.exe organization -n TestOrg -u 10 -d test.com -c 50 -o Spotify
```
MGibson1: isn't this a `vault-organization` command? It doesn't look like the organization command has `-c` or `-o` options

util/DbSeederUtility/Program.cs
```diff
@@ -36,4 +40,25 @@ string domain
         var recipe = new OrganizationWithUsersRecipe(db);
         recipe.Seed(name: name, domain: domain, users: users);
     }
+
+    [Command("vault-organization", Description = "Seed an organization with users and encrypted vault data (ciphers, collections, groups)")]
```
MGibson1, line 44: is this functionally a different command that `organization`? It feels like an expansion of the capabilities of that command with more options

util/DbSeederUtility/README.md
```diff
@@ -28,10 +28,38 @@ DbSeeder.exe <command> [options]
...
+| `-c, --ciphers` | Number of encrypted ciphers to create (optional) |
+| `-o, --org-structure` | Collection structure: Traditional, Spotify, or Modern (optional) |
+
+### Notes
+
+- All users are created with the password `asdfasdfasdf`
+- The owner account is always `owner@{domain}` with Confirmed status
+- Member accounts are `user0@{domain}`, `user1@{domain}`, etc.
+- Without `--org-structure`, a single "Default Collection" is created
+- With `--org-structure`, collections are created from org model (e.g., Spotify creates tribes, chapters, guilds)
```
MGibson1: I'd appreciate some recommendations on when to use each of these options.

util/DbSeederUtility/VaultOrganizationArgs.cs
```diff
@@ -0,0 +1,102 @@
...
+    [Option('c', "ciphers", Description = "Number of login ciphers to create (minimum 1)")]
+    public int Ciphers { get; set; }
+
+    [Option('g', "groups", Description = "Number of groups to create (minimum 1)")]
+    public int Groups { get; set; }
+
+    [Option('m', "mix-user-statuses", Description = "Use realistic status mix (85% confirmed, 5% each invited/accepted/revoked). Requires >= 10 users.")]
+    public bool MixStatuses { get; set; } = true;
+
+    [Option('o', "org-structure", Description = "Org structure for collections: Traditional, Spotify, or Modern")]
```
MGibson1, line 31: It would be nice to include descriptions of valid options in command help output

util/DbSeederUtility/VaultOrganizationArgs.cs
```diff
@@ -0,0 +1,102 @@
...
+    public string? Structure { get; set; }
+
+    [Option('r', "region", Description = "Geographic region for names: NorthAmerica, Europe, AsiaPacific, LatinAmerica, MiddleEast, Africa, or Global")]
+    public string? Region { get; set; }
+
+    public void Validate()
```
MGibson1, line 37: Sorry, I'm not familiar with this cli command framework, but shouldn't validate also check the structure string complies with the enum options?

util/Seeder/Data/UsernamePatterns.cs
```diff
@@ -0,0 +1,57 @@
...
+
+internal sealed record UsernamePattern(
+    UsernamePatternType Type,
+    string FormatDescription,
+    Func<string, string, string, string> Generate);
+
+/// <summary>
+/// Username pattern implementations for different email conventions.
+/// </summary>
+internal static class UsernamePatterns
```
MGibson1, line 13: All of these patterns assume emails, surely we want one which is not an email address.

util/Seeder/Factories/FolderSeeder.cs
MGibson1, line 1: Is this usable? I do not believe a folder data field is present on ciphers

util/Seeder/Factories/OrganizationSeeder.cs
```diff
@@ -1,12 +1,17 @@
...
-using Bit.Infrastructure.EntityFramework.AdminConsole.Models;
 
 namespace Bit.Seeder.Factories;
 
 public class OrganizationSeeder
 {
+    /// <summary>
+    /// Creates an enterprise organization without encryption keys.
+    /// Keys should be generated dynamically using RustSdkService.GenerateOrganizationKeys()
+    /// and assigned to PublicKey/PrivateKey after creation.
```
MGibson1: I would like to ensure this with some type safety through type-state by creating something like an `OrganizationSeed` (name overloaded but I mean something that can become an organization with more input), which has some method on it to populate with required data.

a method on the `OrganizationSeed` would have a `ToOrganization(publicKey: ___, privateKey: ___)` that return the fully formed organization. That way we protect ourselves from scene or factory usages that forget to populate

This pattern should be used anywhere additional data is require to produce a complete object.

Another way to get this effect is to create and return a builder pattern that throws when trying to build an `Organization` without all the required data. I'm fine with that (or any) pattern too. The point is the types themselves push you toward the right behavior

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+
+**Server Structure (flat, stored in Cipher.Data):**
+
+```json
+{ "Name": "2.x...", "Username": "2.y...", "Password": "2.z..." }
+```
+
+The seeder transforms SDK output to server format before database insertion.
+
+### Key Hierarchy
```
MGibson1: This is a weird thing to document here

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+Organization Key (or User Key)
+       │
+       ├──▶ Encrypts Cipher.Key (optional per-cipher key)
+       │
+       └──▶ Encrypts cipher fields directly (if no per-cipher key)
+```
+
+For seeding, we encrypt directly with the organization key.
+
+### Entity Relationships
```
MGibson1: This is a weird thing to document here

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+**Metaphor:** Skilled craftspeople who create one perfect item per call.
+
+**When to use:** Need to create ONE entity (user, cipher, collection) with proper encryption.
+
+**Key characteristics:**
+
+- Create ONE entity per method call
+- Handle encryption/transformation internally
+- Stateless (except for SDK service dependency)
+- Return fully-formed entity ready for persistence
```
MGibson1: Not necessarily. Your OrganizationSeeder does not, for example

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+
+**Metaphor:** Theater scenes with multiple actors and props arranged to tell a complete story.
+
+**When to use:** Need a complete test scenario with proper ID mangling for test isolation.
+
+**Key characteristics:**
+
+- Implement `IScene<TRequest>` or `IScene<TRequest, TResult>`
+- Create complete, realistic test scenarios
+- Handle ID mangling for test isolation
```
MGibson1:
```suggestion
- Handle uniqueness constraint mangling for test isolation
```

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+- Return `SceneResult` with mangle map for test assertions
```
MGibson1:
```suggestion
- Return `SceneResult` with mangle map and optional additional operation result data for test assertions
```

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+
+**Naming:** `{Scenario}Scene` class with `SeedAsync(Request)` method (defined by interface)
+
+#### Queries
+
+**Purpose:** Read-only data retrieval for test assertions and verification.
+
+**Metaphor:** Information desks that answer questions without changing anything.
+
+**When to use:** Need to READ existing seeded data for verification or follow-up operations.
```
MGibson1, line 118:
```suggestion
**When to use:** Need to READ existing seeded data for verification or follow-up operations.

** Example:** Inviting a user to an organization produces a magic link to accept the invite, a query should be used to retrieve that link because it is easier than interfacing with an external smtp catcher.
```

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+
+**Key characteristics:**
+
+- Implement `IQuery<TRequest, TResult>`
+- Synchronous operations (vs. async Scenes)
```
MGibson1: seems unlikely we actually want this, but we can address that later.

util/Seeder/README.md
```diff
@@ -1,18 +1,179 @@
...
+- Read-only (no database modifications)
+- Return typed data for test assertions
+- Often used to verify Scene/Recipe results
```
MGibson1: I would see verifying scene a recipe results as something the scene/recipe should have returned

```suggestion
- Can be used to retrieve side effects due to tested flows
```

util/Seeder/CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+
+The Seeder follows six core patterns:
+
+1. **Factories** - Create ONE entity with encryption. Named `{Entity}Seeder` with `Create{Type}{Entity}()` methods. Do not interact with database.
+
+2. **Recipes** - Orchestrate MANY entities. Named `{DomainConcept}Recipe`. **MUST have `Seed()` method** as primary interface, not `AddToOrganization()` or similar. Use parameters for variations, not separate methods. Compose Factories internally.
+
+3. **Models** - DTOs bridging SDK ↔ Server format. Named `{Entity}ViewDto` (plaintext), `Encrypted{Entity}Dto` (SDK format). Pure data, no logic.
+
+4. **Scenes** - Complete test scenarios with ID mangling. Implement `IScene<TRequest>`. Async, returns `SceneResult` with MangleMap. Named `{Scenario}Scene`.
```
MGibson1:
```suggestion
4. **Scenes** - Complete test scenarios with ID mangling. Implement `IScene<TReques, TResult>`. Async, returns `SceneResult<TResult>` with MangleMap and result property populated with `TResult`. Named `{Scenario}Scene`.
```

util/Seeder/CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+- `GroupsRecipe` - Uses `AddToOrganization()` instead of `Seed()`
+- `OrganizationDomainRecipe` - Uses `AddVerifiedDomainToOrganization()` instead of `Seed()`
+
+## Pattern Decision Tree
+
+```
+Need to create test data?
+├─ ONE entity with encryption? → Factory
+├─ MANY entities as cohesive operation? → Recipe
+├─ Complete test scenario with ID mangling? → Scene
```
MGibson1: The distinction for me is whether the recipe/scene will be served from the seeder api and therefore used by clients for e2e testing.

MGibson1 (commented):
# Seeding
I really like the data production stuff you put together, especially the consistent generation thinking.

What is unclear to me is, since you're accepting a seed rather than a deterministic Random instance, are you using the same value everywhere or somehow iterating it? It seems like everything accepting a required, shared Random and a deterministic usage order would produce predictable results while ensuring we aren't just using the same array index for everything. 

# SDK drift risk

What is the challenge this approach presents regarding drift between the SDK shim being used here, the shimless SDK used in clients, and the non-sdk client code?

theMickster:
> # Seeding
> I really like the data production stuff you put together, especially the consistent generation thinking.
> 
> What is unclear to me is, since you're accepting a seed rather than a deterministic Random instance, are you using the same value everywhere or somehow iterating it? It seems like everything accepting a required, shared Random and a deterministic usage order would produce predictable results while ensuring we aren't just using the same array index for everything. 
> 
> # SDK drift risk
> 
> What is the challenge this approach presents regarding drift between the SDK shim being used here, the shimless SDK used in clients, and the non-sdk client code?

I see where you're headed and I agree that the pseudorandom data points aren't perfect and I agree that the Rust SDK interface is rough and needs refactoring, but that's the point of this POC at this point. Get it working + show meaningful return on investment and iterate fast with byte sized PRs.

For me, this is the first of 5-7 PRs that I want to churn through in the next 2 weeks.

What I do not want (nor what I really ever do with POC work like this) is to have one PR continue on for weeks with many commits and code churn that ends up being impossible for human review.

So big question for the team is whether or not we can accept the work as a feasible starting point knowing that we're going to idéate **fast** with many byte sized PRs in the next 2-3 weeks to get us just one Organization + Vault Recipe knowing that we've got a whole heck of a lot more planned beyond just this little thing here.

util/DbSeederUtility/README.md
theMickster: Yes, I have a local fix ready for this.

util/DbSeederUtility/Program.cs
theMickster: I am intentionally not touching the things written by the admin console team. Too many unwanted side affects. If we want to merge later then great, but not now.

util/DbSeederUtility/VaultOrganizationArgs.cs
theMickster: Sure, let's discuss that towards the end of the cycle in a future PR once we have all the Cipher types working well.

util/Seeder/Data/UsernamePatterns.cs
theMickster: Agreed, but again, let's add that to the mix in a follow-up PR.

util/Seeder/Factories/FolderSeeder.cs
theMickster: Usable? Not sure what you mean... Folders are being made nicely. I have yet to build in some logic to assign Ciphers to folders. Again, that's coming in follow up PRs.

util/Seeder/Factories/OrganizationSeeder.cs
theMickster: Sure, but can that improvement and refactoring wait for follow-up PRs?

util/DbSeederUtility/README.md
theMickster: Let's remove these for now & add that as a TODO for a follow-up. 
I think many of these could be done better with a more all-encompassing how to document. I don't think a notes section does it justice.

util/Seeder/README.md
theMickster: 👍🏼 Removed

util/Seeder/README.md
theMickster: 👍🏼 Removed

util/Seeder/README.md
theMickster: 👍🏼 Removed

util/Seeder/README.md
theMickster: 👍🏼 Removed

util/RustSdk/RustSdkService.cs
theMickster, line 1: @Hinton & @MGibson1 : Sonar Cloud is barking about the `RustSdkService` and it's members being marked as static. I am not considering it for this PR because it's an intrusive change at this point IMO, but is that something for the TODO list?

util/Seeder/Factories/CipherSeeder.cs
theMickster: I moved the brittle, out-of-place error handling out of the `CipherSeeder` and into the `RustSdkService`.
Are we okay as-is for now knowing that more work is coming immediately after? I am trying to keep the blast radius down while we flush out this POC work.
Thoughts @dani-garcia  & @Hinton?

util/Seeder/Factories/CipherSeeder.cs
theMickster, line 1: @Hinton and @MGibson1 
📝 I don't _love_ the size and complexity that I am envisioning for building complex Ciphers (I have some local changes for the other Cipher types in this file and 🤮).  I'm thinking of a strategy pattern long-term for crafting Ciphers.

util/Seeder/Recipes/OrganizationWithVaultRecipe.cs
theMickster, line 1: @Hinton and @MGibson1 
📝 I don't _love_ the size and complexity that I am envisioning for crafting complex organizations. I am not ready to split this up into other _Ingredients_ 😉 yet, but I do think that we will have more sustained success if we create additional creational pattern(s) for organizations.  I'll gather us in the coming weeks to discuss ideas.

util/Seeder/Factories/CipherSeeder.cs
dani-garcia: Yeah I think that's good enough for now. If we need to improve it in the future at least the logic is self contained.

util/Seeder/Factories/OrganizationSeeder.cs
MGibson1: yes, but I think this is a high priority one, since it breaks the documented guarantees of what seeders produce

util/RustSdk/RustSdkService.cs
MGibson1: I don't see any comments by sonar cloud on this PR, can you link to the actual complaint? I don't see any issue with the pattern, isn't that generated by csbindgen, too?

theMickster: Sonar in VS Code was where I saw these. I _think_ I am connected to our instance (I'll double-check). It's the standard issue static refactor. 🤔

> Methods and properties that don't access instance data should be static (csharpsquid:S2325)
image

util/DbSeederUtility/VaultOrganizationArgs.cs
theMickster: I added calls to the transforms so that the other inputs are validated. 
Not sure how long these will last once I start to experiment on a Domain Specific Language (DSL) input file 🤷🏼

util/DbSeederUtility/VaultOrganizationArgs.cs
theMickster: Added to the hitlist https://bitwarden.atlassian.net/browse/PM-31403

util/Seeder/Data/UsernamePatterns.cs
theMickster: Added to the hit list
https://bitwarden.atlassian.net/browse/PM-31403

util/Seeder/Factories/CipherSeeder.cs
theMickster: Added to the hit list https://bitwarden.atlassian.net/browse/PM-31403

util/Seeder/Factories/OrganizationSeeder.cs
theMickster: @MGibson1 I decided to revert the big changes that I made in favor of a simpler approach for now. I would like to tackle the removal of the hardcoded keys in a small PR that only contains those changes because it will cause changes to things owned by admin console that will just slow this whole thing down. 

Let's make a plan for a builder pattern because we are going to need it shortly given the wide variety of organizations to craft. Is that okay?

util/Seeder/Factories/OrganizationSeeder.cs
```diff
@@ -1,13 +1,16 @@
-﻿using Bit.Core.Billing.Enums;
+﻿using Bit.Core.AdminConsole.Entities;
+using Bit.Core.Billing.Enums;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
-using Bit.Infrastructure.EntityFramework.AdminConsole.Models;
 
 namespace Bit.Seeder.Factories;
 
 public class OrganizationSeeder
```
theMickster, line 8: We have an item added to the hit list to refactor this and the associated integration tests owned by admin console. 
https://bitwarden.atlassian.net/browse/PM-31403

dani-garcia (approved):

PR #6912, opened 2026-01-28
Have AppSec own Checkmarx config

withinfocus:
## 🎟️ Tracking

Noticed while reviewing a PR.

## 📔 Objective

Indicates AppSec should own configuration in the `.checkmarx` folder. Also adds some sensible additions to the configuration for development files.

### DIFF
--- a/.checkmarx/config.yml
+++ b/.checkmarx/config.yml
@@ -11,3 +11,7 @@ checkmarx:
         filter: "!test"
       kics:
         filter: "!dev,!.devcontainer"
+      sca:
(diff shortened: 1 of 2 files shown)


mandreko-bitwarden:
It's not based on this PR, but while reviewing this PR, I did notice that there's a conflict on a couple items:

`**/Dockerfile` and `**/entrypoint.sh` are listed twice in the file. The latest changes from #6257 will not be applied here and might need adjustment. Looping in @fntyler

PR #6918, opened 2026-01-28
[PM-31140] Consolidate `unpaid` subscription handling

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31140

## 📔 Objective

Consolidates the handling of subscriptions that transition to `unpaid` status across all entity types (Premium users, Organizations, and Providers).

Previously, each entity type was handled differently:
- **Premium users**: Subscription was immediately canceled
- **Organizations**: A Quartz job was scheduled to cancel the subscription after 7 days  
- **Providers**: `cancel_at` was set on the subscription to 7 days in the future

This PR standardizes on the Provider approach for all entity types:
1. Disable the subscriber (user premium, organization, or provider)
2. Set the subscription's `cancel_at` to 7 days in the future

This gives customers a consistent 7-day grace period to resolve payment issues before their subscription is canceled, regardless of subscription type.

**Additional improvements:**
- Removes the Quartz scheduler dependency for organization cancellation
- Adds `ProrationBehavior.None` to prevent unintended proration charges during cancellation/reactivation
- Includes `CancellationDetails.Comment` for better audit trail when setting cancellation

**Testing:** Unit tests in `SubscriptionUpdatedHandlerTests.cs` were updated to reflect the consolidated behavior and verify the new subscription update options.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/Api/Auth/Models/Request/Accounts/PasswordRequestModel.cs
+++ b/src/Api/Auth/Models/Request/Accounts/PasswordRequestModel.cs
@@ -1,7 +1,5 @@
-﻿#nullable enable
-
-using System.ComponentModel.DataAnnotations;
-using Bit.Api.KeyManagement.Models.Requests;
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Core.KeyManagement.Models.Api.Request;
 
 namespace Bit.Api.Auth.Models.Request.Accounts;
 
--- a/src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
+++ b/src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
@@ -1,5 +1,4 @@
 ﻿using System.ComponentModel.DataAnnotations;
-using Bit.Api.KeyManagement.Models.Requests;
 using Bit.Core.Auth.Models.Api.Request.Accounts;
 using Bit.Core.Auth.Models.Data;
 using Bit.Core.Entities;
--- a/src/Api/Controllers/SsoCookieVendorController.cs
+++ b/src/Api/Controllers/SsoCookieVendorController.cs
@@ -1,0 +1,119 @@
+﻿using Bit.Core.Settings;
+using Bit.Core.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Bit.Api.Controllers;
+
+/// <summary>
+/// Provides an endpoint to read an SSO cookie and redirect to a custom URI
+/// scheme. The load balancer/reverse proxy must be configured such that
+/// requests to this endpoint do not have the auth cookie stripped.
+/// </summary>
+[Route("sso-cookie-vendor")]
+[SelfHosted(SelfHostedOnly = true)]
+public class SsoCookieVendorController(IGlobalSettings globalSettings) : Controller
+{
+    private readonly IGlobalSettings _globalSettings = globalSettings;
+    private const int _maxShardCount = 20;
+    private const int _maxUriLength = 8192;
+
+    /// <summary>
+    /// Reads SSO cookie (shards supported) and redirects to the bitwarden://
+    /// URI with cookie value(s).
+    /// </summary>
+    /// <returns>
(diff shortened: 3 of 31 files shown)


src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -62,183 +54,147 @@ public SubscriptionUpdatedHandler(
...
         }
-    }
 
-    private async Task<bool> IsPremiumSubscriptionAsync(Subscription subscription)
-    {
-        var premiumPlans = await _pricingClient.ListPremiumPlans();
-        var premiumPriceIds = premiumPlans.SelectMany(p => new[] { p.Seat.StripePriceId, p.Storage.StripePriceId }).ToHashSet();
-        return subscription.Items.Any(i => premiumPriceIds.Contains(i.Price.Id));
-    }
+        var now = subscription.TestClock?.FrozenTime ?? DateTime.UtcNow;
```
kdenney, line 179: Oh man, I'm filing this away to remember in the future. I am now wondering if I should have used this instead of using the invoice period end in my current PR haha

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -62,183 +54,147 @@ public SubscriptionUpdatedHandler(
...
+            Status:
+            SubscriptionStatus.Incomplete or
+            SubscriptionStatus.Unpaid
+        } && currentSubscription is
+        {
+            Status: SubscriptionStatus.Active,
+            LatestInvoice.BillingReason: BillingReasons.SubscriptionCreate or BillingReasons.SubscriptionCycle
+        };
+
+    private Task DisableSubscriberAsync(SubscriberId subscriberId, DateTime? currentPeriodEnd) =>
```
kdenney, line 128: ❓ What do you think about moving these methods that operate on a SubscriberId into SubscriberService? It would help clean up this file even more, they will be public so unit tests can target the methods individually which is always a bonus, and I could even see a possibility for these to be used elsewhere in the application at some point so having them in a service makes that easy.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -318,75 +274,6 @@ private async Task RemovePasswordManagerCouponIfRemovingSecretsManagerTrialAsync
         }
     }
 
-    private async Task ScheduleCancellationJobAsync(string subscriptionId, Guid organizationId)
-    {
-        var scheduler = await _schedulerFactory.GetScheduler();
-
-        var job = JobBuilder.Create<SubscriptionCancellationJob>()
```
kdenney, line 325: ❓ This job is now unused. I assume that removing `SubscriptionCancellationJob` from the codebase would cause any scheduled jobs in the queue at the time of release to throw exceptions? I also assume that's why you didn't delete it in this PR? 😆 Is there a follow-up ticket for deleting it?

kdenney (approved):
Nice job!! 🎉 This is way cleaner. I have just a couple of non-blocking questions.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
amorask-bitwarden: Yeah if we remove the whole thing now we'll get a `TypeLoadException` in Quartz when the remaining scheduled jobs run. The earliest the deployment of the job can be removed, I believe, would be 7 days after the release of this PR. I have not gotten a ticket cut for it yet because this hasn't been merged or release so it wouldn't be referencing Production code, but will do so upon this ticket being merged.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
amorask-bitwarden: Great call - took a few changes cuz the `IPushNotificationAdapter` was living in `Billing`, but nothing too crazy. The commit is [here](https://github.com/bitwarden/server/pull/6918/commits/fe0bc1516b86a44cfadc7c2cd31bbb269051046e).

kdenney (approved):

PR #6950, opened 2026-02-04
[PM-21925] Add MasterPasswordSalt Column to User Table

ike-kottlowski:
## 🎟️ Tracking

[PM-21925](https://bitwarden.atlassian.net/browse/PM-21925)

## 📔 Objective

To support the separation of salt from the email we need to track the salt in the user table. This is an additive change. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-21925]: https://bitwarden.atlassian.net/browse/PM-21925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Entities/User.cs
mzieniukbw, line 119: ❓  Are we going to change this too in follow up PR ?

src/Sql/dbo/KeyManagement/Stored Procedures/User_UpdateMasterPassword.sql
```diff
@@ -24,7 +25,8 @@ BEGIN
         [KdfMemory] = @KdfMemory,
         [KdfParallelism] = @KdfParallelism,
         [RevisionDate] = @RevisionDate,
-        [AccountRevisionDate] = @AccountRevisionDate
+        [AccountRevisionDate] = @AccountRevisionDate,
+        [MasterPasswordSalt] = LOWER(LTRIM(RTRIM([User].[Email])))
```
mzieniukbw: It's better to use MasterPasswordSalt from SP arg, make it required and do any mapping (lower, trim) in C# code.
Especially that MP and MP Salt goes together, if you have one, you should expect both. In that case, the `MasterPasswordSalt` SP should be set directly in dapper UserRepository.

util/Migrator/DbScripts_finalization/2026-02-03_00_AlterUserMasterPasswordSaltNotNull.sql
```diff
@@ -0,0 +1,16 @@
+/*
+    Script to finalize the addition of MasterPasswordSalt to User table by making it NOT NULL.
+    This script should be run only after all existing rows have been backfilled with non-NULL values from
+    the transition script 2026-02-03_00_PopulateMasterPasswordSaltColumn.sql.
```
mzieniukbw: The script is called [2026-02-03_00_HydrateMasterPasswordSaltColumn.sql‎](https://github.com/bitwarden/server/pull/6950/changes#diff-a70c7ec16303911ef40b6556753b23bc8862315b3975642df53c74970a35c262)

util/Migrator/DbScripts_transition/2026-02-03_00_HydrateMasterPasswordSaltColumn.sql
```diff
@@ -0,0 +1,65 @@
...
+    ;WITH
+        batch
+        AS
+        (
+            SELECT TOP (1000)
+                [Id]
+            -- smaller batch size to reduce IO/blocking
+            FROM [dbo].[User] WITH (INDEX([IX_User_MasterPasswordSalt]))
+            WHERE [MasterPasswordSalt] IS NULL
+                AND [MasterPassword] IS NOT NULL -- Only backfill users who have a MasterPassword, as those without can have NULL salt
```
mzieniukbw: Question to @bitwarden/dept-dbops, since all users will have `MasterPasswordSalt IS NULL` initially (matching index) but there exist no index for `[MasterPassword] IS NOT NULL`, would the `TOP (1000)` prevent the query going into full table scan plan ?
Or is my suspicion right, that this will cause full table scan multiple times ? (or more like full index scan plan?)

util/Migrator/DbScripts_finalization/2026-02-03_00_AlterUserMasterPasswordSaltNotNull.sql
```diff
@@ -0,0 +1,16 @@
...
+*/
+
+-- Guard: refuse to finalize if any NULLs remain
+IF EXISTS (SELECT 1
+FROM [dbo].[User]
+WHERE
+    [MasterPassword] IS NOT NULL AND
+    [MasterPasswordSalt] IS NULL)
+    THROW 50000, 'MasterPasswordSalt contains NULLs for MasterPassword users; cannot finalize.', 1;
```
mzieniukbw: @bitwarden/dept-dbops If we run this after `util/Migrator/DbScripts_transition/2026-02-03_00_HydrateMasterPasswordSaltColumn.sql` query, but the temporary index is already gone, is that full table scan plan (since it does not match any index) ?

src/Sql/dbo/Stored Procedures/User_Update.sql
```diff
@@ -44,11 +44,18 @@
     @VerifyDevices BIT = 1,
     @SecurityState VARCHAR(MAX) = NULL,
     @SecurityVersion INT = NULL,
-    @SignedPublicKey VARCHAR(MAX) = NULL
+    @SignedPublicKey VARCHAR(MAX) = NULL,
+    @MasterPasswordSalt NVARCHAR(256) = NULL
```
mzieniukbw, line 49: ❓  What is the point of the `MasterPasswordSalt` if we always override it few lines below ?
I honestly don't think we should do any transformations like `LOWER(LTRIM(RTRIM(@Email)))` in stored procedures, except for migrations.

ike-kottlowski: We wanted to be as defensive as possible with these changes, since a mismatch salt could cause irreparable damage to a vault. Placing the `LOWER(LTRIM(RTRIM(@Email)))` here guarantees that it will be set to what we know is correct for the time being. This does arguably create more migration overhead in the future.

Allowing the input to be null protects backwards compatibility where the server can be rolled back without impacting the database. Also we will be accepting the `@MasterPasswordSalt` in the future so this should be viewed as transitional.

The important thing is we don't accidently set a `MasterPasswordSalt` when a user doesn't have a `MasterPassword` which is why I opted the option to overwrite a few lines down. We overwrite to `@Email` now but, eventually when we are certain the Salt is being sent, we will change to the `@MasterPasswordSalt`.

[PM-30355](https://bitwarden.atlassian.net/browse/PM-30355) is the finalization task to remove the hard-coding of the `@Email`.

src/Sql/dbo/KeyManagement/Stored Procedures/User_UpdateMasterPassword.sql
ike-kottlowski: We cannot make the added input required as that is a breaking change for the UserRepository. It is also contradictory to the [EDD documentation](https://contributing.bitwarden.com/contributing/database-migrations/edd#example). 

We can make it required when we are certain the server won't be rolled back to a version that doesn't know about the `@MasterPasswordSalt` input.

src/Core/Entities/User.cs
ike-kottlowski: Yes, I think we should. 

What do you think that should look like? I'm leaning removing the method altogether and just using `User.MasterPasswordSalt`.

util/Migrator/DbScripts/2026-02-11_00_AlterUserAddMasterPasswordSalt.sql
```diff
@@ -0,0 +1,373 @@
...
+        [VerifyDevices],
+        [SecurityState],
+        [SecurityVersion],
+        [SignedPublicKey],
+        [MasterPasswordSalt]
+    FROM
+        [dbo].[User]
+GO
+
+EXECUTE sp_refreshview '[dbo].[UserView]'
```
mkincaid-bw: Since you are modifying `UserView`, there's no need to refresh it (the modification effectively does that). However, you should refresh other views that reference the `User` table (even if they aren't using the new column)

```sql
EXEC sp_refreshview N'[dbo].[EmergencyAccessDetailsView]';
EXEC sp_refreshview N'[dbo].[ProviderUserUserDetailsView]';
EXEC sp_refreshview N'[dbo].[UserEmailDomainView]';
```

util/Migrator/DbScripts_transition/2026-02-03_00_HydrateMasterPasswordSaltColumn.sql
mkincaid-bw: I tested creating the index in this script, as well as creating an index including the `MasterPassword` column (separately) on a copy of the prod DB, and surprisingly, the index with both columns performed far worse. The two column index does an index seek across the entire index, then gets the top 1000 Id's. The original index does a short index seek, then a key lookup. I'm actually a little surprised by this, and I'm currently testing some other scenarios.

util/Migrator/DbScripts/2026-02-11_00_AlterUserAddMasterPasswordSalt.sql
ike-kottlowski: Fixed in https://github.com/bitwarden/server/pull/6950/commits/869eda58c203d663d5e5a65510ed863ced8e8d7c

util/Migrator/DbScripts_finalization/2026-02-03_00_AlterUserMasterPasswordSaltNotNull.sql
ike-kottlowski: updated in https://github.com/bitwarden/server/pull/6950/commits/66760ef7f03719c7d74ba8ba1f0dc2f1aaac7cdc

util/Migrator/DbScripts_transition/2026-02-03_00_HydrateMasterPasswordSaltColumn.sql
mkincaid-bw: I tested this as-is, along with a modified version of the index (leaving out the `ORDER BY`), and both options took around 6 hours to complete, but that was with me changing the `WAITFOR` from 1 second to 0.1 seconds. Running it with the existing 1 second delay could add another 3-4 hours to the script.

util/Migrator/DbScripts_finalization/2026-02-03_00_AlterUserMasterPasswordSaltNotNull.sql
mkincaid-bw: Correct, it will do a full table scan. However, since it's just doing an EXISTS query, it will short-circuit if a matching row is found. Even if it scans the full table, it will probably take several minutes (though not hours or anything that long).

ike-kottlowski: Per internal discussion we will enforce the constraint implicitly on the server and other flows. Verified by Unit and Integration testing.

This file has been removed as vestigial.

src/Sql/dbo/Stored Procedures/User_Update.sql
ike-kottlowski: This has been addressed as Business Logic and should not exist in the database. The writing of the email to the MasterPasswordSalt has been moved to the server code.

src/Sql/dbo/KeyManagement/Stored Procedures/User_UpdateMasterPassword.sql
ike-kottlowski: This has been addressed by moving to the `UserRepository` to manage the setting of the `MasterPasswordSalt` to `User.Email`.

util/Migrator/DbScripts_transition/2026-02-03_00_HydrateMasterPasswordSaltColumn.sql
ike-kottlowski: The hydration script has been moved out and into a different task so that it can be executed out of band this the initialization work to maintain momentum.

[JIRA task split to](https://bitwarden.atlassian.net/browse/PM-32795)

eliykat (dismissed):
Easy review from me - deleting a transient test only. Have not reviewed the rest of the changes.

src/Core/Entities/User.cs
JaredSnider-Bitwarden: @ike-kottlowski because this was added before we built the DB side to add the salt, in my discussions with Patrick and Todd, we determined that we should just keep this and have it be:
`return MasterPasswordSalt ?? Email.ToLowerInvariant().Trim()`

We will also want to put TODO: PM-whatever to clean this up in the final clean up steps to just rip out the accessor and replace usages w/ `User.MasterPasswordSalt`

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -182,12 +182,16 @@ public async Task<DateTime> GetAccountRevisionDateAsync(Guid id)
 
     public override async Task<User> CreateAsync(User user)
     {
+        // Todo: PM-30355: once MasterPasswordSalt is fully detached from Email, MasterPasswordSalt should not be overwritten
+        user.MasterPasswordSalt = user.MasterPassword != null ? user.Email.ToLowerInvariant().Trim() : null;
```
JaredSnider-Bitwarden: ⚠️ :  This logic doesn't allow us to start migrating clients to start sending salt. If they send in a salt that is set on the user, this logic will ignore it and overwrite it.

To start moving us toward being able to accept salts from clients, we need logic like:
```
if(user.MasterPassword != null && string.IsNullOrEmpty(user.MasterPasswordSalt)) {
   user.MasterPasswordSalt = user.Email.ToLowerInvariant().Trim()
}
```
This will give us two main benefits:
1. As soon as this rolls out, new users will start getting salt columns populated w/ their emails. Thus, working towards eventual consistency. 
2. As soon as we update a client to start sending in salt for registration, this will accept the salt without having to be changed again and store it in the db. 

This feedback applies in a few other cases we can discuss next week.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -533,7 +561,8 @@ public UpdateUserData SetMasterPassword(Guid userId, MasterPasswordUnlockData ma
             userEntity.KdfParallelism = masterPasswordUnlockData.Kdf.Parallelism;
             userEntity.RevisionDate = timestamp;
             userEntity.AccountRevisionDate = timestamp;
-            //TODO- PM-30355: Update MasterPasswordSalt to the MasterPasswordUnlockData.Salt instead of matching Email.
+            //TODO- PM-30355: Setting to the Salt here would be safe since the call sites for calling this method use the email as the salt,
+            // but this is an implicit contract that should be revisited. So we set to email defensively.
```
mzieniukbw: ❓ I don't understand why we make this distinction, while we just use `masterPasswordUnlockData.Salt` in Dapper implementation. Is there any downside to not do the same now ?

src/Core/Entities/User.cs
```diff
@@ -111,8 +111,7 @@ public class User : ITableObject<Guid>, IStorableSubscriber, IRevisable, ITwoFac
     /// Allows clients to unlock vault after V1 to V2 key rotation without logout.
     /// </summary>
     public string? V2UpgradeToken { get; set; }
-    // PM-28827 Uncomment below line.
-    // public string? MasterPasswordSalt { get; set; }
+    public string? MasterPasswordSalt { get; set; }
```
mkincaid-bw, line 115: ❓ This column is limited to 256 characters in SQL Server but not in the other DB platforms. Should this have `[MaxLength(256)]`?

util/Migrator/DbScripts_transition/2026-02-23_01_HydrateMasterPasswordSaltColumn.sql
mkincaid-bw, line 1: ❓ Do you know if this script will still run as part of the cloud deployment, or is this only for self-host?

ike-kottlowski: It does not, I left it in for reference. If it's confusing, I can remove it.

util/Migrator/DbScripts/2026-02-25_02_AlterUpdateMasterPassword.sql
```diff
@@ -0,0 +1,33 @@
...
+    @MasterPassword NVARCHAR(300),
+    @MasterPasswordHint NVARCHAR(50) = NULL,
+    @Key VARCHAR(MAX),
+    @Kdf TINYINT,
+    @KdfIterations INT,
+    @KdfMemory INT = NULL,
+    @KdfParallelism INT = NULL,
+    @RevisionDate DATETIME2(7),
+    @AccountRevisionDate DATETIME2(7),
+    @MasterPasswordSalt NVARCHAR(256) = NULL -- NULL for backwards compat.;
```
mkincaid-bw: ⛏️ The ending semi-colon is part of the comment, so it's not part of the actual SQL (not a big deal but also doesn't need to be there and can be confusing as to whether it's intentional or not).

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
ike-kottlowski: I'm trying to be as defensive as possible with these changes. I knew that the call sites for this method all validate the salt is unchanged.

``` csharp
// Prevent a de-synced salt value from creating an un-decryptable unlock method
masterPasswordDataModel.MasterPasswordAuthentication.ValidateSaltUnchangedForUser(user);
masterPasswordDataModel.MasterPasswordUnlock.ValidateSaltUnchangedForUser(user);
```

Which calls the `User.GetMasterPasswordSalt` method which returns the `User.Email.LowerInvariant().Trim()`

I'm confident it will always be the Email for the time being, it's the "implicit contract" I was being defensive about.

mzieniukbw: I would just do this change `masterPasswordUnlockData.Salt` now. We already know how this will look like in Dapper, we know the request, data contracts and validations that are in place, so we don't miss this the next time and have less changes to do.
Opinion: I don't think we should make Dapper of EF special in any way, the business logic should be identical between the two.

eliykat:
@ike-kottlowski I have removed the failing migration tests from `main`, if you'd like to update your branch it'll drop me off as a code reviewer. Thanks!

ike-kottlowski:
Addressing max column length comment. Moving to draft while fighting Entity Framework migrations.

src/Core/Entities/User.cs
ike-kottlowski: https://github.com/bitwarden/server/pull/6950/commits/e6a72c0fad71b5811586f3003ae2b0f10e2009e0

src/Core/Entities/User.cs
JaredSnider-Bitwarden: Per sync discussion, we need to evaluate if we should just leave this or add the MasterPasswordSalt and the null coalesce fallback to email or if that will introduce dangerous code paths for user updates outside of operations where we would want the salt to be written to the user table.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
JaredSnider-Bitwarden: I apologize, this was an incorrect comment. I was thinking that this was up a layer... per our sync discussion, let's review if we need any server fallbacks at all. If we do, then we shouldn't put them in the repository layer where they could run from processes not sending up the salt.

src/Core/Entities/User.cs
ike-kottlowski: I've opted to leave this in for the time being. Once we have more consistency around the `MasterPasswordSalt` we can start to reference it more directly.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
ike-kottlowski: We have opted to change the implementation to trust the client to send the correct information when writing to the database. 

The server will not enforce any fallbacks and will cede control to the client. Meaning if the client sends `null` or `"SomeTastySalt"` for the `Salt` value it will be written.

Reading the `Salt` value is in follow up work.

PR #6954, opened 2026-02-05
[PM-31040] Replace `ISetupIntentCache` with customer-based approach

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31040

## 📔 Objective

**Replace the `ISetupIntentCache` distributed cache with a customer-based approach** that directly associates Stripe `SetupIntent` objects with Stripe `Customer` objects. This eliminates infrastructure dependencies on the distributed cache and simplifies the codebase by removing dead code paths.

### Problem

The `SetupIntentDistributedCache` maintained a bidirectional mapping between subscriber IDs and SetupIntent IDs, which was fragile and caused issues with the bank account payment method flow for organizations and providers.

### Solution

Instead of caching the SetupIntent-to-subscriber relationship:
1. Set the `customer` field on the `SetupIntent` when it's retrieved (during subscription creation or payment method update)
2. Query subscribers by `GatewayCustomerId` when webhooks arrive, using new repository methods
3. Remove all cache-related code and dead code paths

### Changes

**Database Infrastructure (Phase 1-3)**
- Add new stored procedures for querying Organizations, Providers, and Users by `GatewayCustomerId` and `GatewaySubscriptionId`
- Add filtered indexes on `GatewayCustomerId` and `GatewaySubscriptionId` columns for all three entity tables
- Add corresponding repository methods (both Dapper and Entity Framework implementations)
- Generate EF migrations for MySQL, PostgreSQL, and SQLite

**SetupIntent Handling Updates (Phase 4-5)**
- Update `SetupIntentSucceededHandler` to query repositories by customer ID instead of using the cache
- Simplify `StripeEventService` by expanding customer data directly from SetupIntent
- Update `GetPaymentMethodQuery` and `HasPaymentMethodQuery` to query Stripe by customer ID
- Update `OrganizationBillingService`, `ProviderBillingService`, and `UpdatePaymentMethodCommand` to set the customer on SetupIntents

**Dead Code Removal (Phase 6-7)**
- Remove bank account support from `CreatePremiumCloudHostedSubscriptionCommand` (premium users cannot use bank accounts)
- Remove `UpdatePaymentMethod` from `OrganizationBillingService`, `ProviderBillingService`, and `PremiumUserBillingService` (replaced by `UpdatePaymentMethodCommand`)
- Remove `UpdatePaymentSource` from `SubscriberService`
- Remove `SignUpPremiumAsync` and `ReplacePaymentMethodAsync` from `UserService` (orphaned at API layer)
- Remove `Finalize` from `PremiumUserBillingService`
- Remove `ISetupIntentCache` and `SetupIntentDistributedCache`

## 📸 Screenshots

https://github.com/user-attachments/assets/5c721f17-04a5-406e-9b40-6885bfb118d5

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
```diff
@@ -558,7 +555,15 @@ public async Task<Customer> SetupCustomer(
...
+            if (!string.IsNullOrEmpty(setupIntentId))
+            {
+                await stripeAdapter.UpdateSetupIntentAsync(setupIntentId,
+                    new SetupIntentUpdateOptions { Customer = customer.Id });
+            }
+
+            return customer;
         }
         catch (StripeException stripeException) when (stripeException.StripeError?.Code == ErrorCodes.TaxIdInvalid)
         {
```
sbrown-livefront, line 569: ❓Are there any Intent specific errors we'd want to log or throw here?

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
```diff
@@ -577,12 +582,10 @@ async Task Revert()
             // ReSharper disable once SwitchStatementMissingSomeEnumCasesNoDefault
```
sbrown-livefront, line 582: ⛏ Do we want to add a default throw or null in this statement and we can remove the Resharper message?

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
```diff
@@ -402,11 +378,6 @@ async Task Revert()
                 // ReSharper disable once SwitchStatementMissingSomeEnumCasesNoDefault
                 switch (customerSetup.TokenizedPaymentSource!.Type)
                 {
```
sbrown-livefront, line 334: ❓Same here for the default case. But also do we not have to cancel the setup intent here like with providers?

src/Core/Billing/Services/Implementations/SubscriberService.cs
```diff
@@ -819,23 +693,7 @@ public async Task<bool> IsValidGatewaySubscriptionIdAsync(ISubscriber subscriber
 
     #region Shared Utilities
 
-    private async Task AddBraintreeCustomerIdAsync(
```
sbrown-livefront, line 822: ❓Making sure I understand. We don't need the brantree customer associated with stripe anymore?

util/MySqlMigrations/Migrations/20260204232320_AddGatewayIndexes.cs
```diff
@@ -0,0 +1,115 @@
...
+/// <inheritdoc />
+public partial class AddGatewayIndexes : Migration
+{
+    /// <inheritdoc />
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        migrationBuilder.AlterColumn<string>(
+            name: "GatewaySubscriptionId",
+            table: "Provider",
+            type: "varchar(255)",
```
mkincaid-bw, line 16: ❓ These columns are `varchar(50)` columns in SQL Server. Should the EF migrations also use `varchar(50)` instead of `varchar(255)`?

util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
```diff
@@ -359,10 +363,10 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("tinyint unsigned");
 
                     b.Property<string>("GatewayCustomerId")
-                        .HasColumnType("longtext");
+                        .HasColumnType("varchar(255)");
```
mkincaid-bw: Same question as above.

util/Migrator/DbScripts/2026-02-18_00_AddGatewayLookupIndexesAndProcs.sql
```diff
@@ -0,0 +1,141 @@
...
+    CREATE NONCLUSTERED INDEX [IX_Provider_GatewaySubscriptionId]
+        ON [dbo].[Provider]([GatewaySubscriptionId])
+        WHERE [GatewaySubscriptionId] IS NOT NULL;
+END
+GO
+
+-- Add indexes for User
+IF NOT EXISTS (SELECT 1 FROM sys.indexes WHERE name = 'IX_User_GatewayCustomerId')
+BEGIN
+    CREATE NONCLUSTERED INDEX [IX_User_GatewayCustomerId]
```
mkincaid-bw, line 38: ⚠️ Just a heads-up: creating a new index on the `User` table can take several minutes (each of these took close to 1.5 minutes on a backup copy of prod).

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
amorask-bitwarden: Sorry where exactly and what regarding? It's not clear from the line placement.

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
amorask-bitwarden: Resharper kinda sucks for C# switches and enums. Even if you set a default, it will still show the original warning: 
image

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
amorask-bitwarden: See above for Resharper, but no we won't have to cancel it. It auto-cancels on its own after a certain time period so I just left the base behavior as-is. If anything, I'd remove the auto-cancellation from the provider-side to simplify it, but didn't feel the need to as part of this.

src/Core/Billing/Services/Implementations/SubscriberService.cs
amorask-bitwarden: We do - but this private method was removed because we were able to completely get rid of the `UpdatePaymentSource` method (which has been replaced by the `UpdatePaymentMethodCommand`). That's where that linking takes place now.

util/MySqlMigrations/Migrations/20260204232320_AddGatewayIndexes.cs
amorask-bitwarden: Yeah would this basically just be setting a MaxLength(50) on both properties (`GatewayCustomerId`, `GatewaySubscriptionId`) for all entities (`User`, `Organization`, `Provider`)?

Also, since those annotations don't exist currently, is there any danger in converting the existing columns to VARCHAR(50)? I'd assume not.

util/Migrator/DbScripts/2026-02-18_00_AddGatewayLookupIndexesAndProcs.sql
amorask-bitwarden: Is there an alternative approach that's better suited to our purposes here in that case?

bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
sbrown-livefront: My bad. For the catch, were there any SetupIntent specific exceptions we would want to look out for?

amorask-bitwarden: None that I can tell by searching through the [Error codes](https://docs.stripe.com/error-codes) documentation since this is just an update. Stripe is pretty bad about linking operations to potential error codes so the ones we do know about are generally from experience haha.

util/MySqlMigrations/Migrations/20260204232320_AddGatewayIndexes.cs
mkincaid-bw: > Yeah would this basically just be setting a MaxLength(50) on both properties (`GatewayCustomerId`, `GatewaySubscriptionId`) for all entities (`User`, `Organization`, `Provider`)?

I'm not 100% sure on the C# side but I believe so.

> 
> Also, since those annotations don't exist currently, is there any danger in converting the existing columns to VARCHAR(50)? I'd assume not.

The `GatewayCustomerId` and `GatewaySubscriptionId` columns in the DB are already `varchar(50)` for the `User` and `Organization` tables. Looks like those columns in the `Provider` table is `longtext`, so I'm guessing EF is trying to make these consistent. I'm not actually sure if/how this will work, so it should be tested to see if it actually changes the column from a `longtext` to a varchar. If so, then `varchar(50)` would be the correct size.

util/Migrator/DbScripts/2026-02-18_00_AddGatewayLookupIndexesAndProcs.sql
mkincaid-bw: Your stored procedures are querying those columns directly, so the indexes are totally valid for this use-case. I just wanted to make sure you were aware, and we should probably ensure BRE is aware that the DB deployment will take longer than normal. If that is an issue, the DB team can manually create the indexes prior to the deployment.

util/MySqlMigrations/Migrations/20260204232320_AddGatewayIndexes.cs
amorask-bitwarden: Addressed in https://github.com/bitwarden/server/pull/6954/changes/549c5134a959995332af3651e33cff9d264653dc and https://github.com/bitwarden/server/pull/6954/changes/69f804edbcce8e0e971fbbbe735a059a95942a23.

BTreston (dismissed):
approved for AC owned files

mkincaid-bw (approved):

BTreston (approved):

PR #6971, opened 2026-02-10
[PM-30993] Better Error Message for Expired Invitation Tokens

sven-bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30993

## 📔 Objective
The server currently sends a nondescript `Invalid token.` error message when accepting an invalid token. While technically true, we want to be kinder to users accepting expired tokens, giving them a better error message.

### Other Considerations
I thought about enhancing the Tokenable API to be return a `TokenableValidationResult`, which could mirror the existing inheritance hierarchy to return objects that indicate what went wrong. For example, Tokenable returns TokenableValidationResult, ExpiringTokenable returns ExpiringTokenableValidationResult.. each tacking on new metadata.

I opted not to do this, as this is so far the only Tokenable area that is interested in knowing what validation failed, and I didn't want to force the creation of N result classes (one for each Tokenable derivative).

## 📸 Screenshots
Screenshot 2026-02-10 at 7 51 20 AM

### DIFF
--- a/src/Api/Vault/Models/Response/SyncResponseModel.cs
+++ b/src/Api/Vault/Models/Response/SyncResponseModel.cs
@@ -84,7 +84,7 @@ public class SyncResponseModel() : ResponseModel("sync")
                         Parallelism = user.KdfParallelism
                     },
                     MasterKeyEncryptedUserKey = user.Key!,
-                    Salt = user.Email.ToLowerInvariant()
+                    Salt = user.GetMasterPasswordSalt()
                 }
                 : null,
             WebAuthnPrfOptions = webAuthnPrfOptions.Length > 0 ? webAuthnPrfOptions : null,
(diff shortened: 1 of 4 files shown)


JaredSnider-Bitwarden (dismissed):
Nice work!

JaredScar (dismissed):

sven-bitwarden:
@JaredScar @JaredSnider-Bitwarden sorry y'all, my prior PR #6972 had a conflict here 🙏

sven-bitwarden:
@JaredScar @JaredSnider-Bitwarden 🙏  Through QA (which is difficult to QA expired tokens) we found an additional area that needs to benefit from the expired token message. This lead me to consolidate the error message creation logic.

@JaredSnider-Bitwarden I wasn't sure of your preference - we could have all of the methods inside of the static class point to the new method and check if it's null or not to determine the presence of an error. I left it as-is otherwise

src/Core/Auth/UserFeatures/Registration/Implementations/RegisterUserCommand.cs
JaredSnider-Bitwarden, line 1: @sven-bitwarden , could we update the `RegisterUserCommandTests` to include an expired token scenario?

src/Core/Auth/Models/Business/Tokenables/OrgUserInviteTokenable.cs
```diff
@@ -63,6 +63,22 @@ public bool TokenIsValid(Guid orgUserId, string orgUserEmail)
     protected override bool TokenIsValid() =>
         Identifier == TokenIdentifier && OrgUserId != default && !string.IsNullOrWhiteSpace(OrgUserEmail);
 
+    public static string? GetOrgUserInviteValidationError(
```
JaredSnider-Bitwarden: 🎨 : this is a naming nit, but having this be a get validation error method implies the validation has already been done and it is known that the org invite is invalid when actually this method is responsible for validating the org user invite and returning a specific error. Do you think we could change the verbiage to be something like `ValidateOrgUserInvite`?  The shared consolidation is great, though.

JaredSnider-Bitwarden (approved):
Excellent work! Please be sure that the standard org invite and org invite registration flows are re-tested by QA. The tests give great confidence that those will be successful though!

PR #6972, opened 2026-02-10
[PM-4142] Clean Up Pre-Tokenable Token Validation

sven-bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-4142

## 📔 Objective
This ticket was slated to be done a few releases after Tokenables were introduced, to clean up backwards compatible code left to support both old and new token validation. Given that is has been _at least_ a few releases since now and then, we're free to clean up this code - all tokens are provisioned with the newer APIs.


JaredSnider-Bitwarden (approved):
LGTM! Thank you for cleaning this up!
For the QA notes, please ensure that you request testing of the standard accept org invite flow (+ accepting an org invite into an org where MP policy requirements are in place) and registration via org invite flow.

PR #6975, opened 2026-02-10
Ensure that the maria db password is properly set

theMickster:
## 📔 Objective

While tracking down another local problem, I worked to setup all the other database providers and the missing variable caused problems. Claude and I fixed our woes with this change.

### DIFF
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -77,6 +77,7 @@ services:
       - 4306:3306
     environment:
       MARIADB_USER: maria
+      MARIADB_PASSWORD: ${MARIADB_ROOT_PASSWORD}
       MARIADB_DATABASE: vault_dev
       MARIADB_RANDOM_ROOT_PASSWORD: "true"
     volumes:


dev/docker-compose.yml
```diff
@@ -77,6 +77,7 @@ services:
       - 4306:3306
     environment:
       MARIADB_USER: maria
+      MARIADB_PASSWORD: ${MARIADB_ROOT_PASSWORD}
       MARIADB_DATABASE: vault_dev
       MARIADB_RANDOM_ROOT_PASSWORD: "true"
```
withinfocus, line 82: ❓ What's the difference with these two?

vgrassia (approved):

dev/docker-compose.yml
theMickster: `MARIADB_PASSWORD` (the Docker image env var) is required alongside `MARIADB_USER` — without it, the maria user is never created and the connection string fails silently. That's the actual fix.

`${MARIADB_ROOT_PASSWORD}` is just the .env variable being used as the password value. The name is misleading — it's not actually the root password here since MARIADB_RANDOM_ROOT_PASSWORD: "true" overrides that. It's just the only MariaDB-related password variable that already existed in .env, so it got reused. Poor name, but changing it would touch every developer's .env file.

PR #6982, opened 2026-02-10
[PM-30108] import discount from stripe

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30108

## 📔 Objective

Bitwarden portal interface to import a coupon from stripe, set additional properties, and save the data in the database. Index view includes a paginated table to view all discounts.

## 📸 Screenshots

image

image

image

### DIFF
--- a/src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
+++ b/src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
@@ -166,6 +166,15 @@ public class SubscriptionDiscountsController(
 
         try
         {
+            // Check for duplicate coupon to prevent race condition
+            var existing = await subscriptionDiscountRepository.GetByStripeCouponIdAsync(model.StripeCouponId);
+            if (existing != null)
+            {
+                ModelState.AddModelError(nameof(model.StripeCouponId),
+                    "This coupon has already been imported.");
+                return View(model);
+            }
+
             var discount = new SubscriptionDiscount
             {
                 StripeCouponId = model.StripeCouponId,
--- a/test/Admin.Test/Billing/Controllers/SubscriptionDiscountsControllerTests.cs
+++ b/test/Admin.Test/Billing/Controllers/SubscriptionDiscountsControllerTests.cs
@@ -319,6 +319,10 @@ public class SubscriptionDiscountsControllerTests
             IsImported = true
         };
 
+        sutProvider.GetDependency<ISubscriptionDiscountRepository>()
+            .GetByStripeCouponIdAsync(model.StripeCouponId)
+            .Returns((SubscriptionDiscount?)null);
+
         sutProvider.Sut.ModelState.Clear();
         var tempData = new TempDataDictionary(new DefaultHttpContext(), Substitute.For<ITempDataProvider>());
         sutProvider.Sut.TempData = tempData;
@@ -355,6 +359,10 @@ public class SubscriptionDiscountsControllerTests
             IsImported = true
         };
 
+        sutProvider.GetDependency<ISubscriptionDiscountRepository>()
+            .GetByStripeCouponIdAsync(model.StripeCouponId)
+            .Returns((SubscriptionDiscount?)null);
+
         sutProvider.Sut.ModelState.Clear();
         var tempData = new TempDataDictionary(new DefaultHttpContext(), Substitute.For<ITempDataProvider>());
         sutProvider.Sut.TempData = tempData;
@@ -382,6 +390,41 @@ public class SubscriptionDiscountsControllerTests
         Assert.IsType<ViewResult>(result);
         Assert.False(sutProvider.Sut.ModelState.IsValid);
         Assert.Contains("import the coupon", sutProvider.Sut.ModelState[string.Empty]!.Errors[0].ErrorMessage);
+    }
+
+    [Theory, BitAutoData]
+    public async Task Create_DuplicateCoupon_ReturnsViewWithError(
+        SubscriptionDiscount existingDiscount,
+        SutProvider<SubscriptionDiscountsController> sutProvider)
+    {
+        var model = new CreateSubscriptionDiscountModel
+        {
+            StripeCouponId = "TEST123",
+            Name = "Test Coupon",
+            PercentOff = 25,
+            Duration = "once",
+            StartDate = DateTime.UtcNow.Date,
+            EndDate = DateTime.UtcNow.Date.AddMonths(1),
+            IsImported = true
+        };
+
+        // Simulate race condition: another admin imported the same coupon between import and save
+        sutProvider.GetDependency<ISubscriptionDiscountRepository>()
+            .GetByStripeCouponIdAsync(model.StripeCouponId)
+            .Returns(existingDiscount);
+
+        sutProvider.Sut.ModelState.Clear();
+
+        var result = await sutProvider.Sut.Create(model);
+
+        Assert.IsType<ViewResult>(result);
+        Assert.False(sutProvider.Sut.ModelState.IsValid);
+        Assert.Contains("already been imported", sutProvider.Sut.ModelState[nameof(model.StripeCouponId)]!.Errors[0].ErrorMessage);
+
+        // Verify CreateAsync was NOT called since we detected the duplicate
+        await sutProvider.GetDependency<ISubscriptionDiscountRepository>()
+            .DidNotReceive()
+            .CreateAsync(Arg.Any<SubscriptionDiscount>());
     }
 
     [Theory, BitAutoData]


src/Core/Billing/Enums/DiscountAudienceType.cs
```diff
@@ -6,8 +6,14 @@
...
+    /// Discount applies to all users regardless of subscription history.
+    /// This is the default value (0) when audience restrictions are not applied.
+    /// </summary>
+    AllUsers = 0,
+
     /// <summary>
     /// Discount applies to users who have never had a subscription before.
     /// </summary>
-    UserHasNoPreviousSubscriptions = 0
+    UserHasNoPreviousSubscriptions = 1
```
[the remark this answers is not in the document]
kdenney: Good suggestion but this table is unused before this branch. It was created in a different branch but nothing used it.

src/Admin/Billing/Views/SubscriptionDiscounts/Index.cshtml
```diff
@@ -0,0 +1,127 @@
...
+                <tbody>
+                    @foreach (var discount in Model.Items)
+                    {
+                        var dateRange = $"{discount.StartDate.ToString("MM/dd/yyyy")} - {discount.EndDate.ToString("MM/dd/yyyy")}";
+                        <tr>
+                            <td><code>@discount.StripeCouponId</code></td>
+                            <td>
+                                <a asp-action="Edit" asp-route-id="@discount.Id">
+                                    @discount.Name
+                                </a>
```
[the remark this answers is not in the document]
kdenney: This is on purpose. The edit action is being added in another ticket immediately after this one.

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            Coupon coupon;
+            try
+            {
+                var options = new CouponGetOptions();
+                options.AddExpand(StripeConstants.CouponExpandablePropertyNames.AppliesTo);
+                coupon = await stripeAdapter.GetCouponAsync(model.StripeCouponId, options);
+            }
+            catch (StripeException ex)
```
amorask-bitwarden, line 100: ⛏️ This will catch all potential Stripe errors and report them as "Stripe coupon not found". If you want, you can resolve this by checking for error code "resource_missing", but I don't think it's really necessary unless you want to considering it's a GET operation. If not feel free to just resolve the comment.

src/Admin/Billing/Models/CreateSubscriptionDiscountModel.cs
```diff
@@ -0,0 +1,47 @@
...
+    public DateTime EndDate { get; set; } = DateTime.UtcNow.Date.AddMonths(1);
+
+    [Display(Name = "Restrict to users with no previous subscriptions?")]
+    public bool RestrictToNewUsersOnly { get; set; }
+
+    public DiscountAudienceType AudienceType => RestrictToNewUsersOnly
+        ? DiscountAudienceType.UserHasNoPreviousSubscriptions
+        : DiscountAudienceType.AllUsers;
+
+    public bool IsImported => !string.IsNullOrEmpty(Name);
```
amorask-bitwarden: ❓ What do you think about dedicating a settable property to `IsImported`? The reason I bring this up is because, technically, Stripe's `coupon.name` field [is nullable](https://docs.stripe.com/api/coupons/object?api-version=2025-08-27.basil&lang=dotnet#coupon_object-name). If the person who makes the coupon forgets to set a name for whatever reason, the import process will stall out.

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -0,0 +1,196 @@
...
+                    "Coupon not found in Stripe. Please verify the coupon ID.");
+                return View("Create", model);
+            }
+
+            model.Name = coupon.Name;
+            model.PercentOff = coupon.PercentOff;
+            model.AmountOff = coupon.AmountOff;
+            model.Currency = coupon.Currency;
+            model.Duration = coupon.Duration;
+            model.DurationInMonths = coupon.DurationInMonths.HasValue ? (int?)coupon.DurationInMonths.Value : null;
```
amorask-bitwarden: ⛏️ I think this can be simplified to `(int?)coupon.DurationInMonths`

src/Core/Billing/Subscriptions/Repositories/ISubscriptionDiscountRepository.cs
```diff
@@ -20,4 +20,13 @@ public interface ISubscriptionDiscountRepository : IRepository<SubscriptionDisco
...
     Task<SubscriptionDiscount?> GetByStripeCouponIdAsync(string stripeCouponId);
+
+    /// <summary>
+    /// Searches subscription discounts with pagination support.
+    /// Returns discounts ordered by creation date descending (newest first).
+    /// </summary>
+    /// <param name="skip">Number of records to skip (for pagination).</param>
+    /// <param name="take">Number of records to take (page size).</param>
+    /// <returns>A collection of subscription discounts for the requested page.</returns>
+    Task<ICollection<SubscriptionDiscount>> SearchAsync(int skip, int take);
```
amorask-bitwarden: ❓ Should this actually be called `SearchAsync` when we're not filtering by a search term or other parameter? I think it's more like a `List`.

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -0,0 +1,196 @@
...
+                    {
+                        Ids = productIds.ToList()
+                    });
+
+                    model.AppliesToProducts = allProducts
+                        .ToDictionary(product => product.Id, product => product.Name);
+                }
+                catch (StripeException ex)
+                {
+                    logger.LogError(ex, "Failed to fetch the coupon's associated products from Stripe");
```
amorask-bitwarden: ⛏️ I'd suggest getting the `coupon.id` into this log message here (and anywhere else possible). It will make log searching significantly easier.

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -0,0 +1,196 @@
...
+        try
+        {
+            var discount = new SubscriptionDiscount
+            {
+                StripeCouponId = model.StripeCouponId,
+                Name = model.Name,
+                PercentOff = model.PercentOff,
+                AmountOff = model.AmountOff,
+                Currency = model.Currency,
+                Duration = model.Duration!,
```
amorask-bitwarden: ⛏️ I think the `!` operator here is a giveaway that the model's `Duration` property doesn't need to be nullable in the first place, right?

src/Core/Billing/Enums/DiscountAudienceType.cs
amorask-bitwarden, line 5: ❓ Is `IDiscountAudienceFilter` going to be created in a later ticket?

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -0,0 +1,196 @@
...
+            Coupon coupon;
+            try
+            {
+                var options = new CouponGetOptions();
+                options.AddExpand(StripeConstants.CouponExpandablePropertyNames.AppliesTo);
+                coupon = await stripeAdapter.GetCouponAsync(model.StripeCouponId, options);
+            }
+            catch (StripeException ex)
+            {
+                logger.LogError(ex, "Stripe coupon not found: {CouponId}", model.StripeCouponId);
```
cyprain-okeke: I am thinking that logging the coupon IDs might be sensitive, maybe we should log the coupon name instead? What do you think

src/Admin/Billing/Models/CreateSubscriptionDiscountModel.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        if (EndDate < StartDate)
+        {
+            yield return new ValidationResult(
+                "End Date must be on or after Start Date.",
+                new[] { nameof(EndDate) });
+        }
```
cyprain-okeke, line 45: Should we have this added to the validate
i```
f (!PercentOff.HasValue && !AmountOff.HasValue)
      {
          yield return new ValidationResult(
              "Discount must have either a percentage off or amount off.",
              new[] { nameof(PercentOff), nameof(AmountOff) });
      }
```

**Review by cyprain-okeke — CHANGES_REQUESTED** [review]

File: src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs (hunk 9 above)
Comment (kdenney) [reply src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs]: To my knowledge, the ID can't be used in any way to exploit the coupon or discount system so I don't think it is sensitive but please correct me if I am wrong! 😄

File: src/Admin/Billing/Models/CreateSubscriptionDiscountModel.cs (hunk 10 above)
Comment (kdenney) [reply src/Admin/Billing/Models/CreateSubscriptionDiscountModel.cs:45]: Stripe coupons cannot be created without one of these properties and the validation for these already exist in the stripe dashboard when creating a coupon. But I appreciate the attention to detail! 🙌

File: src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs (hunk 3 above)
Comment (kdenney) [reply src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs:100]: Good call! Added in https://github.com/bitwarden/server/pull/6982/commits/527ffcfd14d678d637e53ffc920fc677393ea6d3

File: src/Admin/Billing/Models/CreateSubscriptionDiscountModel.cs (hunk 4 above)
Comment (kdenney) [reply src/Admin/Billing/Models/CreateSubscriptionDiscountModel.cs]: I didn't catch that the name was optional haha! Added your suggestion in https://github.com/bitwarden/server/pull/6982/commits/527ffcfd14d678d637e53ffc920fc677393ea6d3. Thanks!

File: src/Core/Billing/Subscriptions/Repositories/ISubscriptionDiscountRepository.cs (hunk 6 above)
Comment (kdenney) [reply src/Core/Billing/Subscriptions/Repositories/ISubscriptionDiscountRepository.cs]: Renamed in https://github.com/bitwarden/server/pull/6982/commits/527ffcfd14d678d637e53ffc920fc677393ea6d3

File: src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs (hunk 8 above)
Comment (kdenney) [reply src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs]: Yeah, good catch. Updated in https://github.com/bitwarden/server/pull/6982/commits/527ffcfd14d678d637e53ffc920fc677393ea6d3

File: src/Core/Billing/Enums/DiscountAudienceType.cs
Comment (kdenney) [reply src/Core/Billing/Enums/DiscountAudienceType.cs:5]: Yes, that should be added in the "fetch eligible discounts" ticket

File: src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs (hunk 9 above)
Comment (amorask-bitwarden) [reply src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs]: I'd agree a Stripe `Coupon` identifier would not be considered sensitive information in the context of logging. We frequently log `Subscription` and `Customer` IDs for ease in debugging.

File: src/Admin/Billing/Models/SubscriptionDiscountViewModel.cs [hunk 11]
```diff
@@ -0,0 +1,27 @@
...
+    public int? DurationInMonths { get; set; }
+    public DateTime StartDate { get; set; }
+    public DateTime EndDate { get; set; }
+    public DiscountAudienceType AudienceType { get; set; }
+    public DateTime CreationDate { get; set; }
+    public bool IsActive => DateTime.UtcNow >= StartDate && DateTime.UtcNow <= EndDate;
+
+    public string DiscountDisplay => PercentOff.HasValue
+        ? $"{PercentOff.Value:G29}% off"
+        : $"${AmountOff / 100m} off";
```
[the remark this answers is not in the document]
kdenney: We only operate in USD right now.

PR #6984, opened 2026-02-10
Add query for email verification link

MGibson1:
## 📔 Objective

Query to get user email verification link for registration tests.

### DIFF
--- a/util/Seeder/Queries/UserEmailVerificationQuery.cs
+++ b/util/Seeder/Queries/UserEmailVerificationQuery.cs
@@ -1,14 +1,20 @@
 ﻿using System.Globalization;
+using System.Net;
+using Bit.Core.Auth.Models.Business.Tokenables;
 using Bit.Core.Repositories;
+using Bit.Core.Tokens;
 
 namespace Bit.Seeder.Queries;
 
-public class UserEmailVerificationQuery(IUserRepository userRepository) : IQuery<UserEmailVerificationQuery.Request, UserEmailVerificationQuery.Response>
+public class UserEmailVerificationQuery(IUserRepository userRepository,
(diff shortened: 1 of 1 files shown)


util/Seeder/Queries/UserEmailVerificationQuery.cs
```diff
@@ -0,0 +1,38 @@
...
+        public required bool EmailVerified { get; set; }
+    }
+
+    public async Task<Response> Execute(Request request)
+    {
+        var user = await userRepository.GetByEmailAsync(request.Email);
+
+        return new()
+        {
+            Url = Url(string.Empty, request.Email, request.FromMarketing),
```
theMickster: ❓ The passing of the token as always string.Empty is a little confusing and smells like a bug because I am unsure without additional context (in either code comments or the PR description because there's no Jira ticket) just how the URL will be functional because I am reading this as the URL for `pete@bw.com` would look like `/redirect-connector.html#finish-signup?token=&email=pete@bw.com&fromEmail=true`. Wouldn't that then make the verification link non-functional?

util/Seeder/IQuery.cs
theMickster, line 12: ⚠️ Please update the XML comments as well to show the migration to an async pattern.

util/Seeder/Queries/UserEmailVerificationQuery.cs
```diff
@@ -0,0 +1,38 @@
...
+            EmailVerified = user?.EmailVerified ?? false
+        };
+    }
+
+    public string Url(string token, string email, string? fromMarketing = null)
```
theMickster: ⚠️ The method is listed as public but it acts like a private method.

util/Seeder/Queries/UserEmailVerificationQuery.cs
MGibson1: yikes, look like I forgot to fix that stub. Done

PR #6993, opened 2026-02-12
[PM-21179] Add interface to check if user is enrolled in account recovery

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21179

## 📔 Objective

Centralizes the logic for determining whether an organization user is enrolled in account recovery by introducing two methods on `OrganizationUser`:
- Static `IsValidResetPasswordKey(string?)`
- Instance `IsEnrolledInAccountRecovery()`

This removes scattered `string.IsNullOrEmpty` / `string.IsNullOrWhiteSpace` checks across the codebase and replaces them with a single consistent implementation.
It also fixes an inconsistency where some call sites treated whitespace-only keys as valid (`IsNullOrEmpty`) while others did not (`IsNullOrWhiteSpace`).


src/Api/KeyManagement/Validators/OrganizationUserRotationValidator.cs
```diff
@@ -44,7 +44,7 @@ public async Task<IReadOnlyList<OrganizationUser>> ValidateAsync(User user,
                 throw new BadRequestException("All existing reset password keys must be included in the rotation.");
             }
 
-            if (organizationUser.ResetPasswordKey == null)
+            if (!OrganizationUser.IsValidResetPasswordKey(organizationUser.ResetPasswordKey))
```
quexten: I'm double checking but this may re-introduce a bug where users won't be able to rotate their keys. Clients send `""` as "valid" values currently, which is not yet fixed on the client side. Because of that, this change in behavior would break them from rotating keys.

https://github.com/bitwarden/server/pull/6871

quexten: https://bitwarden.atlassian.net/browse/PM-31001

Here is the KM task tracking this behavior on clients.

quexten: I recommend leaving the old behavior, but putting a comment such as:
```
// Should be migrated to: if (!OrganizationUser.IsValidResetPasswordKey(organizationUser.ResetPasswordKey))
// after https://bitwarden.atlassian.net/browse/PM-31001 is resolved
```

r-tome: Nicely spotted! I've added that comment

jrmccannon (approved):

PR #6994, opened 2026-02-12
Add coupon support to invoice preview and subscription creation

cyprain-okeke:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30272
https://bitwarden.atlassian.net/browse/PM-32756

## 📔 Objective

This PR adds coupon/discount code support to the subscription purchase flow, enabling users to apply promotional discounts during both
  tax preview and subscription creation for Premium and Organization subscriptions.

  ---
  **What's Changed**

  **1. Tax Preview with Coupon Support**

  Added optional coupon parameter to tax preview endpoints, allowing users to see discounted pricing before purchase:

  - Organization Tax Preview: POST /api/billing/preview-invoice/organization-subscription-purchase-tax
  - Premium Tax Preview: POST /api/billing/preview-invoice/premium-subscription-purchase-tax

  Both endpoints now accept an optional coupon field in the request body and return tax calculations with the discount applied.

  **2. Subscription Creation with Coupon Support**

  Updated subscription creation endpoints to accept and validate coupon codes:

  - Premium Subscription: POST /api/billing/vnext/accounts/premium-cloud-hosted-subscription
  - Organization Subscription: Handled via OrganizationBillingService.Finalize()

  Coupons are validated server-side before being applied to ensure user eligibility.

  **3. Server-Side Validation**

  Implemented ISubscriptionDiscountService to validate coupon eligibility:
  - Verifies coupon exists in database
  - Checks if coupon is within valid date range
  - Validates user meets audience targeting criteria (e.g., no previous subscriptions)
  - Returns appropriate error messages for invalid/ineligible coupons

  **4. Special Case Handling**

  - Sponsored Plans: User-provided coupons are ignored (organization gets free plan)
  - Standalone Secrets Manager: System coupon takes precedence over user coupons
  - Invalid Coupons: Returns 400 Bad Request with message: "The coupon code is invalid or you are not eligible for this discount."

  ---
###   Technical Implementation

  **API Layer:**
  - Added Coupon property to request models with [MaxLength(50)] validation
  - Updated controllers to extract and pass coupon parameter to commands
  - Input sanitization: whitespace is trimmed before processing

  **Business Logic:**
  - PreviewOrganizationTaxCommand: Applies coupon to Stripe invoice preview
  - PreviewPremiumTaxCommand: Applies coupon to Stripe invoice preview
  - CreatePremiumCloudHostedSubscriptionCommand: Validates and applies coupon to subscription
  - OrganizationBillingService: Validates and applies coupon to organization subscription

Add bug fix for : https://bitwarden.atlassian.net/browse/PM-32756

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,4 +1,8 @@
 {
+  "attribution": {
+    "commit": "",
+    "pr": ""
+  },
   "extraKnownMarketplaces": {
     "bitwarden-marketplace": {
       "source": {


src/Core/Billing/Services/Implementations/SubscriptionDiscountService.cs
```diff
@@ -0,0 +1,34 @@
+﻿#nullable enable
```
amorask-bitwarden: ❌ Not needed

src/Core/Billing/Services/Implementations/SubscriptionDiscountService.cs
```diff
@@ -0,0 +1,34 @@
...
+        var now = DateTime.UtcNow;
+        if (now < discount.StartDate || now > discount.EndDate)
+        {
+            return false;
+        }
+
+        return discount.AudienceType switch
+        {
+            DiscountAudienceType.UserHasNoPreviousSubscriptions =>
+                !user.Premium && string.IsNullOrEmpty(user.GatewaySubscriptionId),
```
amorask-bitwarden: ❌ This checks if the user **has** a premium subscription; not if they've ever **had** a premium subscription. I think we need to check this and also pull their customer to check for any inactive premium subscriptions from the past.

src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
```diff
@@ -22,7 +22,8 @@ public interface IPreviewOrganizationTaxCommand
 {
     Task<BillingCommandResult<(decimal Tax, decimal Total)>> Run(
         OrganizationSubscriptionPurchase purchase,
-        BillingAddress billingAddress);
+        BillingAddress billingAddress,
+        string? coupon);
```
amorask-bitwarden: ❌  This should be part of the `OrganizationSubscriptionPurchase`.

src/Api/Billing/Controllers/VNext/AccountBillingVNextController.cs
```diff
@@ -80,9 +80,9 @@ public async Task<IResult> CreateSubscriptionAsync(
         [BindNever] User user,
         [FromBody] PremiumCloudHostedSubscriptionRequest request)
     {
-        var (paymentMethod, billingAddress, additionalStorageGb) = request.ToDomain();
+        var (paymentMethod, billingAddress, additionalStorageGb, coupon) = request.ToDomain();
```
amorask-bitwarden: ⛏️ At 4 values, I think this is becoming a bit too unwieldy. Can we create a new `PremiumSubscriptionPurchase` in src/Core/Billing/Premium/Models, similar to the `OrganizationSubscriptionPurchase` and pass that into the command instead?

src/Api/Billing/Controllers/PreviewInvoiceController.cs
```diff
@@ -21,8 +21,8 @@ public class PreviewInvoiceController(
     public async Task<IResult> PreviewOrganizationSubscriptionPurchaseTaxAsync(
         [FromBody] PreviewOrganizationSubscriptionPurchaseTaxRequest request)
     {
-        var (purchase, billingAddress) = request.ToDomain();
-        var result = await previewOrganizationTaxCommand.Run(purchase, billingAddress);
+        var (purchase, billingAddress, coupon) = request.ToDomain();
```
amorask-bitwarden: ❌ `coupon` should be part of the `purchase`.

src/Api/Billing/Controllers/PreviewInvoiceController.cs
```diff
@@ -52,8 +52,8 @@ public async Task<IResult> PreviewOrganizationSubscriptionUpdateTaxAsync(
     public async Task<IResult> PreviewPremiumSubscriptionPurchaseTaxAsync(
         [FromBody] PreviewPremiumSubscriptionPurchaseTaxRequest request)
     {
-        var (purchase, billingAddress) = request.ToDomain();
-        var result = await previewPremiumTaxCommand.Run(purchase, billingAddress);
+        var (purchase, billingAddress, coupon) = request.ToDomain();
```
amorask-bitwarden: ❌ Same here.

src/Api/Billing/Models/Requests/Premium/PremiumCloudHostedSubscriptionRequest.cs
```diff
@@ -15,8 +15,9 @@ public class PremiumCloudHostedSubscriptionRequest : IValidatableObject
     [Range(0, 99)]
     public short AdditionalStorageGb { get; set; } = 0;
 
+    public string? Coupon { get; set; }
 
-    public (PaymentMethod, BillingAddress, short) ToDomain()
+    public (PaymentMethod, BillingAddress, short, string?) ToDomain()
```
amorask-bitwarden: ▶️ See [this comment](https://github.com/bitwarden/server/pull/6994/changes#r2801026510).

src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
```diff
@@ -120,6 +124,11 @@ public class PreviewOrganizationTaxCommand(
                         }
                     }
 
+                    if (!string.IsNullOrWhiteSpace(coupon))
```
amorask-bitwarden: ❌ This doesn't validate coupon eligibility as part of the tax estimation process, which can easily result in disjointed data on the client.

If we validate the coupon when determining if it will affect the total the user will pay, but don't validate it when determining the tax, the tax could be discounted while the total could not be. And if we need to validate, it needs to be restricted to Families.

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
```diff
@@ -34,11 +34,21 @@ public class OrganizationBillingService(
...
     ISubscriberService subscriberService,
+    ISubscriptionDiscountService subscriptionDiscountService,
     ITaxService taxService) : IOrganizationBillingService
 {
     public async Task Finalize(OrganizationSale sale)
     {
-        var (organization, customerSetup, subscriptionSetup) = sale;
+        var (organization, customerSetup, subscriptionSetup, owner) = sale;
+
+        if (!string.IsNullOrWhiteSpace(customerSetup?.Coupon) && owner != null)
```
amorask-bitwarden, line 45: ❌ If the `owner` is set to null in the `OrganizationSignup` model that constructed the sale, the validation you've added will be bypassed and the coupon will be applied regardless.

Additionally:
1. This change would break SM Standalone coupon application on release, since that coupon isn't represented as a `SubscriptionDiscount` nor does it have an audience.
2. This change applies to all organization sales regardless of tier even though the ticket specifies these coupons are only for users purchasing Premium or Families.

src/Api/Billing/Models/Requests/Premium/PremiumCloudHostedSubscriptionRequest.cs
```diff
@@ -15,8 +15,9 @@ public class PremiumCloudHostedSubscriptionRequest : IValidatableObject
     [Range(0, 99)]
     public short AdditionalStorageGb { get; set; } = 0;
 
+    public string? Coupon { get; set; }
```
amorask-bitwarden, line 20: ⛏️ The DB only takes 50 characters so we should `[MaxLength(50)]` this.

src/Api/Billing/Models/Requests/PreviewInvoice/PreviewOrganizationSubscriptionPurchaseTaxRequest.cs
```diff
@@ -14,6 +14,8 @@ public record PreviewOrganizationSubscriptionPurchaseTaxRequest
     [Required]
     public required CheckoutBillingAddressRequest BillingAddress { get; set; }
 
-    public (OrganizationSubscriptionPurchase, BillingAddress) ToDomain() =>
-        (Purchase.ToDomain(), BillingAddress.ToDomain());
+    public string? Coupon { get; set; }
```
amorask-bitwarden: ⛏️ The DB only takes 50 characters so we should `[MaxLength(50)]` this.

src/Api/Billing/Models/Requests/PreviewInvoice/PreviewPremiumSubscriptionPurchaseTaxRequest.cs
```diff
@@ -13,5 +13,7 @@ public record PreviewPremiumSubscriptionPurchaseTaxRequest
     [Required]
     public required MinimalBillingAddressRequest BillingAddress { get; set; }
 
-    public (short, BillingAddress) ToDomain() => (AdditionalStorage, BillingAddress.ToDomain());
+    public string? Coupon { get; set; }
```
amorask-bitwarden, line 18: ⛏️ The DB only takes 50 characters so we should `[MaxLength(50)]` this.

src/Core/Billing/Services/ISubscriptionDiscountService.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </returns>
+    /// <remarks>
+    /// This method performs server-side validation to ensure:
+    /// <list type="bullet">
+    /// <item>The discount exists in the database</item>
+    /// <item>The discount is within its valid date range</item>
+    /// <item>The user meets the audience targeting criteria for the discount</item>
+    /// </list>
+    /// </remarks>
+    Task<bool> ValidateDiscountForUserAsync(User user, string stripeCouponId);
```
amorask-bitwarden: ❌ The signature and description of this method are too vague in asserting what it actually does. It doesn't just validate a coupon ID for a user; it specifically checks to see if that coupon ID matches a specific discount audience. That needs to be explicit here. Your options are:
1. Rename the method to emphasize the specific audience the discount is being validated against
2. Take in an audience as a parameter to validate against

amorask-bitwarden (changes requested):
Good start - few changes we need to clean this up.

For testing, can you:
1. Add tests for `SubscriptionDiscountService`
2. Update the tests for `CreatePremiumCloudHostedSubscriptionCommand` to validate coupon functionality instead of passing `null` to all
3. Update the tests for `PreviewPremiumTaxCommand` to validate coupon functionality instead of passing `null` to all

If you want, optionally add more tests for `OrganizationBillingService.Finalize` to cover your cases.

src/Core/Billing/Premium/Models/PremiumSubscriptionPurchase.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Bit.Core.Billing.Payment.Models;
+
+namespace Bit.Core.Billing.Premium.Models;
+
+public record PremiumSubscriptionPurchase
+{
+    public required PaymentMethod PaymentMethod { get; init; }
+    public required BillingAddress BillingAddress { get; init; }
+    public short AdditionalStorageGb { get; init; }
```
amorask-bitwarden: ❓ Should this be nullable?

src/Core/Billing/Services/Implementations/SubscriptionDiscountService.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (HasPremiumPrice(subscription))
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    private static bool HasPremiumPrice(Subscription subscription)
```
amorask-bitwarden: ❌ Couple things here.
1. We should not be hard-coding price IDs in our codebase. I suggest using the `IPricingClient` to get the legacy and active premium price IDs.
2. `Subscription.Items` and `Subscription.Items.Data` are not nullable in Stripe's API so the null-coalescing operators are unnecessary and add confusion.
3. `premium-monthly` has never been a real price ID as we don't offer a monthly version of premium. Please review any Claude-generated code in greater detail before opening up reviews.

src/Core/Billing/Premium/Models/PremiumSubscriptionPurchase.cs
amorask-bitwarden: ❌ This will never be sent with the new approach to premium purchasing, but so we don't have to go through and remove it everywhere, let's make it nullable.

src/Core/Billing/Premium/Models/PremiumPurchasePreview.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Bit.Core.Billing.Premium.Models;
+
+public record PremiumPurchasePreview
+{
+    public short AdditionalStorageGb { get; init; }
```
amorask-bitwarden: ❌ This will never be sent with the new approach to premium purchasing, but so we don't have to go through and remove it everywhere, let's make it nullable.

src/Core/Billing/Premium/Commands/PreviewPremiumTaxCommand.cs
```diff
@@ -47,15 +48,20 @@ public class PreviewPremiumTaxCommand(
...
                 {
                     Price = premiumPlan.Storage.StripePriceId,
-                    Quantity = additionalStorage
+                    Quantity = preview.AdditionalStorageGb
                 });
             }
 
+            if (!string.IsNullOrWhiteSpace(preview.Coupon))
+            {
+                options.Discounts = [new InvoiceDiscountOptions { Coupon = preview.Coupon.Trim() }];
```
amorask-bitwarden: ❌ See comment: https://github.com/bitwarden/server/pull/6994#discussion_r2801042331

src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
```diff
@@ -120,6 +122,11 @@ public class PreviewOrganizationTaxCommand(
                         }
                     }
 
+                    if (!string.IsNullOrWhiteSpace(purchase.Coupon))
+                    {
+                        options.Discounts = [new InvoiceDiscountOptions { Coupon = purchase.Coupon.Trim() }];
```
amorask-bitwarden: ❌ This went unaddressed: https://github.com/bitwarden/server/pull/6994#discussion_r2801042331

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
```diff
@@ -38,7 +38,7 @@ public class OrganizationBillingService(
 {
     public async Task Finalize(OrganizationSale sale)
     {
-        var (organization, customerSetup, subscriptionSetup) = sale;
+        var (organization, customerSetup, subscriptionSetup, owner) = sale;
```
amorask-bitwarden, line 38: ❌ `owner` is no longer used and now we're not doing any coupon validation prior to purchase. My point was not for you to remove the validation code, but to request you fix the order of operations issue that allowed the validation to be bypassed.

src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
```diff
@@ -83,11 +80,20 @@ public Task<BillingCommandResult<None>> Run(
...
         {
             return new BadRequest("Additional storage must be greater than 0.");
         }
 
+        if (!string.IsNullOrWhiteSpace(subscriptionPurchase.Coupon))
+        {
+            var isValid = await subscriptionDiscountService.ValidateDiscountForUserAsync(user, subscriptionPurchase.Coupon.Trim(), DiscountAudienceType.UserHasNoPreviousSubscriptions);
+            if (!isValid)
+            {
+                return new BadRequest("The coupon code is invalid or you are not eligible for this discount.");
```
amorask-bitwarden: ❌ This error message is confusing; the user doesn't enter a coupon code. Aren't these discounts automatically added by our system? In that case, it shouldn't result in a BAD_REQUEST error, it should result in the coupon not being applied.

src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
amorask-bitwarden: Was not addressed.

.claude/settings.local.json
```diff
@@ -0,0 +1,11 @@
+{
```
amorask-bitwarden: ❌ I don't think this should be checked in.

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
```diff
@@ -34,17 +35,36 @@ public class OrganizationBillingService(
...
+    ISubscriptionDiscountRepository subscriptionDiscountRepository,
+    ISubscriptionDiscountService subscriptionDiscountService,
     ITaxService taxService) : IOrganizationBillingService
 {
     public async Task Finalize(OrganizationSale sale)
     {
-        var (organization, customerSetup, subscriptionSetup) = sale;
+        var (organization, customerSetup, subscriptionSetup, owner) = sale;
+
+        // Validate coupon and only apply if valid. If invalid, proceed without the discount.
```
amorask-bitwarden, line 40: ❓ I think this still feels a bit clunky. Can you walk me through the scenario where the `owner` will be `null` or not provided by the invoking method?

src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
```diff
@@ -120,6 +126,24 @@ public class PreviewOrganizationTaxCommand(
                         }
                     }
 
+                    // Validate coupon and only apply if valid. If invalid, proceed without the discount.
+                    if (!string.IsNullOrWhiteSpace(purchase.Coupon))
```
amorask-bitwarden: ⛏️  Can we combine this conditional with the one on 133 so we don't have double nestIng?

src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
```diff
@@ -83,11 +80,22 @@ public Task<BillingCommandResult<None>> Run(
             return new BadRequest("Already a premium user.");
         }
 
-        if (additionalStorageGb < 0)
+        if (subscriptionPurchase.AdditionalStorageGb.HasValue && subscriptionPurchase.AdditionalStorageGb < 0)
```
amorask-bitwarden: ⛏️ `if (subscriptionPurchase.AdditionalStorageGb is < 0)`

src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
```diff
@@ -361,6 +370,11 @@ private async Task<Subscription> CreateSubscriptionAsync(
             OffSession = true
         };
 
+        if (!string.IsNullOrWhiteSpace(coupon))
```
amorask-bitwarden: ❌ This needs to be `validatedCoupon` or your validation will be bypassed entirely.

amorask-bitwarden (changes requested):

.claude/settings.local.json
theMickster, line 1: ❌ Will you please remove this file from the PR?

.claude/settings.local.json
```diff
@@ -0,0 +1,11 @@
...
+  "permissions": {
+    "allow": [
+      "Bash(dotnet format)",
+      "Bash(dotnet format:*)",
+      "Bash(python3:*)"
```
theMickster: ⚠️ Please be careful with wildcards. I recommend you remove this from your local file.

.claude/settings.local.json
cyprain-okeke: removed

.claude/settings.local.json
cyprain-okeke: I have removed it

theMickster: Thank you

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
amorask-bitwarden: @cyprain-okeke Again, my intention was not for you to make an update. I was hoping to discuss this more. These coupons need to be validated using the `SubscriptionDiscountService`, right? 

If so, then we do need to take in the `owner`, but it should not be nullable; it's required. That way, we can ensure we have it and use the owner to validate the coupon by ensuring they've never had premium before. Additionally, it still needs to ensure this is for a Families subscription, similar to what you did for tax.

src/Core/OrganizationFeatures/OrganizationSubscriptions/UpgradeOrganizationPlanCommand.cs
```diff
@@ -230,7 +233,23 @@ await _organizationUserRepository.GetManyByOrganizationAsync(organization.Id,
...
+
+            var owner = await _userRepository.GetByIdAsync(ownerOrganizationUser.UserId.Value);
+
+            if (owner == null)
+            {
+                throw new BadRequestException("Organization owner not found.");
+            }
+
+            var sale = OrganizationSale.From(organization, upgrade, owner);
             await _organizationBillingService.Finalize(sale);
```
amorask-bitwarden: ❌ This doesn't seem right. We're just picking an owner from the list of owners basically at random to validate the discount with? Maybe we do need to rethink this. If being the organization owner is **required** to get the first-time Families discount applied, then maybe owner does need to be nullable in the `OrganizationSale` and it should only be passed in if the `User` executing the operation is the owner of the organization. Then the discount validation also checks for the existence of the owner before trying to validate and apply the discount.

src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
```diff
@@ -81,11 +78,22 @@ public Task<BillingCommandResult<None>> Run(
...
         {
             return new BadRequest("Additional storage must be greater than 0.");
         }
 
+        // Validate coupon and only apply if valid. If invalid, proceed without the discount.
+        string? validatedCoupon = null;
+        if (!string.IsNullOrWhiteSpace(subscriptionPurchase.Coupon))
+        {
+            var isValid = await subscriptionDiscountService.ValidateDiscountForUserAsync(user, subscriptionPurchase.Coupon.Trim(), DiscountAudienceType.UserHasNoPreviousSubscriptions);
+            if (isValid)
```
kdenney: I just want to note here that this should probably throw an exception if the discount is no longer valid. This could happen if the discount was deleted or eligibility changed between the user seeing the discount in the cart summary and creating the transaction. So if we proceed with the transaction without the discount we could be charging the user more than they expected.

src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
```diff
@@ -83,15 +83,16 @@ public Task<BillingCommandResult<None>> Run(
...
+        // Validate coupon if provided. Return error if invalid to prevent charging more than expected.
         string? validatedCoupon = null;
         if (!string.IsNullOrWhiteSpace(subscriptionPurchase.Coupon))
         {
             var isValid = await subscriptionDiscountService.ValidateDiscountForUserAsync(user, subscriptionPurchase.Coupon.Trim(), DiscountAudienceType.UserHasNoPreviousSubscriptions);
-            if (isValid)
+            if (!isValid)
             {
-                validatedCoupon = subscriptionPurchase.Coupon.Trim();
+                return new BadRequest("The discount code is no longer valid. Please remove it or use a different code to proceed with checkout.");
```
kdenney: Is this message shown to the user? Users do not choose discounts and are unable to remove them and the term `discount code` isn't really valid either because there's no coupon code or anything like that involved. If we are displaying this to the user, we might want to change it to something like "The discount is no longer valid. Please try again." or maybe even ask Product what we should say here.

src/Core/Billing/Services/Implementations/SubscriptionDiscountService.cs
```diff
@@ -37,11 +39,19 @@ public async Task<IEnumerable<DiscountEligibility>> GetEligibleDiscountsAsync(Us
...
+        if (discount == null)
         {
             return false;
         }
 
+        if (!IsDiscountActive(discount))
+        {
+            logger.LogWarning("Deleting expired coupon {CouponId} from our table - discount is no longer active",
+                discount.Id);
+            await subscriptionDiscountRepository.DeleteAsync(discount);
```
kdenney: @amorask-bitwarden I know we discussed this in Slack but I'm worried that deleting the discount here could cause confusion especially with our finance team. This is deleting when the discount is outside of the expiration dates. Note that these expiration dates do not come from stripe but are entirely controlled by admins. If we had a way to detect that the coupon is no longer valid in stripe, then I think that would be a good case for deleting but as-written, I'm not sure this really brings much benefit and would result in "why was my discount deleted?" questions. Thoughts?

src/Core/Billing/Organizations/Services/OrganizationBillingService.cs
```diff
@@ -30,17 +30,40 @@ public class OrganizationBillingService(
...
+            {
+                var isValid = await subscriptionDiscountService.ValidateDiscountEligibilityForUserAsync(
+                    owner,
+                    customerSetup.Coupon.Trim(),
+                    DiscountTierType.Families);
+
+                if (isValid)
+                {
+                    validatedCoupon = customerSetup.Coupon.Trim();
+                }
```
kdenney, line 58: ❌  This needs to throw the same bad request exception that is thrown in `CreatePremiumCloudHostedSubscriptionCommand` when the discount is no longer valid.

cyprain-okeke: Added the bad request exception

src/Core/Billing/Services/Implementations/SubscriptionDiscountService.cs
amorask-bitwarden: Agreed I'm not sure why we would delete the `SubscriptionDiscount` record from the database when its not within a `StartDate` and `EndDate` that are set by us in the Bitwarden Admin Portal. That strikes me as confusing. Those values can easily be modified to bring the discount back to Active in our system. If, for whatever reason, the discount was invalid in Stripe itself and could never be applied, it'd make sense to delete it then since it has no use case.

kdenney: From my brief time looking at the stripe docs, it looks like the only specific error we could check for if the subscription creation code throws a StripeException would be:  `coupon_expired`

However, if I remember correctly, we are advising against using Stripes built in expiration date to use the StartDate/EndDate in the portal instead. So that error code gives us essentially nothing.

IMO, at this point we should just remove this code that auto-deletes the discount and in the future we can evaluate possible better ways to detect and auto-delete invalid coupons. @amorask-bitwarden do you agree?

kdenney: Ope, I just noticed that @cyprain-okeke added `IsStripeCouponValidAsync` just below this and it fetches and checks `.Valid` on the coupon from stripe. That's nice that they have that option for checking validity, but IMO the benefit we gain is not worth the expensive stripe call on every subscription purchase with a coupon.

amorask-bitwarden: @kdenney I don't think we'll need to check for a specific error on subscription creation; what @cyprain-okeke added here seems right to me. The `coupon.valid` [property](https://docs.stripe.com/api/coupons/object?api-version=2025-08-27.basil&lang=dotnet#coupon_object-valid) identifies whether or not the coupon can be added to a Customer based on its configuration. If `coupon.valid` is `false`, the application of that coupon by our API, presumably, would result in an exception being thrown, which we'd obviously want to prevent. 

That's separate from whether it's worth deleting the entire coupon, which I'll leave up to you. If we want to be safe and just check `coupon.valid` without doing a deletion of the `SubscriptionDiscount` on an invalid result, I think that's perfectly reasonable.

amorask-bitwarden: Given this is a validity check that will be run iteratively on multiple coupons, I think that's fair and I'm aligned with full removal of it.

PR #6995, opened 2026-02-12
[PM-31052][PM-32469] Add V2UpgradeToken for key rotation without logout

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31052
and https://bitwarden.atlassian.net/browse/PM-32469 to fix EF bug

## 📔 Objective

The V2 upgrade token provides a way for upgrade rotations to retain access to old encrypted data, and both the old and new user-key for a temporary period.
Regular key rotations will still logout  - v2 upgrade token expected to be null.
Exposes the V2 upgrade token in sync's user decryption options response, so other clients will be able to unlock with old and new user key.

(https://bitwarden.atlassian.net/browse/PM-32469) Also fixes a bug in EF UserRepo, where User's V2 fields were not set (security state, version and signed public key)

### DIFF
--- a/util/Migrator/DbScripts/2026-02-19_00_User_V2UpgradeToken.sql
+++ b/util/Migrator/DbScripts/2026-02-19_00_User_V2UpgradeToken.sql
@@ -325,5 +325,9 @@ FROM
     [dbo].[User]
 GO
 
-EXECUTE sp_refreshview 'dbo.UserView'
+EXECUTE sp_refreshview N'[dbo].[EmergencyAccessDetailsView]';
+EXECUTE sp_refreshview N'[dbo].[OrganizationUserUserDetailsView]';
+EXECUTE sp_refreshview N'[dbo].[ProviderUserUserDetailsView]';
+EXECUTE sp_refreshview N'[dbo].[UserEmailDomainView]';
+EXECUTE sp_refreshview N'[dbo].[UserPremiumAccessView]';
 GO


test/Core.Test/KeyManagement/UserKey/RotateUserAccountKeysCommandTests.cs
```diff
@@ -389,7 +390,168 @@ public async Task UpdateUserData_RevisionDateChanged_Success(SutProvider<RotateU
...
+        SutProvider<RotateUserAccountKeysCommand> sutProvider, User user, RotateUserAccountKeysData model)
+    {
+        // Arrange
+        SetTestKdfAndSaltForUserAndModel(user, model);
+        var signatureRepository = sutProvider.GetDependency<IUserSignatureKeyPairRepository>();
+        SetV1ExistingUser(user, signatureRepository);
+        SetV1ModelUser(model);
+
+        var originalSecurityStamp = user.SecurityStamp = Guid.NewGuid().ToString();
+        var wrappedKey1 = "2.key1==|data1==|hmac1==";
```
quexten: Nit: The V1 key that's wrapped would be "7.XXXXXXXXX" since the upgrade migration always is to a COSE key.

test/Core.Test/KeyManagement/Models/Data/V2UpgradeTokenDataTests.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Bit.Core.Test.KeyManagement.Models.Data;
+
+public class V2UpgradeTokenDataTests
+{
+    [Fact]
+    public void ToJson_SerializesCorrectly()
+    {
+        var data = new V2UpgradeTokenData
+        {
+            WrappedUserKey1 = "2.key1==|data1==|hmac1==",
```
quexten: Nit: `7.xxxxxxxxxx`

test/Api.Test/KeyManagement/Models/Request/V2UpgradeTokenRequestModelTests.cs
```diff
@@ -0,0 +1,125 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Api.KeyManagement.Models.Requests;
+using Xunit;
+
+namespace Bit.Api.Test.KeyManagement.Models.Request;
+
+public class V2UpgradeTokenRequestModelTests
+{
+    private const string _validWrappedKey1 = "2.AOs41Hd8OQiCPXjyJKCiDA==|O6OHgt2U2hJGBSNGnimJmg==|iD33s8B69C8JhYYhSa4V1tArjvLr8eEaGqOV7BRo5Jk=";
```
quexten: Nit: `7.xxxxxxxxx`

test/Api.Test/KeyManagement/Controllers/AccountsKeyManagementControllerTests.cs
```diff
@@ -236,6 +237,62 @@ public async Task RotateUserKeyWrongData_Throws(SutProvider<AccountsKeyManagemen
...
+    public async Task RotateUserAccountKeys_WithV2UpgradeToken_PassesTokenToCommand(
+        SutProvider<AccountsKeyManagementController> sutProvider,
+        RotateUserAccountKeysAndDataRequestModel data,
+        User user)
+    {
+        // Arrange
+        data.AccountKeys.SignatureKeyPair = null;
+        data.AccountUnlockData.V2UpgradeToken = new V2UpgradeTokenRequestModel
+        {
+            WrappedUserKey1 = "2.key1==|data1==|hmac1==",
```
quexten: Nit: `7.xxxxxxxxx`

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -233,6 +233,12 @@ public async Task UpdateUserKeyAndEncryptedDataV2Async(Core.Entities.User user,
         userEntity.AccountRevisionDate = user.AccountRevisionDate;
         userEntity.RevisionDate = user.RevisionDate;
 
+        userEntity.SignedPublicKey = user.SignedPublicKey;
```
quexten, line 258: Oof. Makes me wonder if we should QA V2 rotations on EF separately?

test/Core.Test/KeyManagement/UserKey/RotateUserAccountKeysCommandTests.cs
```diff
@@ -389,7 +390,168 @@ public async Task UpdateUserData_RevisionDateChanged_Success(SutProvider<RotateU
...
+        // Assert - Security stamp is updated (logout behavior)
+        Assert.NotEqual(originalSecurityStamp, user.SecurityStamp);
+
+        // Assert - Push notification sent without reason (standard logout)
+        await sutProvider.GetDependency<IPushNotificationService>().Received(1)
+            .PushLogOutAsync(user.Id);
+    }
+
+    [Theory, BitAutoData]
+    public async Task RotateUserAccountKeysAsync_WithExistingToken_WithNewToken_UpdatesToken(
```
quexten, line 508: Hmm, thinking through this scenario, I think we only ever expect *one* upgrade rotation, when going from V1 encryption to V2 encryption. I'm not sure when we would do a second rotation / if we would do this, then I do believe some devices would break.

If a device is on UK1, then goes to UK2, the token T12 allows devices with local method unlocking UK1 to get to UK2.

But if we now update from UK2 to UK3, then the token is T23. If a device still has unlock methods unlocking UK1, then the token would not function / break.

Should we instead only allow the token to be present if:
- The user is a v1 user
- AND The key-rotation contains a V2 cryptographic state
- AND There is no upgrade token present on the user table
OTHERWISE -> This is a regular logout key-rotation?

mzieniukbw: That's true, fixed.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
mzieniukbw: Raised https://bitwarden.atlassian.net/browse/PM-32469 for QA

src/Sql/dbo/Tables/User.sql
```diff
@@ -42,10 +42,11 @@
     [LastKeyRotationDate]              DATETIME2 (7)    NULL,
     [LastEmailChangeDate]              DATETIME2 (7)    NULL,
     [VerifyDevices]                    BIT              DEFAULT ((1)) NOT NULL,
-    [SecurityState]                    VARCHAR (MAX)   NULL,
+    [SecurityState]                    VARCHAR (MAX)    NULL,
     [SecurityVersion]                  INT              NULL,
-    [SignedPublicKey]                  VARCHAR (MAX)   NULL,
+    [SignedPublicKey]                  VARCHAR (MAX)    NULL,
     [MaxStorageGbIncreased]            SMALLINT         NULL,
+    [V2UpgradeToken]                   VARCHAR(MAX)     NULL,
```
mkincaid-bw: ⛏️ Very minor but the majority of our column definitions have a space between the type and the size (ie: `VARCHAR (MAX)`

util/Migrator/DbScripts/2026-02-19_00_User_V2UpgradeToken.sql
```diff
@@ -0,0 +1,329 @@
...
+    [VerifyDevices],
+    [SecurityState],
+    [SecurityVersion],
+    [SignedPublicKey],
+    [V2UpgradeToken]
+FROM
+    [dbo].[User]
+GO
+
+EXECUTE sp_refreshview 'dbo.UserView'
```
mkincaid-bw: FYI, Since you're re-defining the view, you don't need to refresh it. Instead you should refresh any views that reference the updated `User` table that do not get re-defined as part of your migration.

```sql
EXEC sp_refreshview N'[dbo].[EmergencyAccessDetailsView]';
EXEC sp_refreshview N'[dbo].[ProviderUserUserDetailsView]';
EXEC sp_refreshview N'[dbo].[UserEmailDomainView]';
```

src/Core/Constants.cs
```diff
@@ -209,6 +209,7 @@ public static class FeatureFlagKeys
     public const string EnableAccountEncryptionV2KeyConnectorRegistration = "enable-account-encryption-v2-key-connector-registration";
     public const string SdkKeyRotation = "pm-30144-sdk-key-rotation";
     public const string UnlockViaSdk = "unlock-via-sdk";
+    public const string NoLogoutOnKeyUpgradeRotation = "pm-31050-no-logout-key-upgrade-rotation";
```
mkincaid-bw, line 212: ❓ I don't typically flag C# code but my Claude DB reviewer noted that this feature flag is not used server-side. Is this intentional?

mkincaid-bw (changes requested):
Minor change requested.

src/Core/Constants.cs
gbubemismith: Not the author here, but yeah, intentional. We define feature flags server-side so clients can use them, doesn't need to be used by the server itself.

util/Migrator/DbScripts/2026-02-19_00_User_V2UpgradeToken.sql
mzieniukbw: Fixed in fe7b0667

src/Core/Constants.cs
mzieniukbw: This is intentional, we use the feature flag in client's, which when enabled sent alternated request with the V2UpgradeToken (otherwise null), which based on the presence, enabled or disabled the expected behaviour.

quexten (approved):

PR #7023, opened 2026-02-17
[bre-1628] cherry-pick release workflow change

aj-bw:
## 🎟️ Tracking

[bre-1628](https://bitwarden.atlassian.net/browse/bre-1628)

## 📔 Objective

We want to temporarily remove the draft: true key for testing release process enhancements for the next release. This key being passed results in the git tag not being created until we go and manually publish the draft GH release, meaning the following publish workflow will fail.


aj-bw:
> What about the release notes? Is the plan to edit them directly after the release is published?

@vgrassia Yes that is the plan. this is more so a stop gap for this next release only, next sprint I will improve it to retrieve the release notes from the card and pass it to the workflow.

PR #7025, opened 2026-02-17
[PM-31671] Add additional checks when uploading attachment.

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31671


nikwithak:
Fixed by #7053

PR #7030, opened 2026-02-18
Implement plan types, personal ciphers and fix folder assignment

theMickster:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31403
https://bitwarden.atlassian.net/browse/PM-31013
https://bitwarden.atlassian.net/browse/PM-32079
https://bitwarden.atlassian.net/browse/PM-31019

## 📔 Objective

- Implemented the current plan types and preset seeds 
    - I had Claude Code scour the code base for documentation and unit tests to see what plans are active as of now. 
- Implemented a step to create user ciphers
    - Extract the overlapping code into a new class titled `CipherComposer` instead of duplicating in a each cipher step or creating a base class. 
- Implemented a step that will properly create Folders and associate them to Ciphers

### DIFF
--- a/util/Seeder/Factories/PlanFeatures.cs
+++ b/util/Seeder/Factories/PlanFeatures.cs
@@ -109,7 +109,7 @@ public static class PlanFeatures
         var (min, max, avg) = planType switch
         {
             PlanType.Free => (1, 2, 2),
-            PlanType.FamiliesAnnually => (5, 6, 6),
+            PlanType.FamiliesAnnually => (6, 6, 6),
             PlanType.TeamsMonthly => (1, 300, 15),
             PlanType.TeamsAnnually => (1, 100, 7),
             PlanType.TeamsStarter => (10, 10, 10),
--- a/util/DbSeederUtility/README.md
+++ b/util/DbSeederUtility/README.md
@@ -69,7 +69,7 @@ dotnet run -- vault-organization -n ApacOrg -d apac.test -u 17 -c 600 -g 12 --re
 dotnet run -- vault-organization -n IsolatedOrg -d isolated.test -u 5 -c 25 -g 4 -o Spotify --mangle
 
 # With custom password for all accounts
-dotnet run -- vault-organization -n CustomPwOrg -d custom-password-02.test -u 10 -c 100 -g 3 --password "MyTestPassword1"
+dotnet run -- vault-organization -n CustomPwOrg -d custom-password-05.test -u 10 -c 100 -g 3 --password "MyTestPassword1" --plan-type teams-annually
 
 # Free plan org (limited to 2 seats, 2 collections)
 dotnet run -- vault-organization -n FreeOrg -d free.test -u 1 -c 10 -g 1 --plan-type free


util/Seeder/Steps/GenerateFoldersStep.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        var generator = context.RequireGenerator();
+        var userDigests = context.Registry.UserDigests;
+        var distribution = FolderCountDistributions.Realistic;
+
+        for (var index = 0; index < userDigests.Count; index++)
+        {
+            var digest = userDigests[index];
+            var range = distribution.Select(index, userDigests.Count);
+            var count = range.Min + (index % Math.Max(range.Max - range.Min + 1, 1));
```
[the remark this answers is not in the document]
theMickster: Fixed

util/Seeder/Factories/CipherComposer.cs
theMickster, line 1: 📝 This was refactored from the `GenerateCiphersStep.cs` class.  Only new code is the conditional to use the user or org key

util/Seeder/Factories/PlanFeatures.cs
theMickster, line 1: 📝 The intent here is to extract away the application of toggles that control configuring an organization based upon the plan the organization signed up for into a single location. I don't _love_ the `Apply()` but it works well giving us what we need.

util/Seeder/Steps/GenerateCiphersStep.cs
```diff
@@ -90,67 +99,4 @@ public void Execute(SeederContext context)
         context.Registry.CipherIds.AddRange(cipherIds);
         context.CollectionCiphers.AddRange(collectionCiphers);
     }
-
-    private static Cipher CreateLoginCipher(
```
theMickster, line 94: 📝  All of this code now resides in the `CipherComposer.cs`

util/Seeder/Factories/PlanFeatures.cs
theMickster, line 1: Hey @amorask-bitwarden 👋🏼
When you are at a stopping point, I could use your help with a review of the PlanType factory that I have created for the Seeder utility.  I would appreciate a review that I have accurately selected the correct plans that we offer . Then, that the `ApplyTeamsFeatures()` and `ApplyEnterpriseFeatures()` methods correctly apply the features offered in those types of plans.  I collaborated with Claude Code by having it research our online docs + our test cases + the PlanType enum itself. 

Thank you for your help on this file!

MGibson1: It's a good idea to create a custom type, as well. That's an area of poor test coverage, but important to get right.

The previous means of features would be a good approximation of what would be expected in a custom plan, but the admin project's organization "features" is a good start on what would be needed

util/Seeder/Factories/PlanFeatures.cs
```diff
@@ -0,0 +1,219 @@
...
+                $"Invalid planType string '{planTypeString}'. Valid values: free, teams-monthly, " +
+                "teams-annually, teams-starter, enterprise-monthly, enterprise-annually, families-annually.")
+        };
+    }
+
+    /// <summary>
+    /// Deterministic seat count from a log-normal distribution seeded by domain.
+    /// Ranges sourced from our production data.
+    /// </summary>
+    internal static int GenerateRealisticSeatCount(PlanType planType, string domain)
```
nthompson-bitwarden, line 107: this is great 👍

util/DbSeederUtility/VaultOrganizationArgs.cs
```diff
@@ -40,6 +41,9 @@ public class VaultOrganizationArgs : IArgumentModel
     [Option("password", Description = "Password for all seeded accounts (default: asdfasdfasdf)")]
     public string? Password { get; set; }
 
+    [Option("plan-type", Description = "Billing plan type: free, teams-monthly, teams-annually, enterprise-monthly, enterprise-annually, teams-starter, families-annually. Defaults to enterprise-annually.")]
```
amorask-bitwarden, line 44: `teams-starter` is no longer offered, but if the intention is to potentially support the triaging of issues with our legacy `teams-starter` organizations, then it's worth leaving in. If this is only for exemplifying new organizations, it should be removed. Given that `PlanFeatures` is only working with currently offered plans, it seems like the latter.

util/Seeder/Factories/PlanFeatures.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+        var (min, max, avg) = planType switch
+        {
+            PlanType.Free => (1, 2, 2),
+            PlanType.FamiliesAnnually => (5, 6, 6),
```
amorask-bitwarden: The minimum and maximum for Families is always 6 - it's a packaged product (seats are not adjustable).

util/Seeder/Factories/PlanFeatures.cs
```diff
@@ -0,0 +1,219 @@
...
+using Bit.Core.AdminConsole.Entities;
+using Bit.Core.Billing.Enums;
+
+namespace Bit.Seeder.Factories;
+
+/// <summary>
+/// Maps PlanType to organization feature flags.
+/// Values sourced from MockPlans in test/Core.Test/Billing/Mocks/Plans/.
+/// </summary>
+public static class PlanFeatures
```
amorask-bitwarden, line 12: See note here regarding `teams-starter`: https://github.com/bitwarden/server/pull/7030/changes#r2828090840

util/DbSeederUtility/VaultOrganizationArgs.cs
theMickster: The intention is to allow for the flexibility to triage issues of any 'supported feature' (and I use that phrase loosely).  If we have some legacy organizations that use `teams-starter` and there remains any level of pain to support then I want to leave it. I want to Seeder to support teams in supporting all the work we have to do which includes working on the legacy things.  It's not intended to only support the happy-path or new shinny features. 

I think I'm going to leave it for now as it doesn't hurt anything.

util/Seeder/Factories/PlanFeatures.cs
theMickster: Ah, gotcha. I will fix that because that feels like it should be corrected.

util/Seeder/Factories/PlanFeatures.cs
theMickster: Just to round out the conversation... Alex mentioned that 

> PlanType.Custom is not used in any actual production scenarios and can be safely ignored.

PR #7032, opened 2026-02-18
[PM-30109] edit discounts in bitwarden portal

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30109

## 📔 Objective

Adds edit and delete functionality to subscription discounts in the bitwarden portal.

## 📸 Screenshots

image

image


src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -205,5 +206,100 @@ public async Task<IActionResult> Create(CreateSubscriptionDiscountModel model)
         }
     }
 
+    [HttpGet("{id}")]
+    [RequirePermission(Permission.Tools_CreateEditTransaction)]
```
amorask-bitwarden, line 210: ❓ Any particular reason we're using `CreateEditTransaction`?

src/Admin/Billing/Models/EditSubscriptionDiscountModel.cs
```diff
@@ -0,0 +1,67 @@
...
+    public DateTime EndDate { get; set; }
+
+    [Display(Name = "Restrict to users with no previous subscriptions?")]
+    public bool RestrictToNewUsersOnly { get; set; }
+
+    public DiscountAudienceType AudienceType => RestrictToNewUsersOnly
+        ? DiscountAudienceType.UserHasNoPreviousSubscriptions
+        : DiscountAudienceType.AllUsers;
+
+    public string DiscountDisplay => PercentOff.HasValue
```
amorask-bitwarden: ⛏️ I don't think this is ever used in a view unless I'm missing something.

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
```diff
@@ -205,5 +206,100 @@ public async Task<IActionResult> Create(CreateSubscriptionDiscountModel model)
...
+    {
+        var discount = await subscriptionDiscountRepository.GetByIdAsync(id);
+        if (discount == null)
+        {
+            return NotFound();
+        }
+
+        var model = new EditSubscriptionDiscountModel(discount);
+
+        if (model.StripeProductIds != null && model.StripeProductIds.Count != 0)
```
amorask-bitwarden: ⛏️ `if (model.StripeProductIds is { Count: > 0 })`

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
kdenney: Yes I asked product and they agreed that we didn't need to add a new permission and that reusing this would suffice.

src/Admin/Billing/Controllers/SubscriptionDiscountsController.cs
kdenney: I really like this and I wish Jetbrains Rider would suggest this pattern when it could. Maybe it's a setting I need to find...

Thanks for the suggestion!! Added in latest commit.

src/Admin/Billing/Models/EditSubscriptionDiscountModel.cs
kdenney: Good catch! Removed in latest commit.

PR #7035, opened 2026-02-18
[PM-31433] Extension Prompt Account Age Feature Flag

nick-livefront:
## 🎟️ Tracking

[PM-31433](https://bitwarden.atlassian.net/browse/PM-31433)

## 📔 Objective

Adds feature flag associated for the account age extension prompt.

## 📸 Screenshots

N/A

[PM-31433]: https://bitwarden.atlassian.net/browse/PM-31433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -258,6 +258,7 @@ public static class FeatureFlagKeys
     public const string PM30521_AutofillButtonViewLoginScreen = "pm-30521-autofill-button-view-login-screen";
     public const string PM32180_PremiumUpsellAccountAge = "pm-32180-premium-upsell-account-age";
     public const string PM29438_WelcomeDialogWithExtensionPrompt = "pm-29438-welcome-dialog-with-extension-prompt";
+    public const string PM29438_DialogWithExtensionPromptAccountAge = "pm-29438-dialog-with-extension-prompt-account-age";
     public const string PM31039_ItemActionInExtension = "pm-31039-item-action-in-extension";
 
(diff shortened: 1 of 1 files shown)


src/Core/Constants.cs
```diff
@@ -258,6 +258,7 @@ public static class FeatureFlagKeys
     public const string PM30521_AutofillButtonViewLoginScreen = "pm-30521-autofill-button-view-login-screen";
     public const string PM32180_PremiumUpsellAccountAge = "pm-32180-premium-upsell-account-age";
     public const string PM29438_WelcomeDialogWithExtensionPrompt = "pm-29438-welcome-dialog-with-extension-prompt";
+    public const string PM29438_DialogWithExtensionPromptAccountAge = "pm-29438-dialog-with-extension-prompt-account-age";
```
shane-melton, line 261: ❓ The flag has different ticket numbers between the ticket description and the epic description. Not sure which is desired, but wanted to point it out to avoid needing to recreate it in LD if it already exists.

nick-livefront: Good catch, that's my mistake in the description here. I linked the feature work rather than the epic. 

`pm-29438` is correct and is the epic number.

shane-melton: Thanks for clarifying! 🚀

PR #7038, opened 2026-02-18
[PM-24788] Remove Org-Based Policy Requirements Queries

sven-bitwarden:
## 🎟️ Tracking
[PM-24788](https://bitwarden.atlassian.net/browse/PM-24788)

## 📔 Objective
Previously within the policy requirements feature, we could only gather enforcement data for individual users, or for organizations. #6876 enabled retrieving policy requirements for multiple users, which is much more efficient in scenarios where out of an entire organization, we need policy requirements for a handful of users.

This PR adjusts the areas that had previously used the by-org query to load by the user IDs on hand.  It also changes the OrgDataOwnershipPolicyRequirement to return a helpful (pre-existing) model that gives back the org user ID it's applicable to.

[PM-24788]: https://bitwarden.atlassian.net/browse/PM-24788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
@@ -51,7 +51,8 @@ public class ConfirmOrganizationUserCommand : IConfirmOrganizationUserCommand
         IPolicyRequirementQuery policyRequirementQuery,
         IFeatureService featureService,
         ICollectionRepository collectionRepository,
-        IAutomaticUserConfirmationPolicyEnforcementValidator automaticUserConfirmationPolicyEnforcementValidator, ISendOrganizationConfirmationCommand sendOrganizationConfirmationCommand)
+        IAutomaticUserConfirmationPolicyEnforcementValidator automaticUserConfirmationPolicyEnforcementValidator,
+        ISendOrganizationConfirmationCommand sendOrganizationConfirmationCommand)
     {
         _organizationRepository = organizationRepository;
         _organizationUserRepository = organizationUserRepository;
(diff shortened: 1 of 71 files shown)


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/ConfirmOrganizationUserCommand.cs
```diff
@@ -103,13 +103,10 @@ public async Task<List<Tuple<OrganizationUser, string>>> ConfirmUsersAsync(Guid
             .Select(r => r.Item1)
             .ToList();
 
-        if (confirmedOrganizationUsers.Count == 1)
+        if (confirmedOrganizationUsers.Count > 0)
```
sven-bitwarden: The difference between these two methods (== 1 vs > 1) was whether or not it used the by-org or by-user policy requirements query.

Since we removed the by-org query, _and_ the by-user query delegates to the by-multiple-users query, these methods don't need to be separate anymore.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RestoreUser/v1/RestoreOrganizationUserCommand.cs
```diff
@@ -263,22 +263,31 @@ await CreateDefaultCollectionsForConfirmedUsersAsync(organizationId, defaultColl
...
-            var organizationUsersDataOwnershipEnabled = (await policyRequirementQuery
-                    .GetManyByOrganizationIdAsync<OrganizationDataOwnershipPolicyRequirement>(organizationId))
-                .ToList();
+            return;
+        }
 
-            var usersToCreateDefaultCollectionsFor = restoredUsers.Where(x =>
-                organizationUsersDataOwnershipEnabled.Contains(x.Id)
-                && x.Status == OrganizationUserStatusType.Confirmed).ToList();
+        var restoredConfirmedUsers = restoredUsers
```
r-tome, line 274: Consider an early return here if this list is empty

sven-bitwarden: Good call ✅

sven-bitwarden:
@r-tome fixed some merge conflicts and got the tests restored

PR #7014, opened 2026-02-17
[PM-32131] Add UseMyItems organization ability

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32131
## 📔 Objective

Add a `UseMyItems` organization ability. This can be manually toggled off in the Bitwarden Portal in order to disable the creation of My Items when using the Organization Data Ownership policy. It will be on by default and only disabled upon customer request.

This PR adds the ability, handles setting the default value, and allows it to be manually overridden. It does not wire it up to any business logic yet.

This was added following the instructions in https://github.com/bitwarden/server/pull/6970.

**Design decision**
We could structure this as `UseMyItems` (positive) or `DisableMyItems` (negative). I went with `UseMyItems` because:
- My Items is a feature, and this follows the pattern of all other features (i.e. we have `UseSso`, not `DisableSso`)
- it's conceptually clearer to "enable or disable the feature" rather than "enable the flag that disables the feature" (inverted)

**Summary of changes**
- new organization table column with all sprocs and views updated
- data migration to enable the feature for all organizations that already use policies (replicating existing behaviour)
  - note: most LOC changes are due to autogenerated db migrations, but open to feedback if this PR needs to be split up
- data and api models updated
- added to organization license file so it can be preserved in self-hosted instances
- note that we don't have a plan feature defined for this yet, so it is set based on `UsePolicies` to match the migration logic. Billing Team have a ticket to add this, and I have marked the affected locations with `TODO`s, including ticket reference.

PR #7039, opened 2026-02-19
PM-31847 Wire up UseMyItems organization ability

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31847

## 📔 Objective

Follow on from #7014 - now actually wire up this new ability flag into domain code so it can enable/disable creation of My Items collections.

There is some code duplication between ~5 different locations which create My Items collections. However, I have not DRY'd this up because:
- this is a time limited request and we need to avoid large refactors
- it's not clear how to do that, particularly because different code paths use a different `PolicyRequirements` interface

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
@@ -55,7 +55,7 @@ public class OrganizationDataOwnershipPolicyValidator(
     {
         // FIXME: we should use the organizationAbility cache here, but it is currently flaky
         // and it's not obvious how to handle a cache failure.
-        // Revisit once https://bitwarden.atlassian.net/browse/PM-32104 is done
+        // https://bitwarden.atlassian.net/browse/PM-32699
         var organization = await organizationRepository.GetByIdAsync(policyUpdate.OrganizationId);
         if (organization == null)
         {


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
```diff
@@ -52,6 +53,20 @@ public async Task ExecuteSideEffectsAsync(
 
     private async Task UpsertDefaultCollectionsForUsersAsync(PolicyUpdate policyUpdate, string defaultCollectionName)
     {
+        // FIXME: we should use the organizationAbility cache here, but it is currently flaky
```
JimmyVo16, line 56: Hey @eliykat , I’m curious, how are we tracking this? I totally agree with the decision here, but I want to make sure it doesn’t get lost. Maybe we should add this to the ability cache initiative. It doesn’t have to be included in epic PM-32104.

eliykat:
Some additional reviewers were automatically requested for review when the base branch changed. Sorry about that! If you're not a code owner, please ignore.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/OrganizationDataOwnershipPolicyValidator.cs
eliykat: Good question, I do think that epic is probably the best place to track this (and any other instance where we've sidestepped the cache). I'll add a ticket as follow-up work after release of that feature.

JimmyVo16: Okay, I added it to the [tech spec](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2541027329/ACTD+Research+migration+plan+to+the+Bitwarden+extended+cache+pattern#After-the-migration-work).

PR #7040, opened 2026-02-19
[PM-29090][Step 2] Remove the server feature flag

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29090

## 📔 Objective

Removes the feature flag pm-26793-fetch-premium-price-from-pricing-service which has been enabled and validated in production.

  **Changes**

  **This PR cleans up the feature flag by:**

  1. Removed feature flag constant from src/Core/Constants.cs
    - Deleted PM26793_FetchPremiumPriceFromPricingService constant
  2. Simplified PricingClient.ListPremiumPlans() method in src/Core/Billing/Pricing/PricingClient.cs
    - Removed feature flag check that toggled between fetching premium plans from the pricing service vs. using a hardcoded plan
    - Now always fetches premium plans from the pricing service (the enabled behavior)
    - Removed the CurrentPremiumPlan static property which was only used when the flag was disabled
    - Cleaned up unused using directives

### DIFF
--- a/src/Core/Billing/Pricing/PricingClient.cs
+++ b/src/Core/Billing/Pricing/PricingClient.cs
@@ -1,6 +1,5 @@
 ﻿using System.Net;
 using System.Net.Http.Json;
-using Bit.Core.Billing.Constants;
 using Bit.Core.Billing.Enums;
 using Bit.Core.Billing.Pricing.Organizations;
 using Bit.Core.Exceptions;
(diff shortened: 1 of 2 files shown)


amorask-bitwarden (commented):
This flag is used in the Desktop and Browser Extension applications and therefore has to follow this process: https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1152417888/Feature+Flags+Lifecycle

The `server` flag can't be removed for at least 3 release cycles.

PR #7047, opened 2026-02-20
PM-32517 Create Migration for Azure Blob Storage 

prograhamming:
## 🎟️ Tracking

This is a PR for a sub-task ([PM-32517](https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?selectedIssue=PM-32517)) on user story [PM-31923](https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?selectedIssue=PM-31923)

## 📔 Objective

We need to add FileId to the OrganizationReport column in order to track where the file is stored in Azure Blob Storage or the file path location for Self-Hosted.

## 📸 Screenshots

[PM-32517]: https://bitwarden.atlassian.net/browse/PM-32517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-31923]: https://bitwarden.atlassian.net/browse/PM-31923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
```diff
@@ -19,6 +19,7 @@ CREATE TABLE [dbo].[OrganizationReport] (
     [PasswordAtRiskCount]               INT              NULL,
     [CriticalPasswordCount]             INT              NULL,
     [CriticalPasswordAtRiskCount]       INT              NULL,
+    [FileId]                            VARCHAR(100)     NULL,
```
Banrion: ❓ Is 100 the length we use wherever else we use a file id field? I'm asking because we should probably make them the same for consistency.

util/Migrator/DbScripts_transition/2026-02-25_00_OrganizationReport_PopulateFileId.sql
```diff
@@ -0,0 +1,9 @@
+-- Phase 2 (Transition): Populate FileId for existing OrganizationReport records
+-- This runs after deployment, performs batched data migration
+
+-- Set FileId to 'Legacy' for all existing v1 reports (data stored in DB, not blob storage)
+-- This sentinel value makes it easy to distinguish v1 (DB-stored) from v2 (blob-stored) reports
+UPDATE [dbo].[OrganizationReport]
+SET [FileId] = 'Legacy'
```
Banrion: 👍 This will make it easier on the client to determine when migration needs to occur and handles the non-nullable fields.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
prograhamming: We can put the follow that pattern and put it in the SummaryData column. We are the only ones that use a file id column. CipherAttachment and SendFileData are JSON blobs with the file id inside of them.

prograhamming: We are the only ones doing this right now. The other areas of the codebase put the file id in a JSON blob.

Banrion:
Noting here that we may need to make changes to the structure around the report file and store more than the file ID following patterns elsewhere in the codebase. This pull request only adds an id field but we are noticing that other places storing files are also storing more metadata regarding the file.

src/Sql/dbo/Dirt/Tables/OrganizationReport.sql
theMickster: I too am concerned that 100 characters is not nearly enough. 
The PR description states a file location in Azure Blob storage. Is the container location configured somewhere else and this only a file name? Or will this store the entire path to the file?

Also, why the decision on VARCHAR vs NVARCHAR?

util/Migrator/DbScripts_transition/2026-02-25_00_OrganizationReport_PopulateFileId.sql
theMickster, line 1: I disagree that this is easier to distinguish versus a null, and this whole migration feels a little unnecessary. What does having the text 'Legacy' benefit us over a null check in the C# business layer?

src/Sql/dbo_finalization/Dirt/Tables/OrganizationReport.sql
```diff
@@ -0,0 +1,36 @@
...
+    [CriticalApplicationAtRiskCount]    INT              NULL,
+    [MemberCount]                       INT              NULL,
+    [MemberAtRiskCount]                 INT              NULL,
+    [CriticalMemberCount]              INT              NULL,
+    [CriticalMemberAtRiskCount]         INT              NULL,
+    [PasswordCount]                     INT              NULL,
+    [PasswordAtRiskCount]               INT              NULL,
+    [CriticalPasswordCount]             INT              NULL,
+    [CriticalPasswordAtRiskCount]       INT              NULL,
+    [FileId]                            VARCHAR(100)     NOT NULL,
```
mkincaid-bw: The `FileId` column is defined as `NULL` elsewhere but `NOT NULL` here. Also, I don't believe we are using the `dbo_finalization` directory. If the ultimate final state is that the new column ends up as `NOT NULL`, I believe that will need to be handled differently.

util/Migrator/DbScripts/2026-02-25_00_OrganizationReport_AddFileId.sql
```diff
@@ -0,0 +1,149 @@
...
+        [CriticalPasswordCount] = @CriticalPasswordCount,
+        [CriticalPasswordAtRiskCount] = @CriticalPasswordAtRiskCount,
+        [FileId] = @FileId
+    WHERE [Id] = @Id;
+END;
+GO
+
+-- Set FileId to 'Legacy' for all existing v1 reports (data stored in DB, not blob storage)
+-- This sentinel value makes it easy to distinguish v1 (DB-stored) from v2 (blob-stored) reports
+UPDATE [dbo].[OrganizationReport]
```
mkincaid-bw: ❓ Is it possible that new rows can come in between the time that this migration runs and the time that the rest of the code is in use? If so, would that cause issues that new "Legacy" rows come in with a null value?

util/SqliteMigrations/Migrations/20260227203606_2026-02-27_00_OrganizationReport_AddReportFile.cs
mkincaid-bw, line 1: ❓The new column in SQL Server is `varchar(100)` but there are no length constraints in the EF migrations. Should they have `HasMaxLength(100)`?

prograhamming: Just a heads up, we are going to add a ReportFile field instead to store all the file data.

prograhamming:
Closing this pull request due to redoing migrations causing issues and will start from a clean slate.

PR #7053, opened 2026-02-20
[PM-31657] Address Overwriting Attachments 

nick-livefront:
## 🎟️ Tracking

[PM-31657](https://bitwarden.atlassian.net/browse/PM-31657)

## 📔 Objective

Adds `ValidateCipherEditForAttachmentAsync` check to each attachment endpoint so the permission of the user hitting the API directly is validated. 

## 📸 Screenshots

A user with read only access to a cipher overwriting an attachment
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/71e50a1a-cdd7-4a0a-9870-b56d23008528" />|<video src="https://github.com/user-attachments/assets/61f7f66f-633d-4ff7-a9a9-10ff315f5344" />|

[PM-31657]: https://bitwarden.atlassian.net/browse/PM-31657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nick-livefront:
> 💡 Since `RenewFileUploadUrl` and `PostFileForExistingAttachment` don't appear have test coverage yet, would it be worth adding a few tests alongside this change so that we ensure that the vulnerability does not crop up again? Specifically for the read-only user scenario you're fixing, and maybe the org-admin happy path. That would lock in the behavior you're adding and make it harder for someone to accidentally break it.

Yes! Thank you for the callout, added unit tests: [e4c4ab2](https://github.com/bitwarden/server/pull/7053/commits/e4c4ab23a379efcba9e756f04803a6bf9f05f896)

theMickster:
> > 💡 Since `RenewFileUploadUrl` and `PostFileForExistingAttachment` don't appear have test coverage yet, would it be worth adding a few tests alongside this change so that we ensure that the vulnerability does not crop up again? Specifically for the read-only user scenario you're fixing, and maybe the org-admin happy path. That would lock in the behavior you're adding and make it harder for someone to accidentally break it.
> 
> Yes! Thank you for the callout, added unit tests: [e4c4ab2](https://github.com/bitwarden/server/pull/7053/commits/e4c4ab23a379efcba9e756f04803a6bf9f05f896)

Thank you very much for crafting those tests!  Never wanna see this 🐛 again 😆

I will leave it to the vault team to review and approve the PR in more depth

nick-livefront:
@shane-melton - I fixed a merge conflict, can you give this another review when you have time?

PR #7063, opened 2026-02-24
[PM-30271] Create Retrieve Eligible Subscription Discounts Endpoint

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30271

## 📔 Objective

This pull request introduces a new API endpoint for retrieving eligible subscription discounts for a user and implements the supporting backend infrastructure for discount eligibility evaluation. It adds new service interfaces, implementations, audience filter logic, and comprehensive tests to ensure correct discount selection and mapping.

**API and Controller Enhancements:**
- Added a new `GET /discounts` endpoint in `AccountBillingVNextController` that returns all applicable subscription discounts for the authenticated user. This endpoint is feature-flagged and uses a new query interface to fetch eligible discounts. 

**Discount Eligibility Infrastructure:**
- Introduced the `ISubscriptionDiscountService` interface and its implementation `SubscriptionDiscountService` to manage discount eligibility checks, including support for audience-based rules. 
- Added the `IDiscountAudienceFilter` interface and a factory (`IDiscountAudienceFilterFactory`/`DiscountAudienceFilterFactory`) for returning DI filters based on discount targeting rules, with an initial implementations for "UserHasNoPreviousSubscriptions" and "AllUsers". 

**Query and Response Models:**
- Created the `IGetApplicableDiscountsQuery` interface and its implementation to return eligible discounts mapped to the new `SubscriptionDiscountResponseModel`, which includes all relevant discount details. 

**Dependency Injection and Service Registration:**
- Registered all new services, queries, and filter factories in the payment service registration module to ensure proper DI resolution. 

### Stripe Integration and Constants

* Expanded `StripeConstants` to include product IDs and a utility to map Stripe product IDs to discount tiers, and added a method to list subscriptions via Stripe. 

### API Response Model

* Added `SubscriptionDiscountResponseModel` to standardize the API response for discount information, including per-tier eligibility.

**Note: Stripe Product IDs are no longer returned with discounts**
## 📸 Screenshots

https://github.com/user-attachments/assets/e8ba8682-d3d1-4159-be19-a30a53ce5cae

### DIFF
--- a/util/MsSql/Dockerfile
+++ b/util/MsSql/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/mssql/server:2022-CU12-ubuntu-22.04
+FROM mcr.microsoft.com/mssql/server:2022-CU22-ubuntu-22.04
 
 LABEL com.bitwarden.product="bitwarden"
 
@@ -19,7 +19,7 @@ RUN chmod +x /entrypoint.sh \
 
 RUN /opt/mssql/bin/mssql-conf set telemetry.customerfeedback false
 
-HEALTHCHECK --start-period=120s --timeout=3s CMD /opt/mssql-tools/bin/sqlcmd \
-    -S localhost -U sa -P ${SA_PASSWORD} -Q "SELECT 1" || exit 1
+HEALTHCHECK --start-period=120s --timeout=3s CMD /opt/mssql-tools18/bin/sqlcmd \
+    -S localhost -U sa -P ${SA_PASSWORD} -C -Q "SELECT 1" || exit 1
 
 ENTRYPOINT ["/entrypoint.sh"]
--- a/util/MsSql/backup-db.sh
+++ b/util/MsSql/backup-db.sh
@@ -14,7 +14,7 @@ do
   export now=$(date +%Y%m%d_%H%M%S)
 
   # Do a new backup
-  /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P ${SA_PASSWORD} -i /backup-db.sql
+  /opt/mssql-tools18/bin/sqlcmd -S localhost -U sa -P ${SA_PASSWORD} -C -i /backup-db.sql
 
   # Delete backup files older than 30 days
   grep -B1 "BACKUP DATABASE successfully" /var/opt/mssql/log/errorlog | grep -q _$now.BAK &&


src/Core/Billing/Payment/Registrations.cs
```diff
@@ -15,7 +18,12 @@ public static void AddPaymentOperations(this IServiceCollection services)
         services.AddTransient<IUpdateBillingAddressCommand, UpdateBillingAddressCommand>();
         services.AddTransient<IUpdatePaymentMethodCommand, UpdatePaymentMethodCommand>();
 
+        // Discount services
+        services.AddTransient<ISubscriptionDiscountService, SubscriptionDiscountService>();
+        services.AddTransient<IDiscountAudienceFilterFactory, DiscountAudienceFilterFactory>();
```
kdenney: ❓ What do you think about adding the factory class as a singleton? Since it shouldn't ever need instantiated more than once, I don't think... Do you agree?

src/Core/Billing/Services/DiscountAudienceFilters/DiscountAudienceFilterFactory.cs
```diff
@@ -0,0 +1,19 @@
+﻿#nullable enable
+
+using Bit.Core.Billing.Enums;
+
+namespace Bit.Core.Billing.Services.DiscountAudienceFilters;
+
+/// <inheritdoc />
+/// <remarks>
+/// To add support for a new audience type: add an enum value, create a filter class, and add a case here.
+/// </remarks>
```
kdenney, line 11: Nice. Thanks for adding this comment here!

test/Core.Test/Billing/Services/DiscountAudienceFilters/DiscountAudienceFilterFactoryTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    [Fact]
+    public void GetFilter_UnknownAudienceType_ReturnsNull()
+    {
+        // Act
+        var filter = _sut.GetFilter((DiscountAudienceType)99);
+
+        // Assert
+        Assert.Null(filter);
+    }
```
kdenney, line 50: ❓ What do you think about adding a unit test here that would use reflection to loop through all of the possible values in the enum and with the exception of `AllUsers`,  would fail if any don't have a corresponding filter returned, as a guard against the pattern not being followed?

src/Core/Billing/Payment/Registrations.cs
sbrown-livefront: I do! Great callout. I'll update this

test/Core.Test/Billing/Services/DiscountAudienceFilters/DiscountAudienceFilterFactoryTests.cs
sbrown-livefront: Ah, so I understand, if a user adds a new enum but not a filter,  the unit test will show there is no corresponding filter?

Great idea.

kdenney: Yep exactly. 🙌

sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/79ddbc03d0f000e453a7aedc7f6b9a08081f14d5

src/Core/Billing/Payment/Registrations.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/beba0bd596596fe3687ea3f0dc36bfbb8e22fc54

src/Core/Billing/Models/Api/Response/SubscriptionDiscountResponseModel.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using Bit.Core.Billing.Enums;
+using Bit.Core.Billing.Subscriptions.Entities;
+
+namespace Bit.Core.Billing.Models.Api.Response;
+
+public class SubscriptionDiscountResponseModel
+{
+    public string StripeCouponId { get; init; } = null!;
+    public IEnumerable<string>? StripeProductIds { get; init; }
```
kdenney: These product ids are no longer needed, I think. Because we have the eligibility dictionary.

src/Core/Billing/Payment/Registrations.cs
kdenney: Do you think the factory can go back to being a singleton now that it doesn't have to do any DI or retain state?

src/Core/Billing/Services/DiscountAudienceFilters/AllUsersFilter.cs
```diff
@@ -0,0 +1,37 @@
...
+    public Task<IDictionary<DiscountTierType, bool>> IsUserEligible(User user, SubscriptionDiscount discount)
+    {
+        var eligibleTiers = Utilities.GetTierEligibilityDictionary();
+
+        if (discount.StripeProductIds == null || !discount.StripeProductIds.Any())
+        {
+            // If no product IDs are specified, the discount applies to all tiers
+            foreach (var tier in eligibleTiers.Keys.ToList())
+            {
+                eligibleTiers[tier] = true;
```
kdenney, line 21: I think this is a problem if we are going to keep teams and enterprise in the discount tier types enum. Technically the front end could ignore those tiers but I don't love that the API is presenting a state that is incorrect by stating that any discounts are eligible for teams or enterprise plans. I think the easiest thing to do for now is remove teams and enterprise from the enum. What do you think?

src/Core/Billing/Services/DiscountAudienceFilters/UserHasNoPreviousSubscriptionsFilter.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private bool DiscountContainsProductIds(SubscriptionDiscount discount) =>
+        discount.StripeProductIds?.Any() ?? false;
+
+    private bool IsApplicableToFamiliesProduct(SubscriptionDiscount discount) =>
+        discount.StripeProductIds?.Contains(StripeProductIDs.Families) ?? false;
+
+    private bool IsApplicableToPremiumProduct(SubscriptionDiscount discount) =>
+        discount.StripeProductIds?.Contains(StripeProductIDs.Premium) ?? false;
```
kdenney: Suggestion: This could be simplified by having the premium and families methods return true if there are no product IDs specified, removing the need for the `DiscountContainsProductIds` method entirely.

src/Core/Billing/Services/DiscountAudienceFilters/UserHasNoPreviousSubscriptionsFilter.cs
```diff
@@ -0,0 +1,119 @@
...
+        var premiumPriceIds = premiumPlans.Select(p => p.Seat.StripePriceId).ToHashSet();
+        try
+        {
+            var subscriptions = await stripeAdapter.ListSubscriptionsAsync(new SubscriptionListOptions
+            {
+                Customer = user.GatewayCustomerId,
+                Expand = ["data.items.data.price"]
+            });
+            return subscriptions.Data.Any(subscription =>
+                subscription.Items.Data.Any(item => premiumPriceIds.Contains(item.Price.Id)));
```
kdenney, line 99: I really liked your idea in our huddle of making this filter's DI registration use `AddScoped` and cache&reuse the result of the stripe call here and I suggest you go ahead and make that change. It should be a fairly small change but correct me if I'm wrong. I could see the discounts being configured in a way where this would execute more than once per user request and that's going to be an issue that I think we can avoid with minimal changes.

src/Core/Billing/Models/DiscountEligibility.cs
```diff
@@ -0,0 +1,13 @@
...
+
+using Bit.Core.Billing.Enums;
+using Bit.Core.Billing.Subscriptions.Entities;
+
+namespace Bit.Core.Billing.Services;
+
+/// <summary>
+/// Pairs a <see cref="SubscriptionDiscount"/> with its per-tier eligibility matrix.
+/// </summary>
+public record DiscountEligibility(
```
kdenney, line 9: Suggest moving to Core/Billing/Models

kdenney (changes requested):
Looking good and getting closer! 👏 A few notes and suggestions.

src/Core/Billing/Services/DiscountAudienceFilters/AllUsersFilter.cs
sbrown-livefront: I can agree with that. Let's remove them for now.

src/Core/Billing/Services/DiscountAudienceFilters/UserHasNoPreviousSubscriptionsFilter.cs
sbrown-livefront: Ha! I over-complicated this in my head and it came out in the code. Great refactor. Thanks!

src/Core/Billing/Services/DiscountAudienceFilters/UserHasNoPreviousSubscriptionsFilter.cs
sbrown-livefront: It should be minimal. I'll get it implemented and see how we like it.

src/Core/Billing/Models/DiscountEligibility.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/76e7d24eb1ddcf2d892b264b765927b9122f9590

src/Core/Billing/Services/DiscountAudienceFilters/UserHasNoPreviousSubscriptionsFilter.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/8fe6c183c37197a675ae1370c4b3cf979d2e180f. Added additional tests also

src/Core/Billing/Services/DiscountAudienceFilters/UserHasNoPreviousSubscriptionsFilter.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/8fe6c183c37197a675ae1370c4b3cf979d2e180f

src/Core/Billing/Services/DiscountAudienceFilters/AllUsersFilter.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/f56bb190ecd23ae1d8b45168881b93190c528495

src/Core/Billing/Models/Api/Response/SubscriptionDiscountResponseModel.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7063/commits/e56ad3a1b41ac96e305be49589451a474eac0af0

src/Core/Billing/Payment/Registrations.cs
sbrown-livefront: Unfortunately since the filters are scoped, we can't make it a singleton due to captive dependency since we still DI the filters. It's possible to create a scope within the factory to side step this, but it could get messy. I did bump it up to scoped though. Addressed in https://github.com/bitwarden/server/pull/7063/commits/83d0337c036cf97aab7b8c25018a07ae3bc2ae8a

PR #7064, opened 2026-02-24
feat(validation): [PM-32626]

Patrick-Pimentel-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32626

## 📔 Objective

  - Standardized KDF/salt validation: Introduced KdfSettingsValidator.ValidateAuthenticationAndUnlockData() (src/Core/Utilities/KdfSettingsValidator.cs) to consolidate validation that authentication and unlock data share consistent KDF settings and salt values. This is now used by RegisterFinishRequestModel.                                                                                                                                                                                                                                                                               

## 📸 Screenshots

<!-- Required for any UI changes; delete if not applicable. Use fixed width images for better display. -->


src/Api/Auth/Models/Request/Accounts/PasswordRequestModel.cs
```diff
@@ -16,4 +18,29 @@ public class PasswordRequestModel : SecretVerificationRequestModel
...
+        // Enforce: if one is provided, both must be provided.
+        if (AuthenticationData != null && UnlockData != null)
+        {
+            foreach (var validationResult in KdfSettingsValidator.ValidateAuthenticationAndUnlockData(
+                AuthenticationData.ToData(), UnlockData.ToData()))
+            {
+                yield return validationResult;
+            }
+        }
+        else
```
enmande: Are we concerned here about all-or-nothing with the `AuthenticationData` and `UnlockData`? Comments and ticket indicate that these fields "aren't required yet", but it looks like we will fail validation aggressively if they aren't there (which I believe is driven by 🎏 `pm-27086-update-authentication-apis-for-input-password` ([clients](https://github.com/bitwarden/clients/blob/d7e2f59387e3e2ba85552be25156c8f7b47f15ed/libs/angular/src/auth/password-management/change-password/default-change-password.service.ts#L143)))? I.e., if that flag is `false`, all requests to this endpoint fail?

test/Api.Test/Auth/Models/Request/Accounts/PasswordRequestModelTests.cs
```diff
@@ -0,0 +1,246 @@
...
+        var result = model.Validate(new ValidationContext(model)).ToList();
+
+        // Assert
+        Assert.Contains(result, r => r.ErrorMessage != null && r.ErrorMessage.Contains(nameof(PasswordRequestModel.AuthenticationData)));
+    }
+
+    #region Base Validation Preserved
+
+    [Fact]
+    public void Validate_BaseValidation_StillRuns()
```
enmande: Related to the other comment:

This reads like it would catch an error in the base validation (failing on a secret requirement), but that throws before a "flag off" client calling the server would run the new validation.

enmande (commented):
I might misunderstand, so 🐻 with me.

The "flag off" scenario outlined below may not be a concern, but wanted to ensure I asked about it. Thanks!

src/Api/Auth/Models/Request/Accounts/PasswordRequestModel.cs
ike-kottlowski: I wanted replicate what KM was doing in the `v2` flow but didn't consider if the request was `v2`. It's worth adding that check since this would fail aggressively for non `v2` users, as you pointed out. I think my confusion stemmed from this model being used for both KDF changes _and_ Password changes. The KDF change requires both, but the password change is in a transitional state.

`v2` in this case means that both `AuthenticationData` and `UnlockData` are present. 

Thanks for finding this.

ike-kottlowski: I believe this has been addressed in https://github.com/bitwarden/server/pull/7064/commits/f1fc3f8ae1852877e44398a9b58e0fb10408bd6a

test/Api.Test/Auth/Models/Request/Accounts/PasswordRequestModelTests.cs
ike-kottlowski: Renamed in https://github.com/bitwarden/server/pull/7064/commits/561112e53622bfcd173c961f5acf58b70313e655

enmande: Non-blocking ⛏️ 
The rename is clearer, thank you, and the arrange is well-documented for what we're testing, so this works for me. What I was hooked onto -- and didn't articulate well, sorry -- was the assertion that the error message contains `must be supplied`, which is common to both scenarios. _However_, the shape of the input should clearly prove this out; lacking `AuthenticationData` _and_ `UnlockData`, the new validation will necessarily "fall through" as written in f1fc3f8. 🎉

enmande (dismissed):
Thank you~!

test/Api.Test/Auth/Models/Request/Accounts/PasswordRequestModelTests.cs
ike-kottlowski: Added more specific assertion to test in 249fb0344e044a6065eaaa98c0d841cb5697b226

enmande: 🥹 This looks great! Thank you~!

enmande (approved):

PR #7072, opened 2026-02-24
create new dockerfile for SeederApi

AmyLGalles:
## 🎟️ Tracking
[BRE-1618](https://bitwarden.atlassian.net/browse/BRE-1618)
[BRE-1685](https://bitwarden.atlassian.net/browse/BRE-1685)

## 📔 Objective
- Create Dockerfile for the SeederApi service
- Update build workflow platforms to exclude arm/v7 for seederApi only since there were build issues with that and it is not necessary for our use case
- Update NginxConfig.hbs to include a generic `include /etc/bitwarden/nginx/extra-locations/*.conf;` hook inside the generated server block. This allows adding location-only conf files into `bwdata/nginx/extra-locations/` that persist across ./bitwarden.sh update/rebuild cycles without opening additional ports or manually patching default.conf on every restart if `generate_nginx_config` is true on the instance. Flow outlined below
  - [util/Nginx/nginx.conf](https://github.com/bitwarden/server/blob/main/util/Nginx/nginx.conf) gets baked into the nginx docker image during build time and contains line `include conf.d/*.conf;`
  - [util/Setup/NginxConfigBuilder.cs](https://github.com/bitwarden/server/blob/main/util/Setup/NginxConfigBuilder.cs) reads [util/Setup/Templates/NginxConfig.hbs](https://github.com/bitwarden/server/blob/main/util/Setup/Templates/NginxConfig.hbs)
  - it overwrites the `default.conf` for nginx on the selfhost during updates, rebuilds etc

[BRE-1618]: https://bitwarden.atlassian.net/browse/BRE-1618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BRE-1685]: https://bitwarden.atlassian.net/browse/BRE-1685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


util/SeederApi/Dockerfile
```diff
@@ -0,0 +1,115 @@
...
+
+ENV SSL_CERT_DIR=/etc/bitwarden/ca-certificates
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false
+ENV ASPNETCORE_ENVIRONMENT=Production
+ENV ASPNETCORE_URLS=http://+:5000
+EXPOSE 5000
+
+# Set up health check wrapper
+# Get the executable and copy it to any path you want
+COPY --from=ghcr.io/alexaka1/distroless-dotnet-healthchecks:1 / /healthcheck
```
pixman20, line 105: This might be overkill since it'll only run in dev and QA, plus I don't know that this has access to anything sensitive, but I think locking down 3rd party dependencies like GH community provided images would be good.

I verified that Renovate will pick up any updates.

```suggestion
COPY --from=ghcr.io/alexaka1/distroless-dotnet-healthchecks:@sha256:d2a71a595020f13a9283f5e1f93c5f92bc2385c1d9b36e128a00d35863995aba / /healthcheck
```

addisonbeck (approved):
no concerns with the platform changes

mandreko-bitwarden (approved):

util/SeederApi/Dockerfile
pixman20: We'll handle in a follow up PR specific to this Dockerfile change

PR #7076, opened 2026-02-25
[PM-32572]Premium to organization upgrade does not set public/private key pair or default collection

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32572

## 📔 Objective

This PR fixes a bug where upgrading from a Premium individual subscription to an organization plan did not properly initialize the organization's
  cryptographic key pair or create a default collection.

  Bug Fix

  Problem: When users upgraded their Premium subscription to an organization plan (Teams/Families/Enterprise), the resulting organization was
  missing:
  1. Public/Private key pair - Required for organization-level encryption operations
  2. Default collection - Users expect a starter collection to organize their vault items

  This left organizations in an incomplete state, potentially blocking users from properly sharing vault items or performing other organization
  operations that depend on these keys.

  Solution:
  - Added PublicKey and EncryptedPrivateKey as required parameters to the upgrade request
  - Organization entity now properly initializes with these keys during the upgrade flow
  - Added optional CollectionName parameter to allow creating a default collection
  - The owner is automatically granted "Can Manage" access to the default collection
  - Collection creation failures are handled gracefully (logged but don't block the upgrade)
  - Command now returns the new organization ID to enable client navigation

  Changes

  - UpgradePremiumToOrganizationRequest: Added PublicKey, EncryptedPrivateKey, and optional CollectionName fields
  - UpgradePremiumToOrganizationCommand: Sets organization keys and creates default collection if requested
  - AccountBillingVNextController: Updated to pass new parameters through
  - Return type changed from None to Guid to return the created organization ID
  - Comprehensive test coverage for all scenarios including failure cases


test/Core.Test/Billing/Premium/Commands/UpgradePremiumToOrganizationCommandTests.cs
```diff
@@ -887,10 +910,12 @@ public async Task Run_CreatesOrganizationWithCorrectSettings(User user)
...
-        var result = await _command.Run(user, "My Organization", "encrypted-key", PlanType.TeamsAnnually, CreateTestBillingAddress());
+        var result = await _command.Run(user, "My Organization", "encrypted-key", "public-key", "encrypted-private-key", null, PlanType.TeamsAnnually, CreateTestBillingAddress());
 
         // Assert
         Assert.True(result.IsT0);
+        var organizationId = result.AsT0;
+        Assert.NotEqual(Guid.Empty, organizationId);
 
         await _organizationRepository.Received(1).CreateAsync(
             Arg.Is<Organization>(org =>
```
sbrown-livefront, line 853: ⛏ Should the public and private keys be checked here also?

sbrown-livefront (approved):
Looks good! Small non-blocking comment on a test

test/Core.Test/Billing/Premium/Commands/UpgradePremiumToOrganizationCommandTests.cs
sbrown-livefront: ⛏️ I'm not sure this test is named correctly since the removal of a lot of the metadata changes nor is it testing anything we haven't tested I believe. Can we remove it?

PR #7083, opened 2026-02-25
[PM-31884] New send policy to restrict access controls

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31884


harr1424:
Closing in favor of preferred approach

PR #7086, opened 2026-02-26
[PM-32774] Add Melio to global equivalent domains

fluctus:
## 🎟️ Tracking

Community contribution — no Jira ticket.

## 📔 Objective

Add `melio.com` and `meliopayments.com` as global equivalent domains so that Bitwarden users with credentials saved under either domain are not affected by Melio's domain switch (`melio.com` now redirects to `meliopayments.com`).

Changes:
- Added `Melio = 91` to the `GlobalEquivalentDomainsType` enum
- Added the corresponding domain mapping in `StaticStore.cs`

### DIFF
--- a/src/Core/Enums/GlobalEquivalentDomainsType.cs
+++ b/src/Core/Enums/GlobalEquivalentDomainsType.cs
@@ -93,4 +93,5 @@ public enum GlobalEquivalentDomainsType : byte
     TakeawayEU = 88,
     Atlassian = 89,
     Pinterest = 90,
+    Melio = 91,
 }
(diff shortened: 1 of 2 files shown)


fluctus:
Hi @jprusik - tagging you as I've seen you review similar global equivalent domain additions. Would appreciate your review when you get a chance. Thanks!

jprusik:
> Hi @jprusik - tagging you as I've seen you review similar global equivalent domain additions. Would appreciate your review when you get a chance. Thanks!

Hey @fluctus , we're currently in the process of reducing the footprint of our equivalent domain defaults (the equivalent domains feature remains for users who wish to customize it). Excepting some particularly impactful scenarios, we recommend users create and update their login ciphers' domains directly, as that maintains a tighter security relationship between login portals and the ciphers they belong to.

As this case does not represent an extraordinary need, we will not be including it at this time.

PR #7088, opened 2026-02-26
disable claude attribution

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32794

## 📔 Objective

Disable attribution messages output by Claude when it commits or writes a PR.


.claude/settings.json
```diff
@@ -6,5 +6,9 @@
         "repo": "bitwarden/ai-plugins"
       }
     }
+  },
+  "attribution": {
+    "commit": "",
+    "pr": ""
   }
```
SaintPatrck, line 17: ⛏️ Can we move this above `extraKnownMarketplaces` to maintain alphabetic ordering?

PR #7089, opened 2026-02-26
[PM-32796] Fix bulk reinvite 503 timeout by enqueuing invite emails via Azure Queue

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32796

## 📔 Objective

When the `pm-28396-update-join-organization-email-template` flag is enabled, invite emails are sent synchronously through SendGrid one at a time, causing bulk reinvites to time out.

This PR adds an `EnqueueEmail` method to `IMailer` that pre-renders emails and enqueues them to the Azure Storage Queue for background delivery, restoring the same non-blocking behavior the legacy mail path uses.

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationConfirmation/SendOrganizationConfirmationCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationConfirmation/SendOrganizationConfirmationCommand.cs
@@ -59,7 +59,7 @@ public class SendOrganizationConfirmationCommand(IMailer mailer, GlobalSettings 
             }
         };
 
-        await mailer.SendEmail(mail);
+        await mailer.EnqueueEmail(mail);
     }
 
     private async Task SendFamilyFreeConfirmEmailsAsync(List<string> userEmailsList, string organizationName, bool accessSecretsManager)
@@ -78,7 +78,7 @@ public class SendOrganizationConfirmationCommand(IMailer mailer, GlobalSettings 
             }
         };
 
-        await mailer.SendEmail(mail);
+        await mailer.EnqueueEmail(mail);
     }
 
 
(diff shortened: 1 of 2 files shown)


src/Core/Platform/Mail/Mailer/Mailer.cs
```diff
@@ -29,4 +33,34 @@ public async Task SendEmail<T>(BaseMail<T> message) where T : BaseMailView
 
         await mailDeliveryService.SendEmailAsync(mailMessage);
     }
+
+    public async Task EnqueueEmail<T>(BaseMail<T> message) where T : BaseMailView
+    {
+        var content = await renderer.RenderAsync(message.View);
+
```
JimmyVo16, line 40: It looks like the non-queue method has a check for the suppress list. Do we need to add that here?

```C#
  if (message.IgnoreSuppressList)
  {
      metadata.Add("SendGridBypassListManagement", true);
  }
```

src/Core/Platform/Mail/Mailer/Mailer.cs
```diff
@@ -29,4 +33,42 @@ public async Task SendEmail<T>(BaseMail<T> message) where T : BaseMailView
...
+        var queueMessage = new MailQueueMessage
+        {
+            Subject = message.Subject,
+            ToEmails = message.ToEmails,
+            Category = message.Category,
```
JimmyVo16, line 45: I saw a potential null warning for `Category` in my IDE. I looked into it, and the worst-case scenario is that we’re sending an [empty tag](https://github.com/bitwarden/server/blob/3668a445e5e9e781244697c29fc3734a53841cc4/src/Core/Platform/Mail/Delivery/SendGridMailDeliveryService.cs#L67) here. I think this is fine moving forward, given the situation. It’s something we can reassess later.

src/Core/Models/Mail/IMailQueueMessage.cs
```diff
@@ -8,4 +8,7 @@ public interface IMailQueueMessage
     string Category { get; set; }
     string TemplateName { get; set; }
     object Model { get; set; }
+    string? HtmlContent { get; set; }
+    string? TextContent { get; set; }
+    bool IgnoreSuppressList { get; set; }
```
Hinton, line 13: I think putting metadata here would be nicer since it's more generic and avoids splitting the knowledge how `IgnoreSupressList` should be handled.

Hinton (commented):
I can't say this is a particularly nice flow, IMailer -> Enqueued -> old handlebar mailer.

Long term I think my preference would be for every mail to be enqueued which would abstract it away from `IMailer`. It should also not go through the old `HandlebarsMailService` since there is nothing to render.

Two things to consider relevant for this PR:

How are we handling self hosted if they attempt to invite many users? They will most likely run into the same issue, unless we only allow the larger bulk invite in cloud.

This adds a deploy requirement to make sure our mail processors are off before deploy since you're going to get errors if an old mail queue processor receives htmlcontent.

PR #6768, opened 2025-12-22
[PM-29611] Decouple License from Subscription Response

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29611

## 📔 Objective

Refactor the subscription response model to decouple license information from subscription data, implementing proper separation of concerns between billing and licensing functionality.

**Code changes**

  **Backend Changes**

  **New Files**:
 ` - src/Api/Models/Response/LicenseResponseModel.cs` - New response model for license data
`  - test/Api.Test/Billing/Controllers/VNext/AccountBillingVNextControllerTests.cs` - Unit tests for new license endpoint

  **Modified Files:**
 ` - src/Api/Models/Response/SubscriptionResponseModel.cs:`
    - Removed License property
    - Removed Expiration property
    - Removed license-related constructor parameters
    - Maintained all subscription and discount functionality
`  - src/Api/Billing/Controllers/VNext/AccountBillingVNextController.cs:`
    - Added GET /account/billing/vnext/license endpoint
    - Added IUserService and ILicensingService dependencies
`  - src/Api/Billing/Controllers/AccountsController.cs:`
    - Updated GetSubscriptionAsync to remove license generation
    - Removed ILicensingService dependency

  **Test Updates:**
  - Updated SubscriptionResponseModelTests.cs - Fixed 14 tests to use new constructor signatures
  - Updated AccountsControllerTests.cs - Removed licensing service mock

  **Breaking Changes**

  **⚠️ API Breaking Change:** The SubscriptionResponseModel no longer includes License and Expiration properties.

  **Migration Path:**
  - Frontend clients should call GET /account/billing/vnext/license separately to retrieve license information
  - The subscription endpoint continues to return all subscription and billing data

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
[...]

PR #7090, opened 2026-02-26
[PM 32766][Defect] Regression: Unable to upload Premium License to Self Hosted instances

cyprain-okeke:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32766

## 📔 Objective

This PR fixes a regression (PM-32766) where Premium License uploads to self-hosted instances were failing.

  Bug Description

  The server was unable to parse user license files that were wrapped in an envelope structure containing license, expiration, and object fields.
  Previously, the server only supported the flat license format.

  Solution

  - Added a new ReadUserLicenseFromBody helper method in ApiHelpers.cs that supports both formats:
    - Wrapped format: { "license": {...}, "expiration": "...", "object": "license" }
    - Flat format: Direct UserLicense object (backwards compatibility)
  - Updated AccountsController.cs and SelfHostedAccountBillingVNextController.cs to use the new helper method
  - Added unit tests to verify both wrapped and flat license formats are correctly parsed

  Testing

  Unit tests added:
  - ReadUserLicenseFromBody_WrappedLicense_Succeeds - Validates wrapped format parsing
  - ReadUserLicenseFromBody_FlatLicense_Succeeds - Validates backwards compatibility with flat format
## 📸 Screenshots

image

### DIFF
--- a/src/Api/Utilities/ApiHelpers.cs
+++ b/src/Api/Utilities/ApiHelpers.cs
@@ -4,7 +4,6 @@
 using System.Text.Json;
 using Azure.Messaging.EventGrid;
 using Azure.Messaging.EventGrid.SystemEvents;
-using Bit.Core.Billing.Models.Business;
 using Bit.Core.Exceptions;
 using Bit.Core.Utilities;
 using Microsoft.AspNetCore.Mvc;
@@ -13,13 +12,6 @@ namespace Bit.Api.Utilities;
 
 public static class ApiHelpers
 {
-    private class UserLicenseFileWrapper
-    {
-        public UserLicense License { get; set; }
-        public DateTime? Expiration { get; set; }
-        public string Object { get; set; }
-    }
-
     public static string EventGridKey { get; set; }
     public async static Task<T> ReadJsonFileFromBody<T>(HttpContext httpContext, IFormFile file, long maxSize = 51200)
     {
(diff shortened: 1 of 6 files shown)


src/Api/Utilities/ApiHelpers.cs
```diff
@@ -12,6 +13,13 @@ namespace Bit.Api.Utilities;
 
 public static class ApiHelpers
 {
+    private class UserLicenseFileWrapper
+    {
+        public UserLicense License { get; set; }
+        public DateTime? Expiration { get; set; }
+        public string Object { get; set; }
```
kdenney: Why is this `string Object` property needed? Is the form data from the client actually sent in this format?

kdenney (approved):
The code looks ok but I don't understand why it fixes the regression bug. 😅  I have one non-blocking question so I'll go ahead and approve.

amorask-bitwarden (changes requested):
As far as I can tell, the `LicenseResponseModel` that we introduced is not used any differently than just returning the `UserLicense` as JSON to the client.

I suggest reverting the portion of your changes in [this PR](https://github.com/bitwarden/server/pull/6768) where you introduced the `LicenseResponseModel` and just instead return the `UserLicense` as JSON like we have been, but from our new endpoint.

kdenney (approved):

PR #7092, opened 2026-02-26
[PM-28531] GetMetrics added to IOrganizationReportRepository

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28531

## 📔 Objective

1. Added GetMetricsAsync method to the IOrganizationReportRepository 
2. Created a view for Dapper called dbo.OrganizationReportMetricsView
3. Create a proc that reads the view dbo.OrganizationReport_GetMetrics
4. Created unit tests to test all database
5. 
## 📸 Screenshots

None at this time

### DIFF
--- a/src/Infrastructure.Dapper/Dirt/OrganizationReportRepository.cs
+++ b/src/Infrastructure.Dapper/Dirt/OrganizationReportRepository.cs
@@ -211,11 +211,11 @@ public class OrganizationReportRepository : Repository<OrganizationReport, Guid>
         var parameters = new
         {
             OrganizationId = orgId,
-            minDate = until
+            MinRevisionDate = until
         };
 
         var results = await connection.QueryAsync<OrganizationReportMetricsData>(
-            $"[{Schema}].[OrganizationReport_GetMetrics]",
+            $"[{Schema}].[OrganizationReport_ReadByOrganizationIdAndRevisionDate]",
             parameters,
             commandType: CommandType.StoredProcedure);
 


src/Sql/dbo/Dirt/Views/OrganizationReportMetricsView.sql
```diff
@@ -0,0 +1,17 @@
+CREATE VIEW [dbo].[OrganizationReportMetricsView] AS
```
mkincaid-bw: This view is not needed. You can select the data you need from `OrganizationReportView`

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetMetrics.sql
```diff
@@ -0,0 +1,12 @@
+CREATE PROCEDURE [dbo].[OrganizationReport_GetMetrics]
```
mkincaid-bw: This proc should be named something like `OrganizationReportMetrics_ReadByOrganizationIdAndRevisionDate`.
https://contributing.bitwarden.com/contributing/code-style/sql#stored-procedures

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetMetrics.sql
```diff
@@ -0,0 +1,12 @@
...
+    @OrganizationId UNIQUEIDENTIFIER,
+    @minDate DATETIME2(7)
```
mkincaid-bw: This parameter should start with a capital letter and perhaps be named something like `@MinimumRevisionDate`

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetMetrics.sql
```diff
@@ -0,0 +1,12 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    SELECT *
+    FROM dbo.OrganizationReportMetricsView
```
mkincaid-bw: Object names should have square brackets, ie `[dbo].[OrganizationReportMetricsView]`
https://contributing.bitwarden.com/contributing/code-style/sql#general-standards

src/Core/Dirt/Models/Data/OrganizationReportMetricsData.cs
```diff
@@ -5,6 +5,7 @@ namespace Bit.Core.Dirt.Reports.Models.Data;
 public class OrganizationReportMetricsData
```
mkincaid-bw, line 5: This object should be named `OrganizationReportMetrics`

prograhamming:
These metric columns were created because product wanted them for internal dashboards. I don't think we should be using them. I know they are there and it makes sense to use them. However, I think we should be sending a response with the unencrypted data.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetMetrics.sql
voommen-livefront: Thanks for the recommendation for the name. I have applied the name like you suggested.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetMetrics.sql
voommen-livefront: Good eye -thank you!

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetMetrics.sql
voommen-livefront: Updated all object names to be box brackets.

src/Core/Dirt/Models/Data/OrganizationReportMetricsData.cs
voommen-livefront: This will cause a refactor in areas of the code that I didn't work on. Also, a team mate is working on these files and it would cause issues for him.

voommen-livefront:
The ask in the story has changed. It no longer requires a JSON response of actual metrics. The endpoint needs to send out the encrypted summary data for the requested period. Closing this PR as the story is being updated.

PR #7093, opened 2026-02-26
[PM-32764] Archive Sync Notifications Double Sync Solve

Jingo88:
## 🎟️ Tracking

[PM-32764](https://bitwarden.atlassian.net/browse/PM-32764)

## 📔 Objective

The push notification being called from ArchiveCiphersCommand and UnarchiveCiphersCommand would call a full sync on the browser giving users an unnecessary empty vault flash after archiving a cipher. 

Adding a second parameter to exclude the sync on the client that called it.

Screen Recording below includes network tab to show no extra calls after archiving and unarchiving

## 📸 Screen Recording

https://github.com/user-attachments/assets/6401dd25-15fc-410f-9bb0-4100a50cf808

[PM-32764]: https://bitwarden.atlassian.net/browse/PM-32764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Core.Test/Vault/Commands/ArchiveCiphersCommandTest.cs
+++ b/test/Core.Test/Vault/Commands/ArchiveCiphersCommandTest.cs
@@ -43,7 +43,7 @@ public class ArchiveCiphersCommandTest
                 : ids.All(id => cipherList.Contains(cipher))),
             user.Id);
         await sutProvider.GetDependency<IPushNotificationService>().Received(pushNotificationsCalls)
-            .PushSyncCiphersAsync(user.Id);
+            .PushSyncCiphersAsync(user.Id, true);
     }
 
     [Theory]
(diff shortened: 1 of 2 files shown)


src/Core/Platform/Push/IPushNotificationService.cs
```diff
@@ -6,6 +6,7 @@
 using Bit.Core.Tools.Entities;
 using Bit.Core.Vault.Entities;
 using Microsoft.Extensions.Logging;
+using StackExchange.Redis;
```
nick-livefront: ❓ This doesn't look related?

dani-garcia (approved):

PR #7094, opened 2026-02-26
[BW-138] 2FA Account Recovery

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/BW-138

## 📔 Objective

Org admins can recover 2FA on a member's account.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(dotnet build:*)"
+    ]
+  }
+}


plans/2fa-account-recovery.md
```diff
@@ -0,0 +1,435 @@
...
+    IOrganizationRepository organizationRepository,
+    IPolicyQuery policyQuery,
+    IFeatureService featureService,
+    IUserRepository userRepository) : IAdminRecoverAccountValidator
+{
+    public async Task<ValidationResult<RecoverAccountRequest>> ValidateAsync(RecoverAccountRequest request)
+    {
+        // At least one action must be requested
+        if (!request.ResetMasterPassword && !request.ResetTwoFactor)
+            return Invalid(request, new NoActionRequestedError());
```
eliykat, line 157: code style: to my knowledge, we don't use single line if statements.

This validator design is otherwise perfect though, thanks.

plans/2fa-account-recovery.md
```diff
@@ -0,0 +1,435 @@
...
+
+// Map to internal request record
+var commandRequest = model.ToCommandRequest(orgId, targetOrganizationUser);
+
+// Validate
+var validationResult = await _adminRecoverAccountValidator.ValidateAsync(commandRequest);
+if (validationResult.IsError)
+{
+    return Handle(validationResult.AsError);  // maps BadRequestError → 400, NotFoundError → 404
+}
```
eliykat, line 326: We call validation from inside the command, so that command inputs are always guaranteed to be validated. I can see an argument for doing it from the controller so that they're fully decoupled, but it's not our current practice.

plans/2fa-account-recovery.md
```diff
@@ -0,0 +1,435 @@
...
+    foreach (var error in result.Errors)
+    {
+        ModelState.AddModelError(string.Empty, error.Description);
+    }
+    return Handle(new IdentityError(/* ... */));  // existing error handling pattern
+}
+return TypedResults.NoContent();
+```
+
+Add a `Handle(Error)` helper or use existing `Handle(CommandResult)` — need to check how other controllers handle `ValidationResult` errors. The `BaseAdminConsoleController.Handle(CommandResult)` handles `Error` types, so we can map:
```
eliykat: We would use Handle(CommandResult) as this suggests. However, this command returns an IdentityResult for some reason rather than a CommandResult. You can either update the command to return a CommandResult, or - to limit scope - just throw your errors inside the command (including validation failures).

kspearrin:
New proposal seems to have simplified things a bit

plans/2fa-account-recovery.md
```diff
@@ -0,0 +1,312 @@
...
+## Context
+
+The Account Recovery feature currently only supports resetting a user's master password via `PUT /organizations/{orgId}/users/{id}/reset-password`. We are extending this existing endpoint and command to also support resetting (clearing) a member's 2FA methods. Both actions can be performed independently or together in a single request.
+
+## Design Decisions
+
+- **Rename endpoint** `PUT reset-password` → `PUT recover-account` (keep old route as alias for backward compat); add `ResetMasterPassword` and `ResetTwoFactor` booleans to the request
+- **Extend existing command** `AdminRecoverAccountCommand` — add 2FA reset as a conditional action alongside password reset
+- **Internal request record** — command accepts a `RecoverAccountRequest` record instead of primitive parameters; API model maps via `ToCommandRequest()`
+- **Validation inside the command** — all validation stays in `AdminRecoverAccountCommand` (current practice), not a separate validator class, so inputs are always guaranteed validated
```
eliykat: I think Claude misinterpreted me here, I meant just call the validator from inside the command, not move the validation logic itself back inside the command. However, it's probably struggling with competing patterns.

I'm OK to go ahead with the current plan - we can return to these command refactors later, separate to the feature work, and with better close-to-code documentation for Claude.

eliykat (dismissed):

plans/2fa-account-recovery.md
kspearrin: I'll just revert to the previous plan and we can iterate on the code output if necessary.

eliykat (approved):
That also works, thanks!

kspearrin:
Major git merge failure here. Apologies.

PR #7095, opened 2026-02-26
[PM-32594] Add authorization to admin-initiated sponsorship endpoints

cturnbull-bitwarden:
## 🎟️ Tracking

<!-- Paste the link to the Jira or GitHub issue or otherwise describe / point to where this change is coming from. -->
https://bitwarden.atlassian.net/browse/PM-32594

## 📔 Objective

<!-- Describe what the purpose of this PR is, for example what bug you're fixing or new feature you're adding. -->
Adds authorization checks to the `AdminInitiatedRevokeSponsorshipAsync` and `ResendSponsorshipOffer` endpoints in both the cloud and self-hosted sponsorship controllers. Previously, these endpoints only required `[Authorize("Application")]`, meaning any authenticated user could call them regardless of their relationship to the 
sponsoring organization.
                                                                                                                                                                                                                                                                                                                                         
Now uses `[Authorize<ManageUsersRequirement>]` to enforce that the caller is an Owner, Admin, or has the ManageUsers permission in the sponsoring organization — consistent with the existing `GetSponsoredOrganizations` endpoint and `CreateSponsorshipCommand` validation.                                                          

Route parameters on the affected endpoints were renamed from `{sponsoringOrgId}` to `{organizationId}` to work with the existing authorization handler infrastructure. The C# parameter names remain unchanged.


src/Api/Controllers/SelfHosted/SelfHostedOrganizationSponsorshipsController.cs
```diff
@@ -88,8 +89,9 @@ public async Task PostRevokeSponsorship(Guid sponsoringOrgId)
         await RevokeSponsorship(sponsoringOrgId);
     }
 
-    [HttpDelete("{sponsoringOrgId}/{sponsoredFriendlyName}/revoke")]
-    public async Task AdminInitiatedRevokeSponsorshipAsync(Guid sponsoringOrgId, string sponsoredFriendlyName)
+    [Authorize<ManageUsersRequirement>]
```
kdenney, line 92: I was looking over the recommended remediation on the vulnerability ticket and it mentioned to use the same authorization pattern that is currently used elsewhere in this controller. Specifically this bit:

```
if (!await _currentContext.OrganizationOwner(sponsoringOrg.Id) && !await _currentContext.OrganizationAdmin(sponsoringOrg.Id) && !organization.Permissions.ManageUsers)
        {
        }
```

I do prefer the simplicity of using the authorization attribute instead. However, I wanted to double check because I looked into the code for that attribute and it is very slightly different. It is doing this in `BasePermissionRequirement.cs`:

```
    => organizationClaims switch
    {
        { Type: OrganizationUserType.Owner } => true,
        { Type: OrganizationUserType.Admin } => true,
        { Type: OrganizationUserType.Custom } when permissionPicker(organizationClaims.Permissions) => true,
        _ => await isProviderUserForOrg()
    };
```

The main difference being that in the sponsorships controller manual authorization code, if the user has an organization user type of `User` or `Custom`, and they have the `ManageUsers` permission, they are granted access. But using the authorization attribute results in the permission only getting checked if the user type is `Custom` and a type of `User` will only be granted access if they are a provider user for the org.

So the question is... can you even have the `ManageUsers` permission if your org user type is `User`? Probably not, but I didn't know for sure how to go find that out so I wanted to ask. If not, then using the attribute is good with me. 👍

connerbw: Great question. The answer is no - a member with the `User` role can never have `ManageUsers = true` at runtime, so the behavior ends up being the same. A couple of reasons:

- **JWT claim loading**: When building the `CurrentContextOrganization` list from claims, members with the `User` role are created [without loading permissions](https://github.com/bitwarden/server/blob/main/src/Core/Context/CurrentContext.cs#L209-L217). Only `Custom` members [call `SetOrganizationPermissionsFromClaims()`](https://github.com/bitwarden/server/blob/main/src/Core/Context/CurrentContext.cs#L220-L229). The `Permissions` property [defaults to `new Permissions()`](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Context/CurrentContextOrganization.cs#L26) where every boolean is `false`.
- **Invite/update flows**: `SetPermissions()` is [only called when `Type == OrganizationUserType.Custom`](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L619-L622). Permissions passed for other member types are silently ignored.

The inline check `!organization.Permissions.ManageUsers` in `GetSponsoredOrganizations` works by coincidence - it's just evaluating the default `false` for members with the `User` role, not because they could actually have the permission.

[`BasePermissionRequirement`](https://github.com/bitwarden/server/blob/main/src/Api/AdminConsole/Authorization/Requirements/BasePermissionRequirement.cs#L17-L23) is the more intentional design here since it explicitly scopes the permission check to `Custom` members. It also has a minor advantage over the inline code - it allows provider users as a fallback, which the existing inline check in `GetSponsoredOrganizations` is technically missing.

kdenney: Awesome thanks for the detailed reply.

kdenney (approved):

PR #7100, opened 2026-02-27
[PM-32597] - create short-lived signed attachment URL for self-hosted instances

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32597

## 📔 Objective

In self-hosted Bitwarden, cipher attachment download URLs were predictable (`/attachments/{cipherId}/{attachmentId}`) and served by a static file server with no authentication. An attacker with knowledge of a cipher ID and attachment ID could download encrypted attachment files without authorization.

### Changes

**Signed download URLs for local storage** — When `FileUploadType` is `Direct` (local/self-hosted), [`CipherService.GetAttachmentDownloadDataAsync`](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/Implementations/CipherService.cs) now generates a time-limited, cryptographically signed token using ASP.NET Core Data Protection instead of returning a bare static file path. Tokens expire after 1 minute, matching the short-lived nature of Azure SAS URLs used in cloud deployments.

**New download endpoint** — Added `GET /ciphers/attachment/download?token=...` (`[AllowAnonymous]`) to [`CiphersController`](https://github.com/bitwarden/server/blob/main/src/Api/Vault/Controllers/CiphersController.cs). This endpoint validates the signed token, resolves the cipher and attachment, and streams the file. The token itself is the authorization — it can only be obtained through the existing authenticated `GET /ciphers/{id}/attachment/{attachmentId}` endpoint.

**`GetAttachmentReadStreamAsync`** — Added to [`IAttachmentStorageService`](https://github.com/bitwarden/server/blob/main/src/Core/Services/IAttachmentStorageService.cs) and all implementations. [`LocalAttachmentStorageService`](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/Implementations/LocalAttachmentStorageService.cs) returns a `FileStream`; Azure and Noop implementations return `null` (Azure uses SAS URLs directly).

**Null check hardening** — `GetAttachmentData` now explicitly throws `NotFoundException` when the cipher is not found.

### Security considerations

- No change to the client-facing API contract — `GET /ciphers/{id}/attachment/{attachmentId}` still requires `[Authorize("Application")]`
- Token payload is `{cipherId}|{attachmentId}`, encrypted and time-limited via `ITimeLimitedDataProtector`
- Expired/invalid/malformed tokens all return 404 (no information leakage)
- Azure (cloud) deployments are unaffected — they already use time-limited SAS URLs
- The static file server in [`util/Server`](https://github.com/bitwarden/server/blob/main/util/Server/Startup.cs) should separately be configured to stop serving the `/attachments` directory, but this change ensures that even if it does, the API no longer returns those predictable URLs

### Files changed

| File | Change |
|------|--------|
| [CipherService.cs](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/Implementations/CipherService.cs) | Signed URL generation, `ParseAttachmentDownloadToken`, `CreateAttachmentDownloadProtector` |
| [ICipherService.cs](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/ICipherService.cs) | Added `CreateAttachmentDownloadProtector()` |
| [IAttachmentStorageService.cs](https://github.com/bitwarden/server/blob/main/src/Core/Services/IAttachmentStorageService.cs) | Added `GetAttachmentReadStreamAsync` |
| [LocalAttachmentStorageService.cs](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/Implementations/LocalAttachmentStorageService.cs) | Implemented `GetAttachmentReadStreamAsync` |
| [AzureAttachmentStorageService.cs](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/Implementations/AzureAttachmentStorageService.cs) | Implemented `GetAttachmentReadStreamAsync` (returns null) |
| [NoopAttachmentStorageService.cs](https://github.com/bitwarden/server/blob/main/src/Core/Vault/Services/NoopImplementations/NoopAttachmentStorageService.cs) | Implemented `GetAttachmentReadStreamAsync` (returns null) |
| [CiphersController.cs](https://github.com/bitwarden/server/blob/main/src/Api/Vault/Controllers/CiphersController.cs) | Added `DownloadAttachmentAsync` endpoint, null check in `GetAttachmentData` |
| [CiphersControllerTests.cs](https://github.com/bitwarden/server/blob/main/test/Api.Test/Vault/Controllers/CiphersControllerTests.cs) | 6 new tests |
| [CipherServiceTests.cs](https://github.com/bitwarden/server/blob/main/test/Core.Test/Vault/Services/CipherServiceTests.cs) | 8 new tests |


Jingo88 (dismissed):

src/Core/Vault/Services/Implementations/AzureAttachmentStorageService.cs
```diff
@@ -193,6 +193,12 @@ public async Task DeleteAttachmentsForUserAsync(Guid userId)
         await InitAsync(_defaultContainerName);
     }
 
+    public Task<Stream> GetAttachmentReadStreamAsync(Cipher cipher, CipherAttachment.MetaData attachmentData)
```
shane-melton: ⛏️ The return type here should be nullable e.g. `Task<Stream?>` to match the interface and same for the `Noop` implementation.

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -412,17 +419,61 @@ public async Task<AttachmentResponseData> GetAttachmentDownloadDataAsync(Cipher
             throw new NotFoundException();
         }
 
+        var url = await _attachmentStorageService.GetAttachmentDownloadUrlAsync(cipher, data);
+
+        // For local (self-hosted) storage, generate a time-limited signed download URL
```
shane-melton: ❓ 💭  I would have expected this to live inside of the `LocalAttachmentStorageService.GetAttachmentDownloadUrlAsync()` implementation. That avoids needing to check the `FileUploadType == Direct` in the cipher service.

Any reason to keep it here?  It would also improve the separation of concerns - e.g. `IAttachmentStorageService` is responsible for interacting with cipher URLs and protecting tokens.

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -412,17 +419,61 @@ public async Task<AttachmentResponseData> GetAttachmentDownloadDataAsync(Cipher
...
     }
 
+    public ITimeLimitedDataProtector CreateAttachmentDownloadProtector()
+    {
+        return _dataProtectionProvider
+            .CreateProtector(AttachmentDownloadProtectorPurpose)
+            .ToTimeLimitedDataProtector();
+    }
+
+    public static (Guid cipherId, string attachmentId) ParseAttachmentDownloadToken(
```
shane-melton: 🎨 This could also like somewhere in the `IAttachmentStorageService` since its unique to attachments.

src/Core/Vault/Services/Implementations/AzureAttachmentStorageService.cs
jaasen-livefront: @shane-melton That was the original plan but since this file implements `#nullable disable` it forces a `#nullable enable` and `#nullable disable` before and after this task. Is that acceptable?

shane-melton: Ah, missed that at the top of the file. I'm good to leave it unless its trivial enough to make the file null safe while we're here.

jaasen-livefront: Easy fix to just remove `nullable disable` from the file. https://github.com/bitwarden/server/pull/7100/commits/d19cda2068cd6386d6c96ff7bcb1457a91b0fc1e

src/Core/Vault/Services/Implementations/CipherService.cs
jaasen-livefront: fixed https://github.com/bitwarden/server/pull/7100/commits/d19cda2068cd6386d6c96ff7bcb1457a91b0fc1e

src/Core/Vault/Services/Implementations/CipherService.cs
jaasen-livefront: fixed https://github.com/bitwarden/server/pull/7100/commits/d19cda2068cd6386d6c96ff7bcb1457a91b0fc1e

src/Core/Vault/Services/Implementations/LocalAttachmentStorageService.cs
```diff
@@ -11,21 +13,59 @@ public class LocalAttachmentStorageService : IAttachmentStorageService
...
     public FileUploadType FileUploadType => FileUploadType.Direct;
 
     public LocalAttachmentStorageService(
-        IGlobalSettings globalSettings)
+        IGlobalSettings globalSettings,
+        IDataProtectionProvider dataProtectionProvider)
     {
         _baseDirPath = globalSettings.Attachment.BaseDirectory;
         _baseTempDirPath = $"{_baseDirPath}/temp";
         _baseAttachmentUrl = globalSettings.Attachment.BaseUrl;
```
shane-melton: ⛏️ (non-blocking) `_baseAttachmentUrl` can be removed now

test/Api.Test/Vault/Controllers/CiphersControllerTests.cs
shane-melton, line 1: 👍 Thanks for the tests!

shane-melton, line 1: 🎨 (non-blocking) Could we add one more test for the happy path of `DownloadAttachmentAsync`. e.g. Verify a valid token ends up with a `File()` response with the expected stream and file name.

shane-melton (approved):
Looks good! Just two minor, non-blocking nits. 

We should run this by QA before merge since there is no feature flag and its a significant change in underlying behavior.

test/Api.Test/Vault/Controllers/CiphersControllerTests.cs
jaasen-livefront: Done! https://github.com/bitwarden/server/pull/7100/commits/d33ad0b0295a9523dcfbd808d866f901b1c55f2a

src/Core/Vault/Services/Implementations/LocalAttachmentStorageService.cs
jaasen-livefront: Nice catch, removed. https://github.com/bitwarden/server/pull/7100/commits/d33ad0b0295a9523dcfbd808d866f901b1c55f2a

PR #7101, opened 2026-02-27
[PM-32366] Update UseMyItems to use dedicated plan feature

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32366

## 📔 Objective

Billing Team have now added a dedicated `HasMyItems` plan feature, which can be used to set the `UseMyItems` organization ability. We were previously using `UsePolicies` as a proxy.


eliykat:
@sbrown-livefront / @amorask-bitwarden Can I please get your help with the test failures here? They are all on integration tests that are (or should be) consuming this new plan value from the Pricing Service. The failures indicate that the `UseMyItems` value is not being set correctly. The tests all pass locally but not in CI. Is the Pricing Service connection configured differently in CI maybe?

amorask-bitwarden:
> @sbrown-livefront / @amorask-bitwarden Can I please get your help with the test failures here? They are all on integration tests that are (or should be) consuming this new plan value from the Pricing Service. The failures indicate that the `UseMyItems` value is not being set correctly. The tests all pass locally but not in CI. Is the Pricing Service connection configured differently in CI maybe?

@eliykat The wrong branch was deployed to Mothership QA - I've re-deployed and re-ran your tests. You're good to go.

PR #7113, opened 2026-03-02
[PM-31885] Consolidate all Send policies to a single policy

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31885

## 📔 Objective

Consolidate all Send policies, including new policies, to a single policy.

Please see [TBD](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2581823624/Sends+Org+Policies+MS+1) for full context. 

This PR introduces a new `SendControls` policy, that is intended to absorb the existing Send policies `DisableSend` and `DisableHideEmail` and serve as a container for upcoming Send related policies as part of [PM-31883](https://bitwarden.atlassian.net/browse/PM-31883). 

Of note in this PR are multiple `PolicyValidators` used to ensure policy statuses are synced between the old and new implementation across feature flag status changes or potential rollbacks.

[PM-31883]: https://bitwarden.atlassian.net/browse/PM-31883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


harr1424:
@claude please review

util/Migrator/DbScripts_transition/2026-02-28_00_CreateSendControlsPolicies.sql
```diff
@@ -0,0 +1,55 @@
...
+        SELECT
+            COALESCE(ds.OrganizationId, so.OrganizationId) AS OrganizationId,
+            ds.Enabled  AS DisableSendEnabled,
+            so.Enabled  AS SendOptionsEnabled,
+            so.Data     AS SendOptionsData
+        FROM      [dbo].[Policy] ds
+        FULL OUTER JOIN [dbo].[Policy] so
+            ON  ds.OrganizationId = so.OrganizationId
+            AND so.Type = @SendOptionsType
+        WHERE (ds.Type = @DisableSendType OR so.Type = @SendOptionsType)
```
mkincaid-bw: I tested this script against a backup copy of prod and there is an issue. The `[IX_Policy_OrganizationId_Type]` index in the Policy table is a `UNIQUE` index. This subquery returns duplicates, which causes the insert to fail. The suggested replacement query filters out duplicates but you should verify the logic is working as expected.

```suggestion
        SELECT DISTINCT
             COALESCE(ds.OrganizationId, so.OrganizationId) AS OrganizationId,
             ds.Enabled AS DisableSendEnabled,
             so.Enabled AS SendOptionsEnabled,
             so.Data AS SendOptionsData
        FROM 
            [dbo].[Policy] ds
        LEFT JOIN
            [dbo].[Policy] so 
        ON ds.OrganizationId = so.OrganizationId 
            AND so.Type = @SendOptionsType
        WHERE
            ds.Type = @DisableSendType
        UNION
        SELECT
            so.OrganizationId, 
            NULL, 
            so.Enabled, 
            so.Data
        FROM 
            [dbo].[Policy] so
        WHERE 
            so.Type = @SendOptionsType
            AND NOT EXISTS (
                SELECT
                    1 
                FROM
                    [dbo].[Policy] ds 
                WHERE
                    ds.OrganizationId = so.OrganizationId 
                    AND ds.Type = @DisableSendType

```

harr1424: Thanks for the careful review and for running the proposed script against production data, this was not caught during my local testing.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+        Policy? previousPolicyState)
+    {
+        var policyUpdate = policyRequest.PolicyUpdate;
+
+        var sendControlsPolicy = await policyRepository.GetByOrganizationIdTypeAsync(
+            policyUpdate.OrganizationId, PolicyType.SendControls);
+
+        var data = sendControlsPolicy != null
+            ? CoreHelpers.LoadClassFromJsonData<SendControlsPolicyData>(sendControlsPolicy.Data) ?? new SendControlsPolicyData()
+            : new SendControlsPolicyData();
```
eliykat: Please use `policy.GetDataModel` instead of `CoreHelpers`. The return type is non-nullable.


```suggestion
        var data = sendControlsPolicy != null
            ? sendControlsPolicy.GetDataModel<SendControlsPolicyData>()
            : new SendControlsPolicyData();
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyValidator.cs
```diff
@@ -0,0 +1,54 @@
+﻿#nullable enable
```
eliykat: Not required - nullable is the default.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+            OrganizationId = policyUpdate.OrganizationId,
+            Type = PolicyType.SendControls,
+        };
+
+        if (sendControlsPolicy == null)
+        {
+            policy.SetNewId();
+        }
+
+        policy.Enabled = data.DisableSend || data.DisableHideEmail;
```
eliykat: The configuration of a policy (in its `data`) is separate to its enabled/disabled state. I think the proper mapping would be: `sendControlsPolicy.Enabled = disableSendPolicy.Enabled || SendOptionsPolicy.Enabled`. (not the variable names you have here, example only)

This doesn't make a difference for DisableSend, because it's mapped from the Enabled state anyway, but it does for DisableHideEmail.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+            ? CoreHelpers.LoadClassFromJsonData<SendControlsPolicyData>(sendControlsPolicy.Data) ?? new SendControlsPolicyData()
+            : new SendControlsPolicyData();
+
+        data.DisableSend = postUpsertedPolicyState.Enabled;
+
+        var policy = sendControlsPolicy ?? new Policy
+        {
+            OrganizationId = policyUpdate.OrganizationId,
+            Type = PolicyType.SendControls,
+        };
```
eliykat: All the different `policy` objects floating around here are a bit confusing. Maybe handle this upfront:

```
var sendControlsPolicy = await policyRepository.GetByOrganizationIdTypeAsync(
            policyUpdate.OrganizationId, PolicyType.SendControls) ?? new Policy
        {
            Id = CoreHelpers.GenerateComb(),
            OrganizationId = policyUpdate.OrganizationId,
            Type = PolicyType.SendControls,
        };
```

So you can then just deal with this object without worrying whether it existed or not.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+using Bit.Core.Utilities;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyValidators;
+
+/// <summary>
+/// Syncs changes to the DisableSend policy into the SendControls policy row.
+/// Runs regardless of the pm-31885-send-controls feature flag to ensure SendControls
+/// always stays current for when the flag is eventually enabled.
+/// </summary>
+public class DisableSendSyncPolicyValidator(IPolicyRepository policyRepository) : IOnPolicyPostUpdateEvent
```
eliykat: The existing classes are _also_ Validators, but this is not. A more accurate name could be something like


```suggestion
public class DisableSendSyncUpdateEvent(IPolicyRepository policyRepository) : IOnPolicyPostUpdateEvent
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyValidator.cs
```diff
@@ -0,0 +1,75 @@
+﻿#nullable enable
```
eliykat: Not required

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+        Policy postUpsertedPolicyState,
+        Policy? previousPolicyState)
+    {
+        if (!featureService.IsEnabled(FeatureFlagKeys.SendControls))
+        {
+            return;
+        }
+
+        var data = CoreHelpers.LoadClassFromJsonData<SendControlsPolicyData>(postUpsertedPolicyState.Data)
+            ?? new SendControlsPolicyData();
```
eliykat: Should use the getter on the policy, return type is non-nullable.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyEvent.cs
```diff
@@ -0,0 +1,75 @@
...
+            Type = type,
+        };
+
+        if (existing == null)
+        {
+            policy.SetNewId();
+        }
+
+        policy.Enabled = enabled;
+        policy.Data = policyData;
```
eliykat: Also need to bump the `RevisionDate` (making sure to use `TimeProvider`).

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendOptionsSyncPolicyEvent.cs
eliykat, line 1: Feedback on the DisableSend version of this class is also relevant here.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyEventTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        await sutProvider.Sut.ExecutePostUpsertSideEffectAsync(
+            new SavePolicyModel(policyUpdate), postUpsertedPolicy, null);
+
+        await sutProvider.GetDependency<IPolicyRepository>()
+            .Received(1)
+            .UpsertAsync(Arg.Is<Policy>(p =>
+                p.OrganizationId == policyUpdate.OrganizationId &&
+                p.Type == PolicyType.SendControls &&
+                p.Enabled == true &&
+                (CoreHelpers.LoadClassFromJsonData<SendControlsPolicyData>(p.Data)!.DisableSend == true)));
```
eliykat: Tests should also use the getter to access the data model.

util/Migrator/DbScripts_transition/2026-02-28_00_CreateSendControlsPolicies.sql
```diff
@@ -0,0 +1,77 @@
...
+        WHERE 
+            so.Type = @SendOptionsType
+            AND NOT EXISTS (
+                SELECT
+                    1 
+                FROM
+                    [dbo].[Policy] ds 
+                WHERE
+                    ds.OrganizationId = so.OrganizationId 
+                    AND ds.Type = @DisableSendType
```
eliykat: dbops will be a better source of advice on this, but the design of this query is a little unusual to me. "select send options with left join on disable send, but also UNION with select disable send where there are no send options". Couldn't we just select send options UNION select disable send?

util/SqliteMigrations/Migrations/20260306000000_CreateSendControlsPolicies.cs
```diff
@@ -0,0 +1,76 @@
...
+        // SQLite does not support FULL OUTER JOIN; use two separate inserts.
+        // SQLite 3.38+ has json_valid() and json_extract().
+
+        // Insert for orgs that have SendOptions (with or without DisableSend)
+        migrationBuilder.Sql(@"
+            INSERT INTO ""Policy"" (""Id"", ""OrganizationId"", ""Type"", ""Enabled"", ""Data"", ""CreationDate"", ""RevisionDate"")
+            SELECT lower(hex(randomblob(4))) || '-' || lower(hex(randomblob(2))) || '-4' ||
+                   substr(lower(hex(randomblob(2))),2) || '-' ||
+                   substr('89ab',abs(random()) % 4 + 1, 1) ||
+                   substr(lower(hex(randomblob(2))),2) || '-' || lower(hex(randomblob(6))),
```
eliykat: Again I will defer to dbops, but this doens't seem like the right way to generate guids.

util/Migrator/DbScripts_transition/2026-02-28_00_CreateSendControlsPolicies.sql
```diff
@@ -0,0 +1,77 @@
...
+        -- Build JSON: use ISJSON guard for SendOptions.Data
+        N'{"disableSend":' +
+            CASE WHEN ISNULL(combined.DisableSendEnabled, 0) = 1
+                 THEN N'true' ELSE N'false' END +
+        N',"disableHideEmail":' +
+            CASE WHEN combined.SendOptionsData IS NOT NULL
+                      AND ISJSON(combined.SendOptionsData) = 1
+                      AND JSON_VALUE(combined.SendOptionsData, '$.disableHideEmail') = 'true'
+                 THEN N'true' ELSE N'false' END +
+        N'}',
```
eliykat: Could we use one of the json functions here, e.g. https://learn.microsoft.com/en-us/sql/t-sql/functions/json-object-transact-sql?view=sql-server-ver17 ?

util/Migrator/DbScripts_transition/2026-02-28_00_CreateSendControlsPolicies.sql
eliykat, line 1: Data migrations can be pretty high risk - definitely talk to dbops about this. They may want to run it manually to control/monitor performance if it is expected to take a while (especially with json parsing).

You should also write integration tests for this - see https://contributing.bitwarden.com/contributing/testing/database/#testing-data-migrations. Note that these tests are not permanent, but they give you confidence when merging your PR - especially in a non-trivial data migration script like this one, with separate implementations per database provider.

eliykat (changes requested):
Nice use of the policy framework. General approach looks good to me.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    public async Task ExecutePostUpsertSideEffectAsync(
+        SavePolicyModel policyRequest,
+        Policy postUpsertedPolicyState,
+        Policy? previousPolicyState)
+    {
+        if (!featureService.IsEnabled(FeatureFlagKeys.SendControls))
+        {
+            return;
+        }
```
eliykat: Not sure if this guard is needed - you don't expect any updates if the feature flag is off, sure, but if you get one - wouldn't you still want to sync it?

Note that the feature flag state in clients can lag behind server (by whatever the refresh interval is), so it's possible that an admin edits the Send Controls policy on the front-end even though the feature flag has been turned off.

util/SqliteMigrations/Migrations/20260306000000_CreateSendControlsPolicies.cs
harr1424: @mkincaid-bw Sqlite doesn't seem to have a function equivalent to `NEWID()` or `UUID()` etc. do you have any input on the approach used here? For context it's a migration script, I don't see a way to generate GUIDs in C# or something similar. I haven't been able to find any precedence for what we are trying to accomplish here.

util/Migrator/DbScripts_transition/2026-02-28_00_CreateSendControlsPolicies.sql
harr1424: @mkincaid-bw I couldn't find any existing usage of `JSON_OBJECT()` in the solution, and see this was introduced in SQL Server 2022. Could this lead to any potential problems?

util/Migrator/DbScripts_transition/2026-02-28_00_CreateSendControlsPolicies.sql
harr1424: @mkincaid-bw  do you have any input on this?

eliykat: I was thinking about this further, and my suggestion isn't correct. UNION just concatenates rows to the previous result, which is not what you're after here. Your `combined` result is essentially a `PIVOT` operation, changing rows to columns. But I think it can be simplified to avoid this just by breaking it down into multiple CTEs.

Here's what I'm thinking (please note this would need to be verified, tested and adjusted to work with batching):

```sql
      -- Step 1: Extract all DisableSend policies
      WITH DisableSendPolicies AS (
          SELECT
              OrganizationId,
              Enabled AS DisableSendEnabled
          FROM [dbo].[Policy]
          WHERE Type = @DisableSendType
      ),
      -- Step 2: Extract all SendOptions policies
      SendOptionsPolicies AS (
          SELECT
              OrganizationId,
              Enabled AS SendOptionsEnabled,
              Data AS SendOptionsData
          FROM [dbo].[Policy]
          WHERE Type = @SendOptionsType
      ),
      -- Step 3: Combine both policy types (all orgs with either policy)
      CombinedPolicies AS (
          SELECT DISTINCT
              COALESCE(ds.OrganizationId, so.OrganizationId) AS OrganizationId,
              ds.DisableSendEnabled,
              so.SendOptionsEnabled,
              so.SendOptionsData
          FROM DisableSendPolicies ds
          FULL OUTER JOIN SendOptionsPolicies so
              ON ds.OrganizationId = so.OrganizationId
      ),
      -- Step 4: Filter out orgs that already have SendControls policy
      OrgsNeedingMigration AS (
          SELECT
              OrganizationId,
              DisableSendEnabled,
              SendOptionsEnabled,
              SendOptionsData
          FROM CombinedPolicies cp
          WHERE NOT EXISTS (
              SELECT 1
              FROM [dbo].[Policy] sc
              WHERE sc.OrganizationId = cp.OrganizationId
                AND sc.Type = @SendControlsType
          )
      )
      -- Step 5: Insert new SendControls policies
      -- ....etc
```

harr1424:
~~This PR is being closed in favor of an approach that will not require migration.~~

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+        var sendControlsPolicy = await policyRepository.GetByOrganizationIdTypeAsync(
+            policyUpdate.OrganizationId, PolicyType.SendControls) ?? new Policy
+            {
+                Id = CoreHelpers.GenerateComb(),
+                OrganizationId = policyUpdate.OrganizationId,
+                Type = PolicyType.SendControls,
+            };
+
+        var sendControlsPolicyData =
+            sendControlsPolicy.GetDataModel<SendControlsPolicyData>();
```
harr1424: @eliykat could this be obtained directly using `postUpsertedPolicyState`?   I might be confused, but I vaguely recall that using `policyRepository` was a deliberate change here, although looking this over it does appear redundant.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyEvent.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            sendControlsPolicyData.DisableHideEmail =
+                sendOptionsPolicy.GetDataModel<SendOptionsPolicyData>().DisableHideEmail;
+        }
+
+        sendControlsPolicy.SetDataModel(sendControlsPolicyData);
+
+        // Step 2: sync Enabled status. SendControlsPolicy is enabled if either legacy policy is enabled
+        sendControlsPolicy.Enabled = postUpsertedPolicyState.Enabled || (sendOptionsPolicy?.Enabled ?? false);
+
```
mcamirault, line 53: ❓ Do we need to bump the revision date here similarly to how we do when syncing in the other direction?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendOptionsSyncPolicyEvent.cs
```diff
@@ -0,0 +1,48 @@
...
+        var sendControlsPolicyData = sendControlsPolicy.GetDataModel<SendControlsPolicyData>();
+        sendControlsPolicyData.DisableHideEmail = postUpsertedPolicyState.GetDataModel<SendOptionsPolicyData>().DisableHideEmail;
+        sendControlsPolicy.SetDataModel(sendControlsPolicyData);
+
+        // Step 2: sync Enabled status. SendControlsPolicy is enabled if either legacy policy is enabled
+        // Optimization: DisableSendPolicy.Enabled maps to SendControlsPolicy.Data.DisableSend - so we can use that
+        // as a proxy for that legacy policy state
+        sendControlsPolicy.Enabled = postUpsertedPolicyState.Enabled ||
+                                     sendControlsPolicyData.DisableSend;
+
```
mcamirault, line 47: ❓ Same question about revision date from the DisableSend version of this class

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -47,13 +50,39 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
...
+        var sendOptionsTask = _policyRequirementQuery.GetAsync<SendOptionsPolicyRequirement>(userId.Value);
+
+        await Task.WhenAll(sendControlsTask, disableSendTask, sendOptionsTask);
+
+        var sendControlsRequirement = sendControlsTask.Result;
+        var disableSendRequirement = disableSendTask.Result;
+        var sendOptionsRequirement = sendOptionsTask.Result;
+        #endregion
+
+        if (_featureService.IsEnabled(FeatureFlagKeys.SendControls))
```
mcamirault: ❓ Since we'll always be syncing the old policies' changes into the SendControls policy, can we just check it regardless of the feature flag?

mcamirault (commented):
Looks good! A couple questions but nothing major

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
harr1424: The idea here is that if we decide to turn the flag off for any reason, that `SendControls` will no longer be used to determine policy enforcement logic. Here, the `if` block isolates the usage of `SendControls` policy and gates it behind the feature flag. 

It is true that if an admin has saved any policy edits while the flag was on, that these changes have been synced back into the legacy policy rows at the level of the database, but I see this guard as gating logic at a service level. 

Please let me know if you see a strong reason to remove it.

mcamirault: That makes sense, thanks for the explanation!

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyEvent.cs
harr1424: I think this would make sense, it would add parity among the different sync directions, but it also introduces a dependency on `TimeProvider` into these files and associated tests. I also don't see this column being used by any logic in clients, but it seems that by not updating it, we loose visibility that may be useful for auditing the database rows. 

I am in favor of making changes to update the `RevisionDate`, @eliykat do you have any input?

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -88,8 +88,9 @@ public async Task<ListResponseModel<PolicyResponseModel>> GetAll(string orgId)
             throw new NotFoundException();
         }
 
-        var policies = await _policyRepository.GetManyByOrganizationIdAsync(orgIdGuid);
-
+        // Once migration from legacy Send policies > SendControls has run, replace _policyQuery.GetAllAsync
+        // with a direct _policyRepository.GetManyByOrganizationIdAsync call
```
eliykat: Not required - this is a good change you've made.

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -88,8 +88,9 @@ public async Task<ListResponseModel<PolicyResponseModel>> GetAll(string orgId)
...
+        var policies = await _policyQuery.GetAllAsync(orgIdGuid);
         return new ListResponseModel<PolicyResponseModel>(policies.Select(p => new PolicyResponseModel(p)));
```
eliykat: `PolicyStatus` maps to `PolicyStatusResponseModel`. I have checked uses of this endpoint and it is safe to make this change.

src/Api/AdminConsole/Models/Response/Organizations/PolicyResponseModel.cs
eliykat, line 1: Changes are no longer required once you use `PolicyStatusResponseModel`.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyQuery.cs
```diff
@@ -1,14 +1,71 @@
...
+        var results = policies.Select(p => new PolicyStatus(organizationId, p.Type, p)).ToList();
+
+        // Remove this block once migration from legacy Send policies > SendControls has run
+        if (policies.All(p => p.Type != PolicyType.SendControls))
+        {
+            var synthesized = await SynthesizeSendControlsStatusAsync(organizationId);
+            if (synthesized.Enabled)
+            {
+                results.Add(synthesized);
+            }
```
eliykat: `Enabled` check is unnecessary - this represents the status of all policies, whether enabled or not.


```suggestion
            var synthesized = await SynthesizeSendControlsStatusAsync(organizationId);
            results.Add(synthesized);
```

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -47,13 +50,39 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
             return;
         }
 
-        var disableSendRequirement = await _policyRequirementQuery.GetAsync<DisableSendPolicyRequirement>(userId.Value);
+        #region Fetch Policy Requirements Async
```
eliykat: Code regions should not be used: https://contributing.bitwarden.com/contributing/code-style/#code-regions

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -47,13 +50,39 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
...
+        var sendControlsTask = _policyRequirementQuery.GetAsync<SendControlsPolicyRequirement>(userId.Value);
+        var disableSendTask = _policyRequirementQuery.GetAsync<DisableSendPolicyRequirement>(userId.Value);
+        var sendOptionsTask = _policyRequirementQuery.GetAsync<SendOptionsPolicyRequirement>(userId.Value);
```
eliykat: Your syncing logic should prevent you from having to read all locations. You can continue to check the current (old) policies only, because any updates to the new policy will sync back to update them.

After the final data migrations have run (in the future), this code can be updated to read from the new location.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyEvent.cs
eliykat: Yes it's a good call, let's bump the revision date whenever we're making updates. It's OK to add `TimeProvider`.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/DisableSendSyncPolicyEvent.cs
```diff
@@ -0,0 +1,54 @@
...
+                OrganizationId = organizationId,
+                Type = PolicyType.SendControls,
+            };
+
+        var sendOptionsPolicy = await policyRepository.GetByOrganizationIdTypeAsync(
+            organizationId, PolicyType.SendOptions);
+
+        var sendControlsPolicyData = sendControlsPolicy.GetDataModel<SendControlsPolicyData>();
+        sendControlsPolicyData.DisableSend = postUpsertedPolicyState.Enabled;
+        if (sendOptionsPolicy?.Enabled == true)
```
eliykat: This is probably a ⛏️ , but its Enabled status is different to its configuration. For example: you can have a Send Options policy that is disabled, but with `DisableHideEmail` set to true. Let's keep these separate when we map it here as well, so that we map any inactive policy configurations. It still won't do anything unless both policies are Enabled anyway, but it'll make sure we preserve these values.


```suggestion
        if (sendOptionsPolicy != null)
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyEvent.cs
eliykat: You're right, I think that was a copy/paste error as this is the sync back in the other direction. Use `postUpsertedPolicyState`.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/SendControlsSyncPolicyEvent.cs
```diff
@@ -0,0 +1,74 @@
...
+            PolicyType.DisableSend,
+            enabled: postUpsertedPolicyState.Enabled && sendControlsPolicyData.DisableSend,
+            policyData: null);
+
+        var sendOptionsData = new SendOptionsPolicyData { DisableHideEmail = sendControlsPolicyData.DisableHideEmail };
+        await UpsertLegacyPolicyAsync(
+            policyRequest.PolicyUpdate.OrganizationId,
+            PolicyType.SendOptions,
+            enabled: postUpsertedPolicyState.Enabled && sendControlsPolicyData.DisableHideEmail,
+            policyData: CoreHelpers.ClassToJsonData(sendOptionsData));
```
eliykat: Please pass the `IPolicyDataModel` and then use `Policy.SetDataModel` to ensure this is correctly saved.

eliykat (changes requested):
Thanks for sticking with it, good work.

src/Core/AdminConsole/OrganizationFeatures/Policies/Implementations/PolicyQuery.cs
```diff
@@ -1,14 +1,71 @@
...
+    {
+        var disableSendPolicy = await policyRepository.GetByOrganizationIdTypeAsync(
+            organizationId, PolicyType.DisableSend);
+        var sendOptionsPolicy = await policyRepository.GetByOrganizationIdTypeAsync(
+            organizationId, PolicyType.SendOptions);
+
+        var disableSend = disableSendPolicy?.Enabled ?? false;
+        var disableHideEmail = sendOptionsPolicy?.Enabled == true
+            && sendOptionsPolicy.GetDataModel<SendOptionsPolicyData>().DisableHideEmail;
+        var enabled = disableSend || disableHideEmail;
```
eliykat: This is subtly different to the logic in the sync handlers. For Send Options, sync the `DisableHideEmail` value and the `Enabled` value separately. See https://github.com/bitwarden/server/pull/7113#discussion_r2985935741

src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -80,18 +68,17 @@ public async Task<PolicyStatusResponseModel> Get(Guid orgId, PolicyType type)
     }
 
     [HttpGet("")]
-    public async Task<ListResponseModel<PolicyResponseModel>> GetAll(string orgId)
+    [Authorize<ManagePoliciesRequirement>]
+    public async Task<ListResponseModel<PolicyStatusResponseModel>> GetAll(string orgId)
     {
         var orgIdGuid = new Guid(orgId);
```
eliykat: (non-blocking) Parameter should accept a `Guid` directly. I think Claude has picked up on an old pattern here.

eliykat (dismissed):
:shipit:

mcamirault (dismissed):

eliykat (approved):

mcamirault (approved):

PR #7123, opened 2026-03-02
[PM-28519] Remove Emergency Access Contacts for AutoConfirm Org Flows

sven-bitwarden:
## 🎟️ Tracking
[PM-28519](https://bitwarden.atlassian.net/browse/PM-28519)

## 📔 Objective
To prevent unintended access of org keys from users' emergency access contacts, we are eliminating all these contacts from accepted/confirmed users in the organization as they go through user flows, and when the policy is enabled.

### Technical Implementation
We already have extensive policy validation setup through PolicyRequirements, but the entire process is self-contained (between loading && validating data). We needed an extension, at the time a policy is enforced (User is being accepted/confirmed/restored), to know that a policy is enabled and validated.

This was accomplished by exposing an overload to automatic confirmation policy validation that takes a PolicyRequirement, which avoids duplicitous data loading. We can use that requirement to both validate and check enabled status.

### EA Removed On Policy Save
https://github.com/user-attachments/assets/46e5b612-8a41-4bd1-9ccd-652bfd391307

### EA Removed On Invite Accept
https://github.com/user-attachments/assets/6c1f6e92-23e0-4869-89cf-c133eb7e4e23

https://github.com/user-attachments/assets/7762c33b-5a63-4312-a0db-5eb8f6ed4769

### EA Removed on Restore Accepted User
https://github.com/user-attachments/assets/29c535fb-99b7-4889-8a57-d83bea4a6d17

[PM-28519]: https://bitwarden.atlassian.net/browse/PM-28519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RestoreUser/v1/RestoreOrganizationUserCommand.cs
```diff
@@ -364,6 +368,11 @@ private async Task CheckPoliciesBeforeRestoreAsync(OrganizationUser orgUser, boo
             {
                 throw badRequestException;
             }
+
+            if (policyRequirement.IsEnabled(orgUser.OrganizationId))
```
JimmyVo16, line 386: Should we do this `IsEnabled` check first before going through the whole validation process, since it’s a simpler, non-task operation?

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptOrgUserCommand.cs
```diff
@@ -264,5 +273,10 @@ private async Task ValidateAutomaticUserConfirmationPolicyAsync(OrganizationUser
         {
             throw new BadRequestException(error);
         }
+
+        if (policyRequirement.IsEnabled(orgUser.OrganizationId))
```
JimmyVo16, line 277: [Same](https://github.com/bitwarden/server/pull/7123/changes#r2878692680)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RestoreUser/v1/RestoreOrganizationUserCommand.cs
sven-bitwarden: The Validator itself has a bunch of checks, both when the requirement is/isn't enabled. It would end up being a breaking change if I validated only when enabled

sven-bitwarden: If it helps too - the reason we even load the PolicyRequirement here is because the validator only says if the user is compliant or not, but we need to delete EA when both the user is compliant **and** the policy is enabled. IsCompliant doesn't give us a clue if it's enabled

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/AutomaticUserConfirmationPolicyEventHandler.cs
```diff
@@ -42,6 +47,20 @@ public async Task<string> ValidateAsync(PolicyUpdate policyUpdate, Policy? curre
...
+            return;
+        }
+
+        var orgUsers = await organizationUserRepository.GetManyByOrganizationAsync(policyRequest.PolicyUpdate.OrganizationId, null);
+        var orgUserIds = orgUsers.Where(w => w.UserId != null).Select(s => s.UserId!.Value).ToList();
+
+        await deleteEmergencyAccessCommand.DeleteAllByUserIdsAsync(orgUserIds);
+    }
+
+    public Task OnSaveSideEffectsAsync(PolicyUpdate policyUpdate, Policy? currentPolicy) => Task.CompletedTask;
```
JimmyVo16: Just want to preface this by saying that we need documentation around this, and we already have a ticket for it. I’ll move it higher up on my priority list.

The review: You're adding `ExecutePreUpsertSideEffectAsync`, so you need to have `OnSaveSideEffectsAsync` call `ExecutePreUpsertSideEffectAsync` so that this handler gets triggered for both `SavePolicyCommand` and `VNextSavePolicyCommand`. Here’s an [example](https://github.com/bitwarden/server/blob/132db95fb76f86705b203b0b771ccedd383c0f4b/src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/FreeFamiliesForEnterprisePolicyValidator.cs#L22).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RestoreUser/v1/RestoreOrganizationUserCommand.cs
JimmyVo16: Oo I see. I thought this was just for validation, but we’re ensuring that the user is compliant as well. Disregard my previous comment.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyValidators/AutomaticUserConfirmationPolicyEventHandler.cs
sven-bitwarden: makes sense/thanks for the example - updated.

PR #7129, opened 2026-03-03
[PM-31475] Remove decline event log for org data ownership policy

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31475

## 📔 Objective

* Remove the following server-side EventType: OrganizationUser_SelfRevoked = 1518, // User self-revoked due to declining organization data ownership policy
* Remove the call for the server-side Self-Revoke endpoint to generate an event log (it is handled client-side)
* Update associated test cases to remove the expectation of generating an event log


eliykat (changes requested):
Events should be collected server-side wherever possible. A client can always block or intercept the event being logged before it is sent to the server, so it is an unreliable source. We only use client-side event logging where it is a purely client-side event - e.g. viewing a password.

Sorry to send this back at this stage, but I suggest the solution here is:
- keep the server-side event (in AC code)
- update the event display logic to show it properly (fix the empty string)
- remove the duplicate client-side event from Vault code

PR #7136, opened 2026-03-04
[PM-30584] Add support for key-connector-migration setting key

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30584
https://github.com/bitwarden/server/pull/7136
https://github.com/bitwarden/sdk-internal/pull/809
https://github.com/bitwarden/clients/pull/19360

## 📔 Objective

We do not want to keep the same master-key that was used during password derivation, but use a separately sampled key - named "key-connector-key". This means that the conversion flow now requires:
- Posting the key to key-connector
- Setting the "key" field on the user to "key-connector-key-wrapped-user-key"

To keep backwards compatibility, we make the request body optional, but if present, the request body contains the key-connector-key-wrapped-user-key. This is subsequently set to the user object.

This will unblock setting the master-key to state during unlock and login, which will improve unlock time, since we can remove double-kdf-derivation.

### DIFF
--- a/src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
+++ b/src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
@@ -19,7 +19,6 @@ using Bit.Core.KeyManagement.UserKey;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
 using Bit.Core.Tools.Entities;
-using Bit.Core.Utilities;
 using Bit.Core.Vault.Entities;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;


src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -184,7 +184,35 @@ public async Task PostConvertToKeyConnectorAsync()
...
+
+        foreach (var error in result.Errors)
+        {
+            ModelState.AddModelError(string.Empty, error.Description);
+        }
+
+        throw new BadRequestException(ModelState);
+    }
+
+    [HttpPost("key-connector/enroll")]
```
Thomas-Avery, line 201:
```suggestion
    [SelfHosted(SelfHostedOnly = true)]
    [HttpPost("key-connector/enroll")]
```

💭  Non blocking.

Should we restrict this to only be available for self-hosted installs? I've thought this should be the case for all key connector endpoints, but it is a good time to consider it for a new endpoint.

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -184,7 +184,35 @@ public async Task PostConvertToKeyConnectorAsync()
...
+        var user = await _userService.GetUserByPrincipalAsync(User);
+        if (user == null)
+        {
+            throw new UnauthorizedAccessException();
+        }
+
+        if (string.IsNullOrWhiteSpace(model.KeyConnectorKeyWrappedUserKey))
+        {
+            throw new BadRequestException("KeyConnectorKeyWrappedUserKey must be supplied when request body is provided.");
+        }
```
Thomas-Avery: Can we add this request validation to the model instead of in the controller via `IValidatableObject`?

src/Core/Services/Implementations/UserService.cs
```diff
@@ -522,7 +522,7 @@ public async Task<IdentityResult> SetKeyConnectorKeyAsync(User user, string key,
         return IdentityResult.Success;
     }
 
-    public async Task<IdentityResult> ConvertToKeyConnectorAsync(User user)
+    public async Task<IdentityResult> ConvertToKeyConnectorAsync(User user, string keyConnectorKeyWrappedUserKey = null)
```
Thomas-Avery, line 525: Would be nice to have some unit test coverage for this.

Thomas-Avery (changes requested):
Code changes look good. Would be nice to get some coverage for the user service method.

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
quexten: I wrote a note in slack on this. I can't quite figure out how to combine this with integration tests correctly. If you give me a pointer, happy to do it, otherwise I'll skip it here / not dig further since you don't consider it blocking.

PR #7137, opened 2026-03-04
[PM-33091] Add optional Targeting Rules data resource configuration

jprusik:
## 🎟️ Tracking

[PM-33091](https://bitwarden.atlassian.net/browse/PM-33091)

## 📔 Objective

Allow servers to optionally specify the endpoint that should be used to fetch targeting rules. If no endpoint is configured, `fillAssistRules` is excluded from the server configuration entirely

Note, the server does not parse or store rules, only the (optional) resource URI override.

## 📸 Screenshots

example `api/config` response:

Screenshot 2026-03-05 at 12 07 00 PM

[PM-33091]: https://bitwarden.atlassian.net/browse/PM-33091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Settings/GlobalSettings.cs
+++ b/src/Core/Settings/GlobalSettings.cs
@@ -1,6 +1,8 @@
 ﻿// FIXME: Update this file to be null safe and then delete the line below
 #nullable disable
 
+using System.Globalization;
+
 using Bit.Core.Auth.Settings;
 
 namespace Bit.Core.Settings;
@@ -105,7 +107,7 @@ public class GlobalSettings : IGlobalSettings
         {
             return null;
         }
-        return string.Format("{0}/{1}", BaseServiceUri.Vault, name);
+        return string.Format(CultureInfo.InvariantCulture, "{0}/{1}", BaseServiceUri.Vault, name);
     }
 
     public string BuildInternalUri(string explicitValue, string name)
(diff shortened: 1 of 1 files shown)


src/Api/appsettings.Development.json
```diff
@@ -7,6 +7,7 @@
       "admin": "http://localhost:62911",
       "notifications": "http://localhost:61840",
       "sso": "http://localhost:51822",
+      "fillAssistRules": "http://localhost:1495",
```
MGibson1, line 10: Is there a follow up to add this to the vscode launch settings or dev containers?

src/Core/Settings/GlobalSettings.cs
MGibson1, line 1: Go ahead and shut these warnings up by [providing a culture](https://github.com/bitwarden/server/blob/165117fc2c7d46119fd66f988bef636eab80eac3/src/Core/Billing/Extensions/InvoiceExtensions.cs#L55-L56)

jprusik: addressed in cc5d9bf41661ff5f268be3f9c8c0ff94705646b2

src/Api/appsettings.Development.json
jprusik: I hadn't thought about that; I think we probably should, but maybe don't need to worry about it this early as some considerations are likely to change? I mostly wanted to get the base need supported on server early.

I've created a followup task for this in the epic:
https://bitwarden.atlassian.net/browse/PM-33200

MGibson1 (approved):

PR #7139, opened 2026-03-04
[PM-15489] 2fa account recovery

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15489

## 📔 Objective

Allow account recovery to selectively recover password and/or 2FA on org members.


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -546,7 +550,24 @@ public async Task<IResult> PutResetPassword(Guid orgId, Guid id, [FromBody] Orga
...
+        if (validationResult.IsError)
+        {
+            var error = validationResult.AsError;
+            return error switch
+            {
+                NotFoundError notFound => TypedResults.NotFound(new ErrorResponseModel(notFound.Message)),
+                BadRequestError badRequest => TypedResults.BadRequest(new ErrorResponseModel(badRequest.Message)),
+                _ => TypedResults.Json(new ErrorResponseModel(error.Message), statusCode: StatusCodes.Status500InternalServerError)
+            };
+        }
```
kspearrin: This part smells off from what we are are trying to do. I looked at the `DeleteClaimedOrganizationUserAccountCommand` but it is unclear how that translates to this implementation.

Is the validator suppose to be called from within the command (vs controller)? how do those errors propagate back tot he controller to throw the error back to client?

eliykat: Yes, the validator is called within the command.

Controller:
- create commandRequest
- call the command

Command:
- calls validation
- throws if validation fails
- continues with the operation otherwise

Example snippet to handle validation failure:

```c#
        // from within command
        var validationResult = await validator.ValidateAsync(request);
        if (validationResult.IsError)
        {
            // TODO: return a CommandResult, but for now we will just throw
            throw new BadRequestException(validationResult.AsError.Message);
        }

        // otherwise continue with operation
```

---

Optional refactor / additional information:

The ideal end state is that the command returns a `CommandResult`. A CommandResult is a OneOf type (effectively a union) that can hold an error or none (where none indicates success). There is some nice implicit conversion to wrap the error/none for you.

```c#
    // return type must be correct for implicit conversion to work
    public async Task<CommandResult> RecoverAccountAsync(RecoverAccountRequest request)
    {
        var validationResult = await validator.ValidateAsync(request);
        if (validationResult.IsError)
        {
            return validationResult.AsError; // gets converted to a failed CommandResult
        }

       // otherwise continue with operation
       return new None(); // gets converted to a successful CommandResult
```

Then the controller already has logic to handle the commandResult into the appropriate response type:

```c#
return Handle(result);
```

This is a pattern we have adopted to better separate validation vs. action, and to avoid throwing exceptions for business logic.

eliykat (commented):
Commented above

src/Core/Platform/Mail/HandlebarsMailService.cs
```diff
@@ -766,13 +766,15 @@ public async Task SendEnqueuedMailMessageAsync(IMailQueueMessage queueMessage)
         await _mailDeliveryService.SendEmailAsync(message);
     }
 
-    public async Task SendAdminResetPasswordEmailAsync(string email, string? userName, string orgName)
+    public async Task SendAdminResetPasswordEmailAsync(string email, string? userName, string orgName, bool resetMasterPassword, bool resetTwoFactor)
```
BTreston, line 769: This is starting to feel like primitive obsession. Passing in the whole user, org, and request objects might be simpler for this method.

src/Core/AdminConsole/OrganizationFeatures/AccountRecovery/AdminRecoverAccountCommand.cs
```diff
@@ -1,78 +1,139 @@
...
 
-        // Enterprise policy must be enabled
-        var resetPasswordPolicy = await policyQuery.RunAsync(orgId, PolicyType.ResetPassword);
-        if (!resetPasswordPolicy.Enabled)
+        // 2FA reset
+        if (request.ResetTwoFactor)
         {
-            throw new BadRequestException("Organization does not have the password reset policy enabled.");
+            await resetUserTwoFactorCommand.ResetAsync(user!);
         }
```
BTreston, line 58: With this approach, we are calling `ReplaceAsync` on the same user twice in the same flow. Can this be restructured to avoid duplicate writes?

src/Core/AdminConsole/OrganizationFeatures/AccountRecovery/AdminRecoverAccountCommand.cs
```diff
@@ -1,78 +1,139 @@
...
+        if (request.ResetMasterPassword)
         {
-            throw new BadRequestException("Organization does not allow password reset.");
+            var result = await userService.UpdatePasswordHash(user!, request.NewMasterPasswordHash!);
+            if (!result.Succeeded)
+            {
+                return result;
+            }
+
+            user!.RevisionDate = user.AccountRevisionDate = timeProvider.GetUtcNow().UtcDateTime;
```
BTreston: I think an explicit check for `null` is probably more clear than using `!`. Same thing for org

BTreston (changes requested):
I haven't seen the product designs for this so I have a couple of question: Why are we opting to roll this into the existing account recovery flow rather than it having its own an explicit codepath? My assumption is that its just going to be a new check-box in the existing dialog. 
Continuing that thought, do we have concerns about this option being re-usable with how tightly coupled it is to account recovery? I feel like I can easily see this turning into an option under recover account in the `...` menu.

eliykat:
> I haven't seen the product designs for this so I have a couple of question: Why are we opting to roll this into the existing account recovery flow rather than it having its own an explicit codepath? My assumption is that its just going to be a new check-box in the existing dialog. Continuing that thought, do we have concerns about this option being re-usable with how tightly coupled it is to account recovery? I feel like I can easily see this turning into an option under recover account in the `...` menu.

@BTreston There was some discussion about this already, it is indeed a new checkbox in the existing dialog (iirc), but that means we can do it in the same request. If there are separate requests then you can get into a place where (for example) password reset succeeded but 2FA reset failed and the account is left in an intermediate state, and you have to explain this to the user somehow. Combining it is arguably less RESTful but makes it atomic when both are done together, while still supporting doing both operations separately if desired.

src/Core/AdminConsole/OrganizationFeatures/AccountRecovery/AdminRecoverAccountCommand.cs
kspearrin: I am not sure how to resolve that given the @bitwarden/team-auth requirements for it to be in a separate command that they own.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
kspearrin: How about now? Tried to follow your explanation and this new revision feels better.

src/Core/AdminConsole/OrganizationFeatures/AccountRecovery/AdminRecoverAccountCommand.cs
kspearrin: Should be resolved now.

src/Core/AdminConsole/OrganizationFeatures/AccountRecovery/AdminRecoverAccountCommand.cs
BTreston: In that case, I'm fine with leaving it as is. Most likely not worth the added complexity to avoid it.

BTreston (commented):
LGTM aside from the failing test. I think we probably just need to fix the request model in `OrganizationUsersControllerPutResetPasswordTests`

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
eliykat: I agree, this is looking good.

eliykat (dismissed):
Looking good overall, thanks.

One thing I noticed is that the feature flag is only guarding the 2FA reset, but our changes have impacted the entire command. With this level of changes I suggest that the whole command needs to be feature flagged, so that we are not risking any regressions to current functionality (outside of the feature flag).

I think you could do this relatively easily as follows:
1. move the command and new files to a v2 folder/namespace
2. restore the current file from main (now you have both old and new versions)
3. move feature flag logic to the controller to call the correct command depending on the flag

src/Core/Constants.cs
```diff
@@ -147,6 +147,7 @@ public static class FeatureFlagKeys
     public const string ScimRevokeV2 = "pm-32394-scim-revoke-put-v2";
     public const string BulkReinviteUI = "pm-28416-bulk-reinvite-ux-improvements";
     public const string RefactorOrgAcceptInit = "pm-33082-refactor-org-accept-init";
+    public const string AdminResetTwoFactor = "pm-15489-admin-reset-two-factor";
```
BTreston: Feature flag key recently added in LD is `pm-15489-reset-two-factor-account-recovery`

BTreston (changes requested):

src/Api/Models/Request/Organizations/OrganizationUserResetPasswordRequestModel.cs
ike-kottlowski, line 1: 🎨 : should this class be renamed to `OrganizationUserRecoverAccountRequestModel`? It no longer _just_ resets a password.

src/Core/MailTemplates/Handlebars/AdminResetPassword.html.hbs
ike-kottlowski, line 1: 🎨 : When this is migrated to the MJML pattern, this email should be renamed to better reflect it's more general purpose of account recovery instead of the specific `ResetPassword`.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationUsersControllerPutResetPasswordTests.cs
ike-kottlowski, line 1: 🎨: In the future I imagine the file will be renamed to match the new generic `PutRecoverAccount` endpoint.

That being said, I think it would be pertinent to add a companion integration test suite, either added to this one or in another file, for the newly added Two Factor Recovery flows.

As it stands these integration tests do not test the Feature Flag or any of the Two Factor recovery flows.

However, I defer to @bitwarden/team-admin-console since this is their code space.

ike-kottlowski (dismissed):
The Auth changes are good.

I did however have a peruse and I had styling and nitpick comments mostly around file names not accurately reflecting their true use. I understand with file name changes Git has a tendency to identify the files as new thus making it difficult to review actual changes.

I would have liked to see better integration testing around the Feature Flag and the new Two Factor Recovery.

None of the comments are blocking.

eliykat:
I am OOO for the rest of the week but @BTreston can approve on behalf of AC in the meantime.

BTreston (dismissed):

kspearrin:
Had to fix a conflict. It's on auto-merge now.

BTreston (approved):

ttalty (approved):

ike-kottlowski (approved):

Banrion (approved):

addisonbeck (approved):

PR #7145, opened 2026-03-04
Return WebAuthn credential record in create response

iinuwa:
## 🎟️ Tracking

[PM-33110](https://bitwarden.atlassian.net/browse/PM-33110)

### Background

For the device auth key feature, we need to be able to delete a specific WebAuthn credential record from a mobile device when the user disables the setting or logs out. Right now, we cannot do that because the only ID that the caller has is the credential ID, which is generated by the caller. The delete endpoint only takes record IDs, which is distinct from the FIDO2 credential ID. The record ID is not returned from the create endpoint, so the caller cannot use it.

Currently, the delete method requires you to fetch all the credentials, and then the user can select which one they want to delete based on the name of the credential record, and pass the record ID to the delete endpoint. This works in the web vault, but not in the case of device auth keys on mobile, where it doesn’t make sense to list all the keys.

I would like to be able to delete a specific record by ID directly instead. 

### Changes

This PR modifies the WebAuthn create credential endpoint response to return the created object that is saved in the database, which includes the GUID of the generated record. This allows the mobile device to save the ID locally and use it to delete the credential later.

This is backwards compatible, because the response was empty before (unless a client is asserting the empty response). (For the mobile client implementation, I recommend that we require the model so that we can avoid having an optional field stored on devices forever, so the feature will be unavailable on clients connecting to older servers.)

[PM-33110]: https://bitwarden.atlassian.net/browse/PM-33110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Auth/UserFeatures/WebAuthnLogin/ICreateWebAuthnLoginCredentialCommand.cs
+++ b/src/Core/Auth/UserFeatures/WebAuthnLogin/ICreateWebAuthnLoginCredentialCommand.cs
@@ -1,7 +1,4 @@
-﻿// FIXME: Update this file to be null safe and then delete the line below
-#nullable disable
-
-using Bit.Core.Auth.Entities;
+﻿using Bit.Core.Auth.Entities;
 using Bit.Core.Entities;
 using Fido2NetLib;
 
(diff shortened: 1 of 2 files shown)


ike-kottlowski (commented):
Changes look good.

You mention in the description that this should be fine as long as a client doesn't assert `null`. Can you follow up on that to ensure this change won't cause failures on the web client?

iinuwa:
> You mention in the description that this should be fine as long as a client doesn't assert null. Can you follow up on that to ensure this change won't cause failures on the web client?

Sorry, I should have been clear: I was referring to third-party clients there.

The web client does not assert an empty body, just an HTTP success code: https://github.com/bitwarden/clients/blob/main/apps/web/src/app/auth/core/services/webauthn-login/webauthn-login-admin-api.service.ts#L43-L46. I have also confirmed locally that the changes work with the production client.

PR #7155, opened 2026-03-04
Existing device scene

MGibson1:
## 📔 Objective

Scene to create an existing device. Plan is to use for e2e testing login with device without needing to go through login/logout proceedure.

In the future, we may way to hook this up to some kind of push notification system, but that feels very far down the road

### DIFF
--- a/util/Seeder/Scenes/UserDeviceScene.cs
+++ b/util/Seeder/Scenes/UserDeviceScene.cs
@@ -23,6 +23,7 @@ public class UserDeviceScene(IUserRepository userRepository, IDeviceRepository d
 
     public class Result
     {
+        public Guid DeviceId { get; init; }
     }
 
     public async Task<SceneResult<Result>> SeedAsync(Request request)
@@ -39,6 +40,7 @@ public class UserDeviceScene(IUserRepository userRepository, IDeviceRepository d
         return new SceneResult<Result>(
             result: new Result
             {
+                DeviceId = device.Id,
             },
             mangleMap: manglerService.GetMangleMap());
     }


util/Seeder/Scenes/UserDeviceScene.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Bit.Core.Enums;
+using Bit.Core.Repositories;
+using Bit.Seeder;
+using Bit.Seeder.Factories;
+using Bit.Seeder.Services;
+
```
theMickster, line 6: ⚠️ Will you please add the `namespace Bit.Seeder.Scenes;` to this file?

util/Seeder/Scenes/UserDeviceScene.cs
```diff
@@ -0,0 +1,44 @@
...
+        var user = await userRepository.GetByIdAsync(request.UserId);
+        if (user == null)
+        {
+            throw new Exception($"User with ID {request.UserId} not found.");
+        }
+
+        var device = DeviceSeeder.Create(request.UserId, request.Type, request.Name, request.Identifier, request.PushToken);
+        await deviceRepository.CreateAsync(device);
+
+        return new SceneResult<Result>(
```
theMickster, line 40: ❓Are we missing the device id to be returned here? I would expect it to be returned in the result payload.

util/Seeder/Scenes/UserDeviceScene.cs
```diff
@@ -0,0 +1,44 @@
...
+            result: new Result
+            {
+            },
+            mangleMap: manglerService.GetMangleMap());
```
theMickster, line 45: ❓Are we missing something that's supposed to be mangled? We inject it and get the map, but nothing appears to be mangled. It matched the pattern of the other scenes but isn't used.

util/Seeder/Factories/DeviceSeeder.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Bit.Core.Enums;
+
+namespace Bit.Seeder.Factories;
+
+internal static class DeviceSeeder
+{
+    internal static Core.Entities.Device Create(Guid userId, DeviceType deviceType, string deviceName, string identifier, string? pushToken)
```
theMickster: ⚠️ Will you please replace the two instances of `Core.Entities.` with a using block at the top of the file? Thank you, sir
`using Bit.Core.Entities;`

util/Seeder/Scenes/UserDeviceScene.cs
MGibson1: I can include it, but I don't think it's used for anything. It's all keyed by the userId and identifier that's supplied by the client.

util/Seeder/Factories/DeviceSeeder.cs
MGibson1: Added in 87b26aee6f83076fcb954a97c418a9e185c4303e

util/Seeder/Scenes/UserDeviceScene.cs
MGibson1: added in 243ac6be7989853cbbde7ad1d3be37eca26bfa47

util/Seeder/Scenes/UserDeviceScene.cs
MGibson1: added in 7338ac7ab28015d23c9da517ba7d909f2ed46348

util/Seeder/Scenes/UserDeviceScene.cs
MGibson1: I actually think we don't want to mangle anything. The only unique index on the table is combined UserId and Identifier.

We _could_ mangle the identifier, but that means that the legitimate use-case of creating the same device for multiple userIds would be impossible using this scene. That can be handled with a userIds array, but that doesn't really match how the clients work either.

What do you think is the best approach?

As for still returning a mangle map, it's a required part of the result from any scene, since we want to be able to update this independently from calling code.

theMickster: _Ahhh_, gotcha. The _it's a required part of the result from any scene_ is that part of our code that feels awkward and is unclear here without additional context (like a comment but that would then add bloat + negate efforts to write clean code). 

How could we alter our code to ensure we respect that the `SceneResponseModel` requires a dictionary of strings (e.g. the mangle map) but avoid forcing a Scene (or any object for that matter) to consume the mangler that it does not use? 🤔  How many Scenes like this do we have where the client (the e2e test) is telling the Seeder exactly what to create and we just send back a 200 with an empty response? 

I agree that we don't want to mangle in this scene just because the others mangle. We should only be mangling things that must be unique (like our user emails).

util/Seeder/Scenes/UserDeviceScene.cs
theMickster: I am actually okay with an empty response to the e2e test. As I alluded to in my comment about the mangleMap, it was just a little confusing seeing the `manglerService` not mangling and an empty response.

util/Seeder/Scenes/UserDeviceScene.cs
theMickster: Maybe we just have an common empty result return type for these Scenes that's reusable so that anytime anyone sees it we instantly know what's going on without any guessing or deeper reading into the code?

MGibson1: I don't see requiring the service as that big a downside, but you could have a response generator that will do it for you. You're just swapping requiring the `ManglerService` for whatever this response generation service is called, though.

What I don't want is a static method that converts a Result object into a `SceneResult`, since that minor convenience just makes it easier to forget to return mangled strings

PR #7157, opened 2026-03-05
PM 31023 - Creating density models for Seeder

theMickster:
**Not Planning to Publish because this is too large. Really just wanted to get the whole thing working and some eyes on it before breaking it down** 

## 🎟️ Tracking

[PM-31023 - Relational Density Modeling](https://bitwarden.atlassian.net/browse/PM-31023)
[PM-32777 - Baked-In Density Preset Profiles](https://bitwarden.atlassian.net/browse/PM-32777)

## 📔 Objective

Complete the density modeling additions to the Seeder in our presets.  The work represents a sizable shift in the way the Seeder was first allocating entities as it created them.  By leveraging the new JSON density property, we can now make precise adjustments to allocate entity distribution without changing the Seeder. 

### Key changes

- Add 5 density distributions to preset density block
- Create 9 production-calibrated scale presets (XS-XL)
- Fix Hamilton apportionment bug in Distribution
- Reorganize presets into purpose-based folders
- Consolidate docs into Seeds/docs/ with cross-refs
- Add Q5-Q8 verification queries for new distributions
- Deprecate wonka-teams-small and large-enterprise in favor of the production scale presets

<details><summary> Note on the Hamilton apportionment bug</summary>

The Distribution<T>.Select() method divides items into percentage-based buckets using integer truncation, which leaves unclaimed remainder items. The old code silently dumped all remainder onto the last bucket — so a zero-weight HidePasswords bucket would still receive items. The fix uses Hamilton apportionment (largest-remainder method): remainder items go one-at-a-time to whichever buckets lost the most from truncation, and zero-weight buckets are guaranteed to receive exactly zero.

Alexander Hamilton — the first U.S. Secretary of the Treasury. He proposed this method in 1792 to apportion congressional seats among states. The math problem is the same: distribute a fixed number of indivisible items (seats, or in our case collection permissions) proportionally across groups when the proportional shares aren't whole numbers.
</details>

### Where did our distribution statistics come from?

The scale preset archetypes are modeled after three real production organizations analyzed in DBOPS-91: Company A (hierarchical, 2,795 users/74 groups), Company B (flat, 11,491 users/5 groups/13,906 collections), and Company C (balanced, 954 users/99 groups). These profiles revealed that production relationship patterns follow power-law and mega-group distributions — not the uniform round-robin the seeder previously generated. Each scale preset's density parameters (membership skew, collection fan-out, permission weights, orphan rates) were calibrated to reproduce these observed production shapes at five tiers from family (6 users) to mega-corp (10,000 users).

### Why the re-organization or presets?

The seeder is still early-adoption — breaking preset names now costs nearly nothing, but doing it after teams build scripts around them has cost. Purpose-based folders (features/qa/scale/validation) make preset discovery self-documenting so engineers don't need to read a README to find the right one. Consolidating docs into Seeds/docs/ eliminates duplication across scattered READMEs and separates everyday usage from developer-only verification content.

## 🧪 Testing

<details><summary> Expand for detailed instructions</summary>

### Step 1: Verify preset resolution (all 4 folders)

From util/SeederUtility/, run one preset from each folder:

```bash
dotnet run -- seed --preset features.sso-enterprise --mangle
dotnet run -- seed --preset qa.enterprise-basic --mangle
dotnet run -- seed --preset scale.sm-balanced-planet-express --mangle
dotnet run -- seed --preset validation.density-modeling-power-law-test --mangle
```

All four should seed successfully with no errors.

### Step 2: Verify density distributions on a scale preset

Seed a mid-tier and large-tier preset:

```bash
dotnet run -- seed --preset scale.md-balanced-sterling-cooper --mangle
dotnet run -- seed --preset scale.lg-highperm-tyrell-corp --mangle
```

After each, run the verification queries from util/Seeder/Seeds/docs/verification.md against your local MSSQL database. Compare results to the expected-value tables in the same doc.

#### Key things to verify

- Q1: Group membership follows power-law decay (not uniform)
- Q3: Permission percentages match configured weights
- Q4: Orphan cipher count matches configured rate
- Q7: Collections-per-user shows min/max spread (not flat 1-2-3)
- Q8: Multi-collection ciphers present at configured rate

### Step 3: Verify backward compatibility

Seed the no-density validation preset:

```bash
dotnet run -- seed --preset validation.density-modeling-no-density-test --mangle
```

#### Key things to verify

- 0 CollectionGroup records
- uniform round-robin group membership
- every cipher assigned to at least one collection. 

This confirms the null-density path is unchanged.

### Claude Code prompt for verification

**Note**: Mick has a `reading-bw-mssql` skill that automates the pwsh/SqlClient connection pattern. If you'd like it for your Claude Code setup, ask him to share it.

If you'd like Claude Code to run the verification queries for you, use this prompt after seeding:

<blockquote>
1. Read util/Seeder/Seeds/docs/verification.md for the SQL queries.
2. Run Q1 through Q8 against org ID '{paste-org-id-here}' using pwsh with $env:BW_READ_ONLY_MSSQL_CONNECTION_STRING. 
3. Present results as markdown tables and compare against the expected values for {preset-name} in the verification doc.
</blockquote>
</details>

### DIFF
--- a/test/SeederApi.IntegrationTest/DensityModel/MultiCollectionAssignmentTests.cs
+++ b/test/SeederApi.IntegrationTest/DensityModel/MultiCollectionAssignmentTests.cs
@@ -1,5 +1,4 @@
-using Bit.Core.Entities;
-using Xunit;
+﻿using Xunit;
 
 namespace Bit.SeederApi.IntegrationTest.DensityModel;
 


util/Seeder/Seeds/docs/verification.md
```diff
@@ -0,0 +1,355 @@
...
+    FROM [dbo].[CollectionCipher] CC WITH (NOLOCK)
+    INNER JOIN [dbo].[Cipher] CI WITH (NOLOCK) ON CC.CipherId = CI.Id
+    WHERE CI.OrganizationId = @OrgId
+    GROUP BY CC.CipherId
+) T;
+```
+
+---
+
+## Scale Preset Expected Values
```
nthompson-bitwarden, line 187: I wonder if this adds a documentation burden. What if these expected values change? It may be more effective documentation-wise to provide a high level description of the shapes and diagrams, similar to what is shown in PM-31023. Then link to the presets, so readers can see _how_ those high data distribution strategies get mapped to each preset

util/Seeder/Seeds/docs/presets.md
```diff
@@ -0,0 +1,82 @@
...
+|--------|------|-----------------|-------------|--------|---------|
+| sso-enterprise | enterprise-annually | SSO (OIDC, masterPassword) + requireSso policy | verdant-health | starter-team | sso-vault |
+| tde-enterprise | enterprise-annually | SSO (OIDC, trustedDevices/TDE) + requireSso policy | obsidian-labs | starter-team | tde-vault |
+| policy-enterprise | enterprise-annually | All policies except requireSso and require2fa | pinnacle-designs | starter-team | — |
+
+`policy-enterprise` has no ciphers — it exists purely for testing policy enforcement.
+
+## QA
+
+Handcrafted fixture data for visual UI verification. Known users, groups, collections, and permissions you can point at in the web vault.
```
nthompson-bitwarden, line 23: This is overly verbose - QA does not only do visual UI verification with importable data. Maybe just, "Known users, groups, collections, and permissions you can point a client to." is a better representation

util/Seeder/Seeds/docs/presets.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Scale
+
+Production-calibrated presets with density modeling. Realistic relationship patterns (group membership, collection fan-out, permission distribution, cipher assignment) across 5 tiers.
+
+```bash
+dotnet run -- seed --preset scale.{name} --mangle
+```
+
+| Preset | Tier | Archetype | Users | Groups | Collections | Ciphers | Plan |
```
nthompson-bitwarden, line 47: This documentation is duplicated by verification.md. May be a nothingburger - but any updates to this need to then be updated in the verification.md too

util/Seeder/Seeds/docs/presets.md
```diff
@@ -0,0 +1,82 @@
+# Preset Catalog
+
+Complete catalog of all seeder presets, organized by purpose. Use `--mangle` to avoid collisions with existing data.
```
nthompson-bitwarden, line 3: random note here (probably out of scope)- is there a situation where a user would _not_ use --mangle?

theMickster: So I thought about this and, frankly, I am not so sure anymore 🤷🏼 Perhaps we look to invert the flag and, instead, it's a --no-mangle.

Say we're in a _near perfect_ state in Q4 of 2026 with test automation, would you foresee it making more sense that we always use the mangle feature? 🤔

util/Seeder/Seeds/docs/presets.md
theMickster: I'll work to trim these down when I break the work into 3-4 PRs shortly.

util/Seeder/Seeds/docs/presets.md
theMickster: 10-4. Claude got a little excited and I didn't trim that.

util/Seeder/Seeds/docs/verification.md
theMickster: I had Claude add this because I found it more than helpful for Claude to do the verifications for me as we tested. It could do the math of where we were slightly off from expected; and then it went and fixed the various loops/distributions until it was correct. 

Maybe chat about this topic this week?

util/Seeder/Seeds/docs/presets.md
nthompson-bitwarden: I do foresee that. I imagine a "--no-mangle" would be for more rare "one-off" situations. Whereas, we would want most situations to be using mangle to prevent any confusion that might arise if someone did not have mangled data and did end up having data collisions

PR #7158, opened 2026-03-05
[PM-33144] Fix Send password being removed on edit

mandreko-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33144

## 📔 Objective

In the case that a password-protected Send is edited (name, etc), this fixes a scenario where the Send has its password protection removed.


harr1424:
Closing in favor of the implementation proposed in https://github.com/bitwarden/server/pull/7160

PR #7159, opened 2026-03-05
[PM-29129] Add Policy Update Event README

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29129

## 📔 Objective

1. Add README so devs and AI can easily implement new handlers to hook into Policy Update Events.

## 📸 Screenshots

No code changes, just a markdown file.


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+This directory contains the interfaces and infrastructure for the policy save workflow used by `IVNextSavePolicyCommand`.
+Currently, we’re using `IVNextSavePolicyCommand` to transition from the old `IPolicyValidator` pattern.
+
+---
+
+## Overview
+
+When an organization policy is created or updated, the save workflow runs a series of ordered steps. Each step acts like a hook that a handler may listen to by implementing the particular policy event interface.
+
+Note: If you don’t want to hook into these events, you don’t need to create a handler, and your policy will simply upsert to the database with log events.
```
eliykat: I think the terminology could be clearer here: hook, handler, event, etc.

I suggest:
- the policy getting saved is the event
- the class that does something is a handler

Therefore: `IEnforceDependentPoliciesHandler`, `IPolicyValidationHandler`, etc.

The current wording makes it sound like the class is the event. But I think the class is the thing that handles the event (which is also how you describe it here).

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+```
+SaveAsync()
+  ├─ 1. Validate org can use policies
```
eliykat: Avoid contractions:


```suggestion
  ├─ 1. Validate organization can use policies
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+The `PolicyEventHandlerHandlerFactory` resolves the correct handler for a given `PolicyType` and interface at each step. A handler is matched by its `IPolicyUpdateEvent.Type` property. At most one handler of each interface type is permitted per `PolicyType`.
+
+---
+
+## Limitations
+
+1. We don't have a way to keep this whole process idempotent, so if there is an exception at any point that is not being handled, the state will stay where the process failed.
+
+
+
```
eliykat, line 31: Section appears to be unfinished? Also, this is a good callout, but I think it's more about it not being atomic than idempotent. A failure does not roll back any changes already made.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+The base interface that all policy event handlers must implement.
+
+```csharp
+public interface IPolicyUpdateEvent
+{
+    PolicyType Type { get; }
+}
+```
+
+Every handler declares which `PolicyType` it handles via `Type`. All other event interfaces extend this one.
```
eliykat: I suggest this can be omitted given that no other team will ever have to implement this directly.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+```csharp
+public interface IEnforceDependentPoliciesEvent : IPolicyUpdateEvent
+{
+    IEnumerable<PolicyType> RequiredPolicies { get; }
+}
+```
+
+- **Enabling**: Each `PolicyType` in `RequiredPolicies` must already be enabled, otherwise a `BadRequestException` is thrown.
+- **Disabling a required policy**: If any other policy has this policy listed as a requirement and is currently enabled, the disable action is blocked.
```
eliykat: Suggestion: these lines are not required, you've already explained it above.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+Executes side effects **before** the policy is upserted to the database.
+
+```csharp
+public interface IOnPolicyPreUpdateEvent : IPolicyUpdateEvent
+{
+    Task ExecutePreUpsertSideEffectAsync(SavePolicyModel policyRequest, Policy? currentPolicy);
+}
+```
+
+Typical uses: revoking non-compliant users, removing emergency access grants.
```
eliykat: Removing EA is not a typical use case and is probably not a very clear example.

```suggestion
Typical uses: revoking non-compliant users.
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+}
+```
+
+Return an empty string to pass validation. Return a non-empty error message to throw a `BadRequestException` and abort the save.
+
+---
+
+### `IOnPolicyPreUpdateEvent`
+
+Executes side effects **before** the policy is upserted to the database.
```
eliykat:
```suggestion
Executes side effects **before** the policy is upserted to the database. If an exception is thrown, the policy will not be saved.
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+    Task ExecutePostUpsertSideEffectAsync(
+        SavePolicyModel policyRequest,
+        Policy postUpsertedPolicyState,
+        Policy? previousPolicyState);
+}
+```
+
+Typical uses: creating collections, sending notifications that depend on the new policy state.
+
+Note: This is more useful for enabling a policy than for disabling a policy, since when the policy is disabled, there is no easy way to find the users the policy should be enforced on.
```
eliykat: I think this note is confusing/unnecessary. You wouldn't be enforcing a policy that has been disabled.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+### `IPolicyEventHandlerFactory`
+
+Resolves the correct handler for a given `PolicyType` and event interface type.
+
+```csharp
+OneOf<T, None> GetHandler<T>(PolicyType policyType) where T : IPolicyUpdateEvent;
+```
+
+Returns the matching handler, or `None` if the policy type does not implement the requested interface. Throws `InvalidOperationException` if more than one handler is registered for the same `PolicyType` and interface.
```
eliykat, line 120: Probably not required here as it's internal.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+---
+
+## Adding a New Policy Handler
+
+1. Create a class in `PolicyValidators/` implementing `IPolicyUpdateEvent` and any combination of the event interfaces above.
```
eliykat: You don't need to implement `IPolicyUpdateEvent` because it's implemented by one of those other interfaces.


```suggestion
1. Create a class in `PolicyValidators/` implementing any combination of the event interfaces above.
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+2. Set `Type` to the appropriate `PolicyType`.
+3. Register the class as `IPolicyUpdateEvent` (and the legacy interfaces if needed) in `PolicyServiceCollectionExtensions.AddPolicyUpdateEvents()`.
```
eliykat: Are we still using the legacy interfaces - is there any need to register them?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+Note: No changes to `VNextSavePolicyCommand` or `PolicyEventHandlerHandlerFactory` are required.
+
+### Example
+
+`AutomaticUserConfirmationPolicyEventHandler` is a good reference. It requires `SingleOrg`, validates org compliance before enabling, and removes emergency access grants as a pre-save side effect.
```
eliykat:
```suggestion
`AutomaticUserConfirmationPolicyEventHandler` is a good reference. It requires `SingleOrg`, validates organization compliance before enabling, and removes emergency access grants as a pre-save side effect.
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+It must extend `IPolicyUpdateEvent`.
+
+### Step 2: Add a step to `SavePolicyCommand.SaveAsync()` or `VNextSavePolicyCommand.SaveAsync()` during transition
+
+1. Call your method at the appropriate position in the workflow
+2. You can use the existing `ExecutePolicyEventAsync<T>` helper or have your method use `policyEventHandlerFactory` directly to retrieve the handlers.
+3. **Note on cross-policy logic:** `IEnforceDependentPoliciesEvent` is a special case. It scans *all* registered handlers (not just the targeted policy's handler) to find dependents when disabling a policy. If your new interface requires similar cross-policy scanning, you will need to add that logic directly to `SavePolicyCommand` or `VNextSavePolicyCommand.SaveAsync()` during transition rather than using `ExecutePolicyEventAsync<T>`.
+
+### Step 3: Document the interface in the [Interfaces](#interfaces) section of this README and add it to the workflow diagram.
```
eliykat, line 217: I don't expect this to happen often if ever - the current interfaces are fairly comprehensive. I think we could omit this just so the documentation stays focused on adding new policy types.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+---
+
+# IPolicyValidator (Legacy)
```
eliykat: This has been deprecated for a while. When can we get rid of it?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -0,0 +1,258 @@
+# IPolicyUpdateEvent
+
+This is the policy update pattern that we want our system’s end state to follow.
+This directory contains the interfaces and infrastructure for the policy save workflow used by `IVNextSavePolicyCommand`.
+Currently, we’re using `IVNextSavePolicyCommand` to transition from the old `IPolicyValidator` pattern.
```
eliykat: You have a separate section at the end to discuss the old pattern. I suggest this intro paragraph can just be focused on the new pattern.

eliykat (changes requested):
Thanks for coming back to this!

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: I know we're in the process of getting rid of it, and I think it will be done soon. But since it's still in the codebase, I added it to the doc to give more context because removing it is pretty trivial. It's also not something I feel strongly about, so I'm happy with not including it as well.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: Good catch. Also, this is completed. This is the only issue I'm aware of. I'm happy to add more if you know of anything else. I used a bullet list so we can easily add new ones.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: Isn't this normal expected behavior, where if the previous step fails, it wouldn't move to the next one? Therefore, if it fails here, it won't save the policy.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: Yes, which makes this event only relevant to enabling, not disabling. That's why I thought explicitly stating it here gives more context on how to use it, since the system allows you to add it for disabling as well.

But we can remove it if it causes more confusion.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: My intention for this doc is to explain how this whole system works, so I feel like it's relevant. Do you want to narrow the scope to just how to add a new policy handler? If so, we can create another doc or reorganize the sections to explain how the whole system works.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: This feels relevant to the other [discussion](https://github.com/bitwarden/server/pull/7159#discussion_r2905871539). It seems like we should narrow the scope to "adding new policy types," but I feel like this section is important since we can't predict the future. Do you think a separate doc for the whole system would be better, or should we better organize this existing one?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: We're still in a transition state, so right now we need to support both, in my opinion. But it looks like the [epic](https://bitwarden.atlassian.net/browse/PM-25322) is almost done, so maybe I can wait until it's complete before merging this doc (without the legacy pattern)? This will make it cleaner, and we don't have to revisit it later.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: The outcome of this [discussion](https://github.com/bitwarden/server/pull/7159#discussion_r2905966256) will affect this.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
JimmyVo16: Yeah, unfortunately "event" is an overloaded term. The interfaces are called "Policy Update Event," so it would make sense to use "event" for them, but I can make sure to always add the adjective "Policy Update" to avoid confusion. There is also an audit log event, and I make sure to always include an adjective for that as well.

As for handlers, we can add the adjective "policy" (e.g., "policy handler") to reduce confusion.



> the policy getting saved is the event

I'm not sure if this makes sense. It's an HTTP request that comes in, so normally people would call it a "request."

Originally, I was using “hook” interchangeably with “event,” but maybe I can just keep it as a verb to avoid confusion.

How does this sound? I'm open to other suggestions as well.

```
When an organization policy is created or updated, the save workflow runs a series of ordered steps. A policy handler can hook into any step by implementing the corresponding Policy Update Event interface.

Note: If you do not need to hook into any step, you do not need to create a policy handler. The policy will simply upsert to the database with an audit log event.

```

eliykat: That wording looks good to me. I think my confusion with the language is that you write a handler by implementing one of these interface, but the interface isn't called `Handler`, it's called `Event`. It looks like you're implementing an event.

I also agree that event is overloaded; we could avoid it altogether if we wanted.

This is not blocking, just my thoughts. The new wording looks good.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
```diff
@@ -8,13 +8,13 @@ Currently, we’re using `IVNextSavePolicyCommand` to transition from the old `I
 
 ## Overview
 
-When an organization policy is created or updated, the save workflow runs a series of ordered steps. Each step acts like a hook that a handler may listen to by implementing the particular policy event interface.
+When an organization policy is created or updated, the save workflow runs a series of ordered steps. A policy handler can hook into any step by implementing the corresponding Policy Update Event interface.
 
-Note: If you don’t want to hook into these events, you don’t need to create a handler, and your policy will simply upsert to the database with log events.
+Note: If you do not need to hook into any step, you do not need to create a policy handler. The policy will simply upsert to the database with an audit log event.
```
eliykat, line 12: Non-blocking: I misspoke before. Contractions (e.g. "don't") are OK, abbreviations (e.g. "org" instead of "organization") should be minimized.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: OK good point, I see now. Maybe something like this?

> Note: This is more useful for enabling a policy than for disabling a policy, since when the policy is disabled, it does not have any effect.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: That's OK, it's better to get it merged, and we can update it when the old command is deleted.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: Taking another look at this, I think this is OK and well structured. I am most interested in the "how to add a new policy" process, but no harm in documenting the rest of it as well.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: This is OK per my other comment

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: I think it should only be a list if there is more than 1 item. For now it should just be a single paragraph.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: It's the expected behaviour for normal imperative code flow, but here we've built a little system to handle this process, so it's not necessarily obvious how we've designed the error handling. This is alerting the dev to the fact that we will not handle their errors for them, so if they want the update to continue they would have to do that themselves. Or alternatively, it gives them a way of aborting the operation.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyUpdateEvents/README.md
eliykat: I reviewed the epic, looks like we're almost there as you said. Let's keep it in because it's important context.

eliykat:
My comments above are mostly replying to threads inline, you may have to scroll up for proper context.

eliykat:
@JimmyVo16 is there anything blocking this - can we finalize and merge it?

JimmyVo16:
> @JimmyVo16 is there anything blocking this - can we finalize and merge it?

Nope, I just didn’t have time to get back to it before. It’s ready for review now.

eliykat (commented):
My changes have been addressed but the Claude feedback is worth checking before we merge.

eliykat (approved):
Great work, thanks for adding documentation!

PR #7164, opened 2026-03-05
[PM-32829] Cipher Key for unassigned ciphers

nick-livefront:
## 🎟️ Tracking

[PM-32829](https://bitwarden.atlassian.net/browse/PM-32829)

## 📔 Objective

Return the cipher key when fetching ciphers by the organization for EF, without the key SDK decryption fails.

## 📸 Screenshots

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/daab1968-2290-4f2b-bee8-58ca13cf1cbf" />|<video src="https://github.com/user-attachments/assets/80aa96b0-f4a2-46dc-a3dd-504bce3b4da1" />|

[PM-32829]: https://bitwarden.atlassian.net/browse/PM-32829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Infrastructure.EntityFramework/Vault/Repositories/Queries/CipherOrganizationDetailsReadByOrganizationIdQuery.cs
+++ b/src/Infrastructure.EntityFramework/Vault/Repositories/Queries/CipherOrganizationDetailsReadByOrganizationIdQuery.cs
@@ -39,6 +39,7 @@ public class CipherOrganizationDetailsReadByOrganizationIdQuery : IQuery<CipherO
                         RevisionDate = c.RevisionDate,
                         DeletedDate = c.DeletedDate,
                         Key = c.Key,
+                        Reprompt = c.Reprompt,
                         OrganizationUseTotp = o.UseTotp,
                     };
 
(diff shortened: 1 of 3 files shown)


src/Infrastructure.EntityFramework/Vault/Repositories/Queries/CipherOrganizationDetailsReadByOrganizationIdQuery.cs
```diff
@@ -38,6 +38,7 @@ from o in o_g.DefaultIfEmpty()
                         CreationDate = c.CreationDate,
                         RevisionDate = c.RevisionDate,
                         DeletedDate = c.DeletedDate,
+                        Key = c.Key,
```
shane-melton, line 41: 🎨 `Reprompt` is also not being mapped/grabbed here.

shane-melton (changes requested):
We should also update `CipherOrganizationDetailsReadByIdQuery.cs` that is also missing the `Key` mapping

src/Infrastructure.EntityFramework/Vault/Repositories/Queries/CipherOrganizationDetailsReadByOrganizationIdQuery.cs
nick-livefront: Nice catch! [40c0b8e](https://github.com/bitwarden/server/pull/7164/commits/40c0b8eefe55c006d82ccb33c89bed399996b7d7)

test/Infrastructure.EFIntegration.Test/Vault/Repositories/CipherRepositoryTests.cs
shane-melton, line 1: 👍 Thanks for the tests as always!

shane-melton (approved):
🚀

PR #6843, opened 2026-01-14
Update to `IHostBuilder` style

justindbaur:
## 📔 Objective

Update to a style that is not obsoleted in .NET 10. 

https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/10/webhostbuilder-deprecated

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #7171, opened 2026-03-07
[PM-28727] Upgrade to .NET 10

dereknance:
## 🎟️ Tracking

[PM-28727](https://bitwarden.atlassian.net/browse/PM-28727)

## 📔 Objective

This PR upgrades from .NET 8 to 10.

I left most dependencies at their current versions so they can go through the usual Renovate process.  

### Notable Changes

- #7207 With .NET 9, `Base64.IsValid` enforces canonical padding.
- #6843 `WebHostBuilder` is [obsolete](https://learn.microsoft.com/en-us/aspnet/core/breaking-changes/10/webhostbuilder-deprecated?view=aspnetcore-10.0)
- `X509Certificate2` constructors are [obsolete](https://learn.microsoft.com/en-us/dotnet/core/compatibility/cryptography/9.0/x509-certificates)
  - ❗ The now-obsolete constructors allowed X.509, PKCS7, or PKCS12/PFX certs to be passed in. I now attempt to detect PKCS12 or X.509 and call the appropriate loader. I would appreciate a close eye on those changes.
- `NuGetAuditLevel` defaults to `low` with .NET 10, resulting in _any_ dependency with a known vulnerability to block when running `dotnet restore`. I elected to set it to `critical` to only block builds for vulns at that severity, and let the existing Renovate process take care of the rest. ❗ I can be convinced to disable this entirely. 
- `System.Text.Json` 8.0.5 pin removed because [package pruning emits NU1510](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/10.0/nu1510-pruned-references)
- `Microsoft.Extensions.Caching.Memory` 8.0.1 pin removed as Microsoft.Data.SqlClient requires >=9 and was resulting in a package downgrade
- `Microsoft.AspNetCore.HttpOverrides.IPNetwork` and `KnownNetworks` have been [marked as obsolete](https://learn.microsoft.com/en-us/aspnet/core/breaking-changes/10/ipnetwork-knownnetworks-obsolete?view=aspnetcore-10.0)
- .NET 10's configuration binder [now preserves null values](https://learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/10.0/configuration-null-values-preserved) instead of coercing them to empty strings, causing `NullReferenceException` in `.Trim()` calls. Added null-conditional operators in property setters in `GlobalSettings.cs`
- Upgraded SignalR packages, notably `MessagePack` from 2.x to 3.x. The wire format appears compatible for our use case
- New lints required, mostly whitespace

[PM-28727]: https://bitwarden.atlassian.net/browse/PM-28727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Infrastructure.EntityFramework/NotificationCenter/Repositories/Queries/NotificationStatusDetailsViewQuery.cs
```diff
@@ -10,11 +10,9 @@ public class NotificationStatusDetailsViewQuery(Guid userId, ClientType clientTy
...
     public IQueryable<NotificationStatusDetails> Run(DatabaseContext dbContext)
     {
-        var clientTypes = new[] { ClientType.All };
-        if (clientType != ClientType.All)
-        {
-            clientTypes = [ClientType.All, clientType];
-        }
+        var clientTypes = clientType != ClientType.All
+            ? new List<ClientType> { ClientType.All, clientType }
+            : new List<ClientType> { ClientType.All };
```
dereknance, line 15: This was added to fix integration test errors. EF Core 8 appears not to like that the value-type array expression was becoming a `ReadOnlySpan<T>` with .NET 10

test/Core.Test/Utilities/CoreHelpersTests.cs
```diff
@@ -410,7 +410,8 @@ public void TokenIsValid_Success(string unprotectedTokenTemplate, string firstPa
     {
         var protector = new TestDataProtector(string.Format(unprotectedTokenTemplate, CoreHelpers.ToEpocMilliseconds(creationTime)));
 
-        Assert.Equal(isValid, CoreHelpers.TokenIsValid(firstPart, protector, "protected_token", userEmail, id, expirationInHours));
+        var protectedToken = CoreHelpers.Base64UrlEncode(Encoding.UTF8.GetBytes("protected_token"));
+        Assert.Equal(isValid, CoreHelpers.TokenIsValid(firstPart, protector, protectedToken, userEmail, id, expirationInHours));
```
dereknance, line 414: `IDataProtector.Unprotect` requires a valid Base64URL string now.

bitwarden_license/test/Scim.Test/Groups/GetGroupsListQueryTests.cs
```diff
@@ -27,8 +27,7 @@ public async Task GetGroupsList_Success(int count, int startIndex, SutProvider<G
 
         var result = await sutProvider.Sut.GetGroupsListAsync(organizationId, new GetGroupsQueryParamModel { Count = count, StartIndex = startIndex });
 
-        AssertHelper.AssertPropertyEqual(groups.Skip(startIndex - 1).Take(count).ToList(), result.groupList);
-        AssertHelper.AssertPropertyEqual(groups.Count, result.totalResults);
+        Assert.Equal(groups.Count, result.totalResults);
```
dereknance, line 30: This change, and the others like it were a result of `AssertPropertyEqual` performing a comparison of the two lists' `Count` and `Capacity`. With .NET 10 the `Capacity` of one of the list produced by `ToList()` would be slightly larger than the other list, resulting in a failure.

I elected to simply compare `Count` directly instead.

test/Api.Test/SecretsManager/Controllers/AccessPoliciesControllerTests.cs
dereknance, line 1: The changed tests in this file fix a bug where `id` (the Org ID) was `AssertPropertyEqual`'d at the second parameter of `GetManyByOrganizationIdWriteAccessAsync` which is actually the user ID.

I haven't yet exactly nailed down why this bug presents itself with .NET 10.

bitwarden_license/test/Sso.IntegrationTest/Sso.IntegrationTest.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net10.0</TargetFramework>
```
djsmith85: 🎨 Non-blocking:  
Probably just an oversight when these projects got created. Any file that has the TargetFramework updated, should probably be changed to delete the TargetFramework line, so the framework can be picked up via the Directory.BuildProps

bitwarden_license/src/Scim/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ###############################################
 #                 Build stage                 #
 ###############################################
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-alpine3.21 AS build
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:10.0-alpine AS build
```
mimartin12:
```suggestion
FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:10.0-alpine3.23 AS build
```
🤔 I believe we will want to pin to a specific version of alpine, @mandreko-bitwarden any additional thoughts here?

amorask-bitwarden (dismissed):

bitwarden_license/src/Scim/Dockerfile
dereknance: I agree. I inadvertently dropped the alpine version so I'll get that fixed up.

bitwarden_license/test/Sso.IntegrationTest/Sso.IntegrationTest.csproj
dereknance: Good call. Maybe that helps reduce the reviewers required for the .NET 12 upgrade 😆

util/SeederApi/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ###############################################
 #                 Build stage                 #
 ###############################################
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-bookworm-slim AS build
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:10.0-noble AS build
```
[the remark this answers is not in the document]
dereknance: Fixed! e89ee23ea677bab4e4989b85cd1740f21639ae81

bitwarden_license/test/Sso.IntegrationTest/Sso.IntegrationTest.csproj
dereknance: Fixed! 1e030bf1c2dda4de3ff6d992c4d796d67e1f57ed

bitwarden_license/src/Scim/Dockerfile
dereknance: Fixed! 5d24ab0f31223942fa223d560d06bc48ce426bc2

mimartin12 (dismissed):
Dockerfiles LGTM

BTreston (dismissed):
Looks good for AC owned files

prograhamming (dismissed):

dani-garcia (dismissed):

bitwarden_license/src/Sso/Sso.csproj
```diff
@@ -8,9 +8,6 @@
   <PropertyGroup Condition=" '$(RunConfiguration)' == 'Sso' " />
   <PropertyGroup Condition=" '$(RunConfiguration)' == 'Sso-SelfHost' " />
   <ItemGroup>
-    <!-- This is a transitive dependency to Sustainsys.Saml2.AspNetCore2 -->
```
enmande, line 11: 🙏 Thank you for catching and removing this pin.
[Consolidated with 3.0](https://github.com/dotnet/aspnetcore/issues/3756).

enmande (dismissed):
Auth changes LGTM 😻

dani-garcia (approved):

enmande (approved):

amorask-bitwarden (approved):

BTreston (approved):

djsmith85 (approved):
Approving for platform

Directory.Build.props
```diff
@@ -11,6 +11,7 @@
     <Nullable Condition="'$(Nullable)' == '' and '$(IsTestProject)' == 'true'">annotations</Nullable>
     <Nullable Condition="'$(Nullable)' == '' and '$(IsTestProject)' != 'true'">enable</Nullable>
     <TreatWarningsAsErrors Condition="'$(TreatWarningsAsErrors)' == ''">true</TreatWarningsAsErrors>
+    <NuGetAuditLevel>critical</NuGetAuditLevel>
```
mandreko-bitwarden, line 14: I'm not super familiar with this setting, but I saw it's used for Nuget to report vulnerable packages, and defaulted to "low". Are we setting it to "critical" to suppress the low, moderate, and high severity warnings with the assumption that Renovate will take care of it?

bitwarden_license/test/Scim.Test/Groups/GetGroupsListQueryTests.cs
mandreko-bitwarden: Instead of regressing the test, could we do something like this in addition to the `Count`?

```
AssertHelper.AssertPropertyEqual(
            groups.OrderBy(g => g.Name).Skip(startIndex - 1).Take(count).ToList(),
            result.groupList);
```

It's very similar to the original, so I'm not sure if I'm missing something around why the .NET 10 migration caused an issue.

mandreko-bitwarden (changes requested):

Directory.Build.props
dereknance: > Are we setting it to "critical" to suppress the low, moderate, and high severity warnings with the assumption that Renovate will take care of it?

Yes, since there's an established process for vulnerabilities whose severity does not warrant a fix with high urgency, I figured that we would not want to block every every build (local and CI) for those unresolved vulnerability-related upgrades.

bitwarden_license/test/Scim.Test/Groups/GetGroupsListQueryTests.cs
dereknance: Even before .NET 10, the use of `AssertPropertyEqual` wasn't performing any kind of deep comparison of the values contained in the expected and actual lists. It was simply comparing the value of each lists' `Count` and `Capacity` property. That hasn't changed with .NET 10.

In this context, `Capacity` is the total number of elements the internal data structure can hold without resizing[<sup>1</sup>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.capacity?view=net-10.0), so when I saw that `AssertPropertyEqual` was comparing this property of both lists, I chose to reduce the two asserts down to a single assert on the list `Count`s because a difference in `Capacity` has no bearing on the values in the list, nor the `Count`.

1: [`List<T>.Capacity` docs](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.capacity?view=net-10.0)

mimartin12 (approved):

Directory.Build.props
mandreko-bitwarden: I would agree that the default `low` setting may be a bit too noisy for our purposes. However, setting it to `high` would catch any high or critical severity vulnerabilities, which maybe *should* block the build. It would then help trigger either fixing it at that time, or making another PR to fix it, and a rebase to the original PR once merged.

dereknance: I'm cool with that. Would you be okay with me creating a follow-up PR to this one to change this setting to `high` so we can avoid requiring fresh reviews from all involved?

mandreko-bitwarden (approved):

prograhamming (approved):

PR #7173, opened 2026-03-07
[PM-33216] Finalize RequireSsoPolicyRequirement

eliykat:
## 🎟️ Tracking

## 📔 Objective

Finalize the `RequireSsoPolicyRequirement` implementation. We are generally just removing this feature flag for other consumers, but this `PolicyRequirement` is a bit different because it's used every time someone logs in. This means we need to:

1. make sure the query is performant - something we haven't really been pressed on until now
2. keep the feature flag so we have a killswitch if it breaks login functionality

**Context for Auth:** 

This is a minor update to your code as part of finalizing (and removing) the PolicyRequirements feature flag. Your code is now using a vNext endpoint which contains some optimizations specifically for the "hot" login code path. The changes are still feature flagged.

Claude also noticed some unnecessary tests and test mocks, which I have removed while I'm here.

**Context for AC Team and dbops**

For performance - @sven-bitwarden had switched `PolicyRequirementQuery` over to the bulk `PolicyDetails_ReadByUserIdsPolicyType` sproc for both single and bulk queries. I generally support this for simple queries, but here the dbops advice was that this produces a slightly slower result and a significantly more complex query plan. This PR adds a `PolicyDetails_ReadByUserIdPolicyType` sproc for the single user use case. This closely matches the optimizations we already made to the legacy code path, and it should perform equally as well. So after the EDD cycle, we will be left with:

- `PolicyDetails_ReadByUserIdPolicyType` - for a single user
- `PolicyDetails_ReadByUserIdsPolicyType` - for a many users (bulk)

The new sproc is called by the new `IPolicyRequirement.GetAsyncVNext` method which is only consumed by this location. As part of feature flag removal we will make this the standard method.

The single `PolicyDetails_ReadByUserId` sproc we were previously using actually performs far worse than anything else mentioned and will be dropped as part of feature flag cleanup as well.


src/Sql/dbo/Stored Procedures/PolicyDetails_ReadByUserIdPolicyType.sql
```diff
@@ -0,0 +1,77 @@
...
+            OU.[Permissions]
+        FROM
+            [dbo].[OrganizationUserView] OU
+        WHERE
+            OU.[Status] = 0
+            AND OU.[Email] = @UserEmail
+            AND @UserEmail IS NOT NULL
+    ),
+    Providers AS
+      (
```
mkincaid-bw: ⛏️ This CTE is indented more than the rest

eliykat: Should be fixed.

PR #7184, opened 2026-03-09
[PM-30751] - add secure SSRF protection for internal IPs

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30751

## 📔 Objective

Adds centralized SSRF (Server-Side Request Forgery) protection to prevent server-initiated HTTP requests from reaching internal/private/reserved IP ranges when the destination URL is derived from external input.

### Problem

SSRF protection logic existed only in the Icons project (`IPAddressExtension`), and did not cover CGNAT ranges (RFC 6598). Multiple other HTTP clients accepting user-supplied or admin-configured URLs had no IP validation at all.

Additionally, because `SsrfProtectionHandler` is a `DelegatingHandler`, it only validates the initial request URI. When `AllowAutoRedirect = true` (the default on `HttpClientHandler`/`SocketsHttpHandler`), redirected requests are followed internally by the primary handler and never pass back through the SSRF handler. This means an attacker-controlled domain can redirect to internal addresses (e.g., `http://169.254.169.254/latest/meta-data/`) and bypass SSRF protection entirely.

### Solution

**Shared IP validation** — Moved `IsInternal()` logic from `Icons` into `Core/Utilities/IPAddressExtensions.cs` so all projects can use it. Added CGNAT blocking (`100.64.0.0/10`).

**`SsrfProtectionHandler`** — A `DelegatingHandler` that:
- Resolves DNS **before** connecting (closes the gap where redirects could bypass IP checks)
- Validates **all** resolved addresses against the internal IP blocklist
- Blocks with `SsrfProtectionException` if any resolved IP is internal
- Rewrites the request URI to the validated IP while preserving the `Host` header for TLS SNI
- **Handles HTTP redirects manually** (301, 302, 303, 307, 308), validating each redirect hop through the full SSRF check pipeline before following it
- Limits redirect chains to 10 hops
- Blocks redirects to non-HTTP/HTTPS schemes (e.g., `ftp://`)
- Respects RFC 7231 method semantics: 301/302/303 change POST→GET; 307/308 preserve the original method

**One-line opt-in** — `IHttpClientBuilder.AddSsrfProtection()` extension method to wire the handler into any named `HttpClient`. This also **disables `AllowAutoRedirect`** on the primary handler (both `HttpClientHandler` and `SocketsHttpHandler`) so that redirects always pass back through `SsrfProtectionHandler` for validation. This provides defense-in-depth: the handler validates each hop, and the primary handler is prevented from silently following redirects on its own.

### Protected clients

| Client | Risk | Reason |
|--------|------|--------|
| `ChangePasswordUriService` | Highest | User-supplied domain |
| `WebhookIntegrationHandler` | High | Admin-configured URL |
| `DatadogIntegrationHandler` | High | Admin-configured URL |
| `SlackService` | Medium | Integration URLs |
| `TeamsService` | Medium | Integration URLs |
| `Icons` (fetch + redirect) | Existing | Now uses shared code + CGNAT fix |

**Note:** The Icons client already set `AllowAutoRedirect = false` and handled redirects manually via `IconHttpRequest.FollowRedirectsAsync()`, so it was not vulnerable to the redirect bypass. All other clients listed above were vulnerable.

Out-of-scope clients (server-configured URLs not from external input): `BaseIdentityClientService`, `SSOController`, `NotificationHubPushRegistrationService`, `AliveJob`, `HomeController`.

### Changes

- **New:** `src/Core/Utilities/IPAddressExtensions.cs` — shared IP blocklist with CGNAT
- **New:** `src/Core/Utilities/SsrfProtectionHandler.cs` — DNS-resolving delegating handler with redirect-following and per-hop SSRF validation
- **New:** `src/Core/Utilities/HttpClientBuilderSsrfExtensions.cs` — `.AddSsrfProtection()` extension (registers handler + disables `AllowAutoRedirect`)
- **Modified:** `src/Icons/Util/IPAddressExtension.cs` — delegates to shared Core implementation
- **Modified:** `src/Icons/Util/ServiceCollectionExtension.cs` — adds SSRF protection to Icons/ChangePasswordUri clients
- **Modified:** `src/Core/Dirt/EventIntegrations/EventIntegrationsServiceCollectionExtensions.cs` — adds SSRF protection to Webhook, Datadog, Slack, Teams clients
- **New:** `test/Core.Test/Utilities/IPAddressExtensionsTests.cs` — 30 test cases covering all IP ranges + CGNAT boundaries
- **New:** `test/Core.Test/Utilities/SsrfProtectionHandlerTests.cs` — 27 test cases for handler behavior including redirect validation

### Redirect protection test coverage

- Redirect to internal IP (AWS metadata endpoint) → blocked
- Redirect to localhost → blocked
- Multi-hop redirect chain eventually hitting internal IP → blocked
- Redirect to public IP → followed successfully
- Multi-hop redirects through public IPs → followed successfully
- Relative redirects → resolved against original URI and validated
- Non-HTTP scheme redirects (e.g., `ftp://`) → stopped, returns redirect response as-is
- 301/302/303 POST→GET method change
- 307 method preservation

### Future usage

Any new HTTP client that accepts external input can opt in with:
```csharp
services.AddHttpClient("MyClient").AddSsrfProtection();
```

### DIFF
--- a/src/Core/Utilities/SsrfProtectionHandler.cs
+++ b/src/Core/Utilities/SsrfProtectionHandler.cs
@@ -173,8 +173,19 @@ public class SsrfProtectionHandler : DelegatingHandler
             return originalMethod;
         }
 
-        // 301, 302, 303 change POST to GET per RFC 7231
-        return HttpMethod.Get;
+        // 303 always changes to GET per RFC 7231
+        if (statusCode == HttpStatusCode.SeeOther)
+        {
+            return HttpMethod.Get;
+        }
+
+        // 301, 302: only POST changes to GET (historical browser behavior codified in RFC 7231)
+        if (originalMethod == HttpMethod.Post)
+        {
+            return HttpMethod.Get;
+        }
+
+        return originalMethod;
     }
 
     /// <summary>
--- a/test/Core.Test/Utilities/SsrfProtectionHandlerTests.cs
+++ b/test/Core.Test/Utilities/SsrfProtectionHandlerTests.cs
@@ -262,6 +262,40 @@ public class SsrfProtectionHandlerTests
     }
 
     [Theory]
+    [InlineData(HttpStatusCode.MovedPermanently)]  // 301
+    [InlineData(HttpStatusCode.Found)]             // 302
+    public async Task SendAsync_301_302_Redirect_PreservesNonPostMethod(HttpStatusCode redirectCode)
+    {
+        var (client, inner) = CreateClient();
+        var redirectResponse = new HttpResponseMessage(redirectCode);
+        redirectResponse.Headers.Location = new Uri("http://1.1.1.1/final");
+        inner.EnqueueResponse(redirectResponse);
+
+        var request = new HttpRequestMessage(HttpMethod.Head, "http://8.8.8.8/start");
+        var response = await client.SendAsync(request);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        Assert.Equal(2, inner.AllRequests.Count);
+        Assert.Equal(HttpMethod.Head, inner.AllRequests[1].Method);
+    }
+
+    [Fact]
+    public async Task SendAsync_303_Redirect_ChangesNonPostMethodToGet()
+    {
+        var (client, inner) = CreateClient();
+        var redirectResponse = new HttpResponseMessage(HttpStatusCode.SeeOther);
+        redirectResponse.Headers.Location = new Uri("http://1.1.1.1/final");
+        inner.EnqueueResponse(redirectResponse);
+
+        var request = new HttpRequestMessage(HttpMethod.Head, "http://8.8.8.8/start");
+        var response = await client.SendAsync(request);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        Assert.Equal(2, inner.AllRequests.Count);
+        Assert.Equal(HttpMethod.Get, inner.AllRequests[1].Method);
+    }
+
+    [Theory]
     [InlineData(HttpStatusCode.TemporaryRedirect)] // 307
     public async Task SendAsync_307_Redirect_PreservesOriginalMethod(HttpStatusCode redirectCode)
     {


src/Core/Utilities/SsrfProtectionHandler.cs
```diff
@@ -0,0 +1,211 @@
...
+        request.RequestUri = builder.Uri;
+
+        return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Determines the HTTP method to use for a redirect based on the original method and status code.
+    /// </summary>
+    private static HttpMethod GetRedirectMethod(HttpMethod originalMethod, HttpStatusCode statusCode)
+    {
```
Banrion, line 169: :art: **SUGGESTED -- `GetRedirectMethod` changes all methods to GET for 301/302, not just POST**

.NET's own `SocketsHttpHandler` redirect logic changes only POST to GET for 301/302/303 and preserves all other methods (GET stays GET, HEAD stays HEAD, DELETE stays DELETE). This implementation unconditionally returns `HttpMethod.Get` for any non-307/308 redirect, which means a HEAD request that receives a 301 would be changed to GET.

This does not affect the current protected clients (they use GET or POST), but it diverges from standard .NET behavior in a way that could surprise future consumers.

If you want to match the standard behavior:

```csharp
private static HttpMethod GetRedirectMethod(HttpMethod originalMethod, HttpStatusCode statusCode)
{
    // 307 and 308 preserve the original method
    if (statusCode is HttpStatusCode.TemporaryRedirect or (HttpStatusCode)308)
    {
        return originalMethod;
    }

    // 303 always changes to GET per RFC 7231
    if (statusCode == HttpStatusCode.SeeOther)
    {
        return HttpMethod.Get;
    }

    // 301, 302: only POST changes to GET (historical browser behavior codified in RFC 7231)
    if (originalMethod == HttpMethod.Post)
    {
        return HttpMethod.Get;
    }

    return originalMethod;
}
```

Low priority -- the current behavior is safe and correct for existing callers.

jaasen-livefront: @Banrion Thanks for pointing this out! I've adjusted the method and added appropriate test coverage.

PR #7191, opened 2026-03-10
Seeder - Adding density distributions 

theMickster:
## 🎟️ Tracking

[PM-31023 - Relational Density Modeling](https://bitwarden.atlassian.net/browse/PM-31023)
[PM-32777 - Baked-In Density Preset Profiles](https://bitwarden.atlassian.net/browse/PM-32777)

## 📔 Objective

Add 5 density distributions to the Seeder's preset density block and fix the Hamilton apportionment bug in `Distribution<T>`. This is PR 2 of 3 splitting #7157 — PR 1 (#7176, preset reorg) is merged, PR 3 (scale presets + docs) follows.

**New distributions:** collections-per-user, cipher type selection, multi-collection cipher assignment, personal cipher count per user, folder count per user.

**Bug fix:** `Distribution<T>` remainder items now use largest-fraction apportionment instead of dumping on the last bucket. Zero-weight buckets get exactly zero.

<details><summary>More on Hamilton apportionment</summary>

The `Distribution<T>.Select()` method divides items into percentage-based buckets using integer truncation, which leaves unclaimed remainder items. The old code silently dumped all remainder onto the last bucket — so a zero-weight HidePasswords bucket would still receive items. The fix uses Hamilton apportionment (largest-remainder method): remainder items go one-at-a-time to whichever buckets lost the most from truncation.

Alexander Hamilton — the first U.S. Secretary of the Treasury. He proposed this method in 1792 to apportion congressional seats among states. The math problem is the same: distribute a fixed number of indivisible items (seats, or in our case collection permissions) proportionally across groups when the proportional shares aren't whole numbers.
</details>

## 🧪 Testing

<details><summary>Expand for detailed instructions</summary>

### Step 1: Verify backward compatibility

Seed the no-density validation preset:

` dotnet run -- seed --preset validation.density-modeling-no-density-test --mangle`

Verify: 0 CollectionGroup records, uniform round-robin group membership, every cipher assigned to at least one collection. This confirms the null-density path is unchanged.

### Step 2: Verify density distributions
Seed a density validation preset:

`dotnet run -- seed --preset validation.density-modeling-power-law-test --mangle`

Verify: group membership follows power-law decay, permission percentages match configured weights, orphan ciphers present at configured rate.
</details>

### DIFF
--- a/util/Seeder/Pipeline/PresetLoader.cs
+++ b/util/Seeder/Pipeline/PresetLoader.cs
@@ -212,11 +212,6 @@ internal static class PresetLoader
         var identity = cipherTypes.Identity ?? 0;
         var sshKey = cipherTypes.SshKey ?? 0;
 
-        if (login + secureNote + card + identity + sshKey < 0.001)
-        {
-            return null;
-        }
-
         return new Distribution<CipherType>(
             (CipherType.Login, login),
             (CipherType.SecureNote, secureNote),


util/Seeder/Steps/GenerateFoldersStep.cs
```diff
@@ -1,25 +1,26 @@
...
         var userDigests = context.Registry.UserDigests;
-        var distribution = FolderCountDistributions.Realistic;
+        var distribution = density?.FolderDistribution ?? FolderCountDistributions.Realistic;
 
         for (var index = 0; index < userDigests.Count; index++)
         {
             var digest = userDigests[index];
             var range = distribution.Select(index, userDigests.Count);
-            var count = range.Min + (index % Math.Max(range.Max - range.Min, 1));
+            var count = range.Min + (index % Math.Max(range.Max - range.Min + 1, 1));
```
theMickster, line 23: This was a 🐛 fix... `range.Max - range.Min` excluded the max value from the reachable range. With (4, 8) the old formula produced 4-7, never 8. The + 1 makes max inclusive. Same fix applied in `GeneratePersonalCiphersStep.RangeCalculationTests` validates this.

util/Seeder/Steps/CreateCollectionsStep.cs
```diff
@@ -148,14 +147,18 @@ internal int ComputeFanOut(int collectionIndex, int collectionCount, int min, in
         }
     }
 
-    internal static List<CollectionUser> BuildCollectionUsers(
+    internal List<CollectionUser> BuildCollectionUsers(
```
theMickster, line 150: Changed from static to instance because the `new ComputeCollectionsPerUser()` method needs _density to read UserCollectionShape and UserCollectionSkew. Three existing tests updated accordingly... Same assertions, just constructing a step instance first.

util/Seeder/Steps/GenerateCiphersStep.cs
```diff
@@ -95,6 +95,7 @@ public void Execute(SeederContext context)
             {
                 var orphanCount = (int)(count * _density.OrphanCipherRate);
                 var nonOrphanCount = count - orphanCount;
+                var primaryIndices = new int[nonOrphanCount];
```
theMickster, line 98: 📝 The multi-collection loop runs after the primary assignment loop completes. We need to know which collection each cipher was originally assigned to, and the collection index isn't stored on the CollectionCipher entity. Capturing it during the first pass gives O(1) lookup instead of searching back through the list.

util/Seeder/Pipeline/PresetLoader.cs
```diff
@@ -114,7 +120,11 @@ private static void BuildRecipe(string presetName, SeedPreset preset, ISeedReade
 
         if (preset.PersonalCiphers is not null && preset.PersonalCiphers.CountPerUser > 0)
         {
-            builder.AddPersonalCiphers(preset.PersonalCiphers.CountPerUser);
+            builder.AddPersonalCiphers(preset.PersonalCiphers.CountPerUser, density: density);
+        }
+        else if (density?.PersonalCipherDistribution is not null)
+        {
+            builder.AddPersonalCiphers(0, density: density);
```
theMickster, line 127: 📝 The `countPerUser: 0` looks a little strange and that it would skip personal cipher generation entirely, but it doesn't. 

When PersonalCipherDistribution is set on the density profile, the distribution overrides the flat count — each user gets a variable number. The step's early-return guard checks both: `if (countPerUser == 0 && density?.PersonalCipherDistribution is null)`.

util/Seeder/Steps/GeneratePersonalCiphersStep.cs
```diff
@@ -57,4 +72,16 @@ public void Execute(SeederContext context)
         context.Ciphers.AddRange(ciphers);
         context.Registry.CipherIds.AddRange(cipherIds);
     }
+
+    private static int EstimateTotal(int userCount, Distribution<(int Min, int Max)> dist)
```
theMickster, line 76: This is needed because `typeDistribution.Select(globalIndex, total)` needs the exact total cipher count upfront to place each cipher in the correct type bucket. When using a distribution, each user gets a different count, so we can't just multiply `userDigests.Count * countPerUser`. Without pre-computing, the total is wrong and later ciphers overflow into the last bucket (SSHKey) 🤢 🤮. 

This was a real bug we caught in development.

PR #7194, opened 2026-03-10
[PM-32066] - Add Org Ability View

jrmccannon:
## 🎟️ Tracking
[PM-32066](https://bitwarden.atlassian.net/browse/PM-32066)

## 📔 Objective
This adds a dedicated SQL View for the Organization Ability. This includes a new repository method and integration tests validating it correctly maps those fields.  It will be used in future stories.

[PM-32066]: https://bitwarden.atlassian.net/browse/PM-32066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
+++ b/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
@@ -161,7 +161,7 @@ public class OrganizationRepository : Repository<Core.AdminConsole.Entities.Orga
                     Use2fa = e.Use2fa,
                     UseEvents = e.UseEvents,
                     UsersGetPremium = e.UsersGetPremium,
-                    Using2fa = e.Use2fa && e.TwoFactorProviders != null,
+                    Using2fa = e.Use2fa && e.TwoFactorProviders != null && e.TwoFactorProviders != "{}",
                     UseSso = e.UseSso,
                     UseKeyConnector = e.UseKeyConnector,
                     UseResetPassword = e.UseResetPassword,
--- a/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
+++ b/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
@@ -146,7 +146,8 @@ public class OrganizationRepository : Repository<Core.AdminConsole.Entities.Orga
         }
     }
 
-    public async Task<OrganizationAbility> GetAbilityAsync(Guid organizationId)
+#nullable enable
+    public async Task<OrganizationAbility?> GetAbilityAsync(Guid organizationId)
     {
         using (var scope = ServiceScopeFactory.CreateScope())
         {
@@ -181,6 +182,7 @@ public class OrganizationRepository : Repository<Core.AdminConsole.Entities.Orga
                 }).SingleOrDefaultAsync();
         }
     }
+#nullable disable
 
     public async Task<ICollection<Core.AdminConsole.Entities.Organization>> SearchUnassignedToProviderAsync(string name, string ownerEmail, int skip, int take)
     {
--- a/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
+++ b/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
@@ -146,6 +146,42 @@ public class OrganizationRepository : Repository<Core.AdminConsole.Entities.Orga
         }
     }
 
+    public async Task<OrganizationAbility> GetAbilityAsync(Guid organizationId)
+    {
+        using (var scope = ServiceScopeFactory.CreateScope())
+        {
+            var dbContext = GetDatabaseContext(scope);
+            return await GetDbSet(dbContext)
+                .Where(e => e.Id == organizationId)
+                .Select(e => new OrganizationAbility
+                {
+                    Enabled = e.Enabled,
+                    Id = e.Id,
+                    Use2fa = e.Use2fa,
+                    UseEvents = e.UseEvents,
+                    UsersGetPremium = e.UsersGetPremium,
+                    Using2fa = e.Use2fa && e.TwoFactorProviders != null,
+                    UseSso = e.UseSso,
+                    UseKeyConnector = e.UseKeyConnector,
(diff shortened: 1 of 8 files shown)


src/Sql/dbo/Views/OrganizationAbilityView.sql
```diff
@@ -0,0 +1,28 @@
+CREATE VIEW [dbo].[OrganizationAbilityView]
+AS
+SELECT
+    [Id],
+    [UseEvents],
+    [Use2fa],
+    IIF([Use2fa] = 1 AND [TwoFactorProviders] IS NOT NULL AND [TwoFactorProviders] != '{}', 1, 0) AS [Using2fa],
```
JimmyVo16, line 7: Do we need to add this logic here since it already exists in the model object?

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -146,6 +146,44 @@ public async Task<ICollection<OrganizationAbility>> GetManyAbilitiesAsync()
...
+            return await GetDbSet(dbContext)
+                .Where(e => e.Id == organizationId)
+                .Select(e => new OrganizationAbility
+                {
+                    Enabled = e.Enabled,
+                    Id = e.Id,
+                    Use2fa = e.Use2fa,
+                    UseEvents = e.UseEvents,
+                    UsersGetPremium = e.UsersGetPremium,
+                    Using2fa = e.Use2fa && e.TwoFactorProviders != null,
```
JimmyVo16: Looks like this logic missed this part `organization.TwoFactorProviders != "{}"` but we might not need it since the model is handling it.

JimmyVo16 (commented):
Some small questions, but overall it looks good.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
jrmccannon: Switching to use the constructor so its covered in the model.

src/Sql/dbo/Views/OrganizationAbilityView.sql
jrmccannon: It is only built off the Organization with that. This should be evaluated from here as well.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
JimmyVo16: Ok, I see what you mean now.

JimmyVo16 (approved):

src/Sql/dbo/Views/OrganizationAbilityView.sql
```diff
@@ -0,0 +1,28 @@
+CREATE VIEW [dbo].[OrganizationAbilityView]
```
mkincaid-bw, line 1: ❓ Is this view necessary? We already have a view called `OrganizationView` that contains all of the columns in the `Organization` table. Can you just select the columns you need from `OrganizationView`  in your stored procedure?

jrmccannon: This was done according to our [contributing docs](https://contributing.bitwarden.com/contributing/code-style/sql#overview).

mkincaid-bw (approved):

JaredScar (approved):

rkac-bw:
Nice work on the view and per-ID sproc. This is a solid foundation.

One question though: is FusionCache part of this change, like the org abilities example in CACHING.md, or is that happening in a follow-up?

The reason I ask is the current bulk Organization_ReadAbilities only runs around ~1K times per hour. If callers move to per-ID lookups without a cache layer in front of it, that likely turns into something closer to 500–700K DB round trips per hour based on current sync and cipher volume. With FusionCache and stampede protection, that should stay much closer to ~10–15K per hour since you only pay once per unique org per TTL window.

jrmccannon:
> Nice work on the view and per-ID sproc. This is a solid foundation.
> 
> One question though: is FusionCache part of this change, like the org abilities example in CACHING.md, or is that happening in a follow-up?
> 
> The reason I ask is the current bulk Organization_ReadAbilities only runs around ~1K times per hour. If callers move to per-ID lookups without a cache layer in front of it, that likely turns into something closer to 500–700K DB round trips per hour based on current sync and cipher volume. With FusionCache and stampede protection, that should stay much closer to ~10–15K per hour since you only pay once per unique org per TTL window.

FusionCache is a later update we will be making.

rkac-bw:
> > Nice work on the view and per-ID sproc. This is a solid foundation.
> > One question though: is FusionCache part of this change, like the org abilities example in CACHING.md, or is that happening in a follow-up?
> > The reason I ask is the current bulk Organization_ReadAbilities only runs around ~1K times per hour. If callers move to per-ID lookups without a cache layer in front of it, that likely turns into something closer to 500–700K DB round trips per hour based on current sync and cipher volume. With FusionCache and stampede protection, that should stay much closer to ~10–15K per hour since you only pay once per unique org per TTL window.
> 
> FusionCache is a later update we will be making.

Thanks! Makes sense. Is the plan to have FusionCache wired up before the caller migration PRs (#7202,
  #7214, #7240) start landing? Just want to make sure we want to be aware if there will be a window where per-ID calls are hitting the DB uncached.

JimmyVo16:
> Thanks! Makes sense. Is the plan to have FusionCache wired up before the caller migration PRs (https://github.com/bitwarden/server/pull/7202,
https://github.com/bitwarden/server/pull/7214, https://github.com/bitwarden/server/pull/7240) start landing? Just want to make sure we want to be aware if there will be a window where per-ID calls are hitting the DB uncached.

Yes, we’re going to migrate to FusionCache and eventually move away from `Organization_ReadAbilities` completely.

The PRs that you posted will call the new interfaces, but the underlying implementations will still use the existing cache. It will not trigger a database call if they’re not in the cache, since `Organization_ReadAbilities` is called by a scheduler at 10-minute intervals rather than by cache requests, so I don’t think this change will increase the rate of requests.

We have future tickets that will change the implementation to call per request ID and cache it appropriately with FusionCache.

PR #7198, opened 2026-03-11
[PM-32477]PremiumStatusChanged Push Notification

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32477

## 📔 Objective

**Summary**
Implement premium-status push notifications across billing flows so clients stay in sync whenever a user’s premium state changes.

**Details**

**New adapter API for premium changes**
-   Extend IPushNotificationAdapter with NotifyPremiumStatusChangedAsync(User user).
-    Implement in PushNotificationAdapter by delegating to IPushNotificationService.PushPremiumStatusChangedAsync(user).

**Stripe payment and subscription handlers**
   - **PaymentSucceededHandler**
      -  Inject IUserRepository.
      - After EnablePremiumAsync(userId, subscription.GetCurrentPeriodEnd()), load the user and call NotifyPremiumStatusChangedAsync(user) so successful payments trigger a premium-status push.

**SubscriptionDeletedHandler**
- Inject IUserRepository and IPushNotificationAdapter.
- When a user subscription is deleted (userId present), call DisablePremiumAsync and then fetch the user and send NotifyPremiumStatusChangedAsync(user).

**SubscriptionUpdatedHandler**
- Inject IUserRepository and IPushNotificationAdapter.
- In DisableSubscriberAsync user branch: disable premium, fetch the user, then call NotifyPremiumStatusChangedAsync(user).
- In EnableSubscriberAsync user branch: enable premium, fetch the user, then call NotifyPremiumStatusChangedAsync(user).

**Premium subscription commands**
- **CreatePremiumCloudHostedSubscriptionCommand**
    - After saving the user and PushSyncVaultAsync(user.Id), call PushPremiumStatusChangedAsync(user) so new/updated cloud subscriptions immediately push the status change.

- **CreatePremiumSelfHostedSubscriptionCommand**
 - After writing the license and saving the user, call PushPremiumStatusChangedAsync(user) to broadcast the new premium status for self-hosted subscriptions.

- **UpgradePremiumToOrganizationCommand**
  - Inject IPushNotificationService.
  - After downgrading the user’s personal premium while upgrading to an organization, call PushPremiumStatusChangedAsync(user) so clients see the user’s premium flag change.

**Tests**

- **Update billing service tests to:**
  -   Use IUserRepository to return a User instance when subscriptions are enabled/disabled.
  -   Assert IPushNotificationAdapter.NotifyPremiumStatusChangedAsync(user) is called in subscription deleted/updated scenarios.

- **Update premium command tests to:**
  -  Assert IPushNotificationService.PushPremiumStatusChangedAsync is called with a User whose Premium flag matches the expected post-command state (true for new subscriptions, false when upgrading to an organization).

### DIFF
--- a/src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
+++ b/src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
@@ -83,16 +83,21 @@ public class CreatePremiumCloudHostedSubscriptionCommand(
             return new BadRequest("Additional storage must be greater than 0.");
         }
 
-        // Validate coupon if provided. Return error if invalid to prevent charging more than expected.
-        string? validatedCoupon = null;
-        if (!string.IsNullOrWhiteSpace(subscriptionPurchase.Coupon))
+        // Validate all provided coupons. Fail fast if any coupon is invalid to prevent charging more than expected.
+        var validatedCoupons = (subscriptionPurchase.Coupons ?? [])
+            .Where(c => !string.IsNullOrWhiteSpace(c))
+            .Select(c => c.Trim())
+            .ToList();
+
+        if (validatedCoupons.Count > 0)
         {
-            var isValid = await subscriptionDiscountService.ValidateDiscountEligibilityForUserAsync(user, subscriptionPurchase.Coupon.Trim(), DiscountTierType.Premium);
-            if (!isValid)
+            var allValid = await subscriptionDiscountService.ValidateDiscountEligibilityForUserAsync(
+                user, validatedCoupons, DiscountTierType.Premium);
+
+            if (!allValid)
             {
                 return new BadRequest("Discount expired. Please review your cart total and try again");
             }
-            validatedCoupon = subscriptionPurchase.Coupon.Trim();
         }
 
         var premiumPlan = await pricingClient.GetAvailablePremiumPlan();
@@ -127,7 +132,7 @@ public class CreatePremiumCloudHostedSubscriptionCommand(
 
         customer = await ReconcileBillingLocationAsync(customer, subscriptionPurchase.BillingAddress);
 
-        var subscription = await CreateSubscriptionAsync(user.Id, customer, premiumPlan, subscriptionPurchase.AdditionalStorageGb > 0 ? subscriptionPurchase.AdditionalStorageGb : null, validatedCoupon);
+        var subscription = await CreateSubscriptionAsync(user.Id, customer, premiumPlan, subscriptionPurchase.AdditionalStorageGb > 0 ? subscriptionPurchase.AdditionalStorageGb : null, validatedCoupons);
 
         subscriptionPurchase.PaymentMethod.Switch(
             tokenized =>
@@ -307,7 +312,7 @@ public class CreatePremiumCloudHostedSubscriptionCommand(
         Customer customer,
         Pricing.Premium.Plan premiumPlan,
         int? storage,
-        string? validatedCoupon)
+        IReadOnlyList<string> validatedCoupons)
     {
 
         var subscriptionItemOptionsList = new List<SubscriptionItemOptions>
@@ -349,9 +354,11 @@ public class CreatePremiumCloudHostedSubscriptionCommand(
             OffSession = true
         };
 
-        if (!string.IsNullOrWhiteSpace(validatedCoupon))
+        if (validatedCoupons.Count > 0)
         {
-            subscriptionCreateOptions.Discounts = [new SubscriptionDiscountOptions { Coupon = validatedCoupon }];
+            subscriptionCreateOptions.Discounts = validatedCoupons
+                .Select(c => new SubscriptionDiscountOptions { Coupon = c })
+                .ToList();
         }
 
         var subscription = await stripeAdapter.CreateSubscriptionAsync(subscriptionCreateOptions);
--- a/global.json
+++ b/global.json
@@ -6,6 +6,6 @@
   "msbuild-sdks": {
     "Microsoft.Build.Traversal": "4.1.0",
     "Microsoft.Build.Sql": "1.0.0",
-    "Bitwarden.Server.Sdk": "1.4.0"
+    "Bitwarden.Server.Sdk": "1.5.1"
   }
 }
--- a/src/Api/AdminConsole/Models/Request/Organizations/OrganizationCreateRequestModel.cs
+++ b/src/Api/AdminConsole/Models/Request/Organizations/OrganizationCreateRequestModel.cs
@@ -81,6 +81,8 @@ public class OrganizationCreateRequestModel : IValidatableObject
 
     public bool SkipTrial { get; set; }
 
+    public string[] Coupons { get; set; }
+
     public virtual OrganizationSignup ToOrganizationSignup(User user)
     {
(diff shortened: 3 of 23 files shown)


src/Billing/Services/Implementations/PaymentSucceededHandler.cs
```diff
@@ -115,6 +116,8 @@ public async Task HandleAsync(Event parsedEvent)
             }
 
             await userService.EnablePremiumAsync(userId.Value, subscription.GetCurrentPeriodEnd());
+            var user = await userRepository.GetByIdAsync(userId.Value);
+            await pushNotificationAdapter.NotifyPremiumStatusChangedAsync(user!);
```
amorask-bitwarden: ❌ 2 things we should clean up here:
1. `userService.EnablePremiumAsync` makes no guarantees the `User` exists. It just skips the operation if the user is null, which means line 120 could result in an NRE. Not only that, but if the `User` that comes back from the DB is `null`, the `Enable` would be followed by us sending the notification anyway, which is incorrect.
2. This will never be hit for users on the new price because of the missed Price adaptation on line 113 - we should fix that.

src/Core/Platform/Push/IPushNotificationService.cs
```diff
@@ -429,6 +429,21 @@ Task PushRefreshSecurityTasksAsync(Guid userId)
             },
             ExcludeCurrentContext = false,
         });
+
+    Task PushPremiumStatusChangedAsync(Entities.User user)
```
amorask-bitwarden: ❌ We're no longer adding methods like this to the shared `IPushNotificationService`. Please follow the established pattern in the the Billing application's `PushNotificationAdapter` so that we own the code.

src/Billing/Services/Implementations/PushNotificationAdapter.cs
```diff
@@ -68,4 +69,7 @@ public Task NotifyEnabledChangedAsync(Organization organization) =>
             },
             ExcludeCurrentContext = false,
         });
+
+    public Task NotifyPremiumStatusChangedAsync(User user) =>
```
amorask-bitwarden, line 74: ❌ Related to [this comment](https://github.com/bitwarden/server/pull/7198/changes#r2920226293), this should follow the pattern of the methods above.

src/Billing/Services/Implementations/SubscriptionDeletedHandler.cs
```diff
@@ -80,6 +87,8 @@ public async Task HandleAsync(Event parsedEvent)
         else if (userId.HasValue)
         {
             await _userService.DisablePremiumAsync(userId.Value, subscription.GetCurrentPeriodEnd());
+            var user = await _userRepository.GetByIdAsync(userId.Value);
+            await _pushNotificationAdapter.NotifyPremiumStatusChangedAsync(user!);
```
amorask-bitwarden: ❌ Same issue as [here](https://github.com/bitwarden/server/pull/7198/changes#r2920215732).

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -140,7 +143,12 @@ SubscriptionStatus.Incomplete or
 
     private Task DisableSubscriberAsync(SubscriberId subscriberId, DateTime? currentPeriodEnd) =>
         subscriberId.Match(
-            userId => _userService.DisablePremiumAsync(userId.Value, currentPeriodEnd),
+            async userId =>
+            {
+                await _userService.DisablePremiumAsync(userId.Value, currentPeriodEnd);
+                var user = await _userRepository.GetByIdAsync(userId.Value);
+                await _pushNotificationAdapter.NotifyPremiumStatusChangedAsync(user!);
```
amorask-bitwarden: ❌ Same issue as [here](https://github.com/bitwarden/server/pull/7198/changes#r2920215732).

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -162,7 +170,12 @@ private Task DisableSubscriberAsync(SubscriberId subscriberId, DateTime? current
 
     private Task EnableSubscriberAsync(SubscriberId subscriberId, DateTime? currentPeriodEnd) =>
         subscriberId.Match(
-            userId => _userService.EnablePremiumAsync(userId.Value, currentPeriodEnd),
+            async userId =>
+            {
+                await _userService.EnablePremiumAsync(userId.Value, currentPeriodEnd);
+                var user = await _userRepository.GetByIdAsync(userId.Value);
+                await _pushNotificationAdapter.NotifyPremiumStatusChangedAsync(user!);
```
amorask-bitwarden: ❌ Same issue as [here](https://github.com/bitwarden/server/pull/7198/changes#r2920215732).

src/Core/Billing/Premium/Commands/CreatePremiumCloudHostedSubscriptionCommand.cs
```diff
@@ -166,6 +166,7 @@ public Task<BillingCommandResult<None>> Run(
 
         await userService.SaveUserAsync(user);
         await pushNotificationService.PushSyncVaultAsync(user.Id);
+        await pushNotificationService.PushPremiumStatusChangedAsync(user);
```
amorask-bitwarden: ❓ Wouldn't this be redundant? I assume a full sync would re-pull Premium status, right?

src/Core/Billing/Premium/Commands/CreatePremiumSelfHostedSubscriptionCommand.cs
```diff
@@ -61,6 +61,7 @@ public Task<BillingCommandResult<None>> Run(
 
         await userService.SaveUserAsync(user);
         await pushNotificationService.PushSyncVaultAsync(user.Id);
+        await pushNotificationService.PushPremiumStatusChangedAsync(user);
```
amorask-bitwarden: ❓ Same question as [here](https://github.com/bitwarden/server/pull/7198/changes#r2920253701).

src/Core/Models/PushNotification.cs
```diff
@@ -133,3 +133,10 @@ public class AutoConfirmPushNotification
     /// </summary>
     public Guid TargetOrganizationUserId { get; set; }
 }
+
+public class PremiumStatusPushNotification
+{
+    public Guid UserId { get; set; }
+    public bool Premium { get; set; }
+    public DateTime? PremiumExpirationDate { get; set; }
```
amorask-bitwarden: ❓ Is there a client-side use-case for including the expiration date?

src/Core/Models/PushNotification.cs
```diff
@@ -133,3 +133,10 @@ public class AutoConfirmPushNotification
     /// </summary>
     public Guid TargetOrganizationUserId { get; set; }
 }
+
+public class PremiumStatusPushNotification
```
amorask-bitwarden: ❌ Please review the comments on lines 7 and 8 of this file regarding the location this file should live in.

amorask-bitwarden: @cyprain-okeke This still needs to be addressed - the location of the file should be owned to a location we own per the comments at the top of the file.

PR #7201, opened 2026-03-11
[PM-33162] Refactor user key rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33162

## 📔 Objective

The objective of this PR is to refactor the current change password and user key rotation endpoint to use, via composition, a new base data model. This is preparation of adding a new endpoint to support none password change userKey rotation that will share the same base data model and processing logic.

Note to Auth reviewers:
The only auth code changes was moving `MasterPasswordUnlockAndAuthenticationDataModel`  to KM ownership.
`MasterPasswordUnlockAndAuthenticationDataModel` the request model is only used on the KM key rotation endpoint. Let me know if you don't agree and want to still retain ownership.


src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -102,33 +102,40 @@ await _regenerateUserAsymmetricKeysCommand.RegenerateKeysAsync(request.ToUserAsy
 
 
     [HttpPost("key-management/rotate-user-account-keys")]
-    public async Task RotateUserAccountKeysAsync([FromBody] RotateUserAccountKeysAndDataRequestModel model)
+    public async Task PasswordChangeAndRotateUserAccountKeysAsync([FromBody] RotateUserAccountKeysAndDataRequestModel model)
```
Thomas-Avery, line 96: KM team, I've realized changing this method name will effect the SDK name of the generated API bindings. Do we think that is worth it or should I just revert it back to the original name?

quexten: I'm OK with it, it just means we need to fix the breaking API bindings PR. I.e we run the API bindings automation on the SDK repo and do a rename on that branch, and review. Should be fairly low effort?

Patrick-Pimentel-Bitwarden:
1. Could you help me understand why we don't use the unlock and authentication data types separately for key rotation request model?
2. Could we take the time here to update rotate account keys to standardize using the two independent request models (MasterPasswordUnlockData/MasterPasswordAuthenticationData) to make applying validation consistent and bring this endpoint in line with how all the other endpoints that concern unlock and authentication work?

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -102,33 +102,40 @@ await _regenerateUserAsymmetricKeysCommand.RegenerateKeysAsync(request.ToUserAsy
...
+                EmergencyAccesses =
+                    await _emergencyAccessValidator.ValidateAsync(user,
+                        model.AccountUnlockData.EmergencyAccessUnlockData),
+                OrganizationUsers =
+                    await _organizationUserValidator.ValidateAsync(user,
+                        model.AccountUnlockData.OrganizationAccountRecoveryUnlockData),
+                WebAuthnKeys =
+                    await _webauthnKeyValidator.ValidateAsync(user, model.AccountUnlockData.PasskeyUnlockData),
+                DeviceKeys = await _deviceValidator.ValidateAsync(user, model.AccountUnlockData.DeviceKeyUnlockData),
+                V2UpgradeToken = model.AccountUnlockData.V2UpgradeToken?.ToData(),
```
quexten, line 122: Question: If we plan to use only the new endpoint for upgrades anyways, and never this one, then should we even support the V2 token here? Or should we drop it again, and only re-add it to the new token? / Error if it is set here?

Thomas-Avery: End state with the v2 migration flag on, manual rotation either invoked by the "Key rotate" button or the checkbox on the password change would do a v2 upgrade?

The plan was to have support on both.

quexten: Ah, oops. You are right.

src/Api/KeyManagement/Models/Requests/MasterPasswordUnlockDataAndAuthenticationModel.cs
```diff
@@ -45,22 +43,35 @@ public IEnumerable<ValidationResult> Validate(ValidationContext validationContex
...
-            MasterKeyEncryptedUserKey = MasterKeyEncryptedUserKey,
-            MasterPasswordHint = MasterPasswordHint
+            Kdf = new KdfSettings
+            {
+                KdfType = KdfType,
+                Iterations = KdfIterations,
+                Memory = KdfMemory,
+                Parallelism = KdfParallelism,
+            },
+            Salt = Email,
```
ike-kottlowski, line 57: ❓ : Is it possible to add a `public string? MasterPasswordSalt` and for the assignment use a coalesce with the `Email`?

``` csharp
Salt = MasterPasswordSalt ?? Email
```

We are trying to make sure the Email is only used as a fallback and not the primary assignemnt.

ike-kottlowski: per a slack discussion: We are migrating away from this data model toward using `MasterPasswordUnlockResponseModel` which is handled here: https://bitwarden.atlassian.net/browse/PM-30370

Thomas-Avery: Talked to @JaredSnider-Bitwarden about this a bit. I think we need to make sure we're all on the same page. 

The API request model https://github.com/bitwarden/server/blob/main/src/Api/Auth/Models/Request/Accounts/MasterPasswordUnlockDataAndAuthenticationModel.cs currently uses the `Email` property. 

There is two options when it comes to replacing it with `MasterPasswordSalt`: 

- Update the current request model to have a `public string? MasterPasswordSalt`
- Long term KM plans to migrate from this request model to a `MasterPasswordUnlockResponseModel` and a `MasterPasswordAuthenticationDataRequestModel` tracked by [PM-30463](https://bitwarden.atlassian.net/browse/PM-30463)


KM currently doesn't plan to tackle PM-30463 right now and auth is wanting to move forward with adding `public string? MasterPasswordSalt` which KM is good with.

Adding the optional request property is out of scope for this refactor.

Thomas-Avery:
https://github.com/bitwarden/server/pull/7230/changes changed the behavior of `GetMasterPasswordSalt` which effects the `MasterPasswordUnlockData` and `MasterPasswordAuthenticationData`'s `ValidateSaltUnchangedForUser`.  At test runtime `BitAutoData` will optionally fill out `MasterPasswordSalt` which now would take precedent for `GetMasterPasswordSalt` and break expectations in the tests. For the fix just setup all the tests to fill `User.MasterPasswordSalt` as null.

src/Api/KeyManagement/Models/Requests/MasterPasswordUnlockDataAndAuthenticationModel.cs
ike-kottlowski: Optional Param is added in this PR https://github.com/bitwarden/server/pull/7239

PR #7203, opened 2026-03-11
[PM-32885] Fix incorrect behavior for expired sends

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32885

## 📔 Objective

When a Send reached it's deletion date, there was a window of time (up to five minutes) where the Send still existed in the database. There is a cleanup job scheduled to run every five minutes, and during this window users attempting to access such a a Send would receive a randomized error (randomized to prevent enumeration on Send IDs). This caused unexpected behavior in the UI. 

More specifically, auth related logic was returning the same `NeverAuthenticate` type for Send IDs that never existed and for Send IDs that had expired. This PR introduces a `SendInaccessible` result type that is returned whenever a Send exists but can't be accessed (due to expiration, deletion, max access count exceeded, or being disabled by the user) and does not return a randomized error type, instead returning a consistent `send_id_invalid` error. The loss of enumeration protection in this specific case seems appropriate since the Send can no longer be accessed, and in all situations except where a user may re-enable the Send, can not be accessed again in the future. 

The clients already map the `send_id_invalid` error to `unavailable.set(true)`, which shows the correct "Send unavailable" message in the UI.


src/Identity/IdentityServer/RequestValidators/SendAccess/SendInaccessibleRequestValidator.cs
ike-kottlowski, line 1: 🎨 : I think adding the overhead of this RequestValidator is a bit of overkill. I think we can just return

``` csharp
new GrantValidationResult(
            TokenRequestErrors.InvalidGrant,
            SendAccessConstants.SendIdGuidValidatorResults.InvalidSendId,
            customResponse)
```

in the `SendAccessGrantValidator` class for the `SendInaccessible` type.

Since we don't have any special logic here like we do for `NEVER_AUTHENTICATE` the whole file and DI for the above object return feels like too much.

PR #7204, opened 2026-03-12
[PM-33232] Update UX implemented in PM-33219 to resolve VULN-458

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33232

## 📔 Objective

Given a user has an existing 'Who can view' setting configured, when they switch the 'Who can view' option and select Save, then the previous 'Who can view' method is removed and replaced with the newly selected option

Given a user switches the 'Who can view' setting, when they select Save, then the updated setting is reflected immediately in the UI

Given a user switches the 'Who can view' setting, when they select Save, then the updated setting persists after leaving and returning to the screen

Given the above behavior, when tested across web, desktop, and mobile clients, then all clients reflect the same expected behavior

### DIFF
--- a/src/Api/Tools/Models/Request/SendRequestModel.cs
+++ b/src/Api/Tools/Models/Request/SendRequestModel.cs
@@ -3,6 +3,7 @@
 
 using System.ComponentModel.DataAnnotations;
 using System.Text.Json;
+using Bit.Api.Tools.Utilities;
 using Bit.Core.Exceptions;
 using Bit.Core.Tools.Entities;
 using Bit.Core.Tools.Enums;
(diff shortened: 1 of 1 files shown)


src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -247,29 +246,36 @@ private Send ToSendBase(Send existingSend, ISendAuthorizationService authorizati
...
+            AuthType == Core.Tools.Enums.AuthType.Password)
         {
-            existingSend.Password = authorizationService.HashPassword(Password);
-            existingSend.Emails = null;
+            // when password protected Sends are edited, DO NOT re-hash the existing password hash
+            return existingSend;
         }
-        else if (existingSend.AuthType == Core.Tools.Enums.AuthType.Email)
+
+        existingSend.AuthType = AuthType;
```
itsadrago: The older code handled the scenario where no auth type was sent by the client. We have to assume some older clients will still be sending send-related requests

harr1424: Please see 42d99c9300777c764e1e7832335966aa4b634dbe

harr1424: I've also updated the exception message to match the behavior change in 3c6e23fa700939fbd2bccd9326fedbdfd350e0f9

itsadrago: What if an older client is setting the password or email verification list (i.e. `Emails` or `Password` is set in the model), the auth type would be null and so the emails or password information would be lost

harr1424: I've made changes to use the presence of `AuthType` in the request to conditionally execute the new logic. 92d1a41dcfad82539d9fef9d1bdf6eb67576d9a7

itsadrago (approved):

PR #7208, opened 2026-03-13
[PM-33500] - delete attachments from deleted ciphers

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33500

## 📔 Objective

Fix orphaned attachment files left behind when ciphers are deleted in bulk, an organization vault is purged, or an organization is deleted.

When a single cipher is deleted via `CipherService.DeleteAsync`, attachment files are properly removed from blob/disk storage via `DeleteAttachmentsForCipherAsync`. However, three other deletion paths only removed DB records, leaving attachment files permanently orphaned in storage:

1. **`CipherService.DeleteManyAsync`** (bulk cipher delete) — deleted cipher rows but never cleaned up attachment storage.
2. **`CipherService.PurgeAsync`** (org vault purge) — deleted cipher rows but never cleaned up attachment storage.
3. **`OrganizationDeleteCommand.DeleteAsync`** (org deletion) — cascade-deleted cipher rows via SQL stored proc but never cleaned up attachment storage.

### Changes

- **`CipherService.DeleteManyAsync`**: After DB deletion, iterate deleted ciphers and call `DeleteAttachmentsForCipherAsync` for each.
- **`CipherService.PurgeAsync`**: Fetch org cipher IDs before DB deletion, then clean up each cipher's attachments after.
- **`OrganizationDeleteCommand.DeleteAsync`**: Inject `ICipherRepository` and `IAttachmentStorageService`. Fetch org cipher IDs before DB cascade delete, then clean up each cipher's attachments after.
- Updated unit tests for `DeleteManyAsync` (org admin + personal owner paths) and `OrganizationDeleteCommand.Delete_Success` to assert attachment cleanup. Added new `PurgeAsync_WithOrganizationId_DeletesCiphersAndAttachments` test.


JaredScar (approved):

src/Core/AdminConsole/OrganizationFeatures/Organizations/OrganizationDeleteCommand.cs
```diff
@@ -43,8 +50,16 @@ public async Task DeleteAsync(Organization organization)
             catch (GatewayException) { }
         }
 
+        // Fetch cipher IDs before DB deletion so we can clean up attachment storage
+        var orgCiphers = await _cipherRepository.GetManyByOrganizationIdAsync(organization.Id);
```
gbubemismith: I think we can improve this a bit. We can just filter and get ciphers with attachments and only return the ciphers Ids 

```suggestion
      var cipherIdsWithAttachments =
    (await _cipherRepository.GetManyByOrganizationIdAsync(organization.Id))
    .Where(c => c.GetAttachments()?.Count > 0)
    .Select(c => c.Id);
```

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -494,7 +500,17 @@ public async Task PurgeAsync(Guid organizationId)
         {
             throw new NotFoundException();
         }
+
+        // Fetch cipher IDs before DB deletion so we can clean up attachment storage
+        var orgCiphers = await _cipherRepository.GetManyByOrganizationIdAsync(organizationId);
```
gbubemismith: We can do same filtering here and only get ciphers with attachments and select the cipher ids.

gbubemismith (commented):
Added a few suggestions

src/Core/AdminConsole/OrganizationFeatures/Organizations/OrganizationDeleteCommand.cs
```diff
@@ -43,8 +50,16 @@ public async Task DeleteAsync(Organization organization)
...
+
         await _organizationRepository.DeleteAsync(organization);
         await _applicationCacheService.DeleteOrganizationAbilityAsync(organization.Id);
+
+        foreach (var cipher in orgCiphers)
+        {
+            await _attachmentStorageService.DeleteAttachmentsForCipherAsync(cipher.Id);
```
JaredScar: This is fine, but I think it'd be nicer if we followed the format above and added a method such like DeleteAttachmentsForOrganization then supply only the organization ID like we do for the above applicationCacheService. 

The for loop can then live in that method and this file becomes a bit cleaner

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -460,6 +458,12 @@ public async Task DeleteManyAsync(IEnumerable<Guid> cipherIds, Guid deletingUser
             await _cipherRepository.DeleteAsync(deletingCiphers.Select(c => c.Id), deletingUserId);
         }
 
+        // Clean up attachment files from storage
+        foreach (var cipher in deletingCiphers)
+        {
+            await _attachmentStorageService.DeleteAttachmentsForCipherAsync(cipher.Id);
```
JaredScar, line 457: By doing what I suggested above, we can probably also get rid of another for loop here as well

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -494,7 +498,17 @@ public async Task PurgeAsync(Guid organizationId)
...
         }
+
+        // Fetch cipher IDs before DB deletion so we can clean up attachment storage
+        var orgCiphers = await _cipherRepository.GetManyByOrganizationIdAsync(organizationId);
+
         await _cipherRepository.DeleteByOrganizationIdAsync(organizationId);
+
+        foreach (var cipher in orgCiphers)
+        {
+            await _attachmentStorageService.DeleteAttachmentsForCipherAsync(cipher.Id);
```
JaredScar: By having 1 spot where we only supply the organization ID to pass, we can get rid of all these for loops! 😄

JaredScar (changes requested):
Good stuff! Just a few small cleanups and should be good to go

jaasen-livefront:
@gbubemismith @JaredScar I've addressed all the feedback in https://github.com/bitwarden/server/pull/7208/commits/834a1b92a4b327202cd94c5d628a34522acd5ec7. Thanks!

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -494,10 +500,26 @@ public async Task PurgeAsync(Guid organizationId)
...
+
         await _cipherRepository.DeleteByOrganizationIdAsync(organizationId);
+
         await _eventService.LogOrganizationEventAsync(org, EventType.Organization_PurgedVault);
     }
 
+    public async Task DeleteAttachmentsForOrganizationAsync(Guid organizationId)
+    {
+        var ciphersWithAttachments = (await _cipherRepository.GetManyByOrganizationIdAsync(organizationId))
+            .Where(c => c.GetAttachments()?.Count > 0);
```
gbubemismith: ❓ @jaasen-livefront, is there a reason we will need the whole cipher object? since we just need the cipher Ids, we can add a `.Select(c => c.Id)` to this

jaasen-livefront: @gbubemismith Good point! Fixed https://github.com/bitwarden/server/pull/7208/commits/2cac61e19494e97cfc50825079964926b965df55

gbubemismith (dismissed):

jaasen-livefront:
@JaredScar Feedback has been addressed and ready for another look when you have a moment. Ty!

JaredScar (dismissed):

gbubemismith (dismissed):

JaredScar (approved):

shane-melton (approved):

PR #7212, opened 2026-03-13
[PM-3836] Tools - Make Controllers, Services and API Models nullable

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-3836

## 📔 Objective

Update Tools owned code to make Controllers, Services and API Models nullable

See related PR https://github.com/bitwarden/server/pull/7266


src/Api/Tools/Models/Request/Accounts/ImportCiphersRequestModel.cs
```diff
@@ -1,13 +1,10 @@
-﻿// FIXME: Update this file to be null safe and then delete the line below
-#nullable disable
-
-using Bit.Api.Vault.Models.Request;
+﻿using Bit.Api.Vault.Models.Request;
 
 namespace Bit.Api.Tools.Models.Request.Accounts;
 
 public class ImportCiphersRequestModel
 {
```
harr1424, line 6: See [corresponding model in clients](https://github.com/bitwarden/clients/blob/c6bc97cf5c26bd025b107b10015ba3e289275275/libs/common/src/models/request/import-ciphers.request.ts)

src/Api/Tools/Models/Request/Organizations/ImportOrganizationCiphersRequestModel.cs
```diff
@@ -1,14 +1,11 @@
...
-#nullable disable
-
-using Bit.Api.Models.Request;
+﻿using Bit.Api.Models.Request;
 using Bit.Api.Vault.Models.Request;
 
 namespace Bit.Api.Tools.Models.Request.Organizations;
 
 public class ImportOrganizationCiphersRequestModel
 {
```
harr1424, line 7: See [corresponding model in clients](https://github.com/bitwarden/clients/blob/56a3b14583747a499d66960ac5b372ee0857246e/libs/common/src/models/request/import-organization-ciphers.request.ts)

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -146,12 +138,8 @@ public Send ToSend(Guid userId, ISendAuthorizationService sendAuthorizationServi
...
         // FIXME: This method does two things: creates a send and a send file data.
         //        It should only do one thing.
-        var send = ToSendBase(new Send
-        {
-            Type = Type,
-            UserId = (Guid?)userId
-        }, sendAuthorizationService);
-        var data = new SendFileData(Name, Notes, fileName);
+        var send = ToSendBase(new Send { Type = Type, UserId = (Guid?)userId }, sendAuthorizationService);
+        var data = new SendFileData(Name ?? string.Empty, Notes, fileName);
```
harr1424, line 142: Only the CLI can create or edit Sends having no name, the former behavior serialized and deserialized them as 'null', I've taken this opportunity to instead assign an empty string. Suggestions for a different default are welcome.

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -268,7 +264,7 @@ private Send ToSendBase(Send existingSend, ISendAuthorizationService authorizati
                     existingSend.Password = null;
                     break;
                 case Core.Tools.Enums.AuthType.Password:
-                    existingSend.Password = authorizationService.HashPassword(Password);
+                    existingSend.Password = authorizationService.HashPassword(Password!);
```
harr1424, line 267: Clients all enforce that a password protected Send must have a password

src/Api/Tools/Models/Request/SendRequestModel.cs
```diff
@@ -245,7 +241,7 @@ private Send ToSendBase(Send existingSend, ISendAuthorizationService authorizati
     {
         existingSend.Key = Key;
         existingSend.ExpirationDate = ExpirationDate;
-        existingSend.DeletionDate = DeletionDate.Value;
+        existingSend.DeletionDate = DeletionDate!.Value;
```
harr1424, line 244: This is safe since `DeletionDate` is required during model binding.

src/Api/Tools/Models/Response/SendResponseModel.cs
```diff
@@ -39,7 +36,7 @@ public SendResponseModel(Send send)
         AccessId = CoreHelpers.Base64UrlEncode(send.Id.ToByteArray());
         Type = send.Type;
         AuthType = send.AuthType ?? SendUtilities.InferAuthType(send);
-        Key = send.Key;
+        Key = send.Key!;
```
harr1424, line 39: See Send propoerty definition for Key, it is guarenteed to be non-null

src/Api/Tools/Models/Request/SendRequestModel.cs
mcamirault: I think empty string is fine, definitely an improvement over the former behavior. It does seem weird that the CLI is allowed to do that though, was that a deliberate design choice?

src/Core/Tools/SendFeatures/Services/AzureSendFileStorageService.cs
```diff
@@ -11,39 +8,40 @@
...
         var metadata = new Dictionary<string, string>();
         if (send.UserId.HasValue)
         {
             metadata.Add("userId", send.UserId.Value.ToString());
         }
-        else
+        else if (send.OrganizationId.HasValue)
         {
             metadata.Add("organizationId", send.OrganizationId.Value.ToString());
         }
```
mcamirault, line 47: ❓ Should we add logic enforcing that the Send has to have one of `UserId` or `OrganizationId` in its metadata?

src/Core/Tools/SendFeatures/Services/AzureSendFileStorageService.cs
```diff
@@ -106,7 +104,7 @@ public async Task<string> GetSendFileUploadUrlAsync(Send send, string fileId)
             {
                 metadata["userId"] = send.UserId.Value.ToString();
             }
-            else
+            else if (send.OrganizationId.HasValue)
```
mcamirault, line 107: ❓ Same as above

mcamirault (commented):
One question about a possible change and one question for my own curiosity

Looks great overall!

src/Api/Tools/Models/Request/SendRequestModel.cs
harr1424: I am not sure about this, I'd like to raise this with Product next time we sync. Typically, we strive for parity among clients.  I didn't want to increase the scope of this work by making any changes in `clients` without raising this to product first; I think the CLI appeals to power-admins and wasn't sure if this would lead to any breaking changes in scrips used for batch generating Sends.

src/Core/Tools/SendFeatures/Services/AzureSendFileStorageService.cs
harr1424: I don't actually think we would gain much from this, but it also wouldn't hurt to be more defensive. I am not sure if this PR is the most appropriate place to make this change, mainly because I've tried to introduce behavioral changes only as strictly necessary to satisfy the compiler which is now enforcing nullability. 

The throws introduced in this PR are not so much intended to tighten enforcement but instead replace the existing potential `NullReferenceExceptions` with exceptions wrapping an error message, they could have almost all been refactored from no-null-handling to using the [null-forgiving operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving) and this would have more closely matched the behavior when `#nullable disable` was present in these files, but I think the throws both: 
- indirectly document the expected variable state 
- surface the same exception types, but with additional context  

Right now, [OrganizationId](https://github.com/bitwarden/server/blob/6d43cc43e36087cf2a6fcfae8ab325f47667e23e/src/Core/Tools/Entities/Send.cs#L30) isn't actually being used in `clients`, and `UserId` is guaranteed to be assigned before the logic in this code is reached: 

https://github.com/bitwarden/server/blob/f80a89f8cf2467eedc41bef41236d9faf6991d9b/src/Api/Tools/Models/Request/SendRequestModel.cs#L125

https://github.com/bitwarden/server/blob/f80a89f8cf2467eedc41bef41236d9faf6991d9b/src/Api/Tools/Models/Request/SendRequestModel.cs#L141

This assumes that `_userService.GetProperUserId(User)` does return a valid UserId, and if it does not, that case is already [handled](https://github.com/bitwarden/server/blob/a783f609e35af0f01cca0ed60a5b05b14bb91769/src/Api/Tools/Controllers/SendsController.cs#L313).

mcamirault: Sounds good, you're probably right that even if we did make my change we shouldn't do so in this PR

mcamirault (approved):

PR #7216, opened 2026-03-13
[PM-33435] Add new user key rotation endpoint with MP support

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33435

## 📔 Objective

Adds a new endpoint for user key rotation without a password change. This implements the rotation for master password users and setups things for future implementation of key connector and TDE users.

This builds off the refactor done in https://github.com/bitwarden/server/pull/7201.


src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -240,4 +264,36 @@ public async Task<KeyConnectorConfirmationDetailsResponseModel> GetKeyConnectorC
...
+                    request.UnlockData.OrganizationAccountRecoveryUnlockData),
+            WebAuthnKeys = await _webauthnKeyValidator.ValidateAsync(user, request.UnlockData.PasskeyUnlockData),
+            DeviceKeys = await _deviceValidator.ValidateAsync(user, request.UnlockData.DeviceKeyUnlockData),
+            V2UpgradeToken = request.UnlockData.V2UpgradeToken?.ToData(),
+
+            Ciphers = await _cipherValidator.ValidateAsync(user, request.AccountData.Ciphers),
+            Folders = await _folderValidator.ValidateAsync(user, request.AccountData.Folders),
+            Sends = await _sendValidator.ValidateAsync(user, request.AccountData.Sends),
+        };
+    }
```
mzieniukbw: We usually hide it inside of the request itself, i.e. `ToData()`

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -240,4 +264,36 @@ public async Task<KeyConnectorConfirmationDetailsResponseModel> GetKeyConnectorC
         var details = await _keyConnectorConfirmationDetailsQuery.Run(orgSsoIdentifier, user.Id);
         return new KeyConnectorConfirmationDetailsResponseModel(details);
     }
+
+    private async Task MasterPasswordRotateUserAccountKeysAsync(RotateUserKeysRequestModel request, User user)
+    {
+        ArgumentNullException.ThrowIfNull(request.UnlockMethodData.MasterPasswordUnlockData);
```
mzieniukbw: Shouldn't this be done within the validation of the request ?

src/Api/KeyManagement/Models/Requests/UnlockMethodRequestModel.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        if (KeyConnectorKeyWrappedUserKey != null && MasterPasswordUnlockData != null)
+        {
+            yield return new ValidationResult("Invalid request user can't have KeyConnectorKeyWrappedUserKey and MasterPasswordUnlockData");
+        }
+    }
+
+    public UnlockMethod GetUnlockMethod()
```
mzieniukbw: 🤔 I am wondering, whether it might be better to have a property `public required UnlockMethod UnlockMethod {get; init; }`, instead of assuming the decryption type based on other properties existence.

src/Api/KeyManagement/Models/Requests/WrappedAccountCryptographicStateRequestModel.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Bit.Core.KeyManagement.Models.Api.Request;
+using Bit.Core.KeyManagement.Models.Data;
+
+namespace Bit.Api.KeyManagement.Models.Requests;
+
+public class WrappedAccountCryptographicStateRequestModel
```
mzieniukbw, line 7: ❓ We already have `AccountKeysRequestModel`. Might be better to combine this with https://bitwarden.atlassian.net/browse/PM-22384 ?
If this is outside of scope, could we comment on the `AccountKeysRequestModel` that it's deprecated and will be superseded by `WrappedAccountCryptographicStateRequestModel` ?

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
Thomas-Avery: We do normally, but these called the validators which can't be injected into a request class to my knowledge.

See the pre-existing endpoint.

https://github.com/bitwarden/server/blob/a7259df98315ab80653deff2dc7a314a0d599b13/src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs#L113-L129

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
Thomas-Avery: It is, this is to make the complier happy as it doesn't keep track of the switch in `request.UnlockMethodData.GetUnlockMethod()`. (Where the validation happens).

src/Api/KeyManagement/Models/Requests/UnlockMethodRequestModel.cs
Thomas-Avery: Great suggestion, I'll work on converting that over. I think this will solve the  `ArgumentNullException.ThrowIfNull` too.

src/Api/KeyManagement/Models/Requests/WrappedAccountCryptographicStateRequestModel.cs
Thomas-Avery: The problem is `AccountsKeyRequestModel` has been augmented to handle both v1 and v2 encryption payloads (plus has some leftover properties) . @quexten had requested to make a new model that only accepts v2 encryption payloads.

I don't mind adding a comment on `AccountKeysRequestModel` but we will have to make it clear we can only fully move over to `WrappedAccountCryptographicStateRequestModel` after we stop v1 encryption rotation support.

I'm not following on the ticket linked (PM-22384). That one is already done and where we added v2 encryption support to `AccountsKeyRequestModel`?

src/Api/KeyManagement/Models/Requests/UnlockMethodRequestModel.cs
Thomas-Avery: Done in https://github.com/bitwarden/server/pull/7216/commits/dab563e0688fae13065bcb777f1ef36066ad8d8e

src/Api/KeyManagement/Models/Requests/WrappedAccountCryptographicStateRequestModel.cs
Thomas-Avery: I added a note on the new request model https://github.com/bitwarden/server/pull/7216/changes/bc6bfe5a9720847e98711c4f7f60bbd5672d5015. 

I also created a ticket for tracking the tech debt for user key rotation https://bitwarden.atlassian.net/browse/PM-33860. I don't think now is the correct time to mark `AccountKeysRequestModel` as obsolete. It is used in many places that will still need to support v1 encryption payloads.

mzieniukbw: I provided wrong link, i meant https://bitwarden.atlassian.net/browse/PM-23751

Thomas-Avery: Lets leave that out of scope for this PR. Since this is a new endpoint I was planning on doing the QA after merging and introducing usage on the client. For removing those properties I would want to do a regression test.

mzieniukbw (approved):

PR #7227, opened 2026-03-16
[PM-32480] Add endpoint for Stripe billing portal session

cyprain-okeke:
## 🎟️ Tracking

<!-- Paste the link to the Jira or GitHub issue or otherwise describe / point to where this change is coming from. -->
https://bitwarden.atlassian.net/browse/PM-32480

## 📔 Objective

<!-- Describe what the purpose of this PR is, for example what bug you're fixing or new feature you're adding. -->
- Adds new `POST /account/billing/vnext/portal-session` endpoint that returns a Stripe Customer Portal URL                      
- Enables Premium users with active or past_due subscriptions to manage or cancel their subscriptions via Stripe's hosted       portal                                                                                                                          
- Implements security validations to ensure only eligible users can access the portal

## Changes                                                                                                                      
                                                                                                                                  
### API Layer                                                                                                                   
 - Added `CreatePortalSessionAsync` endpoint to `AccountBillingVNextController`                                                  
 - Created `PortalSessionRequest` model with URL validation and XSS prevention                                                   
 - Created `PortalSessionResponse` model for returning the portal URL                                                            
                                                                                                                                 
### Core Layer                                                                                                                  
 - Implemented `ICreateBillingPortalSessionCommand` with comprehensive business logic validation:                                
  - Verifies user has Stripe customer ID                                                                                        
  - Verifies user has active subscription                                                                                       
  - Validates subscription status (only `active` or `past_due` allowed)                                                         
  - Creates Stripe billing portal session with custom return URL                                                                
  - Added `CreateBillingPortalSessionAsync` to `IStripeAdapter` and `StripeAdapter`                                               
  - Registered command in dependency injection container 

## Test Plan                                                                                                                    
                                                                                                                                 
    1. **Success Case**: Call endpoint with authenticated Premium user (active subscription)                                        
       - Verify portal URL is returned                                                                                              
       - Verify URL can be opened and displays subscription management options                                                      
                                                                                                                                  
    2. **Past Due Case**: Test with Premium user in past_due status                                                                 
        - Verify portal URL is returned                                                                                                                                                                                                                                │
    3. **Error Cases**: Test with users who:                                                                                        
       - Have no Stripe customer ID → Returns 400 BadRequest                                                                        
       - Have no subscription → Returns 400 BadRequest                                                                              
       - Have canceled/incomplete subscription → Returns 400 BadRequest                                                             
       - Subscription not found in Stripe → Returns 400 BadRequest                                                                  
                                                                                                                                  
    4. **Validation**: Test request validation                                                                                      
       - Missing return URL → Returns 400                                                                                           
       - Invalid URL format → Returns 400                                                                                           
       - Non-HTTP(S) scheme (javascript:, data:, etc.) → Returns 400             

## 📸 Screenshots

<!-- Required for any UI changes; delete if not applicable. Use fixed width images for better display. -->

### DIFF
--- a/global.json
+++ b/global.json
@@ -6,6 +6,6 @@
   "msbuild-sdks": {
     "Microsoft.Build.Traversal": "4.1.0",
     "Microsoft.Build.Sql": "1.0.0",
-    "Bitwarden.Server.Sdk": "1.4.0"
+    "Bitwarden.Server.Sdk": "1.5.1"
   }
 }
--- a/src/Api/AdminConsole/Models/Request/Organizations/OrganizationCreateRequestModel.cs
+++ b/src/Api/AdminConsole/Models/Request/Organizations/OrganizationCreateRequestModel.cs
@@ -81,6 +81,8 @@ public class OrganizationCreateRequestModel : IValidatableObject
 
     public bool SkipTrial { get; set; }
 
+    public string[] Coupons { get; set; }
+
     public virtual OrganizationSignup ToOrganizationSignup(User user)
     {
         var orgSignup = new OrganizationSignup
@@ -114,6 +116,7 @@ public class OrganizationCreateRequestModel : IValidatableObject
             },
             InitiationPath = InitiationPath,
             SkipTrial = SkipTrial,
+            Coupons = Coupons,
             Keys = Keys?.ToPublicKeyEncryptionKeyPairData()
         };
 
--- a/src/Api/Billing/Models/Requests/Organizations/OrganizationSubscriptionPurchaseRequest.cs
+++ b/src/Api/Billing/Models/Requests/Organizations/OrganizationSubscriptionPurchaseRequest.cs
@@ -20,8 +20,7 @@ public record OrganizationSubscriptionPurchaseRequest : IValidatableObject
 
     public SecretsManagerPurchaseSelections? SecretsManager { get; set; }
 
-    [MaxLength(50)]
-    public string? Coupon { get; set; }
+    public string[]? Coupons { get; set; }
 
     public OrganizationSubscriptionPurchase ToDomain() => new()
     {
@@ -39,7 +38,7 @@ public record OrganizationSubscriptionPurchaseRequest : IValidatableObject
             AdditionalServiceAccounts = SecretsManager.AdditionalServiceAccounts,
             Standalone = SecretsManager.Standalone
         } : null,
-        Coupon = Coupon
+        Coupons = Coupons
     };
 
     public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
--- a/src/Api/Billing/Models/Requests/Premium/PremiumCloudHostedSubscriptionRequest.cs
+++ b/src/Api/Billing/Models/Requests/Premium/PremiumCloudHostedSubscriptionRequest.cs
@@ -16,8 +16,7 @@ public class PremiumCloudHostedSubscriptionRequest : IValidatableObject
     [Range(0, 99)]
     public short AdditionalStorageGb { get; set; } = 0;
 
-    [MaxLength(50)]
-    public string? Coupon { get; set; }
+    public string[]? Coupons { get; set; }
 
     public PremiumSubscriptionPurchase ToDomain()
     {
@@ -36,7 +35,7 @@ public class PremiumCloudHostedSubscriptionRequest : IValidatableObject
             PaymentMethod = paymentMethod,
             BillingAddress = billingAddress,
             AdditionalStorageGb = AdditionalStorageGb,
-            Coupon = Coupon
+            Coupons = Coupons
         };
     }
 
--- a/src/Api/Billing/Models/Requests/PreviewInvoice/PreviewPremiumSubscriptionPurchaseTaxRequest.cs
+++ b/src/Api/Billing/Models/Requests/PreviewInvoice/PreviewPremiumSubscriptionPurchaseTaxRequest.cs
@@ -14,14 +14,13 @@ public record PreviewPremiumSubscriptionPurchaseTaxRequest
     [Required]
     public required MinimalBillingAddressRequest BillingAddress { get; set; }
 
-    [MaxLength(50)]
-    public string? Coupon { get; set; }
+    public string[]? Coupons { get; set; }
 
     public (PremiumPurchasePreview, BillingAddress) ToDomain() => (
         new PremiumPurchasePreview
         {
             AdditionalStorageGb = AdditionalStorage,
-            Coupon = Coupon
+            Coupons = Coupons
         },
         BillingAddress.ToDomain());
 }
--- a/src/Core/Billing/Models/Sales/CustomerSetup.cs
+++ b/src/Core/Billing/Models/Sales/CustomerSetup.cs
@@ -8,7 +8,7 @@ public class CustomerSetup
 {
     public TokenizedPaymentSource? TokenizedPaymentSource { get; set; }
     public TaxInformation? TaxInformation { get; set; }
-    public string? Coupon { get; set; }
+    public string[]? Coupons { get; set; }
 
     public bool IsBillable => TokenizedPaymentSource != null && TaxInformation != null;
 }
--- a/src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
+++ b/src/Core/Billing/Organizations/Commands/PreviewOrganizationTaxCommand.cs
@@ -126,18 +126,26 @@ public class PreviewOrganizationTaxCommand(
                         }
                     }
 
-                    // Validate coupon and only apply if valid. If invalid, proceed without the discount.
-                    // Only Families plans support user-provided coupons
-                    if (!string.IsNullOrWhiteSpace(purchase.Coupon) && purchase.Tier == ProductTierType.Families)
+                    // Validate all coupons at once. If all are eligible, apply them; otherwise skip gracefully.
+                    // Only Families plans support user-provided coupons.
+                    if (purchase is { Coupons.Length: > 0, Tier: ProductTierType.Families })
                     {
-                        var isValid = await subscriptionDiscountService.ValidateDiscountEligibilityForUserAsync(
-                            user,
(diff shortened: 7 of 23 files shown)


src/Api/Billing/Models/Requests/Portal/PortalSessionRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+/// </summary>
+public class PortalSessionRequest : IValidatableObject
+{
+    /// <summary>
+    /// The URL to redirect to after the user completes their session in the billing portal.
+    /// Must be a valid HTTP(S) URL.
+    /// </summary>
+    [Required]
+    [MaxLength(2000)]
+    public string? ReturnUrl { get; set; }
```
SaintPatrck: ❓ I don't see this mentioned as a request property in the linked JIRA ticket, or in the tech breakdown. When the mobile client calls this endpoint, what value should this be and where does it come from?

cyprain-okeke: The ReturnUrl is required by Stripe's billing portal session API (see
  https://docs.stripe.com/api/customer_portal/sessions/create). When users finish managing their subscription in the Stripe
  portal, they click a "Return to Bitwarden" link that redirects to this URL.

  For mobile clients, this should be a deep link (e.g., bitwarden://billing-return or similar app-specific URL scheme) that opens
   the app. The mobile client would:
  1. Open the portal URL in an in-app browser/web view
  2. After the user completes their session, Stripe redirects to the ReturnUrl
  3. The deep link triggers the mobile app to close the web view and return to the billing screen

  This is standard practice for Stripe portal integration on mobile - the same pattern is used across other platforms (see
  https://docs.stripe.com/customer-management/integrate-customer-portal).

src/Api/Billing/Models/Requests/Portal/PortalSessionRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (!Uri.TryCreate(ReturnUrl, UriKind.Absolute, out var uri))
+            {
+                yield return new ValidationResult(
+                    "Return URL must be a valid absolute URL.",
+                    [nameof(ReturnUrl)]);
+                yield break;
+            }
+
+            // Prevent open redirect vulnerabilities by restricting to HTTP(S) schemes
+            if (uri.Scheme != Uri.UriSchemeHttp && uri.Scheme != Uri.UriSchemeHttps)
```
SaintPatrck: If `ReturnUrl` is a deeplink like `bitwarden://foo`, this validation will trigger an error response. Is that intentional?

cyprain-okeke: remove the returnurl from the request, instead pass it from the server base on client

src/Api/Billing/Controllers/VNext/AccountBillingVNextController.cs
```diff
@@ -147,4 +155,24 @@ public async Task<IResult> GetApplicableDiscountsAsync(
         return Handle(result);
     }
 
+    /// <summary>
+    /// Creates a Stripe billing portal session for the authenticated user.
+    /// The portal allows users to manage their subscription, payment methods, and billing history.
+    /// The return URL is automatically determined based on the client type (mobile, web, desktop, etc.).
+    /// </summary>
+    /// <param name="user">The authenticated user</param>
+    /// <returns>Portal session URL for redirection</returns>
```
amorask-bitwarden: ⛏️ I don't think we need code comments on controller methods.

src/Api/Billing/Controllers/VNext/AccountBillingVNextController.cs
```diff
@@ -147,4 +155,24 @@ public async Task<IResult> GetApplicableDiscountsAsync(
...
+    [HttpPost("portal-session")]
+    [InjectUser]
+    public async Task<IResult> CreatePortalSessionAsync([BindNever] User user)
+    {
+        // Mobile clients use deep link callbacks, all others redirect to web vault
+        var returnUrl = DeviceTypes.ToClientType(currentContext.DeviceType) == ClientType.Mobile
+            ? "bitwarden://premium-upgrade-callback"
+            : $"{globalSettings.BaseServiceUri.Vault}/#/settings/subscription/premium";
```
amorask-bitwarden: ❓ Is there a use-case for supporting this endpoint from a non-Mobile device? In the Web Vault, subscriptions are managed there, so until we know we need this, I'd prefer we limit our risk and maybe return a Not Found if the user's not on a mobile device.

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+        new("Unable to create billing portal session. Please contact support for assistance.");
+
+    public Task<BillingCommandResult<string>> Run(User user, string returnUrl) =>
+        HandleAsync<string>(async () =>
+        {
+            if (string.IsNullOrEmpty(user.GatewayCustomerId))
+            {
+                _logger.LogWarning("{Command}: User ({UserId}) does not have a Stripe customer ID",
+                    CommandName, user.Id);
+                return new BadRequest("User does not have a Stripe customer ID.");
```
amorask-bitwarden: ❓ @SaintPatrck If this message gets returned in a 400 response, would the Android app show it to the user in some sort of toast or error message? If so, this should be the default conflict. The user has no control over whether they have a `GatewayCustomerId`. If they're able to reach this point without one, that's a system issue, not a user issue.

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            if (string.IsNullOrEmpty(user.GatewaySubscriptionId))
+            {
+                _logger.LogWarning("{Command}: User ({UserId}) does not have a subscription",
+                    CommandName, user.Id);
+                return new BadRequest("User does not have a Premium subscription.");
```
amorask-bitwarden: ❓ Same question as above.

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+            try
+            {
+                subscription = await stripeAdapter.GetSubscriptionAsync(user.GatewaySubscriptionId);
+            }
+            catch (StripeException stripeException)
+            {
+                _logger.LogError(stripeException,
+                    "{Command}: Failed to fetch subscription ({SubscriptionId}) for user ({UserId})",
+                    CommandName, user.GatewaySubscriptionId, user.Id);
+                return new BadRequest("Unable to verify subscription status.");
```
amorask-bitwarden: ❌ This handling means that any `StripeException` would result in us returning a 400 which, again, the invoker of this command doesn't control. Seems like this should also be the default conflict.

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            if (subscription == null)
```
amorask-bitwarden: ❌ I've mentioned this a few times on other PRs: Subscriptions will never come back null from the `Stripe.net` package; you'll receive a `StripeException` with the error code `resource_missing`. This code path will never be hit.

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            var options = new SessionCreateOptions
+            {
+                Customer = user.GatewayCustomerId,
+                ReturnUrl = returnUrl
+            };
+
+            var session = await stripeAdapter.CreateBillingPortalSessionAsync(options);
+
+            if (session?.Url == null)
```
amorask-bitwarden: ❌ This is not nullable in the [Stripe API](https://docs.stripe.com/api/customer_portal/sessions/object?api-version=2025-08-27.basil&lang=dotnet#portal_session_object-url), this codepath will never be hit.

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                return DefaultConflict;
+            }
+
+            _logger.LogInformation("{Command}: Successfully created billing portal session for user ({UserId})",
```
amorask-bitwarden: ❓ Is this for testing? We should be able to fairly easily see the portal being opened upon a link click if so.

test/Api.Test/Billing/Controllers/VNext/AccountBillingVNextControllerTests.cs
```diff
@@ -33,19 +42,29 @@ public AccountBillingVNextControllerTests()
...
             Substitute.For<Core.Billing.Payment.Queries.IGetCreditQuery>(),
             Substitute.For<Core.Billing.Payment.Queries.IGetPaymentMethodQuery>(),
             _getUserLicenseQuery,
+            _globalSettings,
             Substitute.For<IReinstateSubscriptionCommand>(),
             Substitute.For<Core.Billing.Payment.Commands.IUpdatePaymentMethodCommand>(),
             _updatePremiumStorageCommand,
-            _upgradePremiumToOrganizationCommand,
-            _getApplicableDiscountsQuery);
+            _upgradePremiumToOrganizationCommand);
```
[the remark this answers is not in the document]
cyprain-okeke: resolved

src/Core/Billing/Portal/Commands/CreateBillingPortalSessionCommand.cs
SaintPatrck: @amorask-bitwarden It varies by service and endpoint. We can ignore it if we know it's not user friendly. We generally prefer to use our own message so that we can localize it.

PR #7230, opened 2026-03-16
[PM-21926] [PM-30350] [PM-32389] Read salt from database

ike-kottlowski:
## 🎟️ Tracking

[PM-21926](https://bitwarden.atlassian.net/browse/PM-21926)
[PM-30350](https://bitwarden.atlassian.net/browse/PM-30350)
[PM-32389](https://bitwarden.atlassian.net/browse/PM-32389)

## 📔 Objective

Add MasterPasswordSalt to the response for ReadKdfByEmail and OranizationAccountRevovery in the User Repository.

I've removed a Test pertaining to Emergency Access that tries to check the salt against the `User.Email` which may not be the case in the future. So, it has been removed.

### Notes

The AC Repository updates for account recovery are handled in this PR which is in draft until this one is merged. 

AC PR: https://github.com/bitwarden/server/pull/7231

[PM-21926]: https://bitwarden.atlassian.net/browse/PM-21926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-30350]: https://bitwarden.atlassian.net/browse/PM-30350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-32389]: https://bitwarden.atlassian.net/browse/PM-32389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7231, opened 2026-03-16
[PM-21926] Add salt to Admin Console DTOs

ike-kottlowski:
## 🎟️ Tracking

[PM-21926](https://bitwarden.atlassian.net/browse/PM-21926)

This PR depends on:
https://github.com/bitwarden/server/pull/7230

## 📔 Objective

Add MasterPasswordSalt to DTOs that are owned by Admin Console.

[PM-21926]: https://bitwarden.atlassian.net/browse/PM-21926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/AdminConsole/Models/Response/Organizations/OrganizationUserResponseModel.cs
```diff
@@ -214,6 +215,7 @@ public OrganizationUserResetPasswordDetailsResponseModel(OrganizationUserResetPa
     public int KdfIterations { get; set; }
     public int? KdfMemory { get; set; }
     public int? KdfParallelism { get; set; }
+    public string MasterPasswordSalt { get; set; }
```
jrmccannon, line 218: Is this field nullable? If so, could we do `#nullable enabled` for the blocks required and change the type to `string?` ?

src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserResetPasswordDetails.cs
```diff
@@ -41,6 +42,7 @@ public OrganizationUserResetPasswordDetails(OrganizationUser orgUser, User user,
     public int KdfIterations { get; set; }
     public int? KdfMemory { get; set; }
     public int? KdfParallelism { get; set; }
+    public string MasterPasswordSalt { get; set; }
```
jrmccannon, line 45: Same question here as in `OrganizationUserResetPasswordDetailsResponseModel`

src/Api/AdminConsole/Models/Response/Organizations/OrganizationUserResponseModel.cs
ike-kottlowski: This class has the explicit `#nullable disabled` and modifying this class to be `#nullable` felt outside of the scope of this effort so I opted to leave it `string`.

Are you asking that I enable `#nullable` for `MasterPasswordSalt` only so that we can leverage `string?`?

jrmccannon: That was the thought, but now that I read it back, that's silly. This is fine. Apologies.

PR #7238, opened 2026-03-17
Sm prototype for manage permissions

maxkpower:
Prototype for adding a Manage permission to Secrets Manager access policies, enabling a dedicated tier for administering who can access projects and secrets separate from Read/Write data access. 

This PR should not be merged.

Related PR with more context: https://github.com/bitwarden/clients/pull/19613


bitwarden_license/src/Commercial.Core/SecretsManager/AuthorizationHandlers/AccessPolicies/ProjectServiceAccountsAccessPoliciesAuthorizationHandler.cs
```diff
@@ -84,7 +102,7 @@ await _serviceAccountRepository.ServiceAccountsAreInOrganizationAsync(serviceAcc
         }
 
         // Users can only create access policies for service accounts they have access to.
-        // User can delete and update any service account access policy if they have write access to the project.
+        // Users can delete and update any service account access policy if they have Manage access to the project.
```
cd-bitwarden, line 105: Should we also allow the user to delete and update if they have write access, or are we now restricting this to Manage?

PR #7241, opened 2026-03-17
Arch/cipher scene

MGibson1:
## 📔 Objective

Provide seeder api scenes for cipher creation. To reduce API complexity, I elected to create multiple scenes rather than a single that keys off of a type.

### DIFF
--- a/util/Seeder/Factories/LoginCipherSeeder.cs
+++ b/util/Seeder/Factories/LoginCipherSeeder.cs
@@ -17,7 +17,6 @@ internal static class LoginCipherSeeder
         string? notes = null,
         bool reprompt = false,
         bool deleted = false,
-        bool favorite = false,
         IEnumerable<(string name, string value, int type)>? fields = null)
     {
         var cipherView = new CipherViewDto
@@ -34,7 +33,6 @@ internal static class LoginCipherSeeder
             },
             Reprompt = reprompt ? RepromptTypes.Password : RepromptTypes.None,
             DeletedDate = deleted ? DateTime.UtcNow.AddDays(-1) : null,
-            Favorite = favorite,
             Fields = fields?.Select(f => new FieldViewDto
             {
                 Name = f.name,
--- a/util/Seeder/Scenes/UserCardCipherScene.cs
+++ b/util/Seeder/Scenes/UserCardCipherScene.cs
@@ -27,7 +27,7 @@ public class UserCardCipherScene(IUserRepository userRepository, ICipherReposito
 
     public class Result
     {
-        public required Guid CipherId { get; set; }
+        public required Guid CipherId { get; init; }
     }
 
     public async Task<SceneResult<Result>> SeedAsync(Request request)
--- a/util/Seeder/Scenes/UserIdentityCipherScene.cs
+++ b/util/Seeder/Scenes/UserIdentityCipherScene.cs
@@ -26,7 +26,7 @@ public class UserIdentityCipherScene(IUserRepository userRepository, ICipherRepo
 
     public class Result
     {
-        public required Guid CipherId { get; set; }
+        public required Guid CipherId { get; init; }
     }
 
     public async Task<SceneResult<Result>> SeedAsync(Request request)
--- a/util/Seeder/Scenes/UserLoginCipherScene.cs
+++ b/util/Seeder/Scenes/UserLoginCipherScene.cs
@@ -29,7 +29,7 @@ public class UserLoginCipherScene(IUserRepository userRepository, ICipherReposit
 
     public class Result
     {
-        public required Guid CipherId { get; set; }
+        public required Guid CipherId { get; init; }
     }
 
     public async Task<SceneResult<Result>> SeedAsync(Request request)
@@ -40,7 +40,7 @@ public class UserLoginCipherScene(IUserRepository userRepository, ICipherReposit
             throw new Exception($"User with ID {request.UserId} not found.");
         }
 
-        var cipher = LoginCipherSeeder.Create(request.UserKeyB64, request.Name, userId: request.UserId, username: request.Username, password: request.Password, uri: request.Uri, notes: request.Notes, fields: request.Fields, reprompt: request.Reprompt, deleted: request.Deleted, favorite: request.Favorite);
+        var cipher = LoginCipherSeeder.Create(request.UserKeyB64, request.Name, userId: request.UserId, username: request.Username, password: request.Password, uri: request.Uri, notes: request.Notes, fields: request.Fields, reprompt: request.Reprompt, deleted: request.Deleted);
         if (request.Favorite)
         {
             cipher.Favorites = JsonSerializer.Serialize(new Dictionary<string, bool>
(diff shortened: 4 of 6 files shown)


util/Seeder/Scenes/UserSshKeyCipherScene.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Required]
+        public required Guid UserId { get; set; }
+        [Required]
+        public required string UserKeyB64 { get; set; }
+        [Required]
+        public required string Name { get; set; }
+        public string? PrivateKey { get; set; }
+        public string? PublicKey { get; set; }
+        public string? Fingerprint { get; set; }
+    }
```
theMickster, line 25: ⚠️ We are missing the `Notes` property — all other cipher scenes expose it. Let's ensure we expose it for consistency

```csharp
// In Request class:
public string? Notes { get; set; }

// In SeedAsync, update the Create call:
var cipher = SshKeyCipherSeeder.Create(request.UserKeyB64, request.Name, userId: request.UserId, sshKey: sshKey, notes: request.Notes);
```

util/SeederApi/Commands/DestroyBatchScenesCommand.cs
```diff
@@ -10,7 +10,7 @@ public async Task DestroyAsync(IEnumerable<string> playIds)
     {
         var exceptions = new List<Exception>();
 
-        var deleteTasks = playIds.Select(async playId =>
+        foreach (var playId in playIds)
```
theMickster, line 13: ♻️  The `lock (exceptions)` on line 21 is now unnecessary I think we should remove it.

Now that this is a sequential `foreach` with `await`, there is no concurrent access to the list. The lock is harmless but misleading — it implies thread contention that cannot occur.

```csharp
catch (Exception ex)
{
    exceptions.Add(ex);
    logger.LogError(ex, "Error deleting seeded data: {PlayId}", playId);
}
```

util/Seeder/Scenes/UserSshKeyCipherScene.cs
MGibson1: To that end, was there a reason you didn't expose `Fields` on every cipher type? All ciphers can have them, but they weren't always available in the constructor

theMickster: I hadn't dug enough into the rules of the fields yet to really get a grasp. 
That's something to implement next, though

util/Seeder/Factories/LoginCipherSeeder.cs
```diff
@@ -29,6 +31,8 @@ internal static Cipher Create(
                 Password = password,
                 Uris = string.IsNullOrEmpty(uri) ? null : [new LoginUriViewDto { Uri = uri }]
             },
+            Reprompt = reprompt ? RepromptTypes.Password : RepromptTypes.None,
+            DeletedDate = deleted ? DateTime.UtcNow.AddDays(-1) : null,
```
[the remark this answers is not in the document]
theMickster: @MGibson1 I think this is a legitimate think to fix

util/Seeder/Scenes/UserLoginCipherScene.cs
```diff
@@ -0,0 +1,58 @@
...
+        if (user == null)
+        {
+            throw new Exception($"User with ID {request.UserId} not found.");
+        }
+
+        var cipher = LoginCipherSeeder.Create(request.UserKeyB64, request.Name, userId: request.UserId, username: request.Username, password: request.Password, uri: request.Uri, notes: request.Notes, fields: request.Fields, reprompt: request.Reprompt, deleted: request.Deleted, favorite: request.Favorite);
+        cipher.Favorites = JsonSerializer.Serialize(new Dictionary<string, bool>
+        {
+            { request.UserId.ToString().ToUpperInvariant(), true}
+        });
```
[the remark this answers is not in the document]
theMickster: @MGibson1 I think this needs to be fixed

util/Seeder/Scenes/UserLoginCipherScene.cs
```diff
@@ -0,0 +1,50 @@
...
+        public required Guid UserId { get; set; }
+        [Required]
+        public required string UserKeyB64 { get; set; }
+        [Required]
+        public required string Name { get; set; }
+        public string? Username { get; set; }
+        public string? Password { get; set; }
+        public string? Uri { get; set; }
+        public string? Notes { get; set; }
+        public IEnumerable<(string name, string value, int type)>? Fields { get; set; }
```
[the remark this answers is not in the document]
MGibson1: resolved in follow up

util/Seeder/Factories/LoginCipherSeeder.cs
```diff
@@ -29,6 +32,9 @@ internal static Cipher Create(
...
+            Favorite = favorite,
```
theMickster: ⚠️ I am fairly certain that setting the favorite here is dead code. Will you please double-check your local instance after running the scenes?

util/Seeder/Scenes/UserCardCipherScene.cs
```diff
@@ -0,0 +1,60 @@
...
+        public required string Number { get; set; }
+        public required string ExpMonth { get; set; }
+        public required string ExpYear { get; set; }
+        public required string Code { get; set; }
+        public string? Notes { get; set; }
+    }
+
+    public class Result
+    {
+        public required Guid CipherId { get; set; }
```
theMickster: ♻️ We have some inconsistency between this and our existing scenes. `init` feels more appropriate for result objects that are constructed once and not mutated. 
```suggestion
        public required Guid CipherId { get; init; }
```

util/Seeder/Scenes/UserIdentityCipherScene.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string? Title { get; set; }
+        public string? FirstName { get; set; }
+        public string? MiddleName { get; set; }
+        public string? LastName { get; set; }
+        public string? Notes { get; set; }
+    }
+
+    public class Result
+    {
+        public required Guid CipherId { get; set; }
```
theMickster: ♻️ We have some inconsistency between this and our existing scenes. `init` feels more appropriate for result objects that are constructed once and not mutated. 
```suggestion
        public required Guid CipherId { get; init; }
```

util/Seeder/Scenes/UserLoginCipherScene.cs
```diff
@@ -0,0 +1,61 @@
...
+        public string? Notes { get; set; }
+        public bool Reprompt { get; set; }
+        public bool Deleted { get; set; }
+        public bool Favorite { get; set; }
+        public IEnumerable<(string name, string value, int type)>? Fields { get; set; }
+    }
+
+    public class Result
+    {
+        public required Guid CipherId { get; set; }
```
theMickster: ♻️ We have some inconsistency between this and our existing scenes. `init` feels more appropriate for result objects that are constructed once and not mutated. 
```suggestion
        public required Guid CipherId { get; init; }
```

PR #7242, opened 2026-03-18
[PM-33819] Enforce use of authorize attributes

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33819
## 📔 Objective

Add automated testing to ensure that all our controllers are using:
- an `Authorize` attribute on the class, and
- either an `Authorize` or `AllowAnonymous` attribute on each method - indicating that it has RBAC or has intentionally been exposed as a non-authorized endpoint

This is done via reflection to inspect attribute usage. There is an example of a test for a specific controller, and also a test that automatically covers any controller in the AC Team namespace. As not all controllers have been migrated to use these attributes yet (they manually check for authz), this latter test exempts existing controllers until they are migrated.

The `PoliciesController` is most of the way there already, so I have done some minor cleanup work to get it passing as the proof of concept.

### DIFF
--- a/test/Api.Test/AdminConsole/Controllers/PoliciesControllerTests.cs
+++ b/test/Api.Test/AdminConsole/Controllers/PoliciesControllerTests.cs
@@ -1,5 +1,4 @@
-﻿using System.Security.Claims;
-using System.Text.Json;
+﻿using System.Text.Json;
 using Bit.Api.AdminConsole.Controllers;
 using Bit.Api.AdminConsole.Models.Request;
 using Bit.Api.AdminConsole.Models.Response.Organizations;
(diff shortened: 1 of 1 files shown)


src/Api/AdminConsole/Controllers/PoliciesController.cs
```diff
@@ -124,34 +107,8 @@ public async Task<ListResponseModel<PolicyResponseModel>> GetByToken(Guid orgId,
         return new ListResponseModel<PolicyResponseModel>(responses);
     }
 
-    // TODO: PM-4097 - remove GetByInvitedUser once all clients are updated to use the GetMasterPasswordPolicy endpoint below
```
eliykat, line 127: This ticket is over 2 years old and I could not find any existing consumers.

eliykat:
Holding until `rc` has been cut to be kind to QA.

PR #7249, opened 2026-03-18
[PM-29728] remove pm-28813-data-recovery-tool feature flag

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29728

## 📔 Objective

Remove `pm-28813-data-recovery-tool` feature flagged logic from clients.

### DIFF
--- a/src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
+++ b/src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
@@ -88,7 +88,7 @@ public class AccountsKeyManagementController : Controller
     [HttpPost("key-management/regenerate-keys")]
     public async Task RegenerateKeysAsync([FromBody] KeyRegenerationRequestModel request)
     {
-        if (!_featureService.IsEnabled(FeatureFlagKeys.PrivateKeyRegeneration) && !_featureService.IsEnabled(FeatureFlagKeys.DataRecoveryTool))
+        if (!_featureService.IsEnabled(FeatureFlagKeys.PrivateKeyRegeneration))
         {
             throw new NotFoundException();
         }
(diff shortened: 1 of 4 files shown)


src/Core/Constants.cs
```diff
@@ -204,7 +204,6 @@ public static class FeatureFlagKeys
     public const string DisableType0Decryption = "pm-25174-disable-type-0-decryption";
     public const string ConsolidatedSessionTimeoutComponent = "pm-26056-consolidated-session-timeout-component";
     public const string V2RegistrationTDEJIT = "pm-27279-v2-registration-tde-jit";
-    public const string DataRecoveryTool = "pm-28813-data-recovery-tool";
```
Thomas-Avery, line 207: We can't remove it here because we still need to send it down for older clients.

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
```diff
@@ -88,7 +88,7 @@ public AccountsKeyManagementController(IUserService userService,
     [HttpPost("key-management/regenerate-keys")]
     public async Task RegenerateKeysAsync([FromBody] KeyRegenerationRequestModel request)
     {
-        if (!_featureService.IsEnabled(FeatureFlagKeys.PrivateKeyRegeneration) && !_featureService.IsEnabled(FeatureFlagKeys.DataRecoveryTool))
```
Thomas-Avery, line 91: This was passing the check if `DataRecoveryTool` was true and `PrivateKeyRegeneration` was false. 

Now this wouldn't pass the check that the data recovery tool depends on. I think we need to remove both feature flag checks here.

src/Core/Constants.cs
mzieniukbw: Data recovery tool is Web only. My understanding was that it is safe to remove as long as server and web client's happen at the same time.

src/Api/KeyManagement/Controllers/AccountsKeyManagementController.cs
mzieniukbw: Good catch. Removed in e4442cdb

src/Core/Constants.cs
Thomas-Avery: You're right I forgot it was web only.

Thomas-Avery (approved):
LGTM

PR #7251, opened 2026-03-18
[PM-33415] Fix add-on items being billed immediately instead of prorated

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33415

## 📔 Objective

Fixes a QA bug from the PM-32581 work where storage and SM service accounts were classified as "structural" changes, forcing `ProrationBehavior.AlwaysInvoice` (immediate billing) instead of `ProrationBehavior.CreateProrations`.

**Root cause**: `IsStructural` on `OrganizationSubscriptionChange` treated all `AddItem`, `RemoveItem`, and `UpdateItemQuantity(qty=0)` as structural. There was no way to distinguish between structural changes (adding SM product, changing plan price) and add-on changes (storage, service accounts).

**Fix**: Replace per-change `IsStructural` inspection with a changeset-level `ChargeImmediately` flag. Replace generic builder methods (`AddItem`, `RemoveItem`, etc.) with domain-specific methods (`AddStorage`, `UpdateServiceAccounts`, `EstablishSponsorship`, `ChangePasswordManagerPrice`, etc.) that encapsulate correct Stripe price IDs and proration behavior. Structural methods automatically set `ChargeImmediately = true`; add-on and seat methods leave the default (`false` = prorate).

### DIFF
--- a/src/Core/Billing/Organizations/Models/OrganizationSubscriptionChangeSet.cs
+++ b/src/Core/Billing/Organizations/Models/OrganizationSubscriptionChangeSet.cs
@@ -1,4 +1,4 @@
-﻿using Bit.Core.Models.StaticStore;
+using Bit.Core.Models.StaticStore;
 using OneOf;
 
 namespace Bit.Core.Billing.Organizations.Models;
@@ -9,7 +9,7 @@ namespace Bit.Core.Billing.Organizations.Models;
 public record AddItem(string PriceId, int Quantity);
 
 /// <summary>
-/// Replaces an existing line item's price (e.g. switching from monthly to annual billing).
+/// Replaces an existing line item's price (e.g. upgrading from Teams to Enterprise).
 /// Optionally updates the quantity; if <c>null</c>, the current quantity is preserved.
 /// </summary>
 public record ChangeItemPrice(string CurrentPriceId, string UpdatedPriceId, int? Quantity);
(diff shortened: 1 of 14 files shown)


src/Core/Billing/Organizations/Models/OrganizationSubscriptionChangeSet.cs
```diff
@@ -43,90 +41,171 @@ public static implicit operator OrganizationSubscriptionChange(RemoveItem remove
...
+    public OrganizationSubscriptionChangeSetBuilder EstablishSponsorship(SponsoredPlan sponsoredPlan)
     {
-        _changes.Add(new AddItem(priceId, quantity));
+        _changes.Add(new RemoveItem(currentPlan.PasswordManager.StripePlanId));
+        _changes.Add(new AddItem(sponsoredPlan.StripePlanId, 1));
+        _chargeImmediately = true;
         return this;
     }
 
-    public OrganizationSubscriptionChangeSetBuilder ChangeItemPrice(string currentPriceId, string updatedPriceId, int? quantity = null)
```
amorask-bitwarden, line 112: ❓ What did we do with the `quantity` parameter here? Was it being used by any callers?

amorask-bitwarden: No callers were using the `quantity` parameter. The old generic `ChangeItemPrice` method had `int? quantity = null` as an optional parameter, and all four call sites in `UpgradeOrganizationPlanVNextCommand` omitted it (relying on the `null` default, which tells Stripe to preserve the existing quantity).

The new domain-specific methods (`ChangePasswordManagerPrice`, `ChangeStoragePrice`, `ChangeSecretsManagerSeatPrice`, `ChangeServiceAccountPrice`) all pass `null` for quantity internally — same behavior, just encapsulated.

PR #7253, opened 2026-03-19
[PM-31771]Default payment method not updated within subscription

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31771

## 📔 Objective

This PR fixes a bug where the payment method was not being properly updated when restarting a canceled organization
  subscription through the resubscribe flow.

  **Background**

  When an organization owner restarted a canceled subscription by providing a new payment method via the `/subscription/restart`
  endpoint, the old payment method was still being used to charge the subscription instead of the newly provided payment method.

  **Changes**

  - Modified `RestartSubscriptionCommand.cs`: Added explicit fetching of the customer's current default payment method and passing
  it to the SubscriptionCreateOptions when creating the new subscription
  - Updated `RestartSubscriptionCommandTests.cs`: Added mocks for GetCustomer and assertions to verify DefaultPaymentMethod is
  explicitly set in all success scenarios

  **Technical Details**

  Previously, the subscription restart flow relied on Stripe's OffSession = true parameter to automatically use the customer's
  default payment method. However, this could result in using a stale/cached payment method due to timing issues between updating
   the payment method and creating the subscription.

  The fix explicitly fetches the customer object after the payment method has been updated and passes the DefaultPaymentMethodId
  to the SubscriptionCreateOptions, ensuring the correct payment method is always used regardless of Stripe's caching behavior.

  Testing

  - All 13 existing unit tests pass
  - Added verification that DefaultPaymentMethod is explicitly set in subscription creation across all test scenarios
  - Manual testing confirmed payment method is now properly updated when restarting subscriptions

  Fixes: [Issue tracking payment method update bug during subscription restart]
## 📸 Screenshots

**Before the Fix Video**

https://github.com/user-attachments/assets/21c935b6-619c-43bb-b731-37993f176bc4

**After the Fix Video**

https://github.com/user-attachments/assets/a0182b9d-9527-49f9-a8d9-9643d3ed7be5

### DIFF
--- a/global.json
+++ b/global.json
@@ -6,6 +6,6 @@
   "msbuild-sdks": {
     "Microsoft.Build.Traversal": "4.1.0",
     "Microsoft.Build.Sql": "1.0.0",
-    "Bitwarden.Server.Sdk": "1.4.0"
+    "Bitwarden.Server.Sdk": "1.5.1"
   }
 }


src/Core/Billing/Subscriptions/Commands/RestartSubscriptionCommand.cs
```diff
@@ -162,11 +162,14 @@ private async Task RestartOrganizationSubscriptionAsync(
...
         }
 
+        var customer = await subscriberService.GetCustomer(organization);
+
         var options = new SubscriptionCreateOptions
         {
             AutomaticTax = new SubscriptionAutomaticTaxOptions { Enabled = true },
             CollectionMethod = CollectionMethod.ChargeAutomatically,
             Customer = canceledSubscription.CustomerId,
+            DefaultPaymentMethod = customer.InvoiceSettings?.DefaultPaymentMethodId,
```
amorask-bitwarden, line 172: ❓ This is at the subscription-level. As far as I know, we never set default payment methods on subscriptions directly anywhere else throughout the platform. What happens when a customer updates their payment method in the Web Vault?

PR #7257, opened 2026-03-19
[PM-33895] Filter `[BindNever]` parameters from OpenAPI schema

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33895
https://github.com/bitwarden/sdk-internal/pull/728#discussion_r2841135202
https://bitwarden.slack.com/archives/C05LRC3DBDK/p1771858123530539

## 📔 Objective

Create a new Swagger filter to filter out complex type parameters using `[BindNever]`. This is required to correctly generate the schema for some billing functions that look like this:
```cs
[HttpGet("credit")]
[InjectUser]
public async Task<IResult> GetCreditAsync([BindNever] User user) { ... }
``` 

The issue here is caused by Swashbuckle first expanding `User` into one query parameter per field (`id`, `name`, etc) before trying to filter out the query parameter `user`, which doesn't exist anymore and so it doesn't get correctly filtered out. This causes the OpenAPI schema and the SDK-generated code to have a lot of invalid query parameters:
https://github.com/bitwarden/sdk-internal/blob/7637fdfd66c4403b9f718d85c4cbd5bde15288d9/crates/bitwarden-api-api/src/apis/account_billing_v_next_api.rs#L31
This filter fixes this by manually filtering all the properties of any parameter annotated with `[BindNever]`. I've tested it locally and this fixes the invalid code generation for the billing endpoints.

The cause of this issue seems to me like a Swashbuckle bug so I'm thinking about reporting it upstream, but this filter should at least solve it for the time being.


djsmith85 (approved):
Changes are looking good. Please do look into reporting this issue upstream.

PR #7265, opened 2026-03-19
[PM-34082] Seed passkeys

MGibson1:
## Tracking

[PM-34082](https://bitwarden.atlassian.net/browse/PM-34082)

## 📔 Objective

Allows for seeding passkeys on login data. They won't work, but neither will any seeded username/passwords.

## 📸 Screenshots

[PM-34082]: https://bitwarden.atlassian.net/browse/PM-34082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/util/Seeder/Factories/LoginCipherSeeder.cs
+++ b/util/Seeder/Factories/LoginCipherSeeder.cs
@@ -37,7 +37,7 @@ internal static class LoginCipherSeeder
                 Password = password,
                 Totp = totp,
                 Uris = string.IsNullOrEmpty(uri) ? null : [new LoginUriViewDto { Uri = uri }],
-                Fido2Credentials = passkeys == null ? null : passkeys.Select(r => PasskeyViewFactory.Create(r.rpName, r.userName)).ToList()
+                Fido2Credentials = passkeys == null ? null : passkeys.Select(r => CreateFido2Credential(r.rpName, r.userName)).ToList()
             },
             Reprompt = reprompt ? RepromptTypes.Password : RepromptTypes.None,
             DeletedDate = deleted ? DateTime.UtcNow.AddDays(-1) : null,
@@ -82,11 +82,8 @@ internal static class LoginCipherSeeder
         var encrypted = CipherEncryption.Encrypt(cipherView, encryptionKey);
         return CipherEncryption.CreateEntity(encrypted, encrypted.ToLoginData(), CipherType.Login, organizationId, userId);
     }
-}
 
-public class PasskeyViewFactory
-{
-    public static Fido2CredentialViewDto Create(string rpName, string userName)
+    public static Fido2CredentialViewDto CreateFido2Credential(string rpName, string userName)
     {
         // Generate ECDSA P-256 private key in PKCS#8 format
         using var ecdsa = ECDsa.Create(ECCurve.NamedCurves.nistP256);


util/Seeder/Factories/LoginCipherSeeder.cs
```diff
@@ -77,3 +83,31 @@
...
+{
+    public static Fido2CredentialViewDto Create(string rpName, string userName)
+    {
+        // Generate ECDSA P-256 private key in PKCS#8 format
+        using var ecdsa = ECDsa.Create(ECCurve.NamedCurves.nistP256);
+        var keyValue = CoreHelpers.Base64UrlEncode(ecdsa.ExportPkcs8PrivateKey());
+
+        // Generate 16-byte random user handle and encode as unpadded base64url
+        var userHandleBytes = new byte[16];
+        new Random().NextBytes(userHandleBytes);
```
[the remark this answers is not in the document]
MGibson1: Does not need CSPRNG, just need some random bytes as a user handle, which is an annonymous identifier mapping to a private key in webauthn

util/Seeder/Factories/LoginCipherSeeder.cs
```diff
@@ -75,3 +83,31 @@ internal static Cipher CreateFromSeed(
         return CipherEncryption.CreateEntity(encrypted, encrypted.ToLoginData(), CipherType.Login, organizationId, userId);
     }
 }
+
+public class PasskeyViewFactory
```
theMickster: Two comments: 
1. I don't see this in-use yet outside of the Seeder library, and if that's true, then will you please change the protection level to internal? 

2. Do you mind putting this class into it's own code file?

util/Seeder/Models/CipherViewDto.cs
```diff
@@ -81,6 +81,51 @@ public class LoginViewDto
...
     public string? Totp { get; set; }
+
+    [JsonPropertyName("fido2Credentials")]
+    public List<Fido2CredentialViewDto>? Fido2Credentials { get; set; }
+}
+
+public class Fido2CredentialViewDto
+{
+    [JsonPropertyName("creationDate")]
+    public DateTime CreationDate { get; set; }
```
theMickster: 🤔 Should this have a default value of DateTime.UtcNow to ensure that each seeded passkey has a valid create date?

util/Seeder/Factories/LoginCipherSeeder.cs
MGibson1: I'd thought about doing that, but it's a factory specific to Login Ciphers. Passkeys live within the login cipher data object, so it makes sense to me for this to be limited in scope to the login cipher seeder. That said, I think it probably makes sense to either just make this a method on the `LoginCipherSeeder` class or a private class available to that seeder.

What do you think?

util/Seeder/Models/CipherViewDto.cs
MGibson1: that's consistent with the other DTOs :+1:

util/Seeder/Factories/LoginCipherSeeder.cs
theMickster: The method appears to be a better fit.

PR #7266, opened 2026-03-20
[PM-4128] Remove nullability of Send.Data and Send.Keys

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-4128

## 📔 Objective

While MSSQL [already enforces](https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Tools/Tables/Send.sql) that `Send.Data` and `Send.Key` are not nullable, the `Send` entity has been updated to reflect this and EF migration scripts have been generated. 

Note that as opposed to adding the `required` keyword to the now non-nullable properties, they are assigned a default initial value of empty string, this is ephemeral as the server will assign the key sent in the client request and assemble the JSON data from non-null components of the same request. 

See related PR https://github.com/bitwarden/server/pull/7212


mcamirault (dismissed):

rkac-bw:
Missing NULL backfill before NOT NULL constraint in all 3 EF migrations
 Migration would fail if any not null, is this a concern?

harr1424:
> Missing NULL backfill before NOT NULL constraint in all 3 EF migrations Migration would fail if any not null, is this a concern?

@rkac-bw  This migration assumes that there are no NULL values in columns `Data` or `Key` based upon how `Send` objects are created: 

https://github.com/bitwarden/server/blob/8310ea154695f1217ab4c6dddacfa2c6863b5764/src/Core/Tools/Entities/Send.cs#L46-L64

It would be prudent to query production data for any unexpected NULL values, unless you prefer the backfill occurs regardless.

rkac-bw:
> > Missing NULL backfill before NOT NULL constraint in all 3 EF migrations Migration would fail if any not null, is this a concern?
> 
> @rkac-bw This migration assumes that there are no NULL values in columns `Data` or `Key` based upon how `Send` objects are created:
> 
> https://github.com/bitwarden/server/blob/8310ea154695f1217ab4c6dddacfa2c6863b5764/src/Core/Tools/Entities/Send.cs#L46-L64
> 
> It would be prudent to query production data for any unexpected NULL values, unless you prefer the backfill occurs regardless.

 Cloud is fine — SQL Server has enforced NOT NULL on these columns since the table was created.

  For self-hosted instances on MySQL/MariaDB, PostgreSQL, or SQLite, we'd recommend a defensive backfill to
  protect against any unexpected NULLs. You can just edit the generated migration files and add
  migrationBuilder.Sql() calls at the top of each Up() method:

  MySQL/MariaDB:
  migrationBuilder.Sql("UPDATE `Send` SET `Key` = '' WHERE `Key` IS NULL;");
  migrationBuilder.Sql("UPDATE `Send` SET `Data` = '' WHERE `Data` IS NULL;");

  PostgreSQL:
  migrationBuilder.Sql("UPDATE \"Send\" SET \"Key\" = '' WHERE \"Key\" IS NULL;");
  migrationBuilder.Sql("UPDATE \"Send\" SET \"Data\" = '' WHERE \"Data\" IS NULL;");

  SQLite:
  migrationBuilder.Sql("UPDATE \"Send\" SET \"Key\" = '' WHERE \"Key\" IS NULL;");
  migrationBuilder.Sql("UPDATE \"Send\" SET \"Data\" = '' WHERE \"Data\" IS NULL;");

  No-op if no NULLs exist, prevents a migration failure if they do.

rkac-bw (dismissed):
lgtm

mcamirault (dismissed):

rkac-bw (approved):

mcamirault (approved):

PR #7278, opened 2026-03-23
[PM-33980] Only verify `UseMyItems` when claim exists

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33980

## 📔 Objective

The `2026.3.0` release added `UseMyItems` to the license claims factory and `VerifyData` comparison, but self-hosted instances that updated to `2026.3.0` still hold license files generated before this claim existed. The database migration sets `UseMyItems = true` for orgs with `UsePolicies = true`, but the stale license has no `UseMyItems` claim (defaulting to `false`). When `LicensingService.ValidateOrganizationsAsync()` calls `VerifyData`, the mismatch causes it to return `false`, disabling the organization with "Invalid data."

This PR wraps the `UseMyItems` comparison in a conditional `HasClaim` check so that licenses generated before the claim existed skip the comparison entirely. Once the org syncs a fresh license from cloud `2026.3.0+`, the claim will be present and validated normally.

### DIFF
--- a/src/Core/Billing/Organizations/Models/OrganizationLicense.cs
+++ b/src/Core/Billing/Organizations/Models/OrganizationLicense.cs
@@ -42,6 +42,9 @@ public class OrganizationLicense : ILicense
     /// 1. Use the claims-based system instead of adding properties here
     /// 2. Add new claims to the license token
     /// 3. Validate claims in the <see cref="CanUse"/> and <see cref="VerifyData"/> methods
+    /// 4. In <see cref="VerifyData"/>, wrap new claim comparisons in a conditional
+    ///    HasClaim check so that licenses generated before the claim existed still
+    ///    validate successfully (introduced after PM-33980)
     /// </para>
     /// <para>
     /// This constructor is maintained only for backward compatibility with existing licenses.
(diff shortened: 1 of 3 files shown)


src/Core/Billing/Organizations/Models/OrganizationLicense.cs
```diff
@@ -438,6 +441,11 @@ public bool VerifyData(
             ? organization.PlanType is PlanType.FamiliesAnnually or PlanType.FamiliesAnnually2025
             : organization.PlanType == claimedPlanType;
 
+        // IMPORTANT: UseMyItems is the first claim to require a conditional HasClaim
+        // check because self-hosted instances may hold license files generated before
+        // this claim existed, where GetValue<T> returns the type's default (false),
+        // causing a mismatch that disables the org. Future claims MUST follow this
+        // same pattern. See PM-33980.
```
kdenney, line 448: ❓ @eliykat worked on [this readme file](https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/OrganizationFeatures/OrganizationAbility/README.md) recently with detailed instructions for adding new organization abilities and it contains a detailed section for editing this code for self-host licensing support. I am now wondering if we missed something in those instructions that led to this? Do you think the instructions in that readme file need to also be updated to include these new instructions?

cturnbull-bitwarden (dismissed):

kdenney (commented):
This looks correct, but I have one question about improving documentation.

src/Core/Billing/Organizations/Models/OrganizationLicense.cs
amorask-bitwarden: @kdenney This PR updates the documentation to add the requirement for the conditional check on new licensing properties, but I think we're going to have further, more detailed discussions on the potential deprecation of this validation flow altogether with a stronger focus on the JWT. Those will be part of a separate discussion since this change will be a hotfix.

kdenney (approved):
Thanks for adding to the readme!

src/Core/Billing/Organizations/Models/OrganizationLicense.cs
amorask-bitwarden: (Book keeping) @kdenney Discussed in Slack, but I misunderstood which docs you were referring to. Those have been updated here: https://github.com/bitwarden/server/pull/7278/commits/a3b0de0d6ce67490e4cdbc5e0b938f388bd1df77

sven-bitwarden (approved):

PR #7279, opened 2026-03-23
PM-33964 - Cache seeded RSA key material in LazyLock

theMickster:
## 🎟️ Tracking

[PM-33964](https://bitwarden.atlassian.net/browse/PM-33964)
[PM-33967](https://bitwarden.atlassian.net/browse/PM-33967)

## 📔 Objective

The Seeder's `keypair()` function re-parses the same hardcoded RSA PEM constant and re-derives DER bytes on every call — identical work repeated per user inside `CreateUsersStep`'s `Parallel.For`. This PR caches the invariant material in a `LazyLock<CachedRsaMaterial>` static so PEM parsing and DER conversion happen exactly once, while per-user symmetric key encryption (`encrypt_with_key`) remains per-call.

### Key changes

- **RSA key material caching** (`util/RustSdk/rust/src/lib.rs`) — `LazyLock<CachedRsaMaterial>` caches `Pkcs8PrivateKeyBytes` + `SpkiPublicKeyBytes` from the seeded PEM constant. `keypair()` clones cached DER and encrypts per-user. Thread-safe, immutable, no mutable global state.
- **Rust unit tests** for `keypair()` — validates EncString format, per-user key isolation, and cache stability
- **Integration tests** for `GenerateUserKeys()` — validates all output fields and end-to-end public key org wrapping

## 🧪 Testing

<details><summary>Expand for A/B benchmark and regression results</summary>

### A/B Performance (main vs cached, single run per preset)

| Preset | Users | Ciphers | Main | Cached | Delta |
|---|---|---|---|---|---|
| Large Org Modern (500u/10Kc/85g) | 501 | 10,000 | 38.51s | **32.78s** | **-14.9%** |
| xl-highperm-weyland-yutani | 5,001 | 24,344 | 50.89s | **49.37s** | **-3.0%** |
| zero-knowledge-labs-enterprise | 386 | 265 | 18.05s | **17.89s** | **-0.9%** |
| enterprise-basic (600k KDF) | 5 | 24 | 5.26s | **4.92s** | **-6.4%** |

Performance improved or held steady across all 4 presets. No regressions.

### Entity Count Verification 

| Preset | Users | Groups | Collections | Ciphers | Match |
|---|---|---|---|---|---|
| Large Org Modern | 501 | 85 | — | 10,000 | ✅ |
| xl-highperm-weyland-yutani | 5,001 | 500 | 1,200 | 24,344 | ✅ |
| zero-knowledge-labs-enterprise | 386 | 42 | 53 | 265 | ✅ |
| enterprise-basic (600k KDF) | 5 | 1 | 36 | 24 | ✅ |

### Security Review

- Cached DER bytes are the same key material already present as a PEM `const` — no new security surface
- Per-user `encrypt_with_key` still per-call — zero-knowledge invariant preserved
- `LazyLock` is thread-safe, immutable after init, no interior mutability

</details>

[PM-33964]: https://bitwarden.atlassian.net/browse/PM-33964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-33967]: https://bitwarden.atlassian.net/browse/PM-33967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/SeederApi.IntegrationTest/RustSdkCipherTests.cs
+++ b/test/SeederApi.IntegrationTest/RustSdkCipherTests.cs
@@ -452,4 +452,33 @@ public sealed class RustSdkCipherTests
         Assert.DoesNotContain("BEGIN FAKE OPENSSH PRIVATE KEY", cipher.Data);
         Assert.DoesNotContain("ssh-ed25519", cipher.Data);
     }
+
+    [Fact]
+    public void GenerateUserKeys_ReturnsValidKeys()
+    {
+        var keys = RustSdkService.GenerateUserKeys("test@example.com", "TestPassword123!", 5000);
+
+        Assert.NotNull(keys.MasterPasswordHash);
+        Assert.NotEmpty(keys.MasterPasswordHash);
+        Assert.NotNull(keys.Key);
+        Assert.NotEmpty(keys.Key);
+        Assert.NotNull(keys.EncryptedUserKey);
+        Assert.StartsWith("2.", keys.EncryptedUserKey);
+        Assert.NotNull(keys.PublicKey);
+        Assert.NotEmpty(keys.PublicKey);
+        Assert.NotNull(keys.PrivateKey);
+        Assert.StartsWith("2.", keys.PrivateKey);
+    }
+
+    [Fact]
+    public void GenerateUserKeys_PublicKey_WorksWithOrganizationKeyWrapping()
+    {
+        var userKeys = RustSdkService.GenerateUserKeys("test@example.com", "TestPassword123!", 5000);
+        var orgKeys = RustSdkService.GenerateOrganizationKeys();
+
+        var wrappedOrgKey = RustSdkService.GenerateUserOrganizationKey(userKeys.PublicKey, orgKeys.Key);
+
+        Assert.NotNull(wrappedOrgKey);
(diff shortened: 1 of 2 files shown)


Hinton:
The intention was never to only use a single RSA key for all seeded data but rather have some rotation. How does this fit into that aspiration?

theMickster:
> The intention was never to only use a single RSA key for all seeded data but rather have some rotation. How does this fit into that aspiration?

I agree that sharing a single keypair across all seeded users masks key-mixing bugs that would fail in production. The caching in this PR is still useful for scale testing where we need 5K+ users fast.

I'm creating a follow-up ticket to add a `--keypair-mode` shared|unique CLI flag & preset-level uniqueKeyPairs field. 
My  plan:
 - Fresh RSA-2048 generation via bitwarden_crypto::make_key_pair() as the default (correct behavior)
 - Cached shared key as opt-in for scale presets only
 - Dev/QA/feature presets all default to unique keys

This PR stays as-is (caching the shared path), and the follow-up adds the unique path alongside it.

PR #7281, opened 2026-03-23
Add SeederApi PlayData delete scheduled job

MGibson1:
Use Quartz-based hosted service to clear old play data

We need to stop possible bloat of databases should users of a seeded data fail to appropriately clean up after themselves.

Using the hosted services present in other projects, this adds an alive job and play data delete job to the SeederApi


theMickster:
Couple thoughts
Have we found if there are scenarios where we would not want the Seeded data deleted using this scheduled job?
For clarity, we are focused here on cleaning up after manual testing, automated testing, or both?

MGibson1:
All data tagged with a PlayId (i.e. PlayData) is by definition opted in to be ephemeral by the client requesting it's creation through the `x-play-id` header.

If, for some reason, you want to seed data that is not ephemeral, all you have to do is not include that header data and this job will not touch it.

In some respects, we're assuming good actors here, but the endpoints on publicly accessible lower environments are not publicly accessible.

util/SeederApi/Jobs/DeleteOldPlayDataJob.cs
```diff
@@ -0,0 +1,35 @@
+﻿// FIXME: Update this file to be null safe and then delete the line below
```
theMickster: ⚠️ Are we able to avoid adding a FIXME to brand new code at this time?

util/SeederApi/Jobs/AliveJob.cs
```diff
@@ -0,0 +1,28 @@
...
+using Bit.Core.Jobs;
+using Bit.Core.Settings;
+using Quartz;
+
+namespace Bit.SeederApi.Jobs;
+
+public class AliveJob : BaseJob
+{
+    private readonly GlobalSettings _globalSettings;
+    private HttpClient _httpClient = new HttpClient();
```
theMickster: 🤔 Noting this as a known anti-pattern that causes socket exhaustion. Did we copy this from the src/Admin job?

MGibson1: sure did. I don't know what issue, do we have a fix? Is this bump job unnecessary?

theMickster: The AliveJob is fine — the issue is `private HttpClient _httpClient = new HttpClient()`. Directly instantiating HttpClient as a field is a known antipattern that causes socket exhaustion under load because each instance gets its own connection pool and doesn't respect DNS TTLs.

The fix is to inject `IHttpClientFactory` and call `httpClientFactory.CreateClient()` in the constructor. We'll also need `services.AddHttpClient()` in program startup. The job itself can stay as-is.

util/SeederApi/Startup.cs
```diff
@@ -41,6 +41,9 @@ public void ConfigureServices(IServiceCollection services)
         services.AddQueries();
 
         services.AddControllers();
+
+        Jobs.JobsHostedService.AddJobsServices(services);
```
theMickster, line 54: 🤔 Deleting data with a hosted service always give me a little heartburn. Have we implemented opt-in techniques using environment settings in other places, by chance? I have done that in the past where my teams implemented background services that mutated data and gave us a better feel of control (especially when developing) knowing that one had to explicitly set the setting to ON to mutate/delete data?

MGibson1: Interesting. Are you concerned with unintentional data loss? To some extent the goal of this job is to _enforce_ that play data is ephemeral.

That can be a server setting if we want, I suppose, but maybe a better tweak would be a lifetime?

theMickster: In the ephemeral environments, I am not too concerned because I would not expect an ephemeral environment to live long enough to see the job run. Since the intention is really deleting this ephemeral data from long lived environments then I'm cool with it. 

Only real issue is if you're developing tests locally or on one of the QA team VMs and you're expecting data to be there over a couple days (like a weekend) then you'd better be ready for the Seeder to clean itself up.

Did we add a README.md? (Sorry on the GH app so hard to go back n forth).

MGibson1: no worries, I'm doing a bad job of following up on this PR. I've now updated the readme to callout the delete job's configuration and dependency on play id header

theMickster (approved):

PR #7283, opened 2026-03-23
[PM-32218] Create Session Complete Handler

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32218

## 📔 Objective

Adds support for handling the Stripe `checkout.session.completed` webhook event, enabling the system to update users' premium status when a premium subscription is purchased via Stripe Session Checkout. The changes include new handler implementations, service interface updates, and dependency registrations.

**Stripe Checkout Session Completed Webhook Support:**

* Added a new constant for `checkout.session.completed` in `HandledStripeWebhook` to identify the event type.
* Introduced the `ICheckoutSessionCompletedHandler` interface and its implementation `CheckoutSessionCompletedHandler`, which processes the event by updating the user to premium status if a valid premium subscription is detected. 
* Registered the new handler in dependency injection and updated the `StripeEventProcessor` to route the event to the handler.

**Stripe Service and Adapter Enhancements:**

* Extended `IStripeEventService` and `StripeEventService` with a `GetCheckoutSession` method to extract or fetch the latest Checkout Session from Stripe, and updated logic to retrieve customer metadata for this event type.
* Added `GetCheckoutSessionAsync` to `IStripeAdapter` and implemented it in `StripeAdapter` for retrieving sessions from Stripe's API.

## 📸 Screenshots

https://github.com/user-attachments/assets/c8c7ae08-5efb-4c7a-ae32-fd50894871cd

### DIFF
--- a/src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
+++ b/src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
@@ -34,6 +34,7 @@ public class SubscriptionUpdatedHandler : ISubscriptionUpdatedHandler
     private readonly IProviderRepository _providerRepository;
     private readonly IProviderService _providerService;
     private readonly IPushNotificationAdapter _pushNotificationAdapter;
+    private readonly IPriceIncreaseScheduler _priceIncreaseScheduler;
     private readonly IFeatureService _featureService;
     private readonly ILogger<SubscriptionUpdatedHandler> _logger;
 
@@ -52,6 +53,7 @@ public class SubscriptionUpdatedHandler : ISubscriptionUpdatedHandler
         IProviderRepository providerRepository,
         IProviderService providerService,
         IPushNotificationAdapter pushNotificationAdapter,
+        IPriceIncreaseScheduler priceIncreaseScheduler,
         IFeatureService featureService,
         ILogger<SubscriptionUpdatedHandler> logger)
     {
@@ -71,6 +73,7 @@ public class SubscriptionUpdatedHandler : ISubscriptionUpdatedHandler
         _providerRepository = providerRepository;
         _providerService = providerService;
         _pushNotificationAdapter = pushNotificationAdapter;
+        _priceIncreaseScheduler = priceIncreaseScheduler;
         _featureService = featureService;
         _logger = logger;
     }
@@ -218,6 +221,8 @@ public class SubscriptionUpdatedHandler : ISubscriptionUpdatedHandler
 
     private async Task SetSubscriptionToCancelAsync(Subscription subscription)
     {
+        await _priceIncreaseScheduler.Release(subscription.CustomerId, subscription.Id);
+
         if (subscription.TestClock != null)
         {
             await WaitForTestClockToAdvanceAsync(subscription.TestClock);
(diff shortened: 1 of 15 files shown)


src/Billing/Services/Implementations/StripeEventService.cs
```diff
@@ -82,6 +85,17 @@ public async Task<Subscription> GetSubscription(Event stripeEvent, bool fresh =
...
 
+    public async Task<Session> GetCheckoutSession(Event stripeEvent, bool fresh = false, List<string>? expand = null)
+    {
+        var checkoutSession = Extract<Session>(stripeEvent);
+        if (!fresh)
+        {
+            return checkoutSession;
+        }
+
+        return await stripeAdapter.GetCheckoutSessionAsync(checkoutSession.Id, new SessionGetOptions { Expand = expand });
```
sbrown-livefront, line 96: 📝 I added the Adapter here instead of the Facade. Do we want to update facade calls here to the Adapter?

amorask-bitwarden: Yeah I added facade when I was new and naive 😆

I'll probably get a PR to convert them en-masse in the future and get rid of facade so what you have here is great. No need to change the rest of the file though.

src/Billing/Services/Implementations/CheckoutSessionCompletedHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+    ILogger<CheckoutSessionCompletedHandler> logger)
+    : ICheckoutSessionCompletedHandler
+{
+    public async Task HandleAsync(Event parsedEvent)
+    {
+        var session = await stripeEventService.GetCheckoutSession(parsedEvent, true);
+
+        if (string.IsNullOrWhiteSpace(session.SubscriptionId))
+        {
+            logger.LogWarning("Checkout Session {SessionId} has no subscription ID", session.Id);
```
amorask-bitwarden: ⛏️ I think this is worthy of error-level. No subscription on a completed checkout session is something we'll want to investigate quickly.

src/Billing/Services/Implementations/CheckoutSessionCompletedHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+            return;
+        }
+
+        var subscription = await stripeAdapter.GetSubscriptionAsync(session.SubscriptionId);
+        if (subscription is null)
```
amorask-bitwarden, line 26: ⛏️  Resources coming back from `IStripeAdapter` will never be `null`. Stripe's SDK throws a `StripeException` with error code `resource_missing` so this will never be hit.

src/Billing/Services/Implementations/CheckoutSessionCompletedHandler.cs
sbrown-livefront: Addressed in https://github.com/bitwarden/server/pull/7283/commits/e939b6df9b6b2411c05a6b2f6b0a33926abb4c6c

src/Billing/Services/Implementations/CheckoutSessionCompletedHandler.cs
sbrown-livefront: I cut down the extra adapter call by expanding the session. I think it's possible to be null now with the expand (1 less call though). Addressed in: https://github.com/bitwarden/server/pull/7283/commits/b23b427f192c79c490a4e8f428d6c3e1cea8daf0

src/Billing/Services/Implementations/CheckoutSessionCompletedHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    private async Task UpdateDefaultPaymentMethodAsync(string? defaultPaymentMethodId, string customerId)
+    {
+        if (string.IsNullOrWhiteSpace(defaultPaymentMethodId))
+        {
+            logger.LogWarning("No default payment method found for customer {CustomerId}", customerId);
+            return;
+        }
+
+        await stripeAdapter.UpdateCustomerAsync(customerId, new CustomerUpdateOptions
```
sbrown-livefront, line 82: @amorask-bitwarden I figured out how to set the default payment for the customer after session completion. There is no setting to do this when creating the session, but since it's in `subscription` mode it saves it to the default payment of the subscription. This just adds it to the customer.

PR #7284, opened 2026-03-23
Add Table and Entity to support Bitwarden Receive

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33985

## 📔 Objective

Initial work to add a `Receive` table to the database(s) and a corresponding C# model class.

### DIFF
--- a/src/Core/Tools/Entities/Receive.cs
+++ b/src/Core/Tools/Entities/Receive.cs
@@ -18,7 +18,7 @@ public class Receive : ITableObject<Guid>
     /// <summary>
     /// Identifies the user that created this Receive.
     /// </summary>
-    public Guid? UserId { get; set; }
+    public Guid UserId { get; set; }
 
     /// <summary>
     /// Stores data containing or pointing to the transmitted file(s). JSON.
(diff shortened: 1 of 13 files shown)


src/Core/Tools/Entities/Receive.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    /// <summary>
+    /// Uniquely identifies this Receive.
+    /// </summary>
+    public Guid Id { get; set; }
+
+    /// <summary>
+    /// Identifies the user that created this Receive.
+    /// </summary>
+    public Guid? UserId { get; set; }
```
Thomas-Avery: 💭  It is really easy to make a non-null column into nullable, but not the inverse. Might be a good idea to start out with this as non-nullable.

src/Core/Tools/Entities/Receive.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// <summary>
+    /// Stores data containing or pointing to the transmitted file(s). JSON.
+    /// </summary>
+    public required string Data { get; set; }
+
+    /// <summary>
+    /// Stores the data's encryption key. Encrypted.
+    /// </summary>
+    public required string Key { get; set; }
```
Thomas-Avery:
```suggestion
     /// <summary>
    /// The shared content encryption key (SCEK) wrapped by the owners userKey.
    /// </summary>
    public required string UserKeyWrappedSharedContentEncryptionKey { get; set; }

    /// <summary>
    /// The private key wrapped by the owners userKey.
    /// </summary>
    public required string UserKeyWrappedPrivateKey { get; set; }

    /// <summary>
    /// The public key wrapped by the shared content encryption key (SCEK).
    /// </summary>
    public required string ScekWrappedPublicKey { get; set; }
```

Optional, we can do a follow up PR and exclude this stuff from this one if we want to. 

Looking at this I think it might make sense to have:

- SharedData 
  - JSON blob of the data encrypted by SCEK (name and publicKey).
- FileData
  - And this would be additive when uploaders add another file?

But we should look into that later once we have more stuff implemented.

util/Migrator/DbScripts/2026-03-24_00_AddReceiveTable.sql
```diff
@@ -0,0 +1,25 @@
+-- Add Receive table
+IF OBJECT_ID('[dbo].[Receive]', 'U') IS NULL
```
Thomas-Avery:
```suggestion
IF OBJECT_ID('[dbo].[Receive]) IS NULL
```

Thomas-Avery (approved):
LGTM

harr1424:
Please see https://github.com/bitwarden/server/pull/7306 which is being used to contain an updated Table definition and model definition.

PR #7288, opened 2026-03-24
Added RSA keypair pool + Caching to Seeder's RustSdk

theMickster:
## 🎟️ Tracking

[PM-33964](https://bitwarden.atlassian.net/browse/PM-33964)
[PM-33967](https://bitwarden.atlassian.net/browse/PM-33967)

## 📔 Objective

Replaces the single hardcoded RSA keypair in the Seeder's RustSdk with a pool of 100 pre-generated RSA-2048 keys, cached via `LazyLock` on first access. Users are assigned keypairs by index (`i % 100`), producing 100 distinct public keys across seeded organizations instead of one shared key.

Resolves a finding and supersedes #7279 — all seeded users sharing one RSA keypair masks key-mixing bugs that would fail in production (encrypting with UserA's public key instead of UserB's silently succeeds when both keys are identical). 

## 🧪 Testing

<details><summary>Expand for A/B results and SQL verification</summary>

### A/B: `qa.zero-knowledge-labs-enterprise` (386 users, 265 ciphers)

| Metric | Main | Feature | Match |
|---|---|---|---|
| OrgUsers | 386 | 386 | ✅ |
| Groups | 42 | 42 | ✅ |
| Collections | 53 | 53 | ✅ |
| Ciphers | 265 | 265 | ✅ |
| Distinct PublicKeys | **1** | **100** | ✅ Pool working |
| Wall time | 18.4s | 20.1s | ~comparable |

### A/B: 150-user org (no ciphers, isolates key generation)

| Metric | Main | Feature |
|---|---|---|
| Total Users | 143 | 143 |
| Distinct PublicKeys | **1** | **100** |

### Key distribution (150-user org)

With 150 users across 100 pool slots: 42 keys shared by 2 users, 1 key shared by 3 (owner + user0 + user100 all at index 0), 57 keys used by 1 user. Sum checks out.

### Pool wrapping verification

| User | Pool Index | PublicKey |
|---|---|---|
| owner | 0 | `...2bOrlQtyfqzydFag` |
| user0 | 0 | `...2bOrlQtyfqzydFag` |
| user100 | 0 (wraps) | `...2bOrlQtyfqzydFag` |
| user15 | 15 | `...nibfUTFk0jNAya8n` |
| user50 | 50 | `...sMbf2GY4sit2vHPV` |
| user99 | 99 | `...wFbeDLUwqpRgUh5P` |

### Spot-check: 7 users from `qa.zero-knowledge-labs-enterprise`

**Main** — all 7 share `...l0Vawl/toXzkEvB8` (single hardcoded key)

**Feature** — all 7 distinct:

| Email | PublicKey |
|---|---|
| aaliyah.brown | `...yVM0GHvXhJVWWIDL` |
| abigail.foster | `...0qX8CO7Xmjcin8Sm` |
| adaeze.nwachukwu | `...lbFaJBh0ew4t5Smi` |
| adebola.ogunlade | `...9VnrRPnDIvdFJpGQ` |
| aditya.kapoor | `...1KOVtSNr/Kjm/+WM` |
| adrian.stanescu | `...934G9cuHt0I1y+jJ` |
| adriana.popescu | `...yEh8nGb9Ou0Ktk8D` |

All users log in successfully with default password.

### Security review

- Zero-knowledge invariant preserved — only encrypted `EncString` crosses FFI, plaintext DER stays in Rust process
- `LazyLock<Vec<CachedRsaMaterial>>` is immutable after init, thread-safe, no interior mutability
- No key material in logs or error messages
- PEM constants are test-only material scoped to the Seeder crate

</details>

[PM-33964]: https://bitwarden.atlassian.net/browse/PM-33964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-33967]: https://bitwarden.atlassian.net/browse/PM-33967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/CreateOrganizationUserExtensions.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/InviteUsers/Models/CreateOrganizationUserExtensions.cs
@@ -9,22 +9,31 @@ public static class CreateOrganizationUserExtensions
 {
     public static CreateOrganizationUser MapToDataModel(this OrganizationUserInviteCommandModel organizationUserInvite,
         DateTimeOffset performedAt,
-        InviteOrganization organization) =>
-        new()
+        InviteOrganization organization)
+    {
+        var orgUser = new OrganizationUser
         {
-            OrganizationUser = new OrganizationUser
-            {
-                Id = CoreHelpers.GenerateComb(),
-                OrganizationId = organization.OrganizationId,
-                Email = organizationUserInvite.Email.ToLowerInvariant(),
-                Type = organizationUserInvite.Type,
-                Status = OrganizationUserStatusType.Invited,
-                AccessSecretsManager = organizationUserInvite.AccessSecretsManager,
-                ExternalId = string.IsNullOrWhiteSpace(organizationUserInvite.ExternalId) ? null : organizationUserInvite.ExternalId,
-                CreationDate = performedAt.UtcDateTime,
-                RevisionDate = performedAt.UtcDateTime
-            },
+            Id = CoreHelpers.GenerateComb(),
+            OrganizationId = organization.OrganizationId,
+            Email = organizationUserInvite.Email.ToLowerInvariant(),
+            Type = organizationUserInvite.Type,
+            Status = OrganizationUserStatusType.Invited,
+            AccessSecretsManager = organizationUserInvite.AccessSecretsManager,
+            ExternalId = string.IsNullOrWhiteSpace(organizationUserInvite.ExternalId) ? null : organizationUserInvite.ExternalId,
+            CreationDate = performedAt.UtcDateTime,
+            RevisionDate = performedAt.UtcDateTime
+        };
+
+        if (organizationUserInvite.Type == OrganizationUserType.Custom)
+        {
+            orgUser.SetPermissions(organizationUserInvite.Permissions);
+        }
+
+        return new CreateOrganizationUser
+        {
+            OrganizationUser = orgUser,
             Collections = organizationUserInvite.AssignedCollections,
             Groups = organizationUserInvite.Groups
         };
+    }
 }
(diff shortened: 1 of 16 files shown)


util/RustSdk/rust/src/cipher.rs
MGibson1, line 1: What's up with the formatting changes without content changes? Can you double check you're IDE is using the correct settings for rust formatting? It should be `cargo +"${RUST_NIGHTLY_TOOLCHAIN}" fmt --check`

util/Seeder/Factories/UserSeeder.cs
```diff
@@ -20,12 +20,13 @@ internal static (User user, UserKeys keys) Create(
         bool premium = false,
         UserKeys? keys = null,
         string? password = null,
-        int kdfIterations = 5_000)
+        int kdfIterations = 5_000,
+        uint poolIndex = 0)
```
MGibson1, line 24: **nit**: Seems like this is an unneeded complication to the UserSeeder interface. Every time you use this, you first generate keys so the poolIndex isn't needed.

Unless there is an impact on testing that requires this internal detail to be exposed, I think we're better off simplifying the userSeeder interface.

MGibson1 (approved):

util/RustSdk/rust/src/cipher.rs
theMickster: I think the problem is that cargo format did not run before on a prior code change and thus with this one when I did we got formatting change.

I'll double check in the morning.

util/Seeder/Factories/UserSeeder.cs
theMickster: I'll remove this in a follow-up.

PR #7294, opened 2026-03-24
Receive validation service

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33889

## 📔 Objective

Introduce a service to validate Receives upon creation


harr1424:
Please see https://github.com/bitwarden/server/pull/7306 which is being used to contain an updated Table definition and model definition.

PR #7299, opened 2026-03-24
[SM-1588] Small controller changes to include proper editor name

cd-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/jira/software/c/projects/SM/boards/74?assignee=625cb516fd06270069beaf5d&selectedIssue=SM-1588

## 📔 Objective

Return proper editor name so that we can display who edited the secrets value


src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -217,21 +216,24 @@ public async Task<SecretResponseModel> UpdateSecretAsync([FromRoute] Guid id, [F
                 {
                     editorOrganizationUserId = orgUser.Id;
                 }
-                else
+            }
+            else if (_currentContext.IdentityClientType == IdentityClientType.Organization)
```
jrmccannon, line 220: could we just combine this with the else if above? There's no difference with the code in each conditional block right?

src/Api/SecretsManager/Controllers/SecretsController.cs
```diff
@@ -217,21 +216,24 @@ public async Task<SecretResponseModel> UpdateSecretAsync([FromRoute] Guid id, [F
...
+            {
+                var orgUser = await _organizationUserRepository.GetByOrganizationAsync(secret.OrganizationId, userId);
+                if (orgUser != null)
                 {
-                    throw new NotFoundException();
```
jrmccannon, line 222: Is it intended to no longer throw the 404?

src/Api/SecretsManager/Controllers/SecretVersionsController.cs
```diff
@@ -51,7 +57,7 @@ public async Task<ListResponseModel<SecretVersionResponseModel>> GetVersionsBySe
         {
             // Already verified Secrets Manager access above
             var versionList = await _secretVersionRepository.GetManyBySecretIdAsync(secretId);
-            var responseList = versionList.Select(v => new SecretVersionResponseModel(v));
+            var responseList = await Task.WhenAll(versionList.Select(v => ResolveVersionWithEditorName(v, secret.OrganizationId)));
```
jrmccannon, line 60: This is looping over multiple calls to the database that will scale poorly the more versions are created of a secret. It will also re-query the same data in the event the same user/account has made multiple revisions.

Distinct lists of the service accounts and org ids, then used to populate the response model.

Since there's a bit of logic involved, it should probably be moved to a query class.

If there's no special SQL library being used for bulk retrieval (SqlBulk), then a single Query method is all that's needed and both the Single/Bulk requests can use that.

Something like:
```
public interface IGetSecretVersionsQuery
{
    Task<IEnumerable<SecretVersion>> GetVersionsBySecretAsync(Secret secret);
}

public class GetSecretVersionsQuery : IGetSecretVersionsQuery
{
    public Task<IEnumerable<SecretVersion>> GetVersionsBySecretAsync(Secret secret)
   {
        // Get versions
        // get distinct serviceaccountIds from versions
        // get distinct orguserId from versions
        // query each list to get name of user and service account
             // - SeviceAccountRepository#GetManyByIds
             // - Get list of users by list of org user ids - see below
   }[
}
```

### getting a list of user names by organization user ids

This one is tricky because there's no existing mechanism to retrieve this data. There are two viable options here
1. Call the existing OrgUserRepository#GetManyDetailsByOrganizationAsync_vNext(orgId) to retieve all org users and then you can pull the user name from that list on matching org user ids
 
#### Pros
- It already exists
- Gets the data needed

#### Cons
- Gets more than is needed (only need the name of the user based on the organization user id)

2. Create a new Sproc and repository method that is directly focused on what you need.

`OrganiationUserId[] -> (OrganizationUserId, Name)` : List of UserIds returns a list of UserId and their associated name. This could use the existing view (OrganizationUserUserDetailsView).

#### Pros
- only returning the data needed
- no extra calls

#### Cons
- It doesn't exist and you'd have to write sprocs and repo methods

Personally, I'd probably write the new sprocs and repo methods, but am open to the other option.

src/Api/SecretsManager/Controllers/SecretVersionsController.cs
```diff
@@ -334,4 +340,32 @@ public async Task<IActionResult> BulkDeleteAsync([FromBody] List<Guid> ids)
 
         return Ok();
     }
+
+    private async Task<SecretVersionResponseModel> ResolveVersionWithEditorName(Bit.Core.SecretsManager.Entities.SecretVersion secretVersion, Guid organizationId)
```
jrmccannon, line 344: This should be moved to the query class to build the data necessary for the responses

jrmccannon (changes requested):

cd-bitwarden:
Closing this PR as we are adding this functionality in this ticket https://bitwarden.atlassian.net/browse/SM-1911

PR #7300, opened 2026-03-24
[PM-33896] Update Families organization on schedule transition

cturnbull-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33896

## 📔 Objective

Adds logic to `SubscriptionUpdatedHandler` to detect schedule-triggered Families price changes and update the organization database at renewal time. When a subscription schedule transitions from Phase 1 → Phase 2, the handler detects the price change and updates the org's `PlanType`, `Plan`, `UsersGetPremium`, and `Seats` to the current `FamiliesAnnually` plan values.

### Changes

- **IStripeFacade → IStripeAdapter migration**: Fully migrates `SubscriptionUpdatedHandler` from `IStripeFacade` to `IStripeAdapter`, adding `DeleteSubscriptionDiscountAsync` and `GetTestClockAsync` to `IStripeAdapter`/`StripeAdapter` to enable the migration
- **Schedule-triggered detection**: After existing status-change checks in `HandleAsync`, detects schedule-triggered Families price migrations using: `ScheduleId` present + `PreviousAttributes` has item changes + new price matches `FamiliesAnnually` + metadata has `organizationId`
- **Org DB update**: On match, updates the organization with the current `FamiliesAnnually` plan values
- **Feature flag gated**: All new behavior behind `PM32645_DeferPriceMigrationToRenewal`
- **Tests**: 6 new test cases covering flag on/off, schedule present/absent, org/user subscriptions, price match/mismatch, and item change presence


src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -311,11 +325,64 @@ private async Task WaitForTestClockToAdvanceAsync(TestClock testClock)
...
-            testClock = await _stripeFacade.GetTestClock(testClock.Id);
+            testClock = await _stripeAdapter.GetTestClockAsync(testClock.Id);
             if (testClock.Status == "internal_failure")
             {
                 throw new Exception("Stripe Test Clock encountered an internal failure");
             }
         }
     }
+
+    private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
```
amorask-bitwarden, line 337: ❌ Any exception thrown in here would short-circuit the webhook handler since it runs before all the other operations. Can we wrap the whole thing in a try/catch with logging?

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -84,6 +93,11 @@ public async Task HandleAsync(Event parsedEvent)
             await RemovePendingCancellationAsync(subscription);
         }
 
+        if (_featureService.IsEnabled(FeatureFlagKeys.PM32645_DeferPriceMigrationToRenewal))
+        {
+            await HandleScheduleTriggeredFamiliesMigrationAsync(parsedEvent, subscription);
```
amorask-bitwarden: ⛏️ What do you think about moving this invocation to the `subscriberId.Match` for organization-case on line 103? That would preemptively ensure we're only running this on organization subscriptions. You'd also be able to pass in the Organization ID rather than parsing out the metadata again.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -311,11 +325,64 @@ private async Task WaitForTestClockToAdvanceAsync(TestClock testClock)
...
+    private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
+        Event parsedEvent,
+        Subscription subscription)
+    {
+        if (subscription.ScheduleId == null)
+        {
+            return;
+        }
+
+        if (parsedEvent.Data.PreviousAttributes?.items == null)
```
amorask-bitwarden: ❓ Do we need a stronger check here? We could receive a `customer.subscription.updated` webhook for a Families organization with a schedule for other cases besides a migration; a storage update comes to mind.

I think, ideally, we check if the previous attributes contained the 2019 price and that the current subscription contained the new price, which is exactly the case we're operating on.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
amorask-bitwarden: Discussed in Slack.

PR #7305, opened 2026-03-25
[PM-33899] Release schedule on terminal subscription operations

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33899

## 📔 Objective

Introduces `IPriceIncreaseScheduler` — a shared service that consolidates subscription schedule creation and release logic behind the `PM32645_DeferPriceMigrationToRenewal` feature flag. Previously, schedule creation lived as a private method in `UpcomingInvoiceHandler` and release guards were inlined at each call site.

**`Schedule(Subscription)`** — Creates a two-phase subscription schedule that defers a price increase to the renewal date. Resolves the subscriber type via `SubscriberId` metadata to determine the correct migration path (Premium or Families), builds the appropriate Phase 2 configuration (target price + discount), and creates the schedule in Stripe. Returns `bool` so callers can skip downstream processing (e.g., renewal emails) when a schedule already exists.

**`Release(string customerId, string subscriptionId)`** — Releases any active schedule for the subscription, cancelling a pending deferred price increase. Logs and re-throws on failure, requiring manual release via the Stripe Dashboard.

**Call sites:**
- `UpcomingInvoiceHandler` — calls `Schedule` during `invoice.upcoming` processing (replaces the private `SchedulePriceMigrationAsync`)
- `SubscriptionUpdatedHandler` — calls `Release` when a subscription goes unpaid; calls `Schedule` when a subscription recovers (recreates the schedule)
- `UpgradePremiumToOrganizationCommand` — calls `Release` before upgrading to an org plan
- `SetUpSponsorshipCommand` — calls `Release` before applying a sponsorship

### DIFF
--- a/src/Admin/AdminConsole/Views/Organizations/_ViewInformation.cshtml
+++ b/src/Admin/AdminConsole/Views/Organizations/_ViewInformation.cshtml
@@ -54,6 +54,9 @@
 
     <dt class="col-sm-4 col-lg-3">Limit collection deletion to administrators</dt>
     <dd id="pm-collection-deletion" class="col-sm-8 col-lg-9">@(Model.Organization.LimitCollectionDeletion ? "On" : "Off")</dd>
+
+    <dt class="col-sm-4 col-lg-3">Limit item deletion to Manage collection permission</dt>
+    <dd id="pm-item-deletion" class="col-sm-8 col-lg-9">@(Model.Organization.LimitItemDeletion ? "On" : "Off")</dd>
 </dl>
 
 <h2>Secrets Manager</h2>
--- a/src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
+++ b/src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
@@ -1,8 +1,12 @@
-﻿using Bit.Core.AdminConsole.OrganizationFeatures.Organizations.Interfaces;
+﻿using Bit.Core;
+using Bit.Core.AdminConsole.OrganizationFeatures.Organizations.Interfaces;
 using Bit.Core.AdminConsole.Repositories;
 using Bit.Core.AdminConsole.Services;
+using Bit.Core.Billing;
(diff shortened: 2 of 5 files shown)


src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
```diff
@@ -0,0 +1,280 @@
...
+            _ => ResolvePhase2ForFamiliesAsync(subscription),
+            _ =>
+            {
+                logger.LogWarning(
+                    "Provider subscriptions do not support price increase scheduling ({SubscriptionId})",
+                    subscription.Id);
+                return Task.FromResult<SubscriptionSchedulePhaseOptions?>(null);
+            });
+        }
+        catch (Exception ex)
```
connerbw, line 171: ❓ Should we extend this to rethrow `BillingException` as we did [here](https://github.com/bitwarden/server/blob/billing/PM-33899/release-schedule-terminal/src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs#L411-L418) to retry in the event the pricing service fails? Up to you - that service doesn't go down often, so I can see being overly defensive as not necessary.

amorask-bitwarden: @cturnbull-bitwarden My concern here is that, since I put this in a shared service, we have more call sites than just a single handler, and each call site has different responsibilities. If I re-throw in the service, I'll burn each call site down without actually knowing the exact reason the exception was thrown.

A resolution here would be either:
1. Adding many more exception types that call sites could switch off of to determine what actually happened.
2. Adding some sort of `Reason` string or enum to the `BillingException` that call sites could switch off of to determine what happened.

I think both are probably outside scope here, but let me know what you think.

connerbw: Nah, I think you're right, probably not worth all that effort. I'll resolve this

PR #7306, opened 2026-03-25
[PM-34030] Add receive file storage service

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34030

## 📔 Objective

Roughly implements the existing `AzureSendFileStorage` but for Receives

### DIFF
--- a/src/Core/Tools/Models/Data/ReceiveFileData.cs
+++ b/src/Core/Tools/Models/Data/ReceiveFileData.cs
@@ -60,5 +60,6 @@ public class ReceiveFileData
     /// the expected tolerance and below the maximum supported
     /// file size.
     /// </summary>
-    public bool Validated { get; set; } = false;  // equivalent Send prop defaults to true, let's see what happens
+    public bool Validated { get; set; } = false;  // equivalent Send prop defaults to true
+                                                  // TODO: test both true and false default
 }


src/Sql/dbo/Tools/Tables/Receive.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE [dbo].[Receive]
+(
+    [Id]             UNIQUEIDENTIFIER NOT NULL,
+    [UserId]         UNIQUEIDENTIFIER NOT NULL,
+    [Data]           VARCHAR(MAX)     NOT NULL,
+    [Key]            VARCHAR(MAX)     NOT NULL,
```
mkincaid-bw: ❓ The data model has three key related properties, but the table only has one generic `[Key]` column. Is this intentional?

src/Sql/dbo/Tools/Tables/Receive.sql
```diff
@@ -0,0 +1,23 @@
...
+    [Secret]         NVARCHAR(300)    NOT NULL,
+    [UploadCount]    INT              NOT NULL,
+    [CreationDate]   DATETIME2(7)     NOT NULL,
+    [RevisionDate]   DATETIME2(7)     NOT NULL,
+    [ExpirationDate] DATETIME2(7)     NULL,
+    CONSTRAINT [PK_Receive] PRIMARY KEY CLUSTERED ([Id] ASC),
+    CONSTRAINT [FK_Receive_User] FOREIGN KEY ([UserId]) REFERENCES [dbo].[User] ([Id]),
```
mkincaid-bw: The trailing comma should not be on this line.

src/Sql/dbo/Tools/Tables/Receive.sql
```diff
@@ -0,0 +1,23 @@
...
+);
+
+GO
+CREATE NONCLUSTERED INDEX [IX_Receive_ExpirationDate]
+    ON [dbo].[Receive] ([ExpirationDate] ASC);
```
mkincaid-bw, line 20: ❓Are most rows expected to have a null expiration date? If so, it might make sense to exclude them in the index
```sql
CREATE NONCLUSTERED INDEX [IX_Receive_ExpirationDate]
    ON [dbo].[Receive] ([ExpirationDate] ASC)
    WHERE [ExpirationDate] IS NOT NULL;
```

src/Core/Tools/SendFeatures/Services/AzureReceiveStorageService.cs
```diff
@@ -0,0 +1,117 @@
...
+            await blobClient.SetMetadataAsync(metadata);
+
+            var headers = new BlobHttpHeaders
+            {
+                ContentDisposition = $"attachment; filename=\"{fileId}\""
+            };
+            await blobClient.SetHttpHeadersAsync(headers);
+
+            var length = blobProperties.Value.ContentLength;
+            var valid = minimum <= length || length <= maximum;
```
mkincaid-bw: My Claude reviewer flagged his as a logic bug, says it should be:
```csharp
var valid = minimum <= length && length <= maximum;
```
Not sure if it's correct but I wanted to point it out.

src/Sql/dbo/Tools/Tables/Receive.sql
mkincaid-bw: The EF migrations specify a delete cascade on this foreign key, but the SQL migration does not. If a cascaded delete is desired, then it should be added to the SQAL definition. If not, it should be removed from the EF migrations so that they are consistent.

src/Core/Tools/Models/Data/ReceiveFileData.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Should contain a non-empty string once the file upload completes.
+    /// </value>
+    public string FileName { get; set; } = string.Empty;
+
+    /// <summary>
+    /// When true the uploaded file's length was confirmed within
+    /// the expected tolerance and below the maximum supported
+    /// file size.
+    /// </summary>
+    public bool Validated { get; set; } = false;  // equivalent Send prop defaults to true, let's see what happens
```
mkincaid-bw: I'm confused by this comment, and why it's intentionally different from `Send`, with "let's see what happens".

src/Core/Tools/SendFeatures/Services/AzureReceiveStorageService.cs
```diff
@@ -0,0 +1,117 @@
...
+    public FileUploadType FileUploadType => FileUploadType.Azure;
+
+    public static string ReceiveIdFromBlobName(string blobName) => blobName.Split('/')[0];
+    public static string BlobName(Receive receive, string fileId) => $"{receive.Id}/{fileId}";
+
+    public AzureReceiveFileStorageService(
+        GlobalSettings globalSettings,
+        ILogger<AzureReceiveFileStorageService> logger)
+    {
+        _blobServiceClient = new BlobServiceClient(globalSettings.Send.ConnectionString);
```
mkincaid-bw, line 29: My Claude reviewer flagged this as well, questioning whether or not Receive should have it's own connection string. Not sure if it's intentional, just wanted to point it out.

src/Sql/dbo/Tools/Tables/Receive.sql
harr1424: Every row will receive a value for this column, at most 30 days in the future from the date the Receive was created. We may want to discuss making this column not-nullable. I am not 100% certain but I think the equivalent column in the `Send` table was nullable because rows had both `ExpirationDate` and `DeletionDate`.

src/Core/Tools/SendFeatures/Services/AzureReceiveStorageService.cs
harr1424: This is a legitimate bug that currently exists in the equivalent `Send` file. We are planning to fix this, and I have updated the relevant code in this PR for the receive functionality.

src/Sql/dbo/Tools/Tables/Receive.sql
harr1424: I've opted to remove the cascaded delete from the EF migrations.

src/Core/Tools/Models/Data/ReceiveFileData.cs
harr1424: It seemed unusual to set the result of a validation to true. In C# bools default to false. 

This means the default behavior has been intentionally changed so that if `Validated` isn't explicitly assigned, it will default to `true` meaning validation will pass, I don't yet understand why this would be desirable. 

As we implement this functionality, I wanted to set a breakpoint and explore what impact, if any, this change might have. As this is part of the innovation sprint, we don't yet have enough code implemented to test this out. I've modified the code  comment to read more professionally and less vague (re "let's see what happens") .

src/Core/Tools/SendFeatures/Services/AzureReceiveStorageService.cs
harr1424: Receives should ultimately have a dedicated storage location, but as we work through this proof of concept, I don't think it is worth making the change now, as it will likely require Azure configuration.  During development, I think it is acceptable to share a storage location with Sends.

src/Sql/dbo/Tools/Tables/Receive.sql
harr1424: Apologies, I should have marked this PR as draft last night after updating the Entity. Thank you for the thorough review.

harr1424:
I am leaving this PR open in case @mkincaid-bw wants to review changes made in response to review comments, but PR https://github.com/bitwarden/server/pull/7312 should be considered for review by innovation team devs since it introduces additional functionality needed for getting a controller class implemented.

harr1424:
Closing this due to conflicts, DbOps will review any final solution as part of opening a PR against `main`

PR #7317, opened 2026-03-26
[INNO] - Passkey Directory Endpoint

jrmccannon:
## Associated PRs
- Clients : https://github.com/bitwarden/clients/pull/19787

## 📔 Objective
This PR adds a new Passkey Directory Report endpoint (GET /reports/passkey-directory) that fetches and returns a list of domains supporting passkeys from the https://passkeys-api.2fa.directory/v1/all.json. The query layer (GetPasskeyDirectoryQuery) retrieves the external directory data, parses each entry for passwordless login support, MFA support, and documentation links, and caches results for 24 hours using FusionCache. The endpoint is gated behind the PasskeyDirectoryReport feature flag. This data powers a client-side report that helps organization administrators identify which of their members' credentials could be upgraded to passkeys. Unit tests cover the controller endpoint and the query's parsing/filtering logic.


src/Api/Dirt/Controllers/ReportsController.cs
```diff
@@ -206,4 +204,22 @@ public async Task DropPasswordHealthReportApplication(
 
         await _dropPwdHealthReportAppCommand.DropPasswordHealthReportApplicationAsync(request);
     }
+
+    /// <summary>
+    /// Gets the list of domains that support passkeys from the 2FA Directory
+    /// </summary>
+    /// <returns>List of domains with passkey support details</returns>
+    [HttpGet("passkey-directory")]
```
lastbestdev, line 212: Not critical: instead of placing the endpoint under `/reports/`, I'm wondering if we want to consolidate endpoints that call third party APIs elsewhere. They could share a controller with a base route like `/external` or similar.

Alternatively, we could follow what was done for HIBP and just introduce a new controller for 2FA directory API proxy requests, like `/2fa-directory/`.

It is worth noting that the Inactive 2FA report in the web client has direct calls to the 2FA directory API still, which should also be refactored to make requests through our API like this.

src/Core/Dirt/Reports/ReportFeatures/PasskeyDirectoryReport/README.md
lastbestdev, line 1: Not critical: Where should this README live/is it appropriate to include here?

src/Core/Dirt/Reports/ReportFeatures/GetPasskeyDirectoryQuery.cs
```diff
@@ -0,0 +1,86 @@
...
+    IFusionCache cache,
+    ILogger<GetPasskeyDirectoryQuery> logger)
+    : IGetPasskeyDirectoryQuery
+{
+    public const string HttpClientName = "PasskeyDirectoryHttpClient";
+    public const string CacheName = "PasskeyDirectory";
+
+    private static readonly TimeSpan _cacheDuration = TimeSpan.FromDays(1);
+    private const string _cacheKey = "passkey-directory";
+    private const string _passkeyDirectoryUrl = "https://passkeys-api.2fa.directory/v1/all.json";
```
lastbestdev: Not critical: should this be provided as an environment variable to avoid having to deploy this service to fix a change to the API URL?

lastbestdev (dismissed):
LGTM, no blocking comments

lastbestdev (approved):

Banrion (approved):

PR #7318, opened 2026-03-26
[PM-30483] Remove feature flagged logic around passkey unlock

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30483

## 📔 Objective

Now that passkey unlock has been released, this change, paired with [clients PR#19798](https://github.com/bitwarden/clients/pull/19798) removes the logic related to the passkey unlock feature flag `pm-2035-passkey-unlock`. The feature flag itself will be removed in a future ticket: https://bitwarden.atlassian.net/browse/PM-30484


src/Api/Vault/Controllers/SyncController.cs
```diff
@@ -125,9 +125,7 @@ await _providerUserRepository.GetManyOrganizationDetailsByUserAsync(user.Id,
         var organizationIdsClaimingActiveUser = organizationClaimingActiveUser.Select(o => o.Id);
 
         var organizationAbilities = await GetOrganizationAbilitiesAsync(ciphers);
-        var webAuthnCredentials = _featureService.IsEnabled(FeatureFlagKeys.PM2035PasskeyUnlock)
```
gbubemismith, line 129: ❓ do we also need to remove `PM2035PasskeyUnlock` from the constants.cs

eligrubb: I believe we need to leave the flag in `constants.cs`. The server still needs to report the feature flag for 3 releases after we remove it from clients to maintain backwards compatibility; the flagged logic is part of the extension client. There is a separate ticket, [PM-30484](https://bitwarden.atlassian.net/browse/PM-30484), for removing the flag completely.

Thomas-Avery (approved):

nikwithak (approved):

PR #7332, opened 2026-03-27
[PM-31671] Add SSRF protections to ChangePasswordUriSErvice

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31671

## 📔 Objective

This prevents the service that retrieves the change password URL from exposing internal details, by verifying that the domain does not resolve to an internal IP address. It uses the same approach as the Icons service in following redirects to ensure they are not internal-only services.

Note: When testing locally, all URLs are passing through `127.0.0.1` due to a local proxy - I had to add `UseProxy=false` to the `HttpClientHandler` constructor on `SErviceCollectionExtension.cs:48`. I suspect this is something configured in my localhost dev environment, since the existing IconsService uses this same approach without issue.

### DIFF
--- a/src/Icons/Models/IconUri.cs
+++ b/src/Icons/Models/IconUri.cs
@@ -39,7 +39,7 @@ public class IconUri
     }
 
     /// <summary>
-    /// Represents an ip-validated Uri for use in grabbing an icon.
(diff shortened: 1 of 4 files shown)


shane-melton (approved):
Looks good!

PR #7359, opened 2026-03-30
[PM-30185] Fix email fallback logic to ignore empty primary email

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30185

## 📔 Objective
Changes email fallback logic in SCIM invite to ignore the blank primary email if a valid secondary email is provided


jrmccannon (changes requested):
Could we add a test with the "bad" request and validate that it doesn't throw an error and email is populated?

jrmccannon (approved):

eliykat (approved):

PR #7372, opened 2026-04-01
[PM-34462] Improve role handling in provider controllers

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34462

## 📔 Objective

Refactor provider controllers to improve role handling and make sure we have test coverage.


src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
eliykat, line 1: Changes in this file were required to get Sqlite working. It doesn't support the `LinqToDb.Tools` `In` and `NotIn`. This is never run on EF in prod (it's cloud only), but is required for integration tests.

jrmccannon (approved):

JaredScar (approved):

PR #7393, opened 2026-04-06
[PM-34679] Display Phase 2 prices and discount on org subscription page

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34679

## 📔 Objective

During Phase 1 of a subscription schedule (the 15-day window before Families renewal), the Admin Console's Billing > Subscription page was showing the old plan price (2019 or 2025) and no discount. Stripe had the correct Phase 2 data, but `StripePaymentService.GetSubscriptionAsync()` never consulted the subscription schedule.

This PR adds schedule-aware logic to `GetSubscriptionAsync()` that overlays Phase 2 prices and discounts onto the `SubscriptionInfo` response when an active subscription schedule exists. This follows the same pattern already used for Premium individual subscriptions in `GetBitwardenSubscriptionQuery.GetSchedulePhase2DiscountAsync()`.

**Changes:**
- Added `ApplySchedulePhase2DataAsync()` to `StripePaymentService` — fetches the subscription schedule, matches Phase 2 items by product ID, and overrides line item amounts and discount
- Added `BillingCustomerDiscount(Coupon)` constructor to `SubscriptionInfo` for schedule phase discounts (which expose a Coupon, not a full Discount wrapper)
- 6 unit tests covering: happy path with discount, no Phase 2 discount, no schedule, schedule fetch failure, inactive schedule, and Phase 2 already started

No client-side changes needed — the existing response model fields already carry the correct data once populated with Phase 2 values.

## 📸 Screenshots

[Families subscription in phase 2](https://dashboard.stripe.com/acct_19smIXIGBnsLynRr/test/subscriptions/sub_1TJG7rIGBnsLynRr5FjWc37P)

image

### DIFF
--- a/src/Core/Billing/Services/Implementations/StripePaymentService.cs
+++ b/src/Core/Billing/Services/Implementations/StripePaymentService.cs
@@ -661,6 +661,8 @@ public class StripePaymentService : IStripePaymentService
             subscriptionInfo.CustomerDiscount = new SubscriptionInfo.BillingCustomerDiscount(discount);
         }
 
+        await ApplySchedulePhase2DataAsync(subscription, subscriptionInfo);
+
         var (suspensionDate, unpaidPeriodEndDate) = await GetSuspensionDateAsync(subscription);
 
         if (suspensionDate.HasValue && unpaidPeriodEndDate.HasValue)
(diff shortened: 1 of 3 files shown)


src/Core/Billing/Services/Implementations/StripePaymentService.cs
```diff
@@ -702,6 +704,74 @@ public async Task<SubscriptionInfo> GetSubscriptionAsync(ISubscriber subscriber)
...
+                {
+                    if (phase2Item.Price is not { UnitAmount: not null, ProductId: not null })
+                    {
+                        continue;
+                    }
+
+                    var matchingItem = items.FirstOrDefault(i => i.ProductId == phase2Item.Price.ProductId);
+                    if (matchingItem != null)
+                    {
+                        matchingItem.Amount = phase2Item.Price.UnitAmount.Value / 100M;
```
kdenney, line 749: ❓ Are subscription items' amounts represented in decimal with dollars and cents, while the phase's items' amounts are integers representing cents? I was surprised to see the divide by 100 here so I wanted to double check that this was intentional.

kdenney (approved):
Looks good, just have one question to double check.

src/Core/Billing/Services/Implementations/StripePaymentService.cs
amorask-bitwarden: Yup, Bitwarden's output models have always tended towards providing currency values as 32.50 rather than 3250 to the client.

PR #7408, opened 2026-04-07
[PM-34679] Fix Families 2019 Phase 2 price and discount display

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34679

## 📔 Objective

During the 15-day pre-renewal window (Phase 1), the Billing > Subscription page was showing the old Families 2019 price ($12.00) instead of the upcoming Phase 2 price ($47.88), and the one-time renewal discount strikethrough was missing. PR #7393 fixed this for Families 2025 but not 2019.

**Root cause:** `ApplySchedulePhase2DataAsync` matched Phase 2 schedule items to Phase 1 subscription items by Stripe product ID. Families 2025 and the current plan share the same Stripe product, so the match worked. Families 2019 uses a different, older Stripe product — the match failed and the Phase 2 price was never applied. Additionally, the schedule's coupon `applies_to` data was not being expanded from the Stripe API, so the client couldn't render the discount strikethrough.

**Fix:**
- Added a two-pass matching strategy: Pass 1 matches by product ID (existing behavior); Pass 2 falls back to pairing unmatched Phase 2 items with unmatched non-addon Phase 1 items, overriding `Amount`, `ProductId`, and `Name`
- Expanded `phases.discounts.coupon.applies_to` on the schedule fetch so the discount's product targeting is available to the client for strikethrough rendering

## 📸 Screenshots

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -34,6 +34,22 @@
             "stopAll": true
         },
         {
+            "name": "Admin, API, Identity, Billing",
+            "configurations": [
+                "run-Admin",
+                "run-API",
+                "run-Identity",
+                "run-Billing"
+            ],
+            "presentation": {
+                "hidden": false,
+                "group": "AA_compounds",
+                "order": 4
+            },
+            "preLaunchTask": "buildIdentityApiAdminBilling",
+            "stopAll": true
+        },
+        {
             "name": "API, Identity, SSO",
             "configurations": [
                 "run-API",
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -33,6 +33,22 @@
             "preLaunchTask": "buildIdentityApiAdmin",
             "stopAll": true
         },
+        {
```
sbrown-livefront, line 36: ❓ Did you want to add these?

sbrown-livefront (approved):
Looks good. Just a question

.vscode/launch.json
amorask-bitwarden: @sbrown-livefront Yup! This file's checked in and I've started using VS Code over Rider cuz the Claude plugin is better so I need a launch config for Billing. Thanks for checking though.

PR #7414, opened 2026-04-07
[PM-26043] Fix bug: can't add secrets manager to legacy plans

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26043

## 📔 Objective

Moves `AddSecretsManagerSubscriptionCommand` from OrganizationFeatures/OrganizationSubscriptions/ into billing (Billing/Organizations/Commands/) and rewrites it to use `IUpdateOrganizationSubscriptionCommand` instead of `IStripePaymentService.AddSecretsManagerToSubscription`.

The primary motivation is a bug fix: the old command delegated validation to `OrganizationService.ValidateSecretsManagerPlan`, which internally called `ValidatePlan`. That method threw a `BadRequestException` when `plan.Disabled == true`, blocking existing subscribers on legacy/retired plans from ever adding Secrets Manager to their subscription. The new command intentionally omits the disabled-plan check so that a subscriber's ability to enable Secret's Manager does not depend on whether their plan is still on sale.

#### Bug fix
  - Removes the `plan.Disabled` check that prevented existing subscribers on legacy plans (e.g. `enterprise-annually-2020`) from adding Secrets Manager

#### Refactor
  - Deletes `IStripePaymentService.AddSecretsManagerToSubscription` and its implementation in `StripePaymentService`
  - Deletes the now-unused `SecretsManagerSubscribeUpdate` model
  - Deletes the old command and its interface from OrganizationFeatures/OrganizationSubscriptions/
  - `AddSecretsManagerSubscriptionCommand` moved into billing and all validation logic from `OrganizationService` moved inline into the command


src/Core/Billing/Organizations/Commands/AddSecretsManagerSubscriptionCommand.cs
```diff
@@ -0,0 +1,144 @@
...
+    private static void ValidateOrganization(Organization organization, StaticStorePlan plan, Provider? provider)
+    {
+        if (organization.UseSecretsManager)
+        {
+            throw new BadRequestException("Organization already uses Secrets Manager.");
+        }
+
+        if (!plan.SupportsSecretsManager)
+        {
+            throw new BadRequestException("Invalid Secrets Manager plan selected.");
```
amorask-bitwarden: ⛏️ "Organization's plan does not support Secrets Manager."

src/Core/Billing/Organizations/Commands/AddSecretsManagerSubscriptionCommand.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(organization.GatewayCustomerId))
+            {
+                throw new BadRequestException("No payment method found.");
+            }
+
+            if (string.IsNullOrWhiteSpace(organization.GatewaySubscriptionId))
+            {
+                throw new BadRequestException("No subscription found.");
+            }
```
amorask-bitwarden, line 85: ⛏️ I'd consider these conflicts rather than bad requests. The end user has no control over whether we set these correctly on the backend.

src/Core/Billing/Organizations/Commands/AddSecretsManagerSubscriptionCommand.cs
```diff
@@ -0,0 +1,144 @@
...
+
+    private static void ValidateSecretsManagerPlan(
+        StaticStorePlan plan,
+        int additionalSmSeats,
+        int additionalServiceAccounts,
+        Organization organization)
+    {
+        if (additionalSmSeats < 0)
+        {
+            throw new BadRequestException("You can't subtract Secrets Manager seats!");
```
amorask-bitwarden: ⛏️ Wording is confusing, IMO. They're not subtracting seats from a current SM plan since the command is to add SM. I'd say they're trying to add SM with negative seats.

src/Core/Billing/Organizations/Commands/AddSecretsManagerSubscriptionCommand.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        if (plan.SecretsManager.BaseSeats + additionalSmSeats <= 0)
```
amorask-bitwarden: ❓ I was always confused by this conditional. If the check on 104 passes, is this even possible? The base seats for the free plan is 2. Maybe I'm just getting mixed up by the validation handling free and paid.

src/Core/Billing/Organizations/Commands/AddSecretsManagerSubscriptionCommand.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        if (!plan.SecretsManager.HasAdditionalServiceAccountOption && additionalServiceAccounts > 0)
+        {
+            throw new BadRequestException("Plan does not allow additional Machine Accounts.");
+        }
+
+        if ((plan.ProductTier == ProductTierType.TeamsStarter &&
+             additionalSmSeats > plan.PasswordManager.BaseSeats) ||
+            (plan.ProductTier != ProductTierType.TeamsStarter &&
+             additionalSmSeats > organization.Seats.GetValueOrDefault()))
```
amorask-bitwarden: ⛏️ I think a Teams Starter organization would still have their `organization.Seats` value set to 10. I.e., I don't think the || is necessary. I think you can just check the organization's seat count.

src/Core/Billing/Organizations/Commands/AddSecretsManagerSubscriptionCommand.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            throw new BadRequestException("You cannot have more Secrets Manager seats than Password Manager seats.");
+        }
+
+        if (additionalServiceAccounts < 0)
+        {
+            throw new BadRequestException("You can't subtract Machine Accounts!");
```
amorask-bitwarden: ⛏️ Same as here: https://github.com/bitwarden/server/pull/7414/changes#r3052122133

amorask-bitwarden (dismissed):
Looks good - just a few things I was wondering if we want to clean up from old validation while we're here that I left ⛏️ and ❓ on.

Only other thing I can think of is that the implementation might be clearer if the validation of a free SM plan and a paid SM plan were split.

kdenney:
@amorask-bitwarden All feedback addressed in https://github.com/bitwarden/server/pull/7414/commits/f015195e663d326310ba4632281c3f7d1e832791.

After verifying some assumptions and discussing with you in slack, I was able to simplify the validation a bit and specifically check if the plan is a free plan to replace the `BaseSeats + additionalSmSeats` condition. Hopefully it's a bit easier to understand now.

amorask-bitwarden (dismissed):

amorask-bitwarden (approved):

PR #7417, opened 2026-04-08
[PM-31884] Add access control support to Send Controls policy

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31884

## 📔 Objective

This PR introduces the server logic necessary to support new fields in the Send Controls policy to enable specifying the required access control and - for email access control - the required domains for all Sends created under the policy. When the policy is updated any existing Send that is non-compliant will be marked as Disabled, thus removing the ability to access it but allowing the owner to still see it in the clients and recover the Send's information. If the policy is changed or removed such that existing Sends are no longer non-compliant the Disabled state will be removed.

## 📸 Screenshots

Not applicable


src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +70,55 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+        var orgUserIds = orgUsers.Where(w => w.UserId != null).Select(s => s.UserId!.Value).ToList();
+        var domains = (sendControlsPolicyData.AllowedDomains ?? "").Split(",").Select(d => d.Trim());
+        var enabled = new List<Guid>();
+        var disabled = new List<Guid>();
+        foreach (var userId in orgUserIds)
+        {
+            var userSends = await sendRepository.GetManyByUserIdAsync(userId);
+            foreach (var userSend in userSends)
+            {
+                Console.WriteLine(userSend);
```
harr1424: ❓  Is this left over from local testing?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendControlsPolicyRequirement.cs
```diff
@@ -35,6 +45,8 @@ public override SendControlsPolicyRequirement Create(IEnumerable<PolicyDetails>
                 {
                     DisableSend = result.DisableSend || data.DisableSend,
                     DisableHideEmail = result.DisableHideEmail || data.DisableHideEmail,
+                    WhoCanAccess = result.WhoCanAccess ?? data.WhoCanAccess,
+                    AllowedDomains = result.AllowedDomains ?? data.AllowedDomains
```
harr1424, line 49: 🤔 In practice, the use of the `??` operator will short circuit policy evaluation across multiple orgs.  When multiple orgs are being evaluated for their policy statuses, as soon as one `WhoCanAccess` or `AllowedDomains` is encountered as non-null the rest will be discarded. 

I wouldn't consider this a blocker unless product has defined exactly how they would like to handle this edge case, where a user is affected by more than one access control policy. At one point, they mentioned defining a hierarchy. It's also worth considering how the same situation should be handled with respect to the domains.

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
```diff
@@ -0,0 +1,14 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    -- Set field
+    UPDATE [dbo].[Send]
+    SET
+      [Disabled] = @Disabled
+    WHERE
+        [Id] IN (SELECT * FROM @Ids)
```
harr1424: I think DbOps will request this be de-dented one level

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
```diff
@@ -0,0 +1,21 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    -- Set field
+    UPDATE [dbo].[Send]
+    SET
+      [Disabled] = @Disabled
+    WHERE
+        [Id] IN (SELECT * FROM @Ids)
```
harr1424: Same as above, I suspect this should be de-dented for consistent formatting

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -68,6 +69,24 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
...
+        {
+            var requiredAccessControl = passwordRequired ? "password" : emailsRequired ? "email verification" : "(cannot determine required auth)";
+            throw new BadRequestException($"Due to an Enterpise Policy your Sends must be protected by {requiredAccessControl}");
+        }
+
+        if (emailsRequired && sendControlsRequirement.AllowedDomains != null)
+        {
+            var domains = sendControlsRequirement.AllowedDomains.Split(",").Select(domain => domain.Trim());
+            var emails = send.Emails.Split(",").Select(email => email.Trim());
+            if (emails.Any(email => !domains.Any(domain => email.EndsWith(domain))))
```
harr1424: 🤔 This should probably have some thorough test coverage.

❓ How will `string.EndsWith()` behave if `bitwarden.com` is an allowed domain, but `impersonated-bitwarden.com` is used on a given Send's email OTP field? 

Suggested fix from Claude that seems like a reasonable helper: 
```csharp
static bool EmailMatchesDomain(string email, string domain)
{
    var emailDomain = email.Split('@').LastOrDefault() ?? "";
    return emailDomain.Equals(domain, StringComparison.OrdinalIgnoreCase)
        || emailDomain.EndsWith("." + domain, StringComparison.OrdinalIgnoreCase);
}
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +70,55 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+                // If the policy is no longer in effect then re-enable all Sends
+                if (!postUpsertedPolicyState.Enabled)
+                {
+                    enabled.Add(userSend.Id);
+                } else if (
+                    sendControlsPolicyData.DisableSend ||
+                    (sendControlsPolicyData.DisableHideEmail && (userSend.HideEmail ?? false)) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.PasswordProtected && userSend.AuthType != AuthType.Password) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople && userSend.AuthType != AuthType.Email) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople && domains.Any() && (userSend.Emails ?? "").Split(",").Select(e => e.Trim()).Any(e => !domains.Any(d => e.EndsWith(d)))))
```
harr1424: ⚠️ This is another place that the domain could be bypassed, see [other comment](https://github.com/bitwarden/server/pull/7417/changes#r3052211578)

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
harr1424, line 1: ⚠️ This SP should call `User_BumpAccountRevisionDate` following the convention of other Send SPs:

https://github.com/bitwarden/server/blob/6d43cc43e36087cf2a6fcfae8ab325f47667e23e/src/Sql/dbo/Tools/Stored%20Procedures/Send_Update.sql#L49

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
```diff
@@ -94,4 +94,19 @@ public UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
...
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+        var sends = await GetDbSet(dbContext)
+            .Where(s => ids.Contains(s.Id))
+            .ToListAsync();
+        foreach (var send in sends)
+        {
+            send.Disabled = disabled;
+        }
+
```
harr1424, line 217: ⚠️ I think the RevisionDate should be bumped here: 

https://github.com/bitwarden/server/blob/fae8692d2ab4f2b39ef8a5cbee0801b71b04b4a6/src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs#L44

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +70,55 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+                }
+            }
+        }
+        if (enabled.Count > 0) {
+            await sendRepository.UpdateManyDisabledAsync(enabled, false);
+        }
+        if (disabled.Count > 0)
+        {
+            await sendRepository.UpdateManyDisabledAsync(disabled, true);
+        }
```
harr1424, line 150: ❓ I've added comments about calling `User_BumpAccountRevisionDate` int he repository methods, but do we also want to update the `Send.RevisionDate` on the Send objects themselves when they are modified by policies? 

I included logic to do so for the former policies, it may be a good convention to continue to follow.

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
harr1424: This may take some thought, since the Send owner ID needs to be used for the bump....

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
harr1424: This may take some thought, since the Send owner ID needs to be used for the bump....

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -48,13 +49,13 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
         }
 
         // Once data migration has run, query only SendControls
-        // var sendControlsTask = _policyRequirementQuery.GetAsync<SendControlsPolicyRequirement>(userId.Value);
+        var sendControlsTask = _policyRequirementQuery.GetAsync<SendControlsPolicyRequirement>(userId.Value);
         var disableSendTask = _policyRequirementQuery.GetAsync<DisableSendPolicyRequirement>(userId.Value);
         var sendOptionsTask = _policyRequirementQuery.GetAsync<SendOptionsPolicyRequirement>(userId.Value);
 
         await Task.WhenAll(disableSendTask, sendOptionsTask);
```
harr1424:
```suggestion
        await Task.WhenAll(sendControlsTask, disableSendTask, sendOptionsTask);
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +70,55 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+    {
+        var orgUsers = await organizationUserRepository.GetManyByOrganizationAsync(postUpsertedPolicyState.OrganizationId, null);
+        var orgUserIds = orgUsers.Where(w => w.UserId != null).Select(s => s.UserId!.Value).ToList();
+        var domains = (sendControlsPolicyData.AllowedDomains ?? "").Split(",").Select(d => d.Trim());
+        var enabled = new List<Guid>();
+        var disabled = new List<Guid>();
+        foreach (var userId in orgUserIds)
+        {
+            var userSends = await sendRepository.GetManyByUserIdAsync(userId);
+            foreach (var userSend in userSends)
```
harr1424: ~~Nested for loops are always a concern. The inner loop doesn't seem particularly worrisome, but the outer loop is time bound on the `sendRepository.GetManyByUserIdAsync()` for each user in *all of the orgs*.~~

~~Creating a dedicated method in `SendRepository` to get every `UserId` based upon an input of a list of `OrganizationId` could be really helpful in terms of performance.~~

My comment above was not accurate.

harr1424 (changes requested):
⚠️ Two blockers: 

1. Using `string.EndsWith()` to compare emails against allowed domains allows `impersonated-bitwarden.com` to pass validation enforcing the domain `bitwarden.com`
2. `User_BumpAccountRevisionDate` is not called in the SP or EF methods

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
harr1424: Line 13 specifically

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
harr1424: Line 19

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
```diff
@@ -0,0 +1,14 @@
+CREATE PROCEDURE [dbo].[Send_SetDisabledByIds]
+    @Ids AS [dbo].[GuidIdArray] READONLY,
+    @Disabled BIT
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    -- Set field
+    UPDATE [dbo].[Send]
```
mkincaid-bw: ⛏️ `UPDATE` should be on it's own line.
https://contributing.bitwarden.com/contributing/code-style/sql#update-statements

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
```diff
@@ -0,0 +1,21 @@
+IF OBJECT_ID('[dbo].[Send_SetDisabledByIds]') IS NOT NULL
```
mkincaid-bw: `CREATE OR ALTER` should be used instead of `DROP/CREATE`.
https://contributing.bitwarden.com/contributing/code-style/sql#creating-or-modifying-a-function-or-stored-procedure

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
```diff
@@ -0,0 +1,21 @@
...
+
+CREATE PROCEDURE [dbo].[Send_SetDisabledByIds]
+    @Ids AS [dbo].[GuidIdArray] READONLY,
+    @Disabled BIT
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    -- Set field
+    UPDATE [dbo].[Send]
```
mkincaid-bw: ⛏️ `UPDATE` should be on it's own line.
https://contributing.bitwarden.com/contributing/code-style/sql#update-statements

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: Yep, removed

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyRequirements/SendControlsPolicyRequirement.cs
mcamirault: I believe the plan is to have this policy be dependent on having the Single Org policy enabled first, so the multi-org case should not be a problem.

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
mcamirault: Done

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
mcamirault: Done

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
mcamirault: Good points all, I've added more test coverage and changed the logic for detecting domains to be more robust

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: See above

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
mcamirault: The account revision date is now being bumped

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
mcamirault: See comment on other SQL file

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: Send revision dates should now be getting bumped as well as the user account revision date

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
mcamirault: This may explain some of the difficulties I had testing this :D Good catch

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
mcamirault: Done

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
mcamirault: Done

util/Migrator/DbScripts/2026-04-07_00_SendSetDisabledByIds.sql
mcamirault: Done

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -145,6 +145,19 @@ INNER JOIN
...
+    public async Task UpdateManyDisabledAsync(IEnumerable<Guid> ids, bool disabled, IEnumerable<Guid> userIds)
+    {
+        using var connection = new SqlConnection(ConnectionString);
+        await connection.ExecuteAsync(
+            $"[{Schema}].[Send_SetDisabledByIds]",
+            new { Ids = ids.ToGuidIdArrayTVP(), Disabled = disabled },
+            commandType: CommandType.StoredProcedure);
+        await connection.ExecuteAsync(
+            $"[{Schema}].[User_BumpManyAccountRevisionDates]",
+            new { Ids = ids.ToGuidIdArrayTVP() },
```
harr1424, line 223: ❓ `The first parameter `ids` is passed to both `Send_SetDisabledByIds` and `User_BumpManyAccountRevisionDates`. Could this be an error, did you mean to pass Sends IDs to `Send_SetDisabledByIds` and user IDs to `User_BumpManyAccountRevisionDates`? 

I think you only need to pass `userIds` on line 157 instead: 

```suggestion
            new { userIds = ids.ToGuidIdArrayTVP() },
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +71,70 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+        var disabledSendUserIds = new List<Guid>();
+        foreach (var userId in orgUserIds)
+        {
+            var userSends = await sendRepository.GetManyByUserIdAsync(userId);
+            var userHadSendsEnabled = false;
+            var userHadSendsDisabled = false;
+            foreach (var userSend in userSends)
+            {
+                // If the policy is no longer in effect then re-enable all Sends
+                if (!postUpsertedPolicyState.Enabled)
```
harr1424: ❓ Will this re-enable Sends that the user deliberately disabled themselves?  I am curious how the overall approach distinguishes between Sends that a user explicitly disabled, and those implicitly disabled by policy.

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEventTests.cs
```diff
@@ -135,4 +140,99 @@ await sutProvider.GetDependency<IPolicyRepository>()
...
+        existingSendOptionsPolicy.OrganizationId = policyUpdate.OrganizationId;
+        postUpsertedPolicy.SetDataModel(new SendControlsPolicyData { DisableHideEmail = true, WhoCanAccess = SendWhoCanAccessType.SpecificPeople, AllowedDomains = "duckdodgers.com" });
+
+        sutProvider.GetDependency<IPolicyRepository>()
+            .GetByOrganizationIdTypeAsync(policyUpdate.OrganizationId, PolicyType.DisableSend)
+            .Returns(existingDisableSendPolicy);
+        sutProvider.GetDependency<IPolicyRepository>()
+            .GetByOrganizationIdTypeAsync(policyUpdate.OrganizationId, PolicyType.SendOptions)
+            .Returns(existingSendOptionsPolicy);
+        var orgUserId = new Guid();
```
harr1424: I don't think this will behave [as expected](https://codapi.org/embed/?sandbox=dotnet&code=data%3A%3Bbase64%2Cdc0xCgIxEIXhOnOKKZPGCwQrCxEUBIutYxLiQDaRTLIiy95d1MXKvPJ9xd%2BYUsDLk6sfNYj34N6ukSzaaJjxXHIoZoQZxPpzNZUsTpkcngwlqUDMIMRkCuYSDg63mPwD942cVHqVxr50yKRcb391lxPn6DdDoeqPlLz8BDr2TXTwF1EaQCywvAA%3D). 


```suggestion
        var orgUserId = Guid.NewGuid();
```

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEventTests.cs
```diff
@@ -135,4 +140,99 @@ await sutProvider.GetDependency<IPolicyRepository>()
...
+        var orgUser = new OrganizationUser
+        {
+            UserId = orgUserId
+        };
+        sutProvider.GetDependency<IOrganizationUserRepository>()
+            .GetManyByOrganizationAsync(postUpsertedPolicy.OrganizationId, null)
+            .Returns([ orgUser ]);
+        var nonCompliantSend1 = new Send
+        {
+            Id = new Guid(),
```
harr1424:
```suggestion
            Id =Guid.NewGuid(),
```

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEventTests.cs
```diff
@@ -135,4 +140,99 @@ await sutProvider.GetDependency<IPolicyRepository>()
...
+            HideEmail = true
+        };
+        var nonCompliantSend2 = new Send
+        {
+            Id = new Guid(),
```
harr1424:
```suggestion
            Id = Guid.NewGuid(),
```

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -68,8 +70,33 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
             throw new BadRequestException(
                 "Due to an Enterprise Policy, you are not allowed to hide your email address from recipients when creating or editing a Send.");
         }
+
+        var passwordRequired = sendControlsRequirement.WhoCanAccess == SendWhoCanAccessType.PasswordProtected;
+        var emailsRequired = sendControlsRequirement.WhoCanAccess == SendWhoCanAccessType.SpecificPeople;
+        if ((passwordRequired && send.Password == null) || (emailsRequired && send.Emails == null))
+        {
+            var requiredAccessControl = passwordRequired ? "password" : emailsRequired ? "email verification" : "(cannot determine required auth)";
+            throw new BadRequestException($"Due to an Enterpise Policy your Sends must be protected by {requiredAccessControl}");
```
harr1424: "Due to an enterprise policy...." 

Enterpise (missing r) > enterprise (I don't think it should be capitalized but could be wrong)

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +71,70 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+    {
+        var dataModel = policyRequest.PolicyUpdate.GetDataModel<SendControlsPolicyData>();
+        if (dataModel.AllowedDomains is not null && dataModel.WhoCanAccess != SendWhoCanAccessType.SpecificPeople)
+        {
+            return Task.FromResult("Allowed domains can only be set when the required access type is set to specific people");
+        }
+        return Task.FromResult(string.Empty);
+    }
+
+    public async Task SetDisabledForSendsByPolicyAsync(Policy postUpsertedPolicyState, SendControlsPolicyData sendControlsPolicyData)
```
harr1424: This could be private 

```suggestion
    private async Task SetDisabledForSendsByPolicyAsync(Policy postUpsertedPolicyState, SendControlsPolicyData sendControlsPolicyData)
```

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +71,70 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+            }
+        }
+        if (enabled.Count > 0) {
+            await sendRepository.UpdateManyDisabledAsync(enabled, false, enabledSendUserIds);
+        }
+        if (disabled.Count > 0)
+        {
+            await sendRepository.UpdateManyDisabledAsync(disabled, true, disabledSendUserIds);
+        }
+        return;
```
harr1424: I don't think this return is reachable

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +71,70 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+                {
+                    enabled.Add(userSend.Id);
+                    userHadSendsEnabled = true;
+
+                } else if (
+                    sendControlsPolicyData.DisableSend ||
+                    (sendControlsPolicyData.DisableHideEmail && (userSend.HideEmail ?? false)) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.PasswordProtected && userSend.AuthType != AuthType.Password) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople && userSend.AuthType != AuthType.Email) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople && domains.Any() && (userSend.Emails ?? "").Split(",").Select(e => e.Trim()).Any(e => !domains.Any(d => SendValidationService.SendEmailMatchesDomain(e, d)))))
```
harr1424:
```suggestion
                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople                                                                                           
      && sendControlsPolicyData.AllowedDomains is not null
      && (userSend.Emails ?? "").Split(",").Select(e => e.Trim())                                                                                                       
          .Any(e => !domains.Any(d => SendValidationService.SendEmailMatchesDomain(e, d))))
```
If the admin restricted access using "Specific People," and the domains list was was null or empty then none of the Sends would pass the domain cross-check, even though there are no allowed domains and (I think) they should ALL pass. 

This comment may conflict with requirements established by product that I am not aware of.

harr1424 (changes requested):
Blockers: 

The Dapper implementation of `ISendRepository` has a method `UpdateManyDisabledAsync()` that should be updated such that `ids` representing Send IDs are passed to `Send_SetDisabledByIds` and `userIds` representing User IDs are passed to `User_BumpManyAccountRevisionDates`

`SendControlsSyncPolicyEventTests` are using `new Guid()` which [does not create a unique GUID](https://codapi.org/embed/?sandbox=dotnet&code=data%3A%3Bbase64%2Cdc0xCgIxEIXhOnOKKZPGCwQrCxEUBIutYxLiQDaRTLIiy95d1MXKvPJ9xd%2BYUsDLk6sfNYj34N6ukSzaaJjxXHIoZoQZxPpzNZUsTpkcngwlqUDMIMRkCuYSDg63mPwD942cVHqVxr50yKRcb391lxPn6DdDoeqPlLz8BDr2TXTwF1EaQCywvAA%3D) and therefore `nonCompliantSend1` and `nonCompliantSend2` are indistinguishable. 

Enterprise has been mis-spelled as enterpise, and likely shouldn't be capitalized. @sukhleenb should "Enterprise Policy" be capitalized as such? 

`SendWhoCanAccessType.SpecificPeople && domains.Any() ` results in a set up where the policy is enforced, but no allowed domains are set, meaning all Sends applied by the policy will be disabled. I think if domains is empty, the Sends should not be disabled, but this is a good one to check with product on. 

How does the Sync Event distinguish between Sends disabled explicitly by the user from Sends disabled implicitly by the policy change?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: We've already removed the disabled field from our clients' editors, so currently the only way to explicitly disable a Send is via direct API call. We can probably lock this down even further, as I believe the end goal is to eliminate the explicit "Disabled" state (we should discuss this with Product though)

harr1424: Sounds good, thanks for clarifying!

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
mcamirault: Ah this was a silly copy paste error on my part. Fixed now

test/Core.Test/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEventTests.cs
mcamirault: Fixed

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
mcamirault: Fixed the typo, the capitalization was matching the other error messages in this service but I'll wait to hear from @sukhleenb if they should all be capitalized or not

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: I don't see why not, it's outside any of the `for` and `if` blocks. Can you clarify?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: I adjusted the `domains` variable to make sure it's an empty list when the `AllowedDomains` field is null or empty. The `domains.Any()` should then take care of this scenario, yes?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
harr1424: Sorry, rather than unreachable, I think redundant would be the correct term. I only caught this because my IDE (Rider) grays it out, something I am used to seeing for unused variables, dead code, or in this case a "Redundant control flow jump statement." 

Not a blocker but more of a nit, the function will return at this point regardless.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
harr1424:
```csharp
namespace Demo;

class Program
{
    static void Main(string[] args)
    {
        var domains = (null ?? "").Split(",").Select(d => d.Trim());
        var domainsAny = domains.Any();
        Console.WriteLine(domainsAny);  // True
    }
}
```

This means that any emails will fail the validation, even if `domains` is empty. Maybe this is desired? If the SpecificPeople requirement is enabled, and no domains are listed, the Sends should all be disabled? This is the current behavior as I understand it. 

If so, I feel like the `domains.Any()` is kind of misleading, to me this reads "if SpecificPeople policy is enabled and `domains` has values, then cross-check the emails against the values." 

This could just be me, though. Were you using `.Any()` to ensure it wasn't null? That case was already handled by the null coalescing to empty string.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: Ah thanks I see what you mean now. Interesting that VS Code didn't catch this, it's gotten much better with C# than the last time I tried it but I guess there are still some gaps.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: Ah I was going under the opposite assumption, that since the domain restriction is an optional add-on to requiring email verification then not specifying `domains` meant that any domain was acceptable. But thinking about it again blocking all domains unless explicitly specified is the same pattern as blocking all emails unless explicitly specified (which is what we do now for email verification) so it makes more sense. I've consolidated the handling of the input `emails` and `domains` strings into a function and modified its logic accordingly.

test/Core.Test/Tools/Services/SendValidationServiceTests.cs
```diff
@@ -175,8 +179,77 @@ public async Task ValidateUserCanSaveAsync_WhenPoliciesDoNotApply_Success(
...
+            .Returns(new DisableSendPolicyRequirement { DisableSend = false });
+
+        sutProvider.GetDependency<IPolicyRequirementQuery>().GetAsync<SendOptionsPolicyRequirement>(userId)
+            .Returns(new SendOptionsPolicyRequirement { DisableHideEmail = false });
+        
+        sutProvider.GetDependency<IPolicyRequirementQuery>().GetAsync<SendControlsPolicyRequirement>(userId)
+            .Returns(new SendControlsPolicyRequirement { DisableSend = false, DisableHideEmail = false, WhoCanAccess = SendWhoCanAccessType.SpecificPeople });
+
+        var exception = await Assert.ThrowsAsync<BadRequestException>(() => sutProvider.Sut.ValidateUserCanSaveAsync(userId, send));
+        Assert.Equal("Due to an Enterpise Policy your Sends must be protected by email verification", exception.Message);
```
harr1424:
```suggestion
        Assert.Equal("Due to an Enterprise Policy your Sends must be protected by email verification", exception.Message);
```

test/Core.Test/Tools/Services/SendValidationServiceTests.cs
```diff
@@ -175,8 +179,77 @@ public async Task ValidateUserCanSaveAsync_WhenPoliciesDoNotApply_Success(
...
+            .Returns(new DisableSendPolicyRequirement { DisableSend = false });
+
+        sutProvider.GetDependency<IPolicyRequirementQuery>().GetAsync<SendOptionsPolicyRequirement>(userId)
+            .Returns(new SendOptionsPolicyRequirement { DisableHideEmail = false });
+        
+        sutProvider.GetDependency<IPolicyRequirementQuery>().GetAsync<SendControlsPolicyRequirement>(userId)
+            .Returns(new SendControlsPolicyRequirement { DisableSend = false, DisableHideEmail = false, WhoCanAccess = SendWhoCanAccessType.PasswordProtected });
+
+        var exception = await Assert.ThrowsAsync<BadRequestException>(() => sutProvider.Sut.ValidateUserCanSaveAsync(userId, send));
+        Assert.Equal("Due to an Enterpise Policy your Sends must be protected by password", exception.Message);
```
harr1424:
```suggestion
        Assert.Equal("Due to an Enterprise Policy your Sends must be protected by password", exception.Message);
```

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -145,6 +145,19 @@ INNER JOIN
...
+    public async Task UpdateManyDisabledAsync(IEnumerable<Guid> ids, bool disabled, IEnumerable<Guid> userIds)
+    {
+        using var connection = new SqlConnection(ConnectionString);
+        await connection.ExecuteAsync(
+            $"[{Schema}].[Send_SetDisabledByIds]",
+            new { Ids = ids.ToGuidIdArrayTVP(), Disabled = disabled },
+            commandType: CommandType.StoredProcedure);
+        await connection.ExecuteAsync(
+            $"[{Schema}].[User_BumpManyAccountRevisionDates]",
+            new { UserIds = userIds.ToGuidIdArrayTVP() },
```
harr1424:
```suggestion
            new { Ids = userIds.ToGuidIdArrayTVP() },
```

I believe this needs to match the SP parameter: 
```
CREATE PROCEDURE [dbo].[User_BumpManyAccountRevisionDates]
    @Ids [dbo].[GuidIdArray] READONLY
```

harr1424 (changes requested):
Almost there! I spotted a few more things that need changing:

test/Core.Test/Tools/Services/SendValidationServiceTests.cs
mcamirault: Fixed

test/Core.Test/Tools/Services/SendValidationServiceTests.cs
mcamirault: Fixed

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
mcamirault: I've updated the approach so that we don't have to make two sproc calls

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
```diff
@@ -1,4 +1,4 @@
-CREATE OR ALTER PROCEDURE [dbo].[Send_SetDisabledByIds]
+CREATE PROCEDURE [dbo].[Send_SetDisabledByIds]
```
harr1424: This should likely remain `CREATE OR ALTER`, does the migration script reflect the other changes made here?

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -63,4 +71,68 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+    {
+        var orgUsers = await organizationUserRepository.GetManyByOrganizationAsync(postUpsertedPolicyState.OrganizationId, null);
+        var orgUserIds = orgUsers.Where(w => w.UserId != null).Select(s => s.UserId!.Value).ToList();
+        var enabled = new List<Guid>();
+        var enabledSendUserIds = new List<Guid>();
+        var disabled = new List<Guid>();
+        var disabledSendUserIds = new List<Guid>();
+        foreach (var userId in orgUserIds)
+        {
+            var userSends = await sendRepository.GetManyByUserIdAsync(userId);
```
BTreston: I'm concerned with running this query in a loop for ALL org users. This may be better defined in its own sproc.

BTreston (changes requested):

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: My worry was about loading all of the full Send objects for all org users into memory at once, hence the loop. Would it be better to just get all of the Send IDs first and then process them in batches of 50 or 100? There would still be a loop but the strain on the database would be more evenly distributed

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
mcamirault: I believe this is correct, every other sproc I can find uses `CREATE PROCEDURE` in the `.sql` file but `CREATE OR ALTER PROCEDURE` (or the older `DROP/CREATE` pattern) in the migration script

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
BTreston: That sounds reasonable, I'd have @mkincaid-bw weigh in on it though for a final say

src/Sql/dbo/Tools/Stored Procedures/Send_SetDisabledByIds.sql
harr1424: OK, thanks for looking into the precedent.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: Had an offline conversation with @mkincaid-bw about this and he agreed this approach is fine

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
```diff
@@ -94,4 +94,35 @@ public UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
...
+        await dbContext.UserBumpManyAccountRevisionDatesAsync(userIds);
+        await dbContext.SaveChangesAsync();
+    }
+
+    public async Task<IEnumerable<Guid>> GetIdsByOrganizationIdAsync(Guid organizationId)
+    {
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+        var orgUsers = await dbContext.OrganizationUsers.Where(ou => ou.OrganizationId == organizationId).ToListAsync();
+        var orgUserSendIds = await dbContext.Sends.Where(s => orgUsers.Any(ou => ou.Id == s.UserId)).Select(s => s.Id).ToListAsync();
```
harr1424: ❌ **CRITICAL**: Wrong join key — compares `OrganizationUser.Id` to `Send.UserId`, which will never match.

<details>
<summary>Details and fix</summary>

`ou.Id` is the `OrganizationUser` association row's primary key, while `s.UserId` is the `User` foreign key. These are different GUIDs — a Send belonging to a member of the org will not satisfy `ou.Id == s.UserId`, so this method will return an empty list on EF-backed databases (PostgreSQL, MySQL, SQLite).

The corresponding stored procedure `Send_ReadIdsByOrgId` uses `OrganizationUser.UserId`, not `OrganizationUser.Id`. The EF implementation must do the same.

As a result, `SendControlsSyncPolicyEvent.UpdateSendsByPolicyAsync` is a no-op on non-MSSQL backends: no Sends will be disabled or re-enabled when the SendControls policy changes.

Suggested fix — project UserIds server-side and use `Contains` instead of the in-memory `.Any()`:

```csharp
public async Task<IEnumerable<Guid>> GetIdsByOrganizationIdAsync(Guid organizationId)
{
    using var scope = ServiceScopeFactory.CreateScope();
    var userIds = dbContext.OrganizationUsers
        .Where(ou => ou.OrganizationId == organizationId && ou.UserId.HasValue)
        .Select(ou => ou.UserId!.Value);
    return await dbContext.Sends
        .Where(s => s.UserId.HasValue && userIds.Contains(s.UserId.Value))
        .Select(s => s.Id)
        .ToListAsync();
}
```

Please add dual-ORM parity tests covering policy sync so this kind of mismatch is caught going forward.
</details>

util/Migrator/DbScripts/2026-06-15_00_SendAccessControlPolicyProcs.sql
```diff
@@ -0,0 +1,58 @@
...
+
+    -- Set field
+    UPDATE
+      [dbo].[Send]
+    SET
+      [Disabled] = @Disabled,
+      [RevisionDate] = GETUTCDATE()
+    WHERE
+      [Id] IN (SELECT * FROM @Ids)
+END
```
harr1424, line 31: ❌ **CRITICAL**: Migration drifts from source SP — `User_BumpManyAccountRevisionDates` is missing here but present in `src/Sql`.

<details>
<summary>Details and fix</summary>

The source-of-truth procedure at `src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql` (applied on fresh deployments) calls `User_BumpManyAccountRevisionDates` after the UPDATE. This migration script — applied to every existing production / self-host database — omits that call.

The result is that existing deployments will silently fail to bump user account revision dates when Sends are bulk-enabled/disabled by policy changes, so clients won't know to resync their Send data until some unrelated action triggers a revision bump. Fresh deployments will behave correctly, making the bug environment-dependent and hard to diagnose.

Align the migration with `src/Sql`:

```sql
CREATE OR ALTER PROCEDURE [dbo].[Send_UpdateDisabledByIds]
    @Ids AS [dbo].[GuidIdArray] READONLY,
    @Disabled BIT
AS
BEGIN
    SET NOCOUNT ON

    UPDATE
        [dbo].[Send]
    SET
        [Disabled] = @Disabled,
        [RevisionDate] = GETUTCDATE()
    WHERE
        [Id] IN (SELECT * FROM @Ids)

    EXEC [dbo].[User_BumpManyAccountRevisionDates]
    (
        SELECT DISTINCT
            UserId
        FROM
            [dbo].[Send]
        WHERE
            [Id] IN (SELECT * FROM @Ids)
    )
END
GO
```

(The EF equivalent — `UserBumpManyAccountRevisionDatesAsync` — is already wired up in `SendRepository.UpdateManyDisabledAsync`, so only MSSQL is affected by the drift.)
</details>

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -60,4 +66,49 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+            foreach (var send in sendsChunk)
+            {
+                if (
+                    // If the policy is disabled then we want to re-enable any Sends that were previously disabled
+                    postUpsertedPolicyState.Enabled && 
+                    (sendControlsPolicyData.DisableSend ||
+                    (sendControlsPolicyData.DisableHideEmail && (send.HideEmail ?? false)) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.PasswordProtected && send.AuthType != AuthType.Password) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople && send.AuthType != AuthType.Email) ||
+                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople && !SendValidationService.SendAllEmailsHaveAllowedDomains(send.Emails, sendControlsPolicyData.AllowedDomains))))
```
harr1424:
```suggestion
                    (sendControlsPolicyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople
                    && sendControlsPolicyData.AllowedDomains != null
                    && !SendValidationService.SendAllEmailsHaveAllowedDomains(send.Emails, sendControlsPolicyData.AllowedDomains))))
```

❓ without the suggested change, does the sync logic exactly match the validation logic?

I think the current logic in `SendValidationService` will allow create and edit operations on Sends when the allowed domain list is null, but that the sync logic will disable every existing Send when the allowed domain list is null or empty. 

The suggested change applies the same null guard used in `SendValidationService` to the sync logic.

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
```diff
@@ -60,4 +66,49 @@ private async Task UpsertLegacyPolicyAsync<T>(
...
+                {
```
harr1424, line 95: Claude review suggested that a malformed email could cause a silent abort of the sync logic, what do you think? Are you confident that existing email validation would prevent this kind of situation? 

If not, here is the recommendation. I am undecided on this, since I recall implementing email validation logic on the clients.

```suggestion
  //   - Policy disabled  → fall through to `else` → re-enable every Send (restores anything this policy had previously disabled).                                                                                                                                     
  //   - Policy enabled   → check compliance; non-compliant Sends get disabled, compliant ones get (re-)enabled.                                                                                                                                                       
  if (postUpsertedPolicyState.Enabled && IsNonCompliant(send, sendControlsPolicyData)) 
   ```

With the addition of a helper method: 

```csharp
private static bool IsNonCompliant(Send send, SendControlsPolicyData policyData)
  {
      if (policyData.DisableSend)                                                                                                                                                                                                                                      
      {
          return true;                                                                                                                                                                                                                                                 
      }                
      if (policyData.DisableHideEmail && (send.HideEmail ?? false))                                                                                                                                                                                                    
      {
          return true;                                                                                                                                                                                                                                                 
      }                
      if (policyData.WhoCanAccess == SendWhoCanAccessType.PasswordProtected
          && send.AuthType != AuthType.Password)                                                                                                                                                                                                                       
      {
          return true;                                                                                                                                                                                                                                                 
      }                
      if (policyData.WhoCanAccess == SendWhoCanAccessType.SpecificPeople)
      {                                                                                                                                                                                                                                                                
          if (send.AuthType != AuthType.Email)
          {                                                                                                                                                                                                                                                            
              return true;
          }
          try
          {
              if (!SendValidationService.SendAllEmailsHaveAllowedDomains(send.Emails, policyData.AllowedDomains))
              {                                                                                                                                                                                                                                                        
                  return true;
              }                                                                                                                                                                                                                                                        
          }            
          catch (BadRequestException)
          {                                                                                                                                                                                                                                                            
              // Historical Send rows may contain a malformed email that MailAddress rejects.
              // We can't verify such a Send against the allowed-domains list, so treat it as                                                                                                                                                                          
              // non-compliant and disable it rather than aborting the org-wide sweep.
              return true;                                                                                                                                                                                                                                             
          }
      }                                                                                                                                                                                                                                                                
      return false;    
  }
```

harr1424 (changes requested):
Blocker: Policy sync logic should match how Sends are validated in `SendValidationService`, see suggested change to incorporate a null guard in the sync logic.  Otherwise the divergence will lead to unexpected behavior when admins enable/disable the policy. 

Optional: Consider implications of a malformed email address silently aborting the Sync flow enabling/disabling Sends.

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
mcamirault: Fixed

util/Migrator/DbScripts/2026-06-15_00_SendAccessControlPolicyProcs.sql
mcamirault: Fixed

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: Good catch here! I've added this check to the suggested helper function below

src/Core/AdminConsole/OrganizationFeatures/Policies/PolicyEventHandlers/SendControlsSyncPolicyEvent.cs
mcamirault: I know that our clients have email validation but I think it's almost always worth making the server more defensive just in case. I've added this helper function and the null check you suggested above.

harr1424 (dismissed):

src/Sql/dbo/Tools/Stored Procedures/Send_ReadIdsByOrgId.sql
```diff
@@ -0,0 +1,24 @@
+CREATE PROCEDURE [dbo].[Send_ReadIdsByOrgId]
```
mkincaid-bw: This should be named `Send_ReadIdsByOrganizationId`

src/Sql/dbo/Tools/Stored Procedures/Send_ReadIdsByOrgId.sql
```diff
@@ -0,0 +1,24 @@
...
+    @Id UNIQUEIDENTIFIER
```
mkincaid-bw: The base object here is `Send`, so the input parameter name implies a `SendId`. Since you are querying by `OrganizationId` , the parameter name should also be `@OrganizationId`.

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
```diff
@@ -0,0 +1,30 @@
+CREATE PROCEDURE [dbo].[Send_UpdateDisabledByIds]
+    @Ids AS [dbo].[GuidIdArray] READONLY,
+    @Disabled BIT
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @UserIds [dbo].[GuidIdarray]
```
mkincaid-bw: ⛏️ `GuidIdarray` should be `GuidIdArray`

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
```diff
@@ -0,0 +1,30 @@
...
+
+    -- Set field
+    UPDATE
+        [dbo].[Send]
+    SET
+        [Disabled] = @Disabled,
+        [RevisionDate] = GETUTCDATE()
```
mkincaid-bw: The value for `RevisionDate` should be passed in as a parameter, not set using `GETUTCDATE`.

mkincaid-bw (changes requested):
Some minor changes requested.

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
mcamirault: Fixed

src/Sql/dbo/Tools/Stored Procedures/Send_UpdateDisabledByIds.sql
mcamirault: Fixed

src/Sql/dbo/Tools/Stored Procedures/Send_ReadIdsByOrgId.sql
mcamirault: Fixed

src/Sql/dbo/Tools/Stored Procedures/Send_ReadIdsByOrgId.sql
mcamirault: Fixed

util/Migrator/DbScripts/2026-05-04_00_SendAccessControlPolicyProcs.sql
```diff
@@ -0,0 +1,72 @@
...
+    SET NOCOUNT ON
+
+    DECLARE @UserIds [dbo].[GuidIdArray]
+
+    -- Set field
+    UPDATE
+        [dbo].[Send]
+    SET
+        [Disabled] = @Disabled,
+        [RevisionDate] = GETUTCDATE()
```
mkincaid-bw: This doesn't match the code in `Send_UpdateDisabledByIds.sql` (needs `@RevisionDate` parameter)

util/Migrator/DbScripts/2026-05-04_00_SendAccessControlPolicyProcs.sql
```diff
@@ -0,0 +1,72 @@
...
+        *
+    FROM
+        [dbo].[SendView]
+    WHERE
+        [Id] IN (SELECT * FROM @Ids)
+END
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[Send_ReadIdsByOrganizationId]
+    @Id UNIQUEIDENTIFIER
```
mkincaid-bw: This doesn't match the code in `Send_ReadIdsByOrganizationId.sql` (should be `@OrganizationId`)

util/Migrator/DbScripts/2026-05-04_00_SendAccessControlPolicyProcs.sql
```diff
@@ -0,0 +1,72 @@
...
+
+    -- Get the IDs of all users in an org --
+    DECLARE @OrgUserIds AS [GuidIdArray];
+    INSERT INTO @OrgUserIds
+    SELECT DISTINCT
+        UserId
+    FROM
+        [dbo].[OrganizationUserView]
+    WHERE
+        OrganizationId = @Id
```
mkincaid-bw:
```suggestion
        OrganizationId = @OrganizationId 
```

mkincaid-bw (changes requested):
Couple more small things.

util/Migrator/DbScripts/2026-05-04_00_SendAccessControlPolicyProcs.sql
mcamirault: Fixed

util/Migrator/DbScripts/2026-05-04_00_SendAccessControlPolicyProcs.sql
mcamirault: Fixed

util/Migrator/DbScripts/2026-05-04_00_SendAccessControlPolicyProcs.sql
mcamirault: Fixed

mkincaid-bw (dismissed):
LGTM

harr1424 (dismissed):

BTreston (dismissed):

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -68,14 +68,47 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
...
+            throw new BadRequestException($"Due to an Enterprise Policy your Sends must be protected by {requiredAccessControl}");
+        }
+
+        if (emailsRequired && sendControlsRequirement.AllowedDomains != null)
+        {
+            if (!SendAllEmailsHaveAllowedDomains(send.Emails, sendControlsRequirement.AllowedDomains))
+            {
+                throw new BadRequestException($"Due to an Enterprise Policy your Sends must be protected by email verification and access granted only to the following domain(s): {sendControlsRequirement.AllowedDomains}");
+            }
+        }
```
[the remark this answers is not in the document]
mcamirault: Addressed

mkincaid-bw (approved):

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -89,14 +89,44 @@ public async Task ValidateUserCanSaveAsync(Guid? userId, Send send)
             throw new BadRequestException(
                 "Due to an Enterprise Policy, you are not allowed to hide your email address from recipients when creating or editing a Send.");
         }
+
+        var passwordRequired = sendControlsRequirement.WhoCanAccess == SendWhoCanAccessType.PasswordProtected;
+        var emailsRequired = sendControlsRequirement.WhoCanAccess == SendWhoCanAccessType.SpecificPeople;
+        if ((passwordRequired && send.Password == null) || (emailsRequired && send.Emails == null))
```
harr1424, line 95:
```suggestion
        if ((passwordRequired && string. IsNullOrWhiteSpace(send.Password)) || (emailsRequired && string. IsNullOrWhiteSpace(send.Emails)))
```

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -196,6 +196,35 @@ INNER JOIN
         };
     }
 
+    public async Task UpdateManyDisabledAsync(IEnumerable<Guid> ids, bool disabled)
+    {
+        using var connection = new SqlConnection(ConnectionString);
+        await connection.ExecuteAsync(
+            $"[{Schema}].[Send_UpdateDisabledByIds]",
+            new { Ids = ids.ToGuidIdArrayTVP(), Disabled = disabled, RevisionDate = DateTime.UtcNow },
```
harr1424, line 204: ❓ What does TVP represent here? The helper class doesn't document this function.

harr1424 (approved):

BTreston (approved):

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
mcamirault: Table-valued parameter, it's the standard way of passing arrays of data to our sprocs

PR #7420, opened 2026-04-08
[PM-34773] Fix storage addition during active Phase 2 of schedule

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34773

## 📔 Objective

Fixes a 500 error when adding or removing storage for Premium users or Families organizations whose subscription schedule has transitioned to Phase 2 (the migrated price is active).

The schedule-aware storage commands (introduced in PM-33898) reconstructed both schedule phases when updating storage. After Phase 2 activates, Phase 1's dates are in the past and Stripe rejects updates to ended phases. This PR:

- Skips Phase 1 when its `EndDate` has passed, sending only the active phase(s) to Stripe
- Strips the one-time migration discount from Phase 2 when it is already active, preventing Stripe from re-applying a consumed coupon
- Guards against degenerate schedule states where no updatable phases remain
- Uses `subscription.TestClock.FrozenTime` for accurate phase-ended detection in test clock environments

## 📸 Screenshots

https://github.com/user-attachments/assets/56a3d9b0-c2f8-45e2-8cf2-de564b9cd478

### DIFF
--- a/src/Core/Billing/Organizations/Commands/UpdateOrganizationSubscriptionCommand.cs
+++ b/src/Core/Billing/Organizations/Commands/UpdateOrganizationSubscriptionCommand.cs
@@ -128,15 +128,20 @@ public class UpdateOrganizationSubscriptionCommand(
                 CommandName, activeSchedule.Id, subscription.Id);
 
             var phase1 = activeSchedule.Phases[0];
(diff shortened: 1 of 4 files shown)


sbrown-livefront (approved):
✅

cyprain-okeke (approved):

PR #7425, opened 2026-04-08
[PM-34213] Create attachment event log

shane-melton:
## 🎟️ Tracking

[PM-34213](https://bitwarden.atlassian.net/browse/PM-34213)

## 📔 Objective

Ensure `Cipher_AttachmentCreated` events are logged when an attachment is created/uploaded.

[PM-34213]: https://bitwarden.atlassian.net/browse/PM-34213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gbubemismith (dismissed):

shane-melton:
@gbubemismith Had to move the event logging to when the attachment metadata is created and not when validated asynchronously. The validation step occurs via the Azure webhook which has no current user context for the event log.

gbubemismith (approved):

PR #7426, opened 2026-04-09
[PM-34823] Remove missed uses of PolicyRequirements flag

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34823

## 📔 Objective

We have removed the `PolicyRequirements` feature flag for all uses except for the `RequireSso` policy. One other use crept back in in the v2 `AdminRecoverAccountCommand`, and there were other references not cleaned up in some Tools tests (already removed from the sut). This PR removes those erroneous uses in advance of us turning on the feature flag next regression cycle.


JimmyVo16 (approved):

harr1424 (approved):
Changes to `NonAnonymousSendCommandTests` look good!

PR #7430, opened 2026-04-09
Arch/qa env seeding tweaks

MGibson1:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34880
https://bitwarden.atlassian.net/browse/PM-34881
https://bitwarden.atlassian.net/browse/PM-34886

## 📔 Objective

Various tweaks to smooth qa automation testing.

### DIFF
--- a/src/Core/Billing/Services/Implementations/LicensingService.cs
+++ b/src/Core/Billing/Services/Implementations/LicensingService.cs
@@ -31,8 +31,8 @@ namespace Bit.Core.Billing.Services;
 
 public class LicensingService : ILicensingService
 {
-    const string productionCertThumbprint = "‎B34876439FCDA2846505B2EFBBA6C4A951313EBE";
-    const string developmentCertThumbprint = "207E64A231E8AA32AAF68A61037C075EBEBD553F";
+    private const string productionCertThumbprint = "‎B34876439FCDA2846505B2EFBBA6C4A951313EBE";
+    private const string developmentCertThumbprint = "207E64A231E8AA32AAF68A61037C075EBEBD553F";
     private readonly X509Certificate2 _creationCertificate;
     private readonly List<X509Certificate2> _verificationCertificates;
     private readonly IGlobalSettings _globalSettings;
--- a/src/Core/Billing/Services/Implementations/LicensingService.cs
+++ b/src/Core/Billing/Services/Implementations/LicensingService.cs
@@ -22,6 +22,7 @@ using Bit.Core.Services;
 using Bit.Core.Settings;
 using Bit.Core.Utilities;
 using Duende.IdentityModel;
+using Duende.IdentityServer.Extensions;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
@@ -31,7 +32,10 @@ namespace Bit.Core.Billing.Services;
 
 public class LicensingService : ILicensingService
 {
-    private readonly X509Certificate2 _certificate;
+    const string productionCertThumbprint = "‎B34876439FCDA2846505B2EFBBA6C4A951313EBE";
+    const string developmentCertThumbprint = "207E64A231E8AA32AAF68A61037C075EBEBD553F";
+    private readonly X509Certificate2 _creationCertificate;
+    private readonly List<X509Certificate2> _verificationCertificates;
     private readonly IGlobalSettings _globalSettings;
     private readonly IUserRepository _userRepository;
     private readonly IOrganizationRepository _organizationRepository;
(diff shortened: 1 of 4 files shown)


kdenney (changes requested):
Logic looks good! Only thing I'd ask is if you could add some unit tests to cover the new changes? We've had issues in the past with changes to the licensing logic causing regressions so I like to make sure we cover that area pretty thoroughly. Thanks!

src/Core/Billing/Services/Implementations/LicensingService.cs
```diff
@@ -336,12 +363,20 @@ public ClaimsPrincipal GetClaimsPrincipalFromLicense(ILicense license)
...
+    }
+
+    private ClaimsPrincipal ValidateTokenWithIssuer(string token, string audience, X509SecurityKey issuerKey)
+    {
         var tokenHandler = new JwtSecurityTokenHandler();
         var validationParameters = new TokenValidationParameters
         {
             ValidateIssuerSigningKey = true,
-            IssuerSigningKey = new X509SecurityKey(_certificate),
+            IssuerSigningKey = issuerKey,
```
justindbaur:
```suggestion
            IssuerSigningKeys =_verificationCertificates.Select((c) => new X509SecurityKey(c)),
```

This should also mean you don't need to do any merges since only one would be valid anyways. This just says both of those keys are valid.

kdenney (dismissed):
Approving after our convo in slack. Tests are complicated without giving the licenses to CI. Ideally we refactor this in the future into two services but not needed for this fix.

src/Core/Billing/Services/Implementations/LicensingService.cs
```diff
@@ -63,31 +67,56 @@ public LicensingService(
...
                 "licensing.pfx", _globalSettings.LicenseCertificatePassword)
                 .GetAwaiter().GetResult();
         }
         else
         {
-            _certificate = CoreHelpers.GetCertificate(certThumbprint);
+            _creationCertificate = CoreHelpers.GetCertificate(creationCertThumbprint);
         }
+        // Creation cert can always be used to verify
+        _verificationCertificates.Add(_creationCertificate);
```
kdenney, line 109: I'm not sure if it matters, but looking at this again, I believe this line adds the devCert into `_verificationCertificates` twice if the environment is development.

1. Line 82, `_creationCertificate` is set to `devCert`
2. Line 94, `devCert` is added to `_verificationCertificates`
3. This line, it's added again because it's assigned on line 82

src/Core/Billing/Services/Implementations/LicensingService.cs
```diff
@@ -63,31 +67,56 @@ public LicensingService(
...
+            else
+            {
+                _creationCertificate = prodCert;
+            }
+
+            // non-production environments can use dev cert-generated licenses
+            if (!environment.IsProduction() && devCert != null) // Should already be set by above IsDevelopment check
+            {
+                _verificationCertificates.Add(devCert);
+            }
```
kdenney, line 95: The comment `non-production environments can use dev cert-generated licenses` is confusing here because if the environment is QA, then I believe `devCert` would be null but this comment makes it seem like the devCert would be added to `_verificationCertificates` for all non-production environments. Which is the intended behavior?

kdenney (commented):
After looking at this again, I have a couple questions.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -502,9 +498,13 @@ public static void AddCustomDataProtectionServices(
                     "dataprotection.pfx", globalSettings.DataProtection.CertificatePassword)
                     .GetAwaiter().GetResult();
             }
-            builder
-                .PersistKeysToAzureBlobStorage(globalSettings.Storage.ConnectionString, "aspnet-dataprotection", "keys.xml")
-                .ProtectKeysWithCertificate(dataProtectionCert);
+
+            if (dataProtectionCert != null)
```
amorask-bitwarden: ❓ In `main`, a missing `dataProtectionCert` would result in an `ArgumentNullException` being thrown in `PersistKeysToAzureBlobStorage`, which gives us a fast fail for a misconfiguration.

I might be overthinking this in the sense that `dataProtectionCert` couldn't be null (something else would fail prior to that), but if it could potentially be null, wouldn't we risk our data protection using system default configurations?

src/Core/Billing/Services/Implementations/LicensingService.cs
MGibson1: it does, but I decided to ignore that. If you're worried about duplicate key ids, we can create a distinct set, but I didn't think it mattered for the max three certs in the list

src/Core/Billing/Services/Implementations/LicensingService.cs
MGibson1: fixed in f4da360f632b6d3dc2e2be3798c5b6b6c65a5c67

src/Core/Billing/Services/Implementations/LicensingService.cs
MGibson1: eh, I decided to just calculate distinct: d4802d23b6dfa47f7695c835131075ec644801ad

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
MGibson1: Interesting thought. I hit this because my dev instance does have a storage connection string, but that storage account isn't used for storing my data cert (because of the early return). My feelings is that the fact that storage exists shouldn't imply storing data protection certs there. I'd like that to be a separate setting. I was taking the shorter path to just use storage if it's already being used. This would work fine for existing instances, but might alter the setup operations for new environments.

I could probably be convinced to just turn this off for dev. I had dropped it with the idea that it's possible someone would want this in dev, but it hasn't come up yet.

MGibson1: updated in 2749d31e9ca6f0280a4cdd99cabc037af1834a1a

amorask-bitwarden (dismissed):

src/Core/Billing/Services/Implementations/LicensingService.cs
```diff
@@ -31,7 +31,10 @@ namespace Bit.Core.Billing.Services;
 
 public class LicensingService : ILicensingService
 {
-    private readonly X509Certificate2 _certificate;
+    private const string productionCertThumbprint = "‎B34876439FCDA2846505B2EFBBA6C4A951313EBE";
+    private const string developmentCertThumbprint = "207E64A231E8AA32AAF68A61037C075EBEBD553F";
```
kdenney: ⛏️ Noticed this when I pulled the branch and loaded in Rider.  

`Name 'developmentCertThumbprint' does not match rule 'private_members_with_underscore'. Suggested name is '_developmentCertThumbprint'.`

```suggestion
    private const string _productionCertThumbprint = "‎B34876439FCDA2846505B2EFBBA6C4A951313EBE";
    private const string _developmentCertThumbprint = "207E64A231E8AA32AAF68A61037C075EBEBD553F";
```

src/Core/Billing/Services/Implementations/LicensingService.cs
```diff
@@ -63,31 +66,63 @@ public LicensingService(
...
+        var allowedThumbprints = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            CoreHelpers.CleanCertificateThumbprint(productionCertThumbprint),
+            CoreHelpers.CleanCertificateThumbprint(developmentCertThumbprint)
+        };
+        if (_verificationCertificates is null || _verificationCertificates.Count == 0
+            || _verificationCertificates.Any(c => !allowedThumbprints.Contains(c.Thumbprint)))
+        {
+            throw new Exception("Invalid license verifying certificate.");
+        }
```
kdenney, line 125: This is clean, nice! 😄

kdenney (changes requested):

kdenney (approved):

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -478,10 +478,6 @@ public static void AddCustomDataProtectionServices(
         this IServiceCollection services, IWebHostEnvironment env, GlobalSettings globalSettings)
     {
         var builder = services.AddDataProtection().SetApplicationName("Bitwarden");
-        if (env.IsDevelopment())
```
mzieniukbw, line 481: @MGibson1 this introduced a regression in development self-hosted environment. It is not impossible to run Api Self-Hosted, which now tries to load data protection from `/etc/bitwarden`, which does not exist in most dev environments

PR #7432, opened 2026-04-09
[PM-33866] Revocation Reasons: DDL Edition

sven-bitwarden:
## 🎟️ Tracking
[PM-33866](https://bitwarden.atlassian.net/browse/PM-33866)

## 📔 Objective
We're interested in tracking **why** a user is revoked, in order to enhance UX and unlock the capability to automate restoration of users.

This PR adds RevocationReason with an initial list of reasons based on existing flows and applicable policies, and then executes the painful task of adding the column everywhere.

### Notable Modifications
Restoration and revocation are now implemented by a pair of sprocs that handle it in bulk. The difference between the single and bulk path is negligible, so we're condensing the needed scripts down.

The old sprocs are now being 🗑️ 'd.

### Testing
I have ran the migration on MSSQL and the other 4 databases (MySql, Sqlite, Postgres, MariaDB) with EF using our migrate DB script. I then also ran the integration tests, which exercise the new RevocationReason parameter, as well as revoking/restoring through the new sprocs.

[PM-33866]: https://bitwarden.atlassian.net/browse/PM-33866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


util/Migrator/DbScripts/2026-04-13_00_AddRevocationReasonToOrganizationUser.sql
```diff
@@ -0,0 +1,679 @@
+-- Add RevocationReason column to OrganizationUser
+IF COL_LENGTH('[dbo].[OrganizationUser]', 'RevocationReason') IS NULL
+BEGIN
+    ALTER TABLE [dbo].[OrganizationUser]
+        ADD [RevocationReason] TINYINT NULL;
+END
+GO
+
+-- Refresh OrganizationUserView metadata to pick up the new column (it uses SELECT *)
+EXEC sp_refreshview N'[dbo].[OrganizationUserView]';
```
mkincaid-bw, line 10: It's best practice to refresh all views that select from an underlying table that gets modified (unless they are being re-defined in the PR anyway).

```suggestion
EXEC sp_refreshview N'[dbo].[OrganizationUserView]';
EXEC sp_refreshview N'[dbo].[ProviderOrganizationOrganizationDetailsView]';
EXEC sp_refreshview N'[dbo].[UserPremiumAccessView]';
```

util/Migrator/DbScripts/2026-04-13_01_AddBulkRevocationSprocs.sql
```diff
@@ -0,0 +1,55 @@
...
+        @OrganizationUserIds OUI ON OUI.[Id] = OU.[Id]
+    WHERE
+        OU.[Status] = -1 -- Only restore currently revoked users
+
+    EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationUserIds] @OrganizationUserIds
+END
+GO
+
+-- Drop legacy sprocs replaced by the bulk versions above
+DROP PROCEDURE IF EXISTS [dbo].[OrganizationUser_Deactivate];
```
mkincaid-bw: With our current deployment methodology, DB changes are deployed before code changes. If there is an issue with a deployment, we may roll back the app code changes, but DB changes do not (usually) get rolled back. Therefore, each PR has to support app-current and app-next. Dropping these procs would break the app if we had to rollback, and need to be removed from this PR. They can be done in a future PR once the app code is hardened and we are sure nothing is calling them.

mkincaid-bw (changes requested):

util/Migrator/DbScripts/2026-04-13_01_AddBulkRevocationSprocs.sql
sven-bitwarden: Good point, my bad I had gotten a little ambitious with dropping these in this PR.

src/Sql/dbo/Stored Procedures/OrganizationUser_CreateManyWithCollectionsAndGroups.sql
mkincaid-bw, line 1: I realize you are simply modifying this existing file but can you please rename the file to match the stored proc name:

`OrganizationUser_CreateManyWithCollectionsGroups.sql` --> `OrganizationUser_CreateManyWithCollectionsAndGroups.sql`

Side Note: `OrganizationUser_CreateManyWithCollectionsAndGroups` is the proper name but I think it's better to be consistent with the file name matching the proc name.

sven-bitwarden: No problem, fixed

mkincaid-bw (approved):

jrmccannon (approved):

PR #7442, opened 2026-04-10
[PM-12473] Add dedicated CollectionUser and CollectionGroup authorization handlers

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12473

## 📔 Objective

Extracts user-access and group-access authorization from `BulkCollectionAuthorizationHandler` into dedicated `CollectionUserAuthorizationHandler` and `CollectionGroupAuthorizationHandler`, gated behind a feature flag.


src/Api/AdminConsole/Controllers/GroupsController.cs
```diff
@@ -172,18 +177,13 @@ public async Task<GroupResponseModel> Put(Guid orgId, Guid id, [FromBody] GroupR
             }
         }
```
eliykat, line 189: This is authorization logic that should be encapsulated by the handler.
EDIT: the existing comment is also wrong - we do throw an error here!

src/Api/AdminConsole/Controllers/GroupsController.cs
```diff
@@ -290,4 +292,53 @@ public async Task PostDeleteUser(Guid orgId, Guid id, Guid orgUserId)
     {
         await DeleteUser(orgId, id, orgUserId);
     }
+
+    private async Task PutCollectionAccess_vNext(
+        GroupRequestModel model,
+        ICollection<Core.Models.Data.CollectionAccessSelection> currentAccess)
+    {
+        var (createIds, updateIds, deleteIds) = model.Collections.DiffAccessSelections(currentAccess);
```
eliykat: (no changes required, just sharing my thinking)

This is a great interface. It reflects what I've suspected for a while - the client should send a delta of what's changed, rather than this weird incomplete set we have to jump through hoops to reconcile. e.g. ideally the client would send something like

```
{
  delete: IEnumerable<Guid>
  create: IEnumerable<CollectionCreateRequest>
  update: IEnumerable<CollectionUpdateRequest>
}
```

which neatly aligns with authorization checks and the CRUD operations required by the db. That's a very simple request for the server to consume.

This diff logic you've added lets us work from a delta immediately without breaking the api. Later, we can update the client to send a delta, and remove the glue code.

src/Api/AdminConsole/Controllers/GroupsController.cs
```diff
@@ -290,4 +292,53 @@ public async Task PostDeleteUser(Guid orgId, Guid id, Guid orgUserId)
...
+
+        var allCollections = await _collectionRepository.GetManyByManyIdsAsync(
+            createIds.Concat(updateIds).Concat(deleteIds));
+
+        await AuthorizeCollectionAccessAsync(
+            allCollections.Where(c => createIds.Contains(c.Id)).ToList(), CollectionGroupOperations.Create);
+        await AuthorizeCollectionAccessAsync(
+            allCollections.Where(c => updateIds.Contains(c.Id)).ToList(), CollectionGroupOperations.Update);
+        await AuthorizeCollectionAccessAsync(
+            allCollections.Where(c => deleteIds.Contains(c.Id)).ToList(), CollectionGroupOperations.Delete);
```
eliykat: There is a bug here though, which reflects the weirdness of the current API. If a collection is missing from the request, either:
1. the user has permission to delete it, and it was omitted because it should be deleted -> delete
2. the user does not have permission to delete it, and it was omitted because the client is unaware of it -> do not delete

See the logic following this comment in the endpoint:

```
        // The client only sends collections that the saving user has permissions to edit.
        // We need to combine these with collections that the user doesn't have permissions for, so that we don't
        // accidentally overwrite those
```

Rather than throwing an error, the authorization check just adds back anything the user doesn't have permission to delete.

There is also some duplication: you're producing a diff here, but that logic is also producing its own diff of what's changed. Please see if we can combine/harmonize this logic somehow. Ideally, as suggested in my last comment, `DiffAccessSelections` could act as a glue/transition code to take the API request and translate it to a simple delta that we can act on. So there would be a clean separation of steps:

1. API request comes in
2. map to a delta
3. evaluate authorization

(This is exactly why this is a tech debt ticket 😁 )

eliykat: Claude example of expected failure scenario:

>   Concrete failure scenario
> 
>   A custom-role org user with ManageGroups (but no EditAnyCollection) updates a group. That group has Collection A (which they manage) and Collection B (which they don't manage). The client sends only Collection A. deleteIds contains
>   Collection B. The delete auth check fails → NotFoundException → the update is completely blocked, even though the user is making a legitimate edit.

src/Api/AdminConsole/Controllers/GroupsController.cs
```diff
@@ -130,9 +134,10 @@ public async Task<GroupResponseModel> Post(Guid orgId, [FromBody] GroupRequestMo
         if (model.Collections?.Any() == true)
         {
             var collections = await _collectionRepository.GetManyByManyIdsAsync(model.Collections.Select(a => a.Id));
-            var authorized =
-                (await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.ModifyGroupAccess))
-                .Succeeded;
+            IAuthorizationRequirement requirement = _featureService.IsEnabled(FeatureFlagKeys.CollectionUserCollectionGroupAuthorizationHandlers)
+                ? CollectionGroupOperations.Create
+                : BulkCollectionOperations.ModifyGroupAccess;
+            var authorized = (await _authorizationService.AuthorizeAsync(User, collections, requirement)).Succeeded;
```
eliykat: I'm not sure we should be passing collections as the resource here - the resource should match the action, and we're not creating a collection, we're creating a CollectionGroup association. I'll elaborate in the next comment as I think the approach in `OrganizationUsersController` is better.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -267,12 +269,26 @@ public async Task Invite(Guid orgId, [FromBody] OrganizationUserInviteRequestMod
...
             var collections = await _collectionRepository.GetManyByManyIdsAsync(model.Collections.Select(a => a.Id));
-            var authorized =
-                (await _authorizationService.AuthorizeAsync(User, collections, BulkCollectionOperations.ModifyUserAccess))
-                .Succeeded;
-            if (!authorized)
+
+            if (_featureService.IsEnabled(FeatureFlagKeys.CollectionUserCollectionGroupAuthorizationHandlers))
             {
-                throw new NotFoundException();
+                var resource = new CollectionUserAccessResource(collections.ToList(), null);
```
eliykat: I think `CollectionUserAccessResource` is a good idea. The idea of this interface is that you pass in the resource and the action you want to take on that resource. In db terms the resource is `CollectionUser`, but as we need access to the collection and don't want to fetch it multiple times, I think it's OK to have this DTO that includes the collection as well. Then the handler has access to all the context for authorization decisions. Compare to just passing in the collection - then the handler doesn't know who's being assigned to it or what permissions they're receiving, which could be relevant to authorization.

I don't quite like the `null` parameter to indicate "OrgUser that has not been created yet" but I'm not quite sure how else to handle it.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -408,29 +424,13 @@ public async Task Put(Guid orgId, Guid id, [FromBody] OrganizationUserUpdateRequ
             ? null
             : model.Groups;
```
eliykat, line 427: This is authorization logic that also needs to be inside the authorization layer. However, the way it mutates the request makes it hard to move in there at the moment (the authz handler returns a result but does not change the request like this).

I suggest that:

- this is left here, but
- you also check for it in the handler and fail authorization if the user is trying to add themselves to a group

Similar thoughts as above - if we later update the client to send a delta (of groups to add and remove), then we can remove this code and rely on the authorization check only (because the client will only be sending what it actually wants to do, not this extra info that we are conditionally ignoring). This code in the controller becomes a temporary adapter/glue code rather than permanent authz.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -838,4 +848,72 @@ private async Task<IDictionary<Guid, bool>> GetClaimedByOrganizationStatusAsync(
...
+
+        var allCollections = await _collectionRepository.GetManyByManyIdsAsync(
+            createIds.Concat(updateIds).Concat(deleteIds));
+
+        await AuthorizeCollectionUserAccessAsync(
+            allCollections.Where(c => createIds.Contains(c.Id)), CollectionUserOperations.Create, targetUserId);
+        await AuthorizeCollectionUserAccessAsync(
+            allCollections.Where(c => updateIds.Contains(c.Id)), CollectionUserOperations.Update, targetUserId);
+        await AuthorizeCollectionUserAccessAsync(
+            allCollections.Where(c => deleteIds.Contains(c.Id)), CollectionUserOperations.Delete, targetUserId);
```
eliykat: Same issue here for how the deleted collections are evaluated.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -442,10 +442,20 @@ public async Task Put(Guid orgId, Guid id, [FromBody] OrganizationUserUpdateRequ
         var readonlyCollectionIds = new HashSet<Guid>();
         foreach (var collection in currentCollections)
         {
-            if (!(await _authorizationService.AuthorizeAsync(User, collection, BulkCollectionOperations.ModifyUserAccess))
-                .Succeeded)
+            if (_featureService.IsEnabled(FeatureFlagKeys.CollectionUserCollectionGroupAuthorizationHandlers))
             {
-                readonlyCollectionIds.Add(collection.Id);
+                var resource = new CollectionUserAccessResource(new[] { collection }, organizationUser.UserId);
```
eliykat: 💭  CollectionUser is a link between a Collection and an OrganizationUser, so UserId instead of OrgUserId here is unexpected. Not sure how to resolve.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -221,12 +230,36 @@ public async Task PostBulkCollectionAccess(Guid orgId, [FromBody] BulkCollection
...
+
+            if (model.Groups?.Any() == true)
+            {
+                if (!(await _authorizationService.AuthorizeAsync(User, collections,
+                        CollectionGroupOperations.Create)).Succeeded)
+                {
+                    throw new NotFoundException();
+                }
+            }
+        }
```
eliykat: This change is nested enough that I think a v2 method is better, similar to the Put method.

Also similar concerns about what resource we are passing.

src/Api/Controllers/CollectionsController.cs
```diff
@@ -274,4 +307,94 @@ public async Task PostDeleteMany(Guid orgId, [FromBody] CollectionBulkDeleteRequ
...
+            throw new NotFoundException();
+        }
+
+        if (model.Users != null)
+        {
+            // TargetUserId is null because this endpoint modifies access for multiple users
+            // at the collection level; self-assignment protection is not applicable here
+            // (consistent with the pre-feature-flag behavior).
+            var resource = new CollectionUserAccessResource(new[] { collection }, null);
+            await AuthorizeUserAccessChangesAsync(model.Users, currentAccess.Users, resource);
```
eliykat: This is a leaky abstraction - the caller should pass in the information required and the handler should decide what to do with it, rather than the caller selectively bypassing certain checks.

Also - is it correct that we're bypassing this check?

src/Api/Vault/AuthorizationHandlers/Collections/CollectionGroupAuthorizationHandler.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            context.Fail();
+            return;
+        }
+
+        if (!_currentContext.UserId.HasValue)
+        {
+            context.Fail();
+            return;
+        }
```
eliykat: No need to call `context.Fail()` - the action will not be authorized unless you positively call `context.Succeed`. An early return is enough, and is better practice because it allows any other handlers we may add in the future to potentially authorize the action.

src/Api/AdminConsole/Authorization/Collections/AccessSelectionExtensions.cs
eliykat, line 1: Please add all new files to our code ownership.

I'll open a separate PR to move existing collection files.

src/Api/Vault/AuthorizationHandlers/Collections/CollectionUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,188 @@
...
+            context.Fail();
+            return;
+        }
+
+        _targetOrganizationId = resources.First().OrganizationId;
+
+        if (resources.Any(r => r.OrganizationId != _targetOrganizationId))
+        {
+            throw new BadRequestException("Requested collections must belong to the same organization.");
+        }
```
eliykat: I think this is one issue with the way our handlers are currently structured - it's optimized for the bulk case, but it makes it difficult to understand what the business rules are for any single change.

Ideally the authorization logic would be expressed in an atomic way for a single operation: can I add this (single) user to this (single) collection. Then if we need to provide a bulk path we can do that, and of course make sure it's performant, but the goal is that authz logic is stated separately from other concerns like fetching state or managing the list of multiple collectionUsers.

Claude example to try to better express my thoughts:

```c#
 public class CollectionUserAuthorizationHandler ...                                                                                                                                                                                   
  {                                                                                                                                                                                                                                   
      // Stateful layer: fetch data once, loop, aggregate                                                                                                                                                                               
      protected override async Task HandleRequirementAsync(...)                                                                                                                                                                         
      {                                                                                                                                                                                                                                 
          var org = ...;                                                                                                                                                                                                                
          var managedIds = await GetManagedCollectionIdsAsync();                                                                                                                                                                        
          var orphanedIds = await GetOrphanedCollectionIdsAsync(org);
                                                                                                                                                                                                                                        
          if (resource.Collections.All(c => CanModifyUserAccess(c, org, managedIds, orphanedIds)))    
          {                                                                                                                                  
              context.Succeed(requirement);                                                                                                                                                                                             
          }
      }                                                                                                                                                                                                                                 
                                                                                                                                                       
      // Stateless: expresses the rule for one collection, no I/O so we can loop cheaply                                                                                                         
      private static bool CanModifyUserAccess(                                                                                                         
          Collection collection,                                                                                                                                                                                                        
          CurrentContextOrganization? org,                                                                                                                                                                                              
          HashSet<Guid> managedIds,
          HashSet<Guid> orphanedIds)                                                                                                                                                                                                    
      {                                                                                                                                                
          if (managedIds.Contains(collection.Id))
              return true;
                                                                                                                                                                                                                                        
          return org is { Type: Owner or Admin }
              && orphanedIds.Contains(collection.Id);                                                                                                                                                                                   
      }                                                                                                                                                
  }   
```

The pattern: the handler knows how to get and arrange the data, the static method knows what the rule is. The static method is also trivially unit-testable without mocking anything.

This would be a new pattern for us so I'm open to further discussion or feedback or other ideas.

eliykat (changes requested):
Good work - lots of interesting issues here to dig into in how we structure this. Some of my comments may be open ended so let me know if you want to chat.

I think there will be some iteration here, so if you like you can focus on just 1 handler first, make sure we're happy with that one, then extend the pattern to the rest.

Summary of high level ideas I've touched on below:
1. Our current API is not very good but we don't want to change it right now, so let's add a translation layer: when the request comes in, calculate the delta of what's actually changing, then act on the delta only. Later we'll update the API to provide the delta directly. You've already started doing this, just making the idea explicit.
2. Authz layer should fully encapsulate and contain all authorization / business logic. Currently there is still some logic in the controller, or leaky abstractions where the caller still has to know how authz internals work.
3. Authorization handlers for bulk resources do not express their logic very clearly. We should add a stateless layer that expresses the logic atomically for a single action on a single resource. Then the handler just becomes about connecting the 2 layers.

eliykat (commented):
Good work on this. As discussed, we should take it back to the tech breakdown stage as there are some interesting decisions to be made here and it's worth discussing with the team at a higher level before we get into the code.

Some feedback based on this PR:
- this would ideally be broken up into multiple tickets, maybe 1 per resource and then 1 for the front-end changes
- instead of glue code to transform the current API request into a diff, let's just add new endpoints that have the API we want; then the front-end can toggle between old & new
- I like the layering of state orchestration (handlers) and then static methods (rules) - it makes it really easy to understand and test the actual logic being enforced
- I don't like the bespoke lifetime caching service - it feels like an antipattern in our architecture. We have caching infrastructure provided by our architecture team that we should use if needed - or if not, maybe a few duplicate reads aren't so bad (as long as it's not `O(n)` for example). Or maybe we just refactor the code so that the db query result can be reused in the same scope.
- "adding self" checks are poorly encapsulated - ideally they are in the rules layer rather than the orchestration layer needing to know about it. This is challenging for new orgUsers who don't have an ID yet though.
- we may consider using a query class for state orchestration instead of an authorization handler _if_ it solves some of these problems for us - the interface for handlers can be quite limiting and this is a complex case
- we shouldn't implement single & bulk logic separately - just implement bulk and use it for both cases (unless there are good performance reasons, but that should be fairly rare)

PR #7443, opened 2026-04-10
[PM-32068] - Org Ability Extended Cache

jrmccannon:
## 🎟️ Tracking
[PM-32068](https://bitwarden.atlassian.net/browse/PM-32068)

## 📔 Objective
This implements retrieving the org ability record via the Extended Cache (fusion cache).  This will be used instead of the in memory cache in order to reduce load on the org ability database table and solve the issue with servicebus.

[PM-32068]: https://bitwarden.atlassian.net/browse/PM-32068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrmccannon:
@claude be kind

src/Core/AdminConsole/AbilitiesCache/OrganizationAbilityCacheConstants.cs
eliykat, line 1: `BuildCacheKeyForOrganizationAbility` isn't a constant :P and I think there's little benefit to being in its own class - I suggest these are private properties in the cache service itself.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
```diff
@@ -44,19 +48,29 @@ public async Task<IDictionary<Guid, OrganizationAbility>> GetOrganizationAbiliti
     }
```
eliykat, line 48: Missed this one. Fwiw, I don't think there's a bulk interface for FusionCache - this is for convenience only. It should have a corresponding interface on the new cache service, which would just `await Task.WhenAll` individual cache fetches.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -295,6 +295,8 @@ public static void AddDefaultServices(this IServiceCollection services, GlobalSe
         services.AddTokenizers();
 
         services.AddScoped<IApplicationCacheService, FeatureRoutedCacheService>();
+        services.AddExtendedCache(OrganizationAbilityCacheConstants.CacheName, globalSettings);
+        services.AddScoped<IOrganizationAbilityCacheService, ExtendedOrganizationAbilityCacheService>();
```
eliykat, line 299: It would be better not to expose this at the top level here. Can we have a services extension method - e.g. `AddOrganizationAbilityCache` - which registers both? Similar to how our commands/queries/handlers are registered.

test/Core.Test/AdminConsole/AbilitiesCache/ExtendedOrganizationAbilityCacheServiceTests.cs
eliykat, line 1: It's not a blocker, but I'm not sure of the value of these tests. The class is extremely thin, there's not much to test here by design, and these tests either use mocks or assert against the same prod code that you're testing (i.e. `BuildCacheKeyForOrganizationAbility`). Consider deleting them and/or relying on integration tests instead.

eliykat (changes requested):

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
jrmccannon: That is being covered with this [ticket](https://bitwarden.atlassian.net/browse/PM-32073)

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationAbilityCacheTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        // Act - update the organization via the HTTP endpoint
+        var response = await _client.PutAsJsonAsync($"/organizations/{_organization.Id}", updateRequest);
+
+        // Assert - endpoint succeeded and cache was updated
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+        var cacheService = _factory.GetService<IApplicationCacheService>();
+        var ability = await cacheService.GetOrganizationAbilityAsync(_organization.Id);
+        Assert.NotNull(ability);
+        Assert.Equal(_organization.Id, ability.Id);
```
eliykat, line 94: This does not assert that the ability is updated (cf. the comment above). We need to check that the name and billing email have been updated.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationAbilityCacheTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        var deleteOwnerEmail = $"delete-test-{Guid.NewGuid()}@example.com";
+        await _factory.LoginWithNewAccount(deleteOwnerEmail);
+
+        var signUpResult = await OrganizationTestHelpers.SignUpAsync(
+            _factory,
+            plan: PlanType.EnterpriseAnnually,
+            ownerEmail: deleteOwnerEmail,
+            passwordManagerSeats: 5,
+            paymentMethod: PaymentMethodType.Card);
+        var orgToDelete = signUpResult.Item1;
```
eliykat, line 110: Unnecessary - could use the org created in `InitializeAsync`.

eliykat (changes requested):

eliykat (approved):
Feedback is non-blocking and may be addressed in a separate PR if desired.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationAbilityCacheTests.cs
jrmccannon: will take care of these in the bulk implementation

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationAbilityCacheTests.cs
jrmccannon: will take care of these in the bulk implementation

jrmccannon: They all share the same database so the organization needs to be separate as they might run out of order.

prograhamming (approved):

PR #7444, opened 2026-04-10
billing/pm-24665/license-file-generation-should-fail-for-unpaid-subscription

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24665
## 📔 Objective

The PR adds validation to the license generation endpoint that blocks license file downloads when an organization's subscription is in any of these statuses:

  - canceled - Subscription has been canceled
  - incomplete - Payment setup incomplete
  - incomplete_expired - Payment setup expired before completion
  - unpaid - Payment failed and not resolved (bonus validation)

  Technical Implementation

  Location: src/Core/Billing/Organizations/Queries/GetCloudOrganizationLicenseQuery.cs

  Changes:
  1. Added ValidateSubscriptionForLicenseGeneration() method that checks subscription status before allowing license generation
  2. Throws a BadRequestException with user-friendly error message: "Unable to generate license due to a payment issue. Please update your billing information or contact support for assistance."
  3. Validates at the API endpoint: GET /organizations/{id}/license

  Test Coverage:
  - Added 8 comprehensive test cases covering all blocked statuses and ensuring valid statuses (active, trialing, past_due) still work
## 📸 Screenshots

image


amorask-bitwarden (changes requested):
What about premium?

src/Core/Billing/Licenses/SubscriptionLicenseValidator.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Bit.Core.Billing.Constants;
+using Bit.Core.Exceptions;
+using Bit.Core.Models.Business;
+
+namespace Bit.Core.Billing.Licenses;
+
+public static class SubscriptionLicenseValidator
+{
+    public static void ValidateSubscriptionForLicenseGeneration(SubscriptionInfo subscriptionInfo)
```
amorask-bitwarden: ❌ I don't think there's any reason for this to take a `SubscriptionInfo` when we can just pass in `Subscription`.

src/Core/Billing/Licenses/Queries/GetUserLicenseQuery.cs
```diff
@@ -10,10 +11,14 @@ public interface IGetUserLicenseQuery
 }
 
 public class GetUserLicenseQuery(
-    IUserService userService) : IGetUserLicenseQuery
+    IUserService userService,
+    IStripePaymentService paymentService) : IGetUserLicenseQuery
 {
     public async Task<UserLicense> Run(User user)
     {
+        var subscriptionInfo = await paymentService.GetSubscriptionAsync(user);
```
amorask-bitwarden, line 21: ❌ This work should be done in the query itself and it should just use the `StripeAdapter` to get the subscription per the comment here: https://github.com/bitwarden/server/pull/7444/changes#r3081694476

src/Core/Billing/Organizations/Queries/GetCloudOrganizationLicenseQuery.cs
```diff
@@ -51,6 +48,7 @@ public async Task<OrganizationLicense> GetLicenseAsync(Organization organization
         }
 
         var subscriptionInfo = await GetSubscriptionAsync(organization);
+        SubscriptionLicenseValidator.ValidateSubscriptionForLicenseGeneration(subscriptionInfo);
```
amorask-bitwarden: See comment here: https://github.com/bitwarden/server/pull/7444/changes#r3081694476

amorask-bitwarden (changes requested):

amorask-bitwarden (changes requested):
Looking at this further, I'm not sure an entirely new validator is required for this work, which is evidenced by fact we're only checking the subscription status to make sure the subscription isn't terminal. Is there any reason we can't just check the subscription status in each license query? That way you don't have to double fetch the subscription in the org license query nor did you need an entirely new file.

amorask-bitwarden (approved):

PR #7445, opened 2026-04-10
[PM-35150] Make `Setup` testable and add test for install

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35150

## 📔 Objective

This makes the `Setup` project much more testable. It is an explicit test because the test currently requires openssl and bash which will not work on all development environments. It makes it possible to invoke the program with a custom root directory so that you can save things into a temp directory. This makes it possible to run multiple tests in parallel without worrying about them clobbering each other. It also allows the ability to respond to prompts through mocking instead of dealing with stdin. Right now you still can't assert on logs at all since all tests will go to the same stdout and be unreliable. Making that mockable could be a good next step but for now you just need to assert on the files it creates which is the most important outcome of this utility. 

The one test that I wrote to prove this working is for the install process and I make a lot of assertions about the cert that we generate for TLS. This is because we may soon be switching that code to use managed C# API's instead of using the `openssl` CLI utility. There are still a lot of file assertions that could be made from the `-install` path but I left this short of that.

### DIFF
--- a/util/Setup/Program.cs
+++ b/util/Setup/Program.cs
@@ -153,7 +153,7 @@ public class Program
         Console.WriteLine(string.Empty);
     }
 
-    private static void Update(Application config)
+    private static void Update(Application application)
     {
         // This portion of code checks for multiple certs in the Identity.pfx PKCS12 bag.  If found, it generates
         // a new cert and bag to replace the old Identity.pfx.  This fixes an issue that came up as a result of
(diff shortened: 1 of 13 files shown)


util/Setup/Program.cs
```diff
@@ -135,24 +153,24 @@ private static void Install()
         Console.WriteLine(string.Empty);
     }
 
-    private static void Update()
+    private static void Update(Application config)
```
dereknance: ⛏️ Rename `config` to `application` for consistency with other usage.

```suggestion
    private static void Update(Application application)
```

dereknance (dismissed):
This looks good! Only a single non-blocking nitpick.

dereknance (approved):

PR #7447, opened 2026-04-10
[PM-32069] Add ExtendedProviderAbilityCacheService

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32069
https://bitwarden.atlassian.net/browse/PM-32112

## 📔 Objective

1. Add extended cache for the provider service (IProviderAbilityCacheService).
2. Inject IProviderAbilityCacheService into FeatureRoutedCacheService behind a feature flag.
3. Add unit tests.

## 📸 Screenshots

This code isn't wired up yet, so we can't manually test it, but I feel like the automated test coverage is sufficient.


JimmyVo16:
I looked at the failed Build Docker images (SeederApi, ./util, linux/amd64, linux/arm64, true) job. It looks like a permission issue when logging into Azure. I don't think this is related to my change.

JimmyVo16:
Added @jrmccannon as a reviewer because he has context working with the organization cache. @BTreston, feel free to add feedback, though.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -294,6 +294,10 @@ public static void AddDefaultServices(this IServiceCollection services, GlobalSe
         services.AddOptionality();
         services.AddTokenizers();
 
+        services.AddDistributedCache(globalSettings);
```
jrmccannon: Did you run into an error with this?

I would say there's a chance this would error as it attempts to add more than 1 Distributed cache. It is In several startup files.

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -294,6 +294,10 @@ public static void AddDefaultServices(this IServiceCollection services, GlobalSe
...
+        services.AddExtendedCache(ExtendedProviderAbilityCacheService.CacheName, globalSettings);
+        services.AddSingleton<IProviderAbilityCacheService, ExtendedProviderAbilityCacheService>();
```
jrmccannon: Move this to an extension method similar to [OrgAbility](https://github.com/bitwarden/server/blob/d9aef5c58f2fefbfa9045e52ba9ff477f66eab19/src/Core/AdminConsole/AbilitiesCache/OrganizationAbilityServiceCollectionsExtension.cs#L6).

src/Events/Startup.cs
```diff
@@ -84,6 +84,8 @@ public void ConfigureServices(IServiceCollection services)
 
         // Add event integration services
         services.AddDistributedCache(globalSettings);
+        services.AddExtendedCache(ExtendedProviderAbilityCacheService.CacheName, globalSettings);
+        services.AddSingleton<IProviderAbilityCacheService, ExtendedProviderAbilityCacheService>();
```
jrmccannon: Use extension method instead

test/Core.Test/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheServiceTests.cs
```diff
@@ -0,0 +1,157 @@
...
+using NSubstitute;
+using Xunit;
+using ZiggyCreatures.Caching.Fusion;
+
+using ProviderEntity = Bit.Core.AdminConsole.Entities.Provider.Provider;
+
+namespace Bit.Core.Test.AdminConsole.AbilitiesCache;
+
+[SutProviderCustomize]
+public class ExtendedProviderAbilityCacheServiceTests
```
jrmccannon: Is this testing anything meaningful? I'd say remove these unit tests and replace with some integration tests to validate.

jrmccannon (changes requested):
Just a few things @eliykat had mentioned to me and we should be consistent across the two ability caches.

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
```diff
@@ -0,0 +1,33 @@
...
+    [FromKeyedServices(ExtendedProviderAbilityCacheService.CacheName)] IFusionCache cache,
+    IProviderRepository providerRepository)
+    : IProviderAbilityCacheService
+{
+    public const string CacheName = "ProviderAbilities";
+
+    public async Task<ProviderAbility?> GetProviderAbilityAsync(Guid providerId)
+    {
+        return await cache.GetOrSetAsync<ProviderAbility?>(
+            $"{providerId}",
```
eliykat, line 19: @jrmccannon this is also a different way of constructing the cache key compared to yours - while the guid should be unique, the prefix seems like better practice to me if required to namespace it. Or does the CacheName already namespace it? I don't mind, as long as the approach is consistent.

JimmyVo16: Yes per @justindbaur's comment

image

JimmyVo16: We should update the Org cache since I think not having unnecessary prefixes is a better path.

src/Core/AdminConsole/AbilitiesCache/ProviderAbilityServiceCollectionsExtension.cs
```diff
@@ -0,0 +1,12 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Bit.Core.AdminConsole.AbilitiesCache;
+
+public static class ProviderAbilityServiceCollectionsExtension
+{
+    public static IServiceCollection AddProviderAbilityCache(this IServiceCollection services,
+        GlobalSettings globalSettings) =>
+        services.AddExtendedCache(ExtendedProviderAbilityCacheService.CacheName, globalSettings)
+            .AddSingleton<IProviderAbilityCacheService, ExtendedProviderAbilityCacheService>();
```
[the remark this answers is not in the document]
JimmyVo16: @jrmccannon , I think we should use `AddSingleton` for this since it’s a stateless service. Let me know what you think.

I know you have a PR open for the multiple use case. I can add a request for changes on that one, if you’d like.

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
```diff
@@ -0,0 +1,33 @@
...
+            async _ => await providerRepository.GetAbilityAsync(providerId)
+        );
```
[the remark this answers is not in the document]
JimmyVo16: [comment](https://github.com/bitwarden/server/pull/7447#discussion_r3094506416)

jrmccannon (approved):
Looks good. I took recommendations from this and made mine consistent as well.

Banrion (approved):

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
eliykat: Thanks for the info! I agree.

eliykat (commented):
Nothing more needed from me, you're all over it 👍

PR #7389, opened 2026-04-04
[PM-34595] Add provider authorization attributes

eliykat:
## Tracking

https://bitwarden.atlassian.net/browse/PM-34595

## Objective

Introduces the declarative authorization pattern for Provider endpoints, following the same approach as the existing Organization authorization infrastructure (see `OrganizationRequirementHandler`).

This will let us change from the current imperative pattern, which relies on the `CurrentContext` god class:

```
    [HttpGet("")]
    public async Task<ListResponseModel<ProviderOrganizationOrganizationDetailsResponseModel>> Get(Guid providerId)
    {
        if (!_currentContext.AccessProviderOrganizations(providerId))
        {
            throw new NotFoundException();
        }
```

To a declarative pattern:

```
    [HttpGet("")]
    [Authorize<ProviderUserRequirement>]
    public async Task<ListResponseModel<ProviderOrganizationOrganizationDetailsResponseModel>> Get(Guid providerId)
    {
```

This PR is **plumbing only** — no controller behavior changes. The handler and requirements are registered but not yet wired to any endpoints. A follow-up PR will update the controllers to use these requirements.

There is also some reorganization of the file structure to better accommodate these new files.

## Code changes

New types:
- `IProviderRequirement` — base interface for provider authorization requirements
- `ProviderRequirementHandler` — central handler that extracts the provider from route/claims and delegates to requirement implementations
- `ProviderClaimsExtensions` — helpers to parse `CurrentContextProvider` from a `ClaimsPrincipal`
- `HttpContextExtensions.GetProviderId()` — route param helper mirroring existing `GetOrganizationId()`
- requirements for Provider Users, Provider Admins, and a semantic `ManageProviderUsers` requirement.

PR #7450, opened 2026-04-11
[PM-34595] Update provider controllers to use authz attribute

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34595

## 📔 Objective

Primary objective: update provider controllers to use the `Authorize` attribute and the requirements added in #7389 . Add some basic tests for each controller - each endpoint is not tested comprehensively, but there are 'sentinel' tests to assure us that the attributes are working as intended end-to-end.

Secondary objective: break inheritance relationship between `ProviderClientsController` and the base controller classes still owned and used by Billing. This should be done to clarify our cross-team interfaces, and because the base controller implements its own authorization checks which are now duplicative. Changes:
- bring the `TryGetBillableProvider` logic into our controller as a private method - remove the role check which is handled by the attribute
- update our base controller class to expose response types directly, so that we can use these instead of the equivalent methods on the base billing controller


src/Api/AdminConsole/Controllers/ProviderClientsController.cs
eliykat, line 139: I have not used the base class helper here because I'm not sure this should be a 401 (more likely a 403) and I don't want to encourage other consumers to include error messages in their 401 responses. However, it was out of scope to address further here.

eliykat:
@JimmyVo16 and @sven-bitwarden this will be of interest to you given our tech debt priorities - this is just some work that I already had stashed, and I expect the other controllers will follow this pattern.

src/Api/AdminConsole/Controllers/ProviderUsersController.cs
```diff
@@ -49,59 +48,44 @@ public async Task<ProviderUserResponseModel> Get(Guid providerId, Guid id)
...
-            throw new NotFoundException();
-        }
-
         var invite = ProviderUserInviteFactory.CreateReinvite(new[] { id },
             _userService.GetProperUserId(User).Value, providerId);
         await _providerService.ResendInvitesAsync(invite);
     }
 
     [HttpPost("{id:guid}/accept")]
+    [AllowAnonymous]
```
JimmyVo16: Is this intentional? Just checking if we're introducing a breaking change because of it.

test/Api.Test/AdminConsole/Controllers/ProviderClientsControllerTests.cs
```diff
@@ -278,5 +317,24 @@ await sutProvider.GetDependency<IOrganizationRepository>().Received(1)
...
+        provider.Status = ProviderStatusType.Billable;
+        sutProvider.GetDependency<IProviderRepository>().GetByIdAsync(provider.Id).Returns(provider);
+    }
+
+    private static void AssertUnauthorized(IResult result)
+    {
+        Assert.IsType<UnauthorizedHttpResult>(result);
+    }
+
+    private static void AssertNotFound(IResult result)
```
JimmyVo16, line 334: FYI: I like small, well-named test methods like this. They make things easier to read and keep code in manageable blocks.

src/Api/AdminConsole/Controllers/ProviderUsersController.cs
JimmyVo16: Looks like this is done because of AdminConsoleControllersAuthorizationTests. Was this endpoint anonymous before?

JimmyVo16 (changes requested):
Overall looks good and approved. Just one question about the [AllowAnonymous].

src/Api/AdminConsole/Controllers/ProviderUsersController.cs
eliykat: This was following the pattern of other accept endpoints which use this attribute, but it is a bit counterintuitive. `[AllowAnonymous]` bypasses the class-level `[Authorize("Application")]` which requires an authenticated user; but then the controller code checks JWT claims for an authenticated user anyway.

I have added a new attribute, `[NoopAuthorize]` which (as the name suggests) does nothing, but indicates "yes I have thought about endpoint-specific authorization and intentionally not added any here". It is therefore accepted as valid by `AdminConsoleControllersAuthorizationTests`, because sometimes that is a valid decision, we just want to make sure it's intentional. Unlike `AllowAnonymous`, it does not bypass the class-level authorization policy. (Note that we still need the `user == null` check anyway for null handling.)

Open to other ideas here, we could alternatively do something like `Authorize<AuthenticatedUserRequirement>` which is maybe clearer, but duplicates the class-level policy which we should be able to rely on.

JimmyVo16: We can add those changes in follow-up PRs as long as we’re not introducing any issues in this PR.

JimmyVo16 (approved):

sven-bitwarden (approved):

PR #7228, opened 2026-03-16
PM-31923 adding the whole report endpoints v2

prograhamming:
## 🎟️ Tracking

This is a PR for user story [PM-31923](https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?selectedIssue=PM-31923)

## 📔 Objective

Creating new V2 endpoints for read and update operations on the whole report in the database. This will also include the logic for saving a reportData file in Azure Blob storage and server if self-hosted.

**Documentation:** 
- Get latest report -> [PM-31923_OrganizationReports_GetLatestOrganizationReportAsync.md](https://gist.github.com/prograhamming/67fe8c408ebd07441cf27145bb62b57d)
- Get report by Id ->  [PM-31923_OrganizationReports_GetOrganizationReportAsync.md](https://gist.github.com/prograhamming/1fce449483ad46b637f51a66c06d0aed)
- Create report -> [PM-31923_OrganizationReports_CreateOrganizationReportAsync.md](https://gist.github.com/prograhamming/c4d370fd83bc7b857d25fc121fe5df84)
- Update report metadata -> [PM-31923_OrganizationReports_UpdateOrganizationReport.md](https://gist.github.com/prograhamming/d5a8f529ed6f544d9c1bee59f38dd6d0)
- Get Summary data by date range ->  [PM-31923_OrganizationReports_DIRT-GetOrganizationReportSummaryDataByDateRange.md](https://gist.github.com/prograhamming/01d97c77becb492cef3c0340feaad6e0)
- Upload report file [self-hosted] -> [PM-31923_OrganizationReports_UploadReportFile.md](https://gist.github.com/prograhamming/94d32ba17b921abaa548cae97370a370)
- Download report file [self-hosted] -> [PM-31923_OrganizationReports_DownloadReportFileAsync.md](https://gist.github.com/prograhamming/be46b7489aa07177081f07362358f337)
- Validate Azure file [cloud] -> [PM-31923_OrganizationReports_AzureValidateFileAsync.md](https://gist.github.com/prograhamming/c23c8efd49bf5ceb01c02d84251f9ba0#file-pm-31923_organizationreports_azurevalidatefileasync-md)

[PM-31923]: https://bitwarden.atlassian.net/browse/PM-31923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7465, opened 2026-04-14
[PM-34909] Serve only validated report files from get-latest and self-hosted file endpoints

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34909

## 📔 Objective

Closes correctness gaps in the Access Intelligence report file storage endpoints introduced by [#7228](https://github.com/bitwarden/server/pull/7228), gated behind `pm-31920-access-intelligence-azure-file-storage`:

1. **GetLatest endpoint**: dispatches to a V2-specific SP that filters to validated files, with V1 inline-data fallback so a flag flip doesn't 404 orgs that only have legacy inline reports.
2. **Download endpoint (self-hosted)**: when V2 is on, rejects requests for files that haven't been validated.
3. **Upload endpoint (self-hosted)**: rejects re-uploads against an already-validated file.

When the flag is OFF, all three endpoints behave as they do today.

## Notes

- The design evolved across multiple review rounds (scope change 2026-05-14, regression fix 2026-06-02, defense-in-depth fix 2026-06-03). The commits and review threads cover the why; each commit message is self-describing.
- Jira AC text says `IS NOT NULL` but the column is `NOT NULL` with `string.Empty` default; the shipped predicate is `<> ''`. Will update the AC post-merge.

Manual flag-flip regression test verified locally.

[PM-31923]: https://bitwarden.atlassian.net/browse/PM-31923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-34909]: https://bitwarden.atlassian.net/browse/PM-34909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -141,16 +141,12 @@ public async Task<IActionResult> GetLatestOrganizationReportAsync(Guid organizat
 
         await AuthorizeAsync(organizationId);
 
-        var latestReport = await _getOrganizationReportQuery.GetLatestOrganizationReportAsync(organizationId);
-
-        if (latestReport == null)
-        {
-            throw new NotFoundException();
-        }
```
Banrion, line 149: ⚠️ We still need to check for null in the case that no report exists

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -141,16 +141,12 @@ public async Task<IActionResult> GetLatestOrganizationReportAsync(Guid organizat
...
+        var filterByValidated = _featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2);
+        var latestReport = await _getOrganizationReportQuery.GetLatestOrganizationReportAsync(organizationId, filterByValidated);
 
         var response = new OrganizationReportResponseModel(latestReport);
 
-        if (_featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2))
+        if (filterByValidated)
```
Banrion: ⚠️ Technically correct since this variable is derived from the feature flag, but changing it to the `filterByValidated` can be confusing when reviewing why this if else statement exists. The main logic on this branch is to handle report files differently than reports stored in the database. I suggest reverting this or creating a boolean variables for the feature flag to drive this branch.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -141,16 +141,12 @@ public async Task<IActionResult> GetLatestOrganizationReportAsync(Guid organizat
...
         {
             var fileData = latestReport.GetReportFile();
             if (fileData is { Validated: true })
```
Banrion, line 155: ⛏️ We can remove the check for validation since the query now returns only validated files.

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -449,6 +445,11 @@ public async Task<IActionResult> DownloadReportFileAsync(Guid organizationId, Gu
             throw new NotFoundException();
         }
 
+        if (_featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2) && !fileData.Validated)
```
Banrion, line 451: ⚠️ This route is already gated by the feature flag and validated will always be true from the query. This can be removed

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetLatestByOrganizationId.sql
```diff
@@ -8,5 +9,9 @@ BEGIN
         *
     FROM [dbo].[OrganizationReportView]
     WHERE [OrganizationId] = @OrganizationId
+        AND (
```
mkincaid-bw: This is an anti-pattern in SQL and could cause performance issues and won't scale well. There are a couple better options available:

1. Create a new stored proc without the `@FilterByValidated` and have it do the filtering logic all of time (then call the appropriate one in code)
2. Re-write the logic to perform separate queries based on the value of `@FilterByValidated`. Something like this:
```sql
CREATE OR ALTER PROCEDURE [dbo].[OrganizationReport_GetLatestByOrganizationId]
    @OrganizationId UNIQUEIDENTIFIER,
    @FilterByValidated BIT = 0
AS
BEGIN
    SET NOCOUNT ON;

    IF @FilterByValidated = 1
    BEGIN
        SELECT TOP 1 
            *
        FROM 
            [dbo].[OrganizationReportView]
        WHERE 
            [OrganizationId] = @OrganizationId
            AND JSON_VALUE([ReportFile], '$.Validated') = 'true'
        ORDER BY 
            [RevisionDate] DESC;
    END
    ELSE
    BEGIN
        SELECT TOP 1
            *
        FROM 
            [dbo].[OrganizationReportView]
        WHERE 
            [OrganizationId] = @OrganizationId
        ORDER BY 
            [RevisionDate] DESC;
    END
END
GO
```
This will allow SQL to create separate execution plans for each `SELECT` statement and will perform better overall.

NOTE: This stored procedure is named incorrectly, and should be named something like `OrganizationReport_ReadLatestByOrganizationId`. However, since you are modifying an existing proc, it's fine if you don't want to rename it. If you opt for creating a new stored procedure, it should be named something like `OrganizationReport_ReadLatestByOrganizationIdFilterByValidated`

mkincaid-bw (changes requested):

src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
```diff
@@ -21,13 +21,22 @@ public OrganizationReportRepository(IServiceScopeFactory serviceScopeFactory,
...
-                .Where(p => p.OrganizationId == organizationId)
+
+            IQueryable<Models.OrganizationReport> query = dbContext.OrganizationReports
+                .Where(p => p.OrganizationId == organizationId);
+
+            if (filterByValidated)
+            {
+                query = query.Where(p => p.ReportFile != null
+                    && p.ReportFile.Contains("\"Validated\":true"));
+            }
```
Banrion: ⚠️ Similar to the comment in `OrganizationReport_GetLatestByOrganizationId.sql`, it would be more clear to have two separate query statements rather than modifying one. Not certain that it's an anti-pattern in this situation, but it's not easy to read.

Banrion (changes requested):

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetLatestByOrganizationId.sql
AlexRubik: Conditional WHERE is gone: OrganizationReport_ReadLatestByOrganizationId (OrganizationReport_ReadLatestByOrganizationId.sql) always filters by validated file, and OrganizationReport_GetLatestByOrganizationId (OrganizationReport_GetLatestByOrganizationId.sql) now filters [ReportData] <> '' to skip V2 file-storage rows. Each SP gets its own execution plan. Thanks for flagging! Shipped in [d69fab54a](https://github.com/bitwarden/server/pull/7465/commits/d69fab54a12785a7443d9c8504cf15f2b8242304).

src/Api/Dirt/Controllers/OrganizationReportsController.cs
AlexRubik: Null check moved down a layer. Both GetLatestOrganizationReportAsync and ReadLatestOrganizationReportAsync in GetOrganizationReportQuery (.../Reports/ReportFeatures/GetOrganizationReportQuery.cs) throw NotFoundException on null, covered by _NullResult_ThrowsNotFoundException tests for each in GetOrganizationReportQueryTests.cs. So the controller never sees a null latestReport. Shipped in [c50e572d8](https://github.com/bitwarden/server/pull/7465/commits/c50e572d82a1cf29a06f63095c88ddcf9d378bfa).

src/Api/Dirt/Controllers/OrganizationReportsController.cs
AlexRubik: No longer applies after the scope change. No shared bool threaded through anymore: the controller (OrganizationReportsController.cs) dispatches by method name based on IsEnabled(AccessIntelligenceVersion2). Shipped in [c50e572d8](https://github.com/bitwarden/server/pull/7465/commits/c50e572d82a1cf29a06f63095c88ddcf9d378bfa).

src/Api/Dirt/Controllers/OrganizationReportsController.cs
AlexRubik: Dropped this and kept fileData != null in OrganizationReportsController.cs . Shipped in [c50e572d8](https://github.com/bitwarden/server/pull/7465/commits/c50e572d82a1cf29a06f63095c88ddcf9d378bfa).

src/Api/Dirt/Controllers/OrganizationReportsController.cs
AlexRubik: We look the report row up by report id via GetAuthorizedReportAsync which doesn't filter by validated (only the new ReadLatestByOrganizationIdAsync in the repo layer does, and that's the latest-by-org path, not by-id). The route (DownloadReportFileAsync) also has no [RequireFeature] attribute, just [SelfHosted(SelfHostedOnly = true)]. So with the flag ON, this guard is the only block against streaming an unvalidated file. 

Guard shipped in [091d50362](https://github.com/bitwarden/server/pull/7465/commits/091d503627d57b162479aac4febc7ebe6712e3ab).

src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
AlexRubik: Each method in OrganizationReportRepository.cs is now a single query mirroring its sproc (GetLatestByOrganizationIdAsync and ReadLatestByOrganizationIdAsync). No more conditional Where. 

[d69fab54a](https://github.com/bitwarden/server/pull/7465/commits/d69fab54a12785a7443d9c8504cf15f2b8242304)

src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
```diff
@@ -27,7 +27,27 @@ public async Task<OrganizationReport> GetLatestByOrganizationIdAsync(Guid organi
...
+
+    public async Task<OrganizationReport> ReadLatestByOrganizationIdAsync(Guid organizationId)
+    {
+        using (var scope = ServiceScopeFactory.CreateScope())
+        {
+            var dbContext = GetDatabaseContext(scope);
+            var result = await dbContext.OrganizationReports
+                .Where(p => p.OrganizationId == organizationId
+                    && p.ReportFile != null
+                    && p.ReportFile.Contains("\"Validated\":true"))
```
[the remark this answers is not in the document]
AlexRubik: The substring match is our repo's convention for these cases currently. Introducing this new json extract convention is out of scope for this pr but I still made a comment here https://github.com/bitwarden/server/pull/7465/commits/8561e0098e702477327facb44f97c8b0c65bd7a9

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
```diff
@@ -0,0 +1,13 @@
+CREATE PROCEDURE [dbo].[OrganizationReport_ReadLatestByOrganizationId]
```
mkincaid-bw, line 1: Since `OrganizationReport_GetLatestByOrganizationId` should have been named `OrganizationReport_ReadLatestByOrganizationId`, the naming of this new procedure insinuates these two procs return the same data. If this new proc will eventually replace `OrganizationReport_GetLatestByOrganizationId`, then I'm okay with it; otherwise I think it should be named something else to differentiate that it returns different data, maybe `OrganizationReport_ReadLatestValidatedByOrganizationId`?

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
```diff
@@ -0,0 +1,13 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT TOP 1
+        *
+    FROM [dbo].[OrganizationReportView]
+    WHERE [OrganizationId] = @OrganizationId
+        AND JSON_VALUE([ReportFile], '$.Validated') = 'true'
+    ORDER BY [RevisionDate] DESC
```
mkincaid-bw: SQL Keywords should be on their own line
https://contributing.bitwarden.com/contributing/code-style/sql/#select-statements

```suggestion
    SELECT
        TOP 1 *
    FROM
        [dbo].[OrganizationReportView]
    WHERE
        [OrganizationId] = @OrganizationId
        AND JSON_VALUE([ReportFile], '$.Validated') = 'true'
    ORDER BY
        [RevisionDate] DESC
```

mkincaid-bw (changes requested):

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
AlexRubik: https://github.com/bitwarden/server/pull/7465/commits/d5c0bd9b2d73776175e314711fd98ba22bd867a3

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
AlexRubik: > If this new proc will eventually replace OrganizationReport_GetLatestByOrganizationId, then I'm okay with it

Yes that's the plan, once `pm-31920-access-intelligence-azure-file-storage` flag is permanently ON, the V1 `OrganizationReport_GetLatestByOrganizationId` SP becomes dead code and we remove it, leaving `OrganizationReport_ReadLatestByOrganizationId` as the only latest-by-org SP.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_GetLatestByOrganizationId.sql
```diff
@@ -4,9 +4,13 @@ AS
 BEGIN
     SET NOCOUNT ON
 
-    SELECT TOP 1
-        *
-    FROM [dbo].[OrganizationReportView]
-    WHERE [OrganizationId] = @OrganizationId
-    ORDER BY [RevisionDate] DESC
+    SELECT
+        TOP 1 *
```
mkincaid-bw: Sorry, my previous code suggestion was incorrect.
```suggestion
    SELECT TOP 1
        *
```

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
```diff
@@ -0,0 +1,16 @@
+CREATE PROCEDURE [dbo].[OrganizationReport_ReadLatestByOrganizationId]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT
+        TOP 1 *
```
mkincaid-bw: Same as above
```suggestion
    SELECT TOP 1
        *
```

mkincaid-bw (dismissed):
I committed a couple minor formatting changes to correct my incorrect suggestions. Everything else LGTM.

util/Migrator/DbScripts/2026-05-19_00_AddOrganizationReportReadLatestSp.sql
```diff
@@ -0,0 +1,17 @@
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationReport_ReadLatestByOrganizationId]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT
+        TOP 1 *
```
mkincaid-bw:
```suggestion
    SELECT TOP 1
        *
```

util/Migrator/DbScripts/2026-05-19_01_UpdateOrganizationReportGetLatestSkipEmptyReportData.sql
```diff
@@ -0,0 +1,17 @@
+CREATE OR ALTER PROCEDURE [dbo].[OrganizationReport_GetLatestByOrganizationId]
+    @OrganizationId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT
+        TOP 1 *
```
mkincaid-bw:
```suggestion
    SELECT TOP 1
        *
```

mkincaid-bw (dismissed):
Forgot to update the migration files. Should be good now.

test/Api.Test/Dirt/OrganizationReportsControllerTests.cs
Banrion: ❓ Since this task was updated to use `ReadLatestOrganizationReportAsync`, should the comment `// GetLatestOrganizationReportAsync` and task name, `GetLatestOrganizationReportAsync_WithValidatedFile_ReturnsOkWithDownloadUrl`, be updated?

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
```diff
@@ -0,0 +1,16 @@
...
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT TOP 1
+        *
+    FROM
+        [dbo].[OrganizationReportView]
+    WHERE
+        [OrganizationId] = @OrganizationId
+        AND JSON_VALUE([ReportFile], '$.Validated') = 'true'
```
Banrion: ⚠️ Caught a regression while reviewing: With the current filter, the first read after enabling the flag returns a 404 for orgs that only have an inline report. The last report silently disappears from the UI. Widening the stored procedure to `[ReportData] <> '' OR JSON_VALUE(ReportFile,'$.Validated') = 'true'` fixes this without re-introducing parameter-sniffing (still deterministic, still one execution plan). The client's `FileReportPersistenceService` already has an inline-data fallback path, so no client changes needed.

Keeping the v1 sp narrow for now since the feature flag off client (DefaultReportPersistenceService) is v1-only and can't render file-backed rows. Follow-up prs will remove the dispatcher on both sides to deliver full functional rollback, at which point the v1 sp becomes dead code per the existing flag-permanence cleanup plan.

Requires test updates.

```suggestion
        AND (
            JSON_VALUE([ReportFile], '$.Validated') = 'true'
            OR [ReportData] <> ''
        )
```

util/Migrator/DbScripts/2026-05-19_00_AddOrganizationReportReadLatestSp.sql
```diff
@@ -0,0 +1,17 @@
...
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT TOP 1
+        *
+    FROM
+        [dbo].[OrganizationReportView]
+    WHERE
+        [OrganizationId] = @OrganizationId
+        AND JSON_VALUE([ReportFile], '$.Validated') = 'true'
```
Banrion: ⚠️ Migration changes required to fix [regression](https://github.com/bitwarden/server/pull/7465/changes#r3336747805):

```suggestion
        AND (
            JSON_VALUE([ReportFile], '$.Validated') = 'true'
            OR [ReportData] <> ''
        )
```

src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
```diff
@@ -27,7 +27,31 @@ public async Task<OrganizationReport> GetLatestByOrganizationIdAsync(Guid organi
...
+        {
+            var dbContext = GetDatabaseContext(scope);
+            // Substring match is coupled to SetReportFile (OrganizationReport.cs) writing via
+            // JsonHelpers.IgnoreWritingNull (PascalCase, no whitespace). The Dapper/MSSQL path
+            // uses JSON_VALUE which is format-agnostic; changing the serializer options would
+            // silently return zero rows here without affecting the MSSQL path.
+            var result = await dbContext.OrganizationReports
+                .Where(p => p.OrganizationId == organizationId
+                    && p.ReportFile != null
+                    && p.ReportFile.Contains("\"Validated\":true"))
```
Banrion: ⚠️ EF changes required to fix [regression](https://github.com/bitwarden/server/pull/7465/changes#r3336747805):

Requires updates to test cases


```suggestion
                .Where(p => p.OrganizationId == organizationId
                    && (
                        (p.ReportFile != null && p.ReportFile.Contains("\"Validated\":true"))
                        || p.ReportData != string.Empty
                    ))
```

Banrion (changes requested):

test/Api.Test/Dirt/OrganizationReportsControllerTests.cs
AlexRubik: If you literally mean to swap the test prefix from GetLatestOrganizationReportAsync_... to ReadLatestOrganizationReportAsync_... (and the section comment), then no. Test names in this file are named after the controller method being tested, not the inner method it calls. The test calls sutProvider.Sut.GetLatestOrganizationReportAsync(orgId), which is still the public route handler for GET /reports/organizations/{orgId}/latest; only what it calls internally changed (it now hits the  ReadLatestOrganizationReportAsync when V2 is on). Naming the test ReadLatestOrganizationReportAsync_... would imply there's a controller method by that name.

But I did rename it to GetLatestOrganizationReportAsync_**FlagOn**_WithValidatedFile_ReturnsOkWithDownloadUrl to follow the existing pattern in this file: writes (CreateOrganizationReportAsync tests) use _V1_/_V2_, reads (GetLatest... and Download... tests) use _FlagOn_/_FlagOff_. I know that's unconventional but it'll get cleaned up once the flag goes permanent. Also added a brief hint to the section comment so readers spot the feature flag in the naming.

Commit: https://github.com/bitwarden/server/pull/7465/commits/64cdaa652348dd8b9eeaef38372608539f852f77

src/Sql/dbo/Dirt/Stored Procedures/OrganizationReport_ReadLatestByOrganizationId.sql
AlexRubik: Good catch, thank you https://github.com/bitwarden/server/pull/7465/commits/9abee6f225b62d6cd0de09f17d86ff35d8c69473

util/Migrator/DbScripts/2026-05-19_00_AddOrganizationReportReadLatestSp.sql
AlexRubik: https://github.com/bitwarden/server/pull/7465/commits/9abee6f225b62d6cd0de09f17d86ff35d8c69473

src/Infrastructure.EntityFramework/Dirt/Repositories/OrganizationReportRepository.cs
AlexRubik: https://github.com/bitwarden/server/pull/7465/commits/9abee6f225b62d6cd0de09f17d86ff35d8c69473

test/Api.Test/Dirt/OrganizationReportsControllerTests.cs
```diff
@@ -1401,6 +1512,125 @@ await sutProvider.GetDependency<ICreateOrganizationReportCommand>()
...
+
+        sutProvider.GetDependency<IFeatureService>()
+            .IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2)
+            .Returns(false);
+
+        sutProvider.GetDependency<IGetOrganizationReportQuery>()
+            .GetOrganizationReportAsync(report.Id)
+            .Returns(report);
+
+        var stream = new MemoryStream(new byte[] { 1, 2, 3 });
```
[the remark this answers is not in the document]
AlexRubik: https://github.com/bitwarden/server/pull/7465/commits/addf17559e45c65c906739ab29ff4e7ec4dff421

test/Api.Test/Dirt/OrganizationReportsControllerTests.cs
```diff
@@ -1401,6 +1512,125 @@ await sutProvider.GetDependency<ICreateOrganizationReportCommand>()
...
+        var fileData = new ReportFile { Id = "file-id", FileName = "report.json", Size = 1024, Validated = true };
+        report.SetReportFile(fileData);
+
+        SetupV2Authorization(sutProvider, orgId);
+
+        sutProvider.GetDependency<IGetOrganizationReportQuery>()
+            .GetOrganizationReportAsync(report.Id)
+            .Returns(report);
+
+        var stream = new MemoryStream(new byte[] { 1, 2, 3 });
```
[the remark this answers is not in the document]
AlexRubik: https://github.com/bitwarden/server/pull/7465/commits/addf17559e45c65c906739ab29ff4e7ec4dff421

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -141,19 +141,18 @@ public async Task<IActionResult> GetLatestOrganizationReportAsync(Guid organizat
...
+            : await _getOrganizationReportQuery.GetLatestOrganizationReportAsync(organizationId);
 
         var response = new OrganizationReportResponseModel(latestReport);
 
-        if (_featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2))
+        if (isAccessIntelligenceV2)
         {
             var fileData = latestReport.GetReportFile();
-            if (fileData is { Validated: true })
+            if (fileData != null)
```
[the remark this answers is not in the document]
AlexRubik: https://github.com/bitwarden/server/pull/7465/commits/67a4b1238ffc4b83a0f6cf515970ab31d01df1a8

mkincaid-bw (approved):

Banrion (approved):

PR #7468, opened 2026-04-14
[PM-34828] Fix BW lite sqlite bulk import failure for org imports

mimartin12:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-34828

## 📔 Objective
Fixes a 500 error when logging in with an org account and when importing a vault on SQLite environments.

### Cause
`EventRepository.CreateManyAsync` and two cases of `CipherRepository.CreateAsync` call
  LinqToDB's BulkCopyAsync. When called against a SQLite context, LinqToDB auto-detects which SQLite
  adapter to use by checking for `System.Data.SQLite.dll` or `Microsoft.Data.Sqlite.dll` on disk. Since the
  server publishes as a single-file bundle (PublishSingleFile=true), managed DLLs are embedded in the
  executable and not present on disk ([example](https://github.com/bitwarden/server/blob/main/src/Identity/Dockerfile#L33)).

This surfaced after a change in the build process for the Bitwarden Lite container, which now pulls directly from our existing built containers. 
https://github.com/bitwarden/self-host/commit/08922d41f4f40be9432869b340444a3eb924367e

### Changes in this PR
Added IsSqlite() guards to three methods to route SQLite through EF Core's AddRangeAsync/SaveChangesAsync instead of LinqToDB's BulkCopyAsync:
- EventRepository.CreateManyAsync
- CipherRepository.CreateAsync(userId, ciphers, folders) — personal vault import with folders
- CipherRepository.CreateAsync(ciphers, collections, collectionCiphers, collectionUsers) — org vault import with collections

### DIFF
--- a/src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
+++ b/src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
@@ -48,6 +48,7 @@ public class EventRepository : Repository<Core.Entities.Event, Event, Guid>, IEv
             var tableEvents = entities.Select(e => e as Core.Entities.Event ?? new Core.Entities.Event(e));
             var entityEvents = Mapper.Map<List<Event>>(tableEvents);
             entityEvents.ForEach(e => e.SetNewId());
+            // SQLite does not support LinqToDB BulkCopy; use EF Core directly instead
             if (dbContext.Database.IsSqlite())
             {
                 await dbContext.AddRangeAsync(entityEvents);
--- a/src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
+++ b/src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
@@ -48,7 +48,15 @@ public class EventRepository : Repository<Core.Entities.Event, Event, Guid>, IEv
             var tableEvents = entities.Select(e => e as Core.Entities.Event ?? new Core.Entities.Event(e));
             var entityEvents = Mapper.Map<List<Event>>(tableEvents);
             entityEvents.ForEach(e => e.SetNewId());
-            await dbContext.BulkCopyAsync(entityEvents);
+            if (dbContext.Database.IsSqlite())
+            {
+                await dbContext.AddRangeAsync(entityEvents);
+                await dbContext.SaveChangesAsync();
+            }
+            else
+            {
+                await dbContext.BulkCopyAsync(entityEvents);
+            }
         }
     }
 
--- a/src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
+++ b/src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
@@ -57,6 +57,7 @@ public class EventRepository : Repository<Core.Entities.Event, Event, Guid>, IEv
             else
             {
                 await dbContext.BulkCopyAsync(entityEvents);
+                await dbContext.SaveChangesAsync();
             }
         }
     }
--- a/src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
+++ b/src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
@@ -160,6 +160,7 @@ public class CipherRepository : Repository<Core.Vault.Entities.Cipher, Cipher, G
             var dbContext = GetDatabaseContext(scope);
             var folderEntities = Mapper.Map<List<Folder>>(folders);
             var cipherEntities = Mapper.Map<List<Cipher>>(ciphers);
+            // SQLite does not support LinqToDB BulkCopy; use EF Core directly instead
             if (dbContext.Database.IsSqlite())
             {
                 await dbContext.AddRangeAsync(folderEntities);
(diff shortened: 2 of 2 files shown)


jengstrom-bw (dismissed):

src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
```diff
@@ -48,7 +48,16 @@ public async Task CreateManyAsync(IEnumerable<IEvent> entities)
...
+            // SQLite does not support LinqToDB BulkCopy; use EF Core directly instead
+            if (dbContext.Database.IsSqlite())
+            {
+                await dbContext.AddRangeAsync(entityEvents);
+                await dbContext.SaveChangesAsync();
+            }
+            else
+            {
+                await dbContext.BulkCopyAsync(entityEvents);
+            }
```
Banrion, line 61: ⚠️  Else statement misses `await dbContext.SaveChangesAsync();`

Banrion (changes requested):

src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
mimartin12: Done in https://github.com/bitwarden/server/pull/7468/commits/0c7941e529df11c6948638ea87754f8ad5b4031c

Banrion (approved):

justindbaur (changes requested):
I'm cautious about this change, there could be many other places that are broken and it forces us to do more and more branching when the goal is that we don't have to branch for EF unless absolutely needed. 

I wonder if we could configure the provider explicitly like this:

```
diff --git a/src/Infrastructure.EntityFramework/EntityFrameworkServiceCollectionExtensions.cs b/src/Infrastructure.EntityFramework/EntityFrameworkServiceCollectionExtensions.cs
index e2227167a6..af66106e81 100644
--- a/src/Infrastructure.EntityFramework/EntityFrameworkServiceCollectionExtensions.cs
+++ b/src/Infrastructure.EntityFramework/EntityFrameworkServiceCollectionExtensions.cs
@@ -1,4 +1,5 @@
-﻿using Bit.Core.AdminConsole.Repositories;
+﻿using System.Diagnostics;
+using Bit.Core.AdminConsole.Repositories;
 using Bit.Core.Auth.Repositories;
 using Bit.Core.Billing.Organizations.Repositories;
 using Bit.Core.Billing.Providers.Repositories;
@@ -25,6 +26,7 @@ using Bit.Infrastructure.EntityFramework.Repositories;
 using Bit.Infrastructure.EntityFramework.SecretsManager.Repositories;
 using Bit.Infrastructure.EntityFramework.Tools.Repositories;
 using Bit.Infrastructure.EntityFramework.Vault.Repositories;
+using LinqToDB;
 using LinqToDB.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
@@ -40,9 +42,6 @@ public static class EntityFrameworkServiceCollectionExtensions
             throw new Exception($"Database provider type {provider} was selected but no connection string was found.");
         }
 
-        // TODO: We should move away from using LINQ syntax for EF (TDL-48).
-        LinqToDBForEFTools.Initialize();
-
         services.AddAutoMapper(typeof(UserRepository));
         services.AddDbContext<DatabaseContext>(options =>
         {
@@ -65,6 +64,35 @@ public static class EntityFrameworkServiceCollectionExtensions
             {
                 options.UseSqlServer(connectionString);
             }
+
+            // Configure LinqToDB with the types of drivers we use for each DB type, this allows to us avoid it searching
+            // for assemblies and gives us the ability to publish as a single file.
+            options.UseLinqToDB(linq =>
+            {
+                linq.AddCustomOptions(linqOptions =>
+                {
+                    if (provider == SupportedDatabaseProviders.Postgres)
+                    {
+                        return linqOptions.UsePostgreSQL();
+                    }
+                    else if (provider == SupportedDatabaseProviders.MySql)
+                    {
+                        return linqOptions.UseMySqlConnector();
+                    }
+                    else if (provider == SupportedDatabaseProviders.Sqlite)
+                    {
+                        return linqOptions.UseSQLiteMicrosoft();
+                    }
+                    else if (provider == SupportedDatabaseProviders.SqlServer)
+                    {
+                        return linqOptions.UseSqlServer(provider: LinqToDB.DataProvider.SqlServer.SqlServerProvider.MicrosoftDataSqlClient);
+                    }
+
+                    Debug.Fail("Unknown database provider.");
+
+                    return linqOptions;
+                });
+            });
         });
     }
```

Hopefully by manually telling it the provider we avoid having to have it search for it. If this doesn't work and we need to start branching then we especially need to make sure that these differences are now covered by tests.

jussikuosa:
Hello @justindbaur, I ran into this bug yesterday evening and couldn't import even a single password from apple passwords on iOS, nor did the 1password 1pif import work on macOS.

Looking into the issue, it seems upstream has already fixed this in https://github.com/linq2db/linq2db/pull/5489 for all providers. Updating linq2db to 6.5.0 would also require bumping linq2db.EntityFrameworkCore to 8.0.8.

Would this be a more acceptable fix? (I can try it out if yes)

PR #7469, opened 2026-04-14
[PM-31631] update password pre-login salt response

ike-kottlowski:
## 🎟️ Tracking

[PM-31631](https://bitwarden.atlassian.net/browse/PM-31631)

## 📔 Objective

With the addition of the salt property used to calculate hash values for passwords we need to add `salt` to the response object for password/prelogin. Otherwise the client won't know which algorithm and salt to use to when validating for password authentication.

### Insomnia API Tooling showing request getting both email and null responses from `prelogin/password` endpoint
https://github.com/user-attachments/assets/d1d20f45-cd72-4b32-8aaa-841899fd81bb

[PM-31631]: https://bitwarden.atlassian.net/browse/PM-31631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Identity/Controllers/AccountsController.cs
```diff
@@ -251,21 +236,23 @@ public WebAuthnLoginAssertionOptionsResponseModel GetWebAuthnLoginAssertionOptio
 
     private UserKdfInformation GetDefaultKdf(string email)
     {
-        if (_defaultKdfHmacKey == null)
+        var normalizedEmail = email.Trim().ToLowerInvariant();
```
JaredSnider-Bitwarden, line 240: Can we put a comment here to enshrine that the normalization is very important? 
`// Normalize before use so casing differences in the request do not affect the response.`

JaredSnider-Bitwarden (approved):
Excellent work.

PR #7476, opened 2026-04-15
[PM-32073] - Added Bulk Get Org Ability

jrmccannon:
## 🎟️ Tracking
[PM-32073](https://bitwarden.atlassian.net/browse/PM-32073)

## 📔 Objective
This adds the bulk get implementation of the org ability from the extended cache (FusionCache). It also cleans up some test files.

[PM-32073]: https://bitwarden.atlassian.net/browse/PM-32073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JaredScar (dismissed):
Looks good to me, nice work fellow Jared

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
JimmyVo16: Nit (non-blocker): There will be a provider cache here, so it’ll probably make more sense to add “org” to this cache name. That said, we have a ticket to remove this whole class in the near future, and it’s only a local variable, so I’ll leave it up to you.

JimmyVo16 (commented):
The changes for the multiple cases look good.

Side note: depending on the outcomes of [1](https://github.com/bitwarden/server/pull/7447#discussion_r3094748978) and [2](https://github.com/bitwarden/server/pull/7447#discussion_r3090204694), we should consider adding them to this PR since it’s touching the same area.

JaredScar (dismissed):

JimmyVo16 (dismissed):

JaredScar (approved):

PR #7477, opened 2026-04-15
[PM-34387] Add organization invite link creation endpoint

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34387

## 📔 Objective

Adds a new `POST organizations/{orgId}/invite-link` endpoint that lets organization admins create a shareable invite link scoped to specific email domains.

The endpoint is gated behind the generate invite link feature flag and requires the `ManageUsers` permission like the existing invite users endpoint


src/Api/AdminConsole/Models/Request/Organizations/CreateOrganizationInviteLinkRequestModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Bit.Api.AdminConsole.Models.Request.Organizations;
+
+public class CreateOrganizationInviteLinkRequestModel
+{
+    /// <summary>
+    /// Email domains permitted to accept the invite link (e.g. <c>["acme.com"]</c>).
+    /// </summary>
+    [Required]
+    public IEnumerable<string> AllowedDomains { get; set; } = null!;
```
eliykat: Can we use `required` here as well, or do we have to use `= null!` to satisfy the ASP.NET model binding?

src/Core/AdminConsole/Entities/OrganizationInviteLink.cs
```diff
@@ -6,7 +6,7 @@ namespace Bit.Core.AdminConsole.Entities;
 public class OrganizationInviteLink : ITableObject<Guid>
 {
     public Guid Id { get; set; }
-    public Guid Code { get; set; }
+    public Guid Code { get; set; } = Guid.NewGuid();
```
eliykat, line 16: Should use `CoreHelpers.GenerateComb()`

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        return inviteLink;
+    }
+
+    private static List<string> SanitizeDomains(IEnumerable<string>? domains) =>
+        domains?
+            .Select(d => d?.Trim())
+            .Where(d => !string.IsNullOrEmpty(d))
+            .Cast<string>()
+            .ToList() ?? [];
```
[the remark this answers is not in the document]
eliykat: I agree with [this feedback](https://github.com/bitwarden/server/pull/7477/changes#r3087988156) that some basic validation should be done here. Maybe check how we handle domains elsewhere (e.g. claimed domains, email address validation).

src/Api/AdminConsole/Models/Request/Organizations/CreateOrganizationInviteLinkRequestModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// The invite key encrypted with the organization key.
+    /// </summary>
+    [Required]
```
eliykat, line 20: Is this an EncString? If so, also use the `[EncryptedString]` attribute.

src/Api/AdminConsole/Models/Response/Organizations/OrganizationInviteLinkResponseModel.cs
```diff
@@ -0,0 +1,38 @@
...
+        : base("organizationInviteLink")
+    {
+        ArgumentNullException.ThrowIfNull(inviteLink);
+
+        Id = inviteLink.Id;
+        Code = inviteLink.Code;
+        OrganizationId = inviteLink.OrganizationId;
+        if (!string.IsNullOrWhiteSpace(inviteLink.AllowedDomains))
+        {
+            AllowedDomains = JsonSerializer.Deserialize<IEnumerable<string>>(inviteLink.AllowedDomains)
```
eliykat: Suggestion: move this into a getter on the Entity model. We do this elsewhere, e.g. `Policy.GetDataModel`.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
eliykat, line 1: Just double checking - is this feature limited to any particular plan type? (e.g. enterprise?)

eliykat (changes requested):

src/Core/AdminConsole/Entities/OrganizationInviteLink.cs
r-tome: I don't think we should because `Code` is not a table identifier so we don't need to optimize for indexing. Also Comb GUIDs include the timestamp of creation and that makes it a bit predictable.

src/Api/AdminConsole/Models/Request/Organizations/CreateOrganizationInviteLinkRequestModel.cs
r-tome: Yes, great suggestion!

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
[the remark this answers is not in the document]
r-tome: Good shout, `DomainNameValidatorAttribute` already exists so I'm using that.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return new InviteLinkEncryptedKeyRequired();
+        }
+
+        var existingLink = await organizationInviteLinkRepository.GetByOrganizationIdAsync(request.OrganizationId);
+        if (existingLink != null)
+        {
+            return new InviteLinkAlreadyExists();
+        }
+
```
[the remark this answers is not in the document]
r-tome: I consider this scenario extremely unlikely so its not a concern. @eliykat let me know if you agree with the robot

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
r-tome: 🤯 I never thought of that. Yes, Enterprise only. I'll add another task to add that feature and merge it before this one.

src/Core/AdminConsole/Entities/OrganizationInviteLink.cs
eliykat: I forgot that, but you're right! can you please add a comment?

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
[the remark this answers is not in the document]
eliykat: Agree, not concerned about this

test/Api.Test/AdminConsole/Models/Request/Organizations/CreateOrganizationInviteLinkRequestModelTests.cs
eliykat, line 1: Not blocking, but my 2 cents: these tests are really testing the `EncryptedStringAttribute`, which is separately tested. We can see it's tested and working, we can see it's been applied here - I don't think an extra layer of tests is adding anything.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
```diff
@@ -11,19 +11,21 @@ public class CreateOrganizationInviteLinkCommand(
     TimeProvider timeProvider)
     : ICreateOrganizationInviteLinkCommand
 {
+    private static readonly DomainNameValidatorAttribute _domainValidator = new();
```
eliykat: This is a data validation attribute designed to be applied to request models. e.g. `[DomainNameValidator]`. I don't think we should call it manually here. Either:
- put it on the api request model so the api handles validation, or
- extract its logic to a static class that can be called in different contexts

We have a bit of an inconsistency in where/how we do validation. I think this is OK to do at the api layer as it's simple validation on the input value rather than more complex business rules usually checked by the core layer.

src/Core/AdminConsole/Entities/OrganizationInviteLink.cs
```diff
@@ -14,6 +15,13 @@ public class OrganizationInviteLink : ITableObject<Guid>
     public DateTime CreationDate { get; set; } = DateTime.UtcNow;
     public DateTime RevisionDate { get; set; } = DateTime.UtcNow;
 
+    public IEnumerable<string> GetAllowedDomains() =>
+        JsonSerializer.Deserialize<IEnumerable<string>>(AllowedDomains)
+        ?? throw new JsonException("Failed to deserialize AllowedDomains.");
```
eliykat: I don't think this exception is correct - `Deserialize` will throw its own exception if deserialization fails. Your exception will be thrown if it returns `null` - so really it's a null check.

In that case, throwing on a getter is not a good way to handle nulls. Generally we want to enforce non-null values going into the database but gracefully handle them coming out. In this case you could probably just return an empty sequence; it won't let anyone use the link (because there are no allowed domains) but that is relatively harmless and better than throwing.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
r-tome: I have moved the logic over to `CreateOrganizationInviteLinkRequestModel.Validate` method. I couldn't use the attribute simply because its a list of domains.

r-tome:
The integration test is failing because the Pricing service hasn't been deployed on QA yet.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
eliykat: (non-blocking) 💭 You could create a generic validator that takes another validator and applies it to each item in a list. e.g. `ValidateSequence<DomainNameValidatorAttribute>`. Your solution is also valid.

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationInviteLinksControllerTests.cs
```diff
@@ -35,6 +36,12 @@ public OrganizationInviteLinksControllerTests(ApiApplicationFactory factory)
                 .IsEnabled(FeatureFlagKeys.GenerateInviteLink)
                 .Returns(true);
         });
+        _factory.SubstituteService<IApplicationCacheService>(cacheService =>
+        {
+            cacheService
+                .GetOrganizationAbilityAsync(Arg.Any<Guid>())
+                .Returns(new OrganizationAbility { UseInviteLinks = true });
+        });
```
eliykat, line 44: I'm interested why this is necessary - the signup helper should upsert the new org into the cache. We don't want to have to mock this behavior for every single integration test that may check the cache. Can you please look into this?

eliykat (commented):
LGTM once the base PR is merged, aside from 1 question about the test.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/CreateOrganizationInviteLinkCommand.cs
r-tome: That is a great idea! Gave it a go

test/Api.IntegrationTest/AdminConsole/Controllers/OrganizationInviteLinksControllerTests.cs
r-tome: https://github.com/bitwarden/server/pull/7477#issuecomment-4298207639

I think this is the reason. QA pricing service hasn't updated yet so it does not return `UseInviteLinks` therefore it defaults to false. I just mocked it so we wouldn't depend on QA env.

src/Core/Utilities/ValidateSequenceAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+/// Validates each string element of a collection using <typeparamref name="TValidator"/>, reporting
+/// all invalid items in a single <see cref="ValidationResult"/>.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter)]
+public class ValidateSequenceAttribute<TValidator> : ValidationAttribute
+    where TValidator : ValidationAttribute, new()
+{
+    protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
+    {
+        var items = (value as IEnumerable<string> ?? []).ToList();
```
eliykat: If the type cast fails, we want to fail loudly, because we can't properly validate the value. This validates an empty list instead, which is not the same.

src/Core/Utilities/ValidateSequenceAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+        var validator = new TValidator();
+        var invalid = items.Where(item => !validator.IsValid(item)).ToList();
+
+        if (invalid.Count == 0) return ValidationResult.Success;
```
eliykat: No single line `if` statements

src/Core/Utilities/ValidateSequenceAttribute.cs
eliykat: Is there any need to limit it to `IEnumerable<string>`? Could we do `IEnumerable<object>`?

eliykat (changes requested):

src/Core/Utilities/ValidateSequenceAttribute.cs
r-tome: Not really. I thought to keep it simple but it shouldn't break.

eliykat (dismissed):

src/Core/Utilities/ValidateSequenceAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+    private const string _invalidItemsMessage = "The following items are not valid: {0}";
+
+    protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
+    {
+        if (value is null)
+        {
+            return ValidationResult.Success;
+        }
+
+        var items = ((IEnumerable)value).Cast<object>().ToList();
```
eliykat: This will throw, but maybe not with a helpful error message. I was thinking more like:

```suggestion
        if (value is not IEnumerable<object> items)
        {
            throw new  ArgumentException("ValidateSequenceAttribute can only be used with an IEnumerable");
        }
```

src/Core/Utilities/ValidateSequenceAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+public class ValidateSequenceAttribute<TValidator> : ValidationAttribute
+    where TValidator : ValidationAttribute, new()
+{
+    private const string _invalidItemsMessage = "The following items are not valid: {0}";
+
+    protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
+    {
+        if (value is null)
+        {
+            return ValidationResult.Success;
```
eliykat, line 21: It's not clear whether an empty list should pass validation or not. In our case it's invalid because the user is meant to supply at least 1 domain name, but it really depends on the business logic.

I think your approach is fine because it matches the behaviour of LINQ `.All()`. But we should mention it in the xmldoc. e.g. `An empty list will always pass validation. Add additional validation if an empty list is invalid.` This is our situation because your command checks it, and/or you can use `minLength`. It just needs to be made clear to callers.

r-tome: Clarified that in the xmldoc

eliykat (approved):

PR #7478, opened 2026-04-15
[PM-25056] - Deadlock testing fix

jrmccannon:
## 🎟️ Tracking
[PM-25056](https://bitwarden.atlassian.net/browse/PM-25056)

## 📔 Objective
This identified the root cause for the testing framework deadlock. 

When tests are being run in parallel, tests that called `OrganizationRepo.DeleteAsync` would call the giant sproc that would go across all the other tables to clean up related data. Each opened up its own transaction and would lock tables. This caused the intermittent deadlocks across the different tests as they were all locking and trying to delete records in the same tables.

[PM-25056]: https://bitwarden.atlassian.net/browse/PM-25056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/Infrastructure.IntegrationTest/AdminConsole/OrganizationTestHelpers.cs
+++ b/test/Infrastructure.IntegrationTest/AdminConsole/OrganizationTestHelpers.cs
@@ -4,6 +4,8 @@ using Bit.Core.Billing.Enums;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
 using Bit.Core.Repositories;
+using Bit.Core.Utilities;
+using Microsoft.Data.SqlClient;
 
 namespace Bit.Infrastructure.IntegrationTest.AdminConsole;
 
@@ -16,7 +18,7 @@ public static class OrganizationTestHelpers
 {
     public static Task<User> CreateTestUserAsync(this IUserRepository userRepository, string identifier = "test")
     {
-        var id = Guid.NewGuid();
+        var id = CoreHelpers.GenerateComb();
         return userRepository.CreateAsync(new User
         {
             Id = id,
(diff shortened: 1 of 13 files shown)


sven-bitwarden:
https://kevsoft.net/2025/05/14/banning-api-calls-in-dotnet-without-writing-your-own-analyzer.html

Could we add the banned API analyzer to our integration tests, so that we get an immediate red flag warning if we try to use the method that causes deadlocks? It's a Microsoft package, but not sure if there's some difficulties with adding it. We would also need to just bypass the warnings for the test files that are **trying** to test this method.

shane-melton (dismissed):

jrmccannon:
>Could we add the banned API analyzer to our integration tests, so that we get an immediate red flag warning if we try to use the method that causes deadlocks? It's a Microsoft package, but not sure if there's some difficulties with adding it. We would also need to just bypass the warnings for the test files that are trying to test this method.

That could definitely be something added later on. We'd have to review the package and document the tool for potentially more future uses. @sven-bitwarden

sven-bitwarden (dismissed):

eliykat (changes requested):
As discussed, most of these calls are because we delete our test data after each test. However, test databases should be treated as ephemeral; cleaning up our own test data just makes tests slower and competes for locks.

Please delete all test cleanup logic instead. If we're still having deadlock issues, that should ideally be addressed at a more general level (e.g. how testing framework or the db is configured).

sven-bitwarden (approved):

eliykat (approved):
Thanks!

shane-melton (approved):

PR #7482, opened 2026-04-16
[PM-35243] Exclude 'other' cancel survey reason for Teams/Enterprise

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35243

## 📔 Objective

Adds plan-aware validation to `SubscriberService.CancelSubscription` so that "other" is excluded from valid cancellation survey reasons when the subscriber is a Teams, Enterprise, or Teams Starter organization. This is defense-in-depth for PM-35243, which updates the cancel survey options shown to business-tier customers. The existing behavior — invalid reasons silently become `null` feedback to Stripe — is preserved. Uses the existing `IsBusinessProductTierType()` extension to identify business plans.

## 📸 Screenshots

N/A — no UI changes in this repo.

### DIFF
--- a/src/Core/Billing/Services/Implementations/SubscriberService.cs
+++ b/src/Core/Billing/Services/Implementations/SubscriberService.cs
@@ -66,6 +66,12 @@ public class SubscriberService(
             "unused"
         ];
 
+        if (subscriber is Organization { PlanType: var planType } &&
+            planType.IsBusinessProductTierType())
+        {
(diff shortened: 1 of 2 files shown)


amorask-bitwarden:
Closing — the updated requirements for PM-35243 keep the "Other" option for business plans, so the server-side removal of "other" from valid cancellation reasons is no longer needed.

PR #7486, opened 2026-04-16
[PM-35234] Prevent appending duplicate org user in validator request

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-35234

## 📔 Objective
In the accept flow, we query for all possible org users related to the user accepting the invite (to enforce single org requirement) when auto confirm is enabled. However, at this stage an SSO user will have an org user already provisioned, so the query returns our user accepting the invite. The validator logic does not recognize this is the same user for the same organization and erroneously fails the single organization requirement validation step. 

This PR fixes that by ensuring the orgUser is not appended to the request if it is the same org user as the accepting user.

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptOrgUserCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptOrgUserCommand.cs
@@ -232,11 +232,15 @@ public class AcceptOrgUserCommand : IAcceptOrgUserCommand
         var policyRequirement = await _policyRequirementQuery.GetAsync<AutomaticUserConfirmationPolicyRequirement>(
             user.Id);
 
+        // Ensure that the user is not already in the list of organization users
+        if (!allOrgUsers.Any(x => x.OrganizationId == orgUser.OrganizationId && x.UserId == user.Id))
+        {
+            allOrgUsers.Add(orgUser);
+        }
+
         var error = (await _automaticUserConfirmationPolicyEnforcementValidator.IsCompliantAsync(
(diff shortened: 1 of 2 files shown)


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptOrgUserCommand.cs
```diff
@@ -234,7 +234,9 @@ private async Task HandleAutomaticUserConfirmationPolicyAsync(OrganizationUser o
 
         var error = (await _automaticUserConfirmationPolicyEnforcementValidator.IsCompliantAsync(
                 new AutomaticUserConfirmationPolicyEnforcementRequest(orgUser.OrganizationId,
-                    allOrgUsers.Append(orgUser),
+                    allOrgUsers
+                        .Where(u => u.OrganizationId != orgUser.OrganizationId && u.UserId != orgUser.UserId)
```
jrmccannon: Let's just pull this out to and Add it to the list only if it doesn't exist.

```
        // Ensure that the user is not already in the list of organization users
        if (!allOrgUsers.Any(x => x.OrganizationId == orgUser.OrganizationId && x.UserId == user.Id))
        {
            allOrgUsers.Add(orgUser);
        }

        var error = (await _automaticUserConfirmationPolicyEnforcementValidator.IsCompliantAsync(
                new AutomaticUserConfirmationPolicyEnforcementRequest(orgUser.OrganizationId,
                    allOrgUsers,
                    user),
                policyRequirement))
            .Match(
                error => error.Message,
                _ => string.Empty
            );
```

jrmccannon (changes requested):

BTreston:
@claude sanity check

jrmccannon (approved):

PR #7490, opened 2026-04-17
Add seed script for local development

Hinton:
## 📔 Objective

 Introduces a repeatable, declarative way to seed a local dev environment via the existing `SeederUtility`.                                                                                                        
   
  - **`dev/seed.ps1`** — new script that reads `dev/seeds.json` (and optional, gitignored `dev/seeds.local.json` for per-developer overrides), converts each entry's `args` from camelCase to kebab-case CLI flags, 
  and invokes `dotnet run --project util/SeederUtility` once per seed. Supports `-DryRun` to preview the commands without executing. Fails fast on non-zero exit codes.
  - **`dev/seeds.json`** — default seed set (two individual users + the `qa.enterprise-basic` preset) so new contributors get a consistent baseline dataset.                                                        
  - **`dev/.gitignore`** — ignores `seeds.local.json` so per-developer overrides don't leak into the repo.                                                                                                          
  - **`IndividualUserOptions` / `IndividualArgs` / `RecipeOrchestrator`** — adds an optional `--email` flag to the `individual` command, falling back to the existing `firstName.lastName@individual.example`       
  pattern when omitted. This lets `seeds.json` pin deterministic emails (e.g. `jane@bitwarden.pw`) instead of relying on the generated default.

### DIFF
--- a/dev/.gitignore
+++ b/dev/.gitignore
@@ -19,3 +19,5 @@ signingkey.jwk
 # Reverse Proxy Conifg
 reverse-proxy.conf
 *.crt
+
+seeds.local.json
--- a/dev/seed.ps1
+++ b/dev/seed.ps1
@@ -1,0 +1,77 @@
+#!/usr/bin/env pwsh
+# Runs SeederUtility for each entry defined in seeds.json (and optionally seeds.local.json).
+# Usage: ./seed.ps1 [-DryRun]
+
+param(
+    [switch]$DryRun
+)
+
+$ErrorActionPreference = "Stop"
+
+$globalSeedsPath = Join-Path $PSScriptRoot "seeds.json"
+$localSeedsPath  = Join-Path $PSScriptRoot "seeds.local.json"
+$seederProject   = Join-Path $PSScriptRoot ".." "util" "SeederUtility"
+
+$seeds = @(Get-Content $globalSeedsPath -Raw | ConvertFrom-Json)
+
+if (Test-Path $localSeedsPath) {
+    $seeds += @(Get-Content $localSeedsPath -Raw | ConvertFrom-Json)
+}
+
+function ConvertTo-KebabCase {
+    param([string]$Name)
+    return ($Name -creplace '(?<=[a-z])([A-Z])', '-$1').ToLower()
+}
+
+function Build-CliArgs {
+    param($ArgsObject)
+    $parts = @()
+    foreach ($prop in $ArgsObject.PSObject.Properties) {
+        $flag  = ConvertTo-KebabCase $prop.Name
+        $value = $prop.Value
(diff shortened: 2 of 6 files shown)


dev/seeds.json
theMickster, line 1: ❌ The example below shouldn't encourage the use of a real domain.
Will you please change them to either the coded domain (@individual.example) or the IETF-reserved example.test?

dev/seed.ps1
```diff
@@ -0,0 +1,77 @@
...
+    $command = $seed.command
+    $cliArgs = Build-CliArgs $seed.args
+
+    Write-Host ""
+    Write-Host "[$($i + 1)/$total] $label" -ForegroundColor Cyan
+    Write-Host "  dotnet run --project $seederProject -- $command $($cliArgs -join ' ')" -ForegroundColor DarkGray
+
+    if (-not $DryRun) {
+        dotnet run --project $seederProject -- $command @cliArgs
+        if ($LASTEXITCODE -ne 0) {
```
theMickster, line 65: ♻️ 🤔 Given that we have `$ErrorActionPreference = "Stop"`, I am fairly certain that `Write-Error` on line 66 exits the program and the exit on line 67 is moot. 

If the intent is to propagate the error number up to the caller then I think the below will work better. 

```suggestion
  if ($LASTEXITCODE -ne 0) {
      $code = $LASTEXITCODE
      Write-Host "Seed '$label' failed with exit code $code." -ForegroundColor Red
      exit $code
  }
```

dev/.gitignore
```diff
@@ -19,3 +19,5 @@ signingkey.jwk
 # Reverse Proxy Conifg
 reverse-proxy.conf
 *.crt
+
+seeds.local.json
```
theMickster, line 23: ⛏️ I have one minor gripe about the addition of the `seeds.local.json` and that it actively encourages engineers to make their own setups instead of contributing to our shared `seeds.json`. My preference is that we actively work together in the spirit of the **Unified Test Data** idea to avoid these personal setups in favor  of an engineering standard setup that we evolve together as teams/individuals come up with more use-cases and presets.

theMickster (changes requested):
The domain name in the seeds.json is the only thing that I think must be fixed.

dev/.gitignore
Hinton: There will always be team specific and/or personal preferences for how things should work. This gives you a good way to configuring that. A future expansion might be some way to define your team and those could be checked in.

theMickster (approved):

theMickster:
@Hinton Will you please take a peek at the workflow failures? 
https://github.com/bitwarden/server/pull/7490

PR #7495, opened 2026-04-17
Add -o --output parameters to DB seeder util for preset command

mimartin12:
## 📔 Objective
When working with some automation and the DB seeder utility around the `preset` command, I found it was hard to get machine-readable output in the existing format while using the preset command. This adds a new parameter for output, used with `-o`, `--output` with a new args `json`, `text` (default). 

Using `if (!Console.IsOutputRedirected)` in conjunction with a JSON output lets us pipe directly into tools like `jq` to grab what we need.

For example:
```
# Get all org presets
dotnet run preset --list -o json | jq -r '.organization[]'

# Get all individual presets
dotnet run preset --list -o json | jq -r '.individual[]'
```


util/SeederUtility/Program.cs
```diff
@@ -7,7 +7,11 @@ public class Program
 {
     private static int Main(string[] args)
     {
-        PrintBanner();
+        // Skip the banner when stdout is piped or redirected
```
theMickster: ❌ Let's avoid doing this if condition as it's a code smell and feels like a hack. 
I am fairly certain that `jq` only reads from the stdout and not from the stderr stream. Therefore, in the places PrintBanner() method where we call `Console.WriteLine()` we would change those to `Console.Error.WriteLine` giving interactive users the banner while programatic users avoid it.

util/SeederUtility/Commands/PresetArgs.cs
```diff
@@ -14,6 +20,14 @@ public class PresetArgs : IArgumentModel
     [Option('l', "list", Description = "List all available presets and fixtures")]
     public bool List { get; set; }
 
+    [Option('o', "output", Description = "Output format for --list: text or json (default: text)")]
```
theMickster: ♻️ The letter o has too many domain meanings (organization being foremost); let's take it out and just go with the full name please. 

```suggestion
    [Option("output", Description = "Output format for --list: text or json (default: text)")]
```

util/SeederUtility/Commands/PresetArgs.cs
```diff
@@ -2,6 +2,12 @@
 
 namespace Bit.SeederUtility.Commands;
 
+public enum OutputFormat
```
theMickster: ⚠️ Please move this to it's own file.

theMickster (changes requested):
Couple things to change and one point of discussion to clarify

util/SeederUtility/Commands/PresetArgs.cs
theMickster: Addressed in c7f7ac8befbf562a971c946e6231fe334180c907

util/SeederUtility/Program.cs
mimartin12: Moved out to `Console.Error.WriteLine` as discussed.

theMickster (approved):

PR #7500, opened 2026-04-17
[PM-33417] WebAuthn cache

ike-kottlowski:
## 🎟️ Tracking

[PM-33417](https://bitwarden.atlassian.net/browse/PM-33417)

## 📔 Objective

[PM-33417]: https://bitwarden.atlassian.net/browse/PM-33417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


enmande (dismissed):

src/Identity/IdentityServer/RequestValidators/WebAuthnGrantValidator.cs
```diff
@@ -1,7 +1,4 @@
-﻿// FIXME: Update this file to be null safe and then delete the line below
-#nullable disable
```
enmande, line 2: 🙏 Thank you for getting the `nullable disable`.

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/WebAuthnChallengeCacheProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Bit.Core.Auth.UserFeatures.WebAuthnLogin.Implementations;
+
+internal class WebAuthnChallengeCacheProvider(
+    [FromKeyedServices("persistent")] IDistributedCache distributedCache) : IWebAuthnChallengeCacheProvider
+{
+    private const string _cacheKeyPrefix = "WebAuthnAssertion_";
+    private static readonly DistributedCacheEntryOptions _cacheOptions = new()
+    {
+        AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(17)
```
enmande: ❓ Is this from the lifetime for [WebAuthnLoginAssertionOptionsTokenable](https://github.com/bitwarden/server/blob/a18b60014765ef717621c628c4868c900604872c/src/Core/Auth/Models/Business/Tokenables/WebAuthnLoginAssertionOptionsTokenable.cs#L17)? If so, does it make sense to get that value from the tokenable (e.g., `static` field?).
Recognizing the case for that value changing may be low, perhaps a note of documentation if that's too entangled, so we show why we're choosing this value, and/or something to keep us from drifting if one were to change?

enmande (commented):
Just one question, below. Thank you!

src/Core/Auth/UserFeatures/WebAuthnLogin/Implementations/WebAuthnChallengeCacheProvider.cs
ike-kottlowski: Addressed in [this commit](https://github.com/bitwarden/server/pull/7500/commits/1ee3f6056ea492200afab58e2162bc7253147ae2)

enmande (approved):
✨ thank you!

PR #7505, opened 2026-04-20
[PM-35306] Fix password change not working when using the unlock and authentication data models

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35306

## 📔 Objective

Currently, the KDF request model validates the KDF parameters to be in the acceptable range for KDF changes and new accounts. However, this validation is re-used during password changes, which is incorrect. This prevents users from changing their password until they upgrade their KDF settings.

This PR removes the validation on the request model directly. Validation still happens in the commands that should validate (registration, change KDF).

### DIFF
--- a/src/Core/KeyManagement/Models/Api/Request/KdfRequestModel.cs
+++ b/src/Core/KeyManagement/Models/Api/Request/KdfRequestModel.cs
@@ -1,11 +1,10 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using Bit.Core.Enums;
 using Bit.Core.KeyManagement.Models.Data;
-using Bit.Core.Utilities;
 
 namespace Bit.Core.KeyManagement.Models.Api.Request;
 
-public class KdfRequestModel : IValidatableObject
+public class KdfRequestModel
 {
     [Required]
     public required KdfType KdfType { get; init; }
@@ -23,11 +22,5 @@ public class KdfRequestModel : IValidatableObject
             Memory = Memory,
             Parallelism = Parallelism
         };
-    }
-
-    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
-    {
-        // Generic per-request KDF validation for any request model embedding KdfRequestModel
-        return KdfSettingsValidator.Validate(ToData());
     }
 }
--- a/src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
+++ b/src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
@@ -25,4 +25,17 @@ public class MasterPasswordAuthenticationDataRequestModel
             Salt = Salt
         };
     }
+
+    public bool HasSameKdfConfiguration(MasterPasswordUnlockDataRequestModel unlockData)
+    {
+        if (unlockData == null)
+        {
+            return false;
+        }
+
+        return Kdf.KdfType == unlockData.Kdf.KdfType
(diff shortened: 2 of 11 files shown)


src/Core/KeyManagement/Models/Api/Request/KdfRequestModel.cs
```diff
@@ -24,10 +23,4 @@ public KdfSettings ToData()
             Parallelism = Parallelism
         };
     }
-
-    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
```
quexten, line 28: The validation happens here:
- https://github.com/bitwarden/server/blob/cd4e44cb04f4b80ef035df80df7ff5991cd0e64b/src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs#L110
- https://github.com/bitwarden/server/blob/cd4e44cb04f4b80ef035df80df7ff5991cd0e64b/src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs#L162
- https://github.com/bitwarden/server/blob/cd4e44cb04f4b80ef035df80df7ff5991cd0e64b/src/Core/KeyManagement/Kdf/Implementations/ChangeKdfCommand.cs#L57

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
JaredSnider-Bitwarden: This test needs to be changed to reflect the new  KDF change validation error: 
```suggestion
            Assert.Contains("AuthenticationData and UnlockData must have the same salt.", content);
```

JaredSnider-Bitwarden: This test assertion needs to be changed based on the new KDF validation error:
```suggestion
            Assert.Contains("AuthenticationData and UnlockData must have the same KDF configuration.", content);

```

JaredSnider-Bitwarden (changes requested):
Auth team reviewed synchronously and we are happy w/ the changes, but there are two test tweaks that need to be fixed below:

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
JaredSnider-Bitwarden, line 1: This test also needs to be fixed: `PostKdf_ChangedSaltInAuthenticationData_BadRequest`

quexten:
@mzieniukbw Re: Our conversation in the dev sync. The data models already have a "Validate[salt/kdf]UnchagedForUser". It's just the request models that don't. I don't think we need to introduce anything extra in this PR.

src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
```diff
@@ -25,4 +25,17 @@ public MasterPasswordAuthenticationData ToData()
...
+    {
+        if (unlockData == null)
+        {
+            return false;
+        }
+
+        return Kdf.KdfType == unlockData.Kdf.KdfType
+            && Kdf.Iterations == unlockData.Kdf.Iterations
+            && Kdf.Memory == unlockData.Kdf.Memory
+            && Kdf.Parallelism == unlockData.Kdf.Parallelism;
```
mzieniukbw: This can be simplified.
```suggestion
        return Kdf.ToData().Equals(unlockData.Kdf.ToData());
```

mzieniukbw (dismissed):

mzieniukbw (approved):

JaredSnider-Bitwarden (approved):

PR #7514, opened 2026-04-20
[PM-12475] Add GroupUserAuthorizationHandler and related context for group user …

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12475

## 📔 Objective

Instead of having duplicated code in places for authorization for assigning users to group authorization, implement an authorization service to be used in both places to de-duplicate (is that a word?) code.


eliykat (commented):
Our resource-based authorization patterns need some improvement. This is already being discussed over in #7442, but isn't fully figured out yet. Let's hold on this until @r-tome has finalized a better pattern, then we can implement it here.

PR #7516, opened 2026-04-20
[PM-34615] - add SuppressOnboardingInterstitials setting

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34615

## 📔 Objective

Adds a new `SuppressOnboardingInterstitials` global setting to the server as part of PM-34615. When enabled, this setting signals to clients that onboarding interstitials (post-login welcome dialogs, extension install prompts, setup extension redirects, and premium upsell modals) should be suppressed — useful for QA environments where these prompts interfere with automated testing flows.

### Server changes

- Added `SuppressOnboardingInterstitials` boolean property to `GlobalSettings` and `IGlobalSettings`
- Exposed the property via `ServerSettingsResponseModel` (returned from `GET /config` in the `settings` object)
- Added unit tests verifying the value maps correctly through `ConfigResponseModel`


test/Api.Test/Models/Response/ServerSettingsResponseModelTests.cs
shane-melton, line 1: 👍 Thanks for these!

shane-melton (approved):

PR #7522, opened 2026-04-22
[PM-36158] - Fix org cipher push notifications to fan out per authorized user

jaasen-livefront:
## 🎟️ Tracking

[PM-35168](https://bitwarden.atlassian.net/browse/PM-35168)

## 📔 Objective

This PR refactors org cipher push notification delivery to be **per-user** and fixes several related regressions.

Previously, org cipher push notifications were either silently dropped (NotificationHub / Relay) or broadcast to an org-level SignalR group (Notifications API / AzureQueue) without respecting per-user collection access. This PR moves org cipher fan-out into `MultiServicePushNotificationService`, which resolves the exact set of users who can access the cipher and delivers a targeted `User`-scoped push to each one.

## Changes

**Org cipher push fan-out moved to `MultiServicePushNotificationService`**

- When `PushCipherAsync` is called for an org cipher, the service:
  1. Resolves the collection IDs the cipher belongs to (using the provided `collectionIds` or falling back to `GetCollectionIdsByCipherIdAsync` if none were supplied).
  2. Calls `GetUserIdsByCollectionIdsAsync` to find all confirmed org members with access to those collections (via direct assignment or group membership).
  3. Fans out a `User`-targeted `PushAsync` call per user, concurrently via `Task.WhenAll`.
- If no collection IDs can be resolved, a warning is logged and the push is skipped rather than sending an incorrect broadcast.

**Org cipher handling removed from individual push engines**

- `AzureQueuePushEngine`, `NotificationsApiPushEngine`, `RelayPushEngine`, and `NotificationHubPushEngine` no longer handle org ciphers in `PushCipherAsync`; that responsibility now lives exclusively in `MultiServicePushNotificationService`.
- `HubHelpers` no longer broadcasts cipher notifications to an org-level SignalR group; per-user fan-out upstream means the engine only sees `UserId`-targeted messages.

**`CipherService` fixes**

- `SaveAsync` and `SaveDetailsAsync` now correctly forward `collectionIds` to `PushSyncCipherCreateAsync` (previously always passed `null`).
- `DeleteAsync` now calls `PushCipherAsync(cipher, PushType.SyncLoginDelete, collectionIds)` directly, replacing the old `PushSyncCipherDeleteAsync(cipher)` call that did not carry collection context.
- New private helper `GetCollectionIdsForPushAsync` retrieves collection IDs for org ciphers before push, returning `null` for personal ciphers.

**New repository surface**

- `ICollectionCipherRepository.GetCollectionIdsByCipherIdAsync(Guid cipherId)` — returns the collection IDs a specific cipher belongs to (targeted single-cipher lookup, avoids the O(org) query).
- `ICollectionCipherRepository.GetUserIdsByCollectionIdsAsync(IEnumerable<Guid> collectionIds)` — returns distinct confirmed user IDs with direct or group-based access to the given collections.
- Implemented across Dapper, EF Core, and a new Noop repository.
- `ICollectionCipherRepository` is now registered as a noop singleton in `AddPush` to satisfy DI in push-only service compositions.

**Database**

- New SSDT stored procedures:
  - `CollectionCipher_ReadCollectionIdsByCipherId`
  - `CollectionCipher_ReadUserIdsByCollectionIds`
- Migration scripts added under `util/Migrator/DbScripts/`.

**Tests**

- `AzureQueuePushEngineTests`: Replaced parameterized personal/org test cases with separate `_PersonalCipher_SendsExpectedResponse` and `_OrgCipher_DoesNotSendMessage` variants to clearly document the new engine contract.
- `MultiServicePushNotificationServiceTests`: Added tests for personal cipher delegation, org cipher per-user fan-out, and the null-collection-ID fallback resolution path.
- `CipherServiceTests`: Updated five `DeleteAsync_*` assertions from `PushSyncCipherDeleteAsync` to `PushCipherAsync` to match the production call shape.

## Behavior With Feature Flag **OFF** (`OrgCipherPushFanout = false`)

- **Org cipher sync pushes** are sent as a single org-level broadcast (`NotificationTarget.Organization`) **only to non-mobile engines** (SignalR/Azure Queue/NotificationsApi).
- **Mobile engines** (NotificationHub, Relay) **do not receive** org cipher sync pushes.
- This preserves legacy behavior: web/desktop clients receive org cipher sync notifications, but mobile clients do not (avoiding unnecessary wakeups and full vault syncs).
- **Personal cipher sync pushes** are unaffected and delivered to all clients as before.

---

## Behavior With Feature Flag **ON** (`OrgCipherPushFanout = true`)

- **Org cipher sync pushes** are fanned out per-user (resolved via collection access) and sent as individual user-targeted notifications (`NotificationTarget.User`) to **all engines** (web, desktop, and mobile).
- **Mobile clients** will now receive per-cipher org sync notifications for ciphers they have access to.
- This enables granular, access-aware push delivery for all platforms.

## 📸 Screenshots

N/A — no UI changes.

[PM-35168]: https://bitwarden.atlassian.net/browse/PM-35168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
```diff
@@ -0,0 +1,23 @@
...
+    UNION
+
+    -- Users with group-based collection access
+    SELECT DISTINCT OU.[UserId]
+    FROM [dbo].[CollectionGroup] CG
+    INNER JOIN [dbo].[GroupUser] GU ON GU.[GroupId] = CG.[GroupId]
+    INNER JOIN [dbo].[OrganizationUser] OU ON OU.[Id] = GU.[OrganizationUserId]
+    WHERE CG.[CollectionId] IN (SELECT [Id] FROM @CollectionIds)
+        AND OU.[Status] = 2 -- Confirmed
+END
```
nick-livefront, line 49: This query covers direct CollectionUser and group-based access, but it misses users who have access to all org ciphers implicitly, (OrganizationUser.Type of Owner or Admin) and users with `AccessAll = true`
 
```sql
 -- Users with org-level access (admins, owners, AccessAll)
UNION
SELECT DISTINCT OU.[UserId]
FROM [dbo].[OrganizationUser] OU
INNER JOIN [dbo].[CollectionCipher] CC ON CC.[CollectionId] IN (SELECT [Id] FROM @CollectionIds)
INNER JOIN [dbo].[Collection] COL ON COL.[Id] = CC.[CollectionId]
WHERE OU.[OrganizationId] = COL.[OrganizationId]
  AND OU.[Status] = 2
  AND (OU.[AccessAll] = 1 OR OU.[Type] IN (0, 1))
```

nick-livefront: The same change would be needed on the EF side of things in `src/Infrastructure.EntityFramework/Repositories/CollectionCipherRepository.cs`

src/Core/Vault/Services/Implementations/CipherService.cs
```diff
@@ -992,6 +994,16 @@ private async Task<long> StorageBytesRemainingForCipherAsync(Cipher cipher)
         return storageBytesRemaining;
     }
 
+    private async Task<ICollection<Guid>> GetCollectionIdsForPushAsync(Cipher cipher)
```
nick-livefront: 🎨 This should be nullable, given it can return null

```suggestion
    private async Task<ICollection<Guid?>> GetCollectionIdsForPushAsync(Cipher cipher)
```

nick-livefront (changes requested):

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
jaasen-livefront: @nick-livefront Good point.  `AccessAll` was removed but we should definitely automatically send this to admins and owners.

https://github.com/bitwarden/server/pull/7522/commits/bb3d8b48bb7566c0b2a2950edefa25fe7d2483b1

src/Core/Vault/Services/Implementations/CipherService.cs
jaasen-livefront: @nick-livefront Good suggestion. The file is `#nullable disable` but this is a good opportunity to clean that up. I've done so along with `ICipherService`.

https://github.com/bitwarden/server/pull/7522/commits/bb3d8b48bb7566c0b2a2950edefa25fe7d2483b1

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
nick-livefront: Nice catch on AccessAll. Your current fix will allow any admin or owner to receive notifications, it still should be restricted to if they have access. `AllowAdminAccessToAllCollectionItems` should be checked to see if they have that setting enabled for any owner to admin to access all items.

jaasen-livefront: @nick-livefront Ah yes ofc. Fixed. https://github.com/bitwarden/server/pull/7522/commits/e7d235d416453b8013b10ca355013760814c80d9

nick-livefront: The EF implementation needs the fix as well 😄

jaasen-livefront: 🤦 
https://github.com/bitwarden/server/pull/7522/commits/7b343e6d3138308dc3fcab390234292101f66641

nick-livefront (approved):

src/Core/Platform/Push/Engines/AzureQueuePushEngine.cs
```diff
@@ -33,19 +33,9 @@ public AzureQueuePushEngine(
...
-                RevisionDate = cipher.RevisionDate,
-                CollectionIds = collectionIds,
-            };
-
-            await SendMessageAsync(type, message, true);
-        }
-        else if (cipher.UserId.HasValue)
+        // Org cipher fan-out is handled upstream in MultiServicePushNotificationService,
+        // which resolves per-user access and calls PushAsync directly.
+        if (cipher.UserId.HasValue)
```
justindbaur: If this is true, then `PushCipherAsync` can be deleted from `IPushEngine`. It only exists because of the different behavior of the engines but since this PR is making it so there isn't a difference it shouldn't need to exist anymore.

src/Core/Platform/Push/Engines/MultiServicePushNotificationService.cs
justindbaur, line 1: This is not the place to add this kind of logic. You should likely create your own service that has methods like `PushSyncCipherCreateAsync`, `PushSyncCipherUpdateAsync`, `PushSyncCipherDeleteAsync` and those do all the logic you have here then it can call `IPushNotificationService.PushAsync` then `PushCipherAsync` and associated tests can be deleted.  

This new service is also the place where this needs to be feature flagged. This has the possibility of new load to the database when notifications are being sent, new amount of notifications being sent and new load to the server because clients are now reacting to notifications they did not get before by doing API calls.

justindbaur (changes requested):

src/Core/Platform/Push/Engines/AzureQueuePushEngine.cs
jaasen-livefront: @justindbaur Indeed. Removed.https://github.com/bitwarden/server/pull/7522/commits/2f0aab493cd6689c68b3b11ac371d2aec19eea1f

src/Core/Platform/Push/Engines/MultiServicePushNotificationService.cs
jaasen-livefront: Great idea. Added a service as well as feature flag on/off handling.
https://github.com/bitwarden/server/pull/7522/commits/2f0aab493cd6689c68b3b11ac371d2aec19eea1f

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
```diff
@@ -0,0 +1,36 @@
+CREATE PROCEDURE [dbo].[CollectionCipher_ReadUserIdsByCollectionIds]
+    @CollectionIds AS [dbo].[GuidIdArray] READONLY
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    -- Users with direct collection access
+    SELECT DISTINCT OU.[UserId]
```
mkincaid-bw: ⛏️ Keywords in the `SELECT` statents should be broken out on their own line, like you have in `CollectionCipher_ReadCollectionIdsByCipherId`
https://contributing.bitwarden.com/contributing/code-style/sql/#select-statements

```sql
    SELECT DISTINCT 
        OU.[UserId]
    FROM
        [dbo].[CollectionUser] CU
    INNER JOIN
        [dbo].[OrganizationUser] OU ON OU.[Id] = CU.[OrganizationUserId]
    WHERE
        CU.[CollectionId] IN (SELECT [Id] FROM @CollectionIds)
        AND OU.[Status] = 2 -- Confirmed
```

jaasen-livefront: @mkincaid-bw Ah yes thanks for pointing that out! Fixed https://github.com/bitwarden/server/pull/7522/commits/dee363803fa1695d91c7c6de69dad6bd98a8b9b3

jaasen-livefront:
@justindbaur Ready for another look when you have a chance. Ty!

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
```diff
@@ -0,0 +1,49 @@
...
+
+    -- Users with direct collection access
+    SELECT DISTINCT
+        OU.[UserId]
+    FROM
+        [dbo].[CollectionUser] CU
+    INNER JOIN
+        [dbo].[OrganizationUser] OU ON OU.[Id] = CU.[OrganizationUserId]
+    WHERE
+    CU.[CollectionId] IN (SELECT [Id] FROM @CollectionIds)
```
mkincaid-bw: ⛏️ Total nitpick, can you please indent this 4 more spaces?

util/Migrator/DbScripts/2026-04-30_00_CollectionCipher_ReadUserIdsByCollectionIds.sql
```diff
@@ -0,0 +1,49 @@
...
+
+    -- Users with direct collection access
+    SELECT DISTINCT
+        OU.[UserId]
+    FROM
+        [dbo].[CollectionUser] CU
+    INNER JOIN
+        [dbo].[OrganizationUser] OU ON OU.[Id] = CU.[OrganizationUserId]
+    WHERE
+    CU.[CollectionId] IN (SELECT [Id] FROM @CollectionIds)
```
mkincaid-bw: Same as above.

mkincaid-bw (changes requested):
Minor nit-pick formatting change requested.

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
jaasen-livefront: Fixed! https://github.com/bitwarden/server/pull/7522/commits/aaf2218e192079354627f2d411d15bbf004b1534

util/Migrator/DbScripts/2026-04-30_00_CollectionCipher_ReadUserIdsByCollectionIds.sql
jaasen-livefront: Fixed in https://github.com/bitwarden/server/pull/7522/commits/aaf2218e192079354627f2d411d15bbf004b1534.

Can't wait for eslint to point this out some day!

src/Sql/dbo/Stored Procedures/CollectionCipher_ReadUserIdsByCollectionIds.sql
```diff
@@ -0,0 +1,49 @@
...
+    -- Users with direct collection access
+    SELECT DISTINCT
+        OU.[UserId]
+    FROM
+        [dbo].[CollectionUser] CU
+    INNER JOIN
+        [dbo].[OrganizationUser] OU ON OU.[Id] = CU.[OrganizationUserId]
+    WHERE
+        CU.[CollectionId] IN (SELECT [Id] FROM @CollectionIds)
+            AND OU.[Status] = 2 -- Confirmed
```
mkincaid-bw: Sorry, one more.
```suggestion
        AND OU.[Status] = 2 -- Confirmed
```

jaasen-livefront: @mkincaid-bw Ah nice catch. Fixed https://github.com/bitwarden/server/pull/7522/commits/204fcc8ba3cace3ae6b0d79245187361acfa083c

mkincaid-bw (dismissed):
LGTM

jaasen-livefront:
@justindbaur @nick-livefront Ready for another look when you have some time. Ty!

nick-livefront (dismissed):

src/Core/Platform/Push/PushNotification.cs
```diff
@@ -45,11 +45,17 @@ public record PushNotification<T>
...
+    /// The type of clients the notification should be sent to, if <see langword="null"/> then
     /// <see cref="ClientType.All"/> is inferred.
     /// </summary>
     public ClientType? ClientType { get; init; }
 
+
+    /// <summary>
+    /// When true, only non-mobile engines (SignalR/web/desktop) will deliver this notification. When null or false, all engines process it.
+    /// </summary>
+    public bool? NonMobileOnly { get; init; }
```
justindbaur, line 62: Am I correct in assuming this option can go away when the flag is removed? If that is true then please mention that no-one else should use this option and it will be going away. It should also be part of a ticket that is tracking the removal of the flag.

Could also throw `[EditorBrowsable(EditorBrowsableState.Never)]` on it to discourage additional usage.

src/Core/Platform/Push/PushServiceCollectionExtensions.cs
```diff
@@ -27,6 +28,7 @@ public static IServiceCollection AddPush(this IServiceCollection services, Globa
 
         services.TryAddSingleton(TimeProvider.System);
         services.TryAddSingleton<IPushNotificationService, MultiServicePushNotificationService>();
+        services.TryAddSingleton<ICollectionCipherRepository, Bit.Core.Repositories.Noop.CollectionCipherRepository>();
```
justindbaur: This should be able to be moved elsewhere.

src/Core/Repositories/Noop/CollectionCipherRepository.cs
justindbaur, line 1: We'd normally only have noop repositories for things that are cloud only. I don't think that applies in this case so I'd think this can be removed.

justindbaur (changes requested):

src/Core/Platform/Push/PushNotification.cs
jaasen-livefront: Yes that's correct. Added the comment and `EditorBrowserableState.Never`. 

https://github.com/bitwarden/server/pull/7522/commits/a3bf67f004e179999afe7b86c1160044c7495a05

src/Core/Platform/Push/PushServiceCollectionExtensions.cs
jaasen-livefront: Done!

https://github.com/bitwarden/server/pull/7522/commits/a3bf67f004e179999afe7b86c1160044c7495a05

src/Core/Repositories/Noop/CollectionCipherRepository.cs
jaasen-livefront: Oh ok. Removed.

https://github.com/bitwarden/server/pull/7522/commits/a3bf67f004e179999afe7b86c1160044c7495a05

justindbaur (dismissed):
Looks great to me!

nick-livefront (dismissed):

mkincaid-bw (dismissed):

nick-livefront (approved):

addisonbeck (approved):

justindbaur (approved):

mkincaid-bw (approved):

PR #7530, opened 2026-04-22
Auth/pm 35392/master password service foundation

enmande:
## 🎟️ Tracking

[PM-35392](https://bitwarden.atlassian.net/browse/PM-35392)

## 📔 Objective

Introduces the `MasterPasswordService`, an orchestrating/coordinating surface for Master Password initial set and update operations.

No call sites are connected as part of this PR. For ease of review, those are specced in detail in connected tickets:
- [PM-35393](https://bitwarden.atlassian.net/browse/PM-35393) (Auth APIs)
- [PM-35394](https://bitwarden.atlassian.net/browse/PM-35394) (Admin Console APIs)
- [PM-35395](https://bitwarden.atlassian.net/browse/PM-35395) (Key Management APIs)

A suite of unit tests has been added for the service.

## 📸 Screenshots

N/A

[PM-35392]: https://bitwarden.atlassian.net/browse/PM-35392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-35393]: https://bitwarden.atlassian.net/browse/PM-35393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-35394]: https://bitwarden.atlassian.net/browse/PM-35394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-35395]: https://bitwarden.atlassian.net/browse/PM-35395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Auth/UserFeatures/UserMasterPassword/Data/SetInitialOrUpdateExistingPasswordData.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Bit.Core.KeyManagement.Models.Data;
+
+namespace Bit.Core.Auth.UserFeatures.UserMasterPassword.Data;
+
+public class SetInitialOrUpdateExistingPasswordData
+{
+    public required MasterPasswordAuthenticationData MasterPasswordAuthentication { get; set; }
```
Patrick-Pimentel-Bitwarden: ⛏️ Unlock then authenticate

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,206 @@
+﻿using Bit.Core.Auth.UserFeatures.UserMasterPassword.Data;
+using Bit.Core.Entities;
+using Bit.Core.Repositories;
+using Microsoft.AspNetCore.Identity;
+using OneOf;
+
+namespace Bit.Core.Auth.UserFeatures.UserMasterPassword.Interfaces;
+
+/// <summary>
+/// Centralized mutation point for all master password set, change, and rotate operations.
```
Patrick-Pimentel-Bitwarden:
```suggestion
/// Centralized mutation point for all master password set and update operations.
```

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,206 @@
...
+/// Provides consistent validation, password hashing, and timestamp management across every
+/// flow that establishes or updates a user's master password.
+///
+/// <strong>Compositional, not orchestrating.</strong> This service handles CRUD-like mutations
```
Patrick-Pimentel-Bitwarden: ⛏️  Could you double check how this \<strong> tag renders for you? In rider it doesn't format properly and maybe these are taking up valuable space.

Image

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
```diff
@@ -0,0 +1,280 @@
...
+        EnsureUserIsHydrated(user);
+        var result = await PrepareSetInitialMasterPasswordAsync(user, setInitialData);
+        if (result.IsT1)
+        {
+            return result.AsT1;
+        }
+
+        await _userRepository.ReplaceAsync(user);
+
+        return user;
```
Patrick-Pimentel-Bitwarden: ❓ Would this be a better approach for all the returns? This way it's clear we are using the output of the function call. It's meaningless now but I think there's a smell here to possibly either comment on or change.
```suggestion
        return result.AsT0;
```

src/Infrastructure.Dapper/Repositories/UserRepository.cs
```diff
@@ -488,6 +488,9 @@ public UpdateUserData SetMasterPassword(Guid userId, MasterPasswordUnlockData ma
                     RevisionDate = timestamp,
                     AccountRevisionDate = timestamp,
                     MasterPasswordSalt = masterPasswordUnlockData.Salt
+                    // TODO (PM-35501): Add SecurityStamp so the rotation done in
```
Patrick-Pimentel-Bitwarden, line 491: Comment for LastKeyRotationDate was deleted here. That also needs to be added in 34905


```suggestion
                    // TODO (PM-34905): Add LastPasswordChangeDate here
                    // TODO (PM-35501): Add SecurityStamp so the rotation done in
```

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
```diff
@@ -563,6 +563,9 @@ public UpdateUserData SetMasterPassword(Guid userId, MasterPasswordUnlockData ma
             userEntity.RevisionDate = timestamp;
             userEntity.AccountRevisionDate = timestamp;
             userEntity.MasterPasswordSalt = masterPasswordUnlockData.Salt;
+            // TODO (PM-35501): Persist SecurityStamp so the rotation done in
```
Patrick-Pimentel-Bitwarden, line 566: Missing LastPasswordChangeDate comment too just like dapper

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
```diff
@@ -0,0 +1,280 @@
...
+                var validate = await ValidatePasswordInternalAsync(user,
+                    setInitialData.MasterPasswordAuthentication.MasterPasswordAuthenticationHash);
+                if (!validate.Succeeded)
+                {
+                    throw new BadRequestException(
+                        string.Join("; ", validate.Errors.Select(e => e.Description)));
+                }
+            }
+
+            if (setInitialData.RefreshStamp)
```
Patrick-Pimentel-Bitwarden, line 121: ❓ Do we want to also set the LastPasswordChangeDate here too as a quick fix until addressed later?

Patrick-Pimentel-Bitwarden (commented):
I can't begin to say how giddy I am with the changes you've made. I'm so stoked to see that the job you've done is top notch. Thank you 🙏 

1. Do a verbiage check to make sure we are always using set initial or update existing. No concept of "change" anymore as it's ambiguous and we need to be crystal clear about the operations we are performing. This will be feedback to bring into all the rest of the prs as well.
2. ⛏️ Throughout the MasterPasswordService I left comments on some of the operations happening throughout the set initial and update existing functions and didn't make the comments uniform. Would you mind addressing that? For example the `PrepareSetInitialMasterPasswordAsync` says `// Update time markers on the user` but I omitted that on `PrepareUpdateExistingMasterPasswordAsync`.

src/Core/Auth/UserFeatures/UserMasterPassword/Data/SetInitialOrUpdateExistingPasswordData.cs
enmande: Updated throughout for declarative uses in [b7db39b](https://github.com/bitwarden/server/pull/7530/commits/b7db39bb308744cb5d5c5933006767d940ea360b).
As a follow-up: that changeset is pretty comprehensive if we want to ensure we're thinking of them in order, and _always looking for the pair_, which captures their relationship (dependency/causality). `Unlock` is a source of truth, `Authentication` is derivative. It's a smell throughout the calling layer if we don't see these two together. 
_However_ we have some internals/method bodies like in `PrepareSetInitialMasterPasswordAsync`, where we interrogate `Authentication` first -- it's the validation step before we start working with the key and returning the `user`, for example. I _think_ as-is, this is more readable and appropriate, but it does raise a slight inversion of the principle. 99% of the time, this should be out of sight, and is each piece doing its job. But I do want to make sure I also understand all the reasons why this is an important distinction of _unlock then authenticate_.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: Interesting! It did render for me in VSCode. _However_, `<strong>` [is not a standard tag](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags), so it needs to go. I've re-fit these with, e.g., `<term>`, and other tags which are both standard, and do a more technical job of pulling the semantic weight in ba43405.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: Fixed this instance in ba43405. Need to go over tomorrow with a broader net, so please leave this comment unresolved for now; I will follow up when satisfied with the whole.

src/Core/Auth/UserFeatures/UserMasterPassword/Data/SetInitialOrUpdateExistingPasswordData.cs
Patrick-Pimentel-Bitwarden: Yeah I think it's great to see where the pattern falls apart. In that 1% we should have a comment as to why we can omit the missing check/validation etc to call attention to why we _have_ to stray away. Looks great 👍

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: This should be wrapped up with 22731cb. 
With this change, _operations_ should be consistently couched in set/update language. References to _value comparison for the sake of validation_ consistently use "change" to signal difference (e.g., a salt has changed at time of check and violated consistency, expect the validation to fail). Open to making this be a different verb, but for example "salt remains unchanged" feels neater than "salt is not allowed to _differ_" or similar in validation-focused comments. LMK if you want this iterated on!

src/Core/Auth/UserFeatures/UserMasterPassword/Data/SetInitialOrUpdateExistingPasswordData.cs
enmande: Thanks! In 22731cb, I added [an explicit comment around internals](https://github.com/bitwarden/server/pull/7530/changes/22731cb10f8cfc65dff32bc853eccd72038c6fa3#diff-bbf583c2f1c386eb94d50a66f23b682e8945d126cfa8cfbf5bee310d579fbba1R110).

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
enmande: You're right, and I updated this in 58ffe45.
I leaned into `user` because personally I'm not a fan of the ordinal handling of returns from `OneOf`; I find them a little wonky for reading. _However_, I can see this is both a consistent pattern in the codebase where `OneOf` is used, _and_ is absolutely more true to the reading of the method internals: "We performed an operation on user, got a result, use the result."

Patrick-Pimentel-Bitwarden: What we could do is once we know the result is not an error we could assign a var user = result.T0; That keeps the readability up. If we ever **do** create a new user object separate from the one we pass in from the params then we'll be covered there. Thoughts?

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,206 @@
...
+    /// <param name="updateExistingExistingData">
+    /// Cryptographic and authentication data for the updated password and KDF parameters,
+    /// including <c>MasterPasswordUnlock</c>, <c>MasterPasswordAuthentication</c>,
+    /// and control flags <c>ValidatePassword</c> and <c>RefreshStamp</c>.
+    /// </param>
+    /// <returns>
+    /// On success, the modified <see cref="User"/>. On failure, an array of
+    /// <see cref="IdentityError"/> describing validation failures.
+    /// </returns>
+    Task<OneOf<User, IdentityError[]>> SaveUpdateExistingMasterPasswordAndKdfAsync(User user, UpdateExistingPasswordAndKdfData updateExistingExistingData);
```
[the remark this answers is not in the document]
enmande: Fixed in bf82f2e.

src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
enmande: Restored in d9b98b3.

src/Infrastructure.Dapper/Repositories/UserRepository.cs
enmande: Restored in d9b98b3.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
enmande: In 58ffe45, I have updated all the uses of `OneOf` to continue the use of `result.AsTx` throughout. We could introduce an intermediary variable for readability; I'll use the open/broader review to get feedback on that because I _like_ the readability of an intermediary, but I _think_ I see the ordinal type `result.AsTx` being the prevailing pattern, and it's' a more direct through-line. End of the day, either is more correct and durable than re-using `user`. 😅

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
enmande: I don't think so; I think we need to let the database handle that as persistence layer. It's a strong argument to fast-follow with [PM-34905](https://bitwarden.atlassian.net/browse/PM-34905) for consistency, but I think it belongs only there.

src/Core/Auth/UserFeatures/UserMasterPassword/Data/UpdateExistingPasswordAndKdfData.cs
```diff
@@ -0,0 +1,55 @@
...
+        if (user.UsesKeyConnector)
+        {
+            throw new BadRequestException("Cannot update password of a user with Key Connector.");
+        }
+
+        // Do not validate if kdf is the same here on the user because we are changing it.
+
+        // Validate Salt is unchanged for user
+        MasterPasswordUnlock.ValidateSaltUnchangedForUser(user);
+        MasterPasswordAuthentication.ValidateSaltUnchangedForUser(user);
```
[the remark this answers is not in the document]
enmande: This is a good finding _in isolation_, however not applicable to this PR as written.
KDF-change path is layered on purpose; `ChangeKdfCommand` is responsible for KDF consistency in a change path. [PM-35395](https://bitwarden.atlassian.net/browse/PM-35395) documents this cleanly and lays out steps for implementation _after_ this service is merged. To do _that_ validation here would improperly apply the domain boundary.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,209 @@
...
+/// <para>Source of truth: On SET, the client is the source of truth. On UPDATE,
+/// the server is the source of truth for fields that must not change — it validates the client's
+/// values match what's stored before applying the update.</para>
+/// </summary>
+internal interface IMasterPasswordService
+{
+    /// <summary>
+    /// Inspects the user's current state and dispatches to either
+    /// <see cref="PrepareSetInitialMasterPasswordAsync"/> or
+    /// <see cref="PrepareUpdateExistingMasterPasswordAsync"/> accordingly.
```
JaredSnider-Bitwarden, line 79: 🎨 : the other methods detail "Use for flows that..." but this one does not.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,209 @@
...
+///   </item>
+///   <item>
+///     <c>Build*</c> Returns a deferred <see cref="UpdateUserData"/> delegate for
+///     <see cref="IUserRepository.UpdateUserDataAsync"/> batch transactions. Use when the
+///     password set is part of a larger transactional write that must succeed or fail atomically
+///     (e.g. TDE key + password in a single SQL transaction)
+///   </item>
+/// </list>
+///
+/// <para>Set vs Update contract:</para>
```
JaredSnider-Bitwarden: Per sync discussion, can we improve these headers with a bit more context before we jump in the great explanations below.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// <see cref="IdentityError"/> describing validation failures.
+    /// </returns>
+    Task<OneOf<User, IdentityError[]>> PrepareSetInitialMasterPasswordAsync(User user, SetInitialPasswordData setInitialPasswordData);
+
+    /// <summary>
+    /// Note: This is to be used in the future when a TDE user wants to set a password with self-service.
+    ///
+    /// Applies a new initial master password to the <paramref name="user"/> object and persists
+    /// the updated user. Use when no external transaction coordination is needed.
+    /// </summary>
```
JaredSnider-Bitwarden, line 163: 🎨 : should this have the same requirement details that the other methods do?

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// On success, the modified <see cref="User"/>. On failure, an array of
+    /// <see cref="IdentityError"/> describing validation failures.
+    /// </returns>
+    Task<OneOf<User, IdentityError[]>> PrepareUpdateExistingMasterPasswordAsync(User user, UpdateExistingPasswordData updateExistingData);
+
+    /// <summary>
+    /// Applies a new master password and updated KDF parameters over the user's existing ones
+    /// and persists the updated user to the database. Salt must remain unchanged; KDF 
+    /// validation is intentionally skipped. Use for KDF rotation flows.
+    /// </summary>
```
JaredSnider-Bitwarden, line 274: 🎨 : what do you think about creating a standardized template for each method for stuff like:
1. What the method does
2. When it is expected to be used or the flow
3. What the requirements are for calling it (i.e., user must not have password)

src/Core/Auth/UserFeatures/UserMasterPassword/Data/UpdateExistingPasswordAndKdfData.cs
```diff
@@ -0,0 +1,55 @@
...
+            throw new BadRequestException("User does not have an existing master password to update.");
+        }
+
+        // KDF parameters govern how the master password is stretched into the encryption key.
+        // Key Connector replaces the master password entirely — the encryption key is managed
+        // by an external service — so KDF rotation has no meaningful target. The existing
+        // ChangeKdfCommand blocks this implicitly (CheckPasswordAsync fails against a null
+        // master password), but this guard makes the categorical inapplicability explicit.
+        // Note: org owners/admins cannot be KC users (enforced at conversion time in
+        // UserService.CheckCanUseKeyConnector), so no role-based edge case exists.
```
JaredSnider-Bitwarden, line 53: Thank you for documenting this. That was my immediate thought.

src/Core/Auth/UserFeatures/UserMasterPassword/Data/SetInitialOrUpdateExistingPasswordData.cs
JaredSnider-Bitwarden, line 1: Should thishave similar top level class docs for answering the question detailed here: https://github.com/bitwarden/server/pull/7530/changes#r3138835307
about which flows are expected to use this model?

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden, line 1: Per sync discussion, let's add a TODO: to replace `OneOf` with .NET 10's union.

JaredSnider-Bitwarden, line 1: Stellar docs. 🏅

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
```diff
@@ -0,0 +1,304 @@
...
+    {
+        EnsureUserIsHydrated(user);
+        setInitialData.ValidateDataForUser(user);
+
+        // Note: Keep "unlock and authenticate" pattern in mind.
+        // We name "unlock" first as a naming convention,
+        // e.g., MatherPasswordUnlockAndAuthenticationDataModel. 
+        // The two are complimentary, but mechanically Authenticate comes first.
+        // Eager validation keeps the logic easier to reason about. 
+        // Authentication is the mechanism for validation, unlock is the capability.
```
JaredSnider-Bitwarden: ❓ : I understand this comment in isolation, but I'm struggling to reason about how it relates to the code immediately after it?

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: Applied in b799072.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
```diff
@@ -0,0 +1,304 @@
...
+        EnsureUserIsHydrated(user);
+        setInitialData.ValidateDataForUser(user);
+
+        // Note: Keep "unlock and authenticate" pattern in mind.
+        // We name "unlock" first as a naming convention,
+        // e.g., MatherPasswordUnlockAndAuthenticationDataModel. 
+        // The two are complimentary, but mechanically Authenticate comes first.
+        // Eager validation keeps the logic easier to reason about. 
+        // Authentication is the mechanism for validation, unlock is the capability. 
+        var result = await UpdateExistingPasswordHashAsync(
```
JaredSnider-Bitwarden: ❓ : I found it confusing that the `PrepareSetInitial` method called this `UpdateExistingPasswordHashAsync` helper.  

Would it be more clear to name the helper `UpdateExistingPasswordHashAsync` to be something like `ApplyPasswordHashAsync` clarify this a bit as the helper doesn't care if a password already exists. 

Here's a part of my discussion with claude on it:

> The helper is misnamed. It doesn't actually check or care whether a password already exists — it just:
> 
> Optionally validates the password string through validators
> Hashes it and sets user.MasterPassword
> Optionally refreshes SecurityStamp
> Those three operations are equally necessary when setting an initial password. The name UpdateExistingPasswordHashAsync implies a precondition (existing password) that the method never enforces. It's really just "validate, hash, and stamp" — something like ApplyPasswordHashAsync would describe what it actually does.
> 
> So PrepareSetInitialMasterPasswordAsync calling it isn't conceptually wrong — it's just that the helper's name makes it look wrong. A reviewer reading this cold would have the exact same reaction you did: why is SET calling UPDATE?

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: `Use when:` added in b799072.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: `Constraints:` enumerated in b799072.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
JaredSnider-Bitwarden: A brief note of docs on the helper could also help with the intention.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
```diff
@@ -0,0 +1,304 @@
...
+
+        return user;
+    }
+
+    public async Task<OneOf<User, IdentityError[]>> SaveSetInitialMasterPasswordAsync(
+        User user,
+        SetInitialPasswordData setInitialData)
+    {
+        EnsureUserIsHydrated(user);
+        var result = await PrepareSetInitialMasterPasswordAsync(user, setInitialData);
```
JaredSnider-Bitwarden, line 90: Elegant composition.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// </param>
+    /// <param name="setInitialPasswordData">
+    /// Cryptographic and authentication data required to set the initial password. See
+    /// <see cref="PrepareSetInitialMasterPasswordAsync"/> for field details.
+    /// </param>
+    /// <returns>
+    /// An <see cref="UpdateUserData"/> delegate suitable for inclusion in a batch passed to
+    /// <see cref="IUserRepository.UpdateUserDataAsync"/>.
+    /// </returns>
+    UpdateUserData BuildUpdateUserDelegateSetInitialMasterPassword(User user, SetInitialPasswordData setInitialPasswordData);
```
JaredSnider-Bitwarden, line 220: 🎨 : we should document that we are supporting this method b/c it exists currently but we ideally would prefer callers use the other methods and we would like to drop support for this approach later on.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
JaredSnider-Bitwarden: Or something more akin to the existing `SetKdfStateOnUser` helper.

test/Core.Test/Auth/UserFeatures/UserMasterPassword/MasterPasswordServiceTests.cs
```diff
@@ -0,0 +1,1378 @@
...
+            var user = BuildValidUpdateUser();
+            var data = BuildData(user, saltOverride: "wrong-salt");
+
+            Assert.Throws<BadRequestException>(() => data.ValidateDataForUser(user));
+        }
+    }
+
+    // --- Data model validation: UpdateExistingPasswordAndKdfData ---
+
+    public class UpdateExistingPasswordAndKdfDataTests
```
JaredSnider-Bitwarden: 🎨 , this test file is large, do you think we could move the data model tests into their own files as easy wins?

test/Core.Test/Auth/UserFeatures/UserMasterPassword/MasterPasswordServiceTests.cs
JaredSnider-Bitwarden, line 1: Extremely good coverage. Well done. However, in asking directly for a code cov evaluation, claude found these as possible points to update:

## Code Coverage Assessment: `MasterPasswordService`

All 7 public interface methods are exercised. The main gaps are noted below.

### Well-covered

| Method | Tests |
|---|---|
| `PrepareSetInitialMasterPasswordAsync` | 6 — success, hint, hydration, RefreshStamp both ways, validation failure |
| `SaveSetInitialMasterPasswordAsync` | 3 — persists, validation fail + no-persist, hydration |
| `BuildUpdateUserDelegateSetInitialMasterPassword` | 6 — hydration, happy path, existing-password guard, validate+persist, validate fail+throws, RefreshStamp=true |
| `PrepareUpdateExistingMasterPasswordAsync` | 4 — success, hint, hydration, validation failure |
| `SaveUpdateExistingMasterPasswordAsync` | 2 — persists, validation fail + no-persist |
| `SaveUpdateExistingMasterPasswordAndKdfAsync` | 6 — success, PBKDF2→Argon2id, Argon2id→PBKDF2, validation fail, salt mismatch, no-existing-password |
| `PrepareSetInitialOrUpdateExistingMasterPasswordAsync` | 5 — routes set/update, salt mismatch, validation propagation, hydration |

### Coverage gaps

**Service layer**

1. **`PrepareUpdateExistingMasterPasswordAsync` — `RefreshStamp` behavior**: no `RefreshStamp=true` or `RefreshStamp=false` test. The `Prepare*SetInitial` method has both; the update path does not.
2. **`BuildUpdateUserDelegateSetInitialMasterPassword` — `RefreshStamp=false`**: only stamp rotation is verified; preservation when `false` has no test.
3. **`SaveUpdateExistingMasterPasswordAndKdfAsync` — incomplete success assertions**: the success test only asserts `LastKdfChangeDate`. `LastPasswordChangeDate`, `RevisionDate`, `AccountRevisionDate`, `user.Key`, and `MasterPasswordHint` are all set by the service but not asserted.
4. **`SaveUpdateExistingMasterPasswordAsync` — hydration guard**: has its own `EnsureUserIsHydrated` call but no dedicated test. Low priority — `PrepareUpdateExistingMasterPasswordAsync` (called immediately after) also guards it.

**Data model layer**

5. **`UpdateExistingPasswordData.ValidateDataForUser` — split salt/KDF mismatches**: `SetInitialPasswordDataTests` tests unlock-only and authentication-only salt mismatches in isolation; `UpdateExistingPasswordDataTests` only tests both wrong together. Same gap for KDF — no test where unlock KDF mismatches but authentication KDF is correct.
6. **`UpdateExistingPasswordAndKdfData.ValidateDataForUser` — split salt mismatch**: same gap — no test covering unlock-only vs authentication-only salt divergence.

**Overall: ~90%+ of meaningful paths covered.** 


Thoughts on those?

JaredSnider-Bitwarden (changes requested):
Really excellent work here. Most of the items I found were small improvements or nit picks. Thank you for your efforts to build this foundation!

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: Updated in e1314df.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
enmande: Understand. @Patrick-Pimentel-Bitwarden and I discussed this a good bit. We have a convention of naming `Unlock` first, e.g., [MasterPasswordUnlockAndAuthenticationDataModel](https://github.com/bitwarden/server/blob/main/src/Api/KeyManagement/Models/Requests/MasterPasswordUnlockDataAndAuthenticationModel.cs). For _me_, it took some interrogation and reading to really grok their relationship because they share some of the same data (`KdfSettings`), etc., but fundamentally do different things. And when it comes time to use them, the order we care about them is inverse of the naming -- if you can't _authenticate_, your _unlock_ capability is moot. So I wanted to try and capture something that helped explain some of what I worked through in understanding, having not worked with these data models on the server before.
_Ultimately_, we may not need this comment at all, especially if it's confusing (I would rather remove it than spin on it more I think if we feel strongly about changing it, which I would totally understand from your read of it), but it was my attempt at leaving a breadcrumb for something that gave me non-trivial friction.

Maybe here the friction is good friction, and yields understanding, but I want it to not be opaque to new readers or those unfamiliar with it.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: Upon review, I was incorrect. `union` [is still a preview feature on .NET 11](https://devblogs.microsoft.com/dotnet/csharp-15-union-types/), not 10. Given the timeline, I can still put a comment in, but it seems like much further out than we thought, and that timeline may give us a cross-team opportunity to respond in a specific way?

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
enmande: Updated in 8ec3bfc. Took both of these points: 
- `summary` docs have been added to the helper.
- Both helpers have been renamed to have an `Apply` verb (i.e., also renamed `SetKdfStateOnUser` -> `ApplyKdfStateOnUser`). They are private to the service, and this helps distinguish them from the public API which uses `Update` as a public operation.

src/Core/Auth/UserFeatures/UserMasterPassword/Data/SetInitialOrUpdateExistingPasswordData.cs
enmande: Added in b18c6ac. Annotated all 4 new data models with summary/"use when."

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
enmande: Added in [1915ca3](https://github.com/bitwarden/server/pull/7530/commits/1915ca3936d8eefec5eb428be87bb999e5c27017).

test/Core.Test/Auth/UserFeatures/UserMasterPassword/MasterPasswordServiceTests.cs
enmande: Refactored in 9f42c6c.

test/Core.Test/Auth/UserFeatures/UserMasterPassword/MasterPasswordServiceTests.cs
enmande: Updated in aa5bd3e. All of these seem salient to cover; this service is responsible for a lot 😅 .
This changeset also introduced tests to illustrate the "legacy" PBKDF2 iterations cases for password change.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: The parallelism is so clean. Ty!

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,324 @@
...
+    ///   <item>User must have no existing <c>Key</c> or <c>MasterPasswordSalt</c>.</item>
+    ///   <item>User must not be a Key Connector user.</item>
+    /// </list>
+    /// </para>
+    /// 
+    /// </summary>
+    /// <param name="user">
+    /// The user object to mutate. Must not already have a master password; must have no existing
+    /// <c>Key</c> or <c>MasterPasswordSalt</c>; must not be a Key Connector user.
+    /// Validated via <see cref="SetInitialPasswordData.ValidateDataForUser"/>.
```
JaredSnider-Bitwarden, line 131: Do we need the constraints here since they are enumerated above?
```
    Must not already have a master password; must have no existing
    <c>Key</c> or <c>MasterPasswordSalt</c>; must not be a Key Connector user.
```

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,324 @@
...
+    /// <list type="bullet">
+    ///   <item>User must not already have a master password.</item>
+    ///   <item>User must have no existing <c>Key</c> or <c>MasterPasswordSalt</c>.</item>
+    ///   <item>User must not be a Key Connector user.</item>
+    /// </list>
+    /// </para>
+    ///
+    /// </summary>
+    /// <param name="user">
+    /// The user object to mutate and persist. Must not already have a master password; must have no
```
JaredSnider-Bitwarden, line 165: Same comment as above:https://github.com/bitwarden/server/pull/7530/changes#r3148933667

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,324 @@
...
+    ///
+    /// <para>
+    /// NOTE: This exists to support an existing pattern. Long-term preference is to be able to sunset this method
+    /// and lean on Prepare and Save verbs only. Please prefer those verbs if possible.
+    /// </para>
+    ///
+    /// </summary>
+    /// <param name="user">
+    /// The user whose initial master password state will be written when the returned delegate is
+    /// invoked. Must not already have a master password; must have no existing <c>Key</c> or
```
JaredSnider-Bitwarden, line 209: Same comment as above: https://github.com/bitwarden/server/pull/7530/changes#r3148933667

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
```diff
@@ -0,0 +1,324 @@
...
+    /// <list type="bullet">
+    ///   <item>User must already have a master password.</item>
+    ///   <item>User must not be a Key Connector user.</item>
+    ///   <item>KDF parameters and salt must be unchanged relative to the values in <paramref name="updateExistingData"/>.</item>
+    /// </list>
+    /// </para>
+    ///
+    /// </summary>
+    /// <param name="user">
+    /// The user object to mutate. Must already have a master password;
```
JaredSnider-Bitwarden, line 241: Same comment as above: https://github.com/bitwarden/server/pull/7530/changes#r3148933667

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: @enmande , I think this applies to all the methods.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: Per sync discussion, the repetition allows for the docs to show up on the param which is some added value.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: Per sync discussion, the repetition allows for the docs to show up on the param which is some added value.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: Per sync discussion, the repetition allows for the docs to show up on the param which is some added value.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: Per sync discussion, the repetition allows for the docs to show up on the param which is some added value.

src/Core/Auth/UserFeatures/UserMasterPassword/Interfaces/IMasterPasswordService.cs
JaredSnider-Bitwarden: Ok, that's fair. We can address later on.

src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs
JaredSnider-Bitwarden: Per sync discussion, let's remove these comments. 
Then, let's do this:
1. Rename `ApplyPasswordHashAsync` to `ApplyMasterPasswordAuthenticationHashAsync`
2. Add new `ApplyUserSecurityStampRotation` which is called at the root of the methods instead of inside of `ApplyMasterPasswordAuthenticationHashAsync`

enmande: Great discussion on this; thank you for the time and insight.
1 and 2 are applied in 6cfee8b.
Went one step further on removing the over-indexed comments and added _technical_ detail to the `ApplyMasterPasswordAuthenticationHashAsync` annotation in a672bdc. Also in this commit, renamed the `newPassword` argument to `newAuthenticationHash` for clarity -- it is called `newAuthenticationHash` up until it is hashed into the `User.MasterPassword` value, showing the separation and evolution of the value I think more deliberately. The through-line in to this callsite, as well as the added annotations, should make it clear both _what this is_ and _how we use it_.
This service should be a one-stop shop not just for MP needs, but for detailed documentation as well. 😸

JaredSnider-Bitwarden (approved):
Stellar work. 🏅 :shipit: !

PR #7531, opened 2026-04-22
[PM 33923]Add support for multiple organization discounts to organization subscription page

cyprain-okeke:
## 🎟️ Tracking

 https://bitwarden.atlassian.net/browse/PM-33923

## 📔 Objective

The organization subscription page previously only supported a single discount (CustomerDiscount). This PR updates the billing infrastructure to support multiple concurrent 
  discounts (CustomerDiscounts) — for example, a customer may have both a Milestone 2 migration discount and a Secrets Manager standalone discount at the same time.
                                                                                                                                                                               
  **Changes**                                                                                                                                                                      

  **API Response Models:**                                                                                                                                                         
  - OrganizationSubscriptionResponseModel — changed CustomerDiscount (single) to CustomerDiscounts (array). Organization pages pass through all discounts unfiltered so admins
  can see migration (M3/Families), SM standalone, and M2 discounts.                                                                                                            
  - SubscriptionResponseModel — changed to CustomerDiscounts array, still filtered to M2-only for the individual user subscription page.
                                                                                                                                                                               
  **Core Billing:**                                                                                                                                                              
  - SubscriptionInfo.CustomerDiscount → SubscriptionInfo.CustomerDiscounts (list).
  - StripePaymentService.GetSubscriptionAsync — collects customer-level and subscription-level discounts into a single list, deduplicating by coupon ID. Phase 2 schedule      
  discounts are merged/overridden rather than replacing the whole discount.                                                                                              
  - GetBitwardenSubscriptionQuery — cart-level and product-level discounts are now lists; schedule phase 2 discounts return multiple entries; tax estimation passes multiple   
  coupon IDs to Stripe's upcoming invoice API.                                                                                                                               
  - Cart / CartItem — Discount → Discounts (list) to model multiple discounts per item or per cart.   
## 📸 Screenshots

Screenshot 2026-04-22 at 1 51 17 PM
Screenshot 2026-04-22 at 4 29 53 PM
Screenshot 2026-04-22 at 3 23 20 PM
Screenshot 2026-04-22 at 2 14 46 PM


amorask-bitwarden (changes requested):
❌ There's quite a few changes here to the `GetBitwardenSubscriptionQuery`, which is currently only utilized for Premium. Multiple premium discounts is being addressed in [PM-35121](https://bitwarden.atlassian.net/browse/PM-35121), so I believe this PR's scope extends too far. Can you ensure the changes are specifically focused on surfacing multiple organization discounts?

src/Api/Billing/Controllers/OrganizationsController.cs
```diff
@@ -299,7 +299,7 @@ private async Task<Organization> AdjustOrganizationSeatsForSmTrialAsync(Guid id,
         }
 
         var subscriptionInfo = await paymentService.GetSubscriptionAsync(organization);
-        if (subscriptionInfo?.CustomerDiscount?.Id != StripeConstants.CouponIDs.SecretsManagerStandalone)
+        if (subscriptionInfo?.CustomerDiscounts.Any(d => d.Id == StripeConstants.CouponIDs.SecretsManagerStandalone) != true)
```
amorask-bitwarden:
```suggestion
        if (subscriptionInfo?.CustomerDiscounts.Any(d => d.Id != StripeConstants.CouponIDs.SecretsManagerStandalone))
```

src/Api/Models/Response/SubscriptionResponseModel.cs
```diff
@@ -29,9 +29,10 @@ public SubscriptionResponseModel(User user, SubscriptionInfo subscription, UserL
         License = license;
         Expiration = License.Expires;
 
-        CustomerDiscount = ShouldIncludeDiscount(subscription.CustomerDiscount)
-            ? new BillingCustomerDiscount(subscription.CustomerDiscount!)
-            : null;
+        CustomerDiscounts = subscription.CustomerDiscounts
```
amorask-bitwarden, line 31: ❌ `SubscriptionResponseModel` class was the old Premium subscription response approach before `GetBitwardenSubscriptionQuery` and will be significantly reduced as part of this PR: https://github.com/bitwarden/server/pull/7519

Since this model is not used for Organizations, I don't think the changes in this file are necessary. This is especially true since all the discounts are filtered to only show the price increase discount via `ShouldIncludeDiscount`.

amorask-bitwarden (changes requested):
Good work - still a few things we can remove though to tighten it up.

src/Api/Models/Response/SubscriptionResponseModel.cs
cyprain-okeke: This change was necessary to keep SubscriptionResponseModel compiling. Since SubscriptionInfo.CustomerDiscount (singular) was  replaced by CustomerDiscounts (a list) as part of this PR, every consumer of SubscriptionInfo needed to be updated to use the new property — including this model

amorask-bitwarden (approved):
Thanks for clearing up my question. Changes look good. Nice work!

JimmyVo16 (approved):

r-tome (approved):

PR #7536, opened 2026-04-23
[PM-34883] - Add InjectOrganizationUserAttribute

jrmccannon:
## 🎟️ Tracking
[PM-34883](https://bitwarden.atlassian.net/browse/PM-34883)

## 📔 Objective
This adds an attribute that can be used to decorate api endpoints and allow for the target organizationUser (via the route)
to be validated in a more standardized way. This will pull the id out of the route (id or whatever string is provided via the constructor) and  will validate the target OrganizationUser belongs to the organization from the route. Since we've fetched the org user already, we'll provide it to the controller method so we don't have to get it again.

I've implemented this on `organization/{orgId}/users/{id}/recover-account`.

[PM-34883]: https://bitwarden.atlassian.net/browse/PM-34883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -504,18 +506,17 @@ await _organizationService.UpdateUserResetPasswordEnrollmentAsync(
     /// </summary>
     [HttpPut("{id}/reset-password")]
     [Authorize<ManageAccountRecoveryRequirement>]
-    public Task<IResult> PutResetPassword(Guid orgId, Guid id, [FromBody] OrganizationUserResetPasswordRequestModel model)
-        => PutRecoverAccount(orgId, id, model);
+    [InjectOrganizationUser]
+    public Task<IResult> PutResetPassword(Guid orgId, Guid id, [FromBody] OrganizationUserResetPasswordRequestModel model,
+        [BindNever] OrganizationUser targetOrganizationUser)
```
eliykat: This is following Billing's example, but I think the dual attribute use - `Inject` and then `BindNever` - is awkward. I think we can achieve this through [custom model binding](https://learn.microsoft.com/en-us/aspnet/core/mvc/advanced/custom-model-binding?view=aspnetcore-10.0) which might better fit what we're trying to do here.

<details>
<summary>Untested example</summary>

The attribute:

```c#
  [AttributeUsage(AttributeTargets.Parameter)] 
  public sealed class InjectOrganizationUserAttribute(string organizationUserIdRouteParam = "id")
      : ModelBinderAttribute(typeof(OrganizationUserModelBinder))   
  {
      public string OrganizationUserIdRouteParam { get; } = organizationUserIdRouteParam; 
  }  
```

The binder:

```c#
public class OrganizationUserModelBinder : IModelBinder
  {                                                                                                                                                                                                                            
      public async Task BindModelAsync(ModelBindingContext bindingContext)
      {                                                                                                                                                                                                                        
          var attr = bindingContext.ModelMetadata.Attributes.ParameterAttributes                                 
              ?.OfType<InjectOrganizationUserAttribute>()                                                                                                                                                                        
              .FirstOrDefault()
              ?? new InjectOrganizationUserAttribute(); // this is only used with the attribute so not sure if this is appropriate
                                                                                                                                                                                                                               
          Guid orgId;
          try                                                                                                                                                                                                                  
          {                                                                                                      
              orgId = bindingContext.HttpContext.GetOrganizationId();
          }                                                                                                                                                                                                                    
          catch (InvalidOperationException)
          {
             // not sure if we really need to map this
             // you may also find a better way to return this out instead of throwing
              throw new BadRequestException("Route parameter 'orgId' or 'organizationId' is missing or invalid.");                                                                                                             
          }                                                                                                                                                                                                                    
                                                                                                                                                                                                                               
          var routeValues = bindingContext.ActionContext.RouteData.Values;                                                                                                                                                     
          if (!routeValues.TryGetValue(attr.OrganizationUserIdRouteParam, out var idValue)                       
              || !Guid.TryParse(idValue?.ToString(), out var orgUserId))                                                                                                                                                       
          {                                                                                                                                                                                                                    
              throw new BadRequestException($"Route parameter '{attr.OrganizationUserIdRouteParam}' is missing or invalid.");                                                                                                  
          }                                                                                                                                                                                                                    
                                                                                                                 
          var repo = bindingContext.HttpContext.RequestServices                                                                                                                                                                
              .GetRequiredService<IOrganizationUserRepository>();                                                
              
          var organizationUser = await repo.GetByIdAsync(orgUserId);                                                                                                                                                           
          if (organizationUser is null || organizationUser.OrganizationId != orgId)
          {                                                                                                                                                                                                                    
          }                                                                                                                                                                                                                    
   
          bindingContext.Result = ModelBindingResult.Success(organizationUser);                                                                                                                                                
      }                                                                                                          
  }
```

Usage:
```c#
  [HttpPut("{id}/recover-account")]                                                                                                                                                                                            
  [Authorize<ManageAccountRecoveryRequirement>]                                                                                                                                                                                
  public async Task<IResult> PutRecoverAccount(Guid orgId, Guid id,                                                                                                                                                            
      [FromBody] OrganizationUserResetPasswordRequestModel model,                                                
      [InjectOrganizationUser] OrganizationUser targetOrganizationUser)                                                                                                                                                          
  {
      // ...                                                                                                                                                                                                                   
  }
```
</details>

src/Api/AdminConsole/Attributes/InjectOrganizationUserAttribute.cs
```diff
@@ -0,0 +1,85 @@
...
+///
+/// [HttpPost("{organizationUserId}/accept")]
+/// [InjectOrganizationUser("organizationUserId")]
+/// public async Task<IResult> AcceptAsync(Guid organizationUserId, [BindNever] OrganizationUser organizationUser)
+/// ]]></code>
+/// </example>
+/// <param name="organizationUserIdRouteParam">
+/// Name of the route parameter containing the organization user ID. Defaults to <c>"id"</c>.
+/// </param>
+public class InjectOrganizationUserAttribute(string organizationUserIdRouteParam = "id") : ActionFilterAttribute
```
eliykat: Is there a way we could make this more reusable by other resources - e.g. we'll want similar validation for most of our controllers such as policies, collections, domains, etc. (The answer may be no - I suspect that trying to make it fully generic may be overcomplicated - but worth considering.)

eliykat (commented):
Looks good, I've left some feedback below to explore/consider, but this approach will work just fine as well.

src/Api/AdminConsole/Attributes/InjectOrganizationUserAttribute.cs
jrmccannon: I think it would be tough because there's also a chain of things that could be used to validate (such as validating the org id then the org user.) Maybe we try that after this and iterate on what we have.

eliykat (approved):

PR #7548, opened 2026-04-27
[PM-32211] fix private key before key rotation

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32211

## 📔 Objective

A user may have a corrupt private key. It needs to be fixed before they can rotate their keys or upgrade to v2 encryption. For this, we need to invoke the private key regeneration function and handle status changes for emergency access/organizations.

On key regeneration:
- Update the status of any emergency access grants
  - Invited -> Invited
  - Accepted -> Accepted
  - Confirmed -> Accepted
  - RecoveryInitiated → Accepted
  - RecoveryApproved → Accepted
  - Re-send the email to the emergency access grantor reminding them to confirm the grantee so that emergency access can be re-established
- Update the status of any organization memberships
  - Invited -> Invited
  - Accepted -> Accepted
  - Confirmed -> Accepted
  - Revoked -> Removed
  - In case the user is removed, fire the “user left the organization” event log so the organization has some traceability for what happened.

Other changes:
- The `DatabaseTransactionAction` now explicitly provides connection and transactions for EF, since it was discovered in integration testing for Sqlite, that in fact the action run in separate connection - to be confirmed it is testing or production issue too. Migration to this new data structure and pattern for other KM operations is outside of scope https://bitwarden.atlassian.net/browse/PM-35901


mzieniukbw:
Moving to draft, since i have discovered that the `DatabaseTransactionAction` pattern KM have been using for a while is non transactional in EF. The new pattern will be part of this PR, but only contained to the private key regeneration scope work. Further PR's will be created later for `UpdateUserData` (master password, kdf change) and `UpdateEncryptedDataForKeyRotation` (user key rotation) .

mzieniukbw:
Ready to review. Note, Lint and Tests workflows fail due to separate issue.

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_SetStatusToAccepted.sql
```diff
@@ -0,0 +1,16 @@
+CREATE PROCEDURE [dbo].[EmergencyAccess_SetStatusToAccepted]
```
mkincaid-bw: This does not follow our [naming conventions](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures), and should be something like `EmergencyAccess_UpdateManySetStatus`

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_SetStatusToAccepted.sql
```diff
@@ -0,0 +1,16 @@
...
+    @Ids [dbo].[GuidIdArray] READONLY
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @UtcNow DATETIME2(7) = GETUTCDATE()
+
+    UPDATE EA
```
mkincaid-bw: The `Status` and `RevisionDate` values should be passed in as parameters from the application. Specifying them here accounts to having business logic in the database, which we're not supposed to do.

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_SetStatusToAccepted.sql
```diff
@@ -0,0 +1,16 @@
...
+    SET EA.[Status] = 1, -- Accepted
+        EA.[KeyEncrypted] = NULL,
+        EA.[RevisionDate] = @UtcNow
+    FROM [dbo].[EmergencyAccess] EA
+    INNER JOIN @Ids I ON I.[Id] = EA.[Id]
```
mkincaid-bw: ⛏️ SQL Keywords should be on their own line
https://contributing.bitwarden.com/contributing/code-style/sql/#update-statements

src/Sql/dbo/Stored Procedures/OrganizationUser_SetStatusToAccepted.sql
```diff
@@ -0,0 +1,18 @@
+CREATE PROCEDURE [dbo].[OrganizationUser_SetStatusToAccepted]
```
mkincaid-bw: Same comments as `EmergencyAccess_SetStatusToAccepted`.

mkincaid-bw (changes requested):
A few changes needed.

src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
```diff
@@ -146,4 +146,18 @@ UpdateEncryptedDataForKeyRotation UpdateForKeyRotation(Guid userId,
     /// <param name="organizationUser">The organization user entity with updated properties (status, userId, key)</param>
     /// <returns>An action that can be executed within a transaction</returns>
     Func<DbConnection, DbTransaction, Task> BuildConfirmOwnerAction(OrganizationUser organizationUser);
+
+    /// <summary>
+    /// Returns a delegate that sets organization users to Accepted status and clears Key.
```
quexten:
```suggestion
    /// Returns a delegate that sets organization user status to accepted, and clears their copy of the shared organization key.
```

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -1058,6 +1058,81 @@ public Func<DbConnection, DbTransaction, Task> BuildConfirmOwnerAction(Core.Enti
...
+            {
+                return;
+            }
+
+            using var scope = ServiceScopeFactory.CreateScope();
+            var dbContext = GetTransactionalDatabaseContext(scope, connection, transaction);
+
+            await dbContext.UserBumpAccountRevisionDateByOrganizationUserIdsAsync(ids);
+
+            await dbContext.CollectionUsers
```
quexten: One concern I have here is that this will likely get forgotten if other teams ever implement more org functionality. Can we extract this, and the regular `DeleteAsync` functionality to run the same queries so that a change *must* be done on all at the same time?

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_SetStatusToAccepted.sql
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/650a9fdb90a15af3f9cb3c3cae1a50086d6e526f

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_SetStatusToAccepted.sql
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/650a9fdb90a15af3f9cb3c3cae1a50086d6e526f

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_SetStatusToAccepted.sql
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/650a9fdb90a15af3f9cb3c3cae1a50086d6e526f

src/Sql/dbo/Stored Procedures/OrganizationUser_SetStatusToAccepted.sql
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/650a9fdb90a15af3f9cb3c3cae1a50086d6e526f

src/Core/AdminConsole/Repositories/IOrganizationUserRepository.cs
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/650a9fdb90a15af3f9cb3c3cae1a50086d6e526f

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/ec60e1869a788890deb8ab4050b79b3481a15a85.
Now both Dapper and EF re-use existing delete business logic for our new function.

mkincaid-bw (dismissed):

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
quexten: https://github.com/bitwarden/server/blob/d2f78d9abb3f8876d5c5911cdabe37ddc3bd1961/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs#L127-L153

https://github.com/bitwarden/server/blob/d2f78d9abb3f8876d5c5911cdabe37ddc3bd1961/src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs#L1060-L1135

These still seem to have duplicate logic? Can we join these?

jrmccannon (dismissed):
Interesting way of passing the transaction around.  I am working on a different branch of making the transaction usable via DI.

I would appreciate any feedback you might have on it.  
https://github.com/bitwarden/server/pull/7555

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
mzieniukbw: I already did that in https://github.com/bitwarden/server/pull/7548/changes/ec60e1869a788890deb8ab4050b79b3481a15a85, please re-check on newest changes.

quexten (dismissed):

src/Core/KeyManagement/Commands/RegenerateUserAsymmetricKeysCommand.cs
```diff
@@ -55,17 +69,60 @@ x.Status is EmergencyAccessStatusType.Confirmed or EmergencyAccessStatusType.Rec
...
+        }
+
+        foreach (var ea in eaToReset)
+        {
+            if (ea.GranteeEmail is null || ea.GrantorEmail is null)
+            {
+                continue;
+            }
+
+            await _mailService.SendEmergencyAccessAcceptedEmailAsync(
```
ike-kottlowski, line 125: Do we warn the user that they will be receiving this email in any of the client flows? If I randomly got an email saying something about an emergency access update I'd be concerned for the security of my account.

mzieniukbw: I confirmed this with PM, we don't want to expand scope at this time. This will send an email to grantor and only on cases when the grantor was previously invited.
Also note, the private key regeneration is only needed for a fraction of users (in few thousands count at best) and is necessary step in password manager functionality working properly.

ike-kottlowski (dismissed):

jrmccannon (dismissed):

quexten (dismissed):

mkincaid-bw (dismissed):

src/Infrastructure.Dapper/Auth/Repositories/EmergencyAccessRepository.cs
```diff
@@ -177,6 +178,32 @@ INNER JOIN
         };
     }
 
+    /// <inheritdoc />
+    public DatabaseTransactionAction SetStatusToAcceptedForPublicKeyPairRegeneration(IEnumerable<EmergencyAccess> emergencyAccesses)
```
JaredSnider-Bitwarden: ⚠️ :

## Issues with the current approach

1. **CQS layering violation.** The command ([RegenerateUserAsymmetricKeysCommand.cs:74-83](src/Core/KeyManagement/Commands/RegenerateUserAsymmetricKeysCommand.cs#L74-L83)) filters entities but never mutates them. The actual state contract — "key regen sets `Status = Accepted`, `KeyEncrypted = null`, `RevisionDate = now`" — lives inside the repo ([EF](src/Infrastructure.EntityFramework/Auth/Repositories/EmergencyAccessRepository.cs#L196-L201), [Dapper](src/Infrastructure.Dapper/Auth/Repositories/EmergencyAccessRepository.cs#L194-L203)). A reader of the command can't see what key-regen does to an `EmergencyAccess`.

2. **Single-purpose method on a shared repo.** `SetStatusToAcceptedForPublicKeyPairRegeneration` names the *calling flow*, not the *shape* of the operation. It can't be reused — every new flow that needs a similar update adds another bespoke method. Same problem applies to the sibling `OrganizationUserRepository` methods.

3. **Business decision duplicated across ORMs.** EF and Dapper each re-encode the same three-column mutation. Two places to keep in sync, two places to get wrong.

4. **Diverges from the existing write convention.** `EmergencyAccessService` already uses full-row `ReplaceAsync` everywhere ([EmergencyAccessService.cs](src/Core/Auth/UserFeatures/EmergencyAccess/EmergencyAccessService.cs) — 9+ call sites). The targeted-update safety property this method provides (narrow write set) isn't a property the rest of the codebase relies on for this entity.

## What to do

Replace the per-flow methods with a generic transactional bulk replace:

- **Command** owns the mutation:

    ```csharp
    foreach (var ea in eaToReset)
    {
        ea.Status = EmergencyAccessStatusType.Accepted;
        ea.KeyEncrypted = null;
        ea.RevisionDate = utcNow;
    }
    updateDataActions.Add(_emergencyAccessRepository.ReplaceManyAsync(eaToReset));
    ```

- **Repository** exposes one method per entity:

    ```csharp
    DatabaseTransactionAction ReplaceManyAsync(IEnumerable<EmergencyAccess> entities);
    ```

    Same shape as the existing `ReplaceAsync` — just batched and transaction-aware.

- **Apply the pattern uniformly** to the `OrganizationUserRepository` siblings (`SetStatusToAcceptedForPublicKeyPairRegeneration`, `RemoveForPublicKeyPairRegeneration`) if agreed upon by the AC codeowners (cc: @jrmccannon) so the command is the single place describing what key regeneration does to each entity type.  

- **Concurrent-clobber risk** is the only real argument for the current targeted update, and it's already accepted by `ReplaceAsync` everywhere else in this entity's service. Not a blocker.

Net result: command reads as a self-contained description of the domain operation, repo surface stays generic and reusable, and the EF/Dapper implementations become straightforward bulk-update-by-id with no flow-specific knowledge.

JaredSnider-Bitwarden (changes requested):
Thank you for the change set, but I think we have to discuss a more extensible approach to modifying the EA repository layer 👇

src/Infrastructure.Dapper/Auth/Repositories/EmergencyAccessRepository.cs
JaredSnider-Bitwarden: This also centralizes domain logic testing as well.

JaredSnider-Bitwarden: Following up on this thread — appreciate the back-and-forth in our DM, @mzieniukbw. We landed on a position together and I want to summarize it here for anyone reading later.

**What I'm withdrawing from my original comment**

The `ReplaceMany` / full-row update recommendation. Two reasons made me reconsider:

1. **Concurrent-write clobber risk.** On entities like `OrganizationUser` (and `User` in upcoming forced key rotation flows), unrelated concurrent operations — permission changes, profile updates, etc. — can interleave with a key-handling write. Full-row replace would let one flow overwrite another's columns, and there's existing empirical signal that this class of bug is already contributing to occasional decryption issues. Targeted column updates are a safer default for key-handling paths, not just a stylistic choice.
2. **Scaling cost.** Forced key rotation will touch every user. A full-row replace there is meaningful unnecessary write load and amplifies the concurrency footprint. The targeted update keeps both small.

So the existing targeted-update + `DatabaseTransactionAction` shape is the right foundation. I shouldn't have argued past it on safety grounds.

**What we agreed could still improve, without touching the safety story**

The two concerns I raised — keeping business logic in the command, and keeping the repository surface generic — are orthogonal to the row-replace question. They can be satisfied while preserving the targeted-update behavior:

1. **Pass the mutation values as parameters** instead of hard-coding them inside the repository method. The command names `Accepted`, `null`, and `utcNow` at the call site; the repository just persists the columns it's told to.
2. **Narrow the parameter to `ids`** rather than the entity list. The current implementation only reads `.Id` off each entity anyway, so the narrower signature is honest about what the repo actually needs (and the command keeps the full entity list locally for downstream uses like email notifications).
3. **Rename the method to describe the operation** rather than the calling flow — e.g. `UpdateStatusAndKeyEncryptedByIdsAsync(ids, status, keyEncrypted, revisionDate)`. Matches our existing convention (`UpdateLastActivityByIdAsync`, `UpdateRenewalReminderDateAsync`, `UpdateUserKeyAndEncryptedDataAsync`) and decouples the persistence method from this single caller, so a future flow needing the same column subset can reuse it.

~~Same SQL. Same transactional shape~~. Same safety properties. Just a cleaner contract between command and repository.  Note: same SQL was incorrect as the SQL would need to change to accept params. 

**Where this applies**

Same shape change applies to the `OrganizationUserRepository` siblings (`SetStatusToAcceptedForPublicKeyPairRegeneration`, `RemoveForPublicKeyPairRegeneration`) by the same logic.  (cc: @jrmccannon)

**Why this matters to me**

I want to keep our Auth repositories as clean and limited as possible and centralize the business logic in commands — that gives us proper encapsulation for unit testing and keeps the persistence layer reusable across future flows. The proposed change preserves the safety properties Maciej is rightly defending while moving the "what changes and to what value" decision back into the command, where it's easier to reason about and test.

Thanks for pushing back — the resulting design is better than where I started.

mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/ea1c412147864c8bf9017e4d88273948354edf4c
I went with update by singular id instead, since there is one key per one id, so could not use multiple ids with one key.

JaredSnider-Bitwarden (dismissed):
Thank you very much for making these changes last min. Auth files LGTM!

quexten (dismissed):

jrmccannon (dismissed):
Looks good from the AC perspective. We are currently in the process of adding staged as a new org user status. Tagging @sven-bitwarden for awareness.

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_UpdateStatusAndKeyEncryptedById.sql
```diff
@@ -0,0 +1,19 @@
+CREATE PROCEDURE [dbo].[EmergencyAccess_UpdateStatusAndKeyEncryptedById]
```
mkincaid-bw: We shouldn't use `And` between parameters in stored procedure names ([see docs](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures:~:text=And%20BETWEEN%20PARAMETER%20NAMES%20IN%20PROCEDURE%20NAMES)).

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateStatusAndKeyById.sql
```diff
@@ -0,0 +1,19 @@
+CREATE PROCEDURE [dbo].[OrganizationUser_UpdateStatusAndKeyById]
```
mkincaid-bw: Same comment as above.

mkincaid-bw (changes requested):
Minor naming issue.

src/Sql/dbo/Auth/Stored Procedures/EmergencyAccess_UpdateStatusAndKeyEncryptedById.sql
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/a2c147f4ebb2f3c68f74a46bee792c74f07867ce

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateStatusAndKeyById.sql
mzieniukbw: Fixed in https://github.com/bitwarden/server/pull/7548/commits/a2c147f4ebb2f3c68f74a46bee792c74f07867ce

mkincaid-bw (approved):
LGTM

JaredSnider-Bitwarden (approved):

quexten (approved):

jrmccannon (approved):

PR #7551, opened 2026-04-27
PM-35503 fixed flaky tests due to timing issue.

prograhamming:
## 🎟️ Tracking

[PM-35503](https://bitwarden.atlassian.net/browse/PM-35503?search_id=1963fc16-dd79-4823-8d67-fc11b820e67d)

## 📔 Objective

`RedirectAsync_Success` and `RedirectAsync_IntegrationAlreadyExistsWithNullConfig_Success` within `SlackIntegrationControllerTests.cs` are flaky because they rely on two separate calls to `TimeProvider.GetUtcNow()` returning the same Unix second. `IntegrationOAuthState.ToString()` embeds the timestamp, so if a second boundary is crossed between the test's `FromIntegration` call and the controller's internal `FromIntegration` call, the state strings differ and the assertion fails.

## 📸 Screenshots

[PM-35503]: https://bitwarden.atlassian.net/browse/PM-35503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Banrion (dismissed):

Banrion (approved):
✅ Approving but there is one build error

PR #7552, opened 2026-04-27
[PM-32100] Implement Multi-Provider Ability Lookup

JimmyVo16:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32100

## 📔 Objective
1. Implement Multi-Provider Ability functionality
2. Add unit tests.

I think the unit test coverage is sufficient. No manual testing is needed.


src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
```diff
@@ -13,14 +13,28 @@ public class ExtendedProviderAbilityCacheService(
 {
     public const string CacheName = "ProviderAbilities";
 
-    public async Task<ProviderAbility?> GetProviderAbilityAsync(Guid providerId)
+    public async Task<ProviderAbility?> GetProviderAbilityAsync(Guid providerId, CancellationToken cancellationToken = default)
```
eliykat, line 16: Do we expect to use `CancellationToken`? I don't mind, just curious.

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
```diff
@@ -13,14 +13,28 @@ public class ExtendedProviderAbilityCacheService(
...
+            async (_, _) => await providerRepository.GetAbilityAsync(providerId),
+            token: cancellationToken
         );
     }
 
+    public async Task<Dictionary<Guid, ProviderAbility>> GetProviderAbilitiesAsync(IEnumerable<Guid> providerIds, CancellationToken cancellationToken = default)
+    {
+        var tasks = providerIds
+            .Distinct()
+            .Select(async providerId => (providerId, ability: await GetProviderAbilityAsync(providerId, cancellationToken)));
```
eliykat: Returning a tuple here is a bit awkward and unnecessary. The `providerId` is already present on `ability.Id`. e.g.

```c#
        var tasks = providerIds
            .Distinct()
            .Select(providerId => GetProviderAbilityAsync(providerId, cancellationToken));

        var results = await Task.WhenAll(tasks);

        return results
            .Where(ability => ability != null)
            .ToDictionary(ability => ability.Id);
```

test/Core.Test/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheServiceTests.cs
```diff
@@ -0,0 +1,127 @@
...
+public class ExtendedProviderAbilityCacheServiceTests
+{
+    [Theory, BitAutoData]
+    public async Task GetProviderAbilitiesAsync_ReturnsAbilitiesForAllIds(
+        SutProvider<ExtendedProviderAbilityCacheService> sutProvider,
+        ProviderAbility ability1,
+        ProviderAbility ability2)
+    {
+        // Arrange
+        SetupCacheReturns(sutProvider, ability1, ability2);
```
eliykat: It would be more useful here to seed some abilities that will not be returned as well.

eliykat (changes requested):

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
JimmyVo16: It’s up to the calling code to pass it in, and generally we don’t need it. But since it’s low effort, I feel like we should support it because the extended cache does.

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
JimmyVo16: This is how the org cache does it. I think both patterns are pretty similar in this context, even though your suggestion is less verbose, which is good.

I’m inclined to stick with the existing pattern to stay aligned with the org cache for consistency. Unless you feel strongly about it, then maybe we should change the org cache as well?

test/Core.Test/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheServiceTests.cs
```diff
@@ -13,21 +13,25 @@ namespace Bit.Core.Test.AdminConsole.AbilitiesCache;
...
+        ProviderAbility nonExistAbility1,
+        ProviderAbility nonExistAbility2)
     {
         // Arrange
-        SetupCacheReturns(sutProvider, ability1, ability2);
+        SetupCacheReturns(sutProvider, existAbility1, existAbility2, nonExistAbility1, nonExistAbility2);
 
         // Act
-        var result = await sutProvider.Sut.GetProviderAbilitiesAsync([ability1.Id, ability2.Id]);
+        var result = await sutProvider.Sut.GetProviderAbilitiesAsync([existAbility1.Id, existAbility2.Id, nonExistAbility1.Id, nonExistAbility2.Id]);
```
eliykat: This test is failing - you're seeding all abilities and then fetching all abilities, but asserting (incorrectly) that not all were returned.

My suggestion was not about what abilities exist or do not exist (as your naming suggests). It's about ensuring that only the IDs specified are returned, and IDs that are not specified are not returned.

This can be achieved by:
- removing `nonExistAbility1` and `nonExistAbility2` from your `GetProviderAbilitiesAsync` call (this will make the test pass)
- updating the variable naming - all abilities exist, they're just not all being queried.

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
eliykat: It's not a blocker if you ultimately disagree, but fwiw I do think it's worth changing here - it's minor but it makes a big impact on readability. That said, I see the equivalent org cache ticket has already passed QA, so it's not worth triggering QA on that again. I think a small divergence in the implementations so that one of them is clearer is OK.

eliykat (changes requested):

src/Core/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheService.cs
JimmyVo16: I'm ok with that. My main concern was having to push the org cache through QA again. I made the change.

test/Core.Test/AdminConsole/AbilitiesCache/ExtendedProviderAbilityCacheServiceTests.cs
JimmyVo16: I fixed the test and included our original request. Let me know what you think.

eliykat (approved):

PR #7555, opened 2026-04-27
[PM-25520] - Adding a transaction for mutliple repo calls

jrmccannon:
## 🎟️ Tracking
[PM-25520](https://bitwarden.atlassian.net/browse/PM-25520)

## 📔 Objective
Adding a transaction around subsequent repo calls.

## 📸 Screenshots

[PM-25520]: https://bitwarden.atlassian.net/browse/PM-25520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrmccannon:
Test is failing due to the fact that its creating a SqlServerTestDatabase and it won't be able to pull the connection string to do so. Unsure of how to proceed with this. 

Do we want to add a SQL server db for this test or just skip it in CI?

@eliykat @justindbaur

eliykat:
> Test is failing due to the fact that its creating a SqlServerTestDatabase and it won't be able to pull the connection string to do so. Unsure of how to proceed with this.
> 
> Do we want to add a SQL server db for this test or just skip it in CI?

We definitely don't want to skip it in CI, or it'll never be run in practice.

The CI job does set up and connect to SQL server, e.g. https://github.com/bitwarden/server/blob/master/.github/workflows/test-database.yml#L131-L132.

I assume your issue is caused by the `DapperTransactionManager` getting its connection string from `GlobalSettings`. Have a look at how `DatabaseDataAttribute` works - it injects the integration test database strings into the repository implementations during tests, maybe you can do a similar thing here.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
```diff
@@ -45,6 +48,8 @@ public class PostUserCommand(
         Guid organizationId,
         ScimProviderType scimProvider)
     {
+        await using var transactionScope = await transactionManager.BeginTransactionAsync(IsolationLevel.Serializable);
```
mzieniukbw, line 51: ❓ `Serializable` transaction isolation is the strictest you can get, but it have side effects compared to default, like it can throw error if there was a concurrent write (in different transaciton). Is this by choice here ?
On a side note, this changes the business logic, which i wonder, whether this PR should do.

In most cases this transaction isolation is unnecessary, in which case, it is better to begin the transaction just before the first write, not during the get.

One important note, the copy paste code trend is unavoidable and we will see other places using this isolation level for no reason, just because it was used somewhere else. If we would use serializable isolation level by choice here, i think a reasonable comment should be added explaining why it might be needed here.

@bitwarden/dept-dbops Opinions ?

src/Infrastructure.EntityFramework/Data/EfTransactionManager.cs
```diff
@@ -0,0 +1,50 @@
...
+    public bool HasActiveTransaction => TransactionState.Current is not null;
+
+    public async Task<ITransactionScope> BeginTransactionAsync(
+        IsolationLevel isolationLevel = IsolationLevel.ReadCommitted,
+        CancellationToken cancellationToken = default)
+    {
+        var existing = TransactionState.Current;
+        if (existing is not null)
+        {
+            existing.ReferenceCount++;
```
mzieniukbw: ❓ Since `NestedTransactionScope` takes care of decreasing the reference count on disposal, shouldn't it also own increasing reference count ?
(Same for dapper)

mzieniukbw (commented):
Nice, i love the direction. There is a bit duplication between EF and Dapper, but generally good direction in my opinion.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
jrmccannon: The isolation is needed so that we can get the current seat count and ensure that it won't change before we complete our request. 

For this specific command, multiple requests come into SCIM grab the organization at the same time and attempt to add their user. Currently, the orgservice#inviteuser method grabs resources multiple times including calling out to stripe. An optimized command for inviting users was created, but that introduced an issue where multiple requests would race to update the seat count and all set it to the same value instead of incrementing it.

3 requests come in => get Org: { Seats: 2 } and each add a seat to the org.Seat value. Each request would set org.Seat to 3 when in reality they should set it to 5 (assuming both Org.Seats are occupied). There's also the ability for an Org to set a limit (Org.MaxSeats: 4). So two should succeed and one should fail.

The SCIM protocol requires the server tell it whether the user was safely provisioned so we have to succeed or fail and can't just switch it to an asynchronous add.

With serialized, this would ensure that the seat count is updated correctly as the requests come in.

So that was the reason for this approach. I'd be interested in another way as this is very heavy handed.

src/Infrastructure.EntityFramework/Data/EfTransactionManager.cs
jrmccannon: Good point this isn't even read anywhere so I'll remove it and add a base class the transaction manager.

src/Infrastructure.Dapper/Repositories/BaseRepository.cs
```diff
@@ -29,4 +32,72 @@ public BaseRepository(string connectionString, string readOnlyConnectionString)
...
+        }
+
+        return (new SqlConnection(ConnectionString), null, true);
+    }
+
+    /// <summary>
+    /// Executes an action using the ambient transaction connection (if active) or a new
+    /// owned connection. The connection is opened and disposed automatically when owned.
+    /// </summary>
+    protected async Task<TResult> ExecuteWithConnectionAsync<TResult>(
```
mzieniukbw, line 61: I think we need solid integration test coverage with `ITransactionManager` and without it, for botj Dapper and EF. Include all edge cases and databases. Without it, it's hard to say if it is working as expected / stable to use in prod.

src/Infrastructure.Dapper/Repositories/BaseRepository.cs
```diff
@@ -29,4 +32,72 @@ public BaseRepository(string connectionString, string readOnlyConnectionString)
 
     protected string ConnectionString { get; private set; }
     protected string ReadOnlyConnectionString { get; private set; }
+
+    /// <summary>
+    /// Returns the ambient connection and transaction if an ambient transaction is active,
+    /// or creates a new owned connection. The caller must dispose the connection only if
+    /// <c>Owned</c> is true.
+    /// </summary>
+    protected (SqlConnection Connection, DbTransaction? Transaction, bool Owned) GetConnection()
```
justindbaur: Should this be private? Looks like it should just be a helper for the methods below.

src/Infrastructure.EntityFramework/Repositories/BaseEntityFrameworkRepository.cs
```diff
@@ -31,6 +32,58 @@ public DatabaseContext GetDatabaseContext(IServiceScope serviceScope)
...
+    /// </summary>
+    private (DatabaseContext DbContext, IServiceScope? OwnedScope) GetDatabaseContextOrAmbient()
+    {
+        var holder = TransactionState.Current;
+        if (holder?.DbContext is DatabaseContext ambientContext)
+        {
+            return (ambientContext, null);
+        }
+
+        var scope = ServiceScopeFactory.CreateScope();
```
justindbaur: This way both the begin transaction flow and this have the same type of scope.

```suggestion
    private (DatabaseContext DbContext, AsyncServiceScope? OwnedScope) GetDatabaseContextOrAmbient()
    {
        var holder = TransactionState.Current;
        if (holder?.DbContext is DatabaseContext ambientContext)
        {
            return (ambientContext, null);
        }

        var scope = ServiceScopeFactory.CreateAsyncScope();
```

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -102,10 +105,12 @@ public static SupportedDatabaseProviders AddDatabaseRepositories(this IServiceCo
         if (provider != SupportedDatabaseProviders.SqlServer)
         {
             services.AddPasswordManagerEFRepositories(globalSettings.SelfHosted);
+            services.AddSingleton<ITransactionManager, EfTransactionManager>();
```
justindbaur:
```suggestion
            services.TryAddSingleton<ITransactionManager, EfTransactionManager>();
```

src/SharedWeb/Utilities/ServiceCollectionExtensions.cs
```diff
@@ -102,10 +105,12 @@ public static SupportedDatabaseProviders AddDatabaseRepositories(this IServiceCo
...
         }
         else
         {
             services.AddDapperRepositories(globalSettings.SelfHosted);
+            services.AddSingleton<ITransactionManager, DapperTransactionManager>();
```
justindbaur: Same
```suggestion
            services.TryAddSingleton<ITransactionManager, DapperTransactionManager>();
```

src/Infrastructure.EntityFramework/Repositories/BaseEntityFrameworkRepository.cs
```diff
@@ -31,6 +32,58 @@ public DatabaseContext GetDatabaseContext(IServiceScope serviceScope)
...
+    protected async Task ExecuteWithContextAsync(Func<DatabaseContext, Task> action)
+    {
+        var (dbContext, ownedScope) = GetDatabaseContextOrAmbient();
+        try
+        {
+            await action(dbContext);
+        }
+        finally
+        {
+            ownedScope?.Dispose();
```
justindbaur: This would become: 
```suggestion
            if (ownedScope is not null)
            {
                await ownedScope.DisposeAsync();
            }
```

src/Infrastructure.Dapper/Repositories/BaseRepository.cs
justindbaur, line 61: This is worth a comment saying that this will never use the readonly connection string and you should not use this method if you want to use that. I could see most repository methods switching to this as the default but want some sort of warning that connections using the readonly connection string will not want to move to this.

src/Core/Platform/Data/RootTransactionScope.cs
```diff
@@ -0,0 +1,43 @@
...
+    private bool _disposed;
+
+    public RootTransactionScope(TransactionHolder holder)
+    {
+        _holder = holder;
+    }
+
+    public async Task CommitAsync(CancellationToken cancellationToken = default)
+    {
+        if (_holder.Doomed)
```
justindbaur, line 15:
```suggestion
        ObjectDisposedException.ThrowIf(_disposed, this);

        if (_holder.Doomed)
```

src/Core/Platform/Data/RootTransactionScope.cs
```diff
@@ -0,0 +1,43 @@
...
+                "Cannot commit a transaction that has been marked for rollback by a nested scope.");
+        }
+
+        _holder.Committed = true;
+        await _holder.Transaction.CommitAsync(cancellationToken);
+    }
+
+    public async Task RollbackAsync(CancellationToken cancellationToken = default)
+    {
+        _holder.Doomed = true;
```
justindbaur: Same here

```suggestion
        ObjectDisposedException.ThrowIf(_disposed, this);

        _holder.Doomed = true;
```

src/Core/Platform/Data/ITransactionManager.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// outer transaction's isolation level.
+    /// </summary>
+    Task<ITransactionScope> BeginTransactionAsync(
+        IsolationLevel isolationLevel = IsolationLevel.ReadCommitted,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Returns true if the current async flow has an active ambient transaction.
+    /// </summary>
+    bool HasActiveTransaction { get; }
```
justindbaur: It doesn't look like this is used. I'd remove it.

bitwarden_license/src/Scim/Users/PostUserCommand.cs
mkincaid-bw: I agree that serializable isolation level should be avoided if possible. A couple other options are:

1. Query the data with an UPDATE lock, which locks only that row rather than everything involved in the transaction.
  - Pros: Simple, no schema changes needed
  - Cons: You have to write/maintain 4 separate queries (one for each DB)
2. Use a row version column
  - Pros: Supported natively in SQL Server and EF, can be reused for future concurrency checks
  - Cons: Requires new column on table

I'd recommend the row version column, since it requires minimal code changes and retains optimistic concurrency. The main caveat is that instead of blocking other callers upfront, each caller proceeds as normal and you deal with race conditions (if any) afterwards, like this:

1. Read the org row — you get back Seats = 9 and RowVersion = 42
2. Check the seat limit in C# — passes
3. Issue the update: UPDATE Organization SET Seats = 10 WHERE Id = @id AND RowVersion = 42
4. If 1 row affected — you won the race, commit
5. If 0 rows affected — someone else updated the row between your read and your write; RowVersion is no longer 42, so the WHERE clause matched nothing

Step 5 is the conflict. EF Core detects the 0-rows-affected case and throws DbUpdateConcurrencyException.
```csharp
try
{
    await _organizationRepository.IncrementSeatCountAsync(org);
}
catch (DbUpdateConcurrencyException)
{
    // Option A: retry — reload the org and try again
    // Option B: surface a conflict error to the SCIM caller
    throw new ConflictException("Organization seat count changed concurrently, please retry.");
}
```

src/Infrastructure.EntityFramework/Repositories/BaseEntityFrameworkRepository.cs
```diff
@@ -31,6 +32,58 @@ public DatabaseContext GetDatabaseContext(IServiceScope serviceScope)
...
+    private (DatabaseContext DbContext, IServiceScope? OwnedScope) GetDatabaseContextOrAmbient()
+    {
+        var holder = TransactionState.Current;
+        if (holder?.DbContext is DatabaseContext ambientContext)
+        {
+            return (ambientContext, null);
+        }
+
+        var scope = ServiceScopeFactory.CreateScope();
+        return (GetDatabaseContext(scope), scope);
```
justindbaur: There is a bug happening here if secrets manager code tries to use a transaction, it would just silently not use a transaction when it could. 

```suggestion
        var context = GetDatabaseContext(scope);
        if (holder is not null)
        {
            context.UseTransaction(holder.Transaction);
            holder.DbContext = context;
            holder.Scope = scope;
        }
        return (context, scope);
```

bitwarden_license/src/Scim/Users/PostUserCommand.cs
jrmccannon: I have a branch i was trying something similar with Polly in order to not just fail everything. I'll give that a try.

eliykat: @mkincaid-bw I would also prefer to avoid SERIALIZABLE and use an optimistic solution (even at the expense of a bit of hand-crafting), however we went down this path at @withinfocus 's direction. I would prefer Architecture's confirmation before we spend more time exploring other solutions. (Thoughts on the above suggestions @withinfocus ?)

PR #7556, opened 2026-04-27
[PM-35948] Update send openapi to work for sdk

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35948

## 📔 Objective

Update the functions to have actual return values instead of void returns

Update the endpoints to properly handle the different auth types needed

SDK changes: https://github.com/bitwarden/sdk-internal/pull/961


src/Api/Tools/Controllers/SendsController.cs
```diff
@@ -70,7 +70,7 @@ IHasPremiumAccessQuery hasPremiumAccessQuery
 
     [AllowAnonymous]
     [HttpPost("access/{id}")]
-    public async Task<IActionResult> Access(string id, [FromBody] SendAccessRequestModel model)
+    public async Task<SendAccessResponseModel> Access(string id, [FromBody] SendAccessRequestModel model)
```
harr1424, line 77: ❓  Should this method be decorated with `ProducesResponseType` similar to those that follow?

src/Api/Tools/Controllers/SendsController.cs
```diff
@@ -96,7 +96,7 @@ public async Task<IActionResult> Access(string id, [FromBody] SendAccessRequestM
             await _sendAuthorizationService.AccessAsync(send, model.Password);
         if (sendAuthResult.Equals(SendAccessResult.PasswordRequired))
         {
-            return new UnauthorizedResult();
+            throw new UnauthorizedAccessException();
```
coroiu, line 103: suggestion(non-blocking): Align https://github.com/bitwarden/server/blob/f25cdafe12acd748b3521355f23405c59f0ec7ec/src/Api/Tools/Controllers/SendsController.cs#L157 to also throw for consistency (they return different response bodies)

suggestion(non-blocking): Tests don't check this unhappy path, consider adding new ones

src/Api/Tools/Controllers/SendsController.cs
coroiu: Agreed, I'm also wondering this

question: was dropping the `[ProducesResponseType]` annotations for 400/401/404 intentional here? The other three send-access endpoints in this commit still declare them ([GetSendFileDownloadData](https://github.com/bitwarden/server/blob/f25cdafe12acd748b3521355f23405c59f0ec7ec/src/Api/Tools/Controllers/SendsController.cs#L125-L128), [AccessUsingAuth](https://github.com/bitwarden/server/blob/f25cdafe12acd748b3521355f23405c59f0ec7ec/src/Api/Tools/Controllers/SendsController.cs#L238-L240), [GetSendFileDownloadDataUsingAuth](https://github.com/bitwarden/server/blob/f25cdafe12acd748b3521355f23405c59f0ec7ec/src/Api/Tools/Controllers/SendsController.cs#L288-L290)), so `Access` ends up as the only endpoint without an error matrix in the generated OpenAPI spec. Doesn't affect SDK functionality (error handling is generic) but it's an asymmetry in the docs

src/Api/Tools/Controllers/SendsController.cs
```diff
@@ -117,11 +117,15 @@ public async Task<IActionResult> Access(string id, [FromBody] SendAccessRequestM
             sendResponse.CreatorIdentifier = creator.Email;
         }
 
-        return new ObjectResult(sendResponse);
+        return sendResponse;
     }
 
     [AllowAnonymous]
     [HttpPost("{encodedSendId}/access/file/{fileId}")]
+    [ProducesResponseType<SendFileDownloadDataResponseModel>(StatusCodes.Status200OK)]
```
coroiu, line 129: question: any reason this endpoint kept `Task<IActionResult>` + the explicit 200 annotation, instead of being refactored to `Task<SendFileDownloadDataResponseModel>` like `Access`? The only thing blocking that is the `UnauthorizedResult` return — same `throw` switch as `Access` would unify the two.

src/Api/Tools/Controllers/SendsController.cs
adudek-bw: Good catch! Yes it should!

harr1424 (dismissed):

src/SharedWeb/Swagger/SendAccessBearerOperationFilter.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Bit.SharedWeb.Swagger;
+
+/// <summary>
+/// Replaces the global security requirement with the <c>send-access-bearer</c> scheme for
+/// endpoints that require the <see cref="Policies.Send"/> authorization policy. This causes
+/// the OpenAPI generator to emit an explicit Bearer token parameter rather than injecting the
+/// user session token via middleware.
+/// </summary>
+public class SendAccessBearerOperationFilter : IOperationFilter
```
coroiu: question/issue: Have you generated the bindings to confirm this fix works?

From what I can see using Claude the V2 operations don' get the `send-access-bearer` security in the generated `api.json`. Verified by running `pwsh dev/generate_openapi_files.ps1` on this branch, the V2 operations have `security: [{}]` (empty requirement object, equivalent to "no auth required") instead of the expected `security: [{"send-access-bearer": []}]`. 

According to Claude the cause is that `new OpenApiSecuritySchemeReference("send-access-bearer")` is constructed without the host document, so the reference doesn't serialize as a `$ref`. Fix: convert the filter to `IDocumentFilter` (which receives the document) and pass it to the reference constructor — same pattern as `AddSwaggerServerWithSecurity` in `SharedWeb/Utilities/ServiceCollectionExtensions.cs:867-870`.

adudek-bw: Thanks for the info on the command! I updated the code :)

coroiu: question: Is V1 out of scope? Same IDocumentFilter could potentially also override `[AllowAnonymous]` endpoints in `SendsController` to `security: []`, which could let the SDK drop hand-rolled V1 calls

src/Api/Tools/Controllers/SendsController.cs
itsadrago: @adudek-bw, in addition to `GetSendFileDownloadDataUsingAuth` (as pointed out above),  there is another action (`AccessUsingAuth`) that's still returning `IActionResult`. We should make sure we address returning the correct response model for all the actions in the controller

src/SharedWeb/Swagger/SendAccessBearerOperationFilter.cs
coroiu: @adudek-bw ?

coroiu: Talked with @adudek-bw on Slack where it was clarified that this is already addressed 👍

coroiu (approved):

harr1424 (approved):

PR #7535, opened 2026-04-23
[PM-34565] Save Cancellation Details for Scheduled Subscriptions

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34565

## 📔 Objective

Updates how deferred price increase subscription cancellations and reinstatements are handled. The changes ensure that when a subscription with a pending price increase is cancelled at period end, the associated schedule is released (not updated) with cancellationDetails being updated, a specific metadata flag is set, and reinstatement logic is simplified to clear this flag and reschedule appropriately. Tests have been updated to reflect these new behaviors.

## 📸 Screenshots

https://github.com/user-attachments/assets/712f2aea-d8cf-4fa0-98c3-9c4b7777a0f6

PR #7561, opened 2026-04-29
[PM-35909] Preserve existing discounts during price migration

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35909

## 📔 Objective

Fixes a silent regression introduced by PM-32645's deferred-price-migration scheduler. When a Premium or Families subscription with a pre-existing subscription-level discount (a coupon attached to `Subscription.Discounts`) went through the migration schedule, the discount was silently dropped at Phase 2 activation. Phase 1 preserved discounts correctly, so the bug only fired at renewal — invisible during the ~15-day window between schedule creation and Phase 2 going live, and invisible to our logs and Stripe's. Affected populations: Premium 2019, Families 2019, and Families 2025 customers migrated since ~Q1 2026 with grandfather/partner/NFR/retention coupons.

**The fix:**
- `PriceIncreaseScheduler.ResolvePhase2For{Premium,Families}Async` now reads `subscription.Discounts` and concatenates the migration's Milestone coupon onto that list, rather than replacing with only the Milestone. Premium always appends Milestone2; Families 2019 appends Milestone3; Families 2025 appends nothing (and collapses an empty list to `null` to preserve the historical Stripe-API contract). Existing-discount-first ordering keeps the customer's pre-existing discount as the primary.
- `UpcomingInvoiceHandler` adds `subscriptions.data.discounts` to the customer expand list so `subscription.Discounts` is actually populated for the production webhook path. Without this, the scheduler fix would be a silent no-op for the bug's primary trigger. Added a load-bearing comment citing PM-35909 to discourage future cleanup.
- `ReinstateSubscriptionCommand` adds `Expand = ["discounts"]` to its subscription fetch. This was a third Schedule() call site introduced by PM-34565 (#7535) shortly before this PR; without the expand, the bug regressed silently on the reinstate path (or worse, NRE'd).
- `ResolvePhase2Async` now refuses to construct Phase 2 when handed a subscription whose `Discounts` collection contains null entries — the empirically-confirmed Stripe.NET shape for an unexpanded `discounts` field. Surfaces expand-list misuse as a loud `LogError` with the dropped-discount count, instead of regressing to the pre-fix silent-drop behavior. Defends against future call sites that forget the expand.

**Test changes:**
- Strengthened existing scheduler tests with `Discounts.Count == 1` assertions for the no-existing-discount cases.
- Added 7 new tests covering Premium / Families 2019 / Families 2025 with one or more pre-existing discounts in both `Schedule_*` (integration) and `ResolvePhase2Async_*` (unit) flavors, plus a regression test for the unexpanded-subscription guard.
- Added `Run_FetchesSubscriptionWithDiscountsExpanded` to `ReinstateSubscriptionCommandTests` to pin the new expand option.

## Follow-up: `GetBitwardenSubscriptionQuery` discount-source refactor

Once `PriceIncreaseScheduler` correctly populates Phase 2 with `[existing-discounts..., Milestone]`, several assumptions in `GetBitwardenSubscriptionQuery` no longer hold:
- The pre-fix code only fetched the schedule when `cartLevelDiscount == null`, gating on the (now-broken) assumption that "current discount and scheduled discount are mutually exclusive."
- It surfaced the schedule's discount as a fallback on the seat line item via `?? scheduleDiscount`, only correct because Phase 2 used to contain only the cart-level Milestone.
- The tax-preview override-mode took a single `couponId`, which silently under-reports tax savings now that Phase 2 may contain multiple coupons that all apply at renewal.

**Refactor:**
- Adopt a single source-of-truth rule: when a schedule is attached, Phase 2's discounts are the source for subscription-level coupons (Phase 2 reflects the upcoming-renewal state, including preserved current discounts plus migration coupons). Otherwise `subscription.Discounts` is the source. Customer-level discounts (`subscription.Customer.Discount`) are always included.
- Project both sources to `Stripe.Coupon` and partition by `Coupon.AppliesTo.Products` via a new `PartitionCouponsByScope` helper (renamed from `GetStripeDiscounts`). Cart-level coupons land on `Cart.Discount`; product-level land on the matching `CartItem.Discount`.
- Remove the seat-fallback `?? scheduleDiscount` — cart-level Phase 2 coupons (e.g. the migration Milestone) now surface on `Cart.Discount` directly, where they belong.
- `EstimatePremiumTaxAsync` parameter changes from `string? couponId` to `List<string>? couponIds` so the override-mode tax preview includes every coupon that will apply at renewal.
- Update the schedule expand list from `phases.discounts.coupon` to `phases.discounts.coupon.applies_to` so the partition logic can read `Coupon.AppliesTo.Products`.
- `GetSchedulePhase2CouponsAsync` no longer swallows `StripeException` — it logs and rethrows. Pre-refactor, soft-failing only suppressed a display badge; post-refactor the same swallow would silently corrupt the tax estimate (Phase 2 coupons feed both the discount display *and* `options.Discounts` for the invoice preview), so failure must propagate.
- New `Coupon.AppliesTo(SubscriptionItem)` extension in `DiscountExtensions.cs` mirrors the existing `Discount.AppliesTo` for the partitioned coupon list.

**Behavior change worth flagging for review:** a customer in the migration window with no current cart-level discount used to see the Milestone displayed on the seat line; now they see it as `Cart.Discount`. Visually different, semantically correct.

**Follow-up tests:**
- Renamed `Run_WithSchedulePhase2Discount_IncludesDiscountOnSeats` → `_IncludesDiscountOnCart` and `Run_WithSchedulePhase2AmountOffDiscount_IncludesAmountOffDiscountOnSeats` → `_OnCart` to reflect the new placement.
- New `Run_WithSchedulePhase2ProductLevelCoupon_IncludesDiscountOnSeat` — verifies a Phase 2 coupon with `AppliesTo.Products` populated lands on `Seats.Discount`, exercising the previously-uncovered partition path.
- New `Run_UserOnLegacyPricing_WithCustomerLevelDiscountAndScheduleDiscount_IncludesAllCouponIdsInTaxPreview` — verifies both customer-level and Phase 2 coupon IDs flow into `options.Discounts` (asserts `Count == 2`), the multi-coupon tax-preview path that didn't previously have coverage.
- `Run_WithScheduleStripeException_ReturnsNoDiscount` → `_Rethrows` — schedule fetch failure now propagates instead of soft-failing.
- Test fixture cleanup: `CreateSubscriptionSchedule` parameter `couponId` defaults to `"phase2-coupon"` directly (was a runtime sentinel), and gains an `appliesToProductId` parameter for product-scoped coupons.

## Concerns / follow-ups (not in this PR)

- **`StripePaymentService.ApplySchedulePhase2DataAsync:789`** has the same `phase2.Discounts?.FirstOrDefault()` pattern that PM-35909 fixed in `PriceIncreaseScheduler`. Post-fix, Phase 2 may contain multiple discounts; `FirstOrDefault()` will return whichever Stripe lists first, which is no longer guaranteed to be the migration coupon. The schedule-fetch + status/phase/start-date guard preamble is duplicated between this method and the new `GetSchedulePhase2CouponsAsync`. A future PR should fix the bug there and extract the shared preamble into a helper.
- Multi-discount support on `Cart.Discount` / `CartItem.Discount` — currently single-valued, with a separate Jira ticket tracking the multi-discount surface. This PR keeps the singular contract; the source-of-truth rule lays the groundwork without changing the public shape.

## Out of scope (per ticket, file separately)
- Detection script for already-affected customers since PM-32645 went live.
- Remediation script to re-attach silently-dropped discounts on already-migrated subscriptions.

### DIFF
--- a/src/Core/Billing/Subscriptions/Queries/GetBitwardenSubscriptionQuery.cs
+++ b/src/Core/Billing/Subscriptions/Queries/GetBitwardenSubscriptionQuery.cs
@@ -107,11 +107,8 @@ public class GetBitwardenSubscriptionQuery(
         var additionalStorageItem = subscription.Items.FirstOrDefault(item =>
             plans.Any(plan => plan.Storage.StripePriceId == item.Price.Id));
 
-        var (cartLevelDiscount, productLevelDiscounts) = GetStripeDiscounts(subscription);
-
-        var (scheduleDiscount, scheduleCouponId) = cartLevelDiscount == null
-            ? await GetSchedulePhase2DiscountAsync(subscription)
-            : (null, (string?)null);
+        var coupons = await GetRelevantCouponsAsync(subscription);
+        var (cartLevelCoupon, productLevelCoupons) = PartitionCouponsByScope(coupons);
 
         var availablePlan = plans.First(plan => plan.Available);
         var onCurrentPricing = passwordManagerSeatsItem.Price.Id == availablePlan.Seat.StripePriceId;
@@ -127,7 +124,9 @@ public class GetBitwardenSubscriptionQuery(
         else
         {
             seatCost = availablePlan.Seat.Price;
-            estimatedTax = await EstimatePremiumTaxAsync(subscription, plans, availablePlan, scheduleCouponId);
+            estimatedTax = await EstimatePremiumTaxAsync(
+                subscription, plans, availablePlan,
+                [.. coupons.Select(c => c.Id)]);
         }
 
         var passwordManagerSeats = new CartItem
@@ -135,7 +134,7 @@ public class GetBitwardenSubscriptionQuery(
             TranslationKey = "premiumMembership",
             Quantity = passwordManagerSeatsItem.Quantity,
             Cost = seatCost,
-            Discount = productLevelDiscounts.FirstOrDefault(discount => discount.AppliesTo(passwordManagerSeatsItem)) ?? scheduleDiscount
+            Discount = productLevelCoupons.FirstOrDefault(coupon => coupon.AppliesTo(passwordManagerSeatsItem))
         };
 
         var additionalStorage = additionalStorageItem != null
@@ -144,7 +143,7 @@ public class GetBitwardenSubscriptionQuery(
                 TranslationKey = "additionalStorageGB",
                 Quantity = additionalStorageItem.Quantity,
                 Cost = GetCost(additionalStorageItem),
-                Discount = productLevelDiscounts.FirstOrDefault(discount => discount.AppliesTo(additionalStorageItem))
+                Discount = productLevelCoupons.FirstOrDefault(coupon => coupon.AppliesTo(additionalStorageItem))
             }
             : null;
 
@@ -156,18 +155,16 @@ public class GetBitwardenSubscriptionQuery(
                 AdditionalStorage = additionalStorage
             },
             Cadence = PlanCadenceType.Annually,
-            Discount = cartLevelDiscount,
+            Discount = cartLevelCoupon,
             EstimatedTax = estimatedTax
         };
     }
-
-    #region Utilities
 
     private async Task<decimal> EstimatePremiumTaxAsync(
         Subscription subscription,
         List<PremiumPlan>? plans = null,
         PremiumPlan? availablePlan = null,
-        string? couponId = null)
+        List<string>? couponIds = null)
     {
         try
         {
@@ -185,7 +182,7 @@ public class GetBitwardenSubscriptionQuery(
 
                 options.SubscriptionDetails = new InvoiceSubscriptionDetailsOptions
                 {
-                    Items = subscription.Items.Select(item =>
+                    Items = [.. subscription.Items.Select(item =>
                     {
                         var isSeatItem = plans.Any(plan => plan.Seat.StripePriceId == item.Price.Id);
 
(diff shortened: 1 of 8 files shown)


sbrown-livefront (dismissed):
✅ Sheesh, this was a tough regression. Great find and solution!

src/Core/Billing/Subscriptions/Queries/GetBitwardenSubscriptionQuery.cs
```diff
@@ -223,55 +220,72 @@ private static decimal GetCost(OneOf<SubscriptionItem, List<InvoiceTotalTax>> va
...
+        var coupons = new List<Coupon>();
 
         if (subscription.Customer.Discount.IsValid())
         {
-            discounts.Add(subscription.Customer.Discount);
+            coupons.Add(subscription.Customer.Discount.Coupon);
         }
 
-        discounts.AddRange(subscription.Discounts.Where(discount => discount.IsValid()));
+        if (!string.IsNullOrEmpty(subscription.ScheduleId))
```
sbrown-livefront, line 239: ❓ Do we need to FF this? Probably no other schedules in existence, but I wanted to be sure.

sbrown-livefront (approved):
✅

src/Core/Billing/Subscriptions/Queries/GetBitwardenSubscriptionQuery.cs
amorask-bitwarden: Good q but since we gated schedule creation with the FF - I'm cool with not weaving it into other areas if we don't have to. My understanding is that CS now knows arbitrary schedules are not supported by us unless we build out for them.

PR #7565, opened 2026-04-29
[PM-30852] Add support for TDE user key rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30852

## 📔 Objective

This adds support for TDE users without master passwords to be able to preform a userKey rotation.


src/Core/KeyManagement/UserKey/Models/Data/TdeRotateUserAccountKeysData.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Bit.Core.KeyManagement.UserKey.Models.Data;
+
+public class TdeRotateUserAccountKeysData
+{
+    public required BaseRotateUserAccountKeysData BaseData { get; init; }
+
+    public void ValidateForUser(User user)
+    {
+        var isTdeUser = user is { Key: null, MasterPassword: null, UsesKeyConnector: false };
```
quexten, line 12: A thought: We should probably consider adding a "mainDecryptionType" column on the database instead of dynamically determining it based on several fields every time.

(not here)

quexten (approved):

src/Core/KeyManagement/UserKey/Models/Data/TdeRotateUserAccountKeysData.cs
Thomas-Avery: Had similar thoughts when writing this.

PR #7567, opened 2026-04-30
Fix EF UserRepository batched UpdateUserData transactional behavior

### LINKED ISSUE #7566: UserRepository EF UpdateUserDataAsync does not run delegates in a transaction
### Steps To Reproduce

1. Deploy the server against an Entity Framework provider (PostgreSQL, MySQL, or SQLite) — i.e. any non-MSSQL self-hosted configuration.
2. As an authenticated user, invoke any code path that batches user-state mutations through `IUserRepository.UpdateUserDataAsync(IEnumerable<UpdateUserData>)` or `IUserRepository.SetV2AccountCryptographicStateAsync(...)`. Master-password set/change, Key Connector key set, and `RegisterFinish` master-password unlock data update all flow through here.
3. Construct a batch of two delegates where the second one will throw — e.g. arrange the second delegate's stored mutation to violate a constraint, or inject a faulting delegate after `IUserRepository.SetMasterPassword(...)` in a test harness.
4. Execute the batch and observe the database state after the exception propagates.

### Expected Result

All delegates in the batch commit atomically or none do. The XML doc on `IUserRepository.UpdateUserDataAsync` and `IUserRepository.SetV2AccountCryptographicStateAsync` explicitly states the operations execute "in a single transaction." On a failure mid-batch, every prior write should roll back, leaving the user's cryptographic state (master password, KDF parameters, master-key-wrapped user key, salt, key-connector key) consistent with the pre-call state.

### Actual Result

On the EF code path, each delegate auto-commits independently and the outer transaction wraps nothing. A mid-batch failure leaves earlier delegates' writes durably committed.

Root cause is a mismatch between where the transaction is opened and where the writes are issued:

### Screenshots or Videos

N/A — backend correctness issue. Reproducible via integration test against the EF provider.

### Additional Context

- [`src/Core/Repositories/IUserRepository.cs`](src/Core/Repositories/IUserRepository.cs) (delegate definition, contract)
- [`src/Infrastructure.Dapper/Repositories/UserRepository.cs`](src/Infrastructure.Dapper/Repositories/UserRepository.cs) lines 433–546 (correct, reference behavior)
- [`src/Infrastructure.EntityFramework/Repositories/UserRepository.cs`](src/Infrastructure.EntityFramework/Repositories/UserRepository.cs) lines 511–610 (broken transactional behavior)
- Callers exercising the broken path include [`src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs`](src/Core/Auth/UserFeatures/UserMasterPassword/MasterPasswordService.cs) and [`src/Core/Auth/UserFeatures/UserMasterPassword/TdeSetPasswordCommand.cs`](src/Core/Auth/UserFeatures/UserMasterPassword/TdeSetPasswordCommand.cs).

### Build Version

main @ 27ae3d545 (also reproduces on every prior commit since the UpdateUserData delegate was introduced)

### Environment

Self-Hosted

### Environment Details

- Database provider: PostgreSQL, MySQL, or SQLite (any EF-backed deployment). MSSQL/Dapper deployments are not affected.
- Operating system: provider-agnostic
- Environment: Docker / Kubernetes / bare metal — not deployment-shape-dependent
- Hardware: N/A

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

carladams1299-lab:
## 🎟️ Tracking

Resolves [#7566](https://github.com/bitwarden/server/issues/7566).

## 📔 Objective

Fixes a transactional-correctness bug on the EF code path of `UserRepository`. `UpdateUserDataAsync` and `SetV2AccountCryptographicStateAsync` opened a transaction on a `DatabaseContext` resolved from one `IServiceScope`, but each `UpdateUserData` delegate (`SetMasterPassword`, `SetKeyConnectorUserKey`, `UpdateMasterPasswordUnlockData`) opened its own scope and resolved its own `DatabaseContext`. Each `SaveChangesAsync` therefore auto-committed on a different connection from the one that owned the outer transaction, and the absence of a `try`/`catch` around the action loop meant a mid-batch throw left earlier delegates' writes durably committed. For master-password / KDF / Key Connector mutations this could leave a user with mismatched cryptographic state (e.g. a new master-password hash without the matching master-key-wrapped user key), locking them out of their vault until manual database reconciliation. Self-hosted EF deployments (Postgres / MySQL / SQLite, plus EF-mode MSSQL) were exposed; cloud (MSSQL/Dapper) was not — see issue #7566 for the full mechanism.

The fix flows the active batch `DatabaseContext` into each delegate via an `AsyncLocal<DatabaseContext?>` on the EF `UserRepository`, so all writes in a batch share the connection that owns the open transaction. The `UpdateUserData` delegate signature is preserved (no public-API change). Both methods now wrap their action loops in `try`/`catch` with explicit `RollbackAsync()` on failure. When no batch is active, delegates fall back to a fresh scope, preserving the standalone-invocation path used by existing test helpers. The Dapper `UserRepository` was already correct and is unchanged.

### Test plan

- New integration test `UpdateUserDataAsync_WhenSubsequentActionThrows_RollsBackPriorWrites` batches `[SetMasterPassword, throwing-delegate]`, asserts the call throws, then re-reads the user and asserts the prior delegate's write was rolled back. Runs against every configured `[DatabaseData]` provider — proves the fix on EF backends and regression-protects Dapper.
- All 66 existing unit tests under `Core.Test` for the consumer commands (`MasterPasswordServiceTests`, `TdeSetPasswordCommandTests`, `FinishSsoJitProvisionMasterPasswordCommandTests`, `SetKeyConnectorKeyCommandTests`) continue to pass — confirming the `IUserRepository` contract relied on by production callers is unchanged.

### Out of scope

- Removing `Microsoft.Data.SqlClient.SqlConnection?` / `SqlTransaction?` from the public `UpdateUserData` delegate in `IUserRepository`. That's an API-shape change discussed as a follow-up in the original issue.
- The same transactional-boundary fix for the `UpdateEncryptedDataForKeyRotation` delegate path in `UpdateUserKeyAndEncryptedDataAsync` / `UpdateUserKeyAndEncryptedDataV2Async` — separate delegate type with its own callers.

## 📸 Screenshots

N/A — backend correctness fix.

### DIFF
--- a/src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
+++ b/src/Infrastructure.EntityFramework/Repositories/UserRepository.cs
@@ -14,6 +14,13 @@ namespace Bit.Infrastructure.EntityFramework.Repositories;
 
 public class UserRepository : Repository<Core.Entities.User, User, Guid>, IUserRepository
 {
+    // Carries the DbContext that owns the active UpdateUserData batch transaction so
+    // that delegates returned by SetMasterPassword/SetKeyConnectorUserKey/etc. can
+    // write through the same connection that owns the transaction. Without this, each
+    // delegate would open its own scope and SaveChangesAsync would auto-commit
+    // outside the outer transaction (see issue #7566).
+    private readonly AsyncLocal<DatabaseContext?> _activeBatchDbContext = new();
+
     public UserRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper)
         : base(serviceScopeFactory, mapper, (DatabaseContext context) => context.Users)
     { }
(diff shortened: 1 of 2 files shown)


carladams1299-lab:
Hi, @justindbaur ,
This is my first contribution.
Could you please review my PR?
Thank you.

PR #7568, opened 2026-04-30
[PM-32743] Add ability to create folders during import to orgs

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32743

## 📔 Objective

This PR adds the ability to create folders and folder relationships when importing into an organization, which is required to enable the ability to convert collections into folders when importing into My Items as described in the ticket


src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
```diff
@@ -222,9 +199,43 @@ public async Task ImportIntoOrganizationalVaultAsync(
...
+        {
+            var cipher = ciphers.ElementAtOrDefault(relationship.Key);
+            var folder = folders.ElementAtOrDefault(relationship.Value);
+
+            if (cipher == null || folder == null)
+            {
+                continue;
+            }
+
+            cipher.Folders = $"{{\"{importingUserId.ToString().ToUpperInvariant()}\":" +
```
mcamirault, line 240: This used to be set using the userId on the cipher, but for org ciphers this field is not defined. This should not change what ID is used for personal imports because the same userId that is set on the ciphers is passed into the `ImportIntoIndividualVaultAsync` function as the importing user ID

nick-livefront (dismissed):

src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
```diff
@@ -222,9 +199,43 @@ public async Task ImportIntoOrganizationalVaultAsync(
...
 
         // Create it all
-        await _cipherRepository.CreateAsync(ciphers, newCollections, collectionCiphers, newCollectionUsers);
+        await _cipherRepository.CreateAsync(ciphers, newCollections, collectionCiphers, newCollectionUsers, newFolders);
 
         // push
         await _pushService.PushSyncVaultAsync(importingUserId);
     }
+
+    private async Task<List<Folder>> processFolders(Guid importingUserId, List<Folder> folders, IEnumerable<KeyValuePair<int, int>> folderRelationships, List<CipherDetails> ciphers)
```
harr1424:
```suggestion
    private async Task<List<Folder>> ProcessFolders(Guid importingUserId, List<Folder> folders, IEnumerable<KeyValuePair<int, int>> folderRelationships, List<CipherDetails> ciphers)
```

src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
```diff
@@ -203,6 +204,12 @@ public async Task CreateAsync(IEnumerable<Core.Vault.Entities.Cipher> ciphers,
                 await dbContext.BulkCopyAsync(base.DefaultBulkCopyOptions, collectionUserEntities);
             }
 
+            if(folders.Any())
```
harr1424:
```suggestion
            if (folders.Any())
```

src/Api/Tools/Controllers/ImportCiphersController.cs
harr1424: ❓ Should the new fields `Folders` and/or `FolderRelationships` be validated here as well?

src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
harr1424: It's been some time since I looked at this code, but I remember a few tedious bugs involving `organizationDataOwnershipPolicyRequirement` (see line 53 of this file). I just want to make sure there is no way for this import flow to result in items being added to a user's personal vault? 

If it's possible, we need to respect `organizationDataOwnershipPolicyRequirement` status and disallow any ciphers from ending up in a user's personal vault.

test/Core.Test/Tools/ImportFeatures/ImportCiphersAsyncCommandTests.cs
harr1424, line 1: ❓ Should the new behavior introduced by this PR be tested here? I see the tests were updated to pass based upon the changes, but Folders and FolderRelationships are always empty.

harr1424 (changes requested):

src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
mcamirault: Done

src/Infrastructure.EntityFramework/Vault/Repositories/CipherRepository.cs
mcamirault: Done

src/Api/Tools/Controllers/ImportCiphersController.cs
mcamirault: I added them as new settings-controlled checks, not sure if there's anything else that has to be done when adding new settings like this

test/Core.Test/Tools/ImportFeatures/ImportCiphersAsyncCommandTests.cs
mcamirault: Since there's one test that seems to cover the success case for all of the parameters I updated it to test creating a Folder as well (and ignoring folders that already exist)

src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs
mcamirault: I believe that what determines whether a cipher is in a personal vault is the `UserId` column, and I believe that for personal imports this gets set on [this line of the CipherRequestModel](https://github.com/bitwarden/server/blob/main/src/Api/Vault/Models/Request/CipherRequestModel.cs#L76). The org import endpoint goes through [this function instead](https://github.com/bitwarden/server/blob/main/src/Api/Vault/Models/Request/CipherRequestModel.cs#L196) so I think we're clear. Let me know if my understanding is off though

harr1424: I agree with you, it appears that for org imports the `userId` will be set to null and the `organizationId` used as the `importingUserId`.

src/Api/Tools/Controllers/ImportCiphersController.cs
harr1424: I am not sure either, I like the overall pattern, the individual import uses hard-coded values, which likely isn't preferred. 

I see the added props in the `GlobalSettings.cs` file and values assigned in `appsettings.json` so I think this is complete, it may be worth following up with Product to check if we need to update any documentation for self-hosters who want to configure their desired limits.

harr1424 (dismissed):

nick-livefront (dismissed):

test/Api.Test/Tools/Controllers/ImportCiphersControllerTests.cs
```diff
@@ -247,11 +253,6 @@ public async Task PostImportOrganization_WithAccessImportExport_Succeeds(
         sutProvider.GetDependency<GlobalSettings>()
             .ImportCiphersLimitation = _organizationCiphersLimitations;
 
-        var importCiphersLimitation = new GlobalSettings.ImportCiphersLimitationSettings();
```
mcamirault, line 250: This wasn't used as far as I can tell, removed as cleanup

mcamirault:
@claude Review this PR

harr1424 (approved):

harr1424 (dismissed):

nick-livefront (dismissed):

nick-livefront (approved):

harr1424 (approved):

PR #7576, opened 2026-05-02
[PM-36419] [BEEEP] Add collection management settings to seeder

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36419
## 📔 Objective

Allow control of [collection management settings](https://bitwarden.com/help/collection-management/) via the seeder. These alter organization/collection behavior significantly and go hand-in-hand with different collection/group assignments to set up a customer's preferred access controls.

I wanted to avoid appending lots of nullable parameters to various methods, so I've created an `OrganizationOverrides` bucket to contain them. This should allow for easier modification in in the future as well.


util/Seeder/Factories/PlanFeatures.cs
```diff
@@ -13,13 +14,6 @@ public static class PlanFeatures
 {
     internal static void Apply(Organization org, PlanType planType)
     {
-        // Org-level admin settings — not plan-gated, safe defaults for seeding
-        org.UseAutomaticUserConfirmation = true;
-        org.AllowAdminAccessToAllCollectionItems = true;
-        org.LimitCollectionCreation = true;
-        org.LimitCollectionDeletion = true;
-        org.LimitItemDeletion = true;
```
eliykat, line 21: Opinionated change I wanted to call out: it's correct that these are not plan-gated, however they all default to `false` in production, so omitting them here is a more sensible default. Particularly as we now have overrides you can apply to turn them on if you want. (I've included `UseAutomaticUserConfirmation` in the overrides as well for completeness.)

I _assume_ nobody is relying on these default values yet but we can discuss if this is expected to cause impact anywhere.

theMickster:
Working through a [multi-agent-review](https://github.com/bitwarden/ai-plugins/pull/96) with Claude, brought up an interesting, minor change to wording in the XML comments that worth considering. 

<details>
<summary>♻️ Doc comments say "keep the plan default" but these settings are not plan-gated</summary>

`util/Seeder/Options/OrganizationOverrides.cs:4`
**Caught by:** Code quality agent

Three doc comments in this diff describe a null override as "keeping the plan default":

1. `OrganizationOverrides.cs` class summary: `"Null properties mean 'keep the plan default'."`
2. `PlanFeatures.ApplyOrganizationOverrides` XML summary: `"null means 'keep the plan default'"`
3. `OrganizationVaultOptions.Overrides` summary: `"Null means 'use plan defaults for everything'."`

None of the five settings (`UseAutomaticUserConfirmation`, `AllowAdminAccessToAllCollectionItems`, `LimitItemDeletion`, `LimitCollectionCreation`, `LimitCollectionDeletion`) are set anywhere in the `PlanType` switch. The deleted comment in `PlanFeatures.Apply` was accurate: `// Org-level admin settings — not plan-gated, safe defaults for seeding`. The real fallback when no override is supplied is whatever `OrganizationSeeder.Create` initializes the entity to (the C# `bool` default of `false`), not a plan-derived value.

What "plan default" implies: That the value is determined by which PlanType the org uses — Enterprise gets one default, Teams gets another, Free gets another, etc.

What actually happens: None of these five settings appear anywhere in the PlanType switch in PlanFeatures.Apply. There is no per-plan logic for them — never was. They're org-level admin settings orthogonal to billing plan. If no override is supplied, the value falls back to whatever the Organization entity initializes to: the C# bool default of false.

It's a ♻️ (not⚠️) because the code behaves correctly. No one is getting incorrect data; a future maintainer just gets a misleading mental model.

Why it matters: A developer reading "plan default" in the future might:
  - Look in the PlanType switch for where these defaults come from and find nothing
  - Assume Enterprise orgs get different defaults than Free orgs and write a test around that assumption
  - Add per-plan initialization thinking they're filling in a gap

The accurate wording would be something like "keep the current value" or "leave unchanged from OrganizationSeeder.Create" — meaning null is purely a no-op passthrough, not a delegation to plan logic.

**Suggested fix** — replace "plan default" with "entity default" or a reference to `OrganizationSeeder.Create` in all three locations. Example:

```csharp
/// <summary>
/// Optional overrides applied on top of the organization's initial values.
/// Null properties mean "keep the value set by <see cref="OrganizationSeeder.Create"/>".
/// </summary>
```

</details>

util/Seeder/Factories/PlanFeatures.cs
theMickster: Agreed. 

Let's take this one step further to cement that knowledge in `util/Seeder/CLAUDE.md`

- **Collection management settings are not plan-gated.** `UseAutomaticUserConfirmation`, `AllowAdminAccessToAllCollectionItems`, `LimitCollectionCreation`, `LimitCollectionDeletion`, and `LimitItemDeletion` apply identically across all plan types. They are org-level admin settings, not billing-plan features.
- **These five settings alter access control behavior.** When seeding scenarios that test member vs. admin permissions, collection creation/deletion policies, or item-level access, set them explicitly in the preset rather than relying on defaults.
- **Configurable in presets and CLI.** Use the JSON preset `organization` block (e.g. `"limitCollectionCreation": true`) or the CLI flags: `--limit-collection-creation`, `--limit-collection-deletion`, `--limit-item-deletion`, `--allow-admin-collection-access`, `--auto-confirm-users`.

theMickster (changes requested):
Two minor changes to documentation, but otherwise support this 100%

util/Seeder/Factories/PlanFeatures.cs
eliykat: This is done, except I omitted the reference to `UseAutomaticUserConfirmation`. That is a non-plan-gated organization feature but is not a collection management setting so it doesn't belong in this section. However, for what it's worth I have since added xmldoc to the `Organization` entity which documents this separately.

theMickster (approved):

PR #7579, opened 2026-05-04
[PM-35201] Enhance AdminRecoverAccountValidator to include Accepted status

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35201

## 📔 Objective

Extends the states for password recovery to support the Accepted status


JimmyVo16 (changes requested):
Looks good. Just need to fix the tests. It looks like there are some missing references.

test/Core.Test/AdminConsole/OrganizationFeatures/AccountRecovery/v2/AdminRecoverAccountValidatorTests.cs
```diff
@@ -7,7 +7,6 @@
 using Bit.Core.Enums;
 using Bit.Core.Repositories;
 using Bit.Core.Services;
-using Bit.Core.Test.AdminConsole.AutoFixture;
```
JimmyVo16, line 10: It looks like this is the issue causing the failing test. I think it was removed by accident in a previous commit. I added it back and confirmed locally that the tests are passing.

JimmyVo16 (approved):

PR #7586, opened 2026-05-05
[PM-35624] Fix EF GetCountByOnlyOwnerAsync

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35624

## 📔 Objective
1. Fix the EF query to mirror the stored procedure behavior.
2. Add unit tests to cover edge cases.

Added test cases
  1. A user with no org memberships at all.                                                                                                         
  2. A user who is the only confirmed owner of one org.
  3. A user who shares ownership with another confirmed owner in the same org.                                                                     
  4. A user who is the sole owner of one org but shares ownership in another org.
  5. A user who is an Owner but has only been invited, not confirmed.                                                                       
  6. A user who is the sole confirmed owner of two separate orgs.

## 📸 Screenshots

I ran a sanity check by executing the tests against the old code to confirm they fail, which validates that this change is the fix.

image


eliykat (approved):
No notes!

PR #7588, opened 2026-05-06
[PM-34389] Add refresh endpoint for organization invite links

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34389

## 📔 Objective

Adds `POST /organizations/{orgId}/invite-link/refresh` that replaces an existing invite link with a new one (new ID, code, and encryption key) while preserving the allowed domains.

[Clients PR](https://github.com/bitwarden/clients/pull/20522).

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/InviteLinks/RefreshOrganizationInviteLinkCommand.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/InviteLinks/RefreshOrganizationInviteLinkCommand.cs
@@ -1,0 +1,52 @@
+﻿using Bit.Core.AdminConsole.Entities;
+using Bit.Core.AdminConsole.OrganizationFeatures.InviteLinks.Interfaces;
+using Bit.Core.AdminConsole.Repositories;
+using Bit.Core.AdminConsole.Utilities.v2.Results;
+using Bit.Core.Services;
+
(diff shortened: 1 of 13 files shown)


JimmyVo16:
Reviewing this will be my top priority today.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/RefreshOrganizationInviteLinkCommand.cs
```diff
@@ -0,0 +1,51 @@
...
+        newLink.SetNewId();
+
+        await organizationInviteLinkRepository.RefreshAsync(existing, newLink);
+
+        return newLink;
+    }
+
+    private async Task<bool> OrganizationHasInviteLinksAbilityAsync(Guid organizationId)
+    {
+        var ability = await applicationCacheService.GetOrganizationAbilityAsync(organizationId);
```
JimmyVo16, line 48: Just a comment: I think it's safe to use the ability cache since we should be turning on the new ability cache pattern by the time this change goes out. However, in the event that we fall back to the old cache, just know that there will be a 10-minute lag. This probably isn't a big issue, but it can be frustrating during demos.

JimmyVo16 (approved):

PR #7589, opened 2026-05-06
[PM-36613] Void open invoices for unpaid subscriptions

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36613

## 📔 Objective

Restores the void-open-invoices behavior on canceled Stripe subscriptions that was lost in [#6918](https://github.com/bitwarden/server/pull/6918) (PM-31140). Since `v2026.2.1` shipped on 2026-03-03, invoices on subscriptions whose payment fails and lapse to `unpaid` have been stuck in `open` status indefinitely — Stripe does not auto-void them when the subscription cancels.

### Approach: metadata-gated void at the deletion event

The unpaid lifecycle is a two-step process: when Stripe transitions a subscription to `unpaid`, `SubscriptionUpdatedHandler` calls `SetSubscriptionToCancelAsync` which sets `cancel_at = now + 7d` on the Stripe subscription. Stripe then runs its own clock and emits `customer.subscription.deleted` when the cancellation actually occurs.

The void must run at the *deletion* event, not the *unpaid transition*. Voiding 7 days early would erase invoices on subscriptions that ultimately pay during the grace window and recover.

But `customer.subscription.deleted` also fires for voluntary cancels, off-platform negotiated cancels, and provider migrations — those must leave open invoices intact so ops can reconcile manually. So we need a signal we control that distinguishes "this cancellation came from our unpaid lifecycle" from everything else.

That signal is a metadata stamp on the Stripe subscription:

1. **`SubscriptionUpdatedHandler.SetSubscriptionToCancelAsync`** — when scheduling the 7-day cancellation, stamps `metadata.cancellation_origin = "unpaid_subscription"` on the same `UpdateSubscriptionAsync` call that sets `cancel_at`.
2. **`SubscriptionUpdatedHandler.RemovePendingCancellationAsync`** — when the subscription recovers (`Unpaid → Active`), unsets the metadata by setting the key to empty string (Stripe's documented unset convention, which preserves the other metadata keys).
3. **`SubscriptionDeletedHandler`** — on `customer.subscription.deleted`, checks `metadata.cancellation_origin == "unpaid_subscription"` and, if present, runs the void loop **before** the subscriber-disable branches. Running void first means a downstream disable failure won't strand open invoices — on Stripe re-delivery, the void becomes an idempotent no-op while disable retries.

The new constants live in `StripeConstants.MetadataKeys.CancellationOrigin` and a nested `StripeConstants.CancellationOrigins.UnpaidSubscription`, following the existing convention used by `BillingReasons`, `InvoiceStatus`, etc.

The orphaned `SubscriptionCancellationJob` (dead code since PM-31140 — nothing scheduled it anymore), its test class, and its DI registration are deleted in the same PR.

### Cleanup is best-effort

- A failure inside `ListInvoicesAsync` logs at `Error` and returns. The disable branches still run afterward (they come after the void).
- A per-invoice `VoidInvoiceAsync` failure is logged and the loop continues. `StripeException` (the expected webhook re-delivery race against an already-voided invoice) logs at `Warning`; any other exception (transport-level failures like `HttpRequestException` or `TaskCanceledException`) logs at `Error`.

Pagination is handled by `IStripeAdapter.ListInvoicesAsync` with `SelectAll = true` (auto-paginating via Stripe SDK's `ListAutoPagingAsync`) — no manual cursor management on our side.

**Out of scope:** Backfill of invoices that went `open` between v2026.2.1 deploy and this fix's deploy. Those will not be retroactively voided. Ops decided against a polling background job for backfill; finance/ops can resolve historical leaked invoices manually if/when surfaced.

### Admin Portal symmetry: re-enable and disable hooks

The metadata stamp introduced above gives CS a manual recovery window in the Admin Portal. Both directions on the Edit pages for Organization, Provider, and User are wired through `ISubscriberService`:

1. **Re-enable (Enabled `false → true`, Premium `false → true`)** — `ResumeFromUnpaidCancellationAsync(ISubscriber)`. If the subscription is `unpaid` and carries `cancellation_origin = unpaid_subscription`, clears the pending cancellation (`CancelAtPeriodEnd = false`) and unsets the marker. Mirrors the existing webhook recovery path proven by QA scenarios 3, 8, and 12.
2. **Disable (Enabled `true → false`, Premium `true → false`)** — `ScheduleUnpaidCancellationAsync(ISubscriber)`. If the subscription is `unpaid`, has no `cancel_at`, and lacks the marker, sets `cancel_at = now + 7d` and stamps `cancellation_origin = unpaid_subscription`. Recovery for the broken half-state when the unpaid-transition webhook never fired the schedule.

Both methods are server-side gated and idempotent — safe no-ops on any other subscription state. Both controller hooks use the existing try/catch + `TempData["Warning"]` pattern: DB persistence succeeds even if the Stripe sync fails, with failures logged at `Error` for ops.

### Test coverage

`SubscriptionDeletedHandlerTests` adds:
- Happy path: canceled subscription WITH `cancellation_origin = unpaid_subscription` metadata → all open invoices voided.
- Regression-critical negative test: canceled subscription WITHOUT the metadata → no `ListInvoicesAsync`, no `VoidInvoiceAsync`. Locks in the narrow-scoping decision.
- Per-invoice `StripeException` (webhook re-delivery against already-voided invoice) → loop continues.
- Per-invoice non-Stripe transport exception (`HttpRequestException`) → loop continues.
- `ListInvoicesAsync` failure → does not block subscriber-disable.

`SubscriptionUpdatedHandlerTests` adds:
- Unpaid transition stamps `metadata.cancellation_origin = unpaid_subscription` on the Stripe update.
- Active-from-unpaid transition unsets `metadata.cancellation_origin` (value = empty string).

`SubscriberServiceTests` adds 12 tests for the Admin Portal hooks (6 per direction):
- `ResumeFromUnpaidCancellationAsync`: no `GatewaySubscriptionId`, null subscription, status not `unpaid`, metadata missing, metadata wrong value, happy path (clears cancellation + metadata).
- `ScheduleUnpaidCancellationAsync`: no `GatewaySubscriptionId`, null subscription, status not `unpaid`, already has `cancel_at`, already has matching metadata, happy path (sets `cancel_at = now + 7d` + stamps metadata).


kdenney (dismissed):
This will conflict with my changes in #7480, but it should be simple to resolve and the logic looks good to me.

kdenney (dismissed):

connerbw (dismissed):

connerbw (approved):

eliykat (approved):

PR #7592, opened 2026-05-06
[PM-35300] emails do not match figma

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35300

## 📔 Objective

Fixes up emails to match figma

## 📸 Screenshots
**Browser**
image
**iPhone 12 Pro**
image
**iPad Air**
image
**iPhone SE**
image


src/Core/MailTemplates/Mjml/emails/AdminConsole/AccountRecovery/admin-reset-password.mjml
```diff
@@ -0,0 +1,58 @@
...
+              {{/if}}
+            </ol>
+          </mj-text>
+          <mj-button
+            padding="0px 0px 24px 0px"
+            inner-padding="12px 24px"
+            border-radius="20px"
+            font-weight="600"
+            align="left"
+            href="{{WebVaultUrl}}"
```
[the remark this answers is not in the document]
eliykat: Please address any applicable Claude feedback before requesting team reviews. (If the feedback is wrong then that's OK, but it looks like it needs actioning here.)

src/Core/MailTemplates/Handlebars/AdminResetPassword.html.hbs
eliykat, line 1: Have the `.hbs` files been generated from the MJML per the [README](https://github.com/bitwarden/server/blob/3ae04026b3f8513861d91bbb85865d74a211c285/src/Core/MailTemplates/README.md)? At a glance it looks like these may have been edited separately.

src/Core/MailTemplates/Mjml/emails/AdminConsole/AccountRecovery/admin-reset-password.mjml
eliykat, line 1: Tagging our MJML expert @JimmyVo16 to take a look at this please - looks roughly correct, but should we be using so much custom padding, or do we have pre-built components?

eliykat (changes requested):

src/Core/MailTemplates/Handlebars/AdminResetPassword.html.hbs
JimmyVo16: Looks like the `html.hbs` and `text.hbs` files were merged years ago, and now we’re replacing them with the generated MJML output. This should be fine as long as we’re not using any feature flags (which it doesn’t look like we are), since the old version no longer exists.

src/Core/MailTemplates/Mjml/emails/AdminConsole/AccountRecovery/admin-reset-password.mjml
JimmyVo16: The last guidance we got from UIF was to focus on getting things done and not worry too much about DRY principles. Some factors that led to this decision:

1. UIF didn’t have the bandwidth to come up with established patterns for MJML. I just checked the code, and none of the shared components have been touched in 3+ months, so I don’t think they’ve introduced anything new since then.
2. It’s hard to use shared components because different templates have different text lengths, which causes the output to look different even. This can throw off spacing and affect the intended design.

With that said, we should still use shared components when possible. It looks like we’re already doing that here with `mj-bw-simple-hero` and `footer.mjml`. 


If you find that a shared component doesn’t work for your use case, be mindful about changing it since it will affect other templates, which we would need to retest to make sure nothing breaks. This is why we have `src/Core/MailTemplates/Mjml/emails/AdminConsole/components`, which contains shared components specifically for the AC team. At the time, we didn’t want to interfere with the auth templates.

Another consideration is that UIF will eventually go through and establish proper patterns. It would be a waste if we spent a lot of effort on this now just to have UIF undo it later.

JimmyVo16 (changes requested):
Can you make sure to show the mobile version as well so we can confirm everything looks good? I only saw one screenshot in the ticket and PR.

JaredScar:
> Can you make sure to show the mobile version as well so we can confirm everything looks good? I only saw one screenshot in the ticket and PR.

More images added! @JimmyVo16

JimmyVo16 (approved):

dereknance (approved):
✅ Platform

eliykat (approved):
Thanks @JaredScar and thanks for the second set of eyes @JimmyVo16

PR #7601, opened 2026-05-08
[PM-36678] Fix custom users modifying admins for restore and revoke.

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36678

### DIFF
--- a/src/Core/AdminConsole/Enums/OrganizationUserActionType.cs
+++ b/src/Core/AdminConsole/Enums/OrganizationUserActionType.cs
@@ -1,0 +1,8 @@
+﻿namespace Bit.Core.AdminConsole.Enums;
+
+public enum OrganizationUserActionType
+{
+    Remove,
+    Revoke,
+    Restore,
+}
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeUser/v2/RevokeOrganizationUsersValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeUser/v2/RevokeOrganizationUsersValidator.cs
@@ -1,5 +1,6 @@
 ﻿using Bit.Core.AdminConsole.Models.Data;
 using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Interfaces;
+using Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Validators;
 using Bit.Core.AdminConsole.Utilities.v2.Validation;
 using Bit.Core.Entities;
 using Bit.Core.Enums;
@@ -7,7 +8,9 @@ using static Bit.Core.AdminConsole.Utilities.v2.Validation.ValidationResultHelpe
 
 namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.RevokeUser.v2;
 
-public class RevokeOrganizationUsersValidator(IHasConfirmedOwnersExceptQuery hasConfirmedOwnersExceptQuery)
+public class RevokeOrganizationUsersValidator(
+    IHasConfirmedOwnersExceptQuery hasConfirmedOwnersExceptQuery,
+    ICustomUserActingOnAdminValidator customUserActingOnAdminValidator)
     : IRevokeOrganizationUserValidator
 {
     public async Task<ICollection<ValidationResult<OrganizationUser>>> ValidateAsync(
@@ -16,6 +19,8 @@ public class RevokeOrganizationUsersValidator(IHasConfirmedOwnersExceptQuery has
         var hasRemainingOwner = await hasConfirmedOwnersExceptQuery.HasConfirmedOwnersExceptAsync(request.OrganizationId,
             request.OrganizationUsersToRevoke.Select(x => x.Id) // users excluded because they are going to be revoked
             );
+
+        var customUserCannotRevokeAdmin = await CustomUserCannotRevokeAdminAsync(request.OrganizationUsersToRevoke);
 
         return request.OrganizationUsersToRevoke.Select(x =>
         {
@@ -32,9 +37,20 @@ public class RevokeOrganizationUsersValidator(IHasConfirmedOwnersExceptQuery has
                 { Type: OrganizationUserType.Owner } when request.PerformedBy is not SystemUser
                                                         && !request.PerformedBy.IsOrganizationOwnerOrProvider =>
                     Invalid(x, new OnlyOwnersCanRevokeOwners()),
+                { Type: OrganizationUserType.Admin } when customUserCannotRevokeAdmin =>
+                    Invalid(x, new CustomUserCannotRevokeAdmin()),
 
                 _ => Valid(x)
             };
         }).ToList();
+    }
+
+    // Probes with any admin in the batch. The rule's answer is uniform across every admin
+    // in the same organization, so one cached lookup covers the whole batch.
+    private async Task<bool> CustomUserCannotRevokeAdminAsync(IEnumerable<OrganizationUser> users)
+    {
+        var anyAdmin = users.FirstOrDefault(x => x.Type == OrganizationUserType.Admin);
+        return anyAdmin is not null
+            && await customUserActingOnAdminValidator.IsBlockedAsync(anyAdmin);
     }
 }
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
@@ -1,0 +1,54 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.Context;
+using Bit.Core.Entities;
+using Bit.Core.Enums;
+using Bit.Core.Exceptions;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Validators;
+
+public class CustomUserActingOnAdminValidator(ICurrentContext currentContext) : ICustomUserActingOnAdminValidator
+{
+    // Memoize OrganizationCustom answers for the lifetime of this request-scoped instance.
+    private readonly Dictionary<Guid, bool> _organizationCustomLookup = new();
+
+    public async Task EnforceAsync(OrganizationUser targetUser, OrganizationUserActionType actionType)
+    {
+        if (await IsBlockedAsync(targetUser))
+        {
+            throw new BadRequestException(actionType.ToCustomUserCannotModifyAdminMessage());
+        }
+    }
+
+    public async Task<bool> IsBlockedAsync(OrganizationUser targetUser)
+    {
+        if (targetUser.Type != OrganizationUserType.Admin)
+        {
+            return false;
+        }
+
+        return await IsActingUserCustomAsync(targetUser.OrganizationId);
+    }
+
+    private async Task<bool> IsActingUserCustomAsync(Guid organizationId)
+    {
+        if (_organizationCustomLookup.TryGetValue(organizationId, out var cached))
+        {
+            return cached;
+        }
+
+        var isCustom = await currentContext.OrganizationCustom(organizationId);
+        _organizationCustomLookup[organizationId] = isCustom;
+        return isCustom;
+    }
+}
+
+internal static class OrganizationUserActionExtensions
+{
+    public static string ToCustomUserCannotModifyAdminMessage(this OrganizationUserActionType actionType) => actionType switch
+    {
+        OrganizationUserActionType.Remove => "Custom users can not remove admins.",
+        OrganizationUserActionType.Revoke => "Custom users can not revoke admins.",
+        OrganizationUserActionType.Restore => "Custom users can not restore admins.",
+        _ => throw new ArgumentOutOfRangeException(nameof(actionType), actionType, null),
+    };
+}
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/ICustomUserActingOnAdminValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/ICustomUserActingOnAdminValidator.cs
@@ -1,0 +1,25 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.Entities;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Validators;
+
+public interface ICustomUserActingOnAdminValidator
+{
+    /// <summary>
+    /// Throws <see cref="Bit.Core.Exceptions.BadRequestException"/> when the current acting user
+    /// has the Custom role in the target user's organization and the target user is an Admin.
+    /// Custom users (even those granted the ManageUsers permission) are not permitted to modify
+    /// Admins under the role hierarchy. The exception message is scoped to the supplied actionType.
+    /// </summary>
+    /// <param name="targetUser">The organization user that the acting user is attempting to modify.</param>
+    /// <param name="actionType">The actionType being attempted; selects the error message.</param>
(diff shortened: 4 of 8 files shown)


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.Context;
+using Bit.Core.Entities;
+using Bit.Core.Enums;
+using Bit.Core.Exceptions;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Validators;
+
+public class CustomUserActingOnAdminValidator(ICurrentContext currentContext) : ICustomUserActingOnAdminValidator
```
JimmyVo16, line 9: Hey @sven-bitwarden , this is a new pattern I’m trying out. Happy to get any feedback.

We’re obviously doing the same check in a lot of places, so I was thinking we could abstract it into a single validator.

Benefits:

1. Keeps the logic DRY
2. Handles both single-user and bulk-user org cases without requiring multiple async calls.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/RevokeUser/v2/RevokeOrganizationUsersValidator.cs
```diff
@@ -32,9 +37,20 @@ _ when request.PerformedBy is not SystemUser
...
 
                 _ => Valid(x)
             };
         }).ToList();
     }
+
+    // Probes with any admin in the batch. The rule's answer is uniform across every admin
+    // in the same organization, so one cached lookup covers the whole batch.
+    private async Task<bool> CustomUserCannotRevokeAdminAsync(IEnumerable<OrganizationUser> users)
+    {
```
JimmyVo16, line 51: I still need to iterate on this a bit.

src/Core/AdminConsole/Enums/OrganizationUserActionType.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Bit.Core.AdminConsole.Enums;
+
+public enum OrganizationUserActionType
```
JimmyVo16, line 3: The goal of this was to allow the commands to just pass in the action, while keeping the error messages in the validator. This is mainly to reduce logic in the commands. Also, most of the messages are the same, so centralizing them is a good idea.

I thought about passing in the class type, but that would narrow the calling code to only those types. An enum seems like the easiest solution, but I’m open to ideas.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
eliykat, line 1: This is too narrow - the general pattern here is not "custom users cannot act on admins" (specific), it's "users with lower privileges cannot act on users with higher privileges" (more general).

Here is some example logic from account recovery which fully expresses the rule: https://github.com/bitwarden/server/blob/963c1609997c5d777da522e00d6e0cb498214308/src/Api/AdminConsole/Authorization/Organizations/Handlers/RecoverAccountAuthorizationHandler.cs#L76-L84

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    private async Task<bool> IsActingUserCustomAsync(Guid organizationId)
+    {
+        if (_organizationCustomLookup.TryGetValue(organizationId, out var cached))
+        {
+            return cached;
+        }
+
+        var isCustom = await currentContext.OrganizationCustom(organizationId);
+        _organizationCustomLookup[organizationId] = isCustom;
```
eliykat, line 40: Although this is async, `currentContext` caches its results locally so you don't need a second layer of caching here. (I also generally think that this is an antipattern so would be pretty hesitant to extend this kind of behavior)

eliykat (commented):
A couple of specific comments below, but more generally:

I do think that we need to develop a pattern of having smaller, more composable validators for shared logic. However, in this case, this is an authorization check that I think needs to live in the api layer. It is effectively an extension of RBAC: you must have this role, _but also_ that role must be equal or greater to the user you're acting on.

Therefore I think this should be handled by the authorization layer, potentially generalizing the [RecoveryAccountAuthorizationHandler](https://github.com/bitwarden/server/blob/963c1609997c5d777da522e00d6e0cb498214308/src/Api/AdminConsole/Authorization/Organizations/Handlers/RecoverAccountAuthorizationHandler.cs#L26-L25) which implements this logic.

You could imagine, for example, an attribute that handles this for you in a reusable way:

```c#
    [HttpPut("restore")]
    [Authorize<ManageUsersRequirement>]
    [Authorize<EqualOrGreaterRoleRequirement>]   // additional requirement
    public async Task<ListResponseModel<OrganizationUserBulkResponseModel>> BulkRestoreAsync(Guid orgId, [FromBody] OrganizationUserBulkRequestModel model)
```

Or - and I suspect this is the case - it may be too difficult or inefficient to do it from an attribute, in which case you can do it manually. This also means you have the orgUser objects to pass into the command later:

```c#
        var organizationUsers = await _organizationUserRepository.GetManyAsync(model.Ids);
        var authorized =
            await _authorizationService.AuthorizeAsync(User, organizationUsers, new EqualOrGreaterRoleRequirement());
// check authorized result, then pass orgUsers into command instead of just ids
```

I definitely agree that we need a reusable solution though.

Let me know what you think.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
JimmyVo16: Can you clarify which pattern you’re referring to here: `currentContext` or local caching?

If you mean the local caching approach, could you elaborate a bit more on why you see it as an anti-pattern? I don’t think I fully understand the tradeoffs yet. The main concern I can think of is stale data if the cache is long-lived, but in this case I made sure it only lives for the duration of the request by registering the validator as `AddScoped` in DI.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/ICustomUserActingOnAdminValidator.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.Entities;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Validators;
+
+public interface ICustomUserActingOnAdminValidator
```
JimmyVo16, line 6: Moving the discussion to this [comment](https://github.com/bitwarden/server/pull/7601#pullrequestreview-4268952332) here so it’ll be easier to comment on.

> I think this should be handled by the authorization layer

I’m not against this idea since it’s role-related. One concern is that the *AuthorizationHandler works great for single use cases, but how do we handle bulk actions? We’re running into the same problem as the ticket that checks whether orgUser belongs to the org during bulk actions.

Another concern is that AuthorizationHandler normally just returns pass or fail, while in our case, we return meaningful errors that are displayed to the user depending on the scenario. Changing this would be a breaking change.

edit: cc @eliykat

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
JimmyVo16: My original intention wasn’t to have a general role privileges check, but we can definitely pivot to that.

We have another similar rule: `Only owners can remove other owners` which could also be part of this validator. 

Some outstanding questions we need to address:

1. If we want to turn this into an AuthorizationHandler, then we need to address these [concerns](https://github.com/bitwarden/server/pull/7601/changes#r3227123658).
2. If we want to keep the current pattern, I can rename this validator to something more generic so we can add other rules to it later. In general, I prefer to keep bug fixes small so they’re easier to test and roll out, but I’m also happy to include the other rules here as well. Let me know what you think.

eliykat: ## Scope

> We have another similar rule: Only owners can remove other owners which could also be part of this validator.

Exactly - you can think of them as 3 separate business rules:
- only owners can remove owners
- only owners or admins can remove admins
- only owners, admins or customer users (with Manage Users) can remove anyone else

But in reality these will always go together - so I think they're actually expressing a single larger business rule:
- To remove someone, you must have the same role as them or a superior role

And so it makes sense to cover all 3 cases together in the same place. I think it's worth doing here so that the solution can be reused across all relevant flows (e.g. revoke, remove, delete, reset password... any others?)

## Authorization vs. core layer

You're right about authz handlers not handling bulk cases well - an `AuthorizationResult` can contain one or more reasons ([example](https://github.com/bitwarden/server/blob/main/src/Api/AdminConsole/Controllers/OrganizationUsersController.cs#L527)), but I think it would be really cumbersome to try to correlate those back to the users. I'm OK with keeping it as a single-purpose validator so that the result can flow back through our command/result pattern.

eliykat (commented):
I will leave some additional feedback tomorrow when I have some more time to review. See [response above](https://github.com/bitwarden/server/pull/7601#discussion_r3231730216).

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+public class CustomUserActingOnAdminValidator(ICurrentContext currentContext) : ICustomUserActingOnAdminValidator
+{
+    // Memoize OrganizationCustom answers for the lifetime of this request-scoped instance.
+    private readonly Dictionary<Guid, bool> _organizationCustomLookup = new();
+
+    public async Task EnforceAsync(OrganizationUser targetUser, OrganizationUserActionType actionType)
+    {
+        if (await IsBlockedAsync(targetUser))
+        {
+            throw new BadRequestException(actionType.ToCustomUserCannotModifyAdminMessage());
```
eliykat, line 18: The validator should follow our new pattern of not throwing errors for business logic failure. Have a single public method that returns a value; then callers can decide how to handle that value. (e.g. v1 command can throw, v2 command can return it out as a result)

JimmyVo16: There are two methods in this validator: one for the old approach (v1 method: EnforceAsync), which throws an exception, and another for the new approach (v2 method: IsBlockedAsync), which just returns a bool. 

I can change v2 to return the error message class if you'd like. One concern is that different actions have slightly different error messages. We could pass in an enum similar to how I'm handling it in v1.

I haven't put too much thought into v2 yet, so I'm open to ideas. I'd like to avoid doing multiple checks where the calling code has to inspect the result. Ideally, we would only need a single check.

In v1, we're effectively doing one check since the calling code handles the exception. With v2, we might end up needing a double check.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
JimmyVo16: This is just me thinking out loud here.

1. What if we make the error message more generic, like: "Unable to perform action because your role is below the target user"? Then we could reuse this validator anywhere and encapsulate all the role checks in one place.
2. We could create an authorization handler that simply calls the validator. The validator could also be used directly in commands for bulk actions.

Note: Ideally, it would be nice to have all the validation in one place so it's easier to check and verify, but I understand that authorization is slightly different from other forms of validation.

jrmccannon: >    - To remove someone, you must have the same role as them or a superior role

> And so it makes sense to cover all 3 cases together in the same place. I think it's worth doing here so that the solution can be reused across all relevant flows (e.g. revoke, remove, delete, reset password... any others?)

We need to make sure we are explicit about which roles can perform the revoke / remove action on which other roles as we laid out in the 3 rules. We should use that for the implementation and not rely on the ordinal comparison using enum values.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
jrmccannon: I think having `ICurrentContext` as a request level object that does a bunch of things off the bat every request is the anti-pattern he's referring to. I think it would be more preferred if there were smaller more focused contexts depending on the request that is occurring. `ICurrentContext` does too much at this point. But using it is better than taking something out of CurrentContext to store in another spot.

`ICurrentContext` is populated every request via middleware. By creating another scoped entity, you're caching the cache, which is also an antipattern.

JimmyVo16: It sounds like we want to remove my cache layer and just use ICurrentContext's cache. I'm okay with this. My main goal with adding the cache was to move away from this pattern for bulk operations, so single and bulk validation can be more aligned.

image


Since it's already cached, should we just call await `_currentContext.OrganizationCustom(organizationId);` for each user moving forward?

jrmccannon: You already would be at a certain point. ~It doesn't do a db lookup every time. Just once at the start and then is awaited everywhere else due to how the async/await works (because of lazy loading of provider data).~ See below

Honestly you could probably get away with using the [CurrentContextOrg](https://github.com/bitwarden/server/blob/bb5506e934b72ad8b1aafef76a7aa41c12f2bb97/src/Core/Context/CurrentContext.cs#L471) which is sync. That has access to Permissions and org user type.

Side Note:
```
    public Task<bool> OrganizationCustom(Guid orgId)
    {
        return Task.FromResult(Organizations?.Any(o => o.Id == orgId && o.Type == OrganizationUserType.Custom) ?? false);
    }
```

This shouldn't even be async.

eliykat:
Note: we are discussing this internally, I can have another look at this next week depending on what we decide on.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Validators/CustomUserActingOnAdminValidator.cs
sven-bitwarden: Finally catching up on this - my first/strongest reaction was the same as Thomas' here that this validator is too narrow. IMO the goal of the work is to centralize the business logic of how one user may interact with (manage) another.

eliykat (commented):
We agreed that we'd just do a targeted fix for now to address the bug, with a follow-up ticket to restructure the anti-privilege escalation logic more generally. Let me know if you want to discuss this PR any further at the moment.

When you do come back to the broader question, my main feedback is that it:
- should cover the general rule of "you must have equal or greater permissions than the target member"
- should return a value for the caller to handle, not throw
- should be reusable between all flows that need this
- can be in the validation layer

But happy to discuss any of that further.

JimmyVo16:
After the team discussion, we decided to create another PR to address just the bug using the existing pattern. There will be a follow-up ticket to address how we want to organize validation.

PR #7615, opened 2026-05-12
[PM-35401] Update exception handling in CreateAuthRequestAsync() and PostAdminRequest()

rr-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35401

## 📔 Objective

Adds a `BadRequestException` case to `CreateAuthRequestAsync()` and `PostAdminRequest()`.

### DIFF
--- a/src/Core/Auth/Services/Implementations/AuthRequestService.cs
+++ b/src/Core/Auth/Services/Implementations/AuthRequestService.cs
@@ -113,6 +113,11 @@ public class AuthRequestService : IAuthRequestService
             throw new BadRequestException("User or known device not found.");
         }
 
+        if (_currentContext.UserId.HasValue && user!.Id != _currentContext.UserId.Value)
+        {
+            throw new BadRequestException("User or known device not found.");
+        }
+
         // AdminApproval requests require correlating the user and their organization
         if (model.Type == AuthRequestType.AdminApproval)
         {
--- a/src/Core/Auth/Services/Implementations/AuthRequestService.cs
+++ b/src/Core/Auth/Services/Implementations/AuthRequestService.cs
@@ -113,6 +113,7 @@ public class AuthRequestService : IAuthRequestService
             throw new BadRequestException("User or known device not found.");
         }
 
+        // Ensure authenticated user id matches target user (allows anon scenarios still)
         if (_currentContext.UserId.HasValue && user!.Id != _currentContext.UserId.Value)
         {
             throw new BadRequestException("User or known device not found.");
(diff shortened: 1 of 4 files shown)


test/Core.Test/Auth/Services/AuthRequestServiceTests.cs
```diff
@@ -474,6 +474,46 @@ await sutProvider.GetDependency<IMailService>()
...
+
+        sutProvider.GetDependency<IGlobalSettings>()
+            .PasswordlessAuth.KnownDevicesOnly
+            .Returns(false);
+
+        var exception = await Assert.ThrowsAsync<BadRequestException>(
+            () => sutProvider.Sut.CreateAuthRequestAsync(createModel));
+
+        Assert.Equal("User or known device not found.", exception.Message);
+
```
JaredSnider-Bitwarden, line 515: We should add this: 
```suggestion
        // Pins the failure to the UserId-mismatch guard rather than the
        // identically-worded "user/device not found" guard that fires earlier.
        await sutProvider.GetDependency<IDeviceRepository>()
            .DidNotReceiveWithAnyArgs()
            .GetManyByUserIdAsync(default);
```

src/Core/Auth/Services/Implementations/AuthRequestService.cs
```diff
@@ -113,6 +113,11 @@ public async Task<AuthRequest> CreateAuthRequestAsync(AuthRequestCreateRequestMo
             throw new BadRequestException("User or known device not found.");
         }
 
+        if (_currentContext.UserId.HasValue && user!.Id != _currentContext.UserId.Value)
```
JaredSnider-Bitwarden, line 117: Please add a comment above this line:
`// Ensure authenticated user id matches target user (allows anon scenarios still)
`

JaredSnider-Bitwarden (changes requested):
Nice work. Two tiny tweaks and we are good to go 👇

test/Core.Test/Auth/Services/AuthRequestServiceTests.cs
rr-bw: I don't think the `GetManyByUserIdAsync()` mock is what pins the failure (it should not be called regardless of which specific exception gets thrown).

I've added two comments to the test that I think more accurately describe what's going on - https://github.com/bitwarden/server/pull/7615/changes/5099c6cff900136510c95279bc85e2dd0945d9de

src/Core/Auth/Services/Implementations/AuthRequestService.cs
rr-bw: https://github.com/bitwarden/server/pull/7615/changes/c8bcd1713529e69af17e29da62790405519f7745

test/Core.Test/Auth/Services/AuthRequestServiceTests.cs
JaredSnider-Bitwarden: Ah, ok, that's fair. You're right.

JaredSnider-Bitwarden (approved):

PR #7617, opened 2026-05-12
[PM-36843] Skip relay push registration for non mobile clients

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36843

## 📔 Objective

Skips relaying push registration for non-mobile clients since only mobile clients need to be registered in our cloud to get notifications. Browser extensions will continue to use SignalR on self-host.

Also I added additional details to the logs, for example the log that would be logged without the fix would have been:
```
Request to http://api.localhost/push/register is unsuccessful with status of BadRequest-Bad Request: The model state is invalid.
Validation Errors:
PushToken: The PushToken field is required.
```

### DIFF
--- a/test/Api.IntegrationTest/Api.IntegrationTest.csproj
+++ b/test/Api.IntegrationTest/Api.IntegrationTest.csproj
@@ -16,7 +16,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.Extensions.Diagnostics.Testing" Version="9.3.0" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.Testing" Version="[9.3.0]" />
   </ItemGroup>
 
   <ItemGroup>
--- a/test/Api.IntegrationTest/Api.IntegrationTest.csproj
+++ b/test/Api.IntegrationTest/Api.IntegrationTest.csproj
@@ -16,6 +16,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.Testing" Version="9.3.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


test/Api.IntegrationTest/Platform/RelayPushRegistrationServiceTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        _sut = new RelayPushRegistrationService(
+            httpClientFactory,
+            globalSettings,
+            logger
+        );
+    }
+
+    [Fact]
+    public async Task BrowserExtensionData_ShouldNotLogIssues()
```
justindbaur, line 70: This test failed before the fix.

test/Api.IntegrationTest/Api.IntegrationTest.csproj
```diff
@@ -16,6 +16,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.Testing" Version="9.3.0" />
```
djsmith85: 🎨 We recently pinned all deps, so this should be pinned too

test/Api.IntegrationTest/Platform/RelayPushRegistrationServiceTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    [Fact]
+    public async Task MobileData_ShouldNotLogIssues()
+    {
+        await _sut.CreateOrUpdateRegistrationAsync(
+            new PushRegistrationData("PushToken"),
+            deviceId: Guid.NewGuid().ToString(),
+            userId: Guid.NewGuid().ToString(),
+            identifier: Guid.NewGuid().ToString(),
+            DeviceType.ChromeExtension,
```
djsmith85: ⚠️  The test is a little misleading as the DeviceType is set to ChromeExtension and it's not clear what is prepped to cause mobile not to log?

djsmith85 (changes requested):

test/Api.IntegrationTest/Api.IntegrationTest.csproj
justindbaur: Done in 8386b26

test/Api.IntegrationTest/Platform/RelayPushRegistrationServiceTests.cs
justindbaur: Good call, _technically_ it didn't matter for this test and only that the shape of the `PushRegistrationData` looked mobile-y but it's still good to make it a more realistic payload overall.

Mobile should not log because it can successfully make the API request to cloud and get itself registered. To make that clear-er I added an assertion that the cloud API receives that method call (b4c3a58).

djsmith85 (approved):

PR #7620, opened 2026-05-12
[PM-19551] Add externalId support to groups PATCH endpoint

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19551
## 📔 Objective

1. Add externalId support to groups PATCH endpoint
2. Add unit tests

No manual tests were needed because we have integration tests that provide us with strong confidence.


bitwarden_license/test/Scim.IntegrationTest/Controllers/v2/GroupsControllerPatchTests.cs
```diff
@@ -211,6 +211,197 @@ public async Task Patch_RemoveListMembers_Success()
         Assert.Empty(databaseContext.GroupUsers);
     }
 
+    [Fact]
+    public async Task Patch_ReplaceExternalIdFromPath_Success()
+    {
+        // Arrange
```
r-tome, line 217: :art: this file doesn't use the Arrange Act and Assert comments so to be consistent we should not add these

bitwarden_license/test/Scim.Test/Groups/PatchGroupCommandTests.cs
```diff
@@ -431,6 +431,285 @@ await sutProvider.GetDependency<IGroupRepository>()
                 Arg.Is<DateTime>(d => d == _expectedRevisionDate));
     }
 
+    [Theory]
+    [BitAutoData]
+    public async Task PatchGroup_ReplaceExternalIdFromPath_Success(
+        Organization organization, Group group, string newExternalId)
+    {
+        // Arrange
```
r-tome, line 439: :art: Same thing on this file

r-tome (approved):
Looks good! Just a couple of non-blocking comments

bitwarden_license/test/Scim.IntegrationTest/Controllers/v2/GroupsControllerPatchTests.cs
JimmyVo16: I think it’s good to use Arrange, Act, and Assert since it improves readability. Since this is a new pattern, I think it’s okay to apply it only to new code and leave the old code as-is to avoid extra friction.

PR #7575, opened 2026-05-01
[PM-35393] MasterPasswordService auth integration

enmande:
## 🎟️ Tracking

[PM-35393](https://bitwarden.atlassian.net/browse/PM-35393)

## 📔 Objective

Integrate Auth-domain callers with `MasterPasswordService`.
Wires five password-mutation flows to the service added in #7530 :
- Self-service password change
- TDE offboarding
- Emergency Access Takeover
- Temporary password replacement
- SSO JIT provisioning.

All relevant request models now accept both payload variants ("new": `AuthenticationData` + `UnlockData`, or "legacy": `NewMasterPasswordHash` + `Key`) for backward compatibility.
KDF validation is enforced at the request model layer whenever new payloads are present.
Legacy entry points on `IUserService` are marked `[Obsolete]` with specific replacement guidance. Legacy fields are tracked for removal in [PM-33141](https://bitwarden.atlassian.net/browse/PM-33141).

### Master Password Self-Service

User _enmande_ updates their Master Password using self-service.
- Current Master Password is required and passes validation.
- New Master Password is hashed and set.
- User is logged out on success (behavioral parity with `main`).
- User can log in with the new Master Password. 

https://github.com/user-attachments/assets/47dc0302-588d-4d6b-9195-9379332a0816

### TDE Offboarding

- User _zed_ belongs to an SSO-required _Trusted Device Encryption_ organization.
- Organization Owner _enmande_ converts the organization to _Master Password Encryption_. This is generally described as "TDE Offboarding."
- User _zed_ is required to set a new Master Password at next login; password is successfully set and allows _zed_ to log in and view their vault. 

_NOTE:_ This organization retains SSO requirement throughout; _zed_ remains authenticated to SSO for the duration of this test, so SSO challenge is not represented in this video.

https://github.com/user-attachments/assets/1b1b597c-795a-47e3-8d71-f214ac48d6b9

### Emergency Access Takeover

- User _zed_, who is enrolled in Emergency Access Takeover by organizational policy, has their account taken over ("Recovered") by organization Owner _enmande_.
- _enmande_ sets _zed_'s new temporary Master Password.
- New temporary Master Password is accepted at next login for _zed_'s account.
- Upon login, _zed_'s account requires the setting of a new Master Password.
- User is logged out on change of the Master Password.
- New Master Password hash is set, and enabled subsequent login and access to the vault.

https://github.com/user-attachments/assets/5376ee3b-19fd-4db0-bd98-c914be644c3c

[...]

PR #7628, opened 2026-05-13
feat(kdf-settings-validator): Enforce salt cannot be empty string.

enmande:
## 🎟️ Tracking

[PM-35393](https://bitwarden.atlassian.net/browse/PM-35393)

## 📔 Objective

`[Required]` attribution on `salt`-related fields enforces _by default_ that salt cannot be null, empty, or whitespace. However, this is a [configurable opinion](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.requiredattribute.allowemptystrings?view=netframework-4.8.1&viewFallbackFrom=net-10.0#system-componentmodel-dataannotations-requiredattribute-allowemptystrings). While it should probably not be configured to allow whitespace, because this is a critical path, reinforcement should be present to guard against configuration drift, which would otherwise be invisible to this logic.
`string.IsNullOrWhitespace` properly [captures all of these](https://learn.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?view=netframework-4.8.1&viewFallbackFrom=net-10.0#system-string-isnullorwhitespace(system-string)).

Tests have been added to support.

[PM-35393]: https://bitwarden.atlassian.net/browse/PM-35393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
+++ b/src/Core/KeyManagement/Models/Api/Request/MasterPasswordAuthenticationDataRequestModel.cs
@@ -12,7 +12,7 @@ public class MasterPasswordAuthenticationDataRequestModel
     public required KdfRequestModel Kdf { get; init; }
     [Required]
     public required string MasterPasswordAuthenticationHash { get; init; }
-    [Required]
+    [Required(AllowEmptyStrings = false)]
     [StringLength(256)]
     public required string Salt { get; init; }
 
--- a/src/Core/KeyManagement/Models/Api/Request/MasterPasswordUnlockDataRequestModel.cs
+++ b/src/Core/KeyManagement/Models/Api/Request/MasterPasswordUnlockDataRequestModel.cs
@@ -14,7 +14,7 @@ public class MasterPasswordUnlockDataRequestModel
     [Required]
     [EncryptedString]
     public required string MasterKeyWrappedUserKey { get; init; }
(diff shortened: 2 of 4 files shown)


ike-kottlowski (dismissed):
🎨 I'm thinking we can tighten up the data annotation on the request model themselves as well. This would give us the benefit of failing closer to the controller and the `KdfSettingsValidator` changes act as a safety net that should be unreachable in practice.

``` csharp
// MasterPasswordAuthenticationDataRequestModel.cs
[Required(AllowEmptyStrings = false)]
[StringLength(256)]
public required string Salt { get; init; }

// MasterPasswordUnlockDataRequestModel.cs
[Required(AllowEmptyStrings = false)]
[StringLength(256)]
public required string Salt { get; init; }
```

quexten (dismissed):

JaredSnider-Bitwarden:
Merge plan per discussion with @enmande :  let base PR https://github.com/bitwarden/server/pull/7575 merge, then merge this to main and let testing of this be regressions only since this is lightweight. We don't want to merge this into `auth/pm-35393/master-password-service-auth-integration` as that is mid QA.

Thomas-Avery (approved):
LGTM

ike-kottlowski (approved):

PR #7632, opened 2026-05-13
Auth/PM-37166 - Devices - add client version

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37166

## 📔 Objective

To add a client version to the devices table which can be bumped on login and token refresh. 
To avoid hitting the DB n times where n is the # of device columns we want to bump, I chose to refactor the `Device.LastActivityDate` implementation (#7302 ) to be more generic and only hit the DB on either last activity date bump (only once per day) or client version change - the bump is still protected with a Cosmos backed cache to prevent too much load on the DB. 

## 📸 Screenshots

https://github.com/user-attachments/assets/6aa3bc6a-6a69-47e6-b5ae-3f39a44b7972

PR #7523, opened 2026-04-22
[PM-35489] Move collections to AC ownership

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35489
## 📔 Objective

The collection domain in server is currently unowned. Move all collection-related files to AC ownership.

`CollectionCipher` is intentionally excluded because Vault Team own this relationship.

Some namespaces have been updated where it didn't trigger excessive codeowner reviews. Follow up PRs will align most of the remaining namespaces.

PR #7634, opened 2026-05-14
Auth/PM-37165 - Add Last API Key Rotated Date to User

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37165

## 📔 Objective

We want to add a new column on the user table which tracks last API key rotated dates. 

## 📸 Screenshots

https://github.com/user-attachments/assets/056855a3-13d0-4e9a-a54f-15fe8bc0a579


src/Sql/dbo/Tables/User.sql
```diff
@@ -41,6 +41,7 @@
     [LastKdfChangeDate]                DATETIME2 (7)    NULL,
     [LastKeyRotationDate]              DATETIME2 (7)    NULL,
     [LastEmailChangeDate]              DATETIME2 (7)    NULL,
+    [LastApiKeyRotationDate]           DATETIME2 (7)    NULL,
```
withinfocus: ❌ Always add to the end.

src/Sql/dbo/Views/UserView.sql
```diff
@@ -43,6 +43,7 @@ SELECT
     [LastKdfChangeDate],
     [LastKeyRotationDate],
     [LastEmailChangeDate],
+    [LastApiKeyRotationDate],
```
withinfocus: ❌ Same.

util/MySqlMigrations/Migrations/DatabaseContextModelSnapshot.cs
```diff
@@ -69,6 +69,89 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.ToTable("OrganizationMemberBaseDetails");
                 });
 
+            modelBuilder.Entity("Bit.Infrastructure.EntityFramework.AdminConsole.Models.Collection", b =>
```
Patrick-Pimentel-Bitwarden, line 72: I had claude do a little writeup for why we are seeing this in this pr since it has nothing to do with the scope of your pr.

> Why it shows up in this PR
> 
> DatabaseContextModelSnapshot.cs is a derived artifact — EF only regenerates it when you run dotnet ef migrations add .... PR #7523 changed the namespaces but did not add a new migration, so the snapshot file was left stale (still listing the entities under their old Models.* namespace at their old alphabetical
  position).
> 
> This PR is the first migration generated after that namespace refactor landed. When Jared ran dotnet ef migrations add AddLastApiKeyRotationDateToUserTable, EF re-emitted the entire snapshot from the live model, which:
> 
>  1. Wrote the three Collection entities under their new AdminConsole.Models.* names → at lines ~71-82 (earlier alphabetical position).
>  2. Removed the old blocks from their previous spot lower in the file (see the matching deletions at the old Models.Collection location: - lines 1272-1311 in the diff).

JaredSnider-Bitwarden: I ran into this on the https://github.com/bitwarden/server/pull/7632 PR as well. 
https://github.com/bitwarden/server/pull/7632/changes/33a087f5e32b86605ae3e58356fb8650c355f1d3

Patrick-Pimentel-Bitwarden (commented):
Ready to approve once the new field is being added to the back of the table 👍

src/Sql/dbo/Tables/User.sql
JaredSnider-Bitwarden: @withinfocus , I considered this and consulted https://contributing.bitwarden.com/contributing/code-style/sql#adding-a-column-to-a-table to evaluate if it mattered, but I didn't see anything there. What's the rationale for doing this so I can help enforce it + possibly update the docs to include it?

JaredSnider-Bitwarden: Oh, b/c of ordinal position consumers. Never mind. That makes sense.

JaredSnider-Bitwarden: Fixed with https://github.com/bitwarden/server/pull/7634/changes/c4348b3cfb33d8cffb0383a582300db3111bac84

src/Sql/dbo/Views/UserView.sql
JaredSnider-Bitwarden: Fixed with https://github.com/bitwarden/server/pull/7634/changes/c4348b3cfb33d8cffb0383a582300db3111bac84

Patrick-Pimentel-Bitwarden (approved):
👍 Looks solid

mkincaid-bw (approved):
LGTM

PR #7645, opened 2026-05-15
Migrate to SLNX Style Solution

justindbaur:
## 📔 Objective

Migrate to the [simpler SLNX](https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/) format that became available in the .NET 9 SDK. Removes ~300 lines of fluff that largely just made for annoying git merge issues. 

This was done by running `dotnet sln migrate`.


.github/workflows/test.yml
```diff
@@ -49,9 +49,6 @@ jobs:
           echo "GitHub ref: $GITHUB_REF"
           echo "GitHub event: $GITHUB_EVENT"
 
-      - name: Remove SQL project
-        run: dotnet sln bitwarden-server.sln remove src/Sql/Sql.sqlproj
-
```
justindbaur, line 54: I believe this is from when the SQL project was not able to be built on linux but that has not been the case for a while so I believe this step isn't needed. If I am wrong about that we can add it back as this is still possible with the new solution format.

dani-garcia (dismissed):
Nice! This is so much more readable now.

Also looks like we got merge conflicts on the `sln` file already 😢

dereknance (approved):
Approving before more conflicts arise! 😆

PR #7649, opened 2026-05-15
Auth/PM-37621 - Fix Device.LastActivityDate surfacing legacy NULL rows as DateTime.UtcNow

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37621

## 📔 Objective

To fix an issue with where null `Device.LastActivityDate` values were coming out of the server as new `DateTime.UtcNow`

## 📸 Screenshots

https://github.com/user-attachments/assets/b9f3b075-e437-4f98-a8cb-0b19580c4ca3

### DIFF
--- a/src/Api/Models/Request/DeviceRequestModels.cs
+++ b/src/Api/Models/Request/DeviceRequestModels.cs
@@ -26,7 +26,9 @@ public class DeviceRequestModel
     {
         return ToDevice(new Device
         {
-            UserId = userId == null ? default(Guid) : userId.Value
+            UserId = userId == null ? default(Guid) : userId.Value,
+            // Device creation counts as first activity.
+            LastActivityDate = DateTime.UtcNow,
         });
     }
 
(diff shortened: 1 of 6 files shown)


src/Api/Models/Request/DeviceRequestModels.cs
JaredSnider-Bitwarden, line 1: ℹ️  this is used in the device creation post seen here: https://github.com/bitwarden/server/blob/main/src/Api/Controllers/DevicesController.cs#L91-L99

src/Identity/IdentityServer/RequestValidators/DeviceValidator.cs
JaredSnider-Bitwarden, line 1: This is when we create a device on first success login.

Patrick-Pimentel-Bitwarden (approved):
Changes look good!

PR #7650, opened 2026-05-15
[PM-35830] Add ChangeEmailCommand

ike-kottlowski:
## 🎟️ Tracking

[PM-35830](https://bitwarden.atlassian.net/browse/PM-35830)

## 📔 Objective

Introduces a new `ChangeEmailCommand` that encapsulates the user email change flow.

- Adds `IChangeEmailCommand` / `ChangeEmailCommand` under `Bit.Core.Auth.UserFeatures.UserEmail`, including claimed-domain enforcement, duplicate-email rejection, Stripe customer email sync with rollback on failure, and post-change push notifications (logout vs. settings sync depending on whether the user has a master password).
- Adds `ChangeEmailCommandTests` covering claimed-domain enforcement, disabled claiming orgs, duplicate email, same-user email, Stripe sync, Stripe rollback, and push branching.

This PR introduces the command only; switching callers from `UserService.ChangeEmailAsync` to the new command will follow in a separate PR.

Introduces a Admin Console Query to check if the email domain is allowed to be used.

[PM-35830]: https://bitwarden.atlassian.net/browse/PM-35830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
```diff
@@ -0,0 +1,110 @@
...
+    private readonly IOrganizationDomainRepository _organizationDomainRepository = organizationDomainRepository;
+    private readonly IPushNotificationService _pushService = pushService;
+    private readonly IStripeSyncService _stripeSyncService = stripeSyncService;
+    private readonly IOrganizationRepository _organizationRepository = organizationRepository;
+
+    public async Task ChangeEmailAsync(User user, string newEmail)
+    {
+        // Will throw if the new email violates a claiming organization's domain policy.
+        await EnsureNewEmailMatchesClaimedDomainAsync(user, newEmail);
+
```
eliykat, line 33: As discussed in Slack, I think that the Admin Console layer can handle this.

Validation such as checking the admin role, making sure the user is claimed, and making sure the new email address is claimed, is all part of the admin/organization domain that we will handle.

The cross-team dependency on Auth is much narrower: we don't own the `User` domain and we don't know how to update a user's email (in the narrow sense). So that is the main thing we are looking for here.

I think if you remove the claimed domain check in this method, everything else LGTM. Thank you!

eliykat (dismissed):

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
ike-kottlowski: Spoke with @jrmccannon yesterday and we found the correct approach. This is addressed in this commit: https://github.com/bitwarden/server/pull/7650/commits/fd534bfd785ad176a0e8b2eb1b3583b104793af0

eliykat (commented):
Letting @jrmccannon handle this one as he's discussed the approach with you already.

ike-kottlowski:
@claude please review this PR.

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/OrganizationDomainAllowEmailChangeQuery.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationDomains;
+
+public class OrganizationDomainAllowEmailChangeQuery(
+    IOrganizationRepository organizationRepository,
+    IOrganizationDomainRepository organizationDomainRepository)
+    : IOrganizationDomainAllowEmailChangeQuery
+{
+    private readonly IOrganizationRepository _organizationRepository = organizationRepository;
+    private readonly IOrganizationDomainRepository _organizationDomainRepository = organizationDomainRepository;
```
jrmccannon: These are unnecessary.

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/OrganizationDomainAllowEmailChangeQuery.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <inheritdoc />
+    public async Task<bool> IsAllowedAsync(User user, string newEmailDomain)
+    {
+        // We want to see if the user is currently a claimed account
+        var organizationsWithVerifiedUserEmailDomain =
+            await _organizationRepository.GetByVerifiedUserEmailDomainAsync(user.Id);
+
+        var claimingOrganizations = organizationsWithVerifiedUserEmailDomain
+            .Where(organization => organization is { Enabled: true, UseOrganizationDomains: true });
```
jrmccannon: We can move the select to this and `ToArray()` so that we don't have to inline it below.

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            throw new BadRequestException("Email already in use.");
+        }
+
+        var previousEmail = user.Email;
+        var previousRevisionDate = user.RevisionDate;
+        var previousAccountRevisionDate = user.AccountRevisionDate;
+        var previousLastEmailChangeDate = user.LastEmailChangeDate;
+
+        var now = DateTime.UtcNow;
```
jrmccannon: we should use TimeProvider instead of the System.DateTime

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// Blocks an email change onto a domain claimed by an organization that has the
+    /// BlockClaimedDomainAccountCreation policy enabled. Mirrors the gate enforced by
+    /// RegisterUserCommand so the policy cannot be bypassed via email change.
+    /// </summary>
+    private async Task EnsureNewEmailDomainAllowedByPolicyAsync(User user, string newEmail)
+    {
+        // If the new email domain is the same as the current email domain, we can skip
+        // the checks since it would be a noop in terms of policy and claiming organizations.
+        // Null check for nullable reference types, but the email should always be valid at this point in the code.
+        var newDomain = CoreHelpers.GetEmailDomain(newEmail) ?? throw new BadRequestException("Invalid email address.");
```
jrmccannon: We should use the `EmailValidation.GetDomain` so that it normalizes the domain for string comparison. (this would fail if it was Acme.com vs acme.com)

jrmccannon (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/OrganizationDomainAllowEmailChangeQuery.cs
ike-kottlowski: addressed in https://github.com/bitwarden/server/pull/7650/commits/99dd61710155f0e9ab2706439cb03cb281d6f4d1

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/OrganizationDomainAllowEmailChangeQuery.cs
ike-kottlowski: addressed in https://github.com/bitwarden/server/pull/7650/commits/99dd61710155f0e9ab2706439cb03cb281d6f4d1

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
ike-kottlowski: addressed in https://github.com/bitwarden/server/pull/7650/commits/46a4e23fa42b9eabd34acf5712b592b77927029b

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
ike-kottlowski: addressed in https://github.com/bitwarden/server/pull/7650/commits/46a4e23fa42b9eabd34acf5712b592b77927029b

jrmccannon (dismissed):
Thanks for adding the additional domain checks!

src/Core/Auth/UserFeatures/UserEmail/IChangeEmailCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// Updates the user's email after all upstream authentication and authorization checks have
+    /// passed. On success, signals a settings sync to the user's active sessions; sessions are
+    /// not invalidated.
+    /// </summary>
+    /// <remarks>
+    /// This command is intended to be consumed only once the master password salt is no longer
+    /// coupled to <see cref="User.Email"/>. While that coupling exists, changing the email
+    /// invalidates derived key material for sessions that still hold the old salt; callers must
+    /// either keep using the legacy flow (which logs the user out) or ensure the decoupling has
+    /// shipped before invoking this command.
```
JaredSnider-Bitwarden: 🎨 : this section only applies to password users and we probably should call that out + write a brief note about how changing the email for a passwordless user (TDE) is supported.

🏅 : the surrounding docs are great!

test/Core.Test/Auth/UserFeatures/UserEmail/ChangeEmailCommandTests.cs
JaredSnider-Bitwarden, line 1: Excellent start of this test suite. Claude found these potential gaps, thoughts?

## Gaps worth knowing about (none likely to block, but real)

1. **`BillingEmailAddress()` null branch** — the Stripe guard is `GatewayCustomerId != null && billingEmail != null`. Only the `GatewayCustomerId == null` half is exercised ([ChangeEmailCommandTests.cs:117](test/Core.Test/Auth/UserFeatures/UserEmail/ChangeEmailCommandTests.cs#L117)). A user with `GatewayCustomerId="cus_123"` but a null billing email would hit the same throw — currently untested.

2. **Rollback write itself fails** — the original Claude review flagged that if the rollback `ReplaceAsync` throws, the original Stripe exception is lost. No test pins the current behavior either way.

3. **Same-domain short-circuit is implicit, not asserted** — `_currentEmail` and `_newEmail` both use `example.com`, so most tests rely on the short-circuit (otherwise `IsAllowedAsync` would need to be stubbed). There's no test that explicitly says "when domains match, `IsAllowedAsync` is never called." Easy to add, and would lock in the optimization.

4. **Case-insensitive domain match** — `EmailValidation.GetDomain` lowercases, but no test exercises `OLD@Example.com` → `new@example.com` to confirm the short-circuit still fires on case differences.

5. **`LogWarning` on Stripe failure** — the rollback log message ([ChangeEmailCommand.cs:68](src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs#L68)) is never asserted. Minor; logger assertions in this codebase are inconsistent.

6. **`EmailVerified = true` invariant** — only asserted in the non-Stripe test (line 81). Not asserted on the Stripe happy path.

**Verdict:** thorough enough to ship. If you want to tighten before the caller migration PR, items 1 and 3 are the highest-value additions — they protect actual contract invariants rather than just adding coverage for coverage's sake.

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        await _pushService.PushSyncSettingsAsync(user.Id);
+    }
+
+    /// <summary>
+    /// Blocks an email change onto a domain claimed by an organization that has the
+    /// BlockClaimedDomainAccountCreation policy enabled. Mirrors the gate enforced by
+    /// RegisterUserCommand so the policy cannot be bypassed via email change.
+    /// </summary>
+    private async Task EnsureNewEmailDomainAllowedByPolicyAsync(User user, string newEmail)
```
JaredSnider-Bitwarden: 🎨 : For clarity, it might be best to leave this method generic and just write it as:
`EnsureNewEmailDomainAllowed` + update the summary b/c the actual checks are handled by the `_organizationDomainAllowEmailChangeQuery` which rejects if the user is a claimed account and is changing away from a claimed email domain (no policy requirement) OR if they are changing to an email owned by an org which has the block policy enabled.  I.e., it's not just a policy block that could reject the email change.

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+        var newDomain = EmailValidation.GetDomain(newEmail);
+        if (newDomain == EmailValidation.GetDomain(user.Email))
+        {
+            return;
+        }
+
+        var isAllowed = await _organizationDomainAllowEmailChangeQuery.IsAllowedAsync(user, newDomain);
+        if (!isAllowed)
+        {
+            throw new BadRequestException("This email address is claimed by an organization using Bitwarden.");
```
JaredSnider-Bitwarden: 🎨 : technically, this error is also not accurate. It's one of the possible rejection reasons, but it doesn't capture both scenarios.

JaredSnider-Bitwarden (changes requested):
Looking really good so far. Just some clarity work to be done:

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
ike-kottlowski: Updated return type to allow for branching.

https://github.com/bitwarden/server/pull/7650/commits/42917041e2a5b61158bd16f08ce2097eb06025b4

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
ike-kottlowski: addressed comment in this https://github.com/bitwarden/server/commit/42917041e2a5b61158bd16f08ce2097eb06025b4

src/Core/Auth/UserFeatures/UserEmail/IChangeEmailCommand.cs
ike-kottlowski: added caveat for TDE/passwordless https://github.com/bitwarden/server/pull/7650/commits/7ea6ef89393c6d2f9ad021bd13bf5a7b1b7ec84e

ike-kottlowski:
@claude please review my current changes. Please consider @JaredSnider-Bitwarden comments when responding.

ike-kottlowski:
@claude Please review again. for gap 1 the `user.BillingEmailAddress != null` check is required to satisfy the nullable constraint for C# otherwise there is an error. It's defensive but fine. There is an additional test for the domain case matching, gap 4. We've also removed the throw when a gateway has a value but the customerId is null.

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        var now = _timeProvider.GetUtcNow().UtcDateTime;
+        user.Email = newEmail;
+        user.EmailVerified = true;
+        user.LastEmailChangeDate = now;
+        user.RevisionDate = user.AccountRevisionDate = now;
+
+        await _userRepository.ReplaceAsync(user);
+
+        if (user.Gateway == GatewayType.Stripe)
```
JaredSnider-Bitwarden, line 53: 🏅 The stripe logic change to allow free users with a gateway but no customer id to change their email appears to be a good one, but please be sure we note it as a QA test case if it can be tested since it is a behavior change.

JaredSnider-Bitwarden (approved):
Excellent work! LGTM! Please see my one comment below about QA note implications 👇

jrmccannon (approved):

src/Core/Auth/UserFeatures/UserEmail/ChangeEmailCommand.cs
ike-kottlowski: I'll add it to the self-service ticket. These changes won't be able to be tested until AC has their email change flow implemented. These changes are foundational and not functional at this point in time.

PR #7656, opened 2026-05-18
[PM-37251] Add public invite link GET status endpoint

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37251

## 📔 Objective

Add an unauthenticated endpoint that returns the organization name, seat availability, and SSO details.

This will be used in the clients' invite link landing page on load, before login/signup, so the client can render the right error state (link not found, no seats) or route the joining user through SSO.

[Clients PR](https://github.com/bitwarden/clients/pull/20695)


src/Api/AdminConsole/Models/Response/Organizations/OrganizationInviteLinkStatusResponseModel.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Bit.Core.Models.Api;
+
+namespace Bit.Api.AdminConsole.Models.Response.Organizations;
+
+public class OrganizationInviteLinkStatusResponseModel : ResponseModel
+{
+    public Guid OrganizationId { get; set; }
```
JaredSnider-Bitwarden: Question: Is there any risk of exposing this? The original tech breakdown just had `organizationName`. Should we only include data that we know we need?

JaredSnider-Bitwarden: Or did y'all discover a need to include these b/c it will be needed for subsequent API calls or something?

JaredSnider-Bitwarden (commented):
Everything looks great in terms of fulfilling the data we will need, but I do have one question about `organizationId` below:

src/Api/AdminConsole/Controllers/OrganizationInviteLinksPublicController.cs
```diff
@@ -0,0 +1,32 @@
...
+[Route("organizations/invite-link")]
+[Authorize("Application")]
+[RequireFeature(FeatureFlagKeys.GenerateInviteLink)]
+public class OrganizationInviteLinksPublicController(
+    IGetOrganizationInviteLinkStatusQuery getOrganizationInviteLinkStatusQuery)
+    : BaseAdminConsoleController
+{
+    [AllowAnonymous]
+    [HttpGet("{code:guid}/status")]
+    public async Task<IResult> GetStatus(Guid code)
```
eliykat: `code` is a secret value, so we should keep it out of the GET parameters. It should be in the body of a POST request instead.

src/Api/AdminConsole/Controllers/OrganizationInviteLinksPublicController.cs
eliykat, line 1: Naming: `public` might be confused with the public API. I think you could just add this to the existing invite links controller.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/GetOrganizationInviteLinkStatusQuery.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            return new InviteLinkNotAvailable();
+        }
+
+        var occupied = (await organizationRepository
+            .GetOccupiedSeatCountByOrganizationIdAsync(organization.Id)).Total;
+        var seatsAvailable = !organization.Seats.HasValue
+            || occupied < organization.Seats.Value
+            || (organization.MaxAutoscaleSeats.HasValue
+                && organization.Seats.Value < organization.MaxAutoscaleSeats.Value);
```
eliykat: What if `MaxAutoscaleSeats` is null, indicating there is no autoscale limit?

eliykat (changes requested):
Not a full review, just a few things that jumped out at me coming from the client PR.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/GetOrganizationInviteLinkStatusQuery.cs
```diff
@@ -0,0 +1,72 @@
...
+    public async Task<CommandResult<OrganizationInviteLinkStatus>> GetStatusAsync(Guid code)
+    {
+        var inviteLink = await organizationInviteLinkRepository.GetByCodeAsync(code);
+        if (inviteLink is null)
+        {
+            return new InviteLinkNotFound();
+        }
+
+        var organization = await organizationRepository.GetByIdAsync(inviteLink.OrganizationId);
+        if (organization is null)
```
eliykat:
```suggestion
        if (organization is null or { Enabled: false })
```

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/GetOrganizationInviteLinkStatusQuery.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            return new InviteLinkNotFound();
+        }
+
+        var ability = await applicationCacheService.GetOrganizationAbilityAsync(organization.Id);
+        if (ability is null || !ability.UseInviteLinks)
+        {
+            return new InviteLinkNotAvailable();
+        }
```
eliykat: No need to hit the cache, you already have the org object in scope.

eliykat (dismissed):

src/Api/AdminConsole/Models/Response/Organizations/OrganizationInviteLinkStatusResponseModel.cs
r-tome: There is no risk in exposing the `OrganizationId`. Initially the invite link even had both `OrgId+Code` and I removed it and just kept `Code` because it'll be unique anyway.
I thought we'd need the `OrgId` at some point but we don't -- Code should be enough for this flow. I'll remove it.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/GetOrganizationInviteLinkStatusQuery.cs
r-tome: It would indicate no seats were available. I've fixed it. Well spotted!

src/Api/AdminConsole/Controllers/OrganizationInviteLinksPublicController.cs
r-tome: I thought that placing it in the same controller would be confusing but actually the `[AllowAnonymous]` attribute should be enough to distinguish it from all the other endpoints.

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/GetOrganizationInviteLinkStatusQuery.cs
r-tome: I forgot to remove it after getting the org object to calculate the available seats :doh:

JaredScar:
Please let me know when @eliykat changes are all taken care of. I could also potentially remove myself as a reviewer if he's got AC covered here 😅

eliykat (commented):
My feedback has been addressed, thanks!

@JaredScar Please still do a full review - pretend I'm not here 😁 I will sometimes drop in on PRs if I have something to add, but that may just be a few comments rather than a full PR review.

JaredSnider-Bitwarden (approved):
I only reviewed `src/Api/AdminConsole/Models/Request/Organizations/GetOrganizationInviteLinkStatusRequestModel.cs` and `src/Api/AdminConsole/Models/Response/Organizations/OrganizationInviteLinkStatusResponseModel.cs` as that is what we agreed upon as the expected interface for Auth to consume.

JaredScar (approved):
LGTM

PR #7527, opened 2026-04-22
[PM-33951] automatically confirm pending users on admin login

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33951

## 📔 Objective

When an admin logins to the browser extension for an organization, we want users who accepted an invite to be confirmed as users automatically. When the extension is locked, unlocking should also proceed to accept the users pending confirmation.

PR #7659, opened 2026-05-18
[PM-38273] feat(admin-console): Add InjectOrganizationAttribute and OrganizationModelBinder

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38273

## 📔 Objective

This PR contains the implementation of InjectOrganizationAttribute (renamed to Bind because of conflicting Billing one) needed for [#7527](https://github.com/bitwarden/server/pull/7527 ), split to their own PR to manage PR size.


jrmccannon (changes requested):
We should add unit tests to this and apply it to a simple endpoint.

src/Api/AdminConsole/Controllers/OrganizationUsersController.cs
```diff
@@ -230,7 +231,8 @@ private ListResponseModel<OrganizationUserUserDetailsResponseModel> GetResultLis
 
     [HttpGet("{id}/reset-password-details")]
     [Authorize<ManageAccountRecoveryRequirement>]
-    public async Task<OrganizationUserResetPasswordDetailsResponseModel> GetResetPasswordDetails(Guid orgId, Guid id)
+    public async Task<OrganizationUserResetPasswordDetailsResponseModel> GetResetPasswordDetails(Guid orgId, Guid id,
+        [BindOrganization] Organization organization)
     {
         var organizationUser = await _organizationUserRepository.GetByIdAsync(id);
         if (organizationUser is null || organizationUser.OrganizationId != orgId || organizationUser.UserId is null)
```
jrmccannon: This check should use the Organization Id from the `Organization` object and the `orgId` can be removed as a parameter.

src/Api/AdminConsole/Attributes/BindOrganizationAttribute.cs
```diff
@@ -0,0 +1,60 @@
...
+/// <example>
+/// <code><![CDATA[
+/// [HttpPost("bulk-auto-confirm")]
+/// public async Task<IResult> BulkAutomaticallyConfirmOrganizationUsersAsync(
+///     [BindOrganization] Organization organization,
+///     [FromBody] OrganizationUserBulkConfirmRequestModel model)
+/// ]]></code>
+/// </example>
+[AttributeUsage(AttributeTargets.Parameter)]
+public sealed class BindOrganizationAttribute() : ModelBinderAttribute(typeof(OrganizationModelBinder));
```
jrmccannon, line 26: Lets keep this consistent with the existing `InjectOrganizationUser` that already exists and changed the name to `InjectOrganization`

jrmccannon (changes requested):

src/Api/AdminConsole/Attributes/BindOrganizationAttribute.cs
JaredScar: Clarified in our DMs, but posting here for closing loop. InjectOrganizationAttribute exists via Billing, but is different with how we want to use it. Therefore this will stay named as BindOrganizationAttribute to avoid that ambiguous issue

jrmccannon (dismissed):

jrmccannon (approved):

PR #7661, opened 2026-05-18
[PM-33951] feat(admin-console): Add bulk confirmation and pending auto-confirmation

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33951

## 📔 Objective

This PR contains the database changes needed for [#7527](https://github.com/bitwarden/server/pull/7527 ), split to their own PR to manage PR size.


src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -1045,6 +1045,42 @@ public async Task<bool> ConfirmOrganizationUserAsync(AcceptedOrganizationUserToC
 
     }
 
+    public async Task<ICollection<Guid>> ConfirmManyOrganizationUsersAsync(
+        IEnumerable<AcceptedOrganizationUserToConfirm> usersToConfirm)
+    {
+        using var scope = ServiceScopeFactory.CreateScope();
+        using var dbContext = GetDatabaseContext(scope);
+
+        var usersToConfirmList = usersToConfirm.ToList();
```
r-tome: 💡 : You can avoid this `ToList` if you defer that to the caller by changing the parameter type to `IReadOnlyCollection`

r-tome (changes requested):
Claude pointed out a few things to fix. Let me know when to review this again

mkincaid-bw:
There are similar database changes in PR 7527, as well as this one. Do both PR's need a review from dbops or only this one?

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByOrganizationIdStatus.sql
```diff
@@ -0,0 +1,15 @@
+CREATE PROCEDURE [dbo].[OrganizationUser_ReadByOrganizationIdStatus]
```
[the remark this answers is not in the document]
JaredScar: This is fine. This PR is not going to handle the calling of it @claude -- It just introduces it

JaredScar:
> There are similar database changes in PR 7527, as well as this one. Do both PR's need a review from dbops or only this one?

Valid question! @mkincaid-bw -- Sorry for the confusion. @eliykat suggested we split up that PR into smaller PRs like this one that we merge first, then that original PR will be smaller and easier to review since that big chunk of code are broken into pieces. Theoretically speaking, this should be the only one DB ops needs to worry about 😄

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByOrganizationIdStatus.sql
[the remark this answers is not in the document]
eliykat: @JaredScar Claude is right about this, however I think this has been superseded by the more specific `ReadByPendingAutoConfirm` query. In that case you can just delete this one.

This is probably also the cause of your CI failure.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepository/OrganizationUserRepositoryTests.cs
```diff
@@ -1651,6 +1651,134 @@ public async Task ConfirmOrganizationUserAsync_WhenUserDoesNotExist_ReturnsFalse
         Assert.False(result);
     }
 
+    // -----------------------------------------------------------------------
+    // ConfirmManyOrganizationUsersAsync
+    // -----------------------------------------------------------------------
```
eliykat: We don't use code regions, either via `#region` or by comments.

These test files do get enormous though. Note that this is actually inside an `OrganizationUserRepository` folder - the idea is that you create a separate test file for each method. So you'll move these into 2 files: `ConfirmManyOrganizationUsersTests` and `GetManyPendingAutoConfirmTests`.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepository/OrganizationUserRepositoryTests.cs
```diff
@@ -1651,6 +1651,134 @@ public async Task ConfirmOrganizationUserAsync_WhenUserDoesNotExist_ReturnsFalse
...
+        };
+
+        // Act
+        var confirmedIds = await organizationUserRepository.ConfirmManyOrganizationUsersAsync(usersToConfirm);
+
+        // Assert — only the Accepted user's ID is returned
+        Assert.Single(confirmedIds);
+        Assert.Contains(acceptedOrgUser.Id, confirmedIds);
+
+        // The previously-accepted user is now Confirmed
```
eliykat: Also assert that the other users have not been updated.

eliykat (changes requested):
Thanks @JaredScar for splitting this to a separate PR - I cannot tell you how much easier it is to review smaller PRs.

Aside from my comments below, please check the CI runs:
- db integration test is failing - potentially faulty prod or test logic
- db validation is failing - this ensures that the migration scripts match the `src/Sql` files
- validate new migration naming and order - this is failing because later migrations have overtaken you and you need to bump the dates on your script

More info available in each of the runs, or hit me up on Slack if you have questions.

EDIT: please also add more info to your PR description, e.g.
> This PR contains the database changes needed for #7527, split to their own PR to manage PR size.

The PR link is particularly useful to understand what's going on.

src/Sql/dbo/Stored Procedures/OrganizationUser_ConfirmByIds.sql
```diff
@@ -0,0 +1,43 @@
+CREATE PROCEDURE [dbo].[OrganizationUser_ConfirmByIds]
```
mkincaid-bw: This stored proc name does not follow our [naming standards](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures) (I realize we have `OrganizationUser_ConfirmById` as well, which I previously approved but should not have). It should start with  `OrganizationUser_Update...`, and perhaps be named something like `OrganizationUser_UpdateStatusKey`, or something to that effect.

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByPendingAutoConfirm.sql
```diff
@@ -0,0 +1,16 @@
...
+    SET NOCOUNT ON
+
+    SELECT
+        *
+    FROM
+        [dbo].[OrganizationUserView]
+    WHERE
+        [OrganizationId] = @OrganizationId
+        AND [Status] = 1  -- Accepted
+        AND [Type] = 0    -- User
```
mkincaid-bw: ❓According to `src/Core/AdminConsole/Enums/OrganizationUserType.cs`, Type `0` is `Owner` and Type `2` is `User`. Is the value for `Type` supposed to be `User` or `Owner`?

mkincaid-bw (changes requested):

src/Sql/dbo/Stored Procedures/OrganizationUser_ReadByPendingAutoConfirm.sql
eliykat: Good catch! This is also the cause of the test failure, so good to know our tests are working :D

eliykat (commented):
My changes have been addressed, I can approve once @mkincaid-bw is happy and CI checks are passing.

test/Infrastructure.IntegrationTest/AdminConsole/Repositories/OrganizationUserRepository/ConfirmManyOrganizationUsersTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using Bit.Core.Enums;
+using Bit.Core.Repositories;
+using Xunit;
+
+namespace Bit.Infrastructure.IntegrationTest.AdminConsole.Repositories.OrganizationUserRepository;
+
+public class ConfirmManyOrganizationUsersTests
+{
+    [Theory, DatabaseData]
+    public async Task ConfirmManyOrganizationUsersAsync_MixedBatch_ReturnsOnlyAcceptedIds(
```
r-tome, line 11: Also assert that `RevisionDate` is updated. The simplest way is to capture `var before = DateTime.UtcNow` before the call and assert `updated.RevisionDate >= before` afterwards.

A more thorough alternative (which I'd lean toward but not a blocker!) is to pass `RevisionDate` in as a parameter to the repository method, as we do for example in `GroupRepository.DeleteUserAsync`. This lets the caller (the command/service) generate the timestamp via `TimeProvider`, the repository becomes a pure persistence layer, and the test can assert exact equality. Example: https://github.com/bitwarden/server/blob/d903096a77e386ed89cda74e7c8779358862be30/test/Infrastructure.IntegrationTest/AdminConsole/Repositories/GroupRepositoryTests.cs#L258

util/Migrator/DbScripts/2026-05-27_00_AddOrganizationUserUpdateStatusKey.sql
r-tome, line 1: Bump the date on the migration file names

src/Sql/dbo/Stored Procedures/OrganizationUser_ConfirmByIds.sql
```diff
@@ -0,0 +1,43 @@
...
+    OUTPUT
+        INSERTED.[Id] INTO @UpdatedIds
+    FROM
+        [dbo].[OrganizationUser] OU
+    INNER JOIN
+        @UsersToUpdate U ON U.[Id] = OU.[Id]
+    WHERE
+        OU.[Status] = 1 -- Accepted
+
+    EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationUserIds] @UpdatedIds
```
r-tome: The singular update sproc checks if any rows are updated before calling this. Might be worth it to be consistent and add that here too.


```suggestion
SET @RowCount = @@ROWCOUNT;
    IF @RowCount > 0
        BEGIN
            EXEC [dbo].[User_BumpAccountRevisionDateByOrganizationUserIds] @UpdatedIds
        END
```

r-tome (dismissed):

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateStatusKey.sql
```diff
@@ -0,0 +1,49 @@
...
+    @RevisionDate DATETIME2(7)
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @RowCount INT
+
+    DECLARE @UsersToUpdate AS TABLE (
+        [Id]  UNIQUEIDENTIFIER NOT NULL,
+        [Key] NVARCHAR(MAX)    NULL
```
mkincaid-bw: The `Key` column in the database is `varchar(max)`, not `nvarchar(max)`. The data types should match.
```suggestion
        [Key] VARCHAR(MAX)     NULL
```

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateStatusKey.sql
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    INSERT INTO @UsersToUpdate
+    SELECT
+        [Id],
+        [Key]
+    FROM OPENJSON(@UsersJson)
+    WITH (
+        [Id]  UNIQUEIDENTIFIER '$.Id',
+        [Key] NVARCHAR(MAX)    '$.Key'
```
mkincaid-bw: Same as above.
```suggestion
        [Key] VARCHAR(MAX)     '$.Key'
```

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateStatusKey.sql
```diff
@@ -0,0 +1,49 @@
+CREATE PROCEDURE [dbo].[OrganizationUser_UpdateStatusKey]
```
mkincaid-bw: I just realized that this is updating multiple records, so it should be named `OrganizationUser_UpdateManyStatusKey`

mkincaid-bw (changes requested):
Couple more minor changes.

util/Migrator/DbScripts/2026-05-22_00_AddOrganizationUserUpdateStatusKey.sql
```diff
@@ -0,0 +1,50 @@
...
+    @RevisionDate DATETIME2(7)
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @RowCount INT
+
+    DECLARE @UsersToUpdate AS TABLE (
+        [Id]  UNIQUEIDENTIFIER NOT NULL,
+        [Key] NVARCHAR(MAX)    NULL
```
mkincaid-bw:
```suggestion
        [Key] VARCHAR(MAX)     NULL
```

util/Migrator/DbScripts/2026-05-22_00_AddOrganizationUserUpdateStatusKey.sql
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    INSERT INTO @UsersToUpdate
+    SELECT
+        [Id],
+        [Key]
+    FROM OPENJSON(@UsersJson)
+    WITH (
+        [Id]  UNIQUEIDENTIFIER '$.Id',
+        [Key] NVARCHAR(MAX)    '$.Key'
```
mkincaid-bw:
```suggestion
        [Key] VARCHAR(MAX)     '$.Key'
```

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateManyStatusKey.sql
```diff
@@ -0,0 +1,49 @@
...
+    @RevisionDate DATETIME2(7)
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @RowCount INT
+
+    DECLARE @UsersToUpdate AS TABLE (
+        [Id]  UNIQUEIDENTIFIER NOT NULL,
+        [Key] VARCHAR(MAX)    NULL
```
mkincaid-bw: ⛏️
```suggestion
        [Key] VARCHAR(MAX)     NULL
```

src/Sql/dbo/Stored Procedures/OrganizationUser_UpdateManyStatusKey.sql
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    INSERT INTO @UsersToUpdate
+    SELECT
+        [Id],
+        [Key]
+    FROM OPENJSON(@UsersJson)
+    WITH (
+        [Id]  UNIQUEIDENTIFIER '$.Id',
+        [Key] VARCHAR(MAX)    '$.Key'
```
mkincaid-bw: ⛏️
```suggestion
        [Key] VARCHAR(MAX)     '$.Key'
```

mkincaid-bw (changes requested):
Sorry, I should have flagged the data type changes in the migration script as well. Also have some really minor, petty spacing nit picks 😄

eliykat:
Please re-request review when feedback is addressed and CI is passing, thanks. If you need any help just let me know.

mkincaid-bw (dismissed):
LGTM

eliykat (approved):

eliykat (dismissed):
Migration dates need bumping again 😢

sven-bitwarden (approved):

mkincaid-bw (approved):

PR #7666, opened 2026-05-18
[PM-37593] Add OrganizationUserStatusTypeNew -  🪓 Revoked

sven-bitwarden:
## 🎟️ Tracking
[PM-37593](https://bitwarden.atlassian.net/browse/PM-37593)

## 📔 Objective
This PR implements [part of Phase 1](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2571632646/ACII+Separate+revoked+from+status) of the initiative to separate Revoked from OrganizationUserStatusType.

The overall initiative will have us creating a new home for OrganizationUserStatusType (sans Revoked), populating it for revokes and _checking it first_ for restores, moving all organization user status type checks to account for revoked now being a separate field, and then cleaning up.

[PM-37593]: https://bitwarden.atlassian.net/browse/PM-37593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/AdminConsole/Entities/OrganizationUser.cs
+++ b/src/Core/AdminConsole/Entities/OrganizationUser.cs
@@ -48,6 +48,8 @@ public class OrganizationUser : ITableObject<Guid>, IExternal, IOrganizationUser
     public string? ResetPasswordKey { get; set; }
     /// <inheritdoc cref="OrganizationUserStatusType"/>
     public OrganizationUserStatusType Status { get; set; }
+    /// <inheritdoc cref="OrganizationUserStatusTypeNew"/>
+    public OrganizationUserStatusTypeNew? StatusNew { get; set; }
     /// <summary>
     /// The User's role in the Organization.
     /// </summary>
--- a/src/Core/AdminConsole/Enums/OrganizationUserStatusTypeNew.cs
+++ b/src/Core/AdminConsole/Enums/OrganizationUserStatusTypeNew.cs
@@ -1,0 +1,17 @@
+﻿using Bit.Core.Entities;
+
+namespace Bit.Core.Enums;
+
+/// <summary>
+/// Represents the different stages of a member's lifecycle in an organization.
+/// The <see cref="OrganizationUser"/> object is populated differently depending on their Status.
+/// </summary>
+/// <remarks>
+/// This is effectively a v2 version of OrganizationUserStatusType that severs Revoked as a status type.
+/// </remarks>
+public enum OrganizationUserStatusTypeNew : short
+{
+    Invited = 0,
+    Accepted = 1,
+    Confirmed = 2,
+}
--- a/src/Sql/dbo/Tables/OrganizationUser.sql
+++ b/src/Sql/dbo/Tables/OrganizationUser.sql
@@ -6,7 +6,6 @@
     [Key]                           VARCHAR (MAX)       NULL,
     [ResetPasswordKey]              VARCHAR (MAX)       NULL,
     [Status]                        SMALLINT            NOT NULL,
-    [StatusNew]                     SMALLINT            NULL,
     [Type]                          TINYINT             NOT NULL,
     [ExternalId]                    NVARCHAR (300)      NULL,
     [CreationDate]                  DATETIME2 (7)       NOT NULL,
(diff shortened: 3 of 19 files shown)


src/Sql/dbo/Tables/OrganizationUser.sql
```diff
@@ -6,6 +6,7 @@
     [Key]                           VARCHAR (MAX)       NULL,
     [ResetPasswordKey]              VARCHAR (MAX)       NULL,
     [Status]                        SMALLINT            NOT NULL,
+    [StatusNew]                     SMALLINT            NULL,
```
mkincaid-bw: New columns should be added at the end after the last column in the table.

util/Migrator/DbScripts/2026-05-18_00_AddStatusNewToOrganizationUser.sql
```diff
@@ -0,0 +1,587 @@
+-- Add StatusNew column to OrganizationUser
+IF COL_LENGTH('[dbo].[OrganizationUser]', 'StatusNew') IS NULL
+BEGIN
+    ALTER TABLE [dbo].[OrganizationUser]
+        ADD [StatusNew] SMALLINT NULL;
+END
+GO
+
+-- Refresh the core OrganizationUser view (SELECT * over the table)
+EXEC sp_refreshview N'[dbo].[OrganizationUserView]';
```
mkincaid-bw, line 10: Any view referencing a modified table should be refreshed as well.
```suggestion
EXEC sp_refreshview N'[dbo].[OrganizationUserView]';
EXEC sp_refreshview N'[dbo].[OrganizationUserOrganizationDetailsView]';
EXEC sp_refreshview N'[dbo].[OrganizationUserUserDetailsView]';
EXEC sp_refreshview N'[dbo].[ProviderOrganizationOrganizationDetailsView]';
EXEC sp_refreshview N'[dbo].[UserPremiumAccessView]';
```

mkincaid-bw (changes requested):
Couple  minor changes

JaredScar (dismissed):
Admin Console changes seem fine (only 2 files)

sven-bitwarden:
@mkincaid-bw updated - thanks for the reminder especially on the views ✌️

mkincaid-bw (approved):
LGTM

src/Core/AdminConsole/Enums/OrganizationUserStatusTypeNew.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Bit.Core.Entities;
+
+namespace Bit.Core.Enums;
```
eliykat, line 3: This should be in `Bit.Core.AdminConsole.Enums`. The only reason our current enum isn't is that moving it would trigger code owner review bingo.

This is non-blocking and can be addressed in a later PR.

eliykat (approved):
Comment is non blocking

src/Core/AdminConsole/Entities/OrganizationUser.cs
```diff
@@ -48,6 +48,8 @@ public class OrganizationUser : ITableObject<Guid>, IExternal, IOrganizationUser
     public string? ResetPasswordKey { get; set; }
     /// <inheritdoc cref="OrganizationUserStatusType"/>
     public OrganizationUserStatusType Status { get; set; }
+    /// <inheritdoc cref="OrganizationUserStatusTypeNew"/>
```
eliykat, line 51: This could do with some more explicit documentation at this stage. e.g.

> This is not fully in use yet and should not be used outside the restore/revoke flows.
> It is only used to back up the Status before revoking a user, and restore
> the user to the correct status later. It should be null if the user is not revoked.

(also non-blocking)

JaredScar (approved):

src/Core/AdminConsole/Entities/OrganizationUser.cs
sven-bitwarden: I'll tack it on to my upcoming PR 💅

PR #7667, opened 2026-05-18
[PM-37727] Fix exceptions when serialzing blob-encrypted ciphers

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37727

## 📔 Objective

When saving a blob-encrypted cipher from the client, the Server throws errors because it is expecting the JSON data for parsing. This adds a function call to see if a cipher is blob encrypted, and skip the ciphertype extractions if so.

### DIFF
--- a/src/Core/Vault/Entities/Cipher.cs
+++ b/src/Core/Vault/Entities/Cipher.cs
@@ -32,6 +32,17 @@ public class Cipher : ITableObject<Guid>, ICloneable
         Id = CoreHelpers.GenerateComb();
     }
 
+    public bool IsDataBlobEncrypted()
+    {
+        if (string.IsNullOrWhiteSpace(Data))
+        {
+            return false;
+        }
+
+        var span = Data.AsSpan().TrimStart();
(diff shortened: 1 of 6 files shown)


src/Core/Vault/Entities/Cipher.cs
```diff
@@ -32,6 +32,17 @@ public void SetNewId()
...
 
+    public bool IsDataBlobEncrypted()
+    {
+        if (string.IsNullOrWhiteSpace(Data))
+        {
+            return false;
+        }
+
+        var span = Data.AsSpan().TrimStart();
+        return span.Length > 0 && span[0] != '{';
```
jaasen-livefront, line 43: ⛏️ You can totally ignore this but  `span.Length > 0` can never be false at this point, because `IsNullOrWhiteSpace` already eliminates empty/whitespace strings. We can remove that check:

`return span[0] != '{';`

jaasen-livefront (approved):

shane-melton (approved):

PR #7670, opened 2026-05-19
[PM-37740] Add missing container configuration to aspire

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37740
## 📔 Objective

Adds the remaining Bitwarden services and supporting containers to the Aspire AppHost so the full server stack (matching dev/docker-compose.yml) can be run via Aspire.

**Project resources added**
  - Events, EventsProcessor, Icons, Scim, Sso

**Container resources added**
  - redis (persistent lifetime, append-only volume) for distributed caching
  - idp (SimpleSAMLphp, explicit start) for local SSO testing, with authsources.php bind-mounted from WorkingDirectory

**Other changes**
- Wires Sso into the mail-catcher reference list alongside admin/identity/billing
- ConfigureServices now returns IReadOnlyDictionary<string, IResourceBuilder<ProjectResource>> instead of a fixed 5-tuple, so adding services no longer breaks the call site
- Extracts repeated name[:tag] image parsing into a GetImageParts helper
- Adds BasePort entries for the new services and Redis/Idp config blocks in appsettings.Development.json

### DIFF
--- a/AppHost/AppHost.cs
+++ b/AppHost/AppHost.cs
@@ -9,14 +9,16 @@ builder.ConfigureMigrations()
     .WaitForCompletion(secretsSetup);
 var azurite = builder.ConfigureAzurite();
 var mail = builder.ConfigureMailCatcher();
-var (_, api, billing, _, _) = builder.ConfigureServices(db, secretsSetup, mail, azurite);
+builder.ConfigureRedis();
+builder.ConfigureIdp();
+var services = builder.ConfigureServices(db, secretsSetup, mail, azurite);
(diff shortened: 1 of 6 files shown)


withinfocus (approved):
Good expansion to the full services set but @sbrown-livefront please render your review as well.

sbrown-livefront (approved):
Super clean! Thanks for the additions. @eliykat! I'm glad you added Redis and that readonlydict for services is clutch!

PR #7672, opened 2026-05-19
[PM-37486] Remove IPolicyService and associated dead code

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37486

## 📔 Objective

Migrate remaining `IPolicyService` callers to use `IPolicyRequirementQuery` and subsequently delete `IPolicyService.


util/Migrator/DbScripts/2026-05-19_00_DropUnusedPolicySprocs.sql
```diff
@@ -0,0 +1,9 @@
+-- Drop stored procedures that are no longer called anywhere.
+-- OrganizationUser_ReadByUserIdWithPolicyDetails was used by PolicyService (now deleted).
+-- PolicyDetails_ReadByUserId was never called anywhere.
+
+DROP PROCEDURE IF EXISTS [dbo].[OrganizationUser_ReadByUserIdWithPolicyDetails];
```
mkincaid-bw: It looks like the C# code that calls this proc is being removed in this PR. If the deployment requires a server code rollback, this `DROP` would cause issues since DB changes are not reverted. This `DROP` should be moved to another PR to be done at a later date once this server deployment is hardened.

r-tome: Thanks for calling this out! I have reverted the SQL changes in this branch. I will create a separate task to drop the stored procedures in a future release.

eliykat (dismissed):
🧹

JaredSnider-Bitwarden (approved):
These changes LGTM! 

The expected behavior is intended to be the same but since this isn't feature flagged, I would like to request the following flows be QA tested from an auth perspective:

1. **Login master-password policy** — log in as a member of an org with the Master Password policy configured; confirm the client picks up the correct rules on the direct login response. Repeat for a user in two orgs with conflicting settings — strictest values should win. 

2. `/accounts/verify-password` **reprompt** - As the same user, perform any master-password reprompt action (Export vault is easiest); confirm the client receives the same policy block. Also verify a user in no org can successfully hit this endpoint and gets a default/empty policy block (no 500).

3.  **Accept org invite under Single Org policy**

- User already `Confirmed` in Org A (Single Org policy ON) must be blocked from accepting Org B's invite.
- User only `Invited` (not yet `Accepted`) in Org A must still be able to accept Org B's invite.
- User `Revoked` in Org A must be able to accept Org B's invite.
- Provider-exempt user must be able to accept.

eliykat:
@JaredSnider-Bitwarden 

>  1. **Login master-password policy** — log in as a member of an org with the Master Password policy configured; confirm the client picks up the correct rules on the direct login response. Repeat for a user in two orgs with conflicting settings — strictest values should win.

I take it this is the change in `BaseRequestValidator`. It was calling `PolicyService.GetMasterPasswordPolicyForUserAsync`, but that already used `PolicyRequirementQuery.GetAsyncVNext<MasterPasswordPolicyRequirement>` under the hood. This PR is just removing the indirection; no change in the actual implementation code.

>  2. `/accounts/verify-password` **reprompt** - As the same user, perform any master-password reprompt action (Export vault is easiest); confirm the client receives the same policy block. Also verify a user in no org can successfully hit this endpoint and gets a default/empty policy block (no 500).

Same as above.

>  3. **Accept org invite under Single Org policy**

This has not been changed here. Maybe you're looking at the `InitPendingOrganizationValidator`? That is responsible for initializing a reseller organization created via the Bitwarden Portal. In this case, we are already using the `SingleOrganizationPolicyRequirement` in all the other accept/confirm flows, so we're comfortable with it here.

Let me know if I've missed or misunderstood anything.

JaredSnider-Bitwarden:
> @JaredSnider-Bitwarden
> 
> > 1. **Login master-password policy** — log in as a member of an org with the Master Password policy configured; confirm the client picks up the correct rules on the direct login response. Repeat for a user in two orgs with conflicting settings — strictest values should win.
> 
> I take it this is the change in `BaseRequestValidator`. It was calling `PolicyService.GetMasterPasswordPolicyForUserAsync`, but that already used `PolicyRequirementQuery.GetAsyncVNext<MasterPasswordPolicyRequirement>` under the hood. This PR is just removing the indirection; no change in the actual implementation code.
> 
> > 2. `/accounts/verify-password` **reprompt** - As the same user, perform any master-password reprompt action (Export vault is easiest); confirm the client receives the same policy block. Also verify a user in no org can successfully hit this endpoint and gets a default/empty policy block (no 500).
> 
> Same as above.
> 
> > 3. **Accept org invite under Single Org policy**
> 
> This has not been changed here. Maybe you're looking at the `InitPendingOrganizationValidator`? That is responsible for initializing a reseller organization created via the Bitwarden Portal. In this case, we are already using the `SingleOrganizationPolicyRequirement` in all the other accept/confirm flows, so we're comfortable with it here.
> 
> Let me know if I've missed or misunderstood anything.

@eliykat 
Oh, ok, I'm sorry that I missed that it was already using that under the hood. Request for smoke tests retracted.  Also, the last one was my mistake with vetting claude output - I apologize.

JimmyVo16 (approved):

PR #7678, opened 2026-05-19
[PM-37068] feat: Add business plan cohort branch to UpcomingInvoiceHandler

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37068

## 📔 Objective

Wires the caller side of the Teams/Enterprise 2020 → current price migration that PM-37064 (#7665) built. `AlignOrganizationSubscriptionConcernsAsync` becomes a `ProductTier` dispatcher (future-proof against new plan cadences); the new business branch runs eligibility guards (feature flag on, assignment exists and unscheduled, cohort active, `MigrationPathId` resolves to a known path, org `PlanType` matches the cohort's source plan) before invoking PM-37064's `ScheduleBusinessPriceIncrease` and sending a placeholder renewal email that PM-37070 will replace.

Additional changes:

- Expands `subscriptions.data.customer` on the customer load so PM-37064's scheduler can preserve customer-level discounts into Phase 2.
- Treats a cohort with no `MigrationPathId` as a valid churn-discount-only cohort (silent skip; the scheduler's defensive check is updated to the same convention).
- Treats an unknown/unregistered `MigrationPathId` as a data-corruption signal (Error log, fall through to standard email).
- Treats drift between org `PlanType` and the cohort's source plan as an operational anomaly (Warning log, fall through to standard email).
- Attaches `migration_cohort_id` and `migration_cohort_name` metadata to the Stripe subscription on schedule creation so billing operations can attribute migrations from the Stripe dashboard.
- Incidentally strips pre-existing `#region` markers in `UpcomingInvoiceHandlerTests.cs` per the repo's no-regions convention.

`SendBusinessRenewalEmailAsync` is an info-level placeholder until PM-37070 wires the real template.

### DIFF
--- a/src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
+++ b/src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
@@ -142,6 +142,25 @@ public class PriceIncreaseScheduler(
 
         await CreateAndConfigureScheduleAsync(subscription, phase2);
 
+        try
+        {
+            await stripeAdapter.UpdateSubscriptionAsync(subscription.Id,
+                new SubscriptionUpdateOptions
+                {
+                    Metadata = new Dictionary<string, string>
+                    {
+                        [MetadataKeys.MigrationCohortId] = cohort.Id.ToString(),
+                        [MetadataKeys.MigrationCohortName] = cohort.Name
+                    }
+                });
+        }
+        catch (Exception ex)
+        {
+            logger.LogError(ex,
+                "Failed to attach cohort metadata to subscription ({SubscriptionId}) for cohort ({CohortId}); migration is scheduled but Stripe dashboard attribution will be missing.",
+                subscription.Id, cohort.Id);
+        }
+
         var assignment = await assignmentRepository.GetByOrganizationIdAsync(organizationId);
         if (assignment is null)
         {
--- a/src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
+++ b/src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
@@ -467,9 +467,7 @@ public class PriceIncreaseScheduler(
 
         if (cohort.MigrationPathId is null)
         {
-            logger.LogWarning(
-                "Cohort ({CohortId}) has no MigrationPathId; cannot schedule business price increase for subscription ({SubscriptionId})",
-                cohort.Id, subscription.Id);
+            // Churn-only cohort — no migration to schedule.
             return null;
         }
 
--- a/test/Billing.Test/Services/UpcomingInvoiceHandlerTests.cs
+++ b/test/Billing.Test/Services/UpcomingInvoiceHandlerTests.cs
@@ -3360,9 +3360,48 @@ public class UpcomingInvoiceHandlerTests
     }
 
     [Fact]
+    public async Task HandleAsync_WhenBusinessTier_AndFeatureFlagOff_FallsThroughToStandardEmail()
+    {
+        // Arrange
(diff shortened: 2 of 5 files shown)


src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
```diff
@@ -142,6 +142,25 @@ public async Task<bool> ScheduleBusinessPriceIncrease(
 
         await CreateAndConfigureScheduleAsync(subscription, phase2);
 
+        try
```
sbrown-livefront, line 145: ❓ Should we place this higher in the function since we already have the schedule set at this point? I can't recall if metadata changes could muck up schedule phases.

test/Billing.Test/Services/UpcomingInvoiceHandlerTests.cs
```diff
@@ -3362,6 +3359,668 @@ await _stripeAdapter.DidNotReceive().UpdateCustomerAsync(
...
+        await _mailService.Received(1).SendInvoiceUpcoming(
+            Arg.Is<IEnumerable<string>>(emails => emails.Contains("org@example.com")),
+            Arg.Any<decimal>(),
+            Arg.Any<DateTime>(),
+            Arg.Any<List<string>>(),
+            Arg.Is<bool>(b => b));
+    }
+
+    [Fact]
+    public async Task HandleAsync_WhenBusinessTier_AndCohortHasNoMigrationPath_FallsThroughSilently()
```
sbrown-livefront, line 3540: ❓ Are we going to track these at all for metrics or possible retries with a correct migration path in the future?

sbrown-livefront (commented):
Looks good. Quick questions

src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
amorask-bitwarden: Good question, but I think it's correct where it is. 2 reasons:
1. `subscription.metadata` is one of the few properties that, as far as I can tell from the documentation, does not have to be schedule aware. From [this page](https://docs.stripe.com/billing/subscriptions/subscription-schedules#phase-metadata): "Updating the underlying subscription’s metadata directly doesn’t affect the current phase’s metadata." You **can** use phase `metadata` to schedule transitions in metadata, but the parent subscription's metadata is not affected by those.
2. I think moving it prior to schedule creation might give us a worse failure mode. These pieces of metadata are going to be used for progress reporting. If we get into a scenario where metadata update succeeds and schedule creation fails, we'll be tracking a pending migration that's never going to take place. Additionally, the  reverse failure path (schedule creation succeeds and metadata update fails) is less likely simply due to operational complexity.

Let me know what you think there though.

test/Billing.Test/Services/UpcomingInvoiceHandlerTests.cs
amorask-bitwarden: Not having a migration path is actually a supported approach, hence the lack of logging or tracking. The reason is that a Cohort without a Migration Path **must** have a Churn Discount set. This allows the operators to set up cohorts that will **not** undergo a price migration, but **will** be offered a churn discount at self-serve cancellation, which was one of the requirements.

src/Core/Billing/Pricing/PriceIncreaseScheduler.cs
sbrown-livefront: Makes sense!

sbrown-livefront (approved):

PR #7679, opened 2026-05-20
[PM-36563] Send access event logs

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36563

## 📔 Objective

This PR adds EventTypes, control flow, and test coverage to log Send access events. 

Companion `clients` PR: https://github.com/bitwarden/clients/pull/20740


itsadrago (dismissed):

src/Api/Tools/Controllers/SendsController.cs
```diff
@@ -121,6 +125,13 @@ public async Task<SendAccessResponseModel> Access(string id, [FromBody] SendAcce
             sendResponse.CreatorIdentifier = creator.Email;
         }
 
+        if (send.UserId.HasValue
+            && send.Type == SendType.Text
+            && _featureService.IsEnabled(FeatureFlagKeys.SendEventLogging))
```
Banrion: 🎨 Putting the feature flag check first in the if statement would be a minor benefit in terms of computation since if that flag is false, the other checks don't need to run at all! I see this in multiple places in the file and in other files, but definitely not blocking

Banrion (dismissed):

src/Api/Tools/Controllers/SendsController.cs
harr1424: This should now be adjusted to prioritize the feature flag check  ✅

Banrion (approved):

adudek-bw (approved):
LGTM as well!

PR #7681, opened 2026-05-20
[PM-36964] Add per-org migration cohort assignment to the Admin portal

cyprain-okeke:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-36964

## 📔 Objective

Wire the existing `OrganizationPlanMigrationCohort` / `OrganizationPlanMigrationCohortAssignment` tables into the Admin portal's org Edit page so billing operators can determine eligibility per-org.

Operators can now, on the org Edit page:

- See the org's current cohort assignment.
- Pick a different cohort, or unassign, and submit Save.
- Be blocked at the server when the assignment has already entered the migration pipeline (`MigratedDate`, `ScheduledDate`, or `ChurnDiscountAppliedDate` set). The `<select>` is also disabled client-side as a UX hint, with a hidden input round-tripping the existing cohort id so edits to unrelated fields on locked orgs still save cleanly.

To back the dropdown, this PR also adds `GetManyAsync()` to `IOrganizationPlanMigrationCohortRepository` (Dapper + EF Core) and a new `[dbo].[OrganizationPlanMigrationCohort_ReadMany]` stored procedure. Listing returns active + inactive cohorts ordered by name so operators can pre-stage assignments against an inactive cohort before activation.

## 📸 Screenshots

_TODO: capture the Edit page showing (a) unassigned org with the dropdown enabled, (b) assigned-but-pending org with the dropdown enabled, (c) locked org (Migrated / Scheduled / Churn) with the dropdown disabled and lock reason rendered._

Screenshot 2026-05-20 at 12 02 59 AM

https://github.com/user-attachments/assets/a92de621-5823-4ab0-abd6-2706ffaa71a4


.gitignore
```diff
@@ -244,3 +244,4 @@ bitwarden_license/src/Sso/Sso.zip
 # Claude Code
 CLAUDE.local.md
 .worktrees/
+*.lscache
```
amorask-bitwarden: ⛏ This can be removed - it's been merged to `main` already via another PR.

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -226,7 +234,11 @@ public async Task<IActionResult> Edit(Guid id)
 
         var plans = await _pricingClient.ListPlans();
 
-        return View(new OrganizationEditModel(
+        var migrationCohorts = await _organizationPlanMigrationCohortRepository.GetManyAsync();
```
amorask-bitwarden: ❌ Because the SPROC is Read All, if we use this list in the select dropdown, we'd allow the end user to select a cohort with a Migration Path that has a `From` plan type different from the Organization in question. We should guard against that by filtering those out here.

```csharp
var migrationCohorts = await _organizationPlanMigrationCohortRepository.GetManyAsync();
var visibleCohorts = migrationCohorts
    .Where(c => !c.MigrationPathId.HasValue
                || MigrationPaths.FromId(c.MigrationPathId.Value)?.FromPlan == organization.PlanType)
    .ToList();
```

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -315,12 +341,74 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
+
+                if (submittedCohortId.HasValue)
+                {
+                    cohortToAssign =
+                        await _organizationPlanMigrationCohortRepository.GetByIdAsync(submittedCohortId.Value);
+                    if (cohortToAssign == null)
+                    {
+                        TempData["Error"] = "The selected migration cohort no longer exists.";
+                        return RedirectToAction("Edit", new { id });
+                    }
```
amorask-bitwarden: ⛏ For defense in-depth, we'd also want to guard against the incorrect `PlanType` match here as well:

```csharp
if (cohortToAssign.MigrationPathId.HasValue)
{
    var path = MigrationPaths.FromId(cohortToAssign.MigrationPathId.Value);
    if (path == null || path.FromPlan != organization.PlanType)
    {
        TempData["Error"] = "The selected migration cohort is not compatible with this organization's plan.";
        return RedirectToAction("Edit", new { id });
    }
}
```

src/Admin/AdminConsole/Views/Shared/_OrganizationForm.cshtml
```diff
@@ -62,6 +62,40 @@
                 <label class="form-check-label" asp-for="Enabled"></label>
             </div>
         }
+
+        @if (Model.Organization != null && Model.AvailableMigrationCohorts != null)
+        {
+            var migrationCohortDisabled = !canEditPlan || Model.MigrationCohortLocked;
```
amorask-bitwarden, line 71: This was a miss in the requirements, but we'll want to gate the visibility of this dropdown on 2 things:
1. FF: `pm-35215-business-plan-price-migration`
2. The user being an Admin. You might need to create a new permission for this. I suggest checking with @kdenney on what permission he is using or creating for the cohort management table.

amorask-bitwarden (changes requested):
Nice work! Just a couple things we'll want to clean up before merging.

amorask-bitwarden (dismissed):
Great work.

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -315,12 +360,85 @@ public async Task<IActionResult> Edit(Guid id, OrganizationEditModel model)
...
+                                "The selected migration cohort is not compatible with this organization's plan.";
+                            return RedirectToAction("Edit", new { id });
+                        }
+                    }
+                }
+
+                shouldWriteCohortAssignment = true;
+            }
+        }
+
```
jrmccannon: This is adding a lot to an already large method. Could we extract this out to a private method that returns a tuple or object with the values you need in the post save.

Error messages could be returned as well and then the page return could still happen in the endpoint method still.

jrmccannon (changes requested):

src/Admin/AdminConsole/Controllers/OrganizationsController.cs
cyprain-okeke: Good call — extracted the cohort resolution into a private `ResolveMigrationCohortAssignmentChangeAsync` helper that returns a `MigrationCohortAssignmentChange` record (`AssignmentToReplace`, `CohortToAssign`, `ShouldWrite`, `ErrorMessage`). The endpoint now just checks `ErrorMessage` and keeps owning the page return via `RedirectToAction`. The Edit action shed ~40 lines and all existing tests still pass.

amorask-bitwarden (approved):

rkac-bw (approved):
lgtm

jrmccannon (approved):

PR #7683, opened 2026-05-20
[PM-34776/PM-37797] Add invite link email domain validation endpoint

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34776
https://bitwarden.atlassian.net/browse/PM-37797

## 📔 Objective

Add an anonymous `POST /organizations/invite-link/validate-email-domain` endpoint so clients can check whether an email belongs to an invite link's allowed domains before login/signup.

Introduces `InviteLinkDomainValidator` (will be reused by the accept command).

[Clients PR](https://github.com/bitwarden/clients/pull/20749)


src/Core/AdminConsole/Utilities/InviteLinkDomainValidator.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Bit.Core.AdminConsole.Utilities;
+
+public static class InviteLinkDomainValidator
+{
+    /// <summary>
+    /// Returns whether the given email's domain is contained in the list of allowed domains.
+    /// </summary>
+    public static bool IsEmailDomainAllowed(string? email, IEnumerable<string> allowedDomains)
+    {
+        if (!MailAddress.TryCreate(email, out var mailAddress))
```
jrmccannon: `EmailValidation.IsValidEmail` our preferred method of validating.  I would validate that way then use the `EmailValidation.GetDomain` method (which does the same thing but is a shared method).

jrmccannon (changes requested):

src/Core/AdminConsole/Utilities/InviteLinkDomainValidator.cs
r-tome: I did know that existed! Replaced

jrmccannon (approved):

JaredSnider-Bitwarden (approved):
Request and response models match discussion for auth to consume!

PR #7689, opened 2026-05-21
Document migration workflow in .claude/CLAUDE.md

Hinton:
## 📔 Objective

Adds a Migrations section covering the dual Dapper/EF tracks, paths to the SSDT project and DbScripts, links to the contributing-docs MSSQL migration and SQL code style guides, and the local commands for applying migrations and regenerating EF migrations across all three providers.


.claude/CLAUDE.md
```diff
@@ -66,6 +60,41 @@
 - OpenTelemetry for observability (ADR 0020)
 - Log to standard output (ADR 0021)
 
+## Migrations
```
SaintPatrck: ❓Is this information Claude _always_ needs, or only when dealing with schema changes? If the latter, can we convert this to a new Skill, or fold it into `writing-database-queries`?

SaintPatrck (changes requested):

.claude/CLAUDE.md
Hinton: Only with db schemes. Not sure if the skill was triggered or not but if it was it is in-sufficient. I observed. the following defects.

- Hand wrote MySQL migration because the docker wasn't running. This should be forbidden.
- Used the wrong T-SQL syntax.

SaintPatrck: Gotcha. Can we verify if it was invoked, and if the instructions need to be updated? If it wasn't invoked when you expected we should be able to adjust the skill's `description` or `when_to_use` fields to improve hits from natural language and inference.

Hinton: How do I do that? If you want to take ownership of bringing in this change that's fine too, I just don't want this happening again.

theMickster: I am of the opinon that the majority of the contributions below are important enough to add to the root `CLAUDE.md`. They are surgical and direct instructions for Claude. 

I have a couple small recommendations.

.claude/CLAUDE.md
```diff
@@ -66,6 +60,41 @@
...
+We maintain an SSDT project at `src/Sql/dbo/` (current state of the database) and hand-written dated migrations
+under `util/Migrator/DbScripts/`. Every schema change updates both. Apply locally with `pwsh dev/migrate.ps1`.
+
+See:
+
+- https://contributing.bitwarden.com/contributing/database-migrations/mssql — how to write migrations.
+- https://contributing.bitwarden.com/contributing/code-style/sql — SQL code style (file naming, `IF COL_LENGTH`
+  guards, `CREATE OR ALTER`, etc.).
+
+### EF
```
theMickster: 🎨 Minor suggestion to be 100% sure that we ground claude. 
```suggestion
### Entity Framework (EF)
```

.claude/CLAUDE.md
```diff
@@ -66,6 +60,41 @@
 - OpenTelemetry for observability (ADR 0020)
 - Log to standard output (ADR 0021)
 
+## Migrations
+
+We use two different database providers, Dapper for SQL Server and Entity Framework for SQLite, MySQL and Postgres. When
```
theMickster: 🎨 Minor suggestion to be 100% sure that we ground claude.
```suggestion
We use two different database providers, Dapper for SQL Server and Entity Framework (EF) for SQLite, MySQL and Postgres. When
```

.claude/CLAUDE.md
```diff
@@ -66,6 +60,41 @@
...
+a schema change lands it must be reflected in BOTH tracks (Dapper/MSSQL and EF). Skipping either silently breaks one of
+the four supported databases.
+
+### Dapper
+
+We maintain an SSDT project at `src/Sql/dbo/` (current state of the database) and hand-written dated migrations
+under `util/Migrator/DbScripts/`. Every schema change updates both. Apply locally with `pwsh dev/migrate.ps1`.
+
+See:
```
theMickster: ⚠️ Claude will likely never read these files the way that we have them listed here because it doesn't perform a WebFetch upon starting a coding session.  Instead, I would encourage you to instruct Claude to create our first [rules files](https://code.claude.com/docs/en/memory#organize-rules-with-claude/rules/) based upon these contributing docs.  
I think that these are direct, deterministic instructions that we can codify in those rules files. Then, we can reference them here or in other places with little nudges/hints (@SaintPatrck can give you some tips @Hinton where we did that in the iOS and android repos to nudge Claude).  These Contributing Docs rarely, if ever, change so we don't need to worry about documentation drift at this point IMO.

.claude/CLAUDE.md
```diff
@@ -66,6 +60,41 @@
...
+- https://contributing.bitwarden.com/contributing/database-migrations/mssql — how to write migrations.
+- https://contributing.bitwarden.com/contributing/code-style/sql — SQL code style (file naming, `IF COL_LENGTH`
+  guards, `CREATE OR ALTER`, etc.).
+
+### EF
+
+EF migrations under `util/{Postgres,MySql,Sqlite}Migrations/` are auto-generated from the EF entity classes in
+`src/Infrastructure.EntityFramework/`. Update the entity first, then generate.
+
+You MAY NEVER hand-write a migration `.cs` file or its `.Designer.cs` snapshot. Always:
```
theMickster: 🎨
```suggestion
**NEVER** hand-write a migration `.cs` file or its `.Designer.cs` snapshot. 
**Always:**
```

.claude/CLAUDE.md
Hinton: It should prompt a fetch when it want's to act on it though?

theMickster: No, that's not going to happen the way that you assume it will happen.  The file is a simple file based instruction file treated as plain text. 
Claude code will never invoke a read of those URLs or any URLs embedded in our `CLAUDE.md` files. Any URLs should be used as reference materials for humans to improve a `CLAUDE.md` or similar file. Any information that Claude should leverage during a session needs to be inside .md files as plain text -or- you must create skills that leverage WebFetch or and MCP server to retrieve said data. The later is an expensive process which is why it's best practice to create supplemental example or reference files that Claude Code will progressively load as needed or directed by a human.

.claude/CLAUDE.md
SaintPatrck: >I am of the opinon that the majority of the contributions below are important enough to add to the root CLAUDE.md. They are surgical and direct instructions for Claude.

"Changing the DB schema" is only needed in specific circumstances, so putting it in `CLAUDE.md` directly contradicts the `CONTRIBUTING.md` guidelines that were just added to all of our repos.[^1][^2][^3] 

A lot of this section is already covered in the existing db related skills. Regardless of where it ultimately lands, it should not be duplicated in multiple places. Either move the existing skill content to `CLAUDE.md` or move the net-new content, here, to those skills.

If Claude is failing to invoke those Skills there are other approaches we can take to address that problem, like expanding the `description`, adding `when_to_use`, or adding a reference + usage table with instructions for using repo-specific skills here in `CLAUDE.md`. The latter option has value beyond sessions scoped to this repo, but is not in scope for this PR, so I recommend it as a separate follow-up to address that regardless of the decision of where to put this new information.

[^1]: https://github.com/bitwarden/server/blob/main/.claude/CONTRIBUTING.md?plain=1#L21
[^2]: https://github.com/bitwarden/server/blob/main/.claude/CONTRIBUTING.md?plain=1#L34
[^3]: https://github.com/bitwarden/server/blob/main/.claude/CONTRIBUTING.md?plain=1#L39

.claude/CLAUDE.md
theMickster: Not sure where this landed, but so long as we all understand that Claude Code will pass-over these links without performing a WebFetch then I'm going to resolve this discussion.

.claude/CLAUDE.md
theMickster: Following-up from last week. Where did we land here? Are we going to attempt to move these items into a rules file(s), move and improve existing skills, move them into a `CLAUDE.md` lower in the repo closer to the actual migration scripts, or something else?  

A challenge that we need to work together to overcome is that Oscar clearly had a frustrating experience with Claude attempting to write a migration script and raised the issue as asked in our Claude tooling. I'm not sure that he's got the capacity to do all of the tweaking, testing, adjusting, etc. that goes into skill development at the moment to find the perfect long-term home for these instructions; he just needs Claude Code to help him out.  So what new/old/out-of-the-box ideas can we try to move this change through?

Thoughts @SaintPatrck 

> **Note** when I state 'we' up above I mean like Bitwarden engineering as a collective. Not one particular engineering team or group of people. I tend to think that way for better or for worse 😜

Hinton: I moved things out into a rule. I think a rule is more appropriate than a skill because it *should always* trigger when working in the paths, not sometimes.

theMickster (approved):

SaintPatrck (approved):
Thanks for moving this out of `CLAUDE.md`. 👍

PR #7690, opened 2026-05-21
[PM-37944] Differentiate Send events according to domain

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37944

## 📔 Objective

This PR introduces capability to differentiate when an email-verified Send is created and/or accessed based upon if the accessor email matches a domain claimed by a given organization. 

[companion clients PR](https://github.com/bitwarden/clients/pull/20787)


harr1424:
@claude please review

JaredSnider-Bitwarden (dismissed):
Auth file change is just an extra `SendAccessClaimsPrincipalExtension` for retrieving email from claims. LGTM!

util/Migrator/DbScripts/2026-07-10_00_AddEventSendId.sql
mkincaid-bw, line 1: I was just curious as to why these file names are dated more than a week into the future? Is it expected that this PR will not be merged until 6/26/2026?

src/Sql/dbo/Dirt/Tables/Event.sql
```diff
@@ -30,3 +31,8 @@ GO
 CREATE NONCLUSTERED INDEX [IX_Event_DateOrganizationIdUserId]
     ON [dbo].[Event]([Date] DESC, [OrganizationId] ASC, [ActingUserId] ASC, [CipherId] ASC) INCLUDE ([ServiceAccountId], [GrantedServiceAccountId]);
 
+GO
+CREATE NONCLUSTERED INDEX [IX_Event_OrganizationIdSendIdDate]
```
mkincaid-bw, line 35: 👍I just wanted to leave a note on this. We have some self-host customers with very large `Event` tables. I did a test  locally to simulate this and found that this index took about a minute to build. This is because the `SendId` is added as `NULL`, and this index is filtered where `SendId` is `NOT NULL`, so the `CREATE INDEX` essentially just needs to scan the `Event` table and create an empty index.

util/Migrator/DbScripts/2026-07-10_00_AddEventSendId.sql
harr1424: Hey @mkincaid-bw thanks for looking this over!  Considering how many reviewers are on this very large PR, and our current QA backlog, I think that 6/26/2026 is still extremely eager for a merge date. 

I expect I will need to edit those filenames in the future to have a date well into July, in order to satisfy the `Database testing / Validate new migration naming and order` workflow. 

My apologies if future dating these is bad practice, but for a PR of this size I wanted to avoid any failing CI checks that would dissuade reviewers from beginning their review.

mkincaid-bw: Hey @harr1424,
That's fine. I know many of these larger PR's that hang out for a while often get kicked back due to the file names being out of date. My only concern would be overlapping PR's where another PR modifies the same object(s) that you are modifying. If that PR pulled in your changes as the most current, but then had an earlier migration date and went out during the same deployment as yours, your migrations would overwrite theirs, which would cause issues.

It's not a very common scenario and since dbops has visibility into it, we can likely catch it before it's an issue. I just wanted to make sure it was intentional and with good reason. There is currently another PR outstanding that is also working on the Event table so we'll just need to keep an eye on when they are merged to make sure one PR does not overwrite the changes of another.

mkincaid-bw (dismissed):

src/Core/Tools/SendFeatures/Services/SendEventClassifier.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        var claimedDomainsByOrg = await LoadClaimedDomainsByOrgAsync(confirmedOrgIds);
+
+        // Resolve the accessor as a confirmed member via their linked user account. OrganizationUser.Email
+        // is only populated for pending invites, so a confirmed member can only be matched through the User
+        // record (email -> user -> confirmed memberships), never via OrganizationUser.Email.
+        Guid? accessorUserId = null;
+        var accessorMemberOrgIds = new HashSet<Guid>();
+        var accessorUser = await _userRepository.GetByEmailAsync(accessorEmail!);
```
Banrion: ⚠️ Local Claude Finding - Member match is provider-dependent: `accessorEmail` reaches `GetByEmailAsync` un-normalized. Comparison is case-sensitive on PostgreSQL and SQLite.

**Effect:** an accessor who is a genuine confirmed member, but whose claim email carries any uppercase (Alice@Example.com), is attributed as a member on MSSQL/MySQL yet falls through to claimed-domain or External on Postgres/SQLite — the audit log differs by database backend for the same access. The domain branch still works (it lowercased), so the symptom is "member silently demoted to domain/External" on two of the four supported providers.

**Recommendation:** normalize once at the top of `BuildAccessContextAsync` and use it for both the domain extraction and the user lookup

Banrion (commented):
✅ DIRT Files look good

⚠️ I did a local Claude review along with my DIRT file review, and it found one bug that I thought was important share

src/Core/Tools/SendFeatures/Services/SendEventClassifier.cs
harr1424: @Banrion thank you for catching this! I've just pushed a commit that normalizes the email upstream.

JaredSnider-Bitwarden: @harr1424 , cc: @Banrion :

After seeing the new email normalization land in the auth validator, I took a look with Claude to ensure I understood why we were normalizing there, and it and I came up with some questions (almost exclusively tools domain though):

The normalization at `SendEmailOtpRequestValidator.cs:46` does close the Postgres/SQLite symptom the previous review agent flagged, but only transitively — by the time `BuildAccessContextAsync` is called from `SendsController.cs:320`/`:358`, the email has already been lowercased upstream and stuffed into the JWT claim. A couple of questions before resolving:

1. **Defense in depth at the classifier?** The asymmetry the previous agent originally called out is still there: `ExtractDomain` lowercases at `SendEventClassifier.cs:122` but `GetByEmailAsync(accessorEmail!)` at `SendEventClassifier.cs:60` does not. The classifier now has an implicit "callers must hand me a normalized email" contract that isn't expressed anywhere. Any future caller wiring `accessorEmail` from a different source (a different claim, a header, an internal call) silently reintroduces the provider-dependent bug. Worth normalizing once at the top of `BuildAccessContextAsync` regardless?

2. **Should `Send.Emails` be normalized at write time?** Today nothing on the ingest side lowercases — `SendValidationService.ValidateUserCanSaveAsync` only checks length/prefix, and `SendAuthenticationQuery.EmailOtp` only trims on the read side. The allow-list comparison only works because of `StringComparer.OrdinalIgnoreCase`. If Tools normalized each entry to `Trim().ToLowerInvariant()` in the create/update command (before Data Protection wraps it), every downstream consumer becomes obviously correct on every DB engine without needing to remember the case-folding dance. The Auth-side normalization at L46 is still needed (untrusted input + the OTP cache key at `SendEmailOtpRequestValidator.cs:55` is built via case-sensitive `string.Format`), so this isn't either/or — it's both.

harr1424: @JaredSnider-Bitwarden both of these findings should be addressed by my latest commit.

JaredSnider-Bitwarden: It looks good to me! I'll let @Banrion resolve the comment if they feel the same.

JaredSnider-Bitwarden (dismissed):

JaredSnider-Bitwarden (dismissed):

Banrion (dismissed):

mkincaid-bw (dismissed):

AlexRubik (dismissed):

mcamirault (dismissed):
LGTM

mkincaid-bw (approved):

itsadrago (approved):

ike-kottlowski (approved):

Banrion (approved):

PR #7693, opened 2026-05-22
[PM-37595] Use New OrganizationUserStatusType Columns

sven-bitwarden:
## 🎟️ Tracking
[PM-37595](https://bitwarden.atlassian.net/browse/PM-37595)

## 📔 Objective
This PR implements [phase 2](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2571632646/ACII+Separate+revoked+from+status) of separating revoked from `OrganizationUserStatusType`.

A prior PR added the column to support the work. This PR now begins _lightly_ using the new column. 
When a user is revoked, we will copy the status they were being revoked **from** into the column. In essence, this acts like a "prior status" placeholder.
If/when the user is subsequently restored, _and_ the user has a value present for the new status type, we will restore them from the placeholder. If no value is present (they were revoked before we started tracking this dual-column system), we fallback to existing logic.

[PM-37595]: https://bitwarden.atlassian.net/browse/PM-37595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
```diff
@@ -8,12 +8,15 @@ BEGIN
...
     SET
+        OU.[StatusNew] = OU.[Status], -- Snapshot the pre-update Status before overwriting
         OU.[Status] = -1, -- Revoked
         OU.[RevocationReason] = @RevocationReason
     FROM
         [dbo].[OrganizationUser] OU
     INNER JOIN
         @OrganizationUserIds OUI ON OUI.[Id] = OU.[Id]
+    WHERE
+        OU.[Status] != -1 -- Skip already-revoked rows so existing StatusNew / RevocationReason are preserved
```
eliykat: Nice potential catch

src/Core/AdminConsole/Entities/OrganizationUser.cs
```diff
@@ -99,6 +110,35 @@ public static bool IsValidResetPasswordKey(string? resetPasswordKey)
...
+    /// the prior status from the row's property arrangement, for rows revoked before that snapshot
+    /// was being tracked.
+    /// </summary>
+    public OrganizationUserStatusType GetPriorActiveOrganizationUserStatusType()
+    {
+        // OrganizationUserStatusTypeNew has no Revoked variant, so any populated value is valid.
+        if (StatusNew.HasValue)
+        {
+            return (OrganizationUserStatusType)(short)StatusNew.Value;
+        }
```
eliykat: While your comment is true, and all things going well we should never end up with an invalid value, I am leaning towards being more defensive about it.

The cast does not validate that the enum is in a valid range: https://dotnetfiddle.net/KeF7Ke. So if we do end up with a bad value, it propagates through the system.

We could have an explicit conversion method instead:

```c#
public static class OrganizationUserStatusTypeNewExtensions 
  {                                                 
      public static OrganizationUserStatusType ToOrganizationUserStatusType(this OrganizationUserStatusTypeNew status) 
          => status switch       
          {  
              OrganizationUserStatusTypeNew.Invited => OrganizationUserStatusType.Invited,
              OrganizationUserStatusTypeNew.Accepted => OrganizationUserStatusType.Accepted, 
              OrganizationUserStatusTypeNew.Confirmed => OrganizationUserStatusType.Confirmed, 
              _ => // TODO: how to handle this? Throw, or return null for the caller to handle?
          };      
  }  
```

Open for discussion.

eliykat (commented):
Really tidy and nicely done. Just one thought below.

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
sven-bitwarden: You're too generous, the idea was yours: [PM-37595](https://bitwarden.atlassian.net/browse/PM-37595) ❤️ 

> Scope
> 
> In
> 
> MSSQL — OrganizationUser_RevokeMany: in the current UPDATE statement, copy the row's pre-update Status into StatusNew. Defensively skip rows already in the Revoked state so any existing StatusNew value is preserved

src/Core/AdminConsole/Entities/OrganizationUser.cs
sven-bitwarden: "How to handle" - imo throw 100%. This should be a "your data is corrupted" huge edge case. If we let this value return null, it becomes up to every caller.

This code's lifespan is also until we converge onto one OrganizationUserStatusType, so throwing/having less to clean up is even nicer here imo.

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
eliykat: Well, it's good to be a fan of your own work 😁

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
```diff
@@ -929,10 +929,14 @@ public async Task RevokeManyAsync(IEnumerable<Guid> organizationUserIds, Revocat
...
-        await dbContext.OrganizationUsers.Where(x => organizationUserIds.Contains(x.Id))
+        // Ordering of property updates here is important, see: https://github.com/dotnet/efcore/issues/35361
+        // Potentially fixed in EF 10, and covered by integration tests
+        await dbContext.OrganizationUsers
+            .Where(x => organizationUserIds.Contains(x.Id) && x.Status != OrganizationUserStatusType.Revoked)
             .ExecuteUpdateAsync(s => s
+                .SetProperty(x => x.RevocationReason, reason)
                 .SetProperty(x => x.Status, OrganizationUserStatusType.Revoked)
-                .SetProperty(x => x.RevocationReason, reason));
+                .SetProperty(x => x.StatusNew, x => (OrganizationUserStatusTypeNew?)(short)x.Status));
```
eliykat, line 905: Let's do the same thing here (in the opposite direction) for consistency.

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
eliykat, line 1: CI failures noted in this file

src/Core/AdminConsole/Enums/OrganizationUserStatusTypeNewExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+public static class OrganizationUserStatusTypeNewExtensions
+{
+    public static OrganizationUserStatusType ToOrganizationUserStatusType(this OrganizationUserStatusTypeNew status)
+        => status switch
+        {
+            OrganizationUserStatusTypeNew.Invited => OrganizationUserStatusType.Invited,
+            OrganizationUserStatusTypeNew.Accepted => OrganizationUserStatusType.Accepted,
+            OrganizationUserStatusTypeNew.Confirmed => OrganizationUserStatusType.Confirmed,
+            _ => throw new ArgumentOutOfRangeException(nameof(status), status,
+                "Unknown OrganizationUserStatusTypeNew value."),
```
eliykat, line 14: Re: how to handle this situation. This is not feature flagged, the risk of throwing is that we have unknown invalid data today, and this surfaces it in a way that breaks the app in a very noisy way for users. (i.e. it prevents them from carrying out this action.) That likely leads to CS tickets and a hotfix, or manual dbops intervention to fix the data, which we want to avoid.

However, as you say it does give us stronger guarantees around our data. If you want to go this route, I suggest getting dbops to run a quick query to ensure we have no out of range enum values. If that's good in USPRD, chances are we're good with this approach.

The alternative would be to skip the new logic, but log the failure in a way that we can observe via Datadog and take remedial action. It doesn't throw the brakes on but may allow undefined behavior (although presumably undefined behavior that exists today).

I can go either way on it, just worth discussing. I do think your approach is OK provided we have a quick peek at our current data.

eliykat (changes requested):

src/Core/AdminConsole/Enums/OrganizationUserStatusTypeNewExtensions.cs
sven-bitwarden: I get what you mean. Personally, if I build a model and somehow data has escape what should be airtight control over that model that will cause failures down the road, I consider it undefined behavior and they should encounter errors very loudly..but also hear what you mean re: tickets.

sven-bitwarden: https://bitwarden.slack.com/archives/C065AUDER62/p1779810200481739

I've confirmed there are no enum values out of the valid range [-1, .., 2]

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
sven-bitwarden: Strange, I had to rebase onto main and then force push a fix. Everything worked locally 😖

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
sven-bitwarden: This is ran inside of EF's query translation layer - it's going to be a little harder to work with as an Expression, can't throw or do anything useful other than just map 1:1.

src/Sql/dbo/Stored Procedures/OrganizationUser_RestoreMany.sql
mkincaid-bw, line 1: Hey @sven-bitwarden 
I mistakenly approved the previous PR that created `OrganizationUser_RestoreMany` and `OrganizationUser_RevokeMany`, which are not named according to [our guidelines](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures). Stored procedures should be named in the `{EntityName}_{Action}` format, where the action is a CRUD Action (`Read`, `ReadMany`, `Create`, `Update`, `UpdateMany`, `Delete`, `DeleteMany`, etc). The action can then be followed by a short description of what is is being changed (if if it's not the entire record).

As such, the `RevokeMany` and `RestoreMany` should be changed to something like `OrganizationUser_UpdateManyRevoke` and `OrganizationUser_UpdateManyRestore`

If the `OrganizationUser_RestoreMany` and `OrganizationUser_RevokeMany` procs are already in use, we can switch the code to use the new ones, then drop these in a future release.

My apologies for not catching this in the initial PR.

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
```diff
@@ -8,12 +8,15 @@ BEGIN
     UPDATE
         OU
     SET
+        OU.[StatusNew] = OU.[Status], -- Snapshot the pre-update Status before overwriting
         OU.[Status] = -1, -- Revoked
```
mkincaid-bw, line 11: Values should be passed in as parameters
```suggestion
        OU.[Status] = @Status
```

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
mkincaid-bw: Values should be passed in as parameters
```suggestion
        OU.[Status] != @Status -- Skip already-revoked rows so existing StatusNew / RevocationReason are preserved
```

mkincaid-bw (changes requested):

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
sven-bitwarden: I can parameterize it - but this is a static enum value. We're not setting a variety of statuses, invoking the Revoke stored proc **only** sets their status to -1 (Revoked). Does it still make sense?

sven-bitwarden:
@mkincaid-bw I updated the migration _from_ modifying the existing sprocs _to_ creating new sprocs with the correct name + new behavior. Let me know if we still need to parameterize the revoked enum value.

src/Sql/dbo/Stored Procedures/OrganizationUser_RestoreMany.sql
sven-bitwarden: No problem, taken care of. I admit I didn't check guidelines, I just saw our old sprocs named `Activate` and `Deactivate` and directly replaced with restore/revoke

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
mkincaid-bw: @withinfocus Some of our stored procs have hard-coded values but most procs have them supplied as parameters passed in from the server code. Not sure what the differentiator would be so we could use some guidance here.

src/Core/AdminConsole/Enums/OrganizationUserStatusTypeNewExtensions.cs
eliykat: Thanks for checking!

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationUserRepository.cs
eliykat: Oh good to know - fine as-is then.

eliykat (dismissed):

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
eliykat: My 2 cents: these sprocs are carrying out domain specific behavior - whenever the `OrganizationUser.Status` is changed, there are other updates to the row that are required depending on the transition. `OrganizationUser` is essentially a state machine.

In this case, changing `Status` to revoked (revoking) or from revoked (restoring) requires a specific set of data updates that are not reusable in other state transitions. Making this a parameter only allows misuse because it only has 1 valid value in this context.

mkincaid-bw (dismissed):

src/Sql/dbo/Stored Procedures/OrganizationUser_RevokeMany.sql
mkincaid-bw: @eliykat Thanks for the input/clarification.

sven-bitwarden:
@mkincaid-bw @eliykat Apologies had a merge conflict 😞, please resend your finest reviews

mkincaid-bw (approved):

JaredScar (approved):

eliykat (approved):

PR #7707, opened 2026-05-25
[PM-37092] feat: Add business plan migration handler to SubscriptionUpdatedHandler

cyprain-okeke:
## 🎟️ Tracking

[PM-37092](https://bitwarden.atlassian.net/browse/PM-37092)

## 📔 Objective

Completes the business-plan price-migration loop by wiring a Phase-2 handler into `SubscriptionUpdatedHandler` that reacts to Stripe's schedule-triggered `customer.subscription.updated` webhook, applies the target plan shape to the organization, and marks the cohort assignment migrated.

**What this adds**

- `HandleScheduleTriggeredBusinessMigrationAsync` in `SubscriptionUpdatedHandler` (gated by `PM35215_BusinessPlanPriceMigration`). Layered short-circuits:
  1. `ScheduleId` present → otherwise not a schedule-triggered transition.
  2. `PreviousAttributes.Items.Data` present → otherwise no item-shape change.
  3. Previous items intersect the registered source-price allowlist (built from `MigrationPaths.All`) → otherwise an unrelated schedule update.
  4. Cohort assignment row exists for the org → otherwise the org isn't enrolled.
  5. `assignment.MigratedDate` unset → idempotency guard against Stripe webhook retries.
  6. Cohort + `MigrationPathId` resolve to a registered `MigrationPath` → otherwise paused or forward-compat row.
  7. Target plan's price ID is present in current `subscription.Items` → defensive sanity check that Stripe actually applied the expected price.
  8. Organization row exists → write `organization.ChangePlan(targetPlan)` then stamp `MigratedDate` on the assignment.
- New `Organization.ChangePlan(Plan)` extension that applies structural plan-shape fields only. `UpgradeOrganizationPlanCommand.UpgradePlanAsync` is refactored to call it (behavior-preserving: `UsersGetPremium || PremiumAccessAddon` override is re-applied at the call site).
- `MigrationPaths` registry adds the two Teams 2020 → current paths plus `MigrationPathId` enum entries (`Teams2020AnnualToCurrent = 3`, `Teams2020MonthlyToCurrent = 4`). `MigrationPathIdsSnapshotTests` is updated to lock the new byte values and FromPlan/ToPlan pairs.
- Error handling mirrors the existing Families handler in the same file: `BillingException` from `_pricingClient.GetPlanOrThrow` rethrows so Stripe retries; any other exception is logged and absorbed. All structured logs include `OrganizationId` only — no PII, no plan secrets.

**What this does not do**

- Does not touch the Families migration handler.
- Does not change the catch-all policy or add a transactional wrapper around the two `ReplaceAsync` calls — those would diverge from the Families handler pattern in the same file and require coordinated changes outside this branch.

## Test plan

- [x] `SubscriptionUpdatedHandlerTests` — 17 business-migration tests, including the 4 happy-path matrices (EA/EM/TA/TM 2020 → current), every gating short-circuit, idempotency, the target-price sanity check, the `BillingException` rethrow, and the generic-Exception swallow.
- [x] `OrganizationPlanExtensionsTests` — 3 `UseKeyConnector` arms plus a customer-purchase-column non-mutation guard that enforces the doc contract.
- [x] `MigrationPathIdsSnapshotTests` — byte values, registry IDs, FromPlan/ToPlan pairs, count, and uniqueness all locked.
- [x] `dotnet build src/Billing/Billing.csproj` — 0 errors.
- [x] Full `SubscriptionUpdatedHandlerTests` class — 59/59 passing.
- [x] `OrganizationPlanExtensionsTests` + `MigrationPathIdsSnapshotTests` — 10/10 passing.

[PM-37092]: https://bitwarden.atlassian.net/browse/PM-37092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Billing/Organizations/Extensions/OrganizationPlanExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    ///
+    /// Customer-purchase fields intentionally NOT touched (owned by callers):
+    ///   Seats, MaxStorageGb, SmSeats, SmServiceAccounts, MaxAutoscale*,
+    ///   UseSecretsManager, BusinessName, Enabled.
+    ///
+    /// Forward-application helper: capability flags are projected directly from
+    /// <paramref name="newPlan"/>, so calling with a lesser plan will silently
+    /// drop capabilities. Use only for known-forward plan transitions.
+    /// </summary>
+    public static void ChangePlan(this Organization organization, Plan newPlan)
```
amorask-bitwarden: ⛏ If we don't want this method to be used for downgrades, we should protect it with code, not comments. Otherwise, we can remove this section of the summary.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -525,4 +536,151 @@ private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
...
+                return;
+            }
+
+            if (subscription.ScheduleId == null)
+            {
+                return;
+            }
+
+            var previousSubscription = parsedEvent.Data.PreviousAttributes?.ToObject<Subscription>() as Subscription;
+            if (previousSubscription?.Items?.Data == null)
```
amorask-bitwarden, line 557: ⛏ This case is worth a LogWarning.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -525,4 +536,151 @@ private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
...
+            {
+                return;
+            }
+
+            var assignment = await _cohortAssignmentRepository.GetByOrganizationIdAsync(organizationId);
+            if (assignment == null)
+            {
+                _logger.LogWarning(
```
amorask-bitwarden: ⛏ This log will fire for every organization that isn't in a migration cohort, which would be the majority of them to start. I don't believe it's a warning state. It just means the organization isn't in a cohort yet.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -525,4 +536,151 @@ private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
...
+                    "Schedule-triggered business migration fired for organization ({OrganizationId}) but no cohort assignment row was found",
+                    organizationId);
+                return;
+            }
+
+            if (assignment.MigratedDate.HasValue)
+            {
+                _logger.LogInformation(
+                    "Schedule-triggered business migration already applied for organization ({OrganizationId}); skipping (likely Stripe webhook retry)",
```
amorask-bitwarden: ⛏ I don't think "likely Stripe webhook retry" is true. A `customer.subscription.updated` fires for a wide variety of reasons.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -525,4 +536,151 @@ private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
...
+                    cohort.Id,
+                    cohort.MigrationPathId.Value);
+                return;
+            }
+
+            var sourcePlan = await _pricingClient.GetPlanOrThrow(migrationPath.FromPlan);
+            var sourcePriceId = GetPasswordManagerPriceId(sourcePlan);
+            if (string.IsNullOrEmpty(sourcePriceId) || !previousSubscription.Items.Data.Any(item =>
+                    item.Price?.Id != null && item.Price.Id == sourcePriceId))
+            {
```
amorask-bitwarden, line 622: ⛏ Seems like this situation would be worth a logged warning.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -525,4 +536,151 @@ private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
...
+                return;
+            }
+
+            var targetPlan = await _pricingClient.GetPlanOrThrow(migrationPath.ToPlan);
+            var targetPriceId = GetPasswordManagerPriceId(targetPlan);
+
+            if (!subscription.Items.Any(item => item.Price.Id == targetPriceId))
```
amorask-bitwarden: ❓ If we're checking for null price IDs on line 604, why are we not checking here? I think we should choose one route or the other.

src/Billing/Services/Implementations/SubscriptionUpdatedHandler.cs
```diff
@@ -525,4 +536,151 @@ private async Task HandleScheduleTriggeredFamiliesMigrationAsync(
...
+            {
+                _logger.LogWarning(
+                    "Schedule-triggered business migration for organization ({OrganizationId}): expected target price ({ExpectedPriceId}) for PlanType ({TargetPlanType}) not found in current subscription items; skipping",
+                    organizationId,
+                    targetPriceId,
+                    targetPlan.Type);
+                return;
+            }
+
+            var organization = await _organizationRepository.GetByIdAsync(organizationId);
```
amorask-bitwarden: ⛏ This should probably be one of the first checks we make in this method.

amorask-bitwarden (commented):
Looking good - just a few picks if you don't mind.

amorask-bitwarden (approved):

PR #7713, opened 2026-05-25
[PM-35314] Backfill Revocation Reasons to All Users

sven-bitwarden:
## 🎟️ Tracking
[PM-35314](https://bitwarden.atlassian.net/browse/PM-35314)

## 📔 Objective
The revocation reasons epic gave us the ability to track _why_ a user was revoked. For data consistency's sake, it would behoove us to ensure that _all_ revoked users have a revocation reason.

This PR adds a default reason (Unknown) to all prior users, ensuring our data is fully consistent. This paves the way for future work to separate Revoked from OrganizationUserStatusType, ensuring that when we do so, we don't lose track of revoked users.

[PM-35314]: https://bitwarden.atlassian.net/browse/PM-35314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rkac-bw:
I believe the direction from @withinfocus has been to not have business logic in the db layer.  My concern is that “what does a missing value mean?” is application semantics, and I’d rather keep that logic in server code instead of baking it into a migration.

In this case, the rule is already documented on the enum:

```
/// Revoked for an unknown reason, or migrated from before revocation reasons were tracked.
Unknown = 0,
```

That’s the business rule. I think the cleaner place to apply it is at the read boundary, not through a bulk data update.

For example:

```
public RevocationReason EffectiveRevocationReason =>
    RevocationReason ?? RevocationReason.Unknown;
```

That keeps the rule in code where it is easier to find, test, and change later. It also preserves the NULL signal in the database, so we can still tell the difference between a pre-feature row and a row that was explicitly set to Unknown if that ever matters.

It also avoids writing to a hot, heavily indexed table. That means no lock window, no RevisionDate churn, and no extra client resync amplification from User_BumpAccountRevisionDate.

I don’t think we lose anything by doing it this way. The enum already defines Unknown as the canonical value for rows migrated from before revocation reasons were tracked.

sven-bitwarden:
@rkac-bw This column is going to be come the single source of truth for whether or not a user is revoked - it having having a reason is a bonus. It's similar to a pattern of using `DateTime? DisabledDate` to convey two pieces of information: is a user disabled, and if so, when were they deactivated? [There's an ongoing initiative](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2571632646/ACII+Separate+revoked+from+status) to rip `Revoked` **out** of `OrganizationUserStatusType` - you might have already seen [some work that starts this](https://github.com/bitwarden/server/pull/7666).

withinfocus:
It's not necessarily clear in this isolated migration of course, but this is a larger body of work to transform what's being stored. With that in mind, I think this is acceptable and there will be further database changes to continue this work. Each case will be different but the implication of "logic in the database" is not a worry for me here.

sven-bitwarden:
@rkac-bw How can we proceed with this guy? I'm happy to iterate on the migration or do whatever I can to make it easier.

rkac-bw:
@sven-bitwarden Let me get a backup of prod going and test the migration, get a feel for how long it will take and then we can discuss migration logistics

rkac-bw:
The migration took 40 seconds with batch size 2000 and WAITFOR DELAY '00:00:00.100' added to increase concurrency this will work well

util/Migrator/DbScripts/2026-06-19_00_PopulateOrganizationUserRevocationReason.sql
```diff
@@ -0,0 +1,62 @@
...
+DECLARE @RowsAffected INT = 1;
+
+WHILE @RowsAffected > 0
+BEGIN
+    UPDATE TOP (@BatchSize) [dbo].[OrganizationUser]
+    SET [RevocationReason] = 0  -- RevocationReason.Unknown
+    WHERE [Status] = -1          -- OrganizationUserStatusType.Revoked
+      AND [RevocationReason] IS NULL;
+
+    SET @RowsAffected = @@ROWCOUNT;
```
rkac-bw, line 52: Add for concurrency
`WAITFOR DELAY '00:00:00.100'`

util/Migrator/DbScripts/2026-05-25_00_PopulateOrganizationUserRevocationReason.sql
```diff
@@ -0,0 +1,62 @@
...
+END
+GO
+
+-- ========================================
+-- Backfill RevocationReason = 0 (Unknown) for revoked users
+-- ========================================
+-- Sets RevocationReason to Unknown (0) for every revoked user that pre-dates the
+-- revocation-reason feature. See RevocationReason enum: Unknown is explicitly
+-- defined as "migrated from before revocation reasons were tracked."
+DECLARE @BatchSize INT = 5000;
```
rkac-bw: change to 2000 for concurrency
`DECLARE @BatchSize INT = 2000;`

rkac-bw:
Looks good!

rkac-bw (approved):
LGTM

PR #7716, opened 2026-05-26
[PM-38116] Harden SSRF protection in IPAddressExtensions

mandreko-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38116

## 📔 Objective

Tighten the SSRF block list used by `IPAddressExtensions.IsInternal()` so it matches the IANA IPv4 and IPv6 Special-Purpose Address Registries, and close two classes of bypass:

1. **IPv6 embedded IPv4.** An attacker could previously target an internal IPv4 host (e.g. `169.254.169.254` cloud metadata, `10.0.0.0/8`) by encoding it inside an RFC 6052 NAT64 (`64:ff9b::/96`) or RFC 3056 6to4 (`2002::/16`) address. The IPv6 wrapper passed the old prefix checks and the gateway on the path would translate the destination back to the internal IPv4 host. The new implementation extracts the embedded IPv4 and re-runs the IPv4 reserved-range check.
2. **Over- and under-blocking from prefix string matching.** The old `StartsWith("fe")` check treated all of `fe00::/8` as link-local, when only `fe80::/10` is. Several reserved ranges (192.0.0.0/24 incl. Oracle Cloud metadata, TEST-NETs, 6to4 anycast, benchmarking, multicast, 240.0.0.0/4, RFC 9637 documentation, Teredo, ORCHID, etc.) were not blocked at all.

Implementation uses `System.Net.IPNetwork.Contains` against an explicit table of CIDR ranges per IANA registry.

### DIFF
--- a/src/Core/Utilities/IPAddressExtensions.cs
+++ b/src/Core/Utilities/IPAddressExtensions.cs
@@ -1,4 +1,4 @@
-using System.Net;
+﻿using System.Net;
 using System.Net.Sockets;
 
 namespace Bit.Core.Utilities;
--- a/src/Core/Utilities/IPAddressExtensions.cs
+++ b/src/Core/Utilities/IPAddressExtensions.cs
@@ -1,52 +1,112 @@
-﻿using System.Net;
+using System.Net;
+using System.Net.Sockets;
 
 namespace Bit.Core.Utilities;
 
 /// <summary>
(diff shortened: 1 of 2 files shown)


mandreko-bitwarden:
Closing this as #7721 looks like it should be slightly better.

PR #7721, opened 2026-05-26
[PM-38116] Harden SSRF protection in IPAddressExtensions

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-38116)

## 📔 Objective

Tighten the SSRF block list used by IPAddressExtensions.IsInternal() so it matches the IANA IPv4 and IPv6 Special-Purpose Address Registries, and close two classes of bypass:

IPv6 embedded IPv4. An attacker could previously target an internal IPv4 host (e.g. 169.254.169.254 cloud metadata, 10.0.0.0/8) by encoding it inside an RFC 6052 NAT64 (64:ff9b::/96) or RFC 3056 6to4 (2002::/16) address. The IPv6 wrapper passed the old prefix checks and the gateway on the path would translate the destination back to the internal IPv4 host. The new implementation extracts the embedded IPv4 and re-runs the IPv4 reserved-range check.
Over- and under-blocking from prefix string matching. The old StartsWith("fe") check treated all of fe00::/8 as link-local, when only fe80::/10 is. Several reserved ranges (192.0.0.0/24 incl. Oracle Cloud metadata, TEST-NETs, 6to4 anycast, benchmarking, multicast, 240.0.0.0/4, RFC 9637 documentation, Teredo, ORCHID, etc.) were not blocked at all.
Implementation uses System.Net.IPNetwork.Contains against an explicit table of CIDR ranges per IANA registry.

### DIFF
--- a/src/Core/Utilities/IPAddressExtensions.cs
+++ b/src/Core/Utilities/IPAddressExtensions.cs
@@ -87,26 +87,11 @@ public static class IPAddressExtensions
             }
         }
 
-        foreach (var network in _reservedIPv6Networks)
-        {
-            if (network.Contains(ip))
-            {
-                return true;
-            }
-        }
-
-        return false;
+        return _reservedIPv6Networks.Any(network => network.Contains(ip));
     }
 
     private static bool IsReservedIPv4(IPAddress ip)
     {
-        foreach (var network in _reservedIPv4Networks)
-        {
-            if (network.Contains(ip))
-            {
-                return true;
-            }
-        }
-        return false;
+        return _reservedIPv4Networks.Any(network => network.Contains(ip));
     }
 }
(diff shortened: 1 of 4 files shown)


nick-livefront (dismissed):

mandreko-bitwarden (dismissed):
Before this lands, we should also add: 168.63.129.16/32

https://learn.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16?tabs=windows

jengstrom-bw:
> Before this lands, we should also add: 168.63.129.16/32
> 
> https://learn.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16?tabs=windows

Added [f5b2474](https://github.com/bitwarden/server/pull/7721/commits/f5b2474c50e12618b494a8adb583258463e0a741)

nick-livefront (approved):

PR #7722, opened 2026-05-26
[PM-37170] feat: Add churn-mitigation offer query, command, and API endpoints

amorask-bitwarden:
## 🎟️ Tracking

[PM-37170](https://bitwarden.atlassian.net/browse/PM-37170)

## 📔 Objective

Server-side machinery for the Track A pre-Phase-2 save offer (Mitigating Churn). Adds:

- `IGetChurnMitigationOfferQuery` — eligibility predicate that forks by cohort type. Migration cohorts (`MigrationPathId IS NOT NULL`) require an attached Stripe `SubscriptionSchedule` in Phase 1 with the cohort's churn coupon not already on Phase 2. Churn-only cohorts (`MigrationPathId IS NULL`) require the coupon not currently on `Subscription.Discounts` or `Customer.Discount`, plus a per-assignment one-shot guard (`ChurnDiscountAppliedDate IS NULL`) when the coupon's duration is `once`.
- `RedeemChurnMitigationOfferCommand` — re-validates eligibility via the same query (TOCTOU guard), then mutates the appropriate Stripe surface. Migration cohorts use the modify-phases pattern to append the churn coupon to Phase 2 of the active schedule; churn-only cohorts attach it directly to `Subscription.Discounts`. Set-union semantics so re-running is a no-op (Stripe doesn't dedupe identical coupons in a `discounts` array).
- Two endpoints on `OrganizationBillingVNextController` (`GET .../churn-mitigation-offer`, `POST .../churn-mitigation-offer/redeem`) authorized via `ManageOrganizationBillingRequirement`.

Standalone implementation in the `Bit.Core.Billing.Organizations.PlanMigration` namespace — does not touch `SubscriptionDiscountService`, `IDiscountAudienceFilter`, or any audience-discount artifact. Design rationale: narrower blast radius and no coupling to the existing audience-discount system contracts.

### Design notes worth knowing before reviewing

- **No feature flag** — cohort assignment + `cohort.IsActive` are the gate. Adding a flag would be redundant; ops controls availability via `IsActive`.
- **Self-hosted exclusion** is the controller-level `[SelfHosted(NotSelfHostedOnly = true)]`; no redundant check in the query.
- **Redeem ordering is branch-specific.**
  - *Migration cohort* — Stripe-first, then `ChurnDiscountAppliedDate` stamp. Safe to retry on DB failure because Phase 2 is in the future and Stripe doesn't consume the coupon between attempts, so the next attempt's set-union no-ops at the Stripe layer.
  - *Churn-only cohort* — `ChurnDiscountAppliedDate` stamped first, then `UpdateSubscriptionAsync`, with rollback of the DB stamp if Stripe throws. This defends against the double-application window for `once`-duration coupons: Stripe-first would let a DB-write failure leave the field NULL, and once Stripe consumed the coupon at the next invoice a retry would re-evaluate as eligible and attach a second time. Worst-case failure (Stripe fails + rollback also fails) leaves the field set requiring an ops manual clear — strictly better than silent double-application.
- **Past-phase Stripe error is not explicitly caught.** Stripe doesn't publish a code for "cannot modify past phase" and the codebase convention is to match on `.Code` only. The `EndDate > now` filter and phase-count switch already prevent the call upstream.
- **Modify-phases pattern** matches [PM-37083](https://bitwarden.atlassian.net/browse/PM-37083) / `UpcomingInvoiceHandler.EnableAutomaticTaxAsync` / `UpdateBillingAddressCommand` — canonical for the codebase. The query also mirrors the redeem command's `EndDate > now` phase filter to defend against Stripe's 3-phase schedule normalization (anchor + Phase 1 + Phase 2) after any post-attach subscription mutation.
- **`MigratedDate` analytics caveat:** Marketing-analytics queries on `MigratedDate` are moot until [PM-37092](https://bitwarden.atlassian.net/browse/PM-37092) + a backfill ship. This PR's migration-cohort branch is still correct without it — the eligibility window closes via the schedule's `current_phase` check.
- **GET wire format:** ineligible orgs return 200 + empty body (`Content-Length: 0`, no `Content-Type`). `TypedResults.Ok((T?)null)` does not serialize as the literal JSON `null` token — System.Text.Json elides a top-level null. Angular's `HttpClient` coerces this to `null` so the [PM-37173](https://bitwarden.atlassian.net/browse/PM-37173) consumer can branch on `response === null` without ceremony.

### Tested locally end-to-end

- **Migration-cohort branch:** cohort assignment with `MigrationPathId = 1`, `PriceIncreaseScheduler` ran after test-clock advance to T−7, GET returned the offer record, POST attached the coupon to Phase 2 of the active schedule, follow-up GET returned 200 + empty body (ineligible), `ChurnDiscountAppliedDate` stamped on the assignment row.
- **Churn-only branch:** cohort assignment with `MigrationPathId = NULL`, GET returned the offer, POST attached the coupon directly to `Subscription.Discounts`, follow-up GET returned 200 + empty body (`once`-duration one-shot guard fired), `ChurnDiscountAppliedDate` stamped.

### Out of scope

- Web vault wedge (Angular dialog + cancel-flow integration) — lands in [PM-37173](https://bitwarden.atlassian.net/browse/PM-37173).
- Track B (2019 plans), Track C (Teams Starter) — frozen pending separate strategy work.
- Periodic Stripe coupon re-validation — v1.1 follow-up.
- Audience-discount integration (`SubscriptionDiscountService`, `IDiscountAudienceFilter`, `DiscountTierType`) — explicitly rejected design.

[PM-37170]: https://bitwarden.atlassian.net/browse/PM-37170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-37083]: https://bitwarden.atlassian.net/browse/PM-37083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        existingDiscounts.Add(new SubscriptionDiscountOptions { Coupon = churnDiscountCouponCode });
+
+        await stripeAdapter.UpdateSubscriptionAsync(subscription.Id,
+            new SubscriptionUpdateOptions { Discounts = existingDiscounts });
+
+        assignment.ChurnDiscountAppliedDate = DateTime.UtcNow;
+        assignment.RevisionDate = DateTime.UtcNow;
+        await assignmentRepository.ReplaceAsync(assignment);
```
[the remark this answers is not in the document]
amorask-bitwarden: Resolved here: https://github.com/bitwarden/server/pull/7722/commits/926e02270aa4420a63b981c652a2f7d293ddffed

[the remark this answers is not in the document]
amorask-bitwarden: Resolved here: https://github.com/bitwarden/server/pull/7722/commits/926e02270aa4420a63b981c652a2f7d293ddffed

src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
```diff
@@ -0,0 +1,283 @@
...
+        // we skip the Stripe call (Stripe doesn't dedupe identical coupons in a discounts array),
+        // and we preserve the existing ChurnDiscountAppliedDate -- overwriting it would lose the
+        // original redeem timestamp (relevant for Marketing analytics).
+        if (alreadyApplied)
+        {
+            _logger.LogInformation(
+                "{Command}: Coupon already present on Phase 2 of schedule ({ScheduleId}) for Organization ({OrganizationId}); no Stripe update needed",
+                CommandName, activeSchedule.Id, organization.Id);
+            return new None();
+        }
```
kdenney: This block could move up right after the declaration of `alreadyApplied` to be cleaner and also remove the need for the `if (!alreadyApplied)` conditional currently on line 115.

src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
```diff
@@ -0,0 +1,283 @@
...
+                EndBehavior = SubscriptionScheduleEndBehavior.Release,
+                Phases = phases
+            });
+
+        // ChurnDiscountAppliedDate is informational for migration cohorts (eligibility window
+        // closes via Stripe's current_phase advance + MigratedDate from SubscriptionUpdatedHandler
+        // when PM-37092 lands). If this write fails after the Stripe call succeeds, the set-union
+        // semantics above make a retry a no-op -- harmless.
+        assignment.ChurnDiscountAppliedDate = DateTime.UtcNow;
+        assignment.RevisionDate = DateTime.UtcNow;
```
kdenney: ⛏️ Tiny nitpick but this is inconsistent with how UtcNow is used in the below method. If you change one, I'd probably prefer the pattern of creating a local var for utcnow but totally up to you.

src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
```diff
@@ -0,0 +1,283 @@
...
+                .Select(i => new SubscriptionSchedulePhaseItemOptions { Price = i.PriceId, Quantity = i.Quantity })
+                .ToList(),
+            Discounts = phase.Discounts?
+                .Select(d => new SubscriptionSchedulePhaseDiscountOptions { Coupon = d.CouponId })
+                .ToList(),
+            Metadata = phase.Metadata,
+            ProrationBehavior = phase.ProrationBehavior
+        };
+
+    private async Task<Subscription?> FetchSubscriptionAsync(Organization organization)
```
kdenney: ⛏️ Would prefer this to be named `TryGetSubscriptionAsync` to match the method name in `GetChurnMitigationOfferQuery` but this is a very small nitpick 😅

kdenney (dismissed):
This is real clean; nicely done! I only have 3 nitpicks but nothing blocking.

kdenney:
Also: thanks for the thorough PR description!! I was able to scan the technical breakdown again and then read the PR description and it gave me the perfect context to review. 🙌

src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
amorask-bitwarden: Addressed here: https://github.com/bitwarden/server/pull/7722/commits/07bd1fcb80af7238669e79595ff04e9c56f4353d

src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
amorask-bitwarden: Addressed here: https://github.com/bitwarden/server/pull/7722/commits/07bd1fcb80af7238669e79595ff04e9c56f4353d

src/Core/Billing/Organizations/PlanMigration/Commands/RedeemChurnMitigationOfferCommand.cs
amorask-bitwarden: Addressed here: https://github.com/bitwarden/server/pull/7722/commits/07bd1fcb80af7238669e79595ff04e9c56f4353d

kdenney (approved):

PR #7723, opened 2026-05-26
[PM-35599] Update set-password endpoint to handle V1 encryption when given Auth and Unlock data

rr-bw:
## 🎟️ Tracking

[PM-35599](https://bitwarden.atlassian.net/browse/PM-35599)

Client-side changes:
- https://github.com/bitwarden/clients/pull/20643

## 📔 Objective

Update the `accounts/set-password` endpoint to be able to handle a `SetInitialPasswordRequest` that contains `MasterPasswordAuthenticationData` + `MasterPasswordAuthentication` + legacy `KeysRequest`. In other words, update the endpoint to allow a client to pass `MPAD` + `MPUD` but still do V1 encryption. This update is necessary for the corresponding client-side changes in https://github.com/bitwarden/clients/pull/20643

## Routing matrix

| #  | Scenario                                       | MPAD/MPUD | Keys | AccountKeys | V2 MP JIT flag | Branch        | Outcome                                                                              |
| -- | ---------------------------------------------- | --------- | ---- | ----------- | -------------- | ------------- | ------------------------------------------------------------------------------------ |
| 1  | Modern TDE (V1 or V2 TDE user)                 | ✓         | null | null        | any            | Modern TDE    | `_tdeSetPasswordCommand.SetMasterPasswordAsync` (no cryptographic-state change)      |
| 2  | Modern V1 MP JIT                               | ✓         | set  | null        | OFF            | V1            | `ValidateSaltUnchangedForUser` then `_setInitialMasterPasswordCommandV1`; `Keys.ToUser()` sets keypair |
| 3  | Modern V1 MP JIT                               | ✓         | set  | null        | ON             | V1            | Same as row 2 — V2 MP JIT branch requires `AccountKeys != null`, falls through       |
| 4  | V2 MP JIT (SDK)                                | ✓         | null | set         | ON             | V2 MP JIT     | `_finishSsoJitProvisionMasterPasswordCommand.FinishProvisionAsync`                   |
| 5  | V2 MP JIT shape, flag off                      | ✓         | null | set         | OFF            | V1 guard      | **400 BadRequest** — defensive guard in `SetInitialPasswordV1Async`                  |
| 6  | Old MP JIT (pre-PM-35599 client)               | ✗         | set  | null        | any            | V1            | `_setInitialMasterPasswordCommandV1`; `ToUser()` reads legacy KDF fields; keypair set |
| 7  | Old TDE (pre-PM-35599 client)                  | ✗         | null | null        | any            | V1            | `_setInitialMasterPasswordCommandV1`; legacy fields; no keypair mutation             |
| 8  | Partial: MPAD only OR MPUD only                | partial   | any  | any         | any            | Validator     | 400 — `HasAuthAndUnlockData()` is false; legacy validation demands legacy fields too |
| 9  | MPAD + MPUD with mismatched KDF/salt           | ✓         | any  | any         | any            | Validator     | 400 — `KdfSettingsValidator.ValidateAuthenticationAndUnlockData` rejects             |
| 10 | Both `AccountKeys` and `Keys` set              | ✓/✗       | set  | set         | any            | Validator     | 400 — mutual-exclusion rule in `Validate()`                                          |
| 11 | MPAD hash ≠ legacy `MasterPasswordHash`        | ✓         | any  | any         | any            | Validator     | 400 — cross-shape conflict check in `Validate()`                                    |
| 12 | MPUD key ≠ legacy `Key`                        | ✓         | any  | any         | any            | Validator     | 400 — cross-shape conflict check in `Validate()`                                    |
| 13 | Modern V1 MP JIT with divergent salt           | ✓         | set  | null        | any            | V1 salt check | 400 — `ValidateSaltUnchangedForUser` rejects (Stage 1 PM-27044 invariant)            |

## 📸 Screenshots

Regression testing existing paths (tested with `clients` pointing at `main`)

(Note: the new path (`MPAD` + `MPUD` + legacy `Keys`) will be tested as part of client-side changes in https://github.com/bitwarden/clients/pull/20643)

### MP JIT with legacy properties

A MP decryption org invites a user. The user JIT provisions. Upon setting initial password, the request sends legacy properties: `masterPasswordHash` + `key` + `keys`.

https://github.com/user-attachments/assets/3bd01a1f-6ae9-45b8-a831-8ecc0457e6fe

<br />

### MP JIT with new properties and feature flag on

In this video, the `"enable-account-encryption-v2-jit-password-registration"` feature flag on (proven by a breakpoint in the video).

A MP decryption org invites a user. The user JIT provisions. Upon setting initial password, the request sends `MPAD` + `MPUD` + `accountKeys` (via SDK).

https://github.com/user-attachments/assets/8f8d7d50-b258-4f79-bd44-eb34840a478d

<br />

### TDE JIT with Manage Account Recovery Permission

A TDE org invites a user with the "manage account recovery" permission. The user JIT provisions. Upon setting initial password, the request sends legacy properties: `masterPasswordHash` + `key` + NO keypair.

https://github.com/user-attachments/assets/b66f6b8f-9f72-453f-a500-263fe9f3aae1

[PM-35599]: https://bitwarden.atlassian.net/browse/PM-35599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -250,48 +251,80 @@ public async Task PostSetPasswordAsync([FromBody] SetInitialPasswordRequestModel
...
+            // We require AccountKeys (the new key shape) here, not legacy Keys — otherwise
+            // a modern V1 MP JIT request (MPAD + MPUD + legacy Keys) would be incorrectly routed here
+            // when the flag is on, and `model.ToData().AccountKeys` would be null, breaking the V2 MP
+            // JIT command (which requires AccountKeys per FinishSsoJitProvisionMasterPasswordCommand).
+            if (model.AccountKeys != null &&
+                _featureService.IsEnabled(FeatureFlagKeys.EnableAccountEncryptionV2JitPasswordRegistration))
             {
                 await _finishSsoJitProvisionMasterPasswordCommand.FinishProvisionAsync(user, model.ToData());
+                return;
             }
```
JaredSnider-Bitwarden, line 318: Per discussion with @Thomas-Avery , if we get a V2 payload when the flag is off, we cannot process the request. That is an error state that shouldn't happen unless a BW client is out of sync w/ the FF being turned off, but we should still handle it.

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -250,48 +251,80 @@ public async Task PostSetPasswordAsync([FromBody] SetInitialPasswordRequestModel
...
-        if (model.IsV2Request())
+        // Modern-shape request (MPAD + MPUD set). Try V2 encryption paths if the relevant
+        // feature flag is enabled. If the modern-shape request doesn't match a V2 branch
+        // (e.g., V2 flag off, or modern client doing V1 MP JIT), fall through to V1.
+        if (model.HasAuthAndUnlockData())
         {
-            if (model.IsTdeSetPasswordRequest())
+            // V2 encryption - TDE user with "manage account recovery" permission
+            if (model.IsTdeSetPasswordRequest() &&
+                _featureService.IsEnabled(FeatureFlagKeys.V2RegistrationTDEJIT))
```
JaredSnider-Bitwarden: ⚠️ : we cannot leverage a feature flag here for this particular scenario because, if an issue is discovered with the `V2RegistrationTDEJIT` flag where it is turned off, TDE users without MPs (who can have v1 or v2 encryption for their user key) can still be given the manage account recovery permission and they must be able to set a password. 

`V2RegistrationTDEJIT` just governs SSO + TDE account creation - this process of setting an initial password happens later.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
```diff
@@ -21,16 +23,16 @@ public class SetInitialPasswordRequestModel : IValidatableObject
     [Obsolete("Use AccountKeys instead")]
     public KeysRequestModel? Keys { get; set; }
 
-    [Obsolete("Use MasterPasswordAuthentication instead")]
+    [Obsolete("Use MasterPasswordUnlock instead")]
```
JaredSnider-Bitwarden, line 23: ❓ : why the change?

JaredSnider-Bitwarden (dismissed):
Partial review but leaving this here:

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
rr-bw: No major reason. "Unlock" data is just more consistent with what the `ToUser()` method uses after the changes on this PR.

src/Api/Auth/Controllers/AccountsController.cs
rr-bw: If what you're getting at here is...

- If the request contains `MPAD` + `MPUD` + `AccountKeys`, but the `EnableAccountEncryptionV2JitPasswordRegistration` flag is OFF on the server, then throw.

...then that error state / thrown exception is handled at the top of `SetInitialPasswordV1Async()`:

```csharp

private async Task SetInitialPasswordV1Async(User user, SetInitialPasswordRequestModel model)
{
  // Defensive: V1 cannot consume AccountKeys (the new key shape). If a request carries
  // AccountKeys we'd silently drop the keypair, so fail loudly. This can only happen if
  // the V2 MP JIT flag is off (otherwise the V2 branch above would have handled it) — i.e.,
  // a client/server flag-state mismatch or a non-Angular caller.
  if (model.AccountKeys != null)
  {
       "Request includes V2 AccountKeys but V2 encryption is not enabled.");
  }
  // ...
```

src/Api/Auth/Controllers/AccountsController.cs
rr-bw: https://github.com/bitwarden/server/pull/7723/changes/17dfee5c013ceac335ec068820de983ae7153e40 - includes comments with your note about `V2RegistrationTDEJIT` governing SSO + TDE account creation. Might help avoid confusion with the asymmetry (i.e. not checking a flag for TDE, but checking a flag for MP JIT).

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
```diff
@@ -43,24 +48,51 @@ public class SetInitialPasswordRequestModel : IValidatableObject
...
+        existingUser.Kdf = MasterPasswordUnlock?.Kdf.KdfType ?? Kdf!.Value;
+        existingUser.KdfIterations = MasterPasswordUnlock?.Kdf.Iterations ?? KdfIterations!.Value;
+        existingUser.KdfMemory = MasterPasswordUnlock?.Kdf.Memory ?? KdfMemory;
+        existingUser.KdfParallelism = MasterPasswordUnlock?.Kdf.Parallelism ?? KdfParallelism;
+        existingUser.Key = MasterPasswordUnlock?.MasterKeyWrappedUserKey ?? Key;
+
+        // MasterPasswordSalt column must never be null/empty after a successful password-set
+        // operation. Modern clients send an explicit salt via MPUD; older clients don't send one,
+        // so we fall back to the email-derived V1 salt via GetMasterPasswordSalt().
+        existingUser.MasterPasswordSalt = MasterPasswordUnlock?.Salt ?? existingUser.GetMasterPasswordSalt();
```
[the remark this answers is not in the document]
rr-bw: Addressed in: https://github.com/bitwarden/server/pull/7723/changes/4e168d919c8648923116495864220fd548989eb1

src/Api/Auth/Controllers/AccountsController.cs
ike-kottlowski: That does handle the aforementioned case. Thank you.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
```diff
@@ -115,15 +156,27 @@ public IEnumerable<ValidationResult> Validate(ValidationContext validationContex
...
-        // AccountKeys can be null for TDE users, so we don't check that here
         return MasterPasswordAuthentication != null && MasterPasswordUnlock != null;
     }
 
+    /// <summary>
+    /// True when the request does NOT contain a keypair. TDE users don't send a keypair on the
+    /// request because they already have one. Checks both AccountKeys (new) and Keys (legacy) so
+    /// the predicate is correct for the transitional period where clients may send either key shape.
+    /// </summary>
     public bool IsTdeSetPasswordRequest()
```
ike-kottlowski, line 184: ❓ : If a client could send either `AccountKeys` or `Keys` shouldn't the logic be `||` (OR) not `&&` (AND). 

If `AccountKeys == null` then it's V1 TDESetPassword, if `Keys == null` then it's V2 TDESetPassword?

But both have to be `null` in this case? Maybe the logic works either way?

test/Api.IntegrationTest/Auth/Controllers/AccountsControllerTest.cs
```diff
@@ -999,6 +999,10 @@ await OrganizationTestHelpers.CreateUserAsync(_factory, organization.Id, userEma
         Assert.Null(updatedUser.KdfMemory);
         Assert.Null(updatedUser.KdfParallelism);
 
+        // MasterPasswordSalt column must never be null/empty after a successful set-password.
+        // Legacy V1 path falls back to the email-derived V1 salt.
+        Assert.Equal(userEmail, updatedUser.MasterPasswordSalt);
```
ike-kottlowski, line 1004: 🎨 : this test feels brittle, when the salt is no longer the email then this would break? 

Maybe we could test if the salt is the same as the model input? (which currently does set it to email but allows for better flexibility in the future)

test/Api.IntegrationTest/Auth/Controllers/AccountsControllerTest.cs
```diff
@@ -1020,10 +1024,267 @@ await OrganizationTestHelpers.CreateUserAsync(_factory, organization.Id, userEma
...
+
+        // Verify KDF settings — ToUser reads from MPUD on the modern shape
+        Assert.Equal(KdfType.PBKDF2_SHA256, updatedUser.Kdf);
+        Assert.Equal(600_000, updatedUser.KdfIterations);
+        Assert.Null(updatedUser.KdfMemory);
+        Assert.Null(updatedUser.KdfParallelism);
+
+        // MasterPasswordSalt column must never be null/empty after a successful set-password.
+        // Modern V1 MP JIT path persists the MPUD-provided salt (set to userEmail in the request).
+        Assert.Equal(userEmail, updatedUser.MasterPasswordSalt);
```
ike-kottlowski: 🎨 : Similar to above about the brittle nature of the test. 

If we plan to remove this test wholesale in the future when V2 is standard we might be able to flag this test as to-be-removed?

test/Api.IntegrationTest/Auth/Controllers/AccountsControllerTest.cs
```diff
@@ -1020,10 +1024,267 @@ await OrganizationTestHelpers.CreateUserAsync(_factory, organization.Id, userEma
...
+
+        // Verify KDF settings persisted from MPUD
+        Assert.Equal(KdfType.PBKDF2_SHA256, updatedUser.Kdf);
+        Assert.Equal(600_000, updatedUser.KdfIterations);
+        Assert.Null(updatedUser.KdfMemory);
+        Assert.Null(updatedUser.KdfParallelism);
+
+        // MasterPasswordSalt column must never be null/empty after a successful set-password.
+        // TDE command persists the MPUD-provided salt (the helper sends userEmail as the salt value).
+        Assert.Equal(userEmail, updatedUser.MasterPasswordSalt);
```
ike-kottlowski, line 1259: 🎨 : as above https://github.com/bitwarden/server/pull/7723/changes#r3724148393

test/Api.Test/Auth/Controllers/AccountsControllerTests.cs
```diff
@@ -1101,14 +1103,278 @@ public async Task PostSetPasswordAsync_V2_WhenSettingPasswordFails_ShouldThrowEx
...
+    // Modern clients send a salt via MPUD; ToUser should persist that exact value.
+    [Theory]
+    [BitAutoData]
+    public async Task PostSetPasswordAsync_V1_NewClient_PersistsMpudSaltAsync(
+        User user,
+        SetInitialPasswordRequestModel setInitialPasswordRequestModel)
+    {
+        // Arrange — modern MP JIT client: MPAD + MPUD + legacy Keys (no AccountKeys)
+        // Salt must match the user's email-derived salt (Stage 1 PM-27044 invariant).
+        var emailSalt = user.GetMasterPasswordSalt();
```
ike-kottlowski, line 1297: 👍🏻: In this test since we call `GetMasterPasswordSalt` the salt may or may not be an email thus adding flexibility.

test/Api.Test/Auth/Controllers/AccountsControllerTests.cs
```diff
@@ -1101,14 +1103,278 @@ public async Task PostSetPasswordAsync_V2_WhenSettingPasswordFails_ShouldThrowEx
...
+        // Assert — user.MasterPasswordSalt matches the MPUD-provided salt (email-derived)
+        Assert.Equal(emailSalt, user.MasterPasswordSalt);
+        Assert.False(string.IsNullOrEmpty(user.MasterPasswordSalt));
+    }
+
+    // For older clients that don't send MPUD, MasterPasswordSalt falls back to the email-derived
+    // V1 salt (email.ToLowerInvariant().Trim()) so the column is never null after a successful set.
+    [Theory]
+    [BitAutoData]
+    public async Task PostSetPasswordAsync_V1_OldClient_PersistsEmailDerivedSaltAsync(
```
ike-kottlowski, line 1325: 🎨 : May need to be marked for removal when we unwind?

ike-kottlowski (commented):
Just some more questions. Otherwise looks good.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
rr-bw: `IsTdeSetPasswordRequest()` is not concerned with distinguishing V1 vs. V2 encryption users. It’s simply asking “Is this a TDE user?”, which is determined by the complete absence (hence the use of `&&`) of a key pair of either sort on the request.

The method doesn’t need to distinguish V1 vs. V2 encryption because in a TDE set-password scenario, the user’s cryptographic state was already previously set up (whether V1 or V2), and all we are doing through the `"set-password"` endpoint is just setting the password via `_tdeSetPasswordCommand.SetMasterPasswordAsync()` (i.e. that command does *not* set cryptographic state; it just sets the password).

```csharp
// TDE set-password (for TDE users who obtain the "manage account recovery" permission).
// _tdeSetPasswordCommand handles both V1 and V2 TDE users (it sets the master password
// without touching cryptographic state).
// 
// Note: Why not check the V2RegistrationTDEJIT flag?
// The V2RegistrationTDEJIT flag governs SSO+TDE account creation, not set-password, so
// don't reference it here (no feature-flag gate), because it isn't relevant. A TDE user
// reaching this endpoint already has keys that were set up at registration time, regardless
// of the flag.
if (model.IsTdeSetPasswordRequest())
{
	await _tdeSetPasswordCommand.SetMasterPasswordAsync(user, model.ToData());
	return;
}
```

This is distinct from our handling of MP JIT users, where we do need to distinguish between V1 vs. V2 encryption. The `_finishSsoJitProvisionMasterPasswordCommand.FinishProvisionAsync()` method *does* set cryptographic state - specifically it sets V2 encryption state via `_userRepository.SetV2AccountCryptographicStateAsync()` - which is why that branch must be gated by an `AccountKeys` presence check (not legacy `Keys`) as well as a feature flag check.

```csharp
// V2 encryption - MP JIT.
// We require AccountKeys (the new key shape) here, not legacy Keys — otherwise
// a modern V1 MP JIT request (MPAD + MPUD + legacy Keys) would be incorrectly routed here
// when the flag is on, and `model.ToData().AccountKeys` would be null, breaking the V2 MP
// JIT command (which requires AccountKeys per FinishSsoJitProvisionMasterPasswordCommand).
if (model.AccountKeys != null &&
	_featureService.IsEnabled(FeatureFlagKeys.EnableAccountEncryptionV2JitPasswordRegistration))
{
	await _finishSsoJitProvisionMasterPasswordCommand.FinishProvisionAsync(user, model.ToData());
	return;
}
```

But if the request instead contains `Keys`, then we are dealing with a V1 encryption and so we fall back to `SetInitialPasswordV1Async()`.

test/Api.IntegrationTest/Auth/Controllers/AccountsControllerTest.cs
rr-bw: We can’t assert against the request input for this one because this is a legacy request, which doesn’t contain `MPAD/MPUD` and therefore doesn’t contain a salt.

So the assertion for this test specifically checks that the server-side fallback-to-email is working correctly.

test/Api.IntegrationTest/Auth/Controllers/AccountsControllerTest.cs
rr-bw: Test updated in https://github.com/bitwarden/server/pull/7723/changes/7216ac640dfadfd017c88f67c51afeaef2a7b641. 

As for flagging this test for removal: I figure that once the underlying code - `SetInitialPasswordV1Async()` - gets removed, this test will naturally fail, making it obvious that the test can be removed. So I don’t see a need for a TODO comment to remind us of future cleanup.

`SetInitialPasswordV1Async()` does have to wait for 2 things to happen before it can be removed (see doc comment on the method).

```csharp
/// <summary>
/// Handles setting an initial password for V1 users in the following scenarios:
/// 1. TDE user where request contains legacy fields (MasterPasswordHash + Key)
/// 2. MP JIT user where request contains legacy fields (MasterPasswordHash + Key + Keys)
/// 3. MP JIT user where request contains MPAD + MPUD + legacy Keys (not AccountKeys)
/// </summary>
/// <remarks>
/// TODO: In order to remove this method, all 3 of those scenarios need to become unused. This means
/// removal can only happen when BOTH of the following are true:
/// 
/// 1. Client-side changes in https://bitwarden.atlassian.net/browse/PM-35599 have been merged
///    and have aged out according to the Bitwarden release support policy. This covers scenarios
///    #1-2 above (legacy TDE and legacy MP JIT).
/// 
/// 2. The EnableAccountEncryptionV2JitPasswordRegistration feature flag has been unwound in
///    https://bitwarden.atlassian.net/browse/PM-27327 and the client-side portion of the flag removal has
///    aged out according to the Bitwarden release support policy. This covers scenarios #2-3 above (legacy
///    and modern MP JIT).
/// </remarks>
private async Task SetInitialPasswordV1Async(User user, SetInitialPasswordRequestModel model)
```

If number 1 happens first (the client changes age out), then the test does become moot and we could remove it then. But since the underlying `SetInitialPasswordV1Async()` code implementation must remain until number 2 happens (flag unwinding), I say just keep the test until then as the test still technically covers a code-path.

test/Api.IntegrationTest/Auth/Controllers/AccountsControllerTest.cs
rr-bw: Addressed (indirectly) in https://github.com/bitwarden/server/pull/7723/changes/7216ac640dfadfd017c88f67c51afeaef2a7b641

test/Api.Test/Auth/Controllers/AccountsControllerTests.cs
rr-bw: Test will fail once the underlying `SetInitialPasswordV1Async()` is removed, so we can just naturally remove it then instead of adding a TODO reminder.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
rr-bw: @mzieniukbw Do you believe that the above is an accurate description? Especially regarding the idea that the `_tdeSetPasswordCommand.SetMasterPasswordAsync()` covers (sets a password for) both V1 and V2 encryption TDE users, and therefore there is no need to distinguish between V1 vs. V2 encryption for TDE users?

ike-kottlowski (dismissed):

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
mzieniukbw: It should honestly be `Use MasterPasswordAuthentication or MasterPasswordUnlock instead`. Does not matter to me, can be just one and any.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
```diff
@@ -43,24 +48,59 @@ public class SetInitialPasswordRequestModel : IValidatableObject
...
+        if (AccountKeys != null && Keys != null)
+        {
+            yield return new ValidationResult(
+                $"Cannot specify both {nameof(AccountKeys)} and {nameof(Keys)}. Provide exactly one keypair.",
+                [nameof(AccountKeys), nameof(Keys)]);
+        }
+
+        // If both modern and legacy fields are present for the same value, they must agree —
+        // reject ambiguous input rather than silently letting the modern field win via ?? fallback.
+        if (MasterPasswordAuthentication != null && MasterPasswordHash != null
```
mzieniukbw: That impossible scenario. Furthermore, you should `MasterPasswordAuthentication` or `MasterPasswordHash`, not both at the same time. (Checked existing code and linked clients PR).
If you want to check for correctness, then expect there is at least one - preferably expect that either `MasterPasswordAuthentication` and `MasterPasswordUnlock` are provided (with `Keys` or `AccountKeys` to distinguish modern V1 vs V2) at the same time OR the other legacy. If you have both or none, then that's should be validation error.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
```diff
@@ -43,24 +48,59 @@ public class SetInitialPasswordRequestModel : IValidatableObject
...
+            && MasterPasswordAuthentication.MasterPasswordAuthenticationHash != MasterPasswordHash)
+        {
+            yield return new ValidationResult(
+                $"{nameof(MasterPasswordAuthentication.MasterPasswordAuthenticationHash)} and {nameof(MasterPasswordHash)} are both present but differ. Provide only one.",
+                [nameof(MasterPasswordAuthentication), nameof(MasterPasswordHash)]);
+        }
+
+        if (MasterPasswordUnlock != null && Key != null
```
mzieniukbw: See my other comment, this scenario does not exist in clients.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
```diff
@@ -43,24 +48,59 @@ public class SetInitialPasswordRequestModel : IValidatableObject
...
+        existingUser.MasterPasswordSalt = MasterPasswordUnlock?.Salt ?? existingUser.GetMasterPasswordSalt();
+
         Keys?.ToUser(existingUser);
         return existingUser;
     }
 
     public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
     {
-        if (IsV2Request())
+        if (AccountKeys != null && Keys != null)
```
mzieniukbw, line 76: See my other comment, this scenario does not exist in clients.

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
mzieniukbw: I am fine with this one. We should not expect either. That being said, see my other comment, where i have concerns about gaps in validation. We should really expect one or another, never a mix of MPAD/MPUD and legacy.

mzieniukbw (changes requested):

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
rr-bw: Update here: https://github.com/bitwarden/server/pull/7723/changes/c0e2ba792c237cd07bba3e57cca3134e4973aaea (and related tests https://github.com/bitwarden/server/pull/7723/changes/ed33516207606ec2ba92f08851a65aadc1c885e9)

- `if (hasModern && hasLegacy)` - ensures that we don't have both shapes
- After that, the `if (HasAuthAndUnlockData())` check, followed by the `if (string.IsNullOrEmpty(MasterPasswordHash))` and `if (string.IsNullOrEmpty(Key))` checks, will validate that we have one or the other (i.e. if we receive neither, we yield `"MasterPasswordHash must be supplied."`

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
rr-bw: Update here: https://github.com/bitwarden/server/pull/7723/changes/c0e2ba792c237cd07bba3e57cca3134e4973aaea (and related tests https://github.com/bitwarden/server/pull/7723/changes/ed33516207606ec2ba92f08851a65aadc1c885e9)

- `if (hasModern && hasLegacy)` - ensures that we don't have both shapes
- After that, the `if (HasAuthAndUnlockData())` check, followed by the `if (string.IsNullOrEmpty(MasterPasswordHash))` and `if (string.IsNullOrEmpty(Key))` checks, will validate that we have one or the other (i.e. if we receive neither, we yield `"MasterPasswordHash must be supplied."`

src/Api/Auth/Models/Request/Accounts/SetInitialPasswordRequestModel.cs
rr-bw: I would argue keep this. Even though clients don't send both, the endpoint/model technically does accept both at the same time, so this is just a defensive measure.

mzieniukbw (dismissed):

ike-kottlowski (approved):

mzieniukbw (approved):

PR #7724, opened 2026-05-27
[BEEEP] - Model Binding for Org and Org User Validation

jrmccannon:
## 🎟️ Tracking
[PM-38337](https://bitwarden.atlassian.net/browse/PM-38337)

## 📔 Objective
The purpose of this is to avoid unnecessary calls to the database when evaluating if the values passed through the route parameters are valid. This is scoped to organization and organization users.  When valid, this will return a new object containing the validated organization and organization user.

## 📸 Screenshots

[PM-38337]: https://bitwarden.atlassian.net/browse/PM-38337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eliykat (commented):
To continue discussion from the breakdown:

The nice thing about keeping each binder scoped to a single resource is that they are small, single purpose, and composable. For example:

- org controller: [InjectOrganization]
- group controller: [InjectOrganization], [InjectGroup]
- orgUser controller: [InjectOrganization], [InjectOrganizationUser]
- policy controller: [InjectOrganization], [InjectPolicy]

Instead of having to reimplement both together each time:

- [InjectOrganization]
- [InjectOrganizationAndGroup]
- [InjectOrganizationAndUser]
- [InjectOrganizationAndPolicy]
- (and potentially bulk variants if we add those)

Unless there is any reason why they need to be combined.

It also fits the "interface" of the attribute because each is applied to its respective parameter. `InjectValidOrganizationAndUserAttribute` must be applied to 1 parameter but actually binds 2.

jrmccannon:
> To continue discussion from the breakdown:
> 
> The nice thing about keeping each binder scoped to a single resource is that they are small, single purpose, and composable. For example:
> 
>     * org controller: [InjectOrganization]
> 
>     * group controller: [InjectOrganization], [InjectGroup]
> 
>     * orgUser controller: [InjectOrganization], [InjectOrganizationUser]
> 
>     * policy controller: [InjectOrganization], [InjectPolicy]
> 
> 
> Instead of having to reimplement both together each time:
> 
>     * [InjectOrganization]
> 
>     * [InjectOrganizationAndGroup]
> 
>     * [InjectOrganizationAndUser]
> 
>     * [InjectOrganizationAndPolicy]
> 
>     * (and potentially bulk variants if we add those)
> 
> 
> Unless there is any reason why they need to be combined.
> 
> It also fits the "interface" of the attribute because each is applied to its respective parameter. `InjectValidOrganizationAndUserAttribute` must be applied to 1 parameter but actually binds 2.

Yeah that makes sense.

Did you look at the way of registering the providers? 

Did we want to keep the match and switch methods?

Did we want to keep the extension method for getting the org user Id from the context?

src/Api/AdminConsole/Authorization/HttpContextExtensions.cs
```diff
@@ -68,39 +68,35 @@ private static async Task<IEnumerable<ProviderUserOrganizationDetails>> GetProvi
...
 
     /// <summary>
     /// Parses the {providerId} route parameter into a Guid, or throws if it is not present or is not a valid Guid.
     /// </summary>
     /// <exception cref="InvalidOperationException"></exception>
     public static Guid GetProviderId(this HttpContext httpContext)
+        => httpContext.TryGetRouteParameterAsGuid("providerId")
+           ?? throw new InvalidOperationException(NoProviderIdError);
+
+    private static Guid? TryGetRouteParameterAsGuid(this HttpContext httpContext, string routeParameterName)
```
eliykat: I like `TryGetRouteParameterAsGuid` and support the changes in this file to use it 👍

src/Api/AdminConsole/Authorization/HttpContextExtensions.cs
```diff
@@ -68,39 +68,35 @@ private static async Task<IEnumerable<ProviderUserOrganizationDetails>> GetProvi
...
-        routeValues.TryGetValue("organizationId", out var organizationIdParam);
-        if (organizationIdParam != null && Guid.TryParse(organizationIdParam.ToString(), out var organizationId))
-        {
-            return organizationId;
-        }
-
-        throw new InvalidOperationException(NoOrgIdError);
-    }
+    public static Guid? GetOrganizationUserId(this HttpContext httpContext, string organizationUserRouteParamName = "id")
+        => httpContext.TryGetRouteParameterAsGuid(organizationUserRouteParamName);
```
eliykat: A wrapper method makes sense for `GetOrganizationId` which will be widely used. But this is probably only going to be used in 1 place, which is the `InjectOrganizationUserAttribute`; in which case it can probably just call `TryGetRouteParameterAsGuid` directly. (This is also pretty close to being a bare pass-through, especially because you can override the param name anyway, so what does it really give us?)

src/Core/AdminConsole/Utilities/v2/Validation/ValidationResult.cs
```diff
@@ -17,6 +17,20 @@ public class ValidationResult<TRequest>(TRequest request, OneOf<Error, None> err
...
+    public void SwitchResult(Action<Error> error, Action<TRequest> valid) =>
+        Switch(error, _ => valid(Request));
+
+    /// <summary>
+    /// Returns the result of <paramref name="error"/> when validation failed, or the result of
+    /// <paramref name="valid"/> invoked with the validated <see cref="Request"/> when validation succeeded.
+    /// </summary>
+    public TResult MatchResult<TResult>(Func<Error, TResult> error, Func<TRequest, TResult> valid) =>
+        Match(error, _ => valid(Request));
 }
```
eliykat, line 34: I'm OK with these as a pattern, but aren't they bare wrappers around what the base class gives us directly? In other words, could we just call `Switch` directly?

jrmccannon: The existing switch and map methods don't give us the Request object back to be used within the closure

JimmyVo16:
LGTM, but is this file `src/Core/AdminConsole/Utilities/v2/Shared/Errors.cs` not being referenced anywhere? Also, there are still some discussions ongoing; I can review again once that’s done.

src/Api/AdminConsole/Authorization/HttpContextExtensions.cs
```diff
@@ -12,95 +12,83 @@ public static class HttpContextExtensions
...
 
-    /// <summary>
-    /// Returns the result of the callback, caching it in HttpContext.Features for the lifetime of the request.
-    /// Subsequent calls will retrieve the cached value.
-    /// Results are stored by type and therefore must be of a unique type.
-    /// </summary>
-    public static async Task<T> WithFeaturesCacheAsync<T>(this HttpContext httpContext, Func<Task<T>> callback)
+    /// <param name="httpContext"></param>
+    extension(HttpContext httpContext)
     {
```
eliykat, line 17: Ooh I wasn't aware of this new syntax - very nice.

eliykat (approved):

JimmyVo16 (approved):

PR #7727, opened 2026-05-27
refactor(server): move CloudRegion enum from Setup to Core

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38015

## 📔 Objective

Decouple CloudRegion from Bit.Setup.Enums so any project consuming Core can reference it without a reference back into the Setup utility. Updates the three downstream Setup references (Context.cs, Program.cs, Setup.csproj) to point at the new Bit.Core.Enums namespace. This migration is isolated from the rest of PM-38015 so SHOT can review just the pieces relevant to your team. Platform will be using this enum from Core in related in progress work.

### DIFF
--- a/src/Core/Enums/CloudRegion.cs
+++ b/src/Core/Enums/CloudRegion.cs
@@ -1,6 +1,6 @@
 ﻿using System.ComponentModel.DataAnnotations;
 
-namespace Bit.Setup.Enums;
+namespace Bit.Core.Enums;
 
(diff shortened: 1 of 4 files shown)


djsmith85 (approved):
Changes are looking good.

Initially concerned bringing in this massive dependency, but it's already a transient one, via Migrator.csproj.

mimartin12 (approved):

PR #7732, opened 2026-05-27
[PM-36951] Add Billing Plan Migration Cohorts Admin UI

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36951

## 📔 Objective

Adds the operator-facing Cohorts CRUD UI under the Tools menu of the Bitwarden Portal (`src/Admin`), gated by the `PM35215_BusinessPlanPriceMigration` feature flag. This is the workflow operators use to define, search, edit, and disable the Migration and Churn-only cohorts that drive the business-plan price migration loop introduced in PM-35215.

  **What this adds:**

 - Cohorts list page with name search, Previous/Next pagination, and a Clear search link that surfaces only when a filter is applied. Columns cover Migration Path, Type, coupon codes, and Pending/Scheduled/Migrated counts.
  - Create form scaffolded for the Migration / Churn-only cohort split, with cohort-type-aware cross-field validation (churn coupon required on Churn-only, proactive coupon disallowed on Churn-only, etc.).
  - Edit form composed via `EditCohortViewModel`, with a shared `_CohortFormFields` Razor partial driving Name and coupon inputs across Create and Edit. Active state lives on the form and is persisted through the same POST.
  - Migration-path lock: once a cohort has assignments that have left Pending, the migration-path field renders as static text and the controller ignores any submitted change. Lock decisions flow through a dedicated `GetCohortAssignmentStateQuery`.
  - Stripe coupon validation on save (proactive + churn), duplicate-name rejection, and a delete path that only succeeds when no assignments have left Pending.
  - New stored procedures: `OrganizationPlanMigrationCohort_SearchWithCounts`, `OrganizationPlanMigrationCohort_ReadByName`, `OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId`.

**Where this deviates from the mockups and why**

- The migration path dropdown was moved to the top of the form to enable the option of using a razor partial for the fields shared by the create and edit forms more easily but also so that the "locked path" shows more prominently at the top of the form when editing a locked cohort
- Added a visual lock indicator (see screenshots) when a cohort is locked and cannot change the migration path. (cohorts are "locked" when at least one organization assigned to the cohort has been migrated.
- Removed the "is active" checkbox from the create form because it defaults to OFF, and there's no real reason to activate a cohort before assignments are even loaded. 
- Churn only cohorts list the number of organizations that have activated the churn coupon in the "Migrated" column. The mockup had a "-" in that column, but it is data we are able to represent so it was added to add benefit.
- A "clear search" link was added next to the search box when the user is searching for something to enable an easy way to return to the full list.

## 📸 Screenshots

image

image

image

image

image


src/Admin/Billing/Views/OrganizationPlanMigrationCohorts/_CohortFormFields.cshtml
```diff
@@ -0,0 +1,34 @@
...
+
+<div class="mb-3">
+    <label asp-for="Name" class="form-label"></label>
+    <input asp-for="Name" class="form-control" placeholder="e.g., A1 (a) — Enterprise 2020 Annual" />
+    <span asp-validation-for="Name" class="text-danger"></span>
+    <div class="form-text">Unique, max 255 characters. Used as the join key in CSV uploads.</div>
+</div>
+
+<div class="mb-3">
+    <label asp-for="ProactiveDiscountCouponCode" class="form-label">
```
amorask-bitwarden, line 13: 💡 We've got this field labeled as `(optional)`, but for churn-only cohorts, it's not actually optional, it's forbidden. This is born out in the server-side validation here: https://github.com/bitwarden/server/blob/5ba25d4df80a219ec572fd54441760137a089f43/src/Admin/Billing/Models/OrganizationPlanMigrationCohorts/CohortFormModel.cs#L54-L60

Do you think we can set this one up more like the churn discount field below it where the optionality or display is driven by the script rendered in `Create.cshtml` and `Edit.cshtml`?

src/Admin/Billing/Views/OrganizationPlanMigrationCohorts/_CohortFormFields.cshtml
```diff
@@ -0,0 +1,34 @@
...
+    <span asp-validation-for="ProactiveDiscountCouponCode" class="text-danger"></span>
+    <div class="form-text">
+        Stripe coupon ID applied to Phase 2 of the SubscriptionSchedule at scheduling time. Validated against Stripe at save.
+    </div>
+</div>
+
+<div class="mb-3">
+    <label asp-for="ChurnDiscountCouponCode" class="form-label">
+        Churn discount coupon
+        <span id="churn-optional-label" class="text-muted">(optional)</span>
```
amorask-bitwarden: ⛏️ This really isn't a big deal because it can only happen when the cohort is already locked, but if you have a locked churn-only cohort, this is still going to render as `(optional`) despite it being required. Reason being you're going to early return here when that `<select>` box isn't visible. https://github.com/bitwarden/server/blob/5ba25d4df80a219ec572fd54441760137a089f43/src/Admin/Billing/Views/OrganizationPlanMigrationCohorts/Edit.cshtml#L93-L94

You could do something like this to get around it if you wanted to:
```cshtml
<span id="churn-optional-label" class="text-muted" @(Model.CohortType is CohortType.ChurnOnly ? "style=display:none" : "")>(optional)</span>
```

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            return true;
+        }
+
+        ModelState.AddModelError(nameof(CohortFormModel.Name), "A cohort with this name already exists.");
+        return false;
+    }
+
+    private async Task<bool> ValidateCouponsAsync(CohortFormModel model)
+    {
```
amorask-bitwarden, line 244: ⛏️ Could probably save 2 Stripe calls here if you compare the incoming from the `model` to what's currently saved on the cohort and don't validate if they're the same, but it would break your fast fail on any issues with previously saved coupons so your call.

amorask-bitwarden (dismissed):
Killer work here - this looks great. Just a couple ⛏️ / 💡 if you want to address them, but looks good to me.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
kdenney: Yeah the idea here was to have a safeguard and catch any coupons that are no longer valid to force the operator to solve the issue before being able to edit anything. I thought that was valuable enough to justify the round trips but thanks for the feedback!

src/Admin/Billing/Views/OrganizationPlanMigrationCohorts/_CohortFormFields.cshtml
kdenney: Good call! Fixed in latest push by actually writing out using razor on initial page load (and on postback page loads where validation fails) to hide the label.

src/Admin/Billing/Views/OrganizationPlanMigrationCohorts/_CohortFormFields.cshtml
kdenney: Great point! Thanks for this feedback. I've added razor "hidden" for initial and postback page loads and also javascript: both to hide the entire proactive row when the cohort is churn only. I added another screenshot to the PR description with this row hidden.

amorask-bitwarden (dismissed):

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_ReadByName.sql
```diff
@@ -0,0 +1,13 @@
+CREATE PROCEDURE [dbo].[OrganizationPlanMigrationCohort_ReadByName]
+    @Name NVARCHAR(255)
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT TOP 1
```
mkincaid-bw: The `[Name]` column in the OrganizationPlanMigrationCohort table has a unique constraint, so `TOP 1` is not needed.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_SearchWithCounts.sql
```diff
@@ -0,0 +1,47 @@
+CREATE PROCEDURE [dbo].[OrganizationPlanMigrationCohort_SearchWithCounts]
```
mkincaid-bw: This name (`_SearchWithCounts`) does not follow [our guidelines](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures). This proc should be named something like `OrganizationPlanMigrationCohort_ReadManyWithCountsByName`

mkincaid-bw: I also have some concerns about performance with this proc. I realize it's new but this query might not scale well. This line contains two SQL anti-patterns:

`(@Name IS NULL OR C.[Name] LIKE '%' + @Name + '%')`

1. `@Name IS NULL OR...` - SQL catch-all
2. `LIKE '%' + @Name + '%'` - This eliminates the use of the unique index on the Name column, forcing a table scan.

Also, this query applies pagination after aggregation, which will likely not scale well as the data grows. One way to alleviate that is to paginate before doing the aggregation:

```sql
CREATE PROCEDURE [dbo].[OrganizationPlanMigrationCohort_ReadManyWithCountsByName]
    @Name NVARCHAR(255) = NULL,
    @Skip INT,
    @Take INT
AS
BEGIN
    SET NOCOUNT ON

    ;WITH [PagedCohorts] AS (
        SELECT
            C.[Id], C.[Name], C.[MigrationPathId],
            C.[ProactiveDiscountCouponCode], C.[ChurnDiscountCouponCode],
            C.[IsActive], C.[CreationDate], C.[RevisionDate]
        FROM
            [dbo].[OrganizationPlanMigrationCohortView] C
        WHERE
            (@Name IS NULL OR C.[Name] LIKE '%' + @Name + '%')
        ORDER BY
            C.[CreationDate] DESC, C.[Id] ASC
        OFFSET @Skip ROWS FETCH NEXT @Take ROWS ONLY
    )
    SELECT
        COUNT(
            CASE
                WHEN P.[MigrationPathId] IS NOT NULL AND A.[Id] IS NOT NULL AND A.[ScheduledDate] IS NULL THEN 1
                WHEN P.[MigrationPathId] IS NULL AND A.[Id] IS NOT NULL AND A.[ChurnDiscountAppliedDate] IS NULL THEN 1
            END
            ) AS [Pending],
        COUNT(
                CASE
                WHEN P.[MigrationPathId] IS NOT NULL AND A.[ScheduledDate] IS NOT NULL AND A.[MigratedDate] IS NULL THEN 1
            END
            ) AS [Scheduled],
        COUNT(
            CASE
                WHEN P.[MigrationPathId] IS NOT NULL AND A.[MigratedDate] IS NOT NULL THEN 1
                WHEN P.[MigrationPathId] IS NULL AND A.[ChurnDiscountAppliedDate] IS NOT NULL THEN 1
            END
            ) AS [Migrated],
        P.[Id], P.[Name], P.[MigrationPathId],
        P.[ProactiveDiscountCouponCode], P.[ChurnDiscountCouponCode],
        P.[IsActive], P.[CreationDate], P.[RevisionDate]
    FROM
        [PagedCohorts] P
    LEFT JOIN 
        [dbo].[OrganizationPlanMigrationCohortAssignmentView] A ON A.[CohortId] = P.[Id]
    GROUP BY
        P.[Id], 
        P.[Name], 
        P.[MigrationPathId],
        P.[ProactiveDiscountCouponCode], 
        P.[ChurnDiscountCouponCode],
        P.[IsActive], 
        P.[CreationDate], 
        P.[RevisionDate]
    ORDER BY
        P.[CreationDate] DESC, 
        P.[Id] ASC
END
```

This won't fix the issue with the anti-patterns around the Name but will reduce the scan counts, potentially making the anti-patterns a non-issue.

We can also add the `ChurnDiscountAppliedDate` column to the existing index in the `OrganizationPlanMigrationCohortAssignment` table, which will eliminate a key lookup and further improve performance.
```sql
CREATE NONCLUSTERED INDEX [IX_OrganizationPlanMigrationCohortAssignment_CohortId_ScheduledDate_MigratedDate]
ON [dbo].[OrganizationPlanMigrationCohortAssignment] ([CohortId], [ScheduledDate], [MigratedDate])
INCLUDE ([ChurnDiscountAppliedDate])
WITH (DROP_EXISTING = ON);
```

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId.sql
```diff
@@ -0,0 +1,18 @@
+CREATE PROCEDURE [dbo].[OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId]
+    @CohortId UNIQUEIDENTIFIER
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT COUNT(1)
```
mkincaid-bw: ⛏️ Minor nick pick
```suggestion
    SELECT
        COUNT(1)
```

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId.sql
```diff
@@ -0,0 +1,18 @@
...
+    FROM
+        [dbo].[OrganizationPlanMigrationCohortAssignmentView] A
+        INNER JOIN [dbo].[OrganizationPlanMigrationCohortView] C ON C.[Id] = A.[CohortId]
```
mkincaid-bw: ⛏️ Minor nick pick
```suggestion
    INNER JOIN
        [dbo].[OrganizationPlanMigrationCohortView] C  ON C.[Id] = A.[CohortId]
```

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId.sql
```diff
@@ -0,0 +1,18 @@
+CREATE PROCEDURE [dbo].[OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId]
```
mkincaid-bw: This name (`_CountNonPendingByCohortId`) does not follow [our guidelines](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures). This proc should be named something like `OrganizationPlanMigrationCohortAssignment_ReadNonPendingCountByCohortId`

mkincaid-bw (changes requested):

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_SearchWithCounts.sql
kdenney: All great points; thanks for this feedback! I have added changes for these suggestions in the latest push.

Just FYI: the `@Name IS NULL OR...` clause is just to check if the search parameter passed in to the SP is NULL so that when the user is not searching by a name, it returns all rows by default. 

But let me know if there's a clean way that we can fix the table scan anti-pattern while still enabling the search by name with pagination feature; I'd be happy to make the change.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId.sql
kdenney: Fixed in latest push.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId.sql
kdenney: Fixed in latest push.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohortAssignment_CountNonPendingByCohortId.sql
kdenney: Fixed in latest push.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_SearchWithCounts.sql
kdenney: Fixed in latest push.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_ReadByName.sql
kdenney: Removed the top 1 in latest push.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            // The locked migration path view doesn't post a value for MigrationPathSelection.
+            // Restore from the persisted cohort so [Required] passes and the eventual
+            // ReplaceAsync writes back the unchanged path.
+            model.MigrationPathSelection = cohort.MigrationPathId switch
+            {
+                null => CohortFormModel.NoMigrationPath,
+                var pathId => ((byte)pathId).ToString(),
+            };
+        }
```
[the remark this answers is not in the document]
kdenney: Fixed in latest.

amorask-bitwarden (dismissed):

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_ReadManyWithCountsByName.sql
```diff
@@ -0,0 +1,62 @@
...
+        P.[MigrationPathId],
+        P.[ProactiveDiscountCouponCode],
+        P.[ChurnDiscountCouponCode],
+        P.[IsActive],
+        P.[CreationDate],
+        P.[RevisionDate]
+    FROM
+        [PagedCohorts] P
+        LEFT JOIN [dbo].[OrganizationPlanMigrationCohortAssignmentView] A
+            ON A.[CohortId] = P.[Id]
```
mkincaid-bw: ⛏️ `LEFT JOIN` should be aligned with `FROM`
```suggestion
    LEFT JOIN [dbo].[OrganizationPlanMigrationCohortAssignmentView] A
        ON A.[CohortId] = P.[Id]
```

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_ReadManyWithCountsByName.sql
```diff
@@ -0,0 +1,62 @@
...
+        P.[Id],
+        P.[Name],
+        P.[MigrationPathId],
+        P.[ProactiveDiscountCouponCode],
+        P.[ChurnDiscountCouponCode],
+        P.[IsActive],
+        P.[CreationDate],
+        P.[RevisionDate]
+    ORDER BY
+        P.[CreationDate] DESC, P.[Id] ASC
```
mkincaid-bw: ⛏️ These should be on their own line
```suggestion
        P.[CreationDate] DESC, 
        P.[Id] ASC
```

util/Migrator/DbScripts/2026-06-01_00_AddChurnDiscountAppliedDateToCohortAssignmentIndex.sql
```diff
@@ -0,0 +1,9 @@
+-- Add [ChurnDiscountAppliedDate] as an included column on the cohort assignment lookup index.
+-- This eliminates a key lookup when the cohort counts queries read churn-redemption state,
+-- improving performance of OrganizationPlanMigrationCohort_ReadManyWithCountsByName and
+-- OrganizationPlanMigrationCohortAssignment_ReadNonPendingCountByCohortId.
+CREATE NONCLUSTERED INDEX [IX_OrganizationPlanMigrationCohortAssignment_CohortId_ScheduledDate_MigratedDate]
+    ON [dbo].[OrganizationPlanMigrationCohortAssignment] ([CohortId] ASC, [ScheduledDate] ASC, [MigratedDate] ASC)
+    INCLUDE ([ChurnDiscountAppliedDate])
+    WITH (DROP_EXISTING = ON);
+GO
```
mkincaid-bw: This should be guarded just in case the index doesn't exist yet (DROP_EXISTING will fail in that case)
```suggestion
IF EXISTS (
    SELECT 
        NULL
    FROM 
        sys.indexes 
    WHERE 
        [name] = 'IX_OrganizationPlanMigrationCohortAssignment_CohortId_ScheduledDate_MigratedDate' 
        AND object_id = OBJECT_ID('[dbo].[OrganizationPlanMigrationCohortAssignment]')
)
BEGIN
    CREATE NONCLUSTERED INDEX [IX_OrganizationPlanMigrationCohortAssignment_CohortId_ScheduledDate_MigratedDate]
    ON [dbo].[OrganizationPlanMigrationCohortAssignment] ([CohortId] ASC, [ScheduledDate] ASC, [MigratedDate] ASC)
    INCLUDE ([ChurnDiscountAppliedDate])
    WITH (DROP_EXISTING = ON);
END
ELSE
BEGIN
    CREATE NONCLUSTERED INDEX [IX_OrganizationPlanMigrationCohortAssignment_CohortId_ScheduledDate_MigratedDate]
    ON [dbo].[OrganizationPlanMigrationCohortAssignment] ([CohortId] ASC, [ScheduledDate] ASC, [MigratedDate] ASC)
    INCLUDE ([ChurnDiscountAppliedDate]);
END
GO

```

mkincaid-bw (changes requested):
Couple minor things.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_ReadManyWithCountsByName.sql
kdenney: Fixed in latest push.

src/Sql/dbo/Billing/Stored Procedures/OrganizationPlanMigrationCohort_ReadManyWithCountsByName.sql
kdenney: Fixed in latest push.

util/Migrator/DbScripts/2026-06-01_00_AddChurnDiscountAppliedDateToCohortAssignmentIndex.sql
kdenney: Fixed in latest push.

mkincaid-bw (approved):

amorask-bitwarden (approved):

PR #7734, opened 2026-05-28
[PM-38165] Improve Data Protection error handling and constrain plaintext input

harr1424:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-38165

## 📔 Objective

This PR improves error handling surrounding the use of ASP.NET Data Protection on the `Send.Emails` property. Additionally, it limits plaintext content of `Emails` to 2,500 characters or less, so that protection yields encrypted values well below the 4,000 character limit.


src/Core/Tools/SendFeatures/Services/SendValidationService.cs
```diff
@@ -40,6 +40,27 @@ public SendValidationService(
...
+        // Emails may be sent as a plain-text string < 4000 chars during Send creation and subsequently protected using
+        // ASP.NET Data Protection with encrypted value violating 4000 char limit
+        if (!string.IsNullOrWhiteSpace(send.Emails))
+        {
+            // The "P|" prefix is a server-internal sentinel for Data-Protection-wrapped values.
+            // Clients must not submit a value starting with this prefix.
+            if (send.Emails.StartsWith(Constants.DatabaseFieldProtectedPrefix))
+            {
+                throw new BadRequestException("The Emails field contains an invalid character sequence.");
+            }
```
justindbaur, line 53: You don't need this check the issue isn't new emails being submitted with the `P|` prefix it was the short period of time where data protection wasn't applied to the column so there was a _chance_ that the prefix was mixed between columns that were protected and weren't protected. Since all brand new stuff will be protected it is safe to have that prefix. The issue was reading from the database.

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -214,20 +222,31 @@ private async Task ProtectDataAndSaveAsync(Send send, Func<Task> saveTask)
...
-        send.Emails = string.Concat(Constants.DatabaseFieldProtectedPrefix,
-            _dataProtector.Protect(send.Emails));
+            return true;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "An exception occurred while protecting Emauls for Send {SendId}", send.Id);
+            return false;
+        }
     }
```
justindbaur: Don't do this, there are no known issues of protecting the data and this would only exist to cover up an issue if it did occur. I think now that we know the the issue with unprotecting the data is I would change the `try/catch` from the unprotect call as well and rethrow the exception after logging the send id that caused the issue

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -238,9 +257,10 @@ private bool UnprotectData(Send send)
                 send.Emails.Substring(Constants.DatabaseFieldProtectedPrefix.Length));
             return true;
         }
-        catch (CryptographicException ex)
+
+        catch (Exception ex)
         {
-            _logger.LogWarning(ex, "Failed to unprotect Emails for Send {SendId}.", send.Id);
+            _logger.LogWarning(ex, "An exception occurred while unprotecting Emails for Send {SendId}", send.Id);
             return false;
```
justindbaur: Keep this as `CryptographicException` the more specific exception type the better. I also think this method should still throw on failures as to keep them from being covered up. The important part of fixing the delete send jobs was just not unprotecting the sends for deletion. The fact that this bug lead to a 500 error for a user is likely the only reason we were able to uncover it. If we just logged a warning and filtered it out it could have been much more easily missed.

```
if (send.Emails.Length == 4000)
{
    _logger.LogError("Emails column for Send {SendId} is max length and may have been truncated.", send.Id);
    throw;
}
else
{
    _logger.LogError("Failed to unprotect Emails for Send {SendId}", send.Id);
    throw;
}
```

test/Infrastructure.Dapper.Test/Tools/SendRepositoryTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        public IDataProtector CreateProtector(string purpose) => this;
+
+        public byte[] Protect(byte[] plaintext) =>
+            throw new CryptographicException("Test-induced Protect failure");
+
+        public byte[] Unprotect(byte[] protectedData) =>
+            throw new CryptographicException("Test-induced Unprotect failure");
+    }
+}
```
justindbaur: I don't think this test helps us, this is asserting that failures to Protect lead to a different exception but we never encountered a failure during protection. I would recommend writing a true regression test in the `test/Infrastructure.IntegrationTest` project that tries to save a long email column and have it get truncated. After all my recommended changes the before and after will both lead to a cryptographic exception but when it was truncated it will lead to a different log message that you can assert on. 

^ That might need a minor change to the test project to include a test logger that is easy to assert on so if you have troubles with that please reach out and I can help make that happen.

justindbaur (changes requested):

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
harr1424: I agree, but decided to retain this guard in anticipation of a HackerOne report and future edge case: 

If the plain-text input begins with `P|`, then [logic within SendRepository](https://github.com/bitwarden/server/blob/bd1a91337bcaf657ee7019291fa2e1b41a9edb80/src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs#L229) will assume it has been protected using Data Protection, which means that subsequently this string will fail decryption.

mcamirault (dismissed):

test/Infrastructure.IntegrationTest/CapturingLoggerProvider.cs
justindbaur, line 1: You can use this package to do the same thing: https://github.com/bitwarden/server/blob/9b3e01e09c233f8a6b302e03d86beb7cf5d9d306/test/Identity.Test/Identity.Test.csproj#L10

test/Infrastructure.IntegrationTest/Tools/SendRepositoryTests.cs
```diff
@@ -64,4 +68,41 @@ public async Task GetByDeletionDateAsync_Works(ISendRepository sendRepository)
...
+        ISendRepository sendRepository,
+        Database database,
+        CapturingLoggerProvider logs)
+    {
+        // Only the Dapper SendRepository (SQL Server) carries the truncation-aware
+        // error message. EF providers use a different implementation.
+        if (database.Type != SupportedDatabaseProviders.SqlServer || database.UseEf)
+        {
+            return;
+        }
```
justindbaur: I think we should try and update the EF implementation to act the same.

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
```diff
@@ -184,11 +189,19 @@ INNER JOIN
...
             {
-                _ = UnprotectData(send);
+                try
+                {
+                    UnprotectData(send);
+                }
+                catch (CryptographicException)
+                {
+                    // Logged by UnprotectData; intentionally swallowed here.
+                }
```
justindbaur: What is the desire to swallow here? I would not consider it a successful rotation is the data is not able to be round-tripped.

src/Core/Tools/SendFeatures/Services/SendValidationService.cs
justindbaur: I did not realize you conditionally to the protection based on if the prefix is there. As long as there was is no way to retrieve a protected send (exluding the one for deletion which should be documented) then you should be able to just unconditionally protect without checking the prefix. 

But if you're willing to accept not being able to support that email prefix this is fine.

harr1424: ~~I am going to extend the data protection to the EF implementation [as suggested](https://github.com/bitwarden/server/pull/7734#discussion_r3334164379) and so I think it's best to retain this so that those changes are backwards compatible.~~

src/Infrastructure.Dapper/Tools/Repositories/SendRepository.cs
harr1424: The idea is that since the `Emails` field is not actually used during the rotation (similar to how it wasn't used during the deletion job), that if a `CryptographicException` occurred it could almost certainly be attributed to the problematic truncation in `Emails`, which has been fixed.  Like you mentioned earlier, we probably shouldn't be swallowing any errors in this process. I'll carve this out.

src/Infrastructure.EntityFramework/Repositories/DatabaseContext.cs
```diff
@@ -153,7 +152,6 @@ protected override void OnModelCreating(ModelBuilder builder)
         var dataProtectionConverter = new DataProtectionConverter(dataProtector);
         eUser.Property(c => c.Key).HasConversion(dataProtectionConverter);
         eUser.Property(c => c.MasterPassword).HasConversion(dataProtectionConverter);
-        eSend.Property(c => c.Emails).HasConversion(dataProtectionConverter);
```
harr1424, line 156: I've moved `Emails` logic outside of the converter and into the repository to better align the EF implementation with the Dapper implementation (logging & `GetManyByDeletionDateAsync`).

justindbaur (dismissed):
Looks good, thank you for iterating on this with me!

src/Infrastructure.EntityFramework/Tools/Repositories/SendRepository.cs
```diff
@@ -50,13 +79,24 @@ public SendRepository(IServiceScopeFactory serviceScopeFactory, IMapper mapper)
...
+    }
+
     /// <inheritdoc />
     public async Task<ICollection<Core.Tools.Entities.Send>> GetManyByDeletionDateAsync(DateTime deletionDateBefore)
     {
         using (var scope = ServiceScopeFactory.CreateScope())
         {
             var dbContext = GetDatabaseContext(scope);
             var results = await dbContext.Sends.Where(s => s.DeletionDate < deletionDateBefore).ToListAsync();
+            // Don't unprotect here DeleteSendsJob needs suceed regardless of protected values
```
mcamirault: ⛏️ The most minor of nits...a typo in a comment :). This should be `to succeed` but is not blocking in any way, shape, or form

mcamirault (dismissed):

mcamirault (approved):

PR #7736, opened 2026-05-28
docs: expand CACHING.md (non-Redis L2, sizing/cost, SRE coordination)

withinfocus:
## 🎟️ Tracking

Follow-up to #6682, expanded in review.

## 📔 Objective

#6682 added support for using any `IDistributedCache` (Cosmos DB, SQL Server, EF) as the L2 store under `ExtendedCache`, but the docs in `src/Core/Utilities/CACHING.md` only got a one-line setting rename. This PR closes that gap and, in the course of review, expands the doc to cover two adjacent areas that were also missing or hand-wavy: sizing/throughput/cost, and coordination with the infrastructure team.

### Non-Redis L2 documentation (original scope)

- Overview table and deployment-modes list reflect Cosmos / SQL / EF as valid L2 backends.
- New Option 5 example wires `ExtendedCache` with a non-Redis L2, including the keyed-alias pattern for pairing with the `"persistent"` Cosmos cache.
- "Backend Configuration" subsection rewritten as shared-mode vs keyed-mode resolution, with a prominent backplane caveat (cross-instance L1 invalidation is Redis-only).
- "When NOT to Use" explains *why* `ExtendedCache` may not fit long-lived data (no backplane → cross-instance L1 staleness) instead of the prior blanket "days/weeks → persistent cache."
- New worked example: **Long-Lived Per-Tenant Computed Aggregates** for the `ExtendedCache` + Cosmos pairing, exercising every FusionCache feature honestly.
- Decision tree and Configuration Priority table updated to match.

### Review feedback applied

- `L1`/`L2` vocabulary defined once in the ExtendedCache intro and dropped from the decision tree, which was using it before introduction.
- Cosmos TTL precision: `Duration` → `AbsoluteExpirationRelativeToNow` → per-document `ttl`; container `DefaultTimeToLive` must be enabled for it to apply.
- RU profile / eager-refresh tradeoff called out for the `ExtendedCache` + Cosmos pairing.
- Note that `ExtendedCache` prefixes every key with `cacheName:`, so aliased consumers cannot collide with raw `"persistent"` namespace.
- Fixed Option 5 inline comment that incorrectly implied `AddDistributedCache` registers Cosmos under `"persistent"` in self-hosted (it aliases to the unnamed `IDistributedCache` there).
- Footnoted the Configuration Priority table to point at the opt-in Cosmos pairing path.

### Sizing, throughput, and cost (new)

A new section after the overview table acknowledges that the decision tree silently assumed the dataset fit in the backends it pointed to. Covers:

- **Working-set size**: L1 bounded by process heap; per-backend L2 ceilings (Redis memory cap and shared-instance eviction, Cosmos RU economics, SQL/EF table growth).
- **Read rate**: when L1 absorbs reads vs. becomes pure overhead at high cardinality with low per-key reuse.
- **Write rate**: backplane saturation under sustained writes; eager-refresh's per-key background write cadence.
- **Cost shape** per backend (fixed-ceiling vs. per-RU vs. shared-DB load).
- A 3-question check and a workload-vs-tree-vs-reality table for when sizing overrides the tree.

Decision tree adds gates for working-set size and sustained-high write rate. ExtendedCache Cons and "When NOT to Use" expanded with the two main outs: working sets too large for L1, and sustained-high write rates that saturate the Redis backplane. SSO grants and Org/Provider Abilities "Why" lists now include the size/rate profile that makes `ExtendedCache` appropriate there.

### Coordinating with Infrastructure (new)

A new section calling out that `services.AddExtendedCache(...)` does not stand up the runtime. Covers:

- Loop in SRE / infrastructure early — they have visibility into budgets, utilization, and incident history the code does not.
- Defaults are not production configuration. Uses `CreateIfNotExists = false` + container-level `DefaultTimeToLive` as the worked example.
- The shared application Redis already has tenants — a noisy new cache can starve unrelated callers.
- New backends need provisioning, dashboards, secrets, cost approval — not just code.
- Self-hosted reality differs from cloud; "works in cloud" ≠ "works in every customer's environment."

No code changes — `src/Core/Utilities/CACHING.md` only. Net diff is ~+230 / -45.

## 📸 Screenshots

N/A — docs-only.

### DIFF
--- a/src/Core/Utilities/CACHING.md
+++ b/src/Core/Utilities/CACHING.md
@@ -6,7 +6,10 @@ Caching options available in Bitwarden's server. The server uses multiple cachin
 
 ## Choosing a Caching Option
 
-Use this decision tree to identify the appropriate caching option for your feature. Before you start, skim [Sizing, Throughput, and Cost](#sizing-throughput-and-cost) below — the tree assumes the dataset fits in the backends it points to, and if your working set, read rate, or write rate is extreme along any axis, the sizing constraints can override the tree's answer.
+Use this decision tree to identify the appropriate caching option for your feature. Before you start, skim two sections below:
+
+- [Sizing, Throughput, and Cost](#sizing-throughput-and-cost) — the tree assumes the dataset fits in the backends it points to. If your working set, read rate, or write rate is extreme along any axis, sizing can override the tree's answer.
+- [Coordinating with Infrastructure](#coordinating-with-infrastructure) — caching is shared infrastructure owned by SRE / infra maintainers. Don't assume the defaults in this doc are production-ready or that the runtime can absorb a new workload without coordination.
 
 ```
 Does your data need to be shared across all instances in a horizontally-scaled deployment?
@@ -113,6 +116,22 @@ If any answer is "no", move down the decision tree to a less-feature-rich option
 | 5 K org abilities, read on every request                 | `ExtendedCache`                | `ExtendedCache` (as the tree says) — fits in L1, read-heavy                       |
 | 50 K queue jobs, written and removed within seconds      | `ExtendedCache` w/ backplane   | `IDistributedCache` (default) — backplane traffic would dominate                  |
 | Per-tenant analytics, 100 K orgs, refreshed every 6 h    | `ExtendedCache` paired w/ Cosmos | `ExtendedCache` paired w/ Cosmos — L1 holds the hot subset, L2 holds the long tail |
+
+---
+
+## Coordinating with Infrastructure
+
+Caching is shared infrastructure. The decision tree and sizing section above help you pick the right option for your code, but the runtime that option lands on — Redis cluster sizing, Cosmos container provisioning, RU budgets, SQL Cache table maintenance, monitoring, alerting — is owned by SRE and the infrastructure maintainers, and a `services.AddExtendedCache(...)` call does **not** stand any of it up.
+
+Before standing up a new cache, or significantly changing the load profile of an existing one:
+
+- **Loop in SRE / infrastructure early.** Even a "small" cache adds a new operational concern: memory pressure, RU consumption, eviction churn, monitoring noise. SRE has visibility into existing budgets, utilization, and incident history that this codebase does not. The cost of a five-minute conversation now is much smaller than an unplanned-incident postmortem later.
+- **Do not assume defaults are production-ready.** Connection strings, cluster sizes, container TTL, eviction policies, and RU provisioning are all configured outside this repo. `AddDistributedCache` registers `CosmosCache` with `CreateIfNotExists = false` precisely because the `cache.default` container is provisioned out-of-band — and `DefaultTimeToLive` must be enabled on that container for per-document TTL to take effect at all.
+- **Shared infrastructure has tenants you don't see.** The shared application Redis is already serving the rest of the platform. A new cache that parks a large working set, sustained write rate, or noisy backplane traffic on it can starve unrelated callers. Treat shared-Redis additions the same way you would a new query against a shared database — they need awareness, not just code.
+- **New backends need provisioning, not just code.** Standing up dedicated Redis, a new Cosmos container, or a separately scaled SQL Cache table involves environments, secrets, monitoring, dashboards, and cost — none of which the registration call covers. Plan that work alongside the code change, not after it merges.
+- **Self-hosted reality differs from cloud.** Cloud has Redis + Cosmos baseline; self-hosted deployments may have only Redis, only SQL Server, or neither. The fallback paths described in [Backend Configuration](#backend-configuration) compile and run, but "works in cloud" does not imply "works in every customer's environment." Verify the fallback actually meets the requirement.
+
+If you are unsure whether your workload needs SRE involvement, ask. The decision tree gives you a code answer; the people running the infrastructure give you the production answer.
 
 ---
 
--- a/src/Core/Utilities/CACHING.md
+++ b/src/Core/Utilities/CACHING.md
@@ -15,10 +15,12 @@ Does your data need to be shared across all instances in a horizontally-scaled d
 │   Do you need long-term persistence with TTL (days/weeks)?
 │   ├─ YES
 │   │   │
-│   │   Is cross-instance L1 staleness acceptable?
+│   │   Is it OK for a write on one instance to take a while to reach the
+│   │   in-memory caches on other instances (no cross-instance invalidation)?
 │   │   ├─ NO → Use `IDistributedCache` with persistent keyed service (every read hits the store)
-│   │   └─ YES → Use `ExtendedCache` paired with the persistent cache (L1 memory + Cosmos/SQL/EF L2)
-│   │             (see "Pairing ExtendedCache with Cosmos DB" below)
+│   │   └─ YES → Use `ExtendedCache` paired with the persistent cache
+│   │             (in-memory cache + Cosmos/SQL/EF distributed cache —
+│   │              see "Pairing ExtendedCache with Cosmos DB" below)
 │   │
 │   └─ NO → Use `ExtendedCache`
 │       │
@@ -58,6 +60,8 @@ Does your data need to be shared across all instances in a horizontally-scaled d
 ## `ExtendedCache`
 
 `ExtendedCache` is a wrapper around [FusionCache](https://github.com/ZiggyCreatures/FusionCache) that provides a simple way to register **named, isolated caches** with sensible defaults. The goal is to make it trivial for each subsystem or feature to have its own cache - with optional distributed caching and backplane support - without repeatedly wiring up FusionCache, Redis, and related infrastructure.
+
+> **Vocabulary**: throughout this section, **L1** refers to FusionCache's in-process memory cache, and **L2** refers to the configured `IDistributedCache` (Redis, Cosmos, SQL, EF, or none). FusionCache reads from L1 first, falls back to L2, and writes through to both.
 
 Each named cache automatically receives:
 
@@ -173,7 +177,10 @@ services.AddExtendedCache("MyFeatureCache", globalSettings, new GlobalSettings.E
 // Keyed mode: register an IDistributedCache under the cache name yourself, then call
 // AddExtendedCache. This is how to pair ExtendedCache with the "persistent" (Cosmos)
 // keyed service.
-services.AddDistributedCache(globalSettings); // registers the "persistent" keyed Cosmos cache
+services.AddDistributedCache(globalSettings); // registers the "persistent" keyed cache —
+                                              // Cosmos in cloud (when Cosmos.ConnectionString is set),
+                                              // aliased to the unnamed IDistributedCache
+                                              // (Redis / SQL / EF) in self-hosted
 services.AddKeyedSingleton<IDistributedCache>(
     "MyLongLivedCache",
     (sp, _) => sp.GetRequiredKeyedService<IDistributedCache>("persistent"));
@@ -187,6 +194,9 @@ services.AddExtendedCache("MyLongLivedCache", globalSettings, new GlobalSettings
 // - L1 (memory) + L2 (Cosmos/SQL/EF) caching with stampede protection, eager refresh, fail-safe
 // - NO backplane (Redis-only). L1 entries on other instances will not be invalidated on write —
 //   they expire on their own TTL. See "When NOT to Use" for the staleness tradeoff.
+// - Keys are namespaced from other "persistent" cache consumers: ExtendedCache prefixes every
+//   key with "MyLongLivedCache:", so entries cannot collide with the raw "persistent" namespace
+//   used by direct AddDistributedCache consumers (payment workflow, OAuth grants).
 ```
 
 #### 2. Inject and use the cache:
@@ -293,7 +303,13 @@ public class SsoAuthorizationService
 
 #### Pairing `ExtendedCache` with Cosmos DB
 
-Cosmos DB is registered by `AddDistributedCache` as the keyed `"persistent"` `IDistributedCache` in cloud deployments. To use it as L2 under `ExtendedCache`, register an alias under your cache name and use keyed mode without a Redis connection string (see Option 5 above). This gives you L1 memory + Cosmos L2 with Cosmos's automatic container-level TTL, but no cross-instance L1 invalidation.
+Cosmos DB is registered by `AddDistributedCache` as the keyed `"persistent"` `IDistributedCache` in cloud deployments. To use it as L2 under `ExtendedCache`, register an alias under your cache name and use keyed mode without a Redis connection string (see Option 5 above). This gives you L1 memory + Cosmos L2.
+
+**TTL.** FusionCache's `Duration` (and any per-call `SetDuration`) is translated to `DistributedCacheEntryOptions.AbsoluteExpirationRelativeToNow` and written by `CosmosCache` as a per-document `ttl` on each Cosmos item — that's the authoritative expiry per entry. The Cosmos container must have `DefaultTimeToLive` enabled (`-1` or `>0`) for per-document `ttl` to apply at all; the container value also serves as a default for items that omit `ttl`, but FusionCache always sets a per-item value so the container value never caps it.
+
+**RU profile and the eager-refresh tradeoff.** L1 absorbs most reads, which is the win: hot keys see roughly one Cosmos point-read per process per TTL window instead of one per request. The cost is on the write side — `EagerRefreshThreshold` (default `0.9f`) produces a steady cadence of background factory calls plus L2 writes (~1 per hot key per `Duration` window) that raw `CosmosCache` does not have. If your access pattern doesn't benefit from background refresh (e.g. cold keys, low read concurrency), set `EagerRefreshThreshold = null` to disable it.
+
+Cross-instance L1 invalidation is not provided — see the backplane caveat above.
 
 ### Specific Example: Organization/Provider Abilities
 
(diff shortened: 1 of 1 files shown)


src/Core/Utilities/CACHING.md
```diff
@@ -239,20 +275,25 @@ public class SsoAuthorizationService
...
 
-1. Redis (if configured in `appsettings.json`)
-2. SQL Server / EF Cache (if `IDistributedCache` is registered and no Redis)
-3. Memory-only (default fallback)
+#### Pairing `ExtendedCache` with Cosmos DB
 
-> **Note**: ExtendedCache works seamlessly with any `IDistributedCache` backend. In self-hosted scenarios without Redis, you can configure ExtendedCache to use SQL Server or Entity Framework cache as its distributed layer. This provides local memory caching in front of the database cache, with the option to add Redis later if needed. You won't get the backplane (cross-instance invalidation) without Redis, but you still get stampede protection, eager refresh, and fail-safe mode.
+Cosmos DB is registered by `AddDistributedCache` as the keyed `"persistent"` `IDistributedCache` in cloud deployments. To use it as L2 under `ExtendedCache`, register an alias under your cache name and use keyed mode without a Redis connection string (see Option 5 above). This gives you L1 memory + Cosmos L2 with Cosmos's automatic container-level TTL, but no cross-instance L1 invalidation.
```
JaredSnider-Bitwarden: Claude 🎨:   
This subsection covers L1 + TTL but doesn't touch on the RU cost shift, which is a major reason a reader weighs ExtendedCache+Cosmos against raw CosmosCache. Two specifics:

- L1 absorbs most reads → near-zero Cosmos point-reads on hot keys (the win).
- Eager refresh (EagerRefreshThreshold defaults to 0.9f) produces a steady cadence of background Cosmos writes + DB factory calls per key/TTL window that raw CosmosCache doesn't have.

Suggested addition after the existing sentence:

RU profile: L1 absorbs most reads, reducing Cosmos point-read traffic compared to using CosmosCache directly. Eager refresh produces a steady cadence of background writes (~1 per key per TTL window) and background factory calls against your backing store — disable with EagerRefreshThreshold = null if your access pattern doesn't benefit.

src/Core/Utilities/CACHING.md
```diff
@@ -13,12 +13,19 @@ Does your data need to be shared across all instances in a horizontally-scaled d
 ├─ YES
 │   │
 │   Do you need long-term persistence with TTL (days/weeks)?
-│   ├─ YES → Use `IDistributedCache` with persistent keyed service
+│   ├─ YES
+│   │   │
+│   │   Is cross-instance L1 staleness acceptable?
```
JaredSnider-Bitwarden: 🎨 : `L1` and `L2` first appear in the decision tree at lines 18 and 20, ~30 lines before `FusionCache` is introduced (line 51) and before the doc establishes any layered-cache vocabulary. A reader walking top-to-bottom meets the shorthand without context.

Recommended fix comes in two parts:

1. Decision tree — drop the shorthand. Reword lines 18 and 20 to match the doc's existing "in-memory cache" / "distributed cache" phrasing. The decision tree is the entry point and should read with zero vocabulary debt.
2. `ExtendedCache` section — define once. When FusionCache is introduced around line 53, add one line establishing the L1/L2 convention so the remaining 10 uses (Option 5, Backend Configuration, backplane caveat, pairing subsection, When NOT to Use) have a reference point.

Something like:
Throughout this section, `L1` refers to FusionCache's in-process memory cache and `L2` refers to the configured IDistributedCache (Redis, Cosmos, SQL, EF, or memory).

src/Core/Utilities/CACHING.md
JaredSnider-Bitwarden: Claude 🎨:  "Cosmos's automatic container-level TTL" could be more precise here — Cosmos has two TTL knobs, and FusionCache actually controls the per-item one:

- **Per-document TTL** (`ttl` field on each item) — what FusionCache writes: `Duration` → `DistributedCacheEntryOptions.AbsoluteExpirationRelativeToNow` → Cosmos document `ttl`. See `src/Core/Utilities/ExtendedCacheServiceCollectionExtensions.cs:61`.
- **Container `DefaultTimeToLive`** — a container-level *gate* that must be enabled (`-1` or `>0`) for per-document TTL to apply at all. It also serves as a *default* for items that omit `ttl`. It's not a cap on per-item values; per-item `ttl` always wins.

The current phrasing reads like the container itself is enforcing expiry, which makes it easy to miss that the actual TTL boundary is set by FusionCache.

Suggested rewrite of the last sentence:

> This gives you L1 memory + Cosmos L2. FusionCache's `Duration` is written as a per-document `ttl` on each Cosmos item, so it's the authoritative expiry per entry (the container must have `DefaultTimeToLive` enabled for per-document TTL to apply). Cross-instance L1 invalidation is not provided.

Related: the pre-existing note in the Cosmos DB backend section at `src/Core/Utilities/CACHING.md:617` — "TTL: Cosmos DB container-level TTL (automatic cleanup, no scanning required)" — has the same shape.

src/Core/Utilities/CACHING.md
```diff
@@ -151,6 +159,34 @@ services.AddExtendedCache("SpecializedCache", globalSettings, new GlobalSettings
...
+});
+
+// Keyed mode: register an IDistributedCache under the cache name yourself, then call
+// AddExtendedCache. This is how to pair ExtendedCache with the "persistent" (Cosmos)
+// keyed service.
+services.AddDistributedCache(globalSettings); // registers the "persistent" keyed Cosmos cache
+services.AddKeyedSingleton<IDistributedCache>(
+    "MyLongLivedCache",
+    (sp, _) => sp.GetRequiredKeyedService<IDistributedCache>("persistent"));
+services.AddExtendedCache("MyLongLivedCache", globalSettings, new GlobalSettings.ExtendedCacheSettings
```
JaredSnider-Bitwarden, line 267: Claude 🎨:   
Worth surfacing here that the alias points your ExtendedCache and other direct "persistent" consumers (e.g. payment workflow, OAuth grants) at the same Cosmos container. A reader will reasonably wonder whether keys collide.

They don't — ExtendedCache prefixes every key with `cacheName`: (see src/Core/Utilities/ExtendedCacheServiceCollectionExtensions.cs:53), so MyLongLivedCache entries are naturally namespaced from the raw "persistent" keys. But the doc never says so, leaving the reader to either figure it out or get nervous about it.

Suggested one-liner to add to the Option 5 trailing comment block (around line 187):

Keys are isolated from other "persistent" cache consumers — ExtendedCache prefixes every key with cacheName:, so MyLongLivedCache entries cannot collide with the raw "persistent" namespace shared by direct AddDistributedCache consumers.

src/Core/Utilities/CACHING.md
```diff
@@ -151,6 +159,34 @@ services.AddExtendedCache("SpecializedCache", globalSettings, new GlobalSettings
...
+services.AddExtendedCache("MyFeatureCache", globalSettings, new GlobalSettings.ExtendedCacheSettings
+{
+    UseSharedDistributedCache = true,
+    // No Redis.ConnectionString in globalSettings.DistributedCache.Redis
+});
+
+// Keyed mode: register an IDistributedCache under the cache name yourself, then call
+// AddExtendedCache. This is how to pair ExtendedCache with the "persistent" (Cosmos)
+// keyed service.
+services.AddDistributedCache(globalSettings); // registers the "persistent" keyed Cosmos cache
```
JaredSnider-Bitwarden: Claude 🎨:   
This inline comment holds in cloud but is misleading in self-hosted. AddDistributedCache only registers CosmosCache under "persistent" when DistributedCache.Cosmos.ConnectionString is set — see src/SharedWeb/Utilities/ServiceCollectionExtensions.cs:788. In self-hosted (no Cosmos string), the "persistent" keyed service aliases the unnamed IDistributedCache instead, at src/SharedWeb/Utilities/ServiceCollectionExtensions.cs:799.

The Option 5 pattern still works in self-hosted — the alias just resolves to Redis / SQL / EF rather than Cosmos. But a self-hosted reader following this recipe will reasonably expect Cosmos and be confused about what they're actually getting.

Suggested rewrite of the inline comment:

services.AddDistributedCache(globalSettings); // registers the "persistent" keyed cache — Cosmos in cloud, aliased to the unnamed IDistributedCache (Redis / SQL / EF) in self-hosted

src/Core/Utilities/CACHING.md
JaredSnider-Bitwarden, line 458: Claude 🎨:   
The doc has four worked Specific Example sections — two for ExtendedCache+Redis (SSO grants at line 237, Org/Provider Abilities at 298) and two for raw Cosmos (Payment Workflow at 484, OAuth Grants at 530). The new ExtendedCache+Cosmos pairing has no worked example, leaving the matrix incomplete on the very option this PR is adding.

A natural insertion point is right after the Organization/Provider Abilities bullets at line 361, before "When NOT to Use" at line 363. Keeps both Redis examples first and the Cosmos example next, grouped by backend.

Drafted content below uses a hypothetical OrganizationAnalyticsSnapshot service (same convention as the existing SSO grants example, which uses fictional types). The use case exercises every FusionCache feature honestly — high per-key read frequency, high cardinality, expensive factory, tolerable staleness — so the reader sees when the pairing actually pays off:

---

### Specific Example: Long-Lived Per-Tenant Computed Aggregates

Precomputed per-organization data — e.g. dashboard analytics, materialized usage summaries, or rolled-up compliance state — is expensive to compute, read many times per dashboard session, and stable for hours. Pairing `ExtendedCache` with the persistent Cosmos cache puts an L1 memory cache in front of a durable Cosmos L2, so each instance amortizes the first recomputation across many subsequent reads, and snapshots survive process restarts.

```csharp
// Registration — aliases the keyed "persistent" cache (Cosmos in cloud) under the cache name
services.AddDistributedCache(globalSettings);
services.AddKeyedSingleton<IDistributedCache>(
    "OrganizationAnalytics",
    (sp, _) => sp.GetRequiredKeyedService<IDistributedCache>("persistent"));

services.AddExtendedCache("OrganizationAnalytics", globalSettings, new GlobalSettings.ExtendedCacheSettings
{
    UseSharedDistributedCache = false,
    Duration = TimeSpan.FromHours(6),
    EagerRefreshThreshold = 0.9f,                  // refresh in background at 90% of TTL
    IsFailSafeEnabled = true,
    FailSafeMaxDuration = TimeSpan.FromHours(24),  // serve stale up to 24h on factory failures
});

public class OrganizationAnalyticsService
{
    private readonly IFusionCache _cache;
    private readonly IOrganizationAnalyticsRepository _repository;

    public OrganizationAnalyticsService(
        [FromKeyedServices("OrganizationAnalytics")] IFusionCache cache,
        IOrganizationAnalyticsRepository repository)
    {
        _cache = cache;
        _repository = repository;
    }

    public async Task<OrganizationAnalyticsSnapshot> GetSnapshotAsync(Guid organizationId)
    {
        return await _cache.GetOrSetAsync<OrganizationAnalyticsSnapshot>(
            $"org:{organizationId}:analytics-snapshot",
            async _ => await _repository.ComputeSnapshotAsync(organizationId)
        );
    }

    public async Task InvalidateAsync(Guid organizationId)
    {
        // No backplane on Cosmos — this clears the current instance's L1 and removes the
        // entry from L2. Other instances continue serving their L1 entries until TTL.
        await _cache.RemoveAsync($"org:{organizationId}:analytics-snapshot");
    }
}
```

**Why pair `ExtendedCache` with Cosmos for this pattern:**

- **High per-key read frequency**: Each organization's snapshot is read repeatedly during a dashboard session (multiple widgets, drill-downs, refreshes). L1 absorbs these reads after the first hit per instance, so Cosmos sees roughly one point-read per process per key per TTL window.
- **High key cardinality**: One snapshot per organization. The total dataset can grow large enough that an in-process memory-only cache cannot hold it across instances — durable Cosmos L2 is what makes the L1 layer viable as a working set.
- **Cosmos durability across deploys**: Snapshots remain valid for hours. A process restart does not flush L2, so freshly-started instances warm L1 from Cosmos rather than triggering a thundering herd of recomputation.
- **Expensive factory**: Computing a snapshot involves joins across multiple tables. Stampede protection coalesces concurrent dashboard loads after a TTL boundary into a single recomputation, not N.
- **Eager refresh masks TTL boundaries**: At 90% of `Duration`, the next read returns immediately from L1 while a background factory call refreshes both L1 and L2. Users never observe hard-miss latency at the boundary.
- **Fail-safe keeps dashboards working through transient DB issues**: If the repository call errors, recent stale data is served instead of a user-facing failure.
- **Cross-instance staleness is acceptable**: Snapshot data may lag by minutes after an `InvalidateAsync` on another instance. Without a backplane, the stale L1 entry on instance B persists until its TTL expires — fine for analytics, not appropriate for transactional state.

src/Core/Utilities/CACHING.md
```diff
@@ -815,12 +856,12 @@ public class MyService
...
+| -------------------------------------- | ------------------------------------------------ | -------------------------------------------- | --------------------------------------------------------- |
+| **ExtendedCache**                      | Redis → registered `IDistributedCache` → Memory  | Redis → registered `IDistributedCache` (SQL/EF) → Memory | `GlobalSettings.DistributedCache.Redis.ConnectionString` + any pre-registered `IDistributedCache` |
+| **IDistributedCache** (default)        | Redis                                            | Redis → SQL → EF                             | `GlobalSettings.DistributedCache.Redis.ConnectionString`  |
+| **IDistributedCache** (`"persistent"`) | Cosmos → Redis                                   | Redis → SQL → EF                             | `GlobalSettings.DistributedCache.Cosmos.ConnectionString` |
+| **OAuth Grants** (long-lived)          | Persistent cache (Cosmos)                        | `IGrantRepository` (SQL/EF)                  | Various (see above)                                       |
```
JaredSnider-Bitwarden, line 1024: Claude 🎨:   
A reader coming from the new "Pairing ExtendedCache with Cosmos DB" subsection will likely scroll to this matrix expecting to see Cosmos in the ExtendedCache row — but Cosmos only appears in the IDistributedCache (persistent) row. Cosmos is reachable from ExtendedCache only via the opt-in keyed-mode alias from Option 5, which isn't represented in this matrix anywhere.

Is the matrix intentionally limited to the default resolution paths, or would you want to surface the opt-in keyed-mode pairing here too?

JaredSnider-Bitwarden (changes requested):
Thank you for adding these doc changes! Claude and I found a few things below (included suggestions for fixes). Happy to discuss!

src/Core/Utilities/CACHING.md
withinfocus: Good call — both parts done.

Decision tree (lines 18–21) now reads in plain language: "Is it OK for a write on one instance to take a while to reach the in-memory caches on other instances (no cross-instance invalidation)?" Branches reworded to "in-memory cache + Cosmos/SQL/EF distributed cache" rather than L1/L2.

ExtendedCache section intro now defines the convention once, right after the FusionCache introduction, so the remaining uses downstream have a reference point.

src/Core/Utilities/CACHING.md
withinfocus: Added under Pairing with Cosmos DB:

> RU profile and the eager-refresh tradeoff. L1 absorbs most reads, which is the win: hot keys see roughly one Cosmos point-read per process per TTL window instead of one per request. The cost is on the write side — `EagerRefreshThreshold` (default `0.9f`) produces a steady cadence of background factory calls plus L2 writes (~1 per hot key per `Duration` window) that raw `CosmosCache` does not have. If your access pattern doesn't benefit from background refresh (e.g. cold keys, low read concurrency), set `EagerRefreshThreshold = null` to disable it.

withinfocus: Verified against `CosmosCache` + the `AddDistributedCache` wiring — you're right that "container-level TTL" reads as if the container itself is enforcing expiry. Rewrote the TTL sentence in the Pairing subsection:

> FusionCache's `Duration` (and any per-call `SetDuration`) is translated to `DistributedCacheEntryOptions.AbsoluteExpirationRelativeToNow` and written by `CosmosCache` as a per-document `ttl` on each Cosmos item — that's the authoritative expiry per entry. The Cosmos container must have `DefaultTimeToLive` enabled (`-1` or `>0`) for per-document `ttl` to apply at all; the container value also serves as a default for items that omit `ttl`, but FusionCache always sets a per-item value so the container value never caps it.

Re the pre-existing line at the Cosmos DB backend section ("TTL: Cosmos DB container-level TTL (automatic cleanup, no scanning required)") — same shape, but I left it alone since it's outside this PR's scope. Happy to fix it here in a follow-up commit if you'd prefer, or open a small separate cleanup.

src/Core/Utilities/CACHING.md
withinfocus: Added a bullet to Option 5's trailing comment block:

> Keys are namespaced from other "persistent" cache consumers: ExtendedCache prefixes every key with `"MyLongLivedCache:"`, so entries cannot collide with the raw `"persistent"` namespace used by direct `AddDistributedCache` consumers (payment workflow, OAuth grants).

src/Core/Utilities/CACHING.md
withinfocus: Verified at `ServiceCollectionExtensions.cs:786–800` — you're right, the `"persistent"` keyed cache only points to Cosmos when the Cosmos connection string is set. Updated the inline comment to:

```csharp
services.AddDistributedCache(globalSettings); // registers the "persistent" keyed cache —
                                              // Cosmos in cloud (when Cosmos.ConnectionString is set),
                                              // aliased to the unnamed IDistributedCache
                                              // (Redis / SQL / EF) in self-hosted
```

src/Core/Utilities/CACHING.md
withinfocus: Added the **Specific Example: Long-Lived Per-Tenant Computed Aggregates** section between the Org/Provider Abilities example and *When NOT to Use*, matching the order you suggested (Redis examples first, Cosmos example next).

Used your drafted `OrganizationAnalyticsSnapshot` service mostly verbatim — the fictional-types convention matches the SSO grants example, and the bullet list exercises every FusionCache feature (high read frequency, expensive factory, durability across deploys, eager refresh, fail-safe, tolerable staleness), so readers see exactly when the pairing pays off vs. doesn't.

src/Core/Utilities/CACHING.md
withinfocus: Fair question — went with a footnote rather than expanding the cell. The matrix historically shows what the framework auto-resolves to, and the opt-in keyed alias doesn't fit the column shape cleanly. Marked the ExtendedCache row with † pointing to:

> † This row shows the backend `ExtendedCache` resolves to automatically. `ExtendedCache` can also be wired to Cosmos (or any keyed `IDistributedCache`) by opting into keyed mode and aliasing under the cache name — see Option 5 and the Pairing `ExtendedCache` with Cosmos DB subsection.

Keeps the matrix consistent while making the opt-in path discoverable from here.

JaredSnider-Bitwarden (approved):
These are really helpful improvements. Thank you very much!

PR #7737, opened 2026-05-28
[PM-37087] Release migration schedule before adding org to provider"

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37087

## 📔 Objective

`ProviderBillingService.AddExistingOrganization` cancels and overwrites a standalone Stripe subscription when moving an org under provider management. Both `UpdateSubscriptionAsync(CancelAtPeriodEnd = false)` and `CancelSubscriptionAsync` fail outright when an active migration `SubscriptionSchedule` is attached — which can happen for Track A 2020 orgs (`PlanType` 8–11) that are both eligible for provider addition (per `Organization_ReadAddableToProviderByUserId.sql`) and eligible for a deferred price migration.

This PR applies the canonical "release migration schedule before mutating subscription" pattern used across the epic (PM-37083, PM-37084, PM-37085, PM-37068):

- Inject `IPriceIncreaseScheduler` into `ProviderBillingService` and call `Release(customerId, subscriptionId)` at the top of `AddExistingOrganization`, before the Stripe `UpdateSubscriptionAsync` and `CancelSubscriptionAsync`. `Release` is internally idempotent — no-ops if no active schedule exists.
- After the org transition commits (`ReplaceAsync` + `ProviderOrganization` create + event log), drop any stale cohort assignment row. Placing the cleanup at the bottom of the method means a failure on the assignment delete leaves a stale row to reconcile rather than a half-transitioned org.

### Note on deviation from the ticket spec

The ticket described a single 3-arg `Release(customerId, subscriptionId, organizationId)` call at the top of the method, with the cohort drop as a side effect inside `Release`. That 3-arg overload is shipping in PM-37085 and has not yet merged. This PR adapts by splitting the work into two operations (Release at top, cohort drop at bottom), with a `TODO(PM-37085)` at the cleanup block naming the exact mechanical follow-up — pass `organization.Id` as the third arg to `Release` and delete the cleanup block — once the 3-arg overload lands.

### Tests added

`bitwarden_license/test/Commercial.Core.Test/Billing/Providers/Services/ProviderBillingServiceTests.cs` — 11 tests in the new `AddExistingOrganization` region:

- `_ReleasesMigrationScheduleWithOrgGatewayIds_BeforeStripeOperations` — order regression; verifies `Release` is called with the org's gateway IDs (not the provider's) and runs before the Stripe mutations.
- `_DropsCohortAssignment_AfterOrgTransitionCommits` — Track A 2020 + assignment exists; verifies the assignment is deleted after `ReplaceAsync` + `ProviderOrganization` create + event log commit.
- `_DoesNotCallDelete_AndCompletesFlow_WhenNoCohortAssignment` — Track A 2020 + no assignment; verifies the rest of the flow completes.
- `_ContinuesEntireFlow_AfterReleaseSucceeds` — end-to-end happy path.
- `_DropsCohortAssignmentAndCompletes_ForMoeProvider` — MOE/BusinessUnit branch with discriminating provider plan (`TeamsMonthly`) vs MSP switch fallback (`EnterpriseMonthly`); proves the BusinessUnit branch of `GetManagedPlanTypeAsync` actually ran.
- `_DoesNotDropAssignment_ForCurrentGenOrg` — current-gen plan + no assignment.
- `_DropsStaleAssignment_ForCurrentGenOrg` — current-gen plan + stale assignment row from data drift; verifies the cleanup is plan-tier independent.
- `_SkipsCohortCleanup_WhenBusinessPlanMigrationFlagOff` — verifies the FF gate; even a stale row is not touched when the flag is off.
- `_PropagatesAndShortCircuits_WhenReleaseFails` — Release throws; nothing else runs.
- `_CompletesOrgTransition_BeforeAssignmentLookupFails` — assignment lookup throws after the org transition commits; the failure leaves a reconciliation problem, not a half-transitioned org.
- `_CompletesOrgTransition_BeforeDeleteAssignmentFails` — same invariant for the delete path.

### DIFF
--- a/bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
+++ b/bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
@@ -12,7 +12,6 @@ using Bit.Core.Billing;
 using Bit.Core.Billing.Constants;
 using Bit.Core.Billing.Enums;
 using Bit.Core.Billing.Extensions;
-using Bit.Core.Billing.Organizations.PlanMigration.Repositories;
 using Bit.Core.Billing.Payment.Models;
 using Bit.Core.Billing.Pricing;
 using Bit.Core.Billing.Providers.Entities;
@@ -44,7 +43,6 @@ public class ProviderBillingService(
     IFeatureService featureService,
     IGlobalSettings globalSettings,
     ILogger<ProviderBillingService> logger,
-    IOrganizationPlanMigrationCohortAssignmentRepository organizationPlanMigrationCohortAssignmentRepository,
     IOrganizationRepository organizationRepository,
     IPriceIncreaseScheduler priceIncreaseScheduler,
     IPricingClient pricingClient,
@@ -63,7 +61,8 @@ public class ProviderBillingService(
     {
         await priceIncreaseScheduler.Release(
             organization.GatewayCustomerId,
-            organization.GatewaySubscriptionId);
+            organization.GatewaySubscriptionId,
+            organization.Id);
 
         await stripeAdapter.UpdateSubscriptionAsync(organization.GatewaySubscriptionId,
             new SubscriptionUpdateOptions { CancelAtPeriodEnd = false });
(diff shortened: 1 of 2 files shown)


bitwarden_license/src/Commercial.Core/Billing/Providers/Services/ProviderBillingService.cs
```diff
@@ -148,6 +155,26 @@ await stripeAdapter.CreateCustomerBalanceTransactionAsync(provider.GatewayCustom
...
             providerOrganization,
             EventType.ProviderOrganization_Added);
+
+        // Drop any stale cohort assignment after the org transition commits. Placing the
+        // cleanup here (rather than alongside Release) makes a failure here a reconciliation
+        // problem — stale assignment row — rather than a half-transitioned org. Gated on
+        // PM35215 because the cohort table is only populated when the epic is enabled.
+        // TODO(PM-37085): once Release(customerId, subscriptionId, organizationId) lands,
+        //   1) pass organization.Id as the third arg to the Release call above, and
+        //   2) delete this cohortAssignment lookup/delete block.
```
amorask-bitwarden: ⛏️ The code below is fairly straight forward - I would say this level of commenting is overkill.

amorask-bitwarden (commented):
Change looks good, but if you don't mind cleaning up the large comment, that'd be appreciated.

amorask-bitwarden (approved):

PR #7740, opened 2026-05-29
[PM-30806] Self service change email command

ike-kottlowski:
## 🎟️ Tracking

[PM-30806](https://bitwarden.atlassian.net/browse/PM-30806)

🎏 Feature Flag: pm-30806-self-service-change-email-command
❗ : Do not turn on Feature Flag until Email is no longer used as Salt

## 📔 Objective

Adding a `SelfServiceChangeEmailCommand` that uses the generic `ChangeEmailCommand` to separate the legacy self-service change Email flow from the new self-service change Email Flow. The new change email flow does not modify the key state. This means the `masterpasswordhash` which is submitted to the `AccountsController.PostEmailToken` and `AccountsController.PostEmail` is used for user verification only. The new self-service change email command does not modify user keys or the user master password hash.

Other modifications were to the `ChangeEmailCommand` were to remove the branching logic from the return of `OrganizationDomainAllowEmailChangeQuery`. The query now `throws` directly instead of the `ChangeEmailCommand` throwing an error based on a `switch` from the old return enum. This was done because the on `SelfServiceChangeEmailCommand.InitiateChangeEmailAsync` we need to make sure the email the user wishes to change to is not owned by an organization. Throwing from the query directly makes the implementation DRY.

Testing is added for new classes and tests were modified to account for changes in the `ChangeEmailCommand` and the `organizationDomainAllowEmailChangeQuery`.

[PM-30806]: https://bitwarden.atlassian.net/browse/PM-30806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        return;
+    }
+
+    /// <inheritdoc />
+    public async Task InitiateChangeEmailAsync(User user, string masterPassword, string newEmail)
+    {
+        await CheckPasswordAndKeyConnector(user, masterPassword);
+
+        //? This query runs twice for self-service. Once here and once in the ChangeEmailCommand
```
JaredSnider-Bitwarden: 🎨 : I agree w/ the idea of possibly using a change email valid session `tokenable` to prevent duplicate submission of MP hash, but I do think the `_organizationDomainAllowEmailChangeQuery` running on both initial code email and final email change submission is a good thing b/c a domain can be claimed by an org at any time during the email change process.

src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        var authorizationResult =
+            await _authorizationService.AuthorizeAsync(
+                    _currentContext.HttpContext.User,
+                    user,
+                    KeyConnectorOperations.Use);
+        if (!authorizationResult.Succeeded)
+        {
+            throw new BadRequestException("You cannot change your email when using Key Connector.");
+        }
```
JaredSnider-Bitwarden: ⚠️ : I know we discussed trying to evaluate if we could use this clean authorization check when initially doing this work, but after reviewing its usage here, the authorization check was built to prevent admins and owners from setting a `KeyConnectorKeyWrappedUserKey` in the `SetKeyConnectorKeyCommand`.  The use of it here will introduce new behavior which will prevent admins and owners of Key connector orgs from being able to change their email. 

We need to go back to the `if (user.UsesKeyConnector)` check that the `src/Api/Auth/Controllers/AccountsController.cs PostEmailToken` method uses.  

We should also write an integration test that captures this behavior that KC owners & admins can change their email for the flagged and non-flagged paths.

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -145,11 +157,26 @@ public async Task PostEmail([FromBody] EmailRequestModel model)
             throw new UnauthorizedAccessException();
         }
 
+        if (_featureService.IsEnabled(FeatureFlagKeys.PM30806_SelfServiceChangeEmailCommand))
+        {
+            await _selfServiceChangeEmailCommand.ChangeEmailAsync(
+                user, model.MasterPasswordHash, model.NewEmail, model.Token);
+            return;
+        }
+
```
JaredSnider-Bitwarden, line 170: 🎨 : can we add a TODO: PM-<flag clean up task number> comment to remove the legacy path once the flag is cleaned up?

src/Api/Auth/Controllers/AccountsController.cs
```diff
@@ -115,6 +119,14 @@ public async Task PostEmailToken([FromBody] EmailTokenRequestModel model)
...
         }
 
+        if (_featureService.IsEnabled(FeatureFlagKeys.PM30806_SelfServiceChangeEmailCommand))
+        {
+            await _selfServiceChangeEmailCommand.InitiateChangeEmailAsync(
+                user, model.MasterPasswordHash, model.NewEmail);
+
+            return;
+        }
+
```
JaredSnider-Bitwarden, line 131: Same as https://github.com/bitwarden/server/pull/7740/changes#r3404939875

src/Api/Auth/Models/Request/Accounts/EmailRequestModel.cs
JaredSnider-Bitwarden, line 1: 🎨 : can we add comments with TODO + clean up ticket + clean up instructions about what we will remove?

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/Interfaces/IOrganizationDomainAllowEmailChangeQuery.cs
```diff
@@ -1,23 +1,22 @@
...
+    /// policy enabled.
     /// </remarks>
-    /// <param name="newEmailDomain">The bare domain portion of the target email (e.g. "example.com").</param>
-    /// <returns>
-    /// <see cref="OrganizationDomainAllowEmailChangeDenialReason.Allowed"/> when the change is permitted; otherwise the specific
-    /// denial reason so call sites can branch on it.
-    /// </returns>
-    Task<OrganizationDomainAllowEmailChangeDenialReason> IsAllowedAsync(User user, string newEmailDomain);
+    /// <param name="newEmail">The new email address trying to be set (e.g. "user@example.com").</param>
+    Task IsAllowedAsync(User user, string newEmail);
```
JaredSnider-Bitwarden: Non Blocking Naming nit: 
`IsAllowedAsync` is a predicate name — readers expect it to return `Task<bool>` or `Task<SomeEnum>`. A void-returning method that throws on failure should be named like `EnsureAllowedAsync` or `ValidateAllowedAsync` to match its actual contract.

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
JaredSnider-Bitwarden, line 1: 🎨 : nice tests! Can we add a TODO + the PM clean up ticket # which details what should be cleaned up on flag removal (e.g., test renames and flag mocking removed)?

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/OrganizationDomainAllowEmailChangeQuery.cs
```diff
@@ -11,31 +12,45 @@ public class OrganizationDomainAllowEmailChangeQuery(
...
     /// <inheritdoc />
-    public async Task<OrganizationDomainAllowEmailChangeDenialReason> IsAllowedAsync(User user, string newEmailDomain)
+    public async Task IsAllowedAsync(User user, string newEmail)
     {
-        // We want to see if the user is currently a claimed account
+        var newDomain = EmailValidation.GetDomain(newEmail);
+        if (newDomain == EmailValidation.GetDomain(user.Email))
+        {
+            return;
+        }
```
JaredSnider-Bitwarden: ⚠️: The new same-domain short-circuit skips the `BlockClaimedDomainAccountCreation` policy gate. Grandfathered users at a domain another org has verified-and-blocked can now change to a different local-part at that same domain — legacy code correctly denied this. Can we capture this scenario with tests?  I missed this scenario in the initial change password PR - sorry about that!

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
```diff
@@ -1589,4 +1597,220 @@ private async Task<HttpResponseMessage> PostRotateApiKeyAsync(string masterPassw
         });
         return await _client.SendAsync(message);
     }
+
+    [Fact]
+    public async Task PostEmail_SelfServiceFlagOn_Success_UpdatesEmailWithoutTouchingPasswordOrKey()
```
JaredSnider-Bitwarden, line 1686: ♻️ The ticket calls out the `PushLogOutAsync` → `PushSyncSettingsAsync` swap as one
of the central design decisions for the self-service path ("we can opt to not log
the user out... because their cryptographic state doesn't change"). This
intentional behavior change isn't asserted anywhere — a regression that quietly
restored the logout call would slip through CI.

Suggest pinning it here, alongside the existing no-rotation assertions:

```csharp
await _pushNotificationService.Received(1).PushSyncSettingsAsync(updatedUser.Id);
await _pushNotificationService.DidNotReceive().PushLogOutAsync(Arg.Any<Guid>(), Arg.Any<bool>());
```

JaredSnider-Bitwarden (changes requested):
Nice work! I found a few tweaks we need to make below 👇

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/Interfaces/IOrganizationDomainAllowEmailChangeQuery.cs
ike-kottlowski: addressed in: https://github.com/bitwarden/server/pull/7740/commits/2df8919c7cdefaadc9f57d1d872a7a80e01758c5

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
ike-kottlowski: addressed in: https://github.com/bitwarden/server/pull/7740/commits/8d0a17f0eedbc22925a9cc9d23b1a0773e77b03b

src/Api/Auth/Models/Request/Accounts/EmailRequestModel.cs
ike-kottlowski: addressed in: https://github.com/bitwarden/server/pull/7740/commits/8d0a17f0eedbc22925a9cc9d23b1a0773e77b03b

src/Api/Auth/Controllers/AccountsController.cs
ike-kottlowski: addressed in: https://github.com/bitwarden/server/pull/7740/commits/8d0a17f0eedbc22925a9cc9d23b1a0773e77b03b

src/Api/Auth/Controllers/AccountsController.cs
ike-kottlowski: addressed in: https://github.com/bitwarden/server/pull/7740/commits/8d0a17f0eedbc22925a9cc9d23b1a0773e77b03b

src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
ike-kottlowski: addressed in: https://github.com/bitwarden/server/pull/7740/commits/8dbc3f618ad4a99a6a21dc2324afa34aa9ef201b

src/Core/AdminConsole/OrganizationFeatures/OrganizationDomains/OrganizationDomainAllowEmailChangeQuery.cs
ike-kottlowski: addressed in:
https://github.com/bitwarden/server/pull/7740/commits/e3384ca5b98fbd5f6159a1d79ab7bc7adaf81015

test/Api.IntegrationTest/Controllers/AccountsControllerTest.cs
ike-kottlowski: addressed in:
https://github.com/bitwarden/server/pull/7740/commits/b5197c6914f7560643a462bcd4e37cff2147cb2c

src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
ike-kottlowski: thanks for the feedback. comment removed: https://github.com/bitwarden/server/pull/7740/commits/2df8919c7cdefaadc9f57d1d872a7a80e01758c5

test/Core.Test/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommandTests.cs
```diff
@@ -0,0 +1,302 @@
...
+using NSubstitute.ExceptionExtensions;
+using Xunit;
+
+namespace Bit.Core.Test.Auth.UserFeatures.UserEmail;
+
+[SutProviderCustomize]
+public class SelfServiceChangeEmailCommandTests
+{
+    // Same-domain pair: lets tests bypass the org-domain policy gate via the
+    // EmailValidation.GetDomain short-circuit in EnsureNewEmailDomainAllowedAsync.
```
JaredSnider-Bitwarden: ♻️ Stale test comments referencing the same-domain short-circuit removed in e3384ca5b9

test/Core.Test/Auth/UserFeatures/UserEmail/ChangeEmailCommandTests.cs
JaredSnider-Bitwarden: ♻️ Stale test comments referencing the same-domain short-circuit removed in e3384ca

src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
JaredSnider-Bitwarden: Should we parameterize that test so it covers admins as well as owners?

JaredSnider-Bitwarden (changes requested):
1 test tweak and two stale comment clean ups and I think this is good to go!

src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
ike-kottlowski: good idea! https://github.com/bitwarden/server/pull/7740/commits/470cef5d0a8ca7e749f4896e4cc849da0d517ddd

test/Core.Test/Auth/UserFeatures/UserEmail/ChangeEmailCommandTests.cs
ike-kottlowski: fixed https://github.com/bitwarden/server/pull/7740/commits/0843e7dd9bb5d468b15b0a5f9f36c7941edffd4b

test/Core.Test/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommandTests.cs
ike-kottlowski: fixed https://github.com/bitwarden/server/pull/7740/commits/0843e7dd9bb5d468b15b0a5f9f36c7941edffd4b

src/Core/Auth/UserFeatures/UserEmail/SelfServiceChangeEmailCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+    IOptions<IdentityOptions> identityOptions) : ISelfServiceChangeEmailCommand
+{
+    private readonly IUserService _userService = userService;
+    private readonly IPushNotificationService _pushService = pushService;
+    private readonly UserManager<User> _userManager = userManager;
+    private readonly IUserRepository _userRepository = userRepository;
+    private readonly IMailService _mailService = mailService;
+    private readonly IOrganizationDomainAllowEmailChangeQuery _organizationDomainAllowEmailChangeQuery = organizationDomainAllowEmailChangeQuery;
+    private readonly IChangeEmailCommand _changeEmailCommand = changeEmailCommand;
+    private readonly IdentityOptions _identityOptions = identityOptions.Value;
```
jrmccannon, line 29: Question: Do we need these private fields?

ike-kottlowski: This creates an immutable `readonly` reference that is safer to use. It's also the standard Auth has adopted for our DI bits.

JaredSnider-Bitwarden (approved):
Excellent work. Thank you for the iterations!

jrmccannon (approved):

PR #7742, opened 2026-05-29
[SM-1952] Add access token lifecycle events

maxkpower:
## 🎟️ Tracking

[https://bitwarden.atlassian.net/browse/SM-1952](https://bitwarden.atlassian.net/browse/SM-1952)

## 📔 Objective

Emit event-log entries for access-token lifecycle actions on machine accounts.

- Added two event types `AccessToken_Created` (2306) and `AccessToken_Revoked` (2307). These are logged from `ServiceAccountsController` when an access token is created or revoked, alongside the existing `ServiceAccount_Created` and `ServiceAccount_Deleted` events.

- To ensure the audit trail is accurate `IRevokeAccessTokensCommand.RevokeAsync` now returns the tokens it actually deleted. The revoke event is only logged when at least one token was revoked; requests with empty, stale, or non-matching IDs no longer produce a misleading "revoked" entry.

## 📸 Screenshots

https://github.com/user-attachments/assets/d9bd6668-5afa-4260-bb52-8be163227f88


cd-bitwarden (dismissed):
LGTM!

Banrion (dismissed):

PR #7750, opened 2026-06-01
[PM-38424] Add Kerberos library dependency to setup

notnamed:
## 📔 Objective

The setup container runs database migrations, either manually via the updatedb verb in bitwarden.sh/.ps1, or during a full upgrade. When the database is external and Kerberos authentication is configured, the setup container needs to load the Kerberos libraries need to be installed in order to handle that authentication to the database


mimartin12 (approved):

keithhubner (approved):

djsmith85 (approved):
Approving for platform

PR #7752, opened 2026-06-01
[PM-38328] Speed up JSON Processing

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38328

## 📔 Objective

Speeds up JSON processing by providing singular source of truth which is byte for byte and doesn't need to be serialized everywhere constantly.


test/Core.Test/AdminConsole/Utilities/AdminConsoleJsonContextTests.cs
```diff
@@ -0,0 +1,183 @@
...
+namespace Bit.Core.Test.AdminConsole.Utilities;
+
+/// <summary>
+/// Verifies that <see cref="AdminConsoleJsonContext"/> produces byte-for-byte identical
+/// output to the reflection-based path that uses the same CamelCase naming policy, and
+/// that round-trip deserialization reconstructs every property correctly.
+/// </summary>
+public class AdminConsoleJsonContextTests
+{
+    // ----- Permissions -----
```
r-tome: :art: These type of comments should be removed

src/Core/AdminConsole/Utilities/AdminConsoleJsonContext.cs
r-tome, line 1: `MasterPasswordPolicyData` and `SendOptionsPolicyData` aren't wired up. https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Entities/Policy.cs should be refactored to use something like `AdminConsoleJsonContext.Default.GetTypeInfo(typeof(T))`

https://github.com/bitwarden/server/blob/main/src/Core/AdminConsole/Entities/OrganizationUser.cs should use `AdminConsoleJsonContext.Default.Permissions`

https://github.com/bitwarden/server/blob/0cab8e25c6339a73a1dd021b3072e3ac7fea8101/src/Core/AdminConsole/Models/Data/Organizations/OrganizationUsers/OrganizationUserUserDetails.cs#L80 same thing here

Basically look for where we're using `CoreHelpers.LoadClassFromJsonData` and `CoreHelpers.ClassToJsonData`

r-tome (changes requested):
This is looking good but I think its incomplete. I really like the tests!

Also, the ticket link above is incorrect

src/Core/AdminConsole/Utilities/AdminConsoleJsonContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Text.Json.Serialization;
+using Bit.Core.AdminConsole.Models.Data.Organizations.Policies;
+using Bit.Core.Models.Data;
+
+namespace Bit.Core.AdminConsole.Utilities;
+
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+[JsonSerializable(typeof(Permissions))]
+[JsonSerializable(typeof(MasterPasswordPolicyData))]
+[JsonSerializable(typeof(SendOptionsPolicyData))]
```
r-tome, line 10: `SendControlsPolicyData` should be added here, it also implements `IPolicyDataModel`. Also add tests to make sure its exactly the same as it was

r-tome (changes requested):

eliykat (changes requested):
Interesting idea - I wasn't familiar with this option. Before digging into the implementation, my request is to include performance measurements showing that this results in meaningful improvements.  @r-tome has written performance tests before and can provide guidance, I also see there is a `MicroBenchmarks` assembly that I'm not familiar with, you could ask Justin about that.

You could benchmark this:
- in isolation, and/or
- in the context of one of our flows - where do we actually expect to see a performance improvement in our app?

In particular, this jumped out at me from the [docs](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/reflection-vs-source-generation):
> Whenever JsonSerializer has to serialize or deserialize a type for the first time, it collects and caches this metadata. The metadata collection process takes time and uses memory.

If it is only a one-time cost after pods start, that may not be meaningful; but again, measurements can tell us.

JaredScar:
> Interesting idea - I wasn't familiar with this option. Before digging into the implementation, my request is to include performance measurements showing that this results in meaningful improvements. @r-tome has written performance tests before and can provide guidance, I also see there is a `MicroBenchmarks` assembly that I'm not familiar with, you could ask Justin about that.
> 
> You could benchmark this:
> 
> * in isolation, and/or
> * in the context of one of our flows - where do we actually expect to see a performance improvement in our app?
> 
> In particular, this jumped out at me from the [docs](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/reflection-vs-source-generation):
> 
> > Whenever JsonSerializer has to serialize or deserialize a type for the first time, it collects and caches this metadata. The metadata collection process takes time and uses memory.
> 
> If it is only a one-time cost after pods start, that may not be meaningful; but again, measurements can tell us.

Benchmarking has been done!

Here are the results of changing from `Reflection` to `Source-gen`

Serialize `ResetPasswordDataModel` => 47.9 NS => 39.1 NS
Serialize `Permissions` => 171 NS => 146 NS
Deserialize `Permissions` => 288 NS => 288 NS

Reflection's first-call cost is a one-time hit per pod lifetime.

@r-tome @eliykat

eliykat (commented):
@r-tome Let me know what you think, but if we're talking nanoseconds this doesn't seem worth it. QA did some performance testing - maybe last year? - that you could go look at to identify some other pain points.

r-tome:
> @r-tome Let me know what you think, but if we're talking nanoseconds this doesn't seem worth it. QA did some performance testing - maybe last year? - that you could go look at to identify some other pain points.

Agreed, not worth it. I wasn't liking the divergence we'd have by not using `CoreHelpers` and this settles it. Thanks for putting the effort in this @JaredScar

PR #7757, opened 2026-06-02
[PM-37085] feat: Release migration schedule on organization plan upgrade

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37085

## 📔 Objective

Part of the business-plan price-migration workstream (the "Maintaining Subscription Updates" axis — Gap 2 of 4). When a Track A 2020 org voluntarily upgrades to a different plan while an active deferred-migration Stripe `SubscriptionSchedule` is attached, the upgrade misbehaves: with PM-37083's per-phase price resolution in place, the Phase 1 -> Phase 2 transition collapses into a no-op, and the migration's proactive discount coupon silently inherits into Phase 2 of the plan the user voluntarily chose.

`UpgradeOrganizationPlanVNextCommand` now releases any active migration schedule before delegating to `UpdateOrganizationSubscriptionCommand`, which then applies the upgrade as a normal `UpdateSubscriptionAsync`. A voluntary plan change makes the deferred migration moot.

**Details**

- `IPriceIncreaseScheduler.Release` gains an optional `Guid? organizationId = null`. When provided, after releasing the Stripe schedule it also drops the org's migration cohort assignment (load + `DeleteAsync`) so the org leaves the cohort — otherwise the row lingers with `ScheduledDate` set and cohort views still list it. Existing callers pass no org id and keep today's Stripe-only behavior.
- The cohort drop is coupled with the release (same `try`) but uses its own inner `try/catch` with a cohort-specific log message, so a cleanup failure isn't misreported as a Stripe release failure. It is best-effort: after a successful release, a failed cohort drop is logged for manual cleanup rather than aborting the upgrade (safe direction — schedule gone, cohort retained; the runtime drift check prevents a stale row from re-scheduling).
- The release is driven by the presence of an active schedule, not a feature flag, so it safely no-ops when no schedule exists.

**Testing**

- Unit tests on `PriceIncreaseScheduler.Release` (release + cohort drop, no assignment row, no org id, no active schedule, release throws -> rethrow, cohort delete throws -> swallowed) and `UpgradeOrganizationPlanVNextCommand` (paid upgrade releases with org id, Free->paid skips release, release throws -> upgrade aborts).
- Full `Core.Test` suite passes except 5 pre-existing failures in `Dirt.Services.EventIntegrationHandlerTests` that also fail on `main` (unrelated to this change).

**Out of scope**

- Per-phase price resolution (PM-37083), business-aware schedule recovery (PM-37084), and provider-org-addition release (PM-37087) are handled separately. Auditing whether the other existing `Release` callers should also pass an org id is a follow-up.

### DIFF
--- a/src/Core/Billing/Organizations/Commands/UpgradeOrganizationPlanVNextCommand.cs
+++ b/src/Core/Billing/Organizations/Commands/UpgradeOrganizationPlanVNextCommand.cs
@@ -43,6 +43,7 @@ public class UpgradeOrganizationPlanVNextCommand(
     IOrganizationBillingService organizationBillingService,
     IOrganizationService organizationService,
     IPricingClient pricingClient,
+    IPriceIncreaseScheduler priceIncreaseScheduler,
     IUpdateOrganizationSubscriptionCommand updateOrganizationSubscriptionCommand) : BaseBillingCommand<UpgradeOrganizationPlanVNextCommand>(logger), IUpgradeOrganizationPlanVNextCommand
 {
     protected override Conflict DefaultConflict => new("We had a problem upgrading your plan. Please contact support for assistance.");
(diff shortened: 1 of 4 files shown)


cyprain-okeke (approved):
looks good

PR #7759, opened 2026-06-03
[PM-34778] Implement AcceptOrganizationInviteLinkCommand

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34778

## 📔 Objective

Add the server command that lets a user accept an organization invite link.

`AcceptOrganizationInviteLinkCommand` validates the link and organization, enforces membership and policy rules (single-org, two-factor, reset-password, free-org admin limit), then joins the user by either upgrading a pending invite or creating a new membership (adding a seat if full).

To share the policy-enforcement logic with `AcceptOrgUserCommand`, this branch also extracts `AcceptOrganizationMembershipValidator`, a dedicated validator that consolidates the AutoConfirm, SingleOrg, and two-factor checks. `AcceptOrgUserCommand` is updated to use it, removing its three inline policy methods.

Also pulls seat-availability logic into a shared helper, registers DI, and adds unit tests.

Scope: Core command only. The API endpoint comes in a follow-up branch.


src/Core/AdminConsole/OrganizationFeatures/InviteLinks/AcceptOrganizationInviteLinkCommand.cs
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        var admins = await organizationUserRepository.GetManyByMinimumRoleAsync(organization.Id, OrganizationUserType.Admin);
+        var adminEmails = admins.Select(a => a.Email).Distinct().ToList();
+        if (adminEmails.Count > 0)
+        {
+            await mailService.SendOrganizationAcceptedEmailAsync(organization, user.Email, adminEmails);
+        }
+
+        await pushAutoConfirmNotificationCommand.PushAsync(user.Id, organization.Id);
```
BTreston: Will this collide with @JimmyVo16 's work for milestone 3? Probably worth flagging this for clean up in milestone 3 if it will. --I would also add that we should try and some distinction between the auto confirm policy command and the future auto confirm done by the invite link. I can see that becoming confusing quickly

Also, just a note: This will not auto confirm a user if the org does not have the auto confirm policy and browser extension setting enabled. If it was intended to try and bridge that gap until milestone 3, just know this won't do that. If it is for the existing auto confirm policy feature, then feel free to ignore this 😃

BTreston (dismissed):
Looks good to me, just some non-blocking questions

src/Core/AdminConsole/OrganizationFeatures/InviteLinks/AcceptOrganizationInviteLinkCommand.cs
r-tome: > Will this collide with @JimmyVo16 's work for milestone 3? 

It will not. This was discussed on our dev sync and milestone 3 will be a different endpoint

> we should try and some distinction between the auto confirm policy command and the future auto confirm done by the invite link

Added a comment to clarify this

> This will not auto confirm a user if the org does not have the auto confirm policy and browser extension setting enabled. If it was intended to try and bridge that gap until milestone 3

This is intended to help enforce the existing auto-confirm policy. Milestone 3 will add a new accept link endpoint that will skip this and auto-confirm immediately.

BTreston (dismissed):

BTreston (dismissed):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptMembership/AcceptOrganizationMembershipValidator.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        var singleOrgRequirement = await policyRequirementQuery
+            .GetAsync<SingleOrganizationPolicyRequirement>(request.User.Id);
+        var singleOrgError = singleOrgRequirement.CanJoinOrganization(
+            request.OrganizationId, request.AllOrganizationMemberships);
+        if (singleOrgError is not null)
+        {
+            return Invalid(new AcceptOrganizationMembershipValidationResult(), singleOrgError);
+        }
```
BTreston, line 45: I think this should be handled by the `AutomaticUserConfirmationPolicyEnforcementHandler`, no?

BTreston (commented):
Just 1 question. The validators are v/ nice, keeps the commands much leaner.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptMembership/AcceptOrganizationMembershipValidator.cs
r-tome: It has to be handled by both (PolicyType.SingleOrg and PolicyType.AutomaticUserConfirmation) because the organization can have auto confirm disabled but single org enabled. This is mirroring what `AcceptOrgUserCommand` already did

BTreston (dismissed):

BTreston (approved):
Thank you

PR #7761, opened 2026-06-03
Set fill assist server urls

differsthecat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38486

## 📔 Objective

Set the fillAssistRules URLs


jprusik (approved):

jprusik (approved):
This is the correct value! (apologies for my providing the wrong one previously)

MGibson1 (approved):
Looks good from the server side. the web-specific url handling is all done in the clients, with config files. I think for this, you'll want to update the [base config](https://github.com/bitwarden/clients/blob/47e5d3af09c41df4118caea98299c929dd0b276d/apps/web/config/base.json#L2-L3) to include the new url (see the other configs for formatting, since this is empty).

For other clients, You'll also want to update the [default environments](https://github.com/bitwarden/clients/blob/47e5d3af09c41df4118caea98299c929dd0b276d/libs/common/src/platform/services/default-environment.service.ts#L85-L115).

Finally, I think you should think about not expecting the dev value to be a localhost, but this github link. That way it works for typical dev work and autofill people can update with local config overrides since it's unlikely anyone not working on autofill will have one of these map servers running

PR #7763, opened 2026-06-04
Create exploring-bitwarden-data Claude Skill

theMickster:
## 🎟️ Tracking

N/A

## 📔 Objective

- Adds `exploring-bitwarden-data`, a skill for **read-only** exploration of a local Bitwarden dev DB.
- Was used to craft and verify every fixture under `util/Seeder/Seeds/fixtures`; earlier iterations drove the Seeder itself.
- Read-only by design: a pre-read-only version once wiped a full dev DB after a Seeder validation run.
- Built with `/skill-creator` — eval-gated, so each trim is proven against a recorded baseline, not guessed.

## 🔬 What's inside
- SKILL.md kept lean (invariants + workflow); per-provider runbooks live in `references/*.md`.
- Three providers documented and marked Ready: **MSSQL, PostgreSQL, MySQL**.
- Each connects through a dedicated **read-only login** via the `BW_*_` env contract, plus a **per-session read-only guard**; mutations are also blocked by the `block-mutating-sql.sh` hook (MSSQL).

## ✅ Validation
- Eval set scores **24/24** with the skill vs **62.5%** without; SKILL.md trimmed **605 → 259 lines**, no regression.
- Same fixture query returns identical counts across **MSSQL, PostgreSQL, and MySQL**.
- Read-only enforcement verified per provider — writes refused: MSSQL (`db_datareader` + `-K ReadOnly`); PostgreSQL role → `permission denied`; MySQL user → `ERROR 1142`; plus the session guards.

<details>
<summary> Expand for more details </summary>
<br/>
### Ablation conclusion

Everything else tied: no-skill found the correct joins, the `UserCipherDetails`/`UserCollectionDetails` functions, `DeletedDate` semantics, enum values (more completely than the skill's drifted table — it knew `Staged = 3`), and avoided the encrypted-column LIKE trap by reading `src/Sql` and enum sources. Those surfaces were cut (the deletions and their per-case evidence are this PR's diff and this table). The bar for re-adding any instruction: a case in `evals.json` that fails without it.

### Deferred (parked, with price of admission)

| Capability                                                                   | Build when                                                                                        |
| ---------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
| Write-mode seeding/mutation support ("this shouldn't be just about reading") | A safe write workflow is designed with its own eval set and permission model                      |
| MySQL/PostgreSQL provider references + per-provider read-only hooks          | The multi-provider seed-and-verify loop produces validated per-provider runbooks                  |
| Send-liveness and AuthRequest-expiry grounding rules                         | Both tables are empty database-wide today; a preset seeds them and a case exercises the semantics |

Eval-fixture improvements for the next iteration (grader-flagged): give eval-13 a direct-RO=1/group-RO=0 conflict so the count alone discriminates precedence; give eval-9 an org with mixed `Use*` flags so "all enabled" can't pass trivially.

### Provider validation

Sterling Cooper seeded per provider by flipping `globalSettings:databaseProvider`; identical counts on MSSQL/Dapper, PostgreSQL, and MySQL (status mix 12/13/12/214; ciphers 5,000/25 deleted/50 archived; 50 groups; 500 collections; all payloads `2.`-prefixed EncStrings). Read-only guards in the provider references were verified by rejected write attempts on all three providers.

### Triggering (description optimization, 5 iterations)

Precision 100% (0 false fires across all near-miss negatives, including production asks and write requests), recall 6–17% regardless of description wording — organic auto-triggering is weak for structural reasons (sandboxed trials lacked DB env vars; Claude Code under-consults skills it can substitute with raw Bash). The skill is intended to be reliable via explicit invocation (`user-invocable: true`). Trigger cases: [trigger-eval-set.json](trigger-eval-set.json).

</details>

## 📌 Notes
- Skill relies on **explicit invocation** — description-optimizer showed high precision, low auto-recall (documented in evals).
- Scope is **MSSQL, PostgreSQL, MySQL**.
- Deferred: per-provider eval cases; a companion write-mode skill.


.claude/hooks/block-mutating-sql.sh
```diff
@@ -0,0 +1,176 @@
...
+  'REVOKE[[:space:]]'
+  'DENY[[:space:]]'
+)
+
+for pattern in "${sql_deny_patterns[@]}"; do
+  if echo "$upper_command" | grep -qE "$pattern"; then
+    cat <<HOOKEOF
+{
+  "hookSpecificOutput": {
+    "permissionDecision": "deny"
```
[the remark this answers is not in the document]
theMickster: Yes

.claude/skills/exploring-bitwarden-data/references/schema-discovery-queries.md
```diff
@@ -0,0 +1,70 @@
+# Schema Discovery Queries — Bitwarden (MSSQL)
+
+Read-only `INFORMATION_SCHEMA` and `sys.*` queries for tables, columns, FKs, and view definitions. MSSQL-only — other providers will use their own introspection (MySQL `INFORMATION_SCHEMA`, PostgreSQL `pg_catalog`, SQLite `sqlite_master`).
+
+Read when the schema reference lists ([schema-tables.md](schema-tables.md), [schema-relationships.md](schema-relationships.md), [schema-views.md](schema-views.md)) are insufficient: unlisted tables/columns, full schema enumeration, or fetching a view's definition.
```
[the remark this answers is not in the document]
theMickster: This has been resolved

.claude/skills/exploring-bitwarden-data/evals/baseline-results.md
SaintPatrck, line 1: 🎨 For the baseline, we only need to record the latest/current results. Future changes are expected to meet or exceed them. We don't want to track every version change, or it will become unwieldy. We can trim it down to only noting the fixture and the results.

```markdown
# Baseline

Fixture: `scale.md-balanced-sterling-cooper --mangle` (MSSQL); ground truth via `capture-ground-truth.sh`. One run per case, variance not measured.

## Results (assertions passed / total)

| Eval                                 | Pass rate |
| ------------------------------------- | --------- |
| 0 active-member-count                 | 2/2       |
| 1 member-status-breakdown              | 2/2       |
| 2 user-visible-ciphers                 | 3/3       |
| 3 top-users-direct-collection          | 3/3       |
| 4 active-vs-deleted-ciphers            | 2/2       |
| 5 archived-org-ciphers                 | 2/2       |
| 6 personal-ciphers-of-members          | 2/2       |
| 7 org-plan-and-active-state            | 2/2       |
| 8 encrypted-column-trap                | 2/2       |
| 9 org-abilities-via-view               | 2/2       |
| 13 collection-permission-resolution    | 2/2       |
| **Mean per-eval pass rate**            | **1.00**  |
```

.claude/skills/exploring-bitwarden-data/evals/baseline-results.md
```diff
@@ -0,0 +1,54 @@
...
+| 4 active-vs-deleted-ciphers         | 2/2      | 2/2      | 2/2      |
+| 5 archived-org-ciphers              | 0/2      | 2/2      | 2/2      |
+| 6 personal-ciphers-of-members       | 1/2      | 2/2      | 2/2      |
+| 7 org-plan-and-active-state         | 1/2      | 2/2      | 2/2      |
+| 8 encrypted-column-trap             | 2/2      | 2/2      | 2/2      |
+| 9 org-abilities-via-view            | 2/2      | 2/2      | 2/2      |
+| 13 collection-permission-resolution | 2/2      | 2/2      | 2/2      |
+| **Mean per-eval pass rate**         | **0.71** | **1.00** | **1.00** |
+
+## What the skill flips (the three cases no-skill failed)
```
SaintPatrck: 🎨 `What the skill flips` is implied based on the latest results and should be expressed in documentation. Not tracked in the results table. If someone wanted to re-assess no-skill vs the current state they would run evals without the skill and simply compare those results to the latest. If the core model no longer needs the skill to achieve its goal, we'd have 100% pass rate when it's not used.

.claude/skills/exploring-bitwarden-data/evals/baseline-results.md
```diff
@@ -0,0 +1,54 @@
...
+
+Eval-fixture improvements for the next iteration (grader-flagged): give eval-13 a direct-RO=1/group-RO=0 conflict so the count alone discriminates precedence; give eval-9 an org with mixed `Use*` flags so "all enabled" can't pass trivially.
+
+## Provider validation
+
+Sterling Cooper seeded per provider by flipping `globalSettings:databaseProvider`; identical counts on MSSQL/Dapper, PostgreSQL, and MySQL (status mix 12/13/12/214; ciphers 5,000/25 deleted/50 archived; 50 groups; 500 collections; all payloads `2.`-prefixed EncStrings). Read-only guards in the provider references were verified by rejected write attempts on all three providers.
+
+## Triggering (description optimization, 5 iterations)
+
+Precision 100% (0 false fires across all near-miss negatives, including production asks and write requests), recall 6–17% regardless of description wording — organic auto-triggering is weak for structural reasons (sandboxed trials lacked DB env vars; Claude Code under-consults skills it can substitute with raw Bash). The skill is intended to be reliable via explicit invocation (`user-invocable: true`). Trigger cases: [trigger-eval-set.json](trigger-eval-set.json).
```
SaintPatrck: 🎨 All of this (Ablation reasoning, deferred changes, provider validation, and triggering) is important, but only in the context of this PR. I think it belongs in the PR description, not committed in the baseline result file. 

💭 I'm assuming the deferred changes ledger is a symptom of pointing Claude to the guidelines doc we've started drafting. If so, thank you for exercising it. This, to me, indicates we need to refine the doc's guidance. I suggest tracking the deferred changes in JIRA, if they're meant to be worked on soon-ish. I don't think this is the right place to record them. If they're in JIRA anyone with bandwidth can pickup the work. What do you think?

.claude/hooks/block-mutating-sql.sh
SaintPatrck, line 1: ❓ 🎨 What do you think about scoping this hook to the skill? At the top level like this, it's going to block commands the user may want Claude to run outside of the skill's lifecycle. Scoping it to the Skill only, also means users don't have to register it in `settings.local.json`. It will be registered when the skill is invoked. 

🤔 The docs state hook(s) declared in a Skill or Agent "...are scoped to the component’s lifecycle and only run when that component is active.[^1]" I did some local experimenting and found that if the Skill is invoked in your main session, the hook is registered for the remainder of that session. Not ideal, imo, when you consider a session is not likely to be started just to run this skill. Maybe as a future improvement we tie this skill to an `agent` and `fork` its context. That would prevent the hook from bleeding into the main session context and block intentional mutation commands. 

[^1]: https://code.claude.com/docs/en/hooks#hooks-in-skills-and-agents

SaintPatrck (changes requested):

.claude/hooks/block-mutating-sql.sh
theMickster: What do you mean when you say "scoping this hook to the skill"? The hook isn't in a setting.json file and is only referenced in the Skill.md (see below). 

```
hooks:
  PreToolUse:
    - matcher: "Bash"
      hooks:
        - type: command
          command: "bash ${CLAUDE_PROJECT_DIR}/.claude/hooks/block-mutating-sql.sh"
          timeout: 5
```

That's a bit of an odd/unwanted behavior where the skill bleeds into the session. 
If that's the case, then I think it's best that we ditch it entirely and rely upon a combination of the skill's instructions and the type of database user the Bitwarden engineer created.  My users are all scoped to read-only privs at the database level so the hook is a semi-redundant safe guard; and if another Bitwarden engineer choses to use 'sa' or 'root' to explore their database and Claude janks it up then it's on them; frankly. The whole read-only concept had push back from the beginning. So maybe just ditch this whole concept all together and save it for another time? I'm fine with that.

.claude/skills/exploring-bitwarden-data/SKILL.md
```diff
@@ -0,0 +1,65 @@
...
+hooks:
+  PreToolUse:
+    - matcher: "Bash"
+      hooks:
+        - type: command
+          command: "bash ${CLAUDE_PROJECT_DIR}/.claude/hooks/block-mutating-sql.sh"
+          timeout: 5
+---
+
+# Explorer Bitwarden Database
```
SaintPatrck: ⛏️ Typo?
```suggestion
# Explore Bitwarden Database
```

.claude/skills/exploring-bitwarden-data/SKILL.md
```diff
@@ -0,0 +1,65 @@
+---
+name: exploring-bitwarden-data
+description: Read-only exploration of a local Bitwarden development database — answer business questions from live data, verify seeded fixtures, and introspect schema. Use whenever the user wants to query, count, look up, verify, or explore data in a local Bitwarden database ("how many orgs/users/ciphers", "show me collections", "check what the seeder created", "look up user X", "which orgs have feature Y"), even without the word SQL. Not for authoring stored procedures, migrations, or repository code (use writing-database-queries), and not for seeding or modifying data.
+argument-hint: "[mssql|mysql|postgresql] <question or SQL>"
+user-invocable: true
+allowed-tools: "Bash(which sqlcmd), Bash(sqlcmd:*)"
```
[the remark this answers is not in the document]
theMickster: Not a snowballs chance is h-e-double-hockey-sticks will I let docker exec in the allowed tools for Claude to go rogue. The others are fine. Addressed in 919a505749f11a096761386735d380b0a48bfa3a

.claude/skills/exploring-bitwarden-data/evals/baseline-results.md
theMickster: Addressed in 919a505749f11a096761386735d380b0a48bfa3a
I have moved all comments into the PR description into a `<details>` collapse.

.claude/skills/exploring-bitwarden-data/evals/baseline-results.md
theMickster: Removed & addressed in 919a505749f11a096761386735d380b0a48bfa3a

.claude/skills/exploring-bitwarden-data/evals/baseline-results.md
theMickster: Refactored and addressed in 919a505749f11a096761386735d380b0a48bfa3a

.claude/hooks/block-mutating-sql.sh
SaintPatrck: Hmm. I must have crossed-wires somewhere. Could have sworn I saw instructions to register it in `settings.local.json`. My bad. Thanks for keeping me honest. :feelsbadman:

I agree the behavior is not ideal. I like the idea of omitting the hook for now. We can probably come up with better ways to enforce read-only operations, that doesn't bleed into sessions.

theMickster: Maybe that was because I described them in the readme. no probs. 
Addressed in d93a0ddb77d4b958aa4c48c134447aa6cae85b99

.claude/skills/exploring-bitwarden-data/references/providers/mysql.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash
+[ -n "${BW_MYSQL_SERVER:-}" ] && [ -n "${BW_MYSQL_DB_NAME:-}" ] && \
+[ -n "${BW_MYSQL_USERNAME:-}" ] && [ -n "${BW_MYSQL_PASSWORD:-}" ] && \
+echo "Bitwarden MySQL env vars OK" || echo "MISSING Bitwarden MySQL env var"
+```
+
+## Read-only guard — required on every connection
+
+Connect as the read-only user in `BW_MYSQL_USERNAME` (`bitwarden_data_reader`, `SELECT`-only). The bash hook only inspects `sqlcmd`, so it won't catch `mysql` writes — as defense-in-depth, also open every session with `SET SESSION TRANSACTION READ ONLY;` (verified: writes then fail with `ERROR 1792: Cannot execute statement in a READ ONLY transaction`; autocommit statements are implicit transactions, so plain INSERTs are covered). This guard is per-session, so it only holds if the statement is actually issued.
```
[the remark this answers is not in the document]
theMickster: Addressed in d0f6eb843e1904383ca399a8374f033731f9e9b4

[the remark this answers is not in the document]
theMickster: Addressed in d0f6eb843e1904383ca399a8374f033731f9e9b4

.claude/skills/exploring-bitwarden-data/references/providers/postgresql.md
```diff
@@ -0,0 +1,63 @@
...
+
+```bash
+[ -n "${BW_POSTGRES_SERVER:-}" ] && [ -n "${BW_POSTGRES_DB_NAME:-}" ] && \
+[ -n "${BW_POSTGRES_USERNAME:-}" ] && [ -n "${BW_POSTGRES_PASSWORD:-}" ] && \
+echo "Bitwarden PostgreSQL env vars OK" || echo "MISSING Bitwarden PostgreSQL env var"
+```
+
+## Read-only guard — required on every connection
+
+Connect as the read-only login in `BW_POSTGRES_USERNAME` (`bitwarden_data_reader`, `SELECT`-only). The bash hook only inspects `sqlcmd`, so it won't catch `psql` writes — as defense-in-depth, also open every session with `SET default_transaction_read_only = on;` (verified: writes then fail with `cannot execute INSERT in a read-only transaction`). This guard is per-session, so it only holds if the statement is actually issued.
```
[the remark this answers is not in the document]
theMickster: Addressed in d0f6eb843e1904383ca399a8374f033731f9e9b4

SaintPatrck (approved):

PR #7764, opened 2026-06-04
[PM-38265] Enhance TwoFactorIsEnabledQuery to use a dictionary for two-factor authentication results

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38265

## 📔 Objective

Replace a per-user linear scan over the 2FA results list with a dictionary built once before the loop, reducing the Members page load time


JaredSnider-Bitwarden (approved):
LGTM! Thank you!

PR #7766, opened 2026-06-04
Auth/PM-38588 - Login and Refresh backfill CurrentContext for LaunchDarkly Feature Flag Evaluation

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38588

## 📔 Objective

To backfill `CurrentContext` so that we can actually progressively rollout feature flags in the login and / or token fresh flows (which is currently failing b/c of missing `CurrentContext` data).


src/Identity/IdentityServer/RequestValidators/CustomTokenRequestValidator.cs
JaredSnider-Bitwarden, line 1: Note: this restructuring does allow some claim reading outside the bounds of the feature flag... but I think it is low risk and allows for a cleaner structure.

justindbaur (approved):
Looks good, I had tried to think of a path where refresh tokens are done for things like org api keys which could have been a good integration test to prove out the defensive-ness of the code but I don't believe we support refresh tokens for those (or I couldn't make it happen in our tests).

PR #7768, opened 2026-06-05
[PM-38613] Add timestamp for LastPasswordChangeDate

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38613

## 📔 Objective

When a user key and password change rotation happens we want to indicate a password change has happen by setting the LastPasswordChangeDate.


test/Core.Test/KeyManagement/UserKey/RotateUserAccountKeysCommandTests.cs
```diff
@@ -131,8 +131,15 @@ public async Task PasswordChangeAndRotateUserAccountKeysAsync_V1_Success(SutProv
         sutProvider.GetDependency<IUserService>().CheckPasswordAsync(user, model.OldMasterKeyAuthenticationHash)
             .Returns(true);
 
+        var before = DateTime.UtcNow;
         var result = await sutProvider.Sut.PasswordChangeAndRotateUserAccountKeysAsync(user, model);
+        var after = DateTime.UtcNow;
+
         Assert.Equal(IdentityResult.Success, result);
+        Assert.NotNull(user.LastPasswordChangeDate);
+        Assert.InRange(user.LastPasswordChangeDate.Value, before, after);
```
mzieniukbw, line 142: 🤔  Both are inclusive according to `Assert.InRange` docs and it's likely `user.LastPasswordChangeDate.Value` is identical to before or/and after.
How about we modify user `LastPasswordChangeDate` field to be in the past and use that as `before`, like so:  `user.LastPasswordChangeDate = DateTime.UtcNow.AddDays(-1);`.
(Could do the same about `LastKeyRotationDate`)

mzieniukbw (dismissed):

test/Core.Test/KeyManagement/UserKey/RotateUserAccountKeysCommandTests.cs
Thomas-Avery: Fixed in https://github.com/bitwarden/server/pull/7768/commits/1aa30b82e1d5cfeae89a6d7b8900499ef831fe76

mzieniukbw (approved):

quexten:
Marking as abandoned since @JaredSnider-Bitwarden is implementing the same here https://github.com/bitwarden/server/pull/7804. It will be tested and QA'd as part of https://github.com/bitwarden/server/pull/7804.

PR #7771, opened 2026-06-05
[bre-1835] update SeederApi Base Image

aj-bw:
## 🎟️ Tracking

[bre-1835](https://bitwarden.atlassian.net/browse/bre-1835)

## 📔 Objective
Replace SeederApi distroless base image with aspnet:10.0-noble

The distroless runtime image runs as a hardcoded non-root UID that conflicts with the UID-matching pattern used by other services in the stack. This caused SeederApi to be unable to read shared DataProtection key files at their default permissions, requiring a non-standard filesystem workaround to operate.

This PR swaps the runtime base image to aspnet:10.0-noble and adds the same entrypoint.sh + gosu pattern used by api and identity, so SeederApi runs as the correct UID and reads those files without any workaround.

 See BRE-1835 for full context.
  
Changes:
  - util/SeederApi/Dockerfile — runtime stage only, build stage untouched
  - util/SeederApi/entrypoint.sh — new file, mirrors the api/identity pattern

### DIFF
--- a/util/SeederApi/Dockerfile
+++ b/util/SeederApi/Dockerfile
@@ -89,7 +89,7 @@ RUN . /etc/build-env \
 ###############################################
 #                  App stage                  #
 ###############################################
-FROM mcr.microsoft.com/dotnet/aspnet:10.0-azurelinux3.0-distroless-extra AS app
+FROM mcr.microsoft.com/dotnet/aspnet:10.0-bookworm-slim AS app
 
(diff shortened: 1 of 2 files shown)


mandreko-bitwarden (approved):
Sad that we can't utilize distroless, but sometimes it'll be like that. This is still better than where we were. Progress!

mimartin12 (approved):

PR #7772, opened 2026-06-05
[PM-34053] Add new GET endpoint for fetching key data for user key rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34053

## 📔 Objective

Adds new GET endpoint for fetching all the keys required outside of the sync for a user key rotation.


src/Api/KeyManagement/Models/Responses/EmergencyAccessKeyDataResponseModel.cs
```diff
@@ -0,0 +1,32 @@
...
+    [SetsRequiredMembers]
+    public EmergencyAccessKeyDataResponseModel(EmergencyAccessKeyData data, string obj = _objectName) : base(obj)
+    {
+        ArgumentNullException.ThrowIfNull(data);
+
+        Id = data.Id;
+        GranteeId = data.GranteeId;
+        GranteeName = data.GranteeName;
+        GranteeEmail = data.GranteeEmail;
+        PublicKey = data.PublicKey;
```
quexten, line 20: Just a thought (non-blocking): We probably want to replace public key with a proper identity struct here, because v2 membership will not necessarily work with just a public key but the verifying key + signed public key.

quexten (approved):

src/Api/KeyManagement/Models/Responses/EmergencyAccessKeyDataResponseModel.cs
Thomas-Avery: In the future I think it will be possible for a user to be in both v1 memberships and v2 membership right? 

I would expect this will need to look like two arrays one of the v1 and one of the v2 in that case. Going to keep this out of scope for this work.

PR #7783, opened 2026-06-08
PM-33527 - Don't leave orphaned event logs in storage

prograhamming:
## 🎟️ Tracking
[PM-33527](https://bitwarden.atlassian.net/browse/PM-33527)

## 📔 Objective
In order to stay within GDPR compliance when a organization wants to be deleted we need to remove the event logs from Azure Table Storage. This PR is the database changes needed in order to delete the event logs using jobs and long running processes.

[PM-33527]: https://bitwarden.atlassian.net/browse/PM-33527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BTreston (approved):
Approved for AC files

src/Infrastructure.EntityFramework/Dirt/Models/OrganizationDeleteTask.cs
```diff
@@ -0,0 +1,16 @@
+using AutoMapper;
+
+namespace Bit.Infrastructure.EntityFramework.Dirt.Models;
+
+public class OrganizationDeleteTask : Core.Dirt.Entities.OrganizationDeleteTask
```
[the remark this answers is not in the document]
Banrion: Resolving since this is consistent with existing EF model convention, e.g. User : "Core.Entities.User"

src/Sql/dbo/Dirt/Stored Procedures/OrganizationDeleteTask_ClaimNextPending.sql
```diff
@@ -0,0 +1,45 @@
+CREATE PROCEDURE [dbo].[OrganizationDeleteTask_ClaimNextPending]
```
mkincaid-bw, line 1: The name of the proc doesn't follow the T-SQL [guidelines](https://contributing.bitwarden.com/contributing/code-style/sql/#stored-procedures). This should be named something like `OrganizationDeleteTask_UpdateClaimNextPending`.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
```diff
@@ -158,6 +161,30 @@ BEGIN
     WHERE
         [OrganizationId] = @Id
 
+    -- Atomically enqueue an OrganizationDeleteTask (e.g. for purging Table Storage
+    -- event logs) so downstream cleanup is durably recorded with the deletion.
+    IF @OrganizationDeleteTaskId IS NOT NULL
+    BEGIN
+        DECLARE @OrganizationDeleteTaskDate DATETIME2(7) = COALESCE(@OrganizationDeleteTaskCreationDate, SYSUTCDATETIME())
+
+        INSERT INTO [dbo].[OrganizationDeleteTask]
```
mkincaid-bw, line 170: Any reason why you don't just call `OrganizationDeleteTask_Create` here?

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
```diff
@@ -0,0 +1,27 @@
...
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    DECLARE @Total INT = 0
+    DECLARE @BatchSize INT = 1000
+
+    WHILE @BatchSize > 0
+    BEGIN
+        BEGIN TRANSACTION Event_DeleteManyByOrganizationId
```
mkincaid-bw, line 12: This explicit transaction is unnecessary and potentially harmful. Since the only thing affected by the transaction in this case is the `DELETE` statement, it's already guaranteed to be atomic (it's a single statement). However, with the explicit transaction and no error handling, it's possible that SQL Server could error on the statement and leave the transaction uncommitted.

If you want to guarantee that all records for an org are deleted (or none in the event of a failure), you'd need to move the `BEGIN`/`COMMIT` outside of the `WHILE` loop, and add a `TRY`/`CATCH` block to handle any errors. See [this](https://contributing.bitwarden.com/contributing/code-style/sql/#error-handling) for an example.

If you don't care about all or nothing, then the explicit transaction should be removed completely.

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
```diff
@@ -0,0 +1,27 @@
+CREATE PROCEDURE [dbo].[Event_DeleteManyByOrganizationId]
```
mkincaid-bw, line 1: ❓Is this proc being added mainly for parity or do you expected self-host clients to run it? The main reason I ask is because the functionality does not match cloud (where looping is handled in code vs in the DB). The first call to the proc from code will cause this to loop to completion, and this may timeout for large customers.

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
```diff
@@ -0,0 +1,27 @@
...
+    DECLARE @BatchSize INT = 1000
+
+    WHILE @BatchSize > 0
+    BEGIN
+        BEGIN TRANSACTION Event_DeleteManyByOrganizationId
+
+        DELETE TOP(@BatchSize)
+        FROM
+            [dbo].[Event]
+        WHERE
```
mkincaid-bw, line 17: There is no index on the `OrganizationId` column, so this delete batch will cause thousands of table scans for large customers. Based on my previous question (parity vs actual use), you may want to add a filtered index on `OrganizationId` where `OrganizationId IS NOT NULL`.

mkincaid-bw (changes requested):

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
prograhamming: Good catch — removed the explicit transaction completely in the follow-up PR #7910 (commit 3772c3acd). Since each batched `DELETE` is a single atomic statement and the resumable delete-task design intentionally wants partial progress to persist, all-or-nothing semantics weren't needed.

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
prograhamming: Added in PR #7910 (commit 3772c3acd): a filtered index `IX_Event_OrganizationId ON [dbo].[Event]([OrganizationId]) WHERE [OrganizationId] IS NOT NULL` per your suggestion, plus the EF Core equivalent (unfiltered, since MySQL doesn't support filtered indexes — same approach as `IX_Notification_OrganizationId` / `IX_SecurityTask_OrganizationId`) with generated migrations for all three EF providers.

src/Sql/dbo/Dirt/Stored Procedures/OrganizationDeleteTask_ClaimNextPending.sql
prograhamming: Renamed to `OrganizationDeleteTask_UpdateClaimNextPending` in PR #7910 (commit 3772c3acd), matching its siblings (`_UpdateCompleted` / `_UpdateError` / `_UpdateProgress`). The migration also drops the old proc name idempotently so dev databases that ran the earlier version don't keep an orphaned copy.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
prograhamming: This changed shape in the follow-up PR #7910: the scalar task parameters were replaced with a table-valued parameter (`OrganizationDeleteTaskArray`) and a set-based `INSERT … SELECT`, so any number of task types can be enqueued atomically in the same transaction as the delete. The single-row `OrganizationDeleteTask_Create` proc no longer fits that shape — it's still used for the standalone create path.

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
prograhamming: Self-host clients are expected to run it — cloud stays on the Table Storage implementation. Looping in the proc keeps the same contract as the other implementations ("delete until done" per call). On the timeout concern, PR #7910 mitigates it from a few directions: each batch is now a seek via the new `IX_Event_OrganizationId` filtered index rather than a table scan, the explicit transaction was removed so every batch auto-commits independently, and the caller uses a 3600s command timeout. If a very large customer still times out, progress persists and the durable delete-task queue simply re-claims and resumes on the next run.

src/Infrastructure.EntityFramework/Dirt/Repositories/EventRepository.cs
```diff
@@ -52,6 +52,15 @@ public async Task CreateManyAsync(IEnumerable<IEvent> entities)
...
     }
 
+    public async Task<int> DeleteManyByOrganizationIdAsync(Guid organizationId)
+    {
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+        return await dbContext.Events
+            .Where(e => e.OrganizationId == organizationId)
+            .ExecuteDeleteAsync();
+    }
```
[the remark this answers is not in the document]
prograhamming: Leaving this one as-is intentionally: `Take()` + `ExecuteDeleteAsync` doesn't translate across all three EF providers, and this path is only reachable on self-host EF deployments where a single set-based delete is acceptable. If it ever becomes a problem, the durable delete-task queue retries on failure, so a provider-specific batched implementation can be added later without contract changes.

Banrion:
Superseded by #8182

PR #7798, opened 2026-06-10
Add 3 more Org Scenes

bnagawiecki:
## 📔 Objective

- Update SingleOrganizationScene to respond with the decrypted org key for usage in subsequent org scenes
- Add OrganizationUserScene to allow for seeding of an orgUser in a specific role and status
- Add OrganizationCollectionScene to allow for adding of collection with user and group permissions
- Add OrganizationLoginCipherScene to allow the seeding of a login cipher to specific collections


util/Seeder/Scenes/OrganizationUserScene.cs
```diff
@@ -0,0 +1,83 @@
...
+        var requiresKey = request.OrganizationUserStatusType is OrganizationUserStatusType.Confirmed
+            or OrganizationUserStatusType.Revoked;
+
+        if (requiresKey && string.IsNullOrEmpty(user.PublicKey))
+        {
+            throw new InvalidOperationException(
+                $"User {request.UserId} has no public key; cannot encrypt the organization key for the member.");
+        }
+
+        var organization = await organizationRepository.GetByIdAsync(request.OrganizationId)
```
MGibson1: **issue**: Looks to me like you're probably null checking a `Task`. honestly, this null coalesce into a throw is probably more trouble than it's worth.

```suggestion
        var organization = (await organizationRepository.GetByIdAsync(request.OrganizationId))
```

util/Seeder/Scenes/OrganizationLoginCipherScene.cs
```diff
@@ -0,0 +1,105 @@
...
+        public required string Value { get; set; }
+        public required int Type { get; set; }
+    }
+
+    public class PasskeyRequest
+    {
+        public required string RpId { get; set; }
+        public required string RpName { get; set; }
+        public required string UserName { get; set; }
+    }
```
MGibson1: This should be subclassed, we have org details now, but, as is done with the `Cipher` class, that is represented by a Request subclassed to an `OrganizationRequest`. 

I would like a base class for all cipher requests, then subclasses namespaces to their respective scenes, for user and organization ciphers.

FieldRequest and PasskeyRequest don't need subclasses, but look like they can be associated with the base class

util/Seeder/Scenes/OrganizationLoginCipherScene.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    public class Result
+    {
+        public required Guid CipherId { get; init; }
+    }
+
+    public async Task<SceneResult<Result>> SeedAsync(Request request)
```
MGibson1: A lot of this logic is the same, too. An abstract base class may help ensure proper handling as these evolve. likely this class should only specify the request and create a `Cipher` object for the base class to save / manipulate.

util/Seeder/Scenes/OrganizationCollectionScene.cs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private static List<CollectionAccessSelection>? MapAccessSelections(IEnumerable<AccessSelectionRequest>? selections)
+        => selections?.Select(s => new CollectionAccessSelection
+        {
+            Id = s.Id,
+            ReadOnly = s.ReadOnly,
+            HidePasswords = s.HidePasswords,
+            Manage = s.Manage
+        }).ToList();
```
MGibson1: Looks like `IEnumerable` is accepted
```suggestion
    private static IEnumerable<CollectionAccessSelection>? MapAccessSelections(IEnumerable<AccessSelectionRequest>? selections)
        => selections?.Select(s => new CollectionAccessSelection
        {
            Id = s.Id,
            ReadOnly = s.ReadOnly,
            HidePasswords = s.HidePasswords,
            Manage = s.Manage
        });
```

MGibson1 (changes requested):
a few art comments and one issue to fix, please

util/Seeder/Scenes/OrganizationUserScene.cs
bnagawiecki: Updating this and all other null coalesces in these updates to mirror the pattern used in the other scenes

util/Seeder/Scenes/LoginCipherRequest.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Bit.Seeder.Scenes;
+
+/// <summary>
+/// Shared request fields for seeding a login cipher, regardless of owner (user or organization).
+/// </summary>
+public abstract class LoginCipherRequest
```
MGibson1: let's namespace this under `LoginCipherScene` for consistency

MGibson1 (changes requested):
one last new thing, and I think we're ready. Thank you!

util/Seeder/Scenes/LoginCipherRequest.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.ComponentModel.DataAnnotations;
 
-namespace Bit.Seeder.Scenes;
+namespace Bit.Seeder.Scenes.LoginCipherScene;
```
MGibson1: Sorry, not what I meant. I mean that the `LoginCipherRequest` should be under the `LoginCipherScene`:

```csharp
using System.ComponentModel.DataAnnotations;
using System.Text.Json;
using Bit.Core.Vault.Entities;
using Bit.Core.Vault.Enums;
using Bit.Seeder.Factories;
using Bit.Seeder.Models;
using Bit.Seeder.Services;

namespace Bit.Seeder.Scenes;

/// <summary>
/// Base scene for seeding a login cipher. Builds and persists the cipher from the shared
/// <see cref="LoginCipherRequest"/> fields, deferring the owner-specific bits (key resolution,
/// owner-only customizations, and the save call) to the derived scene.
/// </summary>
public abstract class LoginCipherScene<TRequest>(IManglerService manglerService)
    : IScene<TRequest, LoginCipherScene<TRequest>.Result>
    where TRequest : LoginCipherScene<TRequest>.LoginCipherRequest
{
    public abstract class LoginCipherRequest
    {
        [Required]
        public required Guid UserId { get; set; }
        [Required]
        public required string Name { get; set; }
        public string? Username { get; set; }
        public string? Password { get; set; }
        public string? Totp { get; set; }
        public string? Uri { get; set; }
        public string? Notes { get; set; }
        public bool Reprompt { get; set; }
        public bool Deleted { get; set; }
        public bool Favorite { get; set; }
        public Guid? FolderId { get; set; }
        public IEnumerable<FieldRequest>? Fields { get; set; }
        public IEnumerable<PasskeyRequest>? Passkeys { get; set; }

        public class FieldRequest
        {
            public required string Name { get; set; }
            public required string Value { get; set; }
            public required int Type { get; set; }
        }

        public class PasskeyRequest
        {
            public required string RpId { get; set; }
            public required string RpName { get; set; }
            public required string UserName { get; set; }
        }
    }

    public class Result
    {
        public required Guid CipherId { get; init; }
    }

    /// <summary>Owner resolution result: which key encrypts the cipher and which owner id it carries.</summary>
    protected record CipherOwner(string EncryptionKey, Guid? OrganizationId, Guid? UserId);

    /// <summary>Look up + validate the owner, returning the encryption key and owner ids.</summary>
    protected abstract Task<CipherOwner> ResolveOwnerAsync(TRequest request);

    /// <summary>Persist the cipher (with or without collection links).</summary>
    protected abstract Task PersistAsync(Cipher cipher, TRequest request);

    public async Task<SceneResult<Result>> SeedAsync(TRequest request)
    {
        var owner = await ResolveOwnerAsync(request);

        var cipher = LoginCipherSeeder.Create(new CipherSeed
        {
            Type = CipherType.Login,
            Name = request.Name,
            Notes = request.Notes,
            EncryptionKey = owner.EncryptionKey,
            OrganizationId = owner.OrganizationId,
            UserId = owner.UserId,
            Login = new LoginViewDto
            {
                Username = request.Username,
                Password = request.Password,
                Totp = request.Totp,
                Uris = string.IsNullOrEmpty(request.Uri) ? null : [new LoginUriViewDto { Uri = request.Uri }],
                Fido2Credentials = request.Passkeys?
                    .Select(p => LoginCipherSeeder.CreateFido2Credential(p.RpId, p.RpName, p.UserName)).ToList()
            },
            Fields = request.Fields?.Select(f => new FieldViewDto
            {
                Name = f.Name,
                Value = f.Value,
                Type = f.Type
            }).ToList()
        });

        if (request.Reprompt)
        {
            cipher.Reprompt = CipherRepromptType.Password;
        }
        if (request.Deleted)
        {
            cipher.DeletedDate = DateTime.UtcNow.AddDays(-1);
        }
        if (request.Favorite)
        {
            cipher.Favorites = JsonSerializer.Serialize(new Dictionary<string, bool>
            {
                { request.UserId.ToString().ToUpperInvariant(), true }
            });
        }
        if (request.FolderId.HasValue)
        {
            cipher.Folders = CipherComposer.BuildFoldersJson(new Dictionary<Guid, Guid>
            {
                { request.UserId, request.FolderId.Value }
            });
        }

        await PersistAsync(cipher, request);

        return new SceneResult<Result>(
            result: new Result { CipherId = cipher.Id },
            mangleMap: manglerService.GetMangleMap());
    }
}

```

MGibson1 (changes requested):

MGibson1 (approved):
Great! thank you

PR #7804, opened 2026-06-11
Auth/PM-38811 - KM - Update RotateUserAccountKeysCommand to use MasterPasswordService

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38811

## 📔 Objective

To finish consolidating KM usages of the `MasterPasswordService` which centralizes MP change logic in one location as part of the ongoing separation of MP salt & email work. 

#### Two flows - Change password without key rotation and then with key rotation

https://github.com/user-attachments/assets/cbbe2337-39b8-44cf-98c9-eadaf3198968


eligrubb (approved):
LGTM!

PR #7806, opened 2026-06-11
[DBOPS-177] Update MSSQL Docker image to SQL Server 2025 CU5

mkincaid-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/DBOPS-177

## 📔 Objective

Update SQL Server to version SQL Server 2025 CU5 for self-host.


mimartin12 (approved):

withinfocus (approved):
I'm always in support of upgrading! Confirmed this is the latest CU per https://mcr.microsoft.com/en-us/artifact/mar/mssql/server/tags.

PR #7809, opened 2026-06-12
[PM-31107] Seeder: blob-migration preset + fido2/password-history/linked-field fixture support

shane-melton:
## 🎟️ Tracking

[PM-31107](https://bitwarden.atlassian.net/browse/PM-31107)

## 📔 Objective

Adds Seeder support for testing the SDK V1→V2 blob-encryption key-rotation migration end-to-end.

Two parts:

**1. Fixture loader enhancements** — the cipher fixture schema already advertised `fido2Credentials`, `passwordHistory`, and field `linkedId`, but the deserialization model silently dropped them. The encryption DTOs already supported all three; this wires the fixture-side mapping so they're honored:
- `SeedModels.cs` — added the fields + `SeedFido2Credential`/`SeedPasswordHistory` records
- `CipherSeed.cs` — maps password history, linked-field IDs, and passkeys (key material is synthesized via the existing `CreateFido2Credential` helper; the fixture only supplies rp/user identifiers)
- `cipher.schema.json` — simplified `fido2Credential` to the honored identifiers

Any fixture can now seed passkeys, password history, and linked fields.

**2. New `individual.blob-migration` preset + `blob-migration` cipher fixture** — a premium V1 user whose personal vault holds one of every cipher type (login, card, identity, secure note, SSH key) plus a passkey, password history, custom fields (text/hidden/boolean/linked), a reprompt item, and a favorite. This maximizes coverage of the blob conversion's `type_data` + payload branches.

Usage: `dotnet run -- preset --name individual.blob-migration` → login `blobmigration@individual.example` / `asdfasdfasdf`.

[PM-31107]: https://bitwarden.atlassian.net/browse/PM-31107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/util/Seeder/Models/CipherSeed.cs
+++ b/util/Seeder/Models/CipherSeed.cs
@@ -1,4 +1,6 @@
-﻿using Bit.Core.Vault.Enums;
+﻿using System.Globalization;
+using Bit.Core.Vault.Enums;
+using Bit.Seeder.Factories;
 
 namespace Bit.Seeder.Models;
 
@@ -97,7 +99,7 @@ internal record CipherSeed
         Notes = item.Notes,
         Reprompt = item.Reprompt == 1 ? CipherRepromptType.Password : CipherRepromptType.None,
         Fields = MapFields(item.Fields),
-        Login = MapLogin(item.Login),
+        Login = MapLogin(item.Login, item.PasswordHistory),
         Card = MapCard(item.Card),
         Identity = MapIdentity(item.Identity),
         SecureNote = item.Type == "secureNote" ? new SecureNoteViewDto { Type = 0 } : null,
@@ -119,7 +121,8 @@ internal record CipherSeed
         {
             Name = f.Name,
             Value = f.Value,
-            Type = MapFieldType(f.Type)
+            Type = MapFieldType(f.Type),
+            LinkedId = f.LinkedId
         }).ToList();
 
     private static int MapFieldType(string type) => type switch
@@ -131,7 +134,7 @@ internal record CipherSeed
         _ => throw new ArgumentException($"Unknown field type: '{type}'", nameof(type))
     };
 
-    private static LoginViewDto? MapLogin(SeedLogin? login) =>
+    private static LoginViewDto? MapLogin(SeedLogin? login, List<SeedPasswordHistory>? passwordHistory) =>
         login == null ? null : new LoginViewDto
         {
             Username = login.Username,
@@ -141,8 +144,25 @@ internal record CipherSeed
             {
                 Uri = u.Uri,
                 Match = MapUriMatchType(u.Match)
-            }).ToList()
+            }).ToList(),
+            // Key material is synthesized; the fixture only supplies the relying-party/user identifiers.
+            Fido2Credentials = login.Fido2Credentials?.Select(f => LoginCipherSeeder.CreateFido2Credential(
+                f.RpId ?? "example.com",
+                f.RpName ?? f.RpId ?? "Example",
+                f.UserName ?? login.Username ?? "user")).ToList(),
+            PasswordHistory = MapPasswordHistory(passwordHistory)
         };
+
+    private static List<PasswordHistoryViewDto>? MapPasswordHistory(List<SeedPasswordHistory>? history) =>
+        history?.Select(p =>
+        {
+            var dto = new PasswordHistoryViewDto { Password = p.Password };
+            if (p.LastUsedDate is not null)
+            {
+                dto.LastUsedDate = DateTime.Parse(p.LastUsedDate, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+            }
+            return dto;
+        }).ToList();
 
     private static int MapUriMatchType(string match) => match switch
     {
--- a/util/Seeder/Models/CipherSeed.cs
+++ b/util/Seeder/Models/CipherSeed.cs
@@ -99,7 +99,7 @@ internal record CipherSeed
         Notes = item.Notes,
         Reprompt = item.Reprompt == 1 ? CipherRepromptType.Password : CipherRepromptType.None,
         Fields = MapFields(item.Fields),
-        Login = MapLogin(item.Login, item.PasswordHistory),
+        Login = MapLogin(item.Login),
         Card = MapCard(item.Card),
         Identity = MapIdentity(item.Identity),
         SecureNote = item.Type == "secureNote" ? new SecureNoteViewDto { Type = 0 } : null,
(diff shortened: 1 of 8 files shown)


util/Seeder/Seeds/schemas/cipher.schema.json
theMickster, line 193: ♻️ We have added a new model with `public required string Password { get; init; }` being required. Let's round of the refactor by updating the schema to enforce the same rule. 

```suggestion
"passwordHistory": {
  "type": "object",
  "additionalProperties": false,
  "required": ["password"],
  "properties": {
    "password": { "type": "string" },
    "lastUsedDate": { "type": "string", "description": "ISO date when password was last used." }
  }
}
```

util/Seeder/Models/CipherSeed.cs
```diff
@@ -131,7 +134,7 @@ internal void Validate()
         _ => throw new ArgumentException($"Unknown field type: '{type}'", nameof(type))
     };
 
-    private static LoginViewDto? MapLogin(SeedLogin? login) =>
+    private static LoginViewDto? MapLogin(SeedLogin? login, List<SeedPasswordHistory>? passwordHistory) =>
```
theMickster: ⚠️ / ♻️ Brief architectural note on the enhancements to `MapLogin` that I think we should change.

That new using Bit.Seeder.Factories; is the first time anything in Models/ reaches up into Factories/. Right now it's one-way: factories depend on Models, not the other way around. This flips that and drags ECDSA key synthesis into FromSeedItem, which is intended to only be a plain Seed → ViewDto map. 

I think that this is a straightforward fix..  `CreateFido2Credential` returns a `Fido2CredentialViewDto` (a Models type) and only needs BCL crypto + CoreHelpers —nothing from Factories. Let's move it down into Models and the using goes away. 

LoginCipherScene and CipherComposer already depend on Models, so they're fine.

<details>
<summary> A potential refactor vibed with Claude</summary>

New file — util/Seeder/Data/Generators/Fido2CredentialGenerator.cs:
```csharp
using System.Security.Cryptography;
using System.Text.Json;
using Bit.Core.Utilities;
using Bit.Seeder.Models;

namespace Bit.Seeder.Data.Generators;

/// <summary>
/// Generates a discoverable FIDO2 passkey with real ECDSA P-256 key material for test vault data.
/// Callers supply only the relying-party and user identifiers; all key material is synthesized.
/// </summary>
internal static class Fido2CredentialGenerator
{
    internal static Fido2CredentialViewDto Generate(string rpId, string rpName, string userName)
    {
        // ECDSA P-256 private key in PKCS#8 format
        using var ecdsa = ECDsa.Create(ECCurve.NamedCurves.nistP256);
        var keyValue = CoreHelpers.Base64UrlEncode(ecdsa.ExportPkcs8PrivateKey());

        // 16-byte random user handle, unpadded base64url
        var userHandleBytes = new byte[16];
        RandomNumberGenerator.Fill(userHandleBytes);
        var userHandle = CoreHelpers.Base64UrlEncode(userHandleBytes);

        return new Fido2CredentialViewDto
        {
            Discoverable = JsonSerializer.Serialize(true),
            CredentialId = JsonSerializer.Serialize(Guid.NewGuid()),
            KeyValue = keyValue,
            Counter = "0",
            RpId = rpId,
            RpName = rpName,
            UserHandle = userHandle,
            UserName = userName,
            UserDisplayName = userName,
        };
    }
}
```

Then in `Models/CipherSeed.cs` — drops the Factories using, stop synthesizing, carry the identifiers through
Then in `Factories/LoginCipherSeeder.cs` — delete the moved method + its now-dead usings, synthesize in Create:
```csharp
// Create(CipherSeed options), after building cipherView
        if (options.Login is not null && options.Fido2Credentials is { Count: > 0 })
        {
            cipherView.Login!.Fido2Credentials = options.Fido2Credentials
                .Select(f => Fido2CredentialGenerator.Generate(
                    f.RpId ?? "example.com",
                    f.RpName ?? f.RpId ?? "Example",
                    f.UserName ?? options.Login.Username ?? "user"))
                .ToList();
        }
```
The other two callers — already in Factories/Scenes, both can see Data.Generators (CipherComposer already imports it; the Scene needs the using added):
`Factories/CipherComposer.cs:56`

```csharp
? new List<Fido2CredentialViewDto> { Fido2CredentialGenerator.Generate(company.Domain, company.Name, username) }
```

`Scenes/LoginCipherScene.cs`  (+ add: `using Bit.Seeder.Data.Generators;`)
```csharp
.Select(p => Fido2CredentialGenerator.Generate(p.RpId, p.RpName, p.UserName)).ToList()
```


</details>

theMickster (dismissed):
Couple items to add/adjust to the PR

---
I think that we place a new document that briefly outlines the scenario in to `util/Seeder/Seeds/docs/scenarios/`. The document can be short and direct to start and then we link to it in the `util/Seeder/Seeds/docs/scenarios/README.md`. That allows for both humans & humans working with Claude can easily know what scenario to utilize. 

---
To add a new preset, we need to round out the work by updating `util/Seeder/Seeds/docs/presets.md`. The following appears to be what we'll need.

```
| blob-migration | Premium (1GB) | — | 7 (fixture) | — |
```
---
This pairs along with [this comment](https://github.com/bitwarden/server/pull/7809#discussion_r3427363702)

We need to nest the passwordHistory under login and not the seedVaultItem. 
1. The `SeedLogin` class should have the PasswordHistory list(it moved from `SeedVaultItem`).
```csharp
public List<SeedPasswordHistory>? PasswordHistory { get; init; }
```
2. CipherSeed.cs — MapLogin back to one arg and then password history reads from the login
```csharp
private static LoginViewDto? MapLogin(SeedLogin? login) =>

PasswordHistory = MapPasswordHistory(login.PasswordHistory)
```
3. `Seeds/schemas/cipher.schema.json` — move the property out of item and into the login $def. (Moved from ~lines 65-70 to under the `fido2Credentials` around line ~104)

util/Seeder/Seeds/schemas/cipher.schema.json
theMickster: Resolved in 56e8516de517711ae83de936941818dceee62fdb

util/Seeder/Models/CipherSeed.cs
theMickster: Resolved in 56e8516de517711ae83de936941818dceee62fdb

util/Seeder/Seeds/fixtures/ciphers/blob-migration.json
MGibson1, line 1: Seems like this should be in root and just always used. eventually maybe this is the seed for blob encrypted and not-blob-encrypted data? Having old vaults is part of the point of this project

MGibson1 (approved):

util/Seeder/Seeds/fixtures/ciphers/blob-migration.json
theMickster: Just so I'm tracking with you — by "in root and always used," do you mean making this the default cipher fixture rather than a named one a preset opts into?

PR #7782, opened 2026-06-08
[PM-37469] UseRiskInsights organization ability: migration, plan mapping, admin toggle, license, and endpoint guard

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37469

This PR consolidates all server-side subtasks of PM-37469 into one atomic change:

- [PM-37470](https://bitwarden.atlassian.net/browse/PM-37470) - data migration for existing Enterprise orgs
- [PM-37471](https://bitwarden.atlassian.net/browse/PM-37471) - map Enterprise plan to UseRiskInsights
- [PM-37473](https://bitwarden.atlassian.net/browse/PM-37473) - Admin Portal auto-tick
- [PM-37474](https://bitwarden.atlassian.net/browse/PM-37474) - self-host license VerifyData
- [PM-37475](https://bitwarden.atlassian.net/browse/PM-37475) - re-enable the Access Intelligence endpoint guard

> Heads-up on diff size: most of the +lines are auto-generated. Of the total additions, roughly 11,300 are EF migration snapshot files (the three `*.Designer.cs` files for MySQL/Postgres/SQLite that EF regenerates for the PM-37470 data migration). The actual hand-written, reviewable change is only a few hundred lines across the mapping, admin toggle, license, guard, and the migration scripts. The generated snapshots can be skimmed.

## 📔 Objective

The `UseRiskInsights` organization ability gates Access Intelligence but predates the org-ability pattern, so it was never reliably set on Enterprise orgs. This PR bundles every server-side piece so the migration, the forward-going mapping, the license check, and the re-enabled guard go live in a single atomic deploy, with no window where the guard rejects an org that has not yet been backfilled. The Billing riskInsights pricing key is already live in QA and prod, so the mapping takes effect on deploy.

Each subtask below links to its commits; the code is the source of truth.

### Data migration (PM-37470)
[...]

PR #7818, opened 2026-06-16
[PM-37474] Include UseRiskInsights in self-host license VerifyData

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37474

## 📔 Objective

`OrganizationLicense.VerifyData()` signed the `UseRiskInsights` claim into the self-host license but never validated it against the organization, leaving the last gap in the org-ability wiring for Access Intelligence on self-host. This adds the comparison using the conditional `HasClaim` pattern introduced in PM-33980 (`!claimsPrincipal.HasClaim(...) || claimValue == orgValue`), so that:

- new licenses correctly enforce the `UseRiskInsights` ability on self-host, and
- pre-existing self-host licenses generated before this claim existed continue to validate (the claim is absent, so the comparison is skipped) rather than being treated as invalid and disabling the organization.

This mirrors the existing `UseMyItems` and `UseInviteLinks` clauses in the same method. The claim is already emitted by `OrganizationLicenseClaimsFactory` and synced onto the organization by `Organization.UpdateFromLicense()`; `VerifyData()` was the only consumer missing it.

## 📸 Screenshots

Not applicable — server-side license validation, no UI changes.

## ⏰ Reminders before review

Unit tests cover the three branches of the new clause: claim absent (pre-existing license still validates), claim present and matching, and claim present and mismatching (validation fails).

### DIFF
--- a/src/Core/Billing/Organizations/Models/OrganizationLicense.cs
+++ b/src/Core/Billing/Organizations/Models/OrganizationLicense.cs
@@ -437,6 +437,7 @@ public class OrganizationLicense : ILicense
         var useDisableSmAdsForUsers = claimsPrincipal.GetValue<bool>(nameof(UseDisableSmAdsForUsers));
         var useMyItems = claimsPrincipal.GetValue<bool>(nameof(UseMyItems));
         var useInviteLinks = claimsPrincipal.GetValue<bool>(nameof(UseInviteLinks));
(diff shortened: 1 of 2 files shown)


AlexRubik:
Closing this draft. The self-host license VerifyData change is being consolidated with the rest of the PM-37469 subtasks into a single PR, #7782, so the whole org-ability change ships in one deploy. Work continues there.

PR #7821, opened 2026-06-16
[PM-32108] - Remove FF for org ability cache

jrmccannon:
## 🎟️ Tracking
[PM-32108](https://bitwarden.atlassian.net/browse/PM-32108)
[PM-32109](https://bitwarden.atlassian.net/browse/PM-32109)
## 📔 Objective
This removes the feature flag for the Org ability cache extended cache. Removing deprecated logic and the feature flag since this is not used client side.

[PM-32108]: https://bitwarden.atlassian.net/browse/PM-32108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-32109]: https://bitwarden.atlassian.net/browse/PM-32109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Services/Implementations/FeatureRoutedCacheService.cs
eliykat, line 1: This class only exists as a shim to route between old and new implementations of the cache (as the name suggests). This should be removed and the new implementation should be registered in DI in its place.

eliykat (changes requested):
Please also remove the old implementation of the cache, which should now be unused.

To recap, we should be removing
- `FeatureRoutedCacheService`
- `VCurrentInMemoryApplicationCacheService`
- its weird service bus stuff

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
jrmccannon: I do love touching every file :D

eliykat (approved):

src/Core/Services/Implementations/InMemoryServiceBusApplicationCacheService.cs
```diff
@@ -54,16 +54,6 @@ public override async Task DeleteOrganizationAbilityAsync(Guid organizationId)
         var task = _topicMessageSender.SendMessageAsync(message);
     }
```
eliykat, line 55: These can be removed - the old cache used service bus, the new one doesn't.

src/Core/HostedServices/ApplicationCacheHostedService.cs
```diff
@@ -142,12 +141,12 @@ private async Task ProcessMessageAsync(ServiceBusReceivedMessage message, Cancel
...
-                        await _applicationCacheService.BaseUpsertOrganizationAbilityAsync(upsertedOrg);
+                        await _organizationAbilityCacheService.UpsertOrganizationAbilityAsync(upsertedOrg, cancellationToken);
                     }
                     break;
                 case ApplicationCacheMessageType.DeleteOrganizationAbility:
-                    await _applicationCacheService.BaseDeleteOrganizationAbilityAsync(
-                        (Guid)message.ApplicationProperties["id"]);
+                    await _organizationAbilityCacheService.DeleteOrganizationAbilityAsync(
+                        (Guid)message.ApplicationProperties["id"], cancellationToken);
                     break;
```
eliykat, line 150: This is also supporting message bus infrastructure that we no longer need. Looks like this whole file can be deleted.

eliykat (changes requested):
Spotted some other files to be cleaned up. Also check out `IApplicationCacheServiceBusMessaging` and its implementations - they look suspect.

There's a bit of cleanup here so requesting a double check from @JimmyVo16 who implemented this.

src/Core/HostedServices/ApplicationCacheHostedService.cs
JimmyVo16: We have a [ticket](https://bitwarden.atlassian.net/browse/PM-33046) that will remove this file, so we should probably leave it alone to prevent any circular updates. It uses `BaseUpsertOrganizationAbilityAsync`, which only updates the L1 cache, unlike the regular `UpsertOrganizationAbilityAsync`, which updates both L1 and L2.

src/Core/Services/Implementations/FeatureRoutedCacheService.cs
JimmyVo16: We have a [ticket](https://bitwarden.atlassian.net/browse/PM-33045) to remove this file. 

I think we can remove the feature flag and leave the `Base*` methods to satisfy the `ApplicationCacheHostedService.cs` interface for the [reason](https://github.com/bitwarden/server/pull/7821/changes#r3436574154) I mentioned above.

In hindsight, we didn’t need the new cache in this method since it’s only being used for the service bus, so I think we’re safe to remove it.

src/Core/Services/Implementations/InMemoryServiceBusApplicationCacheService.cs
JimmyVo16: Yes, we have a [ticket](https://bitwarden.atlassian.net/browse/PM-33046) for this.

src/Core/HostedServices/ApplicationCacheHostedService.cs
JimmyVo16: It looks like we also injected the new cache in `BaseUpsertOrganizationAbilityAsync`, but it’s a non-issue. It would probably be safer to leave the service bus stuff for the other ticket, but I don’t think it’s a big deal either way.

JimmyVo16 (commented):
I think we should leave the service bus stuff for the other [ticket](https://bitwarden.atlassian.net/browse/PM-33046), but I’m happy to be convinced otherwise.

src/Core/HostedServices/ApplicationCacheHostedService.cs
jrmccannon: I could just remove these blocks now since i'm cleaning things up. Then later we could go and remove the entire file with the ticket that @JimmyVo16 mentioned.

jaasen-livefront (dismissed):
vault changes look good.

ike-kottlowski (dismissed):
Auth changes are good. Just a service swap and the tests are passing so I'm confident in the changes.

src/Core/HostedServices/ApplicationCacheHostedService.cs
eliykat: You've already started collecting approvals and we have a follow-up ticket, let's follow @JimmyVo16 's advice and leave any further changes.

eliykat (dismissed):
LGTM and we'll leave further service bus cleanup for the later ticket per @JimmyVo16 's advice.

src/Core/HostedServices/ApplicationCacheHostedService.cs
JimmyVo16: Sounds good

JimmyVo16 (dismissed):

kdenney (dismissed):

JimmyVo16 (approved):

ike-kottlowski (approved):

kdenney (approved):

eliykat (approved):

Banrion (approved):
✅ DIRT files

nick-livefront (approved):

PR #7825, opened 2026-06-16
[PM-39164] Gate Access Intelligence report reads on new-architecture flag

Banrion:
## 🎟️ Tracking

[PM-39164](https://bitwarden.atlassian.net/browse/PM-39164)

Related Client PR: https://github.com/bitwarden/clients/pull/21315

## 📔 Objective

This re-points the Access Intelligence report read path from the file-storage flag (`AccessIntelligenceVersion2`) to the new-architecture flag (`AccessIntelligenceNewArchitecture`). The file-storage flag should govern only the write format, so keying reads off it meant a report written as a file no longer loaded once file storage was toggled off, the read fell back to inline-only logic and returned a stale or broken "last report." This change fixes that so a file-backed report still loads in any valid flag state, while write-format selection remains file-flag-gated.

### Changes
- Gate `GetLatestOrganizationReportAsync` (read branch + validated-file download-URL block) on `AccessIntelligenceNewArchitecture` instead of the file-storage flag.
- Gate the `DownloadReportFileAsync` validated-file check on `AccessIntelligenceNewArchitecture`.
- Gate create file report paths on `AccessIntelligenceNewArchitecture` **and** the request shape  (`FileSize.HasValue`), rather than on the file-storage flag (`AccessIntelligenceVersion2`)
- Re-point the file write-lifecycle guards on `AccessIntelligenceNewArchitecture`
- Update/Replace the flag-driven create tests with new-architecture + shape-based tests.

### Additional Changes
- Remove the `[RequestSizeLimit(FileSize501mb)]` override on create (kept on the actual upload endpoint). It accepted large inline bodies the response echoes back, producing oversized responses; file-backed reports upload separately and never need it.

[PM-39164]: https://bitwarden.atlassian.net/browse/PM-39164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lastbestdev (dismissed):

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -98,13 +97,18 @@ public async Task<IActionResult> CreateOrganizationReportAsync(
...
 
-        if (_featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2))
+        // File storage only exists under the new architecture, so gate the file path on the (stable)
+        // AccessIntelligenceNewArchitecture flag. Within that, select the path from the request shape
+        // rather than the file-storage flag (AccessIntelligenceVersion2): the client chooses its shape
+        // from that flag via a config cache that can lag the server by up to an hour (pronounced in
+        // self-hosted), so branching on it here would 500 whenever the two disagree. Honoring the shape
+        // lets file-storage flag staleness degrade gracefully.
+        var isNewArchitecture = _featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceNewArchitecture);
+        if (isNewArchitecture && request.FileSize.HasValue)
```
lastbestdev, line 108: Just checking, creating the report with the NewArch flag on will still fallback to using inline data when the FileSize is not included in the request. This is what we want?

src/Api/Dirt/Controllers/OrganizationReportsController.cs
```diff
@@ -98,13 +97,18 @@ public async Task<IActionResult> CreateOrganizationReportAsync(
 
         await AuthorizeAsync(organizationId);
 
-        if (_featureService.IsEnabled(FeatureFlagKeys.AccessIntelligenceVersion2))
+        // File storage only exists under the new architecture, so gate the file path on the (stable)
+        // AccessIntelligenceNewArchitecture flag. Within that, select the path from the request shape
+        // rather than the file-storage flag (AccessIntelligenceVersion2): the client chooses its shape
+        // from that flag via a config cache that can lag the server by up to an hour (pronounced in
+        // self-hosted), so branching on it here would 500 whenever the two disagree. Honoring the shape
+        // lets file-storage flag staleness degrade gracefully.
```
lastbestdev, line 106: Not blocking, but do we need to include these comments? I'm not sure how much value these add when someone reading the code is faced with information about an implementation decision that was made a while ago.

It is good info, but IMO seems more like information to include in a commit message or the PR description

lastbestdev (approved):

PR #7831, opened 2026-06-17
[PM-38422] feat: Stack customer-level discount with churn and migration discounts

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38422

## 📔 Objective

When a churn-mitigation, proactive, or milestone-migration coupon is applied to a subscription that already carries a customer-level discount (`customer.discount`), the new coupon was written into the subscription's (or schedule phase's) `discounts` array on its own. Because Stripe lets a subscription/phase-level discount override the customer-level one, the pre-existing customer discount was silently dropped from billing.

This adds a shared `DiscountExtensions.MergeDiscountCouponIds` extension (customer coupon first, deduped by id with `StringComparison.Ordinal`, order-preserving) that merges the customer-level discount with the existing subscription/phase discounts and the newly applied coupon so they stack. It is applied wherever a `discounts` array is written and a pre-existing customer discount could be shadowed:

- Churn-mitigation redeem (churn-only subscription update + migration-cohort Phase 2)
- Premium / families / business price-increase scheduler Phase 2 builders
- Premium-storage and billing-address schedule rebuilds, and the worker `UpcomingInvoiceHandler` automatic-tax rebuild — **future phases only** (active phases are mirrored verbatim so a discount is not re-applied on the already-billed period)
- Org subscription-update schedule rebuild — future phase only
- `GetBitwardenSubscriptionQuery` dedups the carried coupon so the tax preview does not double-count it

The customer coupon is carried regardless of `Coupon.Valid` (an active `customer.discount` is one Stripe is already applying). Unblocks migration cohorts that include customers with pre-existing discounts. Unit tests cover the helper, the stacking/repro/self-heal cases, and the active-vs-future-phase boundary at each site.


kdenney (approved):
🔥 Very clean design; nice!

PR #7836, opened 2026-06-18
[PM-39237] fix: Handle mismatched and orphaned migration cohort assignments in admin dropdown

cyprain-okeke:
## 🎟️ Tracking

PM-39237

## 📔 Objective

The admin organization edit page builds the migration cohort dropdown by listing
only cohorts whose migration path matches the organization's current plan. When
an organization's existing assignment points to a cohort outside that filtered
set, the selected value was silently dropped — the dropdown rendered as
"(Not assigned)" or with the wrong selection, hiding the real assignment from
admins.

This PR ensures the currently-assigned cohort is always present in the dropdown
and clearly labeled:

- **Plan mismatch** — when the assigned cohort's migration path no longer matches
  the organization's plan (including a null/retired path id during a multi-stage
  rollout), the option is appended and labeled `(plan mismatch)`. The POST
  resolver already mirrors this guard, blocking a switch to a null-path cohort
  while letting the current value round-trip unchanged.
- **Orphaned assignment** — when the assigned cohort no longer exists in the
  repository, the dropdown shows `(assigned cohort no longer exists)` and a
  warning is logged (org id and cohort id only, no sensitive data).

Covered by 52 unit tests in `OrganizationsControllerTests` (all passing).


src/Admin/AdminConsole/Controllers/OrganizationsController.cs
```diff
@@ -255,6 +257,30 @@ public async Task<IActionResult> Edit(Guid id)
                 .ToList();
             currentAssignment =
                 await _organizationPlanMigrationCohortAssignmentRepository.GetByOrganizationIdAsync(id);
+
+            if (currentAssignment?.CohortId is { } assignedId
+                && visibleCohorts.All(c => c.Id != assignedId))
+            {
+                var assignedCohort = await _organizationPlanMigrationCohortRepository.GetByIdAsync(assignedId);
```
sven-bitwarden, line 264: Do we need to go to the database for this, or can we use the `migrationCohorts` list above? I assume the list would have the one we're looking for

sven-bitwarden (approved):
Approving with one comment that might be a minor optimization, but could also be unnecessary.

PR #7839, opened 2026-06-18
fix broken auto confirm policyRequirement check for org scoped request

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-39254

## 📔 Objective
**Root cause:** `autoConfirmRequirement.IsEnabled()` calls into the user-keyed `IPolicyRequirementQuery`, which looks up policy rows by `OrganizationUser.UserId` (for confirmed users) or `OrganizationUser.Email == User.Email` (for invited users). JIT SSO-provisioned org users have `Status=Invited`, `UserId=set`, `Email=null` satisfying neither branch, so zero rows are returned and `IsEnabled()` returns `false`. The auto-confirm push is then suppressed, leaving JIT users stuck in "Needs Confirmation".

**Fix:** The question being asked, "does this org have auto-confirm turned on?" is org-scoped, not user-scoped. `IPolicyQuery.RunAsync` does an org-keyed lookup (`GetByOrganizationIdTypeAsync`) which is immune to the broken user→OrgUser matching. This pattern is already used in `AutomaticallyConfirmOrganizationUsersValidator` for the same policy.

### DIFF
--- a/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptMembership/AcceptOrganizationMembershipValidator.cs
+++ b/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptMembership/AcceptOrganizationMembershipValidator.cs
@@ -1,4 +1,5 @@
-﻿using Bit.Core.AdminConsole.OrganizationFeatures.Policies;
+﻿using Bit.Core.AdminConsole.Enums;
+using Bit.Core.AdminConsole.OrganizationFeatures.Policies;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies.Enforcement.AutoConfirm;
 using Bit.Core.AdminConsole.OrganizationFeatures.Policies.PolicyRequirements;
 using Bit.Core.AdminConsole.Utilities.v2.Validation;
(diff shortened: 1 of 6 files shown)


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AcceptMembership/AcceptOrganizationMembershipValidator.cs
```diff
@@ -54,9 +56,12 @@ public async Task<ValidationResult<AcceptOrganizationMembershipValidationResult>
             }
         }
 
+        var autoConfirmPolicyStatus = await policyQuery.RunAsync(
+            request.OrganizationId, PolicyType.AutomaticUserConfirmation);
+
```
BTreston, line 61: We just want to flat out know if the organization itself has the policy enabled, not whether there is a valid enforcement path for the user as it exists at run time (because the user may be in an invalid state).

eliykat (dismissed):
This value has drifted too far from where it's used. It's not even used in the validator - it's passed out and only consumed by the command to conditionally call the `IPushAutoConfirmNotificationCommand`.

It's also incomplete - the check should be "do we need to send an autoconfirm push notification?" But that depends on:
- OrgAbility check
- Policy check (what you have here - I agree it's OK to use the policyQuery)
- Role check

Today we would have notifications going round trip only to be rejected by the autoconfirm command because the user isn't eligible.

My recommendation is to call the `IPushAutoConfirmNotificationCommand` unconditionally, but have early returns in that command the user isn't eligible. This should be noted on the xmldoc.

I think this is a relatively small refactor for a bug fix, and I think it's worth doing here because I had to do too much digging to figure out how this value is actually used.

BTreston:
> It's also incomplete - the check should be "do we need to send an autoconfirm push notification?" But that depends on:
> 
> * OrgAbility check
> * Policy check (what you have here - I agree it's OK to use the policyQuery)
> * Role check

@eliykat Should there be a role check to call the push notification? It already filters and goes out to admins/owners/manage user. From the caller perspective, its in an accept flow being initiated by users that would not have any of these roles yet anyway.

eliykat:
> > It's also incomplete - the check should be "do we need to send an autoconfirm push notification?" But that depends on:
> > 
> > * OrgAbility check
> > * Policy check (what you have here - I agree it's OK to use the policyQuery)
> > * Role check
> 
> @eliykat Should there be a role check to call the push notification? It already filters and goes out to admins/owners/manage user. From the caller perspective, its in an accept flow being initiated by users that would not have any of these roles yet anyway.

Sorry, I meant the role check on the member to be confirmed. We only autoconfirm members with the User-role.

sven-bitwarden (approved):

PR #7847, opened 2026-06-22
[PM-36965] feat: Add CSV export of cohort to cohort management table

cyprain-okeke:
## 🎟️ Tracking

[PM-36965](https://bitwarden.atlassian.net/browse/PM-36965) — [Determining Eligibility] Add .csv export of cohort to Cohort management table

## 📔 Objective

Extends the Admin Portal cohort management UI (PM-36951) with an **Export CSV** action that downloads every organization currently assigned to a migration cohort. This is an internal operator tool for auditing cohort contents and cross-referencing migration lists against external data.

**Endpoint:** `GET /migration-cohorts/{id}/export` → `text/csv` with `Content-Disposition: attachment`, gated by the existing `Tools_ManagePlanMigrationCohorts` permission and the `PM35215_BusinessPlanPriceMigration` feature flag (same gates as the rest of the cohort UI).

**Columns:** `OrganizationId, OrganizationName, AssignedAt, ScheduledDate, MigratedDate`.

**How it works:**
- The response is **streamed directly to the body** (not buffered), so large cohorts (thousands of orgs) export without loading every row into memory.
- A new repository method `GetExportRowsByCohortIdAsync` returns a **bounded keyset page** in both ORM tracks (Dapper stored proc + EF Core), ordered by `(CreationDate, Id)` so the new `(CohortId, CreationDate, Id)` index serves both the seek and the sort with no residual Sort operator — important because bulk-loaded cohorts share one `CreationDate` across the whole batch.
- A query object (`IExportCohortAssignmentsQuery`) yields rows as `IAsyncEnumerable`; the controller owns CSV formatting and HTTP streaming.

**Safety / operability:**
- Filename is slugged from the cohort name (guards `Content-Disposition` header injection); `OrganizationName` is defanged against CSV formula injection.
- A mid-stream failure logs with context and aborts the connection, so the operator gets a visibly broken download rather than a silently truncated CSV.
- Reads use the read-only connection string; an info-level audit log records actor + cohort id only (never org names or CSV contents).

**Testing:**
- Core query unit tests (paging/cursor advancement) and Admin controller tests (gates, headers, formula-injection, filename slug, audit log) — 44 unit tests.
- Repository integration tests run across **all four databases** (SQL Server, MySQL, PostgreSQL, SQLite): join surfaces the org name, keyset paging has no gaps/dupes including the all-shared-`CreationDate` bulk case.

**Notes for reviewers:**
- Open product question (non-blocking): final column set and whether export should be reachable from the table row vs. detail page — currently both; defaults implemented per the ticket, to confirm with the operator team.
- Row order in the CSV is not guaranteed identical across DB providers (each uses its native GUID order); the export contract is "every row, once," which is correct for a download.

## 📸 Screenshots

_Internal Admin Portal action (Export CSV link on the cohort list and detail pages); no new design surface — omitted._
Screenshot 2026-06-23 at 10 13 26 AM
[teams-2020-annually-to-current-2026-06-23.csv](https://github.com/user-attachments/files/29244021/teams-2020-annually-to-current-2026-06-23.csv)
Screenshot 2026-06-23 at 10 13 35 AM

[PM-36965]: https://bitwarden.atlassian.net/browse/PM-36965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
+++ b/src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
@@ -150,7 +150,7 @@ public class OrganizationPlanMigrationCohortsController(
 
         foreach (var header in new[]
                  {
-                     "OrganizationId", "OrganizationName", "AssignedAt", "ScheduledDate", "MigratedDate",
+                     "OrganizationId", "OrganizationName", "AssignedDate", "ScheduledDate", "MigratedDate",
                  })
         {
             csv.WriteField(header);
@@ -166,7 +166,7 @@ public class OrganizationPlanMigrationCohortsController(
             {
                 csv.WriteField(row.OrganizationId.ToString());
                 csv.WriteField(SanitizeCsvField(row.OrganizationName));
-                csv.WriteField(row.AssignedAt.ToString("o", CultureInfo.InvariantCulture));
+                csv.WriteField(row.AssignedDate.ToString("o", CultureInfo.InvariantCulture));
                 csv.WriteField(row.ScheduledDate?.ToString("o", CultureInfo.InvariantCulture) ?? string.Empty);
                 csv.WriteField(row.MigratedDate?.ToString("o", CultureInfo.InvariantCulture) ?? string.Empty);
                 await csv.NextRecordAsync();
@@ -180,6 +180,15 @@ public class OrganizationPlanMigrationCohortsController(
             }
 
             await csv.FlushAsync();
+        }
+        catch (OperationCanceledException ex) when (HttpContext.RequestAborted.IsCancellationRequested)
+        {
+            logger.LogInformation(ex,
+                "Cohort CSV export cancelled by the client. CohortId: {CohortId}, RowsWritten: {RowsWritten}",
+                cohort.Id,
+                rowsWritten);
+            aborted = true;
+            HttpContext.Abort();
         }
         catch (Exception ex)
         {
@@ -327,6 +336,10 @@ public class OrganizationPlanMigrationCohortsController(
         return $"{slug}-{DateTime.UtcNow:yyyy-MM-dd}.csv";
     }
 
+    // Defends against CSV formula injection: a spreadsheet treats a cell starting with =, +, -, @,
+    // tab, or CR as a formula, so an operator-controlled org name like "=cmd|'/c calc'!A1" could
+    // execute when the export is opened in Excel/Sheets. Prefixing a single quote forces the cell to
+    // render as literal text. https://owasp.org/www-community/attacks/CSV_Injection
     private static string SanitizeCsvField(string? value)
     {
         if (string.IsNullOrEmpty(value))
--- a/src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
+++ b/src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
@@ -1,4 +1,7 @@
 ﻿using System.ComponentModel.DataAnnotations;
+using System.Globalization;
+using System.Text;
+using System.Text.RegularExpressions;
 using Bit.Admin.Billing.Models.OrganizationPlanMigrationCohorts;
 using Bit.Admin.Enums;
 using Bit.Admin.Utilities;
@@ -8,6 +11,8 @@ using Bit.Core.Billing.Organizations.PlanMigration.Queries;
 using Bit.Core.Billing.Organizations.PlanMigration.Repositories;
 using Bit.Core.Billing.Services;
 using Bit.Core.Services;
+using CsvHelper;
+using CsvHelper.Configuration;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Stripe;
@@ -22,9 +27,15 @@ public class OrganizationPlanMigrationCohortsController(
     IStripeAdapter stripeAdapter,
     ILogger<OrganizationPlanMigrationCohortsController> logger,
     IFeatureService featureService,
-    IGetCohortAssignmentStateQuery getCohortAssignmentStateQuery) : Controller
+    IGetCohortAssignmentStateQuery getCohortAssignmentStateQuery,
+    IExportCohortAssignmentsQuery exportCohortAssignmentsQuery) : Controller
 {
     private const int _defaultPageSize = 25;
+
+    // How many CSV rows to buffer before flushing to the response stream during an export. Keeps
+    // bytes flowing on large cohorts so the connection doesn't idle into a server write-timeout.
+    // Intentionally independent of the export query's DB page size -- this only bounds buffering.
+    private const int _exportFlushEveryRows = 1000;
 
     private bool PlanMigrationCohortsFeatureEnabled() =>
         featureService.IsEnabled(FeatureFlagKeys.PM35215_BusinessPlanPriceMigration);
@@ -112,6 +123,86 @@ public class OrganizationPlanMigrationCohortsController(
 
         var assignmentState = await getCohortAssignmentStateQuery.Run(cohort);
         return View(EditCohortViewModel.From(cohort, CohortFormModel.From(cohort), assignmentState));
+    }
+
+    [HttpGet("{id:guid}/export")]
+    [RequirePermission(Permission.Tools_ManagePlanMigrationCohorts)]
+    public async Task<IActionResult> Export(Guid id)
+    {
+        if (!PlanMigrationCohortsFeatureEnabled()) return NotFound();
+
+        var cohort = await cohortRepository.GetByIdAsync(id);
+        if (cohort == null) return NotFound();
+
+        var fileName = BuildExportFileName(cohort.Name);
+
+        logger.LogInformation(
+            "Cohort CSV export started. Actor: {Actor}, CohortId: {CohortId}",
+            User?.Identity?.Name ?? "unknown",
+            cohort.Id);
+
+        Response.ContentType = "text/csv";
+        Response.Headers.ContentDisposition = $"attachment; filename=\"{fileName}\"";
+
+        var config = new CsvConfiguration(CultureInfo.InvariantCulture) { HasHeaderRecord = false };
+        var writer = new StreamWriter(Response.Body, new UTF8Encoding(false));
+        var csv = new CsvWriter(writer, config);
+
+        foreach (var header in new[]
+                 {
+                     "OrganizationId", "OrganizationName", "AssignedAt", "ScheduledDate", "MigratedDate",
+                 })
+        {
+            csv.WriteField(header);
+        }
+        await csv.NextRecordAsync();
+
+        var sinceFlush = 0;
+        var rowsWritten = 0;
+        var aborted = false;
+        try
+        {
+            await foreach (var row in exportCohortAssignmentsQuery.GetByCohortIdAsync(cohort.Id))
+            {
+                csv.WriteField(row.OrganizationId.ToString());
+                csv.WriteField(SanitizeCsvField(row.OrganizationName));
+                csv.WriteField(row.AssignedAt.ToString("o", CultureInfo.InvariantCulture));
+                csv.WriteField(row.ScheduledDate?.ToString("o", CultureInfo.InvariantCulture) ?? string.Empty);
+                csv.WriteField(row.MigratedDate?.ToString("o", CultureInfo.InvariantCulture) ?? string.Empty);
+                await csv.NextRecordAsync();
+                rowsWritten++;
+
+                if (++sinceFlush >= _exportFlushEveryRows)
+                {
+                    await csv.FlushAsync();
+                    sinceFlush = 0;
+                }
+            }
+
+            await csv.FlushAsync();
+        }
+        catch (Exception ex)
+        {
+            logger.LogError(ex,
+                "Cohort CSV export failed mid-stream after the response started. CohortId: {CohortId}, RowsWritten: {RowsWritten}",
+                cohort.Id,
+                rowsWritten);
+            aborted = true;
+            HttpContext.Abort();
+        }
+        finally
+        {
+            try
+            {
+                await csv.DisposeAsync();
+                await writer.DisposeAsync();
+            }
+            catch when (aborted)
+            {
+            }
+        }
+
+        return new EmptyResult();
     }
 
     [HttpPost("{id:guid}")]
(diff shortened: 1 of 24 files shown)


src/Infrastructure.EntityFramework/Billing/Repositories/OrganizationPlanMigrationCohortAssignmentRepository.cs
```diff
@@ -72,6 +72,37 @@ join c in dbContext.OrganizationPlanMigrationCohorts on a.CohortId equals c.Id
...
+        using var scope = ServiceScopeFactory.CreateScope();
+        var dbContext = GetDatabaseContext(scope);
+
+        var assignments = dbContext.OrganizationPlanMigrationCohortAssignments
+            .Where(a => a.CohortId == cohortId);
+
+        if (afterCreationDate != null)
+        {
+            assignments = assignments.Where(a =>
+                a.CreationDate > afterCreationDate.Value
```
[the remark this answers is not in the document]
cyprain-okeke: False positive — this access is inside the `if (afterCreationDate != null)` guard immediately above (the `.Value` is only reached when it's non-null). The paired `afterId!` is intentional: the cursor contract is both-null-or-both-set, enforced by the sole caller (`ExportCohortAssignmentsQuery`). No change needed.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -114,6 +125,86 @@ public async Task<IActionResult> Edit(Guid id)
...
+        logger.LogInformation(
+            "Cohort CSV export started. Actor: {Actor}, CohortId: {CohortId}",
+            User?.Identity?.Name ?? "unknown",
+            cohort.Id);
+
+        Response.ContentType = "text/csv";
+        Response.Headers.ContentDisposition = $"attachment; filename=\"{fileName}\"";
+
+        var config = new CsvConfiguration(CultureInfo.InvariantCulture) { HasHeaderRecord = false };
+        var writer = new StreamWriter(Response.Body, new UTF8Encoding(false));
```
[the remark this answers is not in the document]
cyprain-okeke: False positive — `writer` and `csv` are disposed explicitly in the `finally` block (`await csv.DisposeAsync()` / `await writer.DisposeAsync()`). They're intentionally not declared with `await using`: disposal flushes buffered bytes, and after a mid-stream `HttpContext.Abort()` that flush throws, so we dispose manually to contain the throw rather than let it mask the real error.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -114,6 +125,86 @@ public async Task<IActionResult> Edit(Guid id)
...
+        var csv = new CsvWriter(writer, config);
```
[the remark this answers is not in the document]
cyprain-okeke: False positive — see the reply above on the StreamWriter: `csv` is disposed in the `finally` block. Manual disposal is deliberate to contain a post-abort flush throw.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -114,6 +125,86 @@ public async Task<IActionResult> Edit(Guid id)
...
+        finally
+        {
+            try
+            {
+                await csv.DisposeAsync();
+                await writer.DisposeAsync();
+            }
+            catch when (aborted)
+            {
+            }
```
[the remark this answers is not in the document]
cyprain-okeke: Addressed in 871ab8bed — the `finally` catch is now `catch (Exception ex) when (aborted)` with a `LogDebug` explaining the swallowed post-abort disposal flush, so it's no longer empty and is diagnosable.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -114,6 +125,86 @@ public async Task<IActionResult> Edit(Guid id)
...
+        }
+        catch (Exception ex)
+        {
+            logger.LogError(ex,
+                "Cohort CSV export failed mid-stream after the response started. CohortId: {CohortId}, RowsWritten: {RowsWritten}",
+                cohort.Id,
+                rowsWritten);
+            aborted = true;
+            HttpContext.Abort();
+        }
```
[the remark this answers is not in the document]
cyprain-okeke: The main streaming catch is intentionally broad and now carries a comment explaining why: once the 200 + headers are committed, *any* failure (DB read, serialization, write) must take the same path — log + abort — because we can't convert it into an error response. Narrowing to IOException/OperationCanceledException would let a DB-layer exception escape after headers are sent, which reintroduces the silent-truncation failure mode this catch exists to prevent. Keeping it broad by design.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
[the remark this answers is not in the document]
cyprain-okeke: Addressed in 871ab8bed — the `finally` catch is now typed (`catch (Exception ex) when (aborted)`) and logs at debug. Kept the `when (aborted)` filter so it only suppresses cleanup failures on the abort path.

test/Admin.Test/Billing/Controllers/OrganizationPlanMigrationCohortsControllerTests.cs
```diff
@@ -698,4 +701,308 @@ await sutProvider.GetDependency<IOrganizationPlanMigrationCohortRepository>()
...
+        // The raw, unprefixed formula must NOT appear at the start of a field.
+        Assert.DoesNotContain(",=cmd", content);
+    }
+
+    [Theory]
+    [BitAutoData("=danger", true)]
+    [BitAutoData("+danger", true)]
+    [BitAutoData("-danger", true)]
+    [BitAutoData("@danger", true)]
+    [BitAutoData("Acme Corp", false)]
```
amorask-bitwarden, line 937: ⛏️ `SanitizeCsvField` guards six characters (`= + - @ \t \r`) but this theory only exercises four. Can you complete the full set of checks?

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -214,6 +313,32 @@ public async Task<IActionResult> Delete(Guid id)
...
+    }
+
+    private static string SanitizeCsvField(string? value)
+    {
+        if (string.IsNullOrEmpty(value))
+        {
+            return string.Empty;
+        }
+
+        return value[0] is '=' or '+' or '-' or '@' or '\t' or '\r'
```
amorask-bitwarden, line 350: ⛏️ This is the only security-relevant portion in the implementation and it's undocumented. Can we add a one-liner naming the threat we're defending against?

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -114,6 +125,94 @@ public async Task<IActionResult> Edit(Guid id)
...
+                if (++sinceFlush >= _exportFlushEveryRows)
+                {
+                    await csv.FlushAsync();
+                    sinceFlush = 0;
+                }
+            }
+
+            await csv.FlushAsync();
+        }
+        catch (Exception ex)
```
amorask-bitwarden, line 193: ❌  This catch/abort block isn't covered by tests, but backs the PR's guarantee that the user isn't going to get an incomplete CSV file. Can we get a test where `IExportCohortAssignmentsQuery.GetByCohortIdAsync` yields a row then throws so we can assert on `HttpContext.Abort` and (optionally) the `LogError` fired? You can probably use a `DefaultHttpContext` subclass and override `Abort()` to just flip a `bool` flag or something along those lines.

src/Admin/Billing/Controllers/OrganizationPlanMigrationCohortsController.cs
```diff
@@ -114,6 +125,94 @@ public async Task<IActionResult> Edit(Guid id)
...
+        {
+            // Deliberately broad: once the 200 + headers are committed we cannot convert any failure
+            // (DB read, serialization, write) into an error response, so every failure mode must take
+            // the same path -- log it and abort the connection so the operator gets a visibly broken
+            // download instead of a silently truncated CSV that looks complete.
+            logger.LogError(ex,
```
amorask-bitwarden, line 199: ⛏️ If a user cancels the download, we'll get an `OperationCancelledException`, but we'll still log it as an error in DataDog. Not a huge deal, but could cause concern if we're looking at logs. Consider catching that exception separately or pivoting your log severity on `HttpContext.RequestAborted.IsCancellationRequested`.

src/Infrastructure.Dapper/Billing/Repositories/OrganizationPlanMigrationCohortAssignmentRepository.cs
```diff
@@ -40,6 +40,25 @@ public async Task<int> GetCohortNonPendingAssignmentsCountAsync(Guid cohortId)
             commandType: CommandType.StoredProcedure);
     }
 
+    public async Task<IReadOnlyList<CohortAssignmentExportRow>> GetExportRowsByCohortIdAsync(
+        Guid cohortId, DateTime? afterCreationDate, Guid? afterId, int take)
+    {
```
amorask-bitwarden, line 45: ❌ Allowing `afterCreationDate` and `afterId` to both be nullable will allow some illegal states through to the DB layer. I think we should add a defensive check:
```csharp
if (afterCreationDate is null != (afterId is null))
    throw new ArgumentException("afterCreationDate and afterId must both be set or both be null.");
```

src/Core/Billing/Organizations/PlanMigration/Models/CohortAssignmentExportRow.cs
```diff
@@ -0,0 +1,21 @@
...
+/// and the Admin controller is responsible for CSV formatting and HTTP streaming. Contains no
+/// Vault Data -- only org-level operational metadata.
+/// </summary>
+/// <param name="Id">
+/// The assignment <c>Id</c>. Used only as the keyset-paging tiebreaker; it is NOT written to the CSV.
+/// </param>
+/// <param name="OrganizationName">The organization's display name (joined from the Organization table).</param>
+/// <param name="AssignedAt">
+/// The date the organization was assigned to the cohort (the assignment <c>CreationDate</c>).
+/// </param>
```
amorask-bitwarden, line 15: ⛏️ Can we finish the rest of the parameter documentation here?

src/Infrastructure.EntityFramework/Billing/Repositories/OrganizationPlanMigrationCohortAssignmentRepository.cs
```diff
@@ -72,6 +72,37 @@ join c in dbContext.OrganizationPlanMigrationCohorts on a.CohortId equals c.Id
         return await query.CountAsync();
     }
 
+    public async Task<IReadOnlyList<CohortAssignmentExportRow>> GetExportRowsByCohortIdAsync(
+        Guid cohortId, DateTime? afterCreationDate, Guid? afterId, int take)
+    {
```
amorask-bitwarden, line 77: ❌ Here as well: https://github.com/bitwarden/server/pull/7847/changes#r3461149782

src/Core/Billing/Organizations/PlanMigration/Models/CohortAssignmentExportRow.cs
```diff
@@ -0,0 +1,21 @@
...
+public record CohortAssignmentExportRow(
+    Guid Id,
+    Guid OrganizationId,
+    string OrganizationName,
+    DateTime AssignedAt,
+    DateTime? ScheduledDate,
```
amorask-bitwarden, line 23: ⛏️ Let's keep these consistent. Either use the `At` or `Date` suffix for all 3.

src/Infrastructure.Dapper/Billing/Repositories/OrganizationPlanMigrationCohortAssignmentRepository.cs
```diff
@@ -40,6 +40,25 @@ public async Task<int> GetCohortNonPendingAssignmentsCountAsync(Guid cohortId)
...
+        await using var connection = new SqlConnection(ReadOnlyConnectionString);
```
amorask-bitwarden, line 51: ❓ I think a check with DB Ops on whether readonly is the right call here is warranted. We'll need to see what kind of timing we have on eventual consistency with our read replica. We don't want users seeing cohort data in the Admin Portal that's inaccurate in the CSV download.

amorask-bitwarden (changes requested):
Great work - few things we should take a look at.

src/Infrastructure.Dapper/Billing/Repositories/OrganizationPlanMigrationCohortAssignmentRepository.cs
amorask-bitwarden: I see this was changed even though I formulated it as a question. Since @bitwarden/dept-dbops will have to review this PR anyway, I'd like to get their perspective on whether we can use the readonly connection string here to prevent load on the primary DB.

cyprain-okeke: Reverted back to ReadOnlyConnectionString — agreed, let's get @bitwarden/dept-dbops's read on whether the replica is fine here before changing it. I am happy to switch to primary if they'd prefer.

mkincaid-bw (approved):
LGTM

src/Infrastructure.Dapper/Billing/Repositories/OrganizationPlanMigrationCohortAssignmentRepository.cs
rkac-bw: Read replica lag over last week obtained from azure portal - observed p99 0.021s, max 0.49s
ReadOnlyConnectionString is the right call imo, single secondary, sub-second lag, keyset stable (CreationDate, Id,  both columns are immutable after insert), and it offloads the primary

amorask-bitwarden (approved):

PR #7853, opened 2026-06-23
[PM-38922] - Adding Problem Details Result

jrmccannon:
## 🎟️ Tracking
[PM-38922](https://bitwarden.atlassian.net/browse/PM-38922)

## 📔 Objective
This work will add the Problem Details result with a custom result type that can be returned via our HTTP API endpoints. It is split out into a separate Project so that it can be consumed as a dependency individually with minimal other packages included.

It wraps the existing `Problem` typed result from `Microsoft.AspNetCore.Http.HttpResults` and allows for us to provide a richer error message including the property that has the error, the i18n key used for translations, and the human readable error message to give context on the server.

This is intended for use with server side form validation for an enhanced user experience

## 📸 Screenshots

[PM-38922]: https://bitwarden.atlassian.net/browse/PM-38922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


justindbaur (dismissed):

justindbaur (approved):

src/Api/AdminConsole/Controllers/ValidationErrorTypedResultsExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Bit.Core.AdminConsole.Utilities.v2.Validation;
+
+namespace Microsoft.AspNetCore.Http.HttpResults;
+
+public static class ValidationErrorTypedResultsExtensions
+{
+    extension(TypedResults)
```
sven-bitwarden, line 7: Image

sven-bitwarden (approved):

PR #7856, opened 2026-06-24
[PM-39443] pin automapper dependency to under v15

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39443

## 📔 Objective

AutoMapper 15+ changed to a commercial license, so we stay on the last MIT-licensed major version (14.x). Adjust Renovate configuration to pin AutoMapper to below v15


harr1424 (dismissed):

.github/renovate.json5
```diff
@@ -122,7 +122,7 @@
     },
     {
       matchPackageNames: [
-        "AutoMapper.Extensions.Microsoft.DependencyInjection",
```
withinfocus, line 125: ❓ Do we still use this other package?

itsadrago: We do. I missed that and so I added the older package name back in

harr1424 (approved):

withinfocus (approved):

PR #7857, opened 2026-06-24
llm(rules): Auth tokenables must be minted via factory.

enmande:
## 🎟️ Tracking

N/A

## 📔 Objective

Auth team wants to standardize going forward: all Auth tokenables must be minted via a factory to ensure they are always injected with the configured lifetime and a static validator, following the `OrgUserInviteTokenableFactory` [pattern](https://github.com/bitwarden/server/blob/main/src/Core/Auth/Models/Business/Tokenables/OrgUserInviteTokenableFactory.cs).

A _rule_ is chosen here over a domain-scoped CLAUDE.md following [contributing guidelines](https://github.com/bitwarden/server/blob/main/.claude/CONTRIBUTING.md#2-shape--how-should-claude-use-it) _(State a rule that only applies to certain file globs_, in this case Auth-domain `Tokenable`s).

The rule is justified by the prior art: the vast majority of Auth `Tokenable`s historically use constructor instantiation only; in recent engagements, Claude has been reluctant to follow the factory pattern, even with steering, due to the volume of prior art. This ensures new additions follow the desired pattern.

### DIFF
--- a/.claude/rules/auth-tokenables.md
+++ b/.claude/rules/auth-tokenables.md
@@ -1,0 +1,24 @@
+---
+paths:
+  - "src/Core/Auth/Models/Business/Tokenables/**"
+---
+
+# Auth tokenables: mint through a factory
+
+New tokenables in this directory MUST be minted through a dedicated factory — never constructed with `new` at the
+call site. A token's expiry and validation are security-relevant: routing every mint through one factory guarantees the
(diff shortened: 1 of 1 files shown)


JaredSnider-Bitwarden (approved):
LGTM! Thank you!

SaintPatrck (approved):

PR #7859, opened 2026-06-24
[PM-37954] Add Command for Provisioning Staged OrganizationUsers

sven-bitwarden:
## 🎟️ Tracking
[PM-37954](https://bitwarden.atlassian.net/browse/PM-37954)

## 📔 Objective
Staged users need merely a fraction of the validation present for invited users - they're lightweight placeholders.  Their validation will take place when the are promoted from Staged -> Invited (far future), or Staged -> Accepted (via invite link).

[PM-37954]: https://bitwarden.atlassian.net/browse/PM-37954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/IProvisionStagedOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Bit.Core.AdminConsole.Entities;
+using Bit.Core.Entities;
+using Bit.Core.Enums;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Provision;
+
+/// <summary>
+/// Provisions organization users in <see cref="Bit.Core.Enums.OrganizationUserStatusType.Staged"/> status.
+/// Staged users are placeholders created by automated directory sync (SCIM / Directory Connector): they do not
+/// yet consume a seat, are not subject to organization policies, and have not been sent an invitation.
```
eliykat: I think these 2 lines belong on the enum definition rather than here - it is defining what a Staged user is. (While you're at it - can you add documentation to the other enum values as well?)

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/IProvisionStagedOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <param name="users">
+    /// The (email, externalId) pairs to stage. Emails already present in the organization, and duplicate
+    /// emails within the batch, are skipped.
+    /// </param>
+    /// <param name="eventSystemUser">The automated system performing the provisioning, used for event attribution.</param>
+    /// <returns>The created Staged organization users. Empty if every email was already present.</returns>
+    Task<ICollection<OrganizationUser>> ProvisionStagedOrganizationUsersAsync(
+        Organization organization,
+        IEnumerable<(string Email, string ExternalId)> users,
+        EventSystemUser eventSystemUser);
```
eliykat: Command conventions (undocumented, sorry! working to fix this):
- method should be called RunAsync
- should use a dedicated request object
- should return a v2 CommandResult

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/IProvisionStagedOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+using Bit.Core.Enums;
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.Provision;
+
+/// <summary>
+/// Provisions organization users in <see cref="Bit.Core.Enums.OrganizationUserStatusType.Staged"/> status.
+/// Staged users are placeholders created by automated directory sync (SCIM / Directory Connector): they do not
+/// yet consume a seat, are not subject to organization policies, and have not been sent an invitation.
+/// </summary>
+public interface IProvisionStagedOrganizationUsersCommand
```
eliykat: (non-blocking, not strongly held): sticking to CRUD verbs might be clearer than Provision. e.g. `CreateStagedOrganizationUserCommand`.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/ProvisionStagedOrganizationUsersCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            organizationUsersToCreate.Add(new OrganizationUser
+            {
+                OrganizationId = organization.Id,
+                UserId = null,
+                Email = email.ToLowerInvariant(),
+                Key = null,
+                Type = OrganizationUserType.User,
+                Status = OrganizationUserStatusType.Staged,
+                // StatusNew is intentionally left null - it is only populated by the revoke flow when
```
eliykat: when?  _when???_ :o

test/Infrastructure.IntegrationTest/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/ProvisionStagedOrganizationUsersCommandTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    [DatabaseTheory, DatabaseData]
+    public async Task ProvisionStagedOrganizationUsersAsync_CreatesStagedMembers(
+        IOrganizationRepository organizationRepository,
+        IOrganizationUserRepository organizationUserRepository)
+    {
+        var organization = await organizationRepository.CreateTestOrganizationAsync();
+
+        // The command is a thin orchestration over the repository; event emission is covered by unit tests,
+        // so a no-op event service keeps this integration test focused on real database persistence.
+        var sut = new ProvisionStagedOrganizationUsersCommand(organizationUserRepository, new NoopEventService());
```
eliykat: This project is for integration testing against the database, so it should call repositories directly, not Core methods. Here you're not adding a new repository method, you're re-using `CreateManyAsync`, so no provision-specific db tests are required. That said, you could make sure that `CreateManyAsync` is tested.

When you add the api endpoint, you can also add api integration tests which will hit the whole stack.

test/Infrastructure.IntegrationTest/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/CreateStagedOrganizationUsersCommandTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            users.Select(u => u.Email.ToLowerInvariant()).OrderBy(email => email),
+            result.Select(organizationUser => organizationUser.Email).OrderBy(email => email));
+        Assert.Equal(
+            users.Select(u => u.ExternalId).OrderBy(externalId => externalId),
+            result.Select(organizationUser => organizationUser.ExternalId).OrderBy(externalId => externalId));
+
+        // Each row is actually persisted as a Staged member with the expected shape.
+        foreach (var created in result)
+        {
+            var persisted = await organizationUserRepository.GetByIdAsync(created.Id);
```
eliykat: There is a `GetManyAsync` method.

eliykat (changes requested):

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/IProvisionStagedOrganizationUsersCommand.cs
sven-bitwarden: They're already there (on the enum), but I did remove the extra details from this guy

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/IProvisionStagedOrganizationUsersCommand.cs
sven-bitwarden: Can do 👍  not fussed one way or the other

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/ProvisionStagedOrganizationUsersCommand.cs
sven-bitwarden: cliffhanger deleted: it just be populated by the revoke flow 😆

test/Infrastructure.IntegrationTest/AdminConsole/OrganizationFeatures/OrganizationUsers/Provision/ProvisionStagedOrganizationUsersCommandTests.cs
sven-bitwarden: I can totally add the API tests later and remove these, but it does feel like we're missing a capability to test commands. Repositories are generally the agnostic building blocks, but the parameters and order-of-operations validity (the broad business logic) comes from the commands

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/StagedUsers/CreateStagedOrganizationUsersCommand.cs
```diff
@@ -1,52 +1,44 @@
...
-        foreach (var (email, externalId) in requestedUsers)
-        {
-            if (!existingEmails.Add(email))
-            {
-                continue;
-            }
-
-            organizationUsersToCreate.Add(new OrganizationUser
+        var organizationUsersToCreate = requestedUsers
+            .Where(w => existingEmails.Add(w.Email))
```
eliykat, line 28: Galaxy brain usage of `Add` to both accumulate the emails and ignore duplicates. :o
That said, you don't actually do anything with `existingEmails` later - do you need to add them at all, or are you just skipping duplicates? It could be:

```c#
 .Where(w => !existingEmails.Contains(w.Email))
```

PS. linq forever, for loops never

eliykat (changes requested):
LGTM, nice and simple, just 1 question.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/StagedUsers/CreateStagedOrganizationUsersCommand.cs
sven-bitwarden: Maybe it's too clever given it's doing a couple of things 🙊 

We have two sets of emails to deduplicate against - the incoming set, and the existing set. The code creates a HashSet of the existing ones, and then iterates over the incoming set _where_ we were able to add an element to the accumulated set, so first-in wins from the request body. If we just did that `.Where`, we could have duplicates coming in from the incoming set

If I wanted to make this more plain, I could do -
```cs
var orgUsersToCreate = requestedUsers
  .DistinctBy(d => d.Email)
  .Where(w => !existingEmails.Contains(w.Email))
  ...
```

Also yes Linq is love

eliykat: Oh that's true. I think your alternative here is slightly clearer (at least to me) because it separates "deduplicate new emails against each other" and "deduplicate new emails against existing emails" as different steps. _But_ I don't think there's anything wrong with what you've done either.

Thanks for clarifying!

eliykat (approved):

lastbestdev (approved):

PR #7860, opened 2026-06-24
add AccessSecretsManager to OrganizationUserScene

nthompson-bitwarden:
## 📔 Objective
- Add AccessSecretsManager to OrganizationUserScene
- Resolve sonar issue on System.Exception

### DIFF
--- a/util/Seeder/Scenes/OrganizationUserScene.cs
+++ b/util/Seeder/Scenes/OrganizationUserScene.cs
@@ -37,6 +37,7 @@ public class OrganizationUserScene(
         [Required]
         public required OrganizationUserStatusType OrganizationUserStatusType { get; set; }
         public Permissions? Permissions { get; set; }
+        public bool AccessSecretsManager { get; set; }
     }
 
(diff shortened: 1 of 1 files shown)


MGibson1 (commented):
Does this not need to ensure org has enough seats to avoid deactivation?

For quick tests this might work, but for persistent seeds it'll create an invalid org.

nthompson-bitwarden:
> Does this not need to ensure org has enough seats to avoid deactivation?
> 
> For quick tests this might work, but for persistent seeds it'll create an invalid org.

Yep! Was working with brad on that which is handled here: https://github.com/bitwarden/server/pull/7861 - _this_ PR should be merged after 7861, which I believe will handle your concern

MGibson1:
Looking at that PR, that'll work, but I'm curious about the long term plans on these org users scenes. Are we assuming that the org will be set up with the right perms and seat count for a complete test?

nthompson-bitwarden:
that is the assumption, yeah

MGibson1 (approved):

PR #7864, opened 2026-06-24
[PM-37514] Support Teams 2019 Migration

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37514

## 📔 Objective

Adds the **Teams 2019 → Teams Current** pricing-migration paths (Monthly and Annual) to the deferred price-increase migration engine (epic PM-37496).

"Teams 2019 Packaged" is a flat base covering **5 included seats**, plus a per-seat overage for seats beyond 5. Current Teams is pure per-seat.

- Orgs with **< 5** occupied seats are billed for actual occupied seats — unused base headroom disappears.
- Orgs with **≥ 5** keep their current purchased seat count.

Unlike Teams Starter (a flat bundle), Teams 2019 carries a per-seat overage line, so `HasNonSeatBasedPasswordManagerPlan()` is **false** for it — that distinction drives most of the changes below.

### Code changes

- **Paths & policy** — registers `Teams2019AnnualToCurrent` / `Teams2019MonthlyToCurrent` (append-only `MigrationPathId`s 9, 10; 7–8 reserved for the in-flight Teams Starter paths). Adds `SeatCountPolicy { Preserve, ActualUsage }` to `MigrationPath`; Teams 2019 uses `ActualUsage`, all other paths default to `Preserve`.
- **Shared plan-migration extensions** (`PlanMigrationExtensions`) — the packaged-source rule and the seat formula are factored into two extensions so the scheduler, the renewal quote, and renewal activation can't drift:
  - `IsPackagedMigrationSource(plan, policy)` — a source is Packaged when it's a flat bundle (`HasNonSeatBasedPasswordManagerPlan()`) **or** its path is `ActualUsage`. Replaces the same expression previously duplicated across `PriceIncreaseScheduler`, `UpcomingInvoiceHandler`, and `SubscriptionUpdatedHandler`.
  - `ResolveMigratedSeatCount(plan, occupied, purchased)` — the single billed-seat formula: `Max(1, occupied)` for a flat bundle, otherwise `occupied < base ? occupied : purchased`. Guards its inputs and throws `ArgumentException` for a Scalable source (`BaseSeats == 0`), which must preserve its line-item quantity rather than resolve from usage.
- **Price mapping** (`OrganizationPlanMigrationPriceMapper`) — a guarded Packaged → Scalable case maps the flat base price (`StripePlanId`) to the target per-seat price; the not-null guard keeps a `null == null` misfire off Scalable sources.
- **Phase 2 schedule** (`PriceIncreaseScheduler`) — for a Packaged source, the base line and the seat-overage line (both map to the same target seat price) collapse into a **single** seat line billed at `ResolveMigratedSeatCount`. Storage and other items pass through unchanged.
- **Renewal email** (`UpcomingInvoiceHandler`) — reuses `BusinessPlanRenewal2020MigrationMail`; the quoted seat count comes from the same `ResolveMigratedSeatCount`, so it matches what Phase 2 bills (occupied for < 5, purchased for ≥ 5).
- **Renewal activation** (`SubscriptionUpdatedHandler`) — source detection uses `IsPackagedMigrationSource` and selects the **base** price for packaged sources (so a sub-5 Teams 2019 org, which has no overage line, is still recognized when the schedule advances). The `organization.Seats` reconciliation (PM-39562, #7869) also fires for Teams 2019 — reconciling `Seats` to the billed quantity.

### Notes for reviewers

- **`MaxAutoscaleSeats`** is preserved by construction — `ChangePlan` doesn't touch it and the reconciliation writes only `Seats`.
- **`ResolveMigratedSeatCount` guard** — Scalable sources (`Preserve` policy) never reach the extension via the call-site gate today; the `ArgumentException` is a backstop so a future caller or a misconfigured `ActualUsage` path fails loudly instead of silently billing the wrong quantity.

- On the org Manage Subscription page, while a Packaged source (Teams 2019 / Teams Starter) is still in Phase 1 (before the scheduled migration applies), there's a plan mismatch: the upcoming (post-migration) plan is what's shown, but the seat total is calculated from the current plan(Teams 2019). The two are out of sync, so the displayed total is wrong — e.g. Total: 5 × $24.00 = $120.00 / year, which pulls the current Teams 2019 seat price into a view that's otherwise presenting the upcoming plan.

  Scope / not a blocker for this PR:
  - The subscription summary price (server-fed via StripePaymentService.GetSubscriptionAsync, including this PR's Phase 2 quantity fix) is correct — only the Manage Subscription page's seat-total calculation is affected.
  - Pre-migration window only. Once the org cuts over to the single Scalable plan, the shown and calculated plans are the same again, so it self-resolves.


src/Core/Billing/Services/Implementations/StripePaymentService.cs
```diff
@@ -778,6 +778,9 @@ private async Task ApplySchedulePhase2DataAsync(Subscription subscription, Subsc
                         fallbackItem.Amount = phase2Item.Price.UnitAmount!.Value / 100M;
                         fallbackItem.ProductId = phase2Item.Price.ProductId;
                         fallbackItem.Name = phase2Item.Price.Nickname;
+                        // A packaged base line (qty 1) collapses onto a scalable seat line whose quantity
+                        // is the migrated seat count; carry it so the Phase 2 preview total is correct.
+                        fallbackItem.Quantity = (int)phase2Item.Quantity;
```
sbrown-livefront, line 783: 📝 @amorask-bitwarden ran into this while testing. The upcoming subscription info was wrong. This applies to team-starter and 2019. I wanted to make sure I didn't overlook any logic here.

amorask-bitwarden: This is correct - great catch.

amorask-bitwarden (approved):
Awesome work - great idea on `SeatCountPolicy` to cover the weird Packaged / Scalable hybrid.

PR #7868, opened 2026-06-25
[PM-39561] Copy ListResponseModel from Api to HttpExtensions

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39561
## 📔 Objective

With PAM we are storing endpoints in a separate dotnet library where they instead are imported into `Api`. We ran into an issue with `ListResponseModel` since it's currently in `Api`.

`ListResponseModel` is a general-purpose API response wrapper. Moving it and `ResponseModel` to the `HttpExtensions` library. Which lets lower-level libraries reference it without taking a dependency on the Api project.


amorask-bitwarden (dismissed):

src/HttpExtensions/ListResponseModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿// FIXME: Update this file to be null safe and then delete the line below
+#nullable disable
```
justindbaur: Can we fix this?

src/HttpExtensions/ResponseModel.cs
```diff
@@ -0,0 +1,22 @@
...
+/// Base class for API response models.
+/// </summary>
+public abstract class ResponseModel
+{
+    public ResponseModel(string obj)
+    {
+        if (string.IsNullOrWhiteSpace(obj))
+        {
+            throw new ArgumentNullException(nameof(obj));
+        }
```
justindbaur:
```suggestion
        ArgumentException.ThrowIfNullOrWhiteSpace(obj);
```

justindbaur (approved):

jrmccannon (approved):

PR #7874, opened 2026-06-26
[PM-39734] Support Minimal API endpoints in OpenAPI

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39734
## 📔 Objective

Our current Swagger implementation using Swashbuckle with `Startup.cs` does not support automatically detecting Minimal Api's. This implements a workaround which creates a custom `EndpointDataSource` which gets registered earlier and are therefore detected by the Swagger generation CLI.

- Add `Bit.HttpExtensions.StandaloneEndpointDataSource` and the `AddOpenApiEndpointDataSource` registration helper. The offline OpenAPI generator (dotnet swagger tofile) never runs the Configure pipeline where Minimal API endpoints are mapped via `UseEndpoints`, so it omits them. A single DI-registered `EndpointDataSource` composes every feature's mapping delegate and is gated to swagger generation only, so it never replaces the runtime composite `EndpointDataSource` that routing and link generation depend on.
- Teach `SwaggerGenOptionsExt.BuildOperationId` and `ActionNameOperationFilter` to derive the operation ID and action name from the endpoint name set via `.WithName(...)` when there are no controller/action route values, as is the case for Minimal API endpoints.
- Guarantee `BuildOperationId` never returns a null/empty operation ID: a Minimal API endpoint mapped without `.WithName()` now falls back to a deterministic id derived from the HTTP method and route template. Swashbuckle writes the selector's value straight onto `operation.OperationId`, so a null/empty id would collapse distinct endpoints together and abort spec generation via `CheckDuplicateOperationIdsDocumentFilter`.
- Add unit tests for `BuildOperationId` (including the route fallback), the `ActionNameOperationFilter` endpoint-name branch, `StandaloneEndpointDataSource` composition, and `AddOpenApiEndpointDataSource` gating/registration.

### Example Output

Example output of a minimal API.

```json
    "/access-requests/history": {
      "get": {
        "tags": [
          "AccessRequests"
        ],
        "operationId": "Pam_AccessRequests_GetHistory",
        "responses": {
          "400": {
            "description": "Bad Request",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/ErrorResponseModel"
                }
              }
            }
          },
          "404": {
            "description": "Not Found",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/ErrorResponseModel"
                }
              }
            }
          },
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/AccessRequestDetailsResponseModelListResponseModel"
                }
              }
            }
          }
        },
        "x-action-name": "GetHistory",
        "x-action-name-snake-case": "get_history",
        "x-source-file": "bitwarden_license/src/Commercial.Pam/Api/Endpoints/AccessRequestEndpoints.cs",
        "x-source-line": 20
      }
    },
```

### DIFF
--- a/src/HttpExtensions/StandaloneEndpointDataSource.cs
+++ b/src/HttpExtensions/StandaloneEndpointDataSource.cs
@@ -19,7 +19,7 @@ namespace Bit.HttpExtensions;
 /// <see cref="EndpointDataSource"/>, so a source-per-feature would let the last-registered hide the rest. See
 /// <see cref="EndpointDataSourceServiceCollectionExtensions.AddOpenApiEndpointDataSource"/>.
 /// </summary>
-public sealed class StandaloneEndpointDataSource : EndpointDataSource
+internal sealed class StandaloneEndpointDataSource : EndpointDataSource
 {
     private readonly EndpointDataSource _source;
 
--- a/src/HttpExtensions/StandaloneEndpointDataSource.cs
+++ b/src/HttpExtensions/StandaloneEndpointDataSource.cs
@@ -1,0 +1,52 @@
+﻿using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Primitives;
+
+namespace Bit.HttpExtensions;
+
+/// <summary>
+/// An <see cref="EndpointDataSource"/> that materializes Minimal API endpoints from one or more route-mapping
+/// delegates outside the request pipeline.
+///
+/// Hosts that use the Startup/<c>Configure</c> pattern map their Minimal API endpoints with <c>UseEndpoints</c>.
+/// The offline OpenAPI generator (<c>dotnet swagger tofile</c>) never executes <c>Configure</c>, so those
+/// endpoints are invisible to ApiExplorer/Swashbuckle and the generated spec omits them. Registering this source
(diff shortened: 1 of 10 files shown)


src/HttpExtensions/StandaloneEndpointDataSource.cs
```diff
@@ -0,0 +1,52 @@
...
+/// The offline OpenAPI generator (<c>dotnet swagger tofile</c>) never executes <c>Configure</c>, so those
+/// endpoints are invisible to ApiExplorer/Swashbuckle and the generated spec omits them. Registering this source
+/// in DI makes the same endpoints discoverable without the request pipeline, while <c>UseEndpoints</c> still picks
+/// them up at runtime.
+///
+/// A single instance composes every feature's mapping delegate: ApiExplorer injects one
+/// <see cref="EndpointDataSource"/>, so a source-per-feature would let the last-registered hide the rest. See
+/// <see cref="EndpointDataSourceServiceCollectionExtensions.AddOpenApiEndpointDataSource"/>.
+/// </summary>
+public sealed class StandaloneEndpointDataSource : EndpointDataSource
```
justindbaur: Can be made internal right?

justindbaur (dismissed):
One minor thing. I've also made tickets for moving to the new entrypoint style and to try out the built in open api doc generation.

justindbaur (approved):

PR #7876, opened 2026-06-26
[PM-38927] - Extract Organziation User Role Validation

jrmccannon:
## 🎟️ Tracking
[PM-38927](https://bitwarden.atlassian.net/browse/PM-38927)

## 📔 Objective
Consolidates logic for Organization User Aciton validation to single class. This will be consumed when checking if an org user can perform an action or if an org user can perform an action on another org user.

[PM-38927]: https://bitwarden.atlassian.net/browse/PM-38927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/OrganizationUserActionValidator.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// authorization needs the gate without a specific target. Returns valid, or a <see cref="MissingManagePermissionError"/>.
+    /// </summary>
+    public static async Task<ValidationResult<OrganizationUserManageMembersRequest>> ValidateCanManageMembersAsync(OrganizationUserManageMembersRequest request)
+    {
+        var isProvider = !IsAuthorizedByRole(request) && await request.IsProviderUserForOrg();
+
+        return IsAuthorizedByRole(request) || isProvider
+            ? Valid(request)
+            : Invalid(request, new MissingManagePermissionError());
+    }
```
eliykat: We can already do simple RBAC checks well at the controller layer via attributes. I don't think we need to duplicate it here. This would be limited to the escalation concern.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/OrganizationUserActionValidator.cs
```diff
@@ -0,0 +1,81 @@
...
+        var canManageResult = await ValidateCanManageMembersAsync(request);
+        if (canManageResult.IsError)
+        {
+            return Invalid(request, canManageResult.AsError);
+        }
+
+        // The acting user can manage members; confirm the target's role is within their authority. A provider
+        // user has Owner-level authority, so treat them as an Owner. The expensive provider lookup already ran
+        // inside the gate above, so we reuse the cheap synchronous role check here rather than repeating it.
+        var effectiveRole = IsAuthorizedByRole(request) ? request.ActingUserRole : OrganizationUserType.Owner;
```
eliykat: I think this could be simplified, and it's better to avoid implicit logic such as:
- if you're not an admin then you must be a provider based on the logic of a separate private method
- if you're a provider we're going to say you're an owner

I would just check the role first, then check provider status if required. If we are doing multiple provider lookups there are better ways to handle that - e.g. push it out to the caller to pass a memoized Func or just a concrete boolean value.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/OrganizationUserActionRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+/// target role, to decide whether the acting user may manage (or assign) that role without escalating privileges.
+/// </summary>
+/// <param name="TargetUserRole">The role being acted upon — an existing member's current role, or the new role being assigned.</param>
+public record OrganizationUserActionRequest(
+    OrganizationUserType? ActingUserRole,
+    Permissions? ActingUserPermissions,
+    Func<Permissions, bool> PermissionPicker,
+    Func<Task<bool>> IsProviderUserForOrg,
+    OrganizationUserType TargetUserRole)
+    : OrganizationUserManageMembersRequest(ActingUserRole, ActingUserPermissions, PermissionPicker, IsProviderUserForOrg);
```
eliykat: This is really flexible but I think it degrades the interface. Simplified:
- `OrganizationUser? actingUser`
- `OrganizationUser targetUser`
- `bool actingUserIsProvider`

Then if this is a service method - they can easily be passed as arguments.

eliykat (changes requested):
A couple of quick comments, but my main feedback - my thinking was to re-introduce a service layer for smaller shared logic like this. e.g. `OrganizationUserValidationService` which can be called by multiple validators associated with different flows, and which would provide a bit more flexibility. That flexibility includes multiple params instead of request objects (where reasonable) which allows it to be easily called from different contexts without them having to share or construct the same request object (which adds a lot of boilerplate). 

i.e. `command -> validator (1:1 relationship with the command) -> domainValidationService (shared logic only, many:many relationship with validators)`.

Not sure if we discussed this - I might be mixing up my conversations. I will try to document some of my thoughts here this sprint.

The case against using dedicated validator classes for this level is:
- confusion between command-specific validators and reusable validators
- proliferation of classes (and associated boilerplate) for small, reusable bits of logic
- excessive request object mapping for them to be reusable

Let me know what you think and we can discuss further.

jrmccannon:
> Let me know what you think and we can discuss further.

That all makes sense and we did have the combo. Went down this path to see how it would look and got a little carried away. I've updated it to match the signature requested and hopefully cleaned up some other things.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/IOrganizationUserValidationService.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <item>Everyone else has no authority.</item>
+    /// </list>
+    /// Callers must separately confirm the acting user can manage members at all.
+    /// For a role change, call this for both the current and the new role.
+    /// </summary>
+    /// <param name="actingUser">The acting user's membership, or <c>null</c> if not a confirmed member.</param>
+    /// <param name="targetUser">The member being managed.</param>
+    /// <param name="actingUserIsProvider">Whether the acting user is a provider user, which grants Owner-level authority.</param>
+    /// <returns><c>null</c> when allowed, otherwise a <see cref="CannotManageTargetUser"/>.</returns>
+    Error? CanManage(OrganizationUser? actingUser, OrganizationUser targetUser, bool actingUserIsProvider);
```
eliykat: (non-blocking) I don't like `actingUserIsProvider` as an interface, this seems like something the method should be able to fetch for itself.

Generally I avoid using `CurrentContext` in `Core` because it leaks API concerns throughout the core layer, and I assume that's why you've avoided it here. But `CurrentContext.ProviderUserForOrg` is the only cached method we have for provider info, and technically speaking it's just a wrapper around the db call rather than API concerns. I think it would be OK to keep using that until we have something better. Open to other ideas though, and this is non-blocking in case I am being contrary here.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/IOrganizationUserValidationService.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Checks whether the acting user can manage the target user without escalating privileges:
+    /// <list type="bullet">
+    /// <item>Owners and provider users can manage anyone.</item>
+    /// <item>Admins can manage anyone except Owners.</item>
+    /// <item>Custom users with ManageUsers can manage Users and other Custom users.</item>
+    /// <item>Everyone else has no authority.</item>
+    /// </list>
+    /// Callers must separately confirm the acting user can manage members at all.
+    /// For a role change, call this for both the current and the new role.
```
eliykat: I take your point about the role change, but if it's only relevant to 1 flow then maybe it doesn't need to be mentioned here. It could just be a comment in the update user flow itself.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/IOrganizationUserValidationService.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Checks whether the acting user can manage the target user without escalating privileges:
+    /// <list type="bullet">
+    /// <item>Owners and provider users can manage anyone.</item>
+    /// <item>Admins can manage anyone except Owners.</item>
+    /// <item>Custom users with ManageUsers can manage Users and other Custom users.</item>
+    /// <item>Everyone else has no authority.</item>
+    /// </list>
+    /// Callers must separately confirm the acting user can manage members at all.
```
eliykat: I'm not sure this is true, I know you had dedicated logic for this before and I asked you to take it out, but looking at the current implementation I still think this would return `false` if the user doesn't have admin permissions.

Maybe this would be clearer guidance, which I think is really what we want to say:

> This should be used in combination with an `AuthorizeAttribute` for the standard RBAC check on the controller endpoint.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/IOrganizationUserValidationService.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Bit.Core.AdminConsole.OrganizationFeatures.OrganizationUsers.OrganizationUserAction;
+
+/// <summary>
+/// Reusable organization-user authorization rules shared by command-specific validators.
+/// </summary>
+public interface IOrganizationUserValidationService
+{
+    /// <summary>
+    /// Checks whether the acting user can manage the target user without escalating privileges:
```
eliykat, line 12: Maybe also good to note:

> This will allow Owners to manage Provider Users, which is suitable for most organization-level concerns. If your operation affects the Provider User as a Provider User you may not want to allow this.

I am thinking of password reset where owners aren't allowed to reset providers' passwords, because it would allow escalation via account takeover. I don't think we need to cover that logic here, but it's worth calling out.

eliykat (commented):
LGTM, a few minor comments/feedback, then good to go.

src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/OrganizationUserAction/OrganizationUserValidationService.cs
```diff
@@ -18,9 +21,16 @@ public class OrganizationUserValidationService : IOrganizationUserValidationServ
...
-            : new CannotManageTargetUser();
+        if (authorizedByRole)
+        {
+            return null;
+        }
+
+        // Provider users aren't org members but hold Owner-level authority. Mirrors CurrentContext.ProviderUserForOrgAsync.
+        var actingUserIsProvider = (await providerUserRepository
+                .GetManyOrganizationDetailsByUserAsync(actingUserId, ProviderUserStatusType.Confirmed))
+            .Any(po => po.OrganizationId == targetUser.OrganizationId);
```
eliykat, line 32: Alas, I realised why you made this a parameter in the first place - because you want to call it in a loop for bulk user updates. It may have to be moved back to the parameter. But see how you go using it in practice.

eliykat (approved):

PR #7880, opened 2026-06-29
Add Seeder query to get email codes

bnagawiecki:
## 📔 Objective

- Add query to get email 2FA or user verification code for a specific user


util/Seeder/Queries/UserEmailTokenCodeQuery.cs
MGibson1, line 1: Given that you need the distributed cache for the product verification to work, it's better done through providing and relying on the `EmailTokenProvider` class than re-implementing here.

It also represents new requirements on the globalSettings provided to the seeder API. It would be nice to document required settings somewhere

MGibson1 (changes requested):

util/Seeder/Queries/UserEmailTokenCodeQuery.cs
bnagawiecki: `EmailTokenProvider` doesn't expose a method to retrieve a previously generated token for a user. It only allows token generation and validation. I wouldn't think that we would want to add that capability to it if the only usage of it is for the seeder and not in normal code paths?

MGibson1: Do you want a previously generated token? I thought the point was to skip email verification, not to validate we saved the token already sent.

bnagawiecki: Yes. The purpose is just to query and retrieve a previously generated token via an API, instead of needing to go find the email and retrieve it from there.

MGibson1: I think we're passing in the night here.

My claim is that generating a new token and invalidating the old one doesn't matter if you don't want to use the email at all.

MGibson1 (approved):
this is a fragility, but unlikely to actually break.

PR #7883, opened 2026-06-29
[PM-36062] refactor: Remove defer-price-migration-to-renewal feature flag

amorask-bitwarden:
## 🎟️ Tracking

Resolves https://bitwarden.atlassian.net/browse/PM-36062

## 📔 Objective

Removes the `pm-32645-defer-price-migration-to-renewal` feature flag (`FeatureFlagKeys.PM32645_DeferPriceMigrationToRenewal`), which is fully rolled out, along with its legacy flag-OFF code:

- **Promotes the deferred-migration path to unconditional.** Premium/Families price migration now always defers to renewal via `PriceIncreaseScheduler` (Stripe subscription schedules). The inline immediate-migration branches in `UpcomingInvoiceHandler` and the schedule-aware automatic-tax handling (`UpcomingInvoiceHandler`, `UpdateBillingAddressCommand`) are no longer flag-gated.
- **Ungates the three subscription-schedule blocks on cancel/reinstate** (`SubscriberService` cancel paths, `ReinstateSubscriptionCommand`), making them unconditional. These were gated `PM32645 || PM35215`; because PM32645 was at 100% the blocks already ran for any subscriber holding an active schedule, and `PM35215_BusinessPlanPriceMigration` is **not** enabled in production. Collapsing the gate to `PM35215`-only would therefore have dropped schedule release/recreate for personal (Premium/Families) subscribers, so the schedule's presence — not a flag — is the gate now, matching `PriceIncreaseScheduler.Release`. PM35215's own logic (business-plan schedule creation, etc.) is untouched, and the now-unused `IFeatureService` is removed from both types.
- **Deletes four orphaned legacy methods** left with no callers once the flag-OFF branches were removed: `UserService.CancelPremiumAsync`, `IOrganizationService.ReinstateSubscriptionAsync(Guid)`, and `IStripePaymentService.CancelSubscriptionAsync(ISubscriber, bool)` / `ReinstateSubscriptionAsync(ISubscriber)`.

No production behavior change — PM32645 was at 100% rollout, so the now-unconditional paths match what production already runs. No API, DTO, or database changes. Tests updated to drop the removed flag-OFF coverage and to re-gate the schedule tests on schedule/metadata presence rather than the flag.


src/Core/Billing/Services/Implementations/SubscriberService.cs
```diff
@@ -228,8 +228,7 @@ private async Task CancelSubscriptionImmediatelyAsync(
         SubscriptionCancellationDetailsOptions? cancellationDetails,
         Dictionary<string, string>? cancellingUserMetadata)
     {
-        if (featureService.IsEnabled(FeatureFlagKeys.PM32645_DeferPriceMigrationToRenewal) ||
-            featureService.IsEnabled(FeatureFlagKeys.PM35215_BusinessPlanPriceMigration))
+        if (featureService.IsEnabled(FeatureFlagKeys.PM35215_BusinessPlanPriceMigration))
```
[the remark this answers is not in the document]
amorask-bitwarden: Resolved here: https://github.com/bitwarden/server/pull/7883/commits/fe25ce39fb8f232cb55b4f309f6f422f75005d9f

src/Billing/Services/Implementations/UpcomingInvoiceHandler.cs
```diff
@@ -269,67 +269,10 @@ private async Task<bool> ScheduleFamiliesPriceMigrationAsync(
...
-                        {
-                            Id = seatAddOnItem.Id,
-                            Deleted = true
-                        });
-                    }
-                }
-
-                await organizationRepository.ReplaceAsync(organization);
-                await stripeAdapter.UpdateSubscriptionAsync(subscription.Id, options);
+                return true;
```
kdenney, line 252: Would you mind adding a comment here (and the other locations where you are early-returning true) to explain why we want to just return true here if `!scheduled`? I know you only removed the flag in this PR but I assume you know why and I realized while reading through this that I don't lol so a comment would be helpful. 😄

kdenney (dismissed):
Looks good. I have one non-blocking request but that's it.

src/Billing/Services/Implementations/UpcomingInvoiceHandler.cs
amorask-bitwarden: Good call - added here: https://github.com/bitwarden/server/pull/7883/commits/496e464915a7db7ff24598d6330b5310ac0b3791

JimmyVo16 (dismissed):

src/Billing/Services/Implementations/UpcomingInvoiceHandler.cs
kdenney: Thanks! Looks like you have broken integration tests too:

```/home/runner/work/server/server/test/SeederApi.IntegrationTest/Pipeline/RecipeOrchestratorIntegrationTests.cs(123,24): error CS7036: There is no argument given that corresponds to the required parameter 'LicensingService' of 'SeederDependencies.SeederDependencies(DatabaseContext, IMapper, IPasswordHasher<User>, IManglerService, ILicensingService)' [/home/runner/work/server/server/test/SeederApi.IntegrationTest/SeederApi.IntegrationTest.csproj]```

kdenney: Which is weird because you didn't edit those files lol. Maybe it just needs reran..I'll try that.

edit: nope, both unit and integration tests are failing. Might be a bad merge got into main.

kdenney (dismissed):

JimmyVo16 (approved):

kdenney (approved):

PR #7892, opened 2026-06-30
[PM-39210] fix: Block converting a Secrets-Manager-enabled org to a Business Unit Portal

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39210

## 📔 Objective

Converting an organization to a Business Unit Portal (BUP) is the only existing-org provider-attach path that is both unguarded against Secrets Manager and billing-destructive. `BusinessUnitConverter` gated only on Enterprise tier (the tier that carries SM) and never checked `UseSecretsManager`, so an SM-enabled org could be converted — after which the Password Manager line is swapped to the BUP price and the org's `GatewaySubscriptionId` is nulled, orphaning the Secrets Manager subscription line. The org could no longer self-serve its SM seat count.

This adds a `UseSecretsManager` guard to both validators in `BusinessUnitConverter`, mirroring the existing precedent in `ProviderService` (`ProviderService.cs:447-451`):

- **`ValidateInitiationAsync`** — appends a user-facing problem string ("...Please contact Customer Support to convert this organization to a business unit.") to the returned problems list, so the Admin operator sees an actionable message and no provider is created.
- **`ValidateFinalizationAsync`** — fails via the file's existing log-and-throw `Fail()` convention (`BillingException`), as defense-in-depth.

Both guards are placed **before** the Stripe `GetSubscription` call, so a disqualified org short-circuits without a wasted round-trip and well before any destructive mutation.

**Tests** (`BusinessUnitConverterTests.cs`):
- `InitiateConversion_OrganizationUsesSecretsManager_ReturnsError` — SM-on Enterprise org is blocked; asserts no provider is created.
- `FinalizeConversion_OrganizationUsesSecretsManager_ThrowsBillingException` — asserts the throw and that `GetSubscription` is never called (proving the short-circuit).
- Extended `InitiateConversion_ValidationFails_ReturnsErrors` to set `UseSecretsManager = true` on a non-Enterprise org, proving the SM gate is **independent** of the tier check (the accumulating path).
- Hardened the two `_Succeeds_` tests with `UseSecretsManager = false` to avoid `BitAutoData` flakiness now that the gate exists.

**Open question for Billing:** the guard keys off the `UseSecretsManager` flag (matching the `ProviderService` precedent), so it blocks any SM-enabled org even if there is no actively-billed SM line on the Stripe subscription. This is the safe default; if conversion should be permitted for orgs that have the flag set without a billable SM line, that is a follow-up scoping decision.


amorask-bitwarden (changes requested):
These are both valuable validations, but we should prevent the option from showing in the Admin Portal at all if the organization uses Secrets Manager in order to avoid confusion for the end user.

amorask-bitwarden (approved):
Looks good - thanks for making that change.

sven-bitwarden (approved):

PR #7878, opened 2026-06-29
[PM-39745] Scaffold PAM access-rule endpoints

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39745
## 📔 Objective

Add a minimal, self-contained slice of the PAM access-rule surface so client teams can start generating server-side bindings from the OpenAPI spec, ahead of the full PAM implementation landing.

- Add Commercial.Pam project (references Core + HttpExtensions only) and register it in the solution and Api.csproj (#if !OSS)
- Add AccessRuleRequestModel and AccessRuleResponseModel as standalone DTOs (no domain coupling), with docs on Name, Description and Conditions
- Map the five access-rule endpoints (GET all, GET, POST, PUT, DELETE) with the shared PAM endpoint chain; handler bodies are stubs that throw NotImplementedException, so only the wire contract is defined
- Wire AddEndpointsApiExplorer + AddCommercialPamServices + MapPamEndpoints into Startup
- Add Commercial.Pam.Test covering the endpoint contract (routes, names, methods, group, error responses, return types)

PR #7893, opened 2026-06-30
[PM-39798] Scaffold PAM access-request endpoints

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39798

## 📔 Objective

Extends the PAM OpenAPI scaffold to the **access-request** resource so client teams can generate bindings for the requester/approver surface ahead of the full PAM implementation, mirroring the access-rule slice in #7878.

- **Standalone DTOs (no PAM-domain coupling):** `AccessRequestDetailsResponseModel`, `AccessLeaseResponseModel`, `AccessRequestDecisionResponseModel`, `AccessDecisionRequestModel`, plus a standalone `AccessDecisionVerdict` enum so the wire contract carries the `0 = deny` / `1 = approve` values. These reproduce the field-level shape and docs of the eventual domain-backed models without referencing `Bit.Pam.*`, keeping the project's Core + HttpExtensions-only references intact.
- **Seven endpoints** mapped through the shared PAM endpoint chain (`WithPamDefaults`): `GET inbox`, `GET history`, `GET mine`, `GET {id}`, `POST {id}/decision`, `POST {id}/activate`, `POST {id}/revoke`. Handler bodies are stubs that throw `NotImplementedException` — only the wire contract (routes, names, methods, status codes, return types) is defined; behavior lands with the rest of PAM.
- **Wiring:** register `AccessRequestEndpointsHandler` in `AddCommercialPamServices` and add the `/access-requests` group to `MapPamEndpoints`.
- **Tests:** `AccessRequestEndpointsTests` locks the contract (routes/names/methods, internal doc group, `ErrorResponseModel` 400/404, handler return types). The existing `AccessRuleEndpointsTests` now filters by tag, since `MapPamEndpoints` serves more than one group.

Verified locally: `Commercial.Pam.Test` (28 passing), `dotnet format --verify-no-changes`, and `dotnet swagger tofile … internal` — the generated spec emits all seven `Pam_AccessRequests_*` paths, the four DTOs, and `AccessDecisionVerdict` as an integer enum (`Deny`/`Approve`).


bitwarden_license/src/Services/Pam/Api/Models/Response/AccessLeaseResponseModel.cs
```diff
@@ -0,0 +1,56 @@
...
+    public Guid RequestId { get; set; }
+
+    /// <summary>The cipher the lease grants access to.</summary>
+    public Guid CipherId { get; set; }
+
+    /// <summary>The collection the cipher belongs to.</summary>
+    public Guid CollectionId { get; set; }
+
+    /// <summary>The access rule that gated the cipher at grant time. Not tracked in v1.</summary>
+    public string? RuleId { get; set; }
```
abergs, line 31: I think we should finish our out-of-band discussion on RuleId/CollectionId

bitwarden_license/src/Services/Pam/Api/Models/Response/AccessLeaseResponseModel.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// <summary>The organization that owns the cipher.</summary>
+    public Guid OrganizationId { get; set; }
+
+    /// <summary>The user the lease was granted to (the original requester).</summary>
+    public Guid RequesterId { get; set; }
+
+    /// <summary><c>active | expired | revoked</c>.</summary>
+    public string Status { get; set; } = null!;
```
abergs: Should this be enum?

patriksvensson: We probably want to serialize it in the response to the client as a string, but the representation here should probably be an enum.

bitwarden_license/src/Services/Pam/Api/Models/Response/AccessRequestDecisionResponseModel.cs
```diff
@@ -0,0 +1,34 @@
...
+/// — the request's full decision log, oldest first.
+///
+/// For an automatic (access-rule) decision <see cref="Id"/>/<see cref="Name"/>/<see cref="Email"/> are null; for a
+/// human decision they carry the approver (name/email denormalized by the server, null only when the user could not be
+/// resolved).
+/// </summary>
+public class AccessRequestDecisionResponseModel
+{
+    /// <summary><c>human | automatic</c>.</summary>
+    public string DeciderKind { get; set; } = "human";
```
patriksvensson: Maybe this should be an enum as well?

bitwarden_license/src/Services/Pam/Api/Models/Response/AccessRequestDetailsResponseModel.cs
```diff
@@ -0,0 +1,82 @@
...
+    public Guid CollectionId { get; set; }
+
+    /// <summary>The organization that owns the cipher.</summary>
+    public Guid OrganizationId { get; set; }
+
+    /// <summary>The member who opened the request.</summary>
+    public Guid RequesterId { get; set; }
+
+    /// <summary><c>pending | approved | activated | denied | cancelled | expired</c>.</summary>
+    public string Status { get; set; } = null!;
```
patriksvensson: Maybe this should be an enum?

bitwarden_license/src/Services/Pam/Api/Models/Response/AccessRequestDetailsResponseModel.cs
```diff
@@ -0,0 +1,82 @@
...
+    public IEnumerable<AccessRequestDecisionResponseModel> Decisions { get; set; } = null!;
+
+    /// <summary>Set once an approved request has produced a lease.</summary>
+    public Guid? ProducedLeaseId { get; set; }
+
+    /// <summary>
+    /// The produced lease's status (<c>active | expired | revoked</c>), or null when no lease exists. The inbox uses
+    /// this to keep an ended lease out of the "active" group so it is not offered for revocation.
+    /// </summary>
+    public string? ProducedLeaseStatus { get; set; }
```
patriksvensson: Maybe this should be a nullable enum?

patriksvensson (commented):
In general, it looks good to me, but I think we should have a discussion about whether we should be using enums for statuses and similar fields.

bitwarden_license/src/Services/Pam/Api/Models/Response/AccessLeaseResponseModel.cs
Hinton: All request/response enums are ints in network.

patriksvensson (approved):

PR #7896, opened 2026-06-30
[PM-39557] Replace invite link key columns with an opaque Invite blob

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39557

## 📔 Objective

Store the invite link's cryptographic material as a single opaque `Invite` blob plus a `SupportsConfirmation` flag instead of the separate `EncryptedInviteKey`/`EncryptedOrgKey` columns, so the server stays a pure storage/transport layer and the blob's format can evolve without AC changes.


util/MySqlMigrations/Migrations/20260630144227_OrganizationInviteLinkInviteBlob.cs
```diff
@@ -0,0 +1,49 @@
...
+public partial class OrganizationInviteLinkInviteBlob : Migration
+{
+    /// <inheritdoc />
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        migrationBuilder.DropColumn(
+            name: "EncryptedOrgKey",
+            table: "OrganizationInviteLink");
+
+        migrationBuilder.RenameColumn(
```
r-tome, line 17: EF decided to rename this column instead of dropping and creating a new one. I guess there is no harm because this feature hasn't been rolled out yet.

src/Api/AdminConsole/Models/Request/Organizations/RefreshOrganizationInviteLinkRequestModel.cs
```diff
@@ -1,28 +1,20 @@
...
     [Required]
-    [EncryptedString]
-    public required string EncryptedInviteKey { get; set; }
-
-    /// <summary>
-    /// The organization key encrypted for the invite link. Currently unused; will be populated in a future stage.
-    /// </summary>
-    [EncryptedString]
-    public string? EncryptedOrgKey { get; set; }
+    public required string Invite { get; set; }
```
JimmyVo16, line 15: I think we need to include the boolean SupportsConfirmation in both the creation and refresh flows as well, because we need to track this data.

src/Sql/dbo/Tables/OrganizationInviteLink.sql
```diff
@@ -1,13 +1,13 @@
...
+    [Code]                 UNIQUEIDENTIFIER NOT NULL,
+    [OrganizationId]       UNIQUEIDENTIFIER NOT NULL,
+    [AllowedDomains]       NVARCHAR(MAX)    NOT NULL,
+    [Invite]               NVARCHAR(MAX)    NOT NULL,
+    [SupportsConfirmation] BIT              NOT NULL,
+    [CreationDate]         DATETIME2(7)     NOT NULL,
+    [RevisionDate]         DATETIME2(7)     NOT NULL,
     CONSTRAINT [PK_OrganizationInviteLink] PRIMARY KEY CLUSTERED ([Id] ASC),
     CONSTRAINT [FK_OrganizationInviteLink_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]) ON DELETE CASCADE
 );
```
mkincaid-bw, line 13: The net result of the migration script will add these columns at the end, so the table definition will look like this:
```sql
CREATE TABLE [dbo].[OrganizationInviteLink]
(
    [Id]                   UNIQUEIDENTIFIER NOT NULL,
    [Code]                 UNIQUEIDENTIFIER NOT NULL,
    [OrganizationId]       UNIQUEIDENTIFIER NOT NULL,
    [AllowedDomains]       NVARCHAR(MAX)    NOT NULL,
    [CreationDate]         DATETIME2(7)     NOT NULL,
    [RevisionDate]         DATETIME2(7)     NOT NULL,
    [Invite]               NVARCHAR(MAX)    NOT NULL,
	[SupportsConfirmation] BIT              NOT NULL,
    CONSTRAINT [PK_OrganizationInviteLink] PRIMARY KEY CLUSTERED ([Id] ASC),
    CONSTRAINT [FK_OrganizationInviteLink_Organization] FOREIGN KEY ([OrganizationId]) REFERENCES [dbo].[Organization] ([Id]) ON DELETE CASCADE
);
```

util/Migrator/DbScripts/2026-06-30_00_OrganizationInviteLinkInviteBlob.sql
```diff
@@ -0,0 +1,108 @@
+-- Replace EncryptedInviteKey/EncryptedOrgKey with an Invite blob column and a SupportsConfirmation flag
+
+-- Drop the old key columns
+IF COL_LENGTH('[dbo].[OrganizationInviteLink]', 'EncryptedInviteKey') IS NOT NULL
+BEGIN
+    ALTER TABLE [dbo].[OrganizationInviteLink]
+        DROP COLUMN [EncryptedInviteKey], [EncryptedOrgKey];
+END
+GO
```
mkincaid-bw: ⛏️ Can you please split these into two statements and drop only one column per `IF` statement?

```suggestion
IF COL_LENGTH('[dbo].[OrganizationInviteLink]', 'EncryptedInviteKey') IS NOT NULL
BEGIN
    ALTER TABLE [dbo].[OrganizationInviteLink]
        DROP COLUMN [EncryptedInviteKey];
END
GO

IF COL_LENGTH('[dbo].[OrganizationInviteLink]', 'EncryptedOrgKey') IS NOT NULL
BEGIN
    ALTER TABLE [dbo].[OrganizationInviteLink]
        DROP COLUMN [EncryptedOrgKey];
END
GO

File: util/Migrator/DbScripts/2026-06-30_00_OrganizationInviteLinkInviteBlob.sql [hunk 5]
```diff
@@ -0,0 +1,108 @@
...
+
+-- Clear throwaway rows so [Invite] can be added NOT NULL without a default (flag never enabled in production)
+DELETE FROM [dbo].[OrganizationInviteLink];
```
mkincaid-bw: These scripts need to be idempotent so we shouldn't have a blanket `DELETE` in there. The current `OrganizationInviteLink` table in prod is empty, so you can probably just remove it. 

Alternatively, you could add something like this to the script BEFORE dropping the `EncryptedInviteKey`/`EncryptedOrgKey` columns so you know that the table has not been modified to the new schema yet:

```sql
IF  COL_LENGTH('[dbo].[OrganizationInviteLink]', 'EncryptedInviteKey') IS NOT NULL
AND COL_LENGTH('[dbo].[OrganizationInviteLink]', 'EncryptedOrgKey')    IS NOT NULL
BEGIN
    DELETE FROM [dbo].[OrganizationInviteLink];
END
GO
```

util/MySqlMigrations/Migrations/20260630144227_OrganizationInviteLinkInviteBlob.cs
mkincaid-bw: Like the SQL migration, this will cause the table to end up with columns out of order from the definition defined in `src/Sql/dbo/Tables/OrganizationInviteLink.sql`, so I think these should be written so that the existing columns are dropped and then the new ones added at the end.

If you don't want to do that, you could have the SQL Server migration follow the EF pattern and drop `EncryptedOrgKey`, rename `EncryptedInviteKey` to `Invite`, then add `SupportsConfirmation`. It doesn't really matter except that the end-result of the tables should be consistent across environments.

mkincaid-bw (changes requested):
Some minor changes requested.

src/Api/AdminConsole/Models/Request/Organizations/RefreshOrganizationInviteLinkRequestModel.cs
r-tome: Good call! I have updated both flows to accept that property although the commands have it hardcoded as `false`. That is so that when we eventually ship auto-confirmation admins will have to manually enable it.

r-tome:
@mkincaid-bw thanks for the review and suggestions! I have followed your suggestion to update sql to match EF and rename the existing column and add the new one at the end.

util/Migrator/DbScripts/2026-06-30_00_OrganizationInviteLinkInviteBlob.sql
r-tome: Resolved by the migration script redesign. Now we drop `EncryptedOrgKey` and rename `EncryptedInviteKey` separately

util/Migrator/DbScripts/2026-06-30_00_OrganizationInviteLinkInviteBlob.sql
r-tome: `DELETE` statement removed

src/Api/AdminConsole/Models/Request/Organizations/RefreshOrganizationInviteLinkRequestModel.cs
JimmyVo16: > That is so that when we eventually ship auto-confirmation admins will have to manually enable it.

Is the goal here to provide a safeguard? Could we just rely on the client to send the correct value, which would be false in Milestone 1? My concern is that we'd have to track this and remember to change it later.

mkincaid-bw (dismissed):
DB Changes LGTM but I just wanted to leave a note since this could be flagged by others:

This PR technically violates EDD with the column drop/rename. However, the `OrganizationInviteLink` table in prod is currently empty and from what I understand, the code has never been enabled in prod, so we should be fine. I also checked QA and the table is empty there as well.

src/Api/AdminConsole/Models/Request/Organizations/RefreshOrganizationInviteLinkRequestModel.cs
r-tome: Yes, that was the intent. I wanted to prevent that any link would have auto confirm enabled before we release it. Thinking back on it I think you're right that we should just accept whatever the clients sends and not have to update this later

r-tome:
> DB Changes LGTM but I just wanted to leave a note since this could be flagged by others:
> 
> This PR technically violates EDD with the column drop/rename. However, the `OrganizationInviteLink` table in prod is currently empty and from what I understand, the code has never been enabled in prod, so we should be fine. I also checked QA and the table is empty there as well.

@mkincaid-bw  I agree this violates EDD. Thanks for checking prod and QA and since those tables are empty and this feature hasn't been released no data is at risk.

mkincaid-bw (approved):

JimmyVo16 (approved):

PR #7899, opened 2026-06-30
[PM-20344] Throw BadRequestException when uploading attachments with …

mcamirault:
…same name as existing attachment

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20344

## 📔 Objective

This PR adds a BadRequestException when trying to upload an attachment with the same name as an existing attachment. This will be prevented client-side by [this PR](https://github.com/bitwarden/clients/pull/21596) for every client except the CLI, which will be covered by this server change.


mcamirault:
Nevermind, this won't work since the file name is encrypted

PR #7901, opened 2026-06-30
[PM-39805] fix: Expose SM migration-grace service accounts on subscription response

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39805

## 📔 Objective

After a Teams Starter 2023 → Teams (Monthly) migration, the Secrets Manager subscription response priced the permanent migration-grace service accounts as billable. The web vault's "Additional machine accounts" field showed 55 (30 grace + 25 paid) at $55/month, while the billing summary and Stripe correctly billed 25 @ $25/month — two conflicting machine-account costs for the same subscription.

The grace allotment (e.g. 30 = source SM baseline 50 − target baseline 20) is written to Stripe subscription metadata as `migration_grace_service_accounts` at migration, but was never surfaced on the subscription response, so the client could not exclude it.

This PR exposes that count on the subscription/organization response via `SubscriptionInfo.ServiceAccountGrace` (read through `SubscriptionExtensions.GetMigrationGraceServiceAccounts`), letting the client subtract the free, permanent grace from the billable additional-machine-accounts figure. The field then reports 25 @ $1 = $25, matching the billing summary and Stripe.

Unit tests added in `StripePaymentServiceTests` (Core.Test) and `OrganizationsControllerTests` (Api.Test).

### DIFF
--- a/src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
+++ b/src/Api/AdminConsole/Models/Response/Organizations/OrganizationResponseModel.cs
@@ -156,6 +156,7 @@ public class OrganizationSubscriptionResponseModel : OrganizationResponseModel
         bool hideSensitiveData) : this(organization, plan)
     {
         Subscription = subscription.Subscription != null ? new BillingSubscription(subscription.Subscription) : null;
+        SmServiceAccountsGrace = subscription.Subscription?.ServiceAccountGrace;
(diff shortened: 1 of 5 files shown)


sbrown-livefront (approved):
✅ Great fix

JaredScar (approved):

PR #7903, opened 2026-06-30
Seeder - Bump RustSdk to the latest release of sdk-internal

theMickster:
## 🎟️ Tracking

N/A

## 📔 Objective

We are quite a bit behind in our RustSdk project's alignment with the `sdk-internal` repo's crypto crate. 
Ensure that our toolchain is in alignment. 
Ensure that our `cargo fmt` is clean. 

## 🧪 Testing

Completed a full end-to-end test of the SeederApi and the SeederUtility by creating individual free, individual premium, and enterprise organizations.  Then, leveraged the forthcoming [Skill(querying-bitwarden-database)](https://github.com/bitwarden/server/pull/7763) to have Claude Code perform all of the read-only tests found in [verification.md](https://github.com/bitwarden/server/blob/main/util/Seeder/Seeds/docs/verification.md), and manual spot checking of the accounts and vaults via web app.

PR #7909, opened 2026-07-01
Refactor Skill(bump-rust-sdk) to align with current workflow & best practices for skill evaluation

theMickster:
## 🎟️ Tracking

N/A

## 📔 Objective

- Consolidated the skill into one lean file & split the frontmatter into `description` + `when_to_use` per the Claude Code skills spec. Less context loaded on every trigger, one authoritative procedure instead of an overview echoing a detail file.
- Purged and refactored outdated guidance that future RustkSdk bumps would have blindly followed 
- Fixed an over-trigger in `description`/`when_to_use` 
- Made the skill's first eval set** (`evals/`) — behavior cases + a lean README — closing the gap the Bitwarden AI Review Guidelines flag: _"the absence of evals is itself the finding."_
- Ran a rigorous 3× blind behavior benchmark (via `skill-creator`) that caught a regression & confirmed the skill's load-bearing value (details below).
- **Why it matters:** this skill drives the Seeder's Rust crypto layer; following its stale guidance would pin the _wrong_ `bitwarden-crypto` commit and misdirect future bumps. The eval set now protects that correctness. Zero-knowledge invariant is unaffected — RustSdk is `util/` **test** infrastructure, not production runtime.

## Why now?

We needed to bump the RustSdk project. See #7903 for more details and the quantifiable results of having a quality Claude Skill

## 🧪 Testing

Everything here is framed against our AI Review Guidelines. Their review gate for a net-new or changed LLM capability is: an **eval set of real cases**, a **recorded baseline**, and **additions traceable to a case that fails without them** — with **ablation** as the additive linkage and **three-tier blinding** for rigor. This skill had replaced a core method (the NPM→git-SHA mapping) with **zero** eval coverage. This PR closes that gap — and is intended as our reference example of doing evals correctly.

### The `evals/` folder

| File         | Purpose                                                                                                                                                                                                     |
| ------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `evals.json` | 5 behavior cases in the `skill-creator` schema — each a real `prompt`, an `expected_output`, and binary `expectations`. Cases 4 & 5 carry a `notes` field recording their **ablation outcome** (see below). |
| `README.md`  | Brief orientation: what the five cases cover and how to run them (`skill-creator` Benchmark mode). Deliberately bare — no fluff.                                                                            |

### The behavior benchmark

Run through `skill-creator`'s existing-skill benchmark flow: for each case, **with-skill** vs. **baseline (no skill)**, **3 independent runs per config**, graded by an **independent, config-blind** grader.

<details>
<summary><b>Methodology</b> — 24 runs + 4 blind graders, mutation-safe</summary>

- **24 runs** = 4 cases × {with-skill, baseline} × 3 reps. With-skill runs read the actual `SKILL.md` + references; baseline runs answer from their own expertise and may read repo code but **not** the skill.
- Every run was **read-only / advice-only** (no edits, no `cargo`/`git`/`npm`/network) — mutation-safe against the working tree. Each wrote its answer to an **opaquely-labeled** file (`A`–`F`; the with/baseline mapping withheld from the grader).
- **4 independent grader agents** (one per case) graded the six files against that case's `expectations`, **binary pass/fail, no partial credit** (per `skill-creator`'s `grader.md`), without knowing which file came from which config.
- **Case 1 (live NPM→SHA derivation) was skipped** — it needs the npm registry plus sibling `sdk-internal`/`clients` clones that aren't available in the eval environment. Stated, not silently dropped.

</details>

### Results

| Case                        | Checks | With-skill (3×)                    | Baseline (3×)  | Verdict                                                                                                           |
| --------------------------- | ------ | ---------------------------------- | -------------- | ----------------------------------------------------------------------------------------------------------------- |
| **2** — NPM→git-SHA mapping | 3      | **100%** (9/9)                     | **67%** (6/9)  | ✅ Skill win — baseline invents a wrong git-tag/release method; only the skill reaches the WASM-embedded-SHA read |
| **3** — MSRV / toolchain    | 3      | **100%** (9/9)                     | **67%** (6/9)  | ✅ Skill win — the "match the **MSRV**, _not_ sdk-internal's dev toolchain" nuance; one baseline rep inverts it   |
| **4** — `cargo update -p`   | 2      | **33%** (2/6) ⚠️ _trimmed variant_ | **100%** (6/6) | ⚠️ **Regression caught → instruction restored**                                                                   |
| **5** — breaking-change fix | 3      | **78%** (7/9)                      | **78%** (7/9)  | ⚖️ Borderline — **kept** (leans to earning its keep; see `evals.json` notes)                                      |

### What the benchmark caught (the point of doing it)

An earlier **single run** suggested case 4 (`cargo update -p`) was "unearned" — baseline recommends it anyway — so we trimmed the instruction. The **3× blind re-run on the trimmed skill exposed that as a regression:** without the explicit line, the skill's own Step 5 `cargo build` becomes a distractor and 2/3 with-skill reps recommend build-driven re-resolution, dropping **below** baseline (33% vs 100%). **The instruction was earned; we restored it** (the restored skill was not re-benchmarked at 3× — the earlier single run _with_ the instruction present had scored 2/2 on this case). The rigorous run also revised the single-run reads on cases 3 (parity → clear win) and 5 (clear win → parity). This is exactly why the Guidelines require a recorded baseline and ablation rather than a one-shot impression.

<details>
<summary><b>Per-case blind grades</b> (A/C/E = with-skill, B/D/F = baseline — mapping withheld from grader)</summary>

**Case 2** (E1 reject run-number · E2 WASM-SHA method · E3 don't endorse run-number)

| File     | E1  | E2  | E3  | Tally |
| -------- | --- | --- | --- | ----- |
| A (with) | ✓   | ✓   | ✓   | 3/3   |
| C (with) | ✓   | ✓   | ✓   | 3/3   |
| E (with) | ✓   | ✓   | ✓   | 3/3   |
| B (base) | ✓   | ✗   | ✓   | 2/3   |
| D (base) | ✓   | ✗   | ✓   | 2/3   |
| F (base) | ✓   | ✗   | ✓   | 2/3   |

**Case 3** (E1 channel→1.88.0 · E2 match MSRV not dev-toolchain · E3 rustc-too-old failure)

| File         | E1  | E2             | E3  | Tally    |
| ------------ | --- | -------------- | --- | -------- |
| A/C/E (with) | ✓   | ✓              | ✓   | 3/3 each |
| B/D (base)   | ✓   | ✗              | ✓   | 2/3 each |
| F (base)     | ✓   | ✗ (inverts it) | ✓   | 2/3      |

**Case 4** (E1 targeted `cargo update -p` · E2 bare-update-churn) — _skill trimmed for this run_

| File         | E1  | E2  | Tally    |
| ------------ | --- | --- | -------- |
| A (with)     | ✗   | ✗   | 0/2      |
| C (with)     | ✗   | ✗   | 0/2      |
| E (with)     | ✓   | ✓   | 2/2      |
| B/D/F (base) | ✓   | ✓   | 2/2 each |

**Case 5** (E1 compile-fail · E2 `make(SymmetricKeyAlgorithm::Aes256CbcHmac)` · E3 import enum)

| File       | E1  | E2              | E3  | Tally    |
| ---------- | --- | --------------- | --- | -------- |
| A/E (with) | ✓   | ✓               | ✓   | 3/3 each |
| C (with)   | ✓   | ✗ (guesses API) | ✗   | 1/3      |
| B/D (base) | ✓   | ✓               | ✓   | 3/3 each |
| F (base)   | ✓   | ✗ (guesses API) | ✗   | 1/3      |

The blind grader clustered A/C/E vs B/D/F cleanly on cases 2–4 with no config labels, and on case 5 the two flaky reps split one per config (with-skill `C`, baseline `F`) — evidence the blinding held and it wasn't rubber-stamping "with-skill = better."

</details>

<details>
<summary><b>Three-tier blinding</b> (per the Guidelines)</summary>

- **Subject blind** — prompts are phrased as genuine user asks and don't telegraph the expected method; case 2 actively proposes the _wrong_ method and checks the skill resists it.
- **Observer blind** — runs executed by independent subagents writing to opaque files; the orchestrator didn't coach mid-run.
- **Grader blind** — graders received files `A`–`F` with the with/baseline mapping withheld and graded each on its own substance.

</details>

<details>
<summary><b>Caveats</b> (stated, not buried)</summary>

- **3 reps is small.** Rates are indicative, not definitive; cases 4-with and 5 (both configs) show genuine run-to-run variance.
- Baseline runs could read repo code (not the skill); for case 5 they drew the exact API from OSS/training knowledge. `api-surface.md` is at `2.0.0`, so with-skill's case-5 answers came from the worked example, not the surface doc.
- Graders were config-blind but outputs can self-reference "the skill"; they were instructed to grade substance only, and the case-5 cross-config flakiness indicates they did.
- Benchmark artifacts (`grading.json` per case) were kept in `/tmp`, out of the repo.

</details>

### Ablation outcomes recorded in `evals.json`

Per the Guidelines, an instruction is only earned if removing it regresses a case. Both borderline cases now carry their verdict in-file:

- **Case 4 — EARNED (restored).** Removing `cargo update -p` regressed with-skill to ~33% (below baseline's 100%), because Step 5's `cargo build` misdirects without it. Load-bearing precisely because it counteracts the skill's own steer.
- **Case 5 — borderline, KEPT.** Parity in aggregate, but baseline only matched by recalling a version-specific API from memory (one rep guessed a _wrong_ constructor — a wrong key silently entering the Seeder is what this guards), while the with-skill answer comes from cheap on-demand reference content. Retained; revisit only if the example is later shown to mislead.


.claude/commands/bump-rust-sdk.md
```diff
@@ -17,23 +17,21 @@ Before starting, read these files to understand the current state:
...
 
 ## Execution
 
 Follow the skill's process in order:
 
 1. **Identify target** — Find the `@bitwarden/sdk-internal` NPM version at the release tag
-2. **Map to git SHA** — Query the GitHub Actions API for the publish workflow run number
-3. **Analyze breaking changes** — Compare old rev to new rev across the three crates, using the API surface reference
+2. **Map to git SHA** — Read the `main (<short-sha>)` baked into `bitwarden_wasm_internal_bg.wasm` in the published npm tarball, then `git rev-parse` it (not the GitHub Actions run number — that method is gone)
+3. **Analyze breaking changes** — Compare old rev to new rev for `bitwarden-crypto`, using the API surface reference
```
SaintPatrck: ♻️ Probably out of scope for this PR, but the fact we're updating this command to maintain parity with the corresponding Skill is a prompt smell. Every future change to the Skill may require this command to be updated also. We should try to strip the Skill instructions repeated here to avoid that drift and treat the Skill as the source of truth. Keep the command focused on what Skill(s) to invoke and how to handle their output. Not restating the Skill instructions.

SaintPatrck (dismissed):
Thanks for adding evals. This is a great start.

Left one comment about a prompt smell, but I think fixing it goes beyond the scope of this PR. Would love to see that addressed in a subsequent PR, if you agree it's out of scope for this one.

.claude/commands/bump-rust-sdk.md
theMickster: I looked this closer and we don't need the command. The skill is sufficient for our purposed with the Seeder + RustSdk. Removed in a0cb4510ea1b0a7345bb2ed3be99c8872e82f5fd

SaintPatrck (approved):

PR #7910, opened 2026-07-01
Dirt/pm 33527/multi delete task types

prograhamming:
## 🎟️ Tracking &  📔 Objective

This PR is stacked on top of this [pull request](https://github.com/bitwarden/server/pull/7783) in order to make the original ask more flexible with different types.


util/Migrator/DbScripts/2026-06-30_00_OrganizationDeleteByIdMultiTaskEnqueue.sql
```diff
@@ -0,0 +1,235 @@
...
+        [CreationDate] DATETIME2(7)     NOT NULL);
+END
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[Organization_DeleteById]
+    @Id UNIQUEIDENTIFIER,
+    @OrganizationDeleteTaskId UNIQUEIDENTIFIER = NULL,
+    @OrganizationDeleteTaskType TINYINT = NULL,
+    @OrganizationDeleteTaskCreationDate DATETIME2(7) = NULL,
+    @OrganizationDeleteTasks [dbo].[OrganizationDeleteTaskArray] READONLY
```
Banrion: ❓ Should `@OrganizationDeleteTasks` argument default to `NULL`?

⚠️ Remove `@OrganizationDeleteTaskId`, `@OrganizationDeleteTaskType`,  `@OrganizationDeleteTaskCreationDate` arguments and usage since they are replaced by `@OrganizationDeleteTasks`

util/Migrator/DbScripts/2026-06-30_00_OrganizationDeleteByIdMultiTaskEnqueue.sql
```diff
@@ -0,0 +1,235 @@
...
+IF TYPE_ID('[dbo].[OrganizationDeleteTaskArray]') IS NULL
+BEGIN
+    CREATE TYPE [dbo].[OrganizationDeleteTaskArray] AS TABLE (
+        [Id]           UNIQUEIDENTIFIER NOT NULL,
+        [TaskType]     TINYINT          NOT NULL,
+        [CreationDate] DATETIME2(7)     NOT NULL);
+END
+GO
+
+CREATE OR ALTER PROCEDURE [dbo].[Organization_DeleteById]
```
Banrion, line 7: ⚠️ Update existing migrations and sql scripts from https://github.com/bitwarden/server/pull/7783 rather than creating new ones since those have not been merged into main yet

util/Migrator/DbScripts/2026-06-30_00_OrganizationDeleteByIdMultiTaskEnqueue.sql
```diff
@@ -0,0 +1,235 @@
...
+            [TaskType],
+            [CreationDate],
+            [CreationDate]
+        FROM
+            @OrganizationDeleteTasks
+    END
+    -- Legacy single-task path. Retained so callers still running the previous server
+    -- version keep working during a rolling deployment; safe to remove once fully rolled.
+    ELSE IF @OrganizationDeleteTaskId IS NOT NULL
+    BEGIN
```
Banrion: ⚠️ Remove if statement since `@OrganizationDeleteTasks` replaces `@OrganizationDeleteTaskId`

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
```diff
@@ -161,9 +162,32 @@ BEGIN
...
+            [TaskType],
+            [CreationDate],
+            [CreationDate]
+        FROM
+            @OrganizationDeleteTasks
+    END
+    -- Legacy single-task path. Retained so callers still running the previous server
+    -- version keep working during a rolling deployment; safe to remove once fully rolled.
+    ELSE IF @OrganizationDeleteTaskId IS NOT NULL
     BEGIN
```
Banrion: ⚠️ Remove if statement since `@OrganizationDeleteTasks` replaces `@OrganizationDeleteTaskId` usage

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
```diff
@@ -2,7 +2,8 @@
     @Id UNIQUEIDENTIFIER,
     @OrganizationDeleteTaskId UNIQUEIDENTIFIER = NULL,
     @OrganizationDeleteTaskType TINYINT = NULL,
-    @OrganizationDeleteTaskCreationDate DATETIME2(7) = NULL
+    @OrganizationDeleteTaskCreationDate DATETIME2(7) = NULL,
+    @OrganizationDeleteTasks [dbo].[OrganizationDeleteTaskArray] READONLY
```
Banrion: ⚠️ Remove the three single task arguments in preference of the one multiple tasks argument

❓ Should `@OrganizationDeleteTasks` default to NULL

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -32,22 +33,37 @@ public OrganizationRepository(
...
 
-    public Task DeleteAndCreateDeleteTaskAsync(Organization organization, OrganizationDeleteTaskType taskType)
-        => DeleteInternalAsync(organization, taskType);
+    public Task DeleteAndCreateDeleteTasksAsync(Organization organization,
+        IEnumerable<OrganizationDeleteTaskType> taskTypes)
+        => DeleteInternalAsync(organization, taskTypes);
 
-    private async Task DeleteInternalAsync(Organization organization, OrganizationDeleteTaskType? taskType)
+    private async Task DeleteInternalAsync(Organization organization,
+        IEnumerable<OrganizationDeleteTaskType> taskTypes)
```
Banrion, line 43: ❓ Can/do we want to keep taskTypes optional?

Banrion (changes requested):

src/Infrastructure.EntityFramework/AdminConsole/Repositories/OrganizationRepository.cs
```diff
@@ -305,8 +314,25 @@ await dbContext.OrganizationConnections.Where(oc => oc.OrganizationId == organiz
...
+                var deleteTask = new Dirt.Models.OrganizationDeleteTask
+                {
+                    OrganizationId = organization.Id,
+                    TaskType = deleteTaskType,
+                    CreationDate = creationDate,
+                    RevisionDate = creationDate,
+                };
+                deleteTask.SetNewId();
+                await dbContext.OrganizationDeleteTasks.AddAsync(deleteTask);
+            }
```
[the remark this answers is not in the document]
prograhamming: Done in 12d38fef — the mapping is now a `Select` projection passed to `AddRangeAsync`, matching the Dapper implementation's existing `.Select(...).ToList()` approach.

src/Infrastructure.EntityFramework/Dirt/Models/OrganizationDeleteTask.cs
```diff
@@ -0,0 +1,16 @@
+﻿using AutoMapper;
+
+namespace Bit.Infrastructure.EntityFramework.Dirt.Models;
+
+public class OrganizationDeleteTask : Core.Dirt.Entities.OrganizationDeleteTask
```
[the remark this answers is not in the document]
prograhamming: Won't fix — sharing the Core entity's class name (in a different namespace) is the established convention for EF models throughout this codebase, e.g. `EF.Vault.Models.Cipher : Core.Vault.Entities.Cipher` and `EF.AdminConsole.Models.Organization : Core.AdminConsole.Entities.Organization`. Renaming this one model would make it inconsistent with that pattern and the AutoMapper profile conventions.

util/Migrator/DbScripts/2026-06-30_00_OrganizationDeleteByIdMultiTaskEnqueue.sql
prograhamming: Done in 31551b33 — the three single-task arguments and their legacy enqueue branch are removed; the proc now takes only `@Id` plus the TVP.

On defaulting to `NULL`: a TVP can't take an explicit default (assigning `NULL` to a table type is an operand type clash), and it doesn't need one — per the [CREATE PROCEDURE docs](https://learn.microsoft.com/sql/t-sql/statements/create-procedure-transact-sql#arguments), a TVP omitted at call time passes an empty table, which the `IF EXISTS` guard already handles.

Note this migration file no longer exists — per the other thread it was folded into the base PR's migration, now named `2026-07-14_03_AlterOrganizationDeleteByIdEnqueueDeleteTask.sql`.

util/Migrator/DbScripts/2026-06-30_00_OrganizationDeleteByIdMultiTaskEnqueue.sql
prograhamming: Done in 31551b33 — this file is deleted and the TVP change is folded into #7783's existing migration. The stack's migrations were subsequently renamed (a7d65991) to sort after main's latest, so the combined script is now `2026-07-14_03_AlterOrganizationDeleteByIdEnqueueDeleteTask.sql`. Since this PR now edits migrations owned by #7783, the two need to merge together (or this one rebases after #7783 lands) — noted in the PR description.

util/Migrator/DbScripts/2026-06-30_00_OrganizationDeleteByIdMultiTaskEnqueue.sql
prograhamming: Done in 31551b33 — the legacy `ELSE IF @OrganizationDeleteTaskId IS NOT NULL` branch is removed; the TVP `INSERT ... SELECT` is now the only enqueue path. (This file was consolidated into `2026-07-14_03_AlterOrganizationDeleteByIdEnqueueDeleteTask.sql`.)

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
prograhamming: Done in 31551b33 — removed; the TVP `INSERT ... SELECT` is now the only enqueue path in the SSDT definition as well.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
prograhamming: Done in 31551b33 — the three single-task parameters are removed in preference of `@OrganizationDeleteTasks`.

On `= NULL`: TVPs can't be given an explicit default and don't need one — they're implicitly optional. Per the [CREATE PROCEDURE docs](https://learn.microsoft.com/sql/t-sql/statements/create-procedure-transact-sql#arguments): "If a procedure contains table-valued parameters, and the parameter is missing in the call, an empty table is passed in." The `IF EXISTS (SELECT 1 FROM @OrganizationDeleteTasks)` guard handles that case.

src/Infrastructure.Dapper/AdminConsole/Repositories/OrganizationRepository.cs
prograhamming: Kept it required for now: this method exists specifically to enqueue delete tasks, and its only caller (`OrganizationDeleteCommand`) always passes a non-empty list — the no-task path is already served by the `DeleteAsync(organization)` overload, which passes `[]` internally. Making `taskTypes` optional would blur the two intents. That said, it's a one-line change if you'd prefer the flexibility — happy to add a default.

Banrion: I'm good with keeping it!

JimmyVo16 (approved):

src/Sql/dbo/Dirt/Tables/Event.sql
```diff
@@ -30,3 +30,7 @@ GO
 CREATE NONCLUSTERED INDEX [IX_Event_DateOrganizationIdUserId]
     ON [dbo].[Event]([Date] DESC, [OrganizationId] ASC, [ActingUserId] ASC, [CipherId] ASC) INCLUDE ([ServiceAccountId], [GrantedServiceAccountId]);
 
+GO
+CREATE NONCLUSTERED INDEX [IX_Event_OrganizationId]
+    ON [dbo].[Event]([OrganizationId] ASC) WHERE [OrganizationId] IS NOT NULL;
+
```
mkincaid-bw, line 42: Merged PR [7690](https://github.com/bitwarden/server/pull/7690) adds a new column `SendId` and new index to this table as well, so your `Event.sql` is stale.

src/Sql/dbo/Dirt/User Defined Types/OrganizationDeleteTaskArray.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TYPE [dbo].[OrganizationDeleteTaskArray] AS TABLE (
```
mkincaid-bw: New datatypes [should not by created](https://contributing.bitwarden.com/contributing/code-style/sql/#user-defined-types). If you need to pass an object, JSON is the preferred type.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
```diff
@@ -1,8 +1,6 @@
 ﻿CREATE PROCEDURE [dbo].[Organization_DeleteById]
     @Id UNIQUEIDENTIFIER,
-    @OrganizationDeleteTaskId UNIQUEIDENTIFIER = NULL,
-    @OrganizationDeleteTaskType TINYINT = NULL,
-    @OrganizationDeleteTaskCreationDate DATETIME2(7) = NULL
+    @OrganizationDeleteTasks [dbo].[OrganizationDeleteTaskArray] READONLY
```
mkincaid-bw: Per my previous comment, this should be a `VARCHAR(MAX)`/`NVARCHAR(MAX)` and processed with `JSON_VALUE`/`OPENJSON` as needed.

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
```diff
@@ -9,8 +9,6 @@ BEGIN
 
     WHILE @BatchSize > 0
```
mkincaid-bw: I don't generally focus too much on the C# code but my Claude review helper noted this about this procedure (not sure if the design is intentional):

> The job/repository layer was correctly redesigned (bounded run budget, lease refresh per batch, explicit "bounded per call" contract), but the only implementation still calls the unchanged Event_DeleteManyByOrganizationId, which loops internally until the entire org's events are deleted (commandTimeout: 3600). A large org can still blow past the 10-minute lease and get reclaimed by a concurrent run.

mkincaid-bw (changes requested):

src/Sql/dbo/Dirt/User Defined Types/OrganizationDeleteTaskArray.sql
prograhamming: Done in 6c95b9c2 — the `OrganizationDeleteTaskArray` type is removed. `Organization_DeleteById` now takes `@OrganizationDeleteTasks NVARCHAR(MAX) = NULL` (a JSON array of `{ Id, TaskType, CreationDate }`) parsed with `OPENJSON ... WITH`, and the Dapper repository serializes with `JsonSerializer` like the other JSON-taking procs. The migration script also drops the type for anyone who ran the earlier revision.

src/Sql/dbo/Stored Procedures/Organization_DeleteById.sql
prograhamming: Done in 6c95b9c2 — the parameter is now `NVARCHAR(MAX) = NULL` and the tasks are read via `OPENJSON(@OrganizationDeleteTasks) WITH ([Id] UNIQUEIDENTIFIER, [TaskType] TINYINT, [CreationDate] DATETIME2(7))`.

src/Sql/dbo/Dirt/Tables/Event.sql
prograhamming: Done in 6c95b9c2 — `Event.sql` now matches main's post-#7690 shape (`SendId` column plus the filtered `IX_Event_OrganizationIdSendIdDate` index) with this PR's `IX_Event_OrganizationId` added after it. The diff against the stacked base branch will show the `SendId` lines as additions since that branch predates #7690, but the file content now agrees with main so the eventual merge won't clobber it.

src/Sql/dbo/Stored Procedures/Event_DeleteManyByOrganizationId.sql
prograhamming: Good catch — fixed in 6c95b9c2. `Event_DeleteManyByOrganizationId` now takes `@MaxRows INT = 50000` and stops once that cap is hit, returning the count so the job calls it again after refreshing the lease via its per-batch progress update; the Dapper `commandTimeout` drops from 3600 to 300 accordingly. The EF implementation had the same problem (`ExecuteDeleteAsync` over the whole org) and is now bounded to 1000 rows per call. A large org therefore never holds a single call longer than one bounded chunk, and the 4-minute run budget / 10-minute lease math holds.

src/Infrastructure.EntityFramework/Dirt/Configurations/EventEntityTypeConfiguration.cs
```diff
@@ -23,6 +23,11 @@ public void Configure(EntityTypeBuilder<Event> builder)
             index,
             e => new { e.ServiceAccountId, e.GrantedServiceAccountId });
 
+        builder
+            .HasIndex(e => e.OrganizationId)
+            .IsClustered(false)
+            .HasDatabaseName("IX_Event_OrganizationId");
```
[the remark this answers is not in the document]
prograhamming: This follows the repo's existing precedent: merged PR #7690 added `IX_Event_OrganizationIdSendIdDate` filtered on MSSQL but unfiltered on the EF side, with a comment explaining that MySQL doesn't support filtered indexes. In 6c95b9c2 I added the same explanatory comment to this index's EF configuration rather than introducing per-provider `HasFilter` branching for one index — keeping the three EF providers uniform matches how the codebase already handles this divergence.

Banrion:
Superseded by #8182

PR #7912, opened 2026-07-01
Relax cert check on development builds

justindbaur:
## 📔 Objective

Relates to: #7837 

I wanted to have a different error message for if the protection cert couldn't be found in the user cert store vs from not having enough settings for it to be retrieved at all but right now a lot of developers have a stub thumbprint of `______` from when we used to have to set that setting. But development environments don't need any data protection setup. I am going to be removing that stubbed value but I can also relax the exception to actually only happen when that unprotect cert is needed (non-development).


dereknance (approved):
I'm annoyed I didn't catch this too 🤦🏻

PR #7914, opened 2026-07-02
[PM-39783] Support policies enabled by default

eliykat:
## 🎟️ Tracking

Tools Team need the new Send policy enabled by default. This is not something we support today:

- PolicyQuery maps a missing policy row to the disabled state
- PolicyRequirements doesn't expose missing policy rows - factories only receive `PolicyDetails` for policies that are actually in the database (due to the db query using an `INNER JOIN`).

We don't want to actually insert rows, so we need to handle the null state in code. Also note that client-side enforcement will be handled by Tools Team and is out of scope.

This is a minimal solution to unblock Tools Team this sprint, without overhauling the domain or impacting other policies.

This is done by:
- adding a new `PolicyDefaultState = ENABLED | DISABLED` property to the `IPolicyRequirementFactory`
- PolicyRequirementsQuery uses this to pivot to a new db query, which LEFT JOINs on policies rather than INNER JOIN. It returns a `PolicyDetailsWithState` subclass so that we can disambiguate between the 3 states (enabled, disabled, null)
- PolicyRequirementsQuery then includes all `PolicyDetails` for both enabled and null policies

The benefit of this is that it doesn't introduce new types to downstream consumers: we don't need to refactor `IPolicyRequirementsFactory` or the `IPolicyRequirement` - they just receive additional `PolicyDetails` to process. The underlying db state is abstracted away, minimizing the blast radius.

`PolicyQuery` also leans on this factory setting to synthesize enabled-by-default policies. This is a new coupling with the Factory, but it means there's a single source of truth. In particular, this is used by the Admin Console policies page, so the default value will flow through when the admin is configuring policies.

### DIFF
--- a/src/Core/AdminConsole/Models/Data/Organizations/Policies/PolicyDetails.cs
+++ b/src/Core/AdminConsole/Models/Data/Organizations/Policies/PolicyDetails.cs
@@ -37,3 +37,20 @@ public class PolicyDetails
     public Permissions GetOrganizationUserCustomPermissions()
         => CoreHelpers.LoadClassFromJsonData<Permissions>(OrganizationUserPermissionsData);
 }
+
+/// <summary>
+/// A <see cref="PolicyDetails"/> that also carries the raw enabled state of the underlying policy row, including the
+/// case where no row exists. Used by the query path for policies that are enabled by default so that "no row" can be
+/// distinguished from "disabled".
+/// </summary>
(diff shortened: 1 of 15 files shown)


eliykat:
Not sure this will fully solve the problem it was intended to address - closing for now pending further communications with Tools Team over what they actually need.

PR #7915, opened 2026-07-02
Add OrganizationUserInviteQuery

bnagawiecki:
## 📔 Objective

- Add query to get a user's organization invite link


MGibson1 (approved):
Same fragility as other link queries, but we're accepting that for now.

PR #7916, opened 2026-07-02
Opt the Aspire project out of package locking

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39513

## 📔 Objective

Drop the packages.lock.json file for the aspire project since it pulls in a package that is arch dependent which is causing everyone that has to do a package lock file update to likely do that change and then have CI do it differently.


withinfocus (approved):
A necessary change for maintainability.

